PR #798, opened 2026-05-15
docs(uipath-maestro-flow): extract shared action-node structure

uisherif:
> ⚠️ For discussion — do not merge until ready. This is the first of six proposed PRs from a review of `uipath-maestro-flow`. Each is small and independent. Critique scope, framing, and content before merge.

## The problem

Coding agents using `uipath-maestro-flow` load more context than any single task requires, and a measurable share of that load is content the agent has already read upstream. Measured against `main`:

- 78 markdown files, 12,376 lines in the skill.
- A typical "build a flow with a trigger and an action node" task loads ~2,400 cumulative lines across ~10 files before the agent can act.
- The 24 plugin folders each pair a `planning.md` with an `impl.md`. The ~15 action-node `impl.md` files repeat the same boilerplate per node:
  - The standard node JSON skeleton (`id` / `type` / `typeVersion` / `display` / `inputs` / `outputs`).
  - The standard `outputs` block with `"source": "=result.response"` and `"source": "=result.Error"` — identical across all action nodes.
  - The Registry-Validation pattern (`uip maestro flow registry get <type>`).
  - The Adding / Editing cross-reference to `editing-operations.md`.

Across the action-node `impl.md` files, that boilerplate accounts for roughly 200–300 lines of repetition. The agent re-reads it for every plugin it touches.

## Why this fix

1. One canonical doc — `references/shared/action-nodes.md` — owns the standard JSON skeleton, the standard `outputs` block, the registry-validation pattern, and the Adding/Editing cross-reference.
2. Demonstrated on two plugins of different sizes. `script/impl.md` is small (96 lines) and `http/impl.md` is the largest action-node impl (221 lines). Migrating both shows the template works at both ends.
3. Other action-node plugins (rpa, connector, queue, transform, loop, delay, summarize, batch-transform, hitl, etc.) deliberately left out so this PR stays a reviewable proof-of-concept. If the template shape is agreed, the remaining migrations are mechanical and ship as follow-up PRs.

## Files changed

| File | Before | After | Δ |
|---|---|---|---|
| `references/shared/action-nodes.md` (new) | — | 64 | +64 |
| `plugins/script/impl.md` | 96 | 79 | −17 |
| `plugins/http/impl.md` | 221 | 196 | −25 |
| Total | 317 | 339 | +22 |

[...]

PR #800, opened 2026-05-15
docs(uipath-maestro-flow): compress duplicated rules in author/CAPABILITY.md

uisherif:
> ⚠️ For discussion — do not merge until ready. PR #3 of six from the same `uipath-maestro-flow` review. Independent of #798 and #799.

## The problem

`uipath-maestro-flow/references/author/CAPABILITY.md` is 143 lines, the biggest of the four capability indexes, and an agent loads it on every "create or edit a flow" task. Some content is restated word-for-word inside the file, and some is a long restatement of content the agent has already seen (or will see) in canonical home docs:

- SKILL.md line 75 ↔ this file's anti-pattern at L76: double-nested solution layout rule, restated verbatim (~140 words).
- Rule #15 (L36) ↔ anti-pattern L87: `outputs` block on action nodes, same examples.
- Rule #13 (L34), anti-pattern L86, anti-pattern L87: the `flow format` 96×96 fact stated three times within ~60 lines.
- Rule #12 (L33, ~250 words) and anti-pattern L95 (~140 words): the `=js:` prefix rule. Both restate `shared/node-output-wiring.md`, the canonical home, which is already linked from both sites.

## Why this fix

`author/CAPABILITY.md` should be a navigation/rule index, not a re-teaching ground. Where a rule has a canonical home (SKILL.md or a `shared/` doc), the capability index should state the rule in one line and link to the home. This:

1. Cuts always-loaded context for every authoring task.
2. Removes the agent's incentive to re-verify rules it has already internalized from upstream.
3. Keeps every rule and anti-pattern intact and accurate — only restatements move.

## What changes

- **Rule #12 (`=js:` prefix):** compressed from ~155 words to ~50 words + link to `shared/node-output-wiring.md` and `shared/variables-and-expressions.md`.
- **Anti-pattern L76 (double-nested solution):** collapsed to one sentence + link to SKILL.md rule #6.
- **Anti-pattern L86 (`flow format` skipped):** collapsed to one sentence preserving the MST-9061 cross-reference; removed the 96×96 restatement that duplicates rule #13.
- **Anti-pattern L87 (outputs block):** deleted; rule #15 covers it.
- **Anti-pattern L95 (`=js:` missing):** collapsed to one sentence + link to rule #12 and `shared/node-output-wiring.md`.

All other rules and anti-patterns unchanged. Rule numbering (#1–#16) is preserved so inbound links from other docs continue to resolve.

## Expected impact

| Measure | Before | After | Δ |
|---|---|---|---|
| Lines | 143 | 142 | −1 |
| Words | 2,774 | 2,526 | **−248 (−9%)** |
| Characters | 23,545 | 21,640 | −1,905 (−8%) |

Line count is misleading because each rule/anti-pattern is a single long markdown bullet; the savings show up in words and characters. The original PR draft estimated ~600 words; the actual diff lands closer to ~250 because some restatements carry unique context (e.g., the `flow format` anti-pattern keeps its MST-9061 reference) that I preserved rather than dropping.

More importantly: each compressed restatement removes a re-decision point. The agent currently sees the `=js:` rule three times in this file's neighborhood (SKILL.md rule #9 fragment, rule #12 here, anti-pattern L95 here) and once in `node-output-wiring.md`. After this PR, it sees it twice (once as rule, once as anti-pattern), both pointing at the canonical home.

## Validation

- All rules and anti-patterns kept; only restatements collapsed.
- `.maintenance/check-links.sh` and `check-anchors.sh`: no new failures introduced. Pre-existing failures in `summarize/impl.md`, `agent/{planning,impl}.md`, `batch-transform/impl.md`, `editing-operations-json.md`, and `evaluate/CAPABILITY.md` are unchanged.
- Inbound links from other docs that reference this file's rule numbers (#12, #15, etc.) continue to work — rule numbers are unchanged.

## Out of scope

Same out-of-scope list as #798 and #799.


rockymadden (approved):
Looks good to me.

PR #801, opened 2026-05-15
docs(uipath-maestro-flow): add shared connection-binding workflow doc

uisherif:
> ⚠️ For discussion — do not merge until ready. PR #4 of six from the same `uipath-maestro-flow` review. Independent of #798, #799, #800.

## The problem

Three plugins author the same connection-fetch workflow (list → ping → refresh-retry → STOP-and-ask) with cosmetic variation:

- `plugins/connector/impl.md` Step 1 — connector activities (adds `--folder-key`).
- `plugins/connector-trigger/impl.md` Step 1a/1c — triggers (two-stage; reads `byoaConnection`).
- `plugins/http/impl.md` Step 2 — managed HTTP in connector mode (the longest restatement, full STOP block inline).

The HTTP file alone restates the STOP-and-ask block in 16 lines that duplicate guidance the connector plugins reach via cross-references. There is no canonical home inside the skill for the common kernel — each plugin tells its own version and points downstream to the `uipath-platform` skill for IS-side rules.

## Why this fix

Create `references/shared/connection-binding.md` as the canonical home for the kernel — list + ping + refresh-retry + STOP block + records to carry forward. Each plugin links to it for the common parts and keeps only its plugin-specific divergences (BYOA flow on triggers, `--folder-key` on activities, manual-mode escape on HTTP).

Establishing the canonical home matters more than the immediate line savings — future action-node plugins that need connections (queue, etc. when authenticated) can link rather than re-author.

## What changes

- **New `references/shared/connection-binding.md` (~50 lines):** standard list + ping, refresh-retry, STOP-and-ask, records to carry forward, plus a "where plugins diverge" table.
- **`plugins/http/impl.md` Step 2:** the largest migration — replace the inline list/ping/refresh/STOP block with a link to the shared doc. HTTP-specific note retained: it's the only plugin where "Switch this node to manual mode" is a valid STOP fallback.
- **`plugins/connector/impl.md` Step 1:** trim the inline bash block; link to the shared doc; keep the activity-specific `--folder-key` requirement and the "If a connector key fails" recovery hint inline.
- **`plugins/connector-trigger/impl.md` Step 1a:** redirect the empty-list recovery to the shared doc; keep the trigger-specific framing ("triggers must use a real connection here").

Connector-trigger's Steps 1b/1c (the BYOA flow and trigger-objects detour) stay untouched — they're trigger-specific, not shared.

## Expected impact

| File | Before | After | Δ |
|---|---|---|---|
[...]

PR #802, opened 2026-05-15
docs(uipath-maestro-flow): add common-tasks.md master index

uisherif:
> ⚠️ For discussion — do not merge until ready. PR #5 of six from the same `uipath-maestro-flow` review. Independent of #798–#801.

## The problem

When an agent knows what it wants to build (e.g. "add a script node", "diagnose MST-9107", "publish to Studio Web"), it currently still has to load the 143-line `author/CAPABILITY.md` (or one of the other three capability indexes) just to find the right plugin or workflow doc.

The four `CAPABILITY.md` files are valuable for tradeoff decisions (Studio Web vs Orchestrator, inline agent vs published agent, decision vs switch), but they are heavyweight for the case where the agent already knows the journey.

## Why this fix

Add a single top-level `references/common-tasks.md` that maps the ~35 most-common named tasks directly to their target files. An agent that recognises its task in the list can skip the CAPABILITY layer and go straight to the relevant plugin / workflow / shared doc.

The CAPABILITY indexes stay as the home for exploratory navigation and tradeoff content — the new file is purely a shortcut.

## What changes

- **New `references/common-tasks.md` (~57 lines):** a table of named tasks → file paths, plus a short "When to use a `CAPABILITY.md` instead" footer.
- **`SKILL.md` Capability router section:** add a one-paragraph callout above the existing capability table pointing to `references/common-tasks.md`. Existing table left intact — both paths coexist.

## Expected impact

For a task the agent recognises (about 70% of typical journeys based on the audit), the read path shrinks from:

`SKILL.md (88) → author/CAPABILITY.md (143) → plugin doc`

to:

`SKILL.md (88) → common-tasks.md (57) → plugin doc`

A ~86-line saving on the routing layer per task, repeated for every recognized task.

For tasks not in the index, behavior is unchanged — the agent reads SKILL.md → the relevant CAPABILITY.md as before.

## Validation

- `.maintenance/check-links.sh`: no new failures introduced (11 pre-existing broken links unchanged).
- `.maintenance/check-orphans.sh`: clean. The new file is reachable from SKILL.md.
- All 35+ links in the index point at files that exist on `main`.

## Out of scope

Same out-of-scope list as #798–#801. The connection-binding row in this index intentionally points at `connector/impl.md` Step 1 rather than `shared/connection-binding.md` (which lives on #801's branch) to keep this PR independently mergeable. If #801 lands first, a follow-up can update the row.

### DIFF
--- a/skills/uipath-maestro-flow/references/common-tasks.md
+++ b/skills/uipath-maestro-flow/references/common-tasks.md
@@ -1,0 +1,56 @@
+# Common tasks — direct routing
+
+For the most-frequent named journeys, this index points straight at the file that handles the task. Use it to skip the four `CAPABILITY.md` indexes when you already know what you want to do. For navigation that explores tradeoffs, start from [SKILL.md](../SKILL.md) → the relevant `CAPABILITY.md`.
+
+## Index
+
+| I want to... | Go to |
+|---|---|
+| **Create a brand-new flow project** | [author/references/greenfield.md](author/references/greenfield.md) |
+| **Edit an existing flow** | [author/references/brownfield.md](author/references/brownfield.md) |
+| **Add a script node** | [author/references/plugins/script/impl.md](author/references/plugins/script/impl.md) |
+| **Add an HTTP node (Managed HTTP Request)** | [author/references/plugins/http/impl.md](author/references/plugins/http/impl.md) |
+| **Add a connector activity** | [author/references/plugins/connector/impl.md](author/references/plugins/connector/impl.md) |
+| **Add a connector trigger** | [author/references/plugins/connector-trigger/impl.md](author/references/plugins/connector-trigger/impl.md) |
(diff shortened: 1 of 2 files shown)


skills/uipath-maestro-flow/references/common-tasks.md
```diff
@@ -0,0 +1,56 @@
+# Common tasks — direct routing
+
+For the most-frequent named journeys, this index points straight at the file that handles the task. Use it to skip the four `CAPABILITY.md` indexes when you already know what you want to do. For navigation that explores tradeoffs, start from [SKILL.md](../SKILL.md) → the relevant `CAPABILITY.md`.
+
+## Index
```
rockymadden, line 5: I get this, but is there a reason we shouldn't integrate directly into SKILL.md?

rockymadden:
Closing due to age. Please feel free to reopen if needed.

PR #814, opened 2026-05-18
fix: Added advanced field option and offset support

AditiGoyalUipath:
1. Skills update to support advanced option in field like field length,max min value
2. Offset support in get API.

### DIFF
--- a/skills/uipath-data-fabric/references/records-query.md
+++ b/skills/uipath-data-fabric/references/records-query.md
@@ -70,7 +70,7 @@ uip df records query <entity-id> \
 }
 ```
 
-> `start`, `limit`, `offset`, and `pageSize` are **not** valid body keys — use `--limit` and `--cursor` CLI flags instead. The CLI translates them into the underlying offset-based API call.
+> Pagination is CLI-flag only: `--limit`, `--cursor`, `--offset`. Do not put `start`, `limit`, `offset`, or `pageSize` inside `--body` — the CLI translates these flags into the underlying offset-based API call.
 
 ### Operators
 
--- a/skills/uipath-data-fabric/references/records-query.md
+++ b/skills/uipath-data-fabric/references/records-query.md
@@ -15,6 +15,24 @@ Response: `{ TotalCount, Records, HasNextPage, NextCursor?, CurrentPage?, TotalP
 - Use `HasNextPage` to check if more records exist
 - Pass the `NextCursor` string value to `--cursor` to fetch the next page
 
+## Pagination
+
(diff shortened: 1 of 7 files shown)


skills/uipath-data-fabric/references/records-query.md
```diff
@@ -52,7 +70,7 @@ uip df records query <entity-id> \
 }
 ```
 
-> `start` and `limit` are **not** valid body keys — use `--limit` and `--cursor` CLI flags instead.
+> `start`, `limit`, `offset`, and `pageSize` are **not** valid body keys — use `--limit`, `--cursor`, and `--offset` CLI flags instead. The CLI translates them into the underlying offset-based API call.
```
SuphalAthlur: Wondering if the first part is a necessary clarification

skills/uipath-data-fabric/references/records-query.md
```diff
@@ -23,7 +41,7 @@ uip df records query <entity-id> \
   --output json
 ```
 
-Pagination for query also uses `--limit` and `--cursor` flags — not body keys.
+Pagination for query also uses `--limit`, `--cursor`, and `--offset` flags — not body keys.
```
SuphalAthlur, line 44: Is `--cursor` used for querying VDO entities alone?

SuphalAthlur (approved):
And do we have these guardrails enforced in the CLI command? Things like is the constraint applied permitted for the chosen field type etc.

skills/uipath-data-fabric/references/records-query.md
AditiGoyalUipath: No this is supported in SDK for cursor based pagination. User can use limit based(Which is supported in datafabric API) or cursor based pagination (supported via sdk)

AditiGoyalUipath:
> And do we have these guardrails enforced in the CLI command? Things like is the constraint applied permitted for the chosen field type etc.

Guardrails are enforced in typescript SDK which CLI invokes to perform any action.

skills/uipath-data-fabric/references/records-query.md
AditiGoyalUipath: Can you clarify this ?

SuphalAthlur: As in, can't we have something like this? Why do we need to mention properties that are invalid, we could directly provide the valid flags right?

> use `--limit`, `--cursor`, and `--offset` CLI flags ONLY. The CLI translates them into the underlying offset-based API call.

PR #833, opened 2026-05-18
feat(audit): /audit-verbs CLI-verb reachability pipeline + /lint-task axis

uipreliga:
## Summary

Adds a static-analysis pipeline that verifies every `uip <verb>` reference in this repo against the real CLI, with a per-PR enforcement gate that blocks doc drift before it ships.

Four layers:

1. **Catalog** — `scripts/build-uip-catalog.py` walks `uip --help --output json` recursively (top-level + plugin groups, after `uip tools install` for every available plugin) and writes `assets/uip-catalog-snapshot.json` (1115 verbs in uip 1.2.0). Marks broken/Click-style plugins and platform-specific tools (e.g. `rpa`, which needs Windows + Studio Helm and can't install on the Linux runner) as **unwalkable** so consumers downgrade those to *Uncertain* instead of false-positive *Stale*.

2. **Checkers**
   - `scripts/check-cli-verbs.py` — parses `command_pattern` regex AST (via `sre_parse`) in coder-eval task YAMLs, enumerates literal verb tokens after `uip`, matches against the catalog. Emits **High** (verb not in catalog) / **Medium** (retired per `.claude/rules/cli-renames.md`) / **Info** (pattern too dynamic).
   - `scripts/check-skill-verbs.py` — scans every `.md` under `skills/` for `uip <verb>` references in prose, code blocks, and tables. Tokeniser tuned to ignore placeholders (`<X>.flow`, `[FILE]`, `$VAR`), inline shell comments, dot-separated argument values (registry keys like `core.action.script`), and prose noise (English conjunctions, modal verbs, em-dash separators). Emits **Stale** (no catalog prefix at all) / **Stale (soft)** (valid prefix + suspicious tail) / **Uncertain** (under unwalkable group).
   - Both expose `--json` and `--report PATH` flags.

3. **Enforcement gate** (new in this PR — `.github/workflows/verb-gate.yml`)
   - Mirrors the `detect → matrix` pattern from `smoke-skills.yml` / `activation-gate.yml`: only skills whose markdown the PR touched are gated. Infrastructure edits (catalog snapshot, checker script, the workflow itself) fan out to every skill so CLI-wide drift gets caught the moment a refresh PR lands.
   - **Hard-fail** only on `Stale && matched_prefix == null` — verbs the catalog doesn't know at all. Findings with a valid prefix but a suspicious tail surface as `::warning` annotations on the PR (positional argument values mis-parsed as sub-verbs).
   - No `paths:` filter so this is wireable as a required check: `detect` short-circuits with `skip=true` for PRs that don't touch any skill markdown, and GitHub counts the skipped gated job as a pass.

4. **Surface**
   - New `/audit-verbs` slash command runs both audits and writes `tests/reports/cli-verb-audit.md` + `tests/reports/skill-verb-audit.md` (gitignored — generated on demand).
   - New **CLI verb reachability** axis on `/lint-task` (per-YAML check at PR time).
   - Nightly `.github/workflows/refresh-uip-catalog.yml` rebuilds the catalog and opens a PR when it drifts, **using `secrets.GH_PAT` so the bot-PR triggers required-check workflows** (PRs opened via `GITHUB_TOKEN` do not fire `pull_request:` events, which would let auto-merge bypass the gate).
   - Scoped auto-merge (`.github/workflows/automerge-catalog-refresh.yml`) — listens on `check_suite: completed`, filters open PRs by branch pattern `auto/refresh-uip-catalog-*` + label `automated`, re-verifies the branch pattern per-PR, and squash-merges when every check is SUCCESS/SKIPPED/NEUTRAL. **Does not use** GitHub's repo-wide `Allow auto-merge` setting — that capability is repo-wide and any human could opt in; this workflow enforces the scope in code, so only refresh PRs ever get auto-merged.
   - Docs: `tests/README.md` Authoring Workflow + `.claude/rules/test-writing.md` reference `/audit-verbs` (conditional, after `/lint-task`).

Bonus: `task-authoring.md` merged into `test-writing.md` so there's one rules file for test authoring instead of two.

## Seed entries for the 1.2.0 catalog

The branch ships against `uip 1.2.0`, which retired two verb families with substantial repo footprint:

| Retired         | Canonical            | Repo refs at time of writing |
|-----------------|----------------------|------------------------------|
| `flow`          | `maestro flow`       | ~25 files (100 leaf verbs moved under the `maestro` group) |
| `solution new`  | `solution init`      | 9 files                      |

Both are registered in `.claude/rules/cli-renames.md` so the task linter classifies them as **Medium** (retired) with the canonical replacement suggested, rather than **High** (unknown). The remaining doc footprint gets swept opportunistically as each touched file flows through the gate — same pattern documented in `verb-gate.yml`.

## Latent classifier bugs surfaced + fixed

Wiring the first rename entries surfaced two bugs in `scripts/check-cli-verbs.py` that the previously-empty rename table had been hiding:

1. **`load_renames` kept markdown backticks in dict keys.** A row formatted as `| `solution new` | `solution init` |` produced keys like `'`solution new`'` that no extracted verb could ever match. Stripped in the loader.
2. **`classify` preferred any catalog longest-prefix over a more-specific renames entry.** `solution new` (retired) was getting shadowed by the still-walkable parent group `solution` and mis-classified as reachable. Now: when the renames match is more specific than the catalog prefix, prefer the renames verdict.

Both bugs pinned with regression tests (`test_renames_entry_shadows_shallower_catalog_prefix`, `test_backticked_rename_cells_strip_to_bare_verbs`).

## What this catches

Audits against this branch's HEAD, with catalog at uip 1.2.0:

| Audit | Findings |
|---|---|
| Task YAMLs (Linux CI) | 0 High / 2 Medium / 97 Info |
| Task YAMLs (local macOS, no `rpa` plugin) | 6 High (all `rpa` patterns; pass on CI Linux via `--install-tools`) / 2 Medium / 97 Info |
| Skill docs (gate verdict) | **35 hard-fail** (Stale, no prefix match) / 266 soft-stale (warning) / 437 Uncertain |

The 35 hard-fails are the real doc-drift this PR's gate is designed to catch. The bulk are `uip flow ...` references that now need `uip maestro flow ...` (the cli-renames entry reclassifies these for the task linter, but the skill gate only knows Stale/Uncertain — those docs get swept the next time each file is edited).

## End-to-end refresh chain after this lands

1. 04:00 UTC cron fires `refresh-uip-catalog.yml`.
2. Builds catalog against `@uipath/cli@latest` on a Linux runner (platform-specific prefixes auto-marked unwalkable).
3. If the snapshot diffs, opens a PR via `GH_PAT` so `pull_request:` workflows fire.
4. Verb gate fans out to all 20 skills (because the snapshot changed); failures surface as per-line annotations.
5. `automerge-catalog-refresh.yml` listens for check-suite completion; merges + deletes branch only when every check is green for a PR matching the refresh branch pattern.
6. A red gate leaves the PR open with annotations attached — the next day's cron sees the existing PR and skips, so failures don't pile up multiple refreshes deep until a human resolves them.

## Deployment prerequisites

After merge, two settings need to be true on `main` for the chain to be fully hands-off:

- **`secrets.GH_PAT`** is owned by a Repo Admin (currently the case — the repo's two rulesets allow Repo Admin role to bypass approval, so the automerger can squash without a human review).
- **Squash merges allowed** in repo settings (Settings → General → Pull Requests).

The repo-wide `Allow auto-merge` setting is **not** required — the dedicated automerger workflow handles merge in code with an explicit branch-pattern filter.

## Tests

`tests/scripts/test_verb_checkers.py` covers 13 regression cases against `check-cli-verbs.py` / `check-skill-verbs.py`, including the false-positive classes fixed in this PR: `<X>.ext` placeholders, inline `#` shell comments, dot-separated argument values, prose noise filter, unwalkable-group classification, exit-code contracts, plus the two classifier bugs fixed during this round (renames-shadows-shallower-catalog-prefix, backtick-stripping in load_renames).


bai-uipath:
Substantive infrastructure that solves a real problem (silent CLI-doc rot). Catalog + checkers + per-skill gate + nightly refresh with scoped auto-merge — the factoring is clean and the failure modes are mostly handled well.

## Major issue — duplication (severity 7/10)

`uipath-maestro-flow` and `uipath-maestro-bpmn` already ship an identical 266-line live-CLI verb checker at `.maintenance/check-uip-commands.sh`. The PR doesn't acknowledge it. The skill-local checker queries `uip --help --output json` live (no stale catalog), and supports a per-line `<!-- uip-check-skip -->` opt-out that maestro docs rely on for intentional historical references (e.g. CLI version-comparison tables). After this PR, those exact lines hit the gate's hard-fail branch with no escape — a concrete usability regression.

**Requested change: port-and-delete.** Port `<!-- uip-check-skip -->` and `difflib.get_close_matches` suggestions into `scripts/check-skill-verbs.py`, then delete both `.maintenance/check-uip-commands.sh` files and update their READMEs to point at `/audit-verbs`. One source of truth, opt-out preserved.

## Overengineering worth trimming

- **`cli-renames.md` ships empty** (severity 5/10) — whole Medium-severity code path, table parser, report section, and a test, all gated on a file with zero rows. Defer until the first verb is actually retired. Saves ~80 LOC.
- **`sre_parse` is a deprecated CPython internal** (severity 5/10) — silenced via `warnings.catch_warnings()`. A simple `split` on whitespace handles ~95% of real `command_pattern` values; reserve AST walking for cases that need it.
- **`PROSE_NOISE` has grown to ~20 English words across two review rounds** (severity 4/10) — signals the tokenizer regex is too permissive. Tightening to `uip\s+[a-z][a-z0-9-]*\s+[a-z][a-z0-9-]*` (matching the skill-local checker) eliminates the maintenance burden.

bai-uipath (approved):

aoltean16 (approved):

.github/workflows/refresh-uip-catalog.yml
```diff
@@ -0,0 +1,105 @@
...
+      # carries divergent 1.0.0-alpha.* prereleases under the same scope; we
+      # want the stable public releases for the catalog. Same pattern as
+      # smoke-skills.yml / smoke-rpa-skills.yml.
+      - name: Configure npm to use public registry for @uipath
+        if: steps.check.outputs.exists != 'true'
+        run: npm config set @uipath:registry https://registry.npmjs.org/
+
+      - name: Install uip CLI
+        if: steps.check.outputs.exists != 'true'
+        run: npm install -g @uipath/cli@latest
```
baishalighosh: this will resolve it against published cli which will drift by 2 weeks , we need to catch regressions faster, so this should have a mode to check against latest alpha by default and if someone is interested against published cli they should do that

assets/uip-catalog-snapshot.json
```diff
@@ -0,0 +1,1110 @@
+{
```
baishalighosh, line 1: HAVE the snapshot suffix the cli version in the file name, then we can find deltas or gaps across diff versions of cli easily

baishalighosh (approved):

rockymadden (approved):

.github/workflows/refresh-uip-catalog.yml
uipreliga: Done in d0f37dd2 — the refresh workflow now installs `@uipath/cli@alpha` from `npm.pkg.github.com` using `secrets.GH_PAT`, matching the pattern already established in smoke-skills.yml. The catalog will track every merge to `cli/main` instead of lagging public npm by ~2 weeks.

Kept the manual-trigger path (`workflow_dispatch`) the same so anyone who wants to audit against a specific cut release can still `npm install -g @uipath/cli@<version>` locally and run `python3 scripts/build-uip-catalog.py --install-tools` themselves.

assets/uip-catalog-snapshot.json
uipreliga: Holding the line on the single-file design, with the reasoning written down for the record:

- **Version is already discoverable inside the JSON** — `jq -r .cli_version` and `jq -r .generated_at` both work on the live snapshot, and the daily refresh commit message embeds `($CLI_VERSION, $VERB_COUNT verbs)` so `git log` gives a chronological version timeline.
- **Cross-version diffs work out of the box via git** — `git log --follow -- assets/uip-catalog-snapshot.json` lists every refresh; `git diff <sha1> <sha2> -- assets/uip-catalog-snapshot.json` shows the verbs added/removed between any two versions. That's the same primitive the catalog refresher in d0f37dd2 relies on (`git diff --quiet` to decide whether to open a PR).
- **Single-path consumers stay simple** — `scripts/check-cli-verbs.py`, `scripts/check-skill-verbs.py`, `.github/workflows/verb-gate.yml`, and `/audit-verbs` all read `assets/uip-catalog-snapshot.json` unconditionally. Suffixing per CLI version would push version-discovery logic into every consumer ("glob the assets dir, find the newest, parse the version out of the filename...") for no extra capability.
- **Versioned archive is a follow-up if it pays off** — if cross-version analysis becomes a real workflow (rather than a `git diff` away), the right shape is probably a script that materializes `assets/archive/uip-catalog-<version>.json` on demand from the git history, not a checked-in fan-out. Happy to file as a separate ticket if you've got a concrete use case in mind.

uipreliga:
Thanks for the thorough review. Working through the four points:

## Major issue — duplication (7/10)

**`<!-- uip-check-skip -->` opt-out: shipped** in d0f37dd2. The marker is now honored by `scripts/check-skill-verbs.py` with the same semantics as the maestro `.maintenance/check-uip-commands.sh` (any occurrence on the line suppresses checking). Pinned with regression test `test_skip_marker_suppresses_line`. Across the 9 skills that were failing the gate, hard-fail findings dropped from 35 → 25, with `uipath-maestro-bpmn` going fully clean and `uipath-maestro-flow` from 8 → 1 — the suppressed lines are exactly the intentional historical references you flagged (version-comparison tables, fallback-prefix sentences).

**Port-and-delete: deferring to follow-up.** Keeping both checkers alive for now because they target different runtimes:

- `scripts/check-skill-verbs.py` runs in CI on a Linux runner against the catalog snapshot. Sub-second for the whole repo, no `uip` binary needed, lets the gate fan out across 20 skills in one matrix job.
- `.maintenance/check-uip-commands.sh` runs locally against a live `uip --help` query and is invoked from `.maintenance/check-all.sh` alongside the other pre-commit checks (link, anchor, plugin-pair, …). Coupling it to the catalog snapshot would change the maintenance workflow for the maestro skills.

The immediate usability regression (intentional opt-outs failing the gate) is resolved. Coordinating the delete + porting `difflib.get_close_matches` suggestions is a follow-up worth its own PR. Filed as a TODO; happy to take it on after this lands.

## Overengineering worth trimming

- **`cli-renames.md` ships empty (5/10) — resolved by catalog bump.** The branch now ships against `uip 1.2.0`, which retired two verb families with real repo footprint: `flow → maestro flow` (100 leaf verbs, ~25 referencing files) and `solution new → solution init` (9 referencing files). Both rows are seeded; the Medium-severity code path is now exercised by 2 task YAMLs (`tests/tasks/uipath-agents/lowcode/{batch_transform,deeprag}/e2e_enable_tool.yaml`). Wiring the table also surfaced two latent classifier bugs the empty table had been hiding (`load_renames` not stripping markdown backticks, `classify` preferring shallow catalog prefix over more-specific renames entry); both pinned with regression tests in c20d82a4.

- **`sre_parse` deprecation (5/10) — acknowledged, deferring.** CI runs Python 3.13 and the `warnings.catch_warnings()` block keeps the surface clean. Agree it'll need migration when Python 3.15 drops; the AST walker is the only consumer in `extract_verb_paths`, so the migration is contained. Filing as tech debt rather than blocking this PR — would rather not invent a custom splitter for the 5% of patterns that need real AST work just to dodge the deprecation warning.

- **`PROSE_NOISE` list growing (4/10) — fair smell, want to push back on the tightening.** The stricter `uip\s+[a-z][a-z0-9-]*\s+[a-z][a-z0-9-]*` regex you suggested would lose single-token verbs (`uip login`, `uip logout`, `uip mcp`, `uip skills`) by requiring two kebab-token segments. About a third of the catalog top-level entries are single-token, so the tighter regex moves the false-positive problem from "prose noise" to "single-verb commands ignored entirely." Open to a different shape that keeps the single-token case — e.g. permit `uip <token>` only when `<token>` is in the catalog top-level set, fall back to the noise filter otherwise. Worth a follow-up to design; I'd rather not shrink coverage in this PR.

PR #841, opened 2026-05-19
perf(maestro-case): batch caseplan.json writes per tasks.md section

jundayin:
## Summary

- Replaces the per-T-entry Read→Edit contract on `caseplan.json` with a **per-section Edit-batch**: one Read + N Edits (skip re-Read between siblings) + one validate per `tasks.md` section.
- CLI-gated sections (Phase 2 §4.6 non-connector `tasks describe`, Phase 3 §9.7 connector `case spec`) use **gather-then-write** — collect all CLI results first, then enter the Read → N-Edits → validate batch.
- **Edit, not whole-file Write**, inside a batch. T01 scaffold is the only legal Write — Edit preserves untouched fields and avoids the big-file-Write slowness the `Write` path exhibited.
- `tasks.md` (Phase 1) build stays per-entry — markdown appends are cheap and the incremental plan has user-review value.

## Why

The per-T-entry contract was the largest single perf lever in the skill: a 30-T-entry case ran ~60 `caseplan.json` tool calls (30 Reads + 30 Edits) plus validates. The "re-Read before next mutation" rule generated redundant Reads of a growing JSON file.

The new contract collapses Reads to one per section (~21 fewer Reads on a 30-T-entry case) while preserving the per-T-entry Edit cadence — so the user still sees ~30 incremental Edits flow through, with new TodoWrite-per-T-entry transitions for explicit progress.

## Expected impact

| Case size | Time saved | % reduction |
|---|---|---|
| 10 T-entries | ~30-60s | ~10-15% |
| 30 T-entries | ~2-4 min | ~15-25% |
| 50+ T-entries | ~4-7 min | ~20-30% |

Savings come from skipped Reads + reduced context bloat + fewer reasoning interludes. **Edit count, Edit size, and Edit time per call are unchanged** — only the redundant Reads between Edits go away.

## Files changed

- `SKILL.md` — Anti-pattern split: Phase 1 keeps per-entry prohibition; Phase 2+3 explicitly mandates per-section Edit-batch with all contract details.
- `references/case-editing-operations.md` — New canonical section `Per-section batch write contract — canonical`. Pre-flight Item 13 retargeted from per-plugin to per-section. Validation Cadence section + Anti-patterns updated. Whole-file Write now forbidden inside a batch.
- `references/implementation.md` — `Per-plugin execution` rewritten to reference the canonical contract. Step 9 + 9.7 restructured as gather-then-write. Steps 9.8, 10, 11 rewritten as per-section / per-(scope, target) / per-target Edit batches. Phase 2 + Phase 3 todo-seed sections gained per-T-entry sub-items guidance.

## Audit trail preservation

- **Same incremental Edit visibility** — each Edit still carries one T-entry's mutation; the transcript shows N Edits per section, same per-T-entry resolution as before.
- **NEW: TodoWrite per T-entry** — each section seeds per-entry todos that transition `pending` → `in_progress` → `completed` visibly during the batch. Old contract had only section-level todos.
- **Per-section validate** — section boundary is the audit checkpoint (already required by Pre-flight Item 13).
- **Recovery unchanged** — on any mid-batch interruption: re-Read both files, scan for next un-applied T-entry, resume. Section boundary is the recovery point.

## Test plan

- [ ] Run evals 1, 2, 3, 6, 7 from `tests/` against this branch
- [ ] Run a v19 fixture end-to-end on `main` and this branch; diff resulting `caseplan.json` files with prefixed-ID normalization (Stage_001, Stage_002, ...) — identical structural output confirms no field drops
- [ ] Run a v20 fixture end-to-end on both branches; same diff check
- [ ] Manual smoke: connector-heavy case (≥3 connector tasks) to exercise Phase 3 §9.7 gather-then-write
- [ ] Spot-check a Phase 2 publish: confirm Studio Web renders the skeleton correctly (catches frontend-render regressions validate doesn't catch)

## Out of scope

- Phase 1 `tasks.md` building stays per-entry — separate decision, separate future PR if ever needed
- Token-bloat fixes (v19/v20 dual recipes, SKILL.md Rule trimming, connector doc merge) — independent and additive, tracked in the saved perf-bottleneck analysis
- `get-connection` per-task re-run — needs a correctness audit before changing
- Hard-stop UX (Phase 5+6 merge into `Debug / Publish / Done`) — UX not perf


charlesliu9:
Check the "Primitive Operations / Write inconsistency — The Primitive Operations subsections in case-editing-operations.md (lines 217–266) still reference "Write" as the final mutation step. These were not updated in this PR but now conflict with the new "Edit only" contract. This is pre-existing but the PR makes it contradictory. Should be addressed as a follow-up at minimum." comment

charlesliu9 (approved):

PR #846, opened 2026-05-19
fix(troubleshoot) manifest stale cli paths

MarinRzv:
## Summary

  Aligns the `uipath-troubleshoot` scenario manifests, skill docs, and smoke
  validator with the `uip` CLI v1.0.4 subcommand restructure. CI smoke had
  been failing on 9 stale subcommand paths; the local Windows smoke was
  silently passing them because of two compounding bugs (wrong PATH binary
  + `.CMD` shim false-Success).

  ## What changed

  **Smoke validator** (`tests/tasks/uipath-troubleshoot/_shared/scripts/verify_manifest_commands.py`):
  - Walks the command tree level-by-level via JSON `--help` instead of a flat probe.
  - Validates path, flags, and positional count separately. Flags check against the leaf's `Options` plus a hardcoded set of CLI-wide globals (`--output`, `--log-level`, `--log-file`, `-h`, `--help`, `--help-all`) that aren't relisted in
   per-command help.
  - Catches the Windows `.CMD`-shim false-Success quirk: unknown subcommands return `Result=Success` with the parent's help. Discriminator: `Data.Command == path[-1]`.
  - Strict kebab-case tokenizer so subcommands like `credential-stores`, `element-executions` aren't misclassified as positionals.

  **Manifest rewrites** — subcommand restructure in v1.0.4:
  - `or assets list` → `resource assets list` (10 scenarios)
  - `or libraries list` → `resource libraries list`
  - `or releases list` / `releases get` → `or processes list` / `processes get`
  - `orchestrator credential-stores list` → `or credential-stores list`
  - `or folders list-current-user` → `or folders list` (default = user folders; `--all` for tenant-wide)
  - `is connections list <connector-positional>` → `is connections list`
  - `rpa get-errors` → `rpa validate`

  **Manifest deletions** — vestigial mocks no skill path reaches:
  - 4× `or robots list` rules in foreground scenarios (skill never instructs the agent to list robots; admin-tool's `robot-accounts list` is tenant-scoped, not folder-scoped — original semantic doesn't survive)
  - 2× `or resources --help` / `resources --help` probes
  - 3 duplicate-rule reconciliations after rewrites (with orphan canned files removed)

  **Skill doc fixes** — same v1.0.4 alignment for what the skill instructs the agent to run:
  - `or folders users list <fk>` → `or users list-in-folder --folder-key <fk>`
  - `--top <N>` → `--limit <N>` on jobs/queue-items list
  - `--queue-key` → `--queue-definition-key` on queue-items
  - Dropped `--folder-key` from `or jobs get` / `or processes get` (keys are tenant-unique)
  - Dropped `--use-studio` from `rpa validate` and `rpa activities get-default-xaml` (Studio-IPC flag removed)
  - `uip resources --help` → `uip resource --help` (singular tool name)
  - `rpa focus-activity` no longer exists at all — replaced the "highlight in Studio" instruction with a manual prompt to the user

  ## Local verification

  - Validator: **160 / 160** unique match signatures pass against installed uip CLI 1.0.4 (`python tests/tasks/uipath-troubleshoot/_shared/scripts/verify_manifest_commands.py --uip 'C:\Users\<u>\AppData\Roaming\npm\uip.cmd'`).
  - E2E smoke on a representative rewritten scenario (`getasset-permission-denied`): score **1.000** (skill_triggered 1.00, llm_judge 1.00), status SUCCESS. Agent correctly invoked `uip resource assets list --folder-key …` via the new
  manifest match.


dmorosanu (approved):

costin-uipath (approved):

tests/tasks/uipath-troubleshoot/argument-null-exception/fixtures/mocks/responses/manifest.json
```diff
@@ -69,12 +65,12 @@
     {
       "pattern": "or processes list",
       "min": 0,
-      "description": "OPTIONAL — Agent may inspect the release to rule out caller-supplied In arguments (H3)."
+      "description": "OPTIONAL \u2014 Agent may inspect the release to rule out caller-supplied In arguments (H3)."
```
costin-uipath: are these kind of changes intended?

PR #856, opened 2026-05-19
test: docker sandbox for smoke tests

akshaylive:
## Summary

Migrate all 115 smoke tests to Docker driver for improved test isolation and consistency. Pass UiPath CLI authentication environment variables to container. Configure experiment defaults with env_passthrough_extra.

## Changes

- Migrate all smoke tests to docker sandbox driver
- Pass UIPATH CLI auth env vars (ENABLE_ENV_AUTH, AUTH_TOKEN, ORG/TENANT names/IDs) to Docker container
- Pass Bedrock model and task parallelism to container
- Add docker sandbox configuration to experiment defaults
- Fix docker permissioning for test execution

## Test Plan

- [ ] Run `make smoke` to verify all smoke tests pass with Docker driver
- [ ] Verify environment variables are properly passed to containers
- [ ] Check test execution logs for any permission or auth issues

🤖 Generated with Claude Code


uipreliga:
## Code Review Summary

Ran a multi-model review (Opus 4.7, Gemini-3, Codex). All three converged on the same critical issues.

### Critical (flagged by all three reviewers)

1. **Lost deterministic pass-rate gate — Linux smoke can now report green on real task failures.** `smoke-skills.yml:423-427`
   The previous step parsed every `task.json` and enforced `pass_rate >= 95%`. It was replaced with `if: steps.smoke.outcome == 'failure' || 'cancelled'`. But the `Run smoke tests` step is `continue-on-error: true`, and `coder-eval run` typically exits 0 when individual tasks fail their `success_criteria` (it aggregates and reports). Net effect: tasks can deterministically fail (`final_status != SUCCESS`) and the job still goes green as long as the LLM-reviewer score stays ≥ 0.7. The Windows-equivalent workflow (`smoke-rpa-skills.yml`) still has the deterministic gate — Linux now lacks parity. **Confirm `coder-eval run`'s exit-code semantics before merge; if it does exit 0 on partial failures, restore a `weighted_score` / `final_status` gate.**

2. **`defaults.sandbox` block may be silently dropped.** `tests/experiments/default.yaml:16-28`
   The installed `coder_eval` package's `ExperimentDefaults` model has no `sandbox` field, and the installed `sandbox.py` declares `driver: Literal["tempdir"]` only — so docker driver support and `defaults.sandbox` likely require a newer coder-eval than what's installed locally. CI installs a different version via `uv pip install --system .` in `.coder_eval/`, so this *probably* works in CI but isn't validated. If the schema isn't honored, `UIPATH_CLI_AUTH_TOKEN` etc. never reach the container and authenticated tasks will silently fail. **Confirm the CI-installed coder-eval version honors `defaults.sandbox`; consider adding a fail-fast schema-check step.**

3. **Intentional-failure debug task committed.** `tests/tasks/test-failure/test-failure.yaml`
   Tagged `[temporary]` and instructs the agent to `sleep 30` and `exit 1`. Tag filter (`--tags smoke`) should exclude it, but it must not ship to `main`. **Delete before merge.**

### High

4. **`smoke-rpa-skills.yml:385-389` — error downgraded to warning masks catastrophic runs.** Changing "No task.json files found" from `::error:: + sys.exit(1)` to `::warning:: + sys.exit(0)` means a smoke run that crashed before producing any task output now reports green. The path filter only fires this job when RPA code changed, so zero outputs ≠ "no tests selected" — it means the harness broke. **Revert, or detect the "no smoke tasks matched the glob" case upstream of coder-eval invocation.**

5. **Secret-exfiltration surface via artifact upload.** `smoke-skills.yml:355-374` + `tests/experiments/default.yaml:22-28`
   The upload now includes `/tmp/runs/**/artifacts` (entire tree), and the docker passthrough forwards a real bearer token (`UIPATH_CLI_AUTH_TOKEN`) into the container. If any task writes env or a CLI config cache under its artifacts dir, the token ends up in a workflow-run artifact downloadable by anyone with repo read access. GitHub `::add-mask::` only redacts log output, not file payloads. **Either confirm coder-eval scrubs artifacts, narrow the upload glob to known-safe filenames, or add an explicit scrub step.** Token TTL is short, but org/tenant IDs are long-lived.

### Medium

6. **`smoke-skills.yml:309` permission-fix step has inconsistent error handling.** `sudo chown -R … && sudo chmod -R 755 … 2>/dev/null || true` — the `|| true` covers only the trailing `chmod`, and `&&` short-circuits if chown fails. Asymmetric. Also assumes `sudo` is available (fine on hosted ubuntu-latest, breaks on self-hosted). **Split into two lines, apply `|| true` consistently to both, and add a comment explaining why this exists (Docker container writes root-owned files).**

### Low / Informational

7. **DRY — 113 per-task `driver: tempdir → docker` flips.** Mechanically correct given current schema uncertainty. If/when `defaults.sandbox.driver` becomes supported upstream, collapse to a single default and revert per-task lines. Reviewers agreed: not blocking.

8. **3 smoke-tagged tasks remain on tempdir** (`uipath-platform/users_import_robot_smoke.yaml`, `uipath-troubleshoot/smoke-manifest-commands/task.yaml`, `uipath-agents/antipattern_openai_agents_hitl/...`). If intentional (host CLI / pre_run hooks), add an inline comment in each.

9. **`tests/experiments/default.yaml:16-28`** — `docker:` subkey applies to all tasks, including the ~190 still on `tempdir`. Benign but confusing. One-line comment noting "applied only when driver: docker" helps future readers.

10. **`smoke-skills.yml:64` workflow-file path-filter expansion** — every workflow edit now triggers a 120-min full fan-out. Intentional for testing this PR; acceptable in steady state.

11. **`smoke-skills.yml:350-353` artifact cleanup** — `-path "*/artifacts/*"` constraint is correct and `-exec ... +` is right. Worth an inline comment ("these are bulky and not needed for triage").

### Positive Observations

- Splitting "Verify Docker is available" and "Build coder-eval Docker image" into discrete steps gives clean CI failure attribution.
- Routing output to `/tmp/runs` is the right call — bind-mounting into the checkout would create root-owned files that pollute git status.
- `Fail if tests failed` correctly catches `cancelled` so a 120-min timeout is not silently green.
- `if: always()` on all post-run steps preserves artifact-and-report-first ordering.

### Overall Assessment

The Docker plumbing (build, run-dir, permission-fix, cleanup) is the right shape and the 113 driver flips are mechanically clean. The blocking issues are not in the migration itself but in the **gating logic regression on Linux** (#1), the **untested schema assumption for `defaults.sandbox`** (#2), and the **debug task** (#3). Confirm `coder-eval run`'s exit semantics, decide whether to restore the deterministic gate, delete the test-failure task, and either narrow the artifact upload or confirm artifact scrubbing — then it's mergeable.

---
🤖 Multi-model review: Opus 4.7 + Gemini-3 + Codex

akshaylive:
1. will fix
2. you're hallucinating
3. intentional
4. intentional
5. intentional/required
6. that's fine

uipreliga:
## Code Review — `akshaya/docker`

3 reviewers (Opus 4.6, Gemini 3.1 Pro, GPT-5.3 Codex) ran in parallel. Findings deduplicated below.

### Critical

- **`.github/workflows/smoke-skills.yml:425,436` — Pass-rate gate reads dead path; will fail every run.** *(Opus + Gemini, verified manually)* Step `Enforce pass-rate threshold (>= 95%)` keeps `working-directory: tests` (line 425) but the Python heredoc reads relative `pathlib.Path('runs')` (line 436). Run output moved to `/tmp/runs/`, so `tests/runs/` is empty → `n=0` → `rate=0.0` → `sys.exit(1)`. The job will go red on every smoke run, not just regressions. **Fix:** change line 436 to `pathlib.Path('/tmp/runs')` and drop `working-directory: tests` (line 425), matching the sibling threshold step at line 389 which was already updated.

### High

- **`.github/workflows/smoke-skills.yml:213-218` — Docker image build is missing the index URL.** *(Opus + Gemini)* The `make docker-image` step passes `UV_INDEX_UIPATH_USERNAME`/`UV_INDEX_UIPATH_PASSWORD` (uv named-index credential convention). These bind only when a matching `UV_INDEX_UIPATH_URL` env var or a `[[tool.uv.index]] name = \"uipath\"` entry exists. The earlier install step (208-211) uses `UV_EXTRA_INDEX_URL=https://user:pass@.../simple/` instead — different convention. If the `.coder_eval` project doesn't declare a named `uipath` index, the build cannot authenticate. **Fix:** either also pass `UV_EXTRA_INDEX_URL` for consistency with the install step, or confirm/add the named-index declaration in `.coder_eval`.

- **`.github/workflows/smoke-rpa-skills.yml:387` — Zero-task gate softened from fail to pass.** *(all three reviewers)* Was `::error` + `exit 1`, now `::warning` + `exit 0`. Combined with `continue-on-error: true` on the run step, a job where no task ever executed (install crash, helm crash, glob mismatch, runner skew) now reports green. **Fix:** restore `sys.exit(1)`, or guard the soft path to cases where `task_globs` was legitimately empty.

- **`tests/tasks/test-failure/test-failure.yaml` — Intentional-failure debug scaffolding left in `tests/tasks/`.** *(Opus + Codex)* Tagged `[temporary]`, so `--tags smoke` excludes it from normal runs. But the file is under the canonical task discovery root — any tooling that ignores tags (lint, dashboards, future scripts) will pick it up. The 30s sleep also burns runner minutes on every infra-PR fan-out. **Fix:** delete before merge, or relocate to a non-discovered dir (e.g. `tests/debug/`).

### Medium

- **`tests/experiments/default.yaml:18-28` — `UIPATH_CLI_AUTH_TOKEN` passed into agent-driven sandbox.** *(Opus + Codex)* The Docker sandbox runs agent-authored code. Passing the CLI auth token and tenant IDs in means a prompt-injected or compromised task can exfiltrate them. Likely acceptable for an alpha smoke tenant, but worth a comment documenting the trust boundary, or scoping the token narrower.

- **`.github/workflows/smoke-skills.yml:368` — `/tmp/runs/**/artifacts` directory glob inflates uploads.** *(Codex + Opus)* Adding the whole `artifacts/` directory alongside the specific file globs uploads everything under it. Cleanup at 348 strips only `.venv` / `node_modules`. Other transient outputs (build caches, image dumps, downloaded packages) ship to GitHub artifacts and count against retention. **Fix:** narrow to specific file types, or expand cleanup before upload.

### Positive

- 116 task YAMLs flipped uniformly (`tempdir`→`docker`) — verified clean diff, no field bleed.
- Adding `^\.github/workflows/smoke-skills\.yml\$` to the infra-change path filter is correct — workflow edits now self-exercise.
- \`find ... -path \"*/artifacts/*\"\` cleanup expression is logically correct for matching nested `.venv`/`node_modules` under `artifacts/`.
- \"Verify Docker is available\" pre-step catches missing-docker before the build step.

### Overall

Solid migration intent, but **the pass-rate gate bug at smoke-skills.yml:425/436 will make every CI run fail** — that needs fixing before merge regardless of anything else. After that, drop the `test-failure.yaml` debug task, restore the zero-tasks fail-closed gate in the RPA workflow, and resolve the uv credential convention mismatch in the Docker build step.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) via /code-review (Opus 4.7 + Gemini 3.1 Pro + GPT-5.3 Codex)

akshaylive:
1. fixed
2. not required
3. intentional
4. it was intentional, but sure i deleted it
5. intentional
6. intentional

uipreliga:
## Code Review Summary

Three reviewers (Gemini-3.1-Pro, GPT-5.3-Codex, Opus 4.6) flagged largely overlapping issues. Findings consolidated below.

### Critical / High Issues

- **🔴 HIGH — Silent-pass regression on RPA gate.** `.github/workflows/smoke-rpa-skills.yml:387-389` — empty `task_jsons` was `::error + sys.exit(1)`, now `::warning + sys.exit(0)`. Combined with `continue-on-error: true` on the smoke step (line 298) and the `Fail if smoke step was cancelled` step only catching `outcome == 'cancelled'`, any catastrophic infra failure (Helm/CLI install fail, all tasks crash pre-task.json) now passes the required check green. **Flagged by all 3 reviewers.** Suggested direction: keep the hard fail, or only soft-warn when an upstream signal indicates "no tasks selected" rather than "tasks ran but produced no output."

- **🟠 HIGH — Stale documentation.** `tests/experiments/skill-comparison-playbook.md:55` still says "Use `sandbox.driver: tempdir` (all existing tasks already do this)" — after this PR, no tasks do on Linux smoke. New tasks copy-pasting from the playbook will drift.

### Medium Issues

- **🟡 Dual env-var list (DRY hazard).** `tests/experiments/default.yaml:18-28` vs the workflow `env:` block (`smoke-skills.yml:260-294`). Adding a new env var requires editing both files; forgetting `env_passthrough_extra` silently drops the var inside the container with no error. **Flagged by all 3.** Suggested direction: add a comment in `default.yaml` explicitly calling out the sync contract, or support a glob (`UIPATH_CLI_*`) in coder-eval.

- **🟡 UV credentials handed to docker build.** `smoke-skills.yml:213-218` exports `UV_INDEX_UIPATH_USERNAME/PASSWORD` as env to `make docker-image`. Without confirming `.coder_eval/Dockerfile` uses BuildKit `--mount=type=secret`, the creds may persist in image layers via `ENV` / `RUN pip install` URLs. **Flagged by Codex + Opus.** Follow-up: audit `.coder_eval/Dockerfile` before merge.

- **🟡 Cleanup `find` swallows all errors.** `smoke-skills.yml:348` — `2>/dev/null || true` hides legitimate problems (e.g., wrong path, permissions). **Flagged by all 3.** Suggested direction: drop the `2>/dev/null`, keep the `|| true`.

- **🟡 RPA task likely incompatible with docker.** `tests/tasks/uipath-rpa/uia_google_search.yaml` drives Chrome via `UIAutomation`. If the coder-eval image lacks Chrome + X server, this task can only validate-compile — but it's in the smoke set and now points at `driver: docker`. Confirm container ships browser/display, or carve this task out. Note: RPA tasks run on the Windows workflow which still uses tempdir, so the practical impact here depends on whether the Linux fan-out picks this task up.

### Low Issues

- **🟢 Cross-workflow drift.** Linux uses `--run-dir /tmp/runs`; RPA workflow keeps relative `tests/runs/`. Not wrong but undocumented. Add a one-line comment near `--run-dir` explaining why Linux diverges. **Flagged by Codex + Opus.**

- **🟢 RPA artifact upload lacks `**/artifacts` pattern** that Linux added. Minor consistency gap.

- **🟢 Docker Hub rate limit risk.** `smoke-skills.yml:202-205` pulls `hello-world` every run; subject to 100-pull/6h anonymous limit on shared runner IPs. Use `docker info > /dev/null` instead. **Flagged by Gemini + Codex.**

- **🟢 Self-trigger regex lacks inline rationale.** `smoke-skills.yml:64` — workflow file is now in the full-suite trigger regex; add an explaining comment.

### Positive Observations

- Mechanical 115-file `tempdir → docker` flip is clean; sampled YAMLs unaffected by semantic content.
- `--run-dir /tmp/runs` consistently applied across all three Linux gates (summarize, reviewer threshold, pass-rate) — no orphaned `tests/runs` reference left in `smoke-skills.yml`.
- Cleanup scope guard `-path "*/artifacts/*"` prevents nuking the runner's own venv — good defensive design.
- Self-trigger regex extension prevents the "edit workflow, see no smoke effect" footgun.
- `test-failure/test-failure.yaml` deletion is clean; no other repo references.
- Strong concurrency cancellation + `if: always()` artifact uploads preserve good post-mortem observability.

### Overall Assessment

The Linux docker migration itself is solid and well-scoped. The blocking concern is the RPA gate softening — it converts catastrophic CI failures into green checks on a required workflow, and was flagged independently by all three reviewers. Secondary blocker before merge: audit the `.coder_eval` Dockerfile to confirm UV credentials aren't baked into image layers.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code) — multi-model review (Gemini-3.1-Pro + GPT-5.3-Codex + Opus 4.6)

akshaylive:
Updated RPA and docker compatibility issues. Also updated documentation.

2026-05-19-152000-create-a-uipath-flow-project-named-bellevueweath.txt
```diff
@@ -0,0 +1,2188 @@
+ ▐▛███▜▌   Claude Code v2.1.144
```
bai-uipath: this looks like an accidental commit

tests/experiments/default.yaml
```diff
@@ -13,6 +13,20 @@ defaults:
     task_timeout: 900
     turn_timeout: 900
 
+  sandbox:
```
bai-uipath, line 16: can we do the docker setting on this experiment file instead of 100+ individual task files? will be painful to merge this otherwise.

bai-uipath (commented):
are all the smoke tests passing?

2026-05-19-152000-create-a-uipath-flow-project-named-bellevueweath.txt
akshaylive: whoops

tests/experiments/default.yaml
akshaylive: Yes but I want to break this into two. 1/ (this PR) - move smoke to docker, 2/ (next PR) split into `default-smoke.yaml` and `default-rpa.yaml` and clean up what you mentioned. Also need to make equivalent makefile changes and announce on channels about the change.

akshaylive:
> are all the smoke tests passing?

They are **very** similar (compared to the baseline, there's +/- 1 failed test).

bai-uipath (approved):
## Blockers for the nightly VM

### 1. Docker daemon — fixed
The `coder-eval-runner` VM had no docker installed. I installed Docker CE + enabled the daemon + added `azureuser` to the docker group. `docker ps` works under sudo now; will work passwordless next SSH session.

### 2. `daily.sh` doesn't build the image
`tests/README.md` (and the workflow) now require `make docker-image`, but `dashboard/scripts/ci/daily.sh` only does `npm install -g @uipath/cli@alpha` + `uv venv` setup. Without a step that builds `coder-eval-agent:latest`, every smoke task would fail at sandbox setup tomorrow. Either:
- patch `daily.sh` to run `make docker-image` after the venv block, or
- have the docker sandbox driver auto-build on first miss.

Also: `make docker-image` needs `UV_INDEX_UIPATH_USERNAME` as a build secret. It's commented out in the VM's `coder_eval/.env` (only `UV_INDEX_UIPATH_PASSWORD` is set).

### 3. Auth model mismatch (the big one)
The VM uses **file-based** auth — `dashboard/scripts/ci/refresh-auth.sh` runs on a separate timer, mints a ROPC token, and writes it to `~/.uipath/.auth`. The `tempdir` sandbox sees that file because it shares `$HOME` with the host.

`env_passthrough_extra` in this PR passes `UIPATH_CLI_ENABLE_ENV_AUTH`, `UIPATH_CLI_AUTH_TOKEN`, `UIPATH_CLI_ORGANIZATION_NAME/ID`, `UIPATH_CLI_TENANT_NAME/ID` — these are the **env-mode** auth vars, and **none of them are set on the VM**. So inside the container:
- file-based auth: no `~/.uipath/.auth` (different filesystem)
- env-based auth: vars are empty

Net result: all 115 smoke tasks fail with auth errors on tomorrow's run.

Three reasonable fixes — pick one:
- **(a)** extend `refresh-auth.sh` to also export `UIPATH_CLI_*` env vars to a sourced file, and have `daily.sh` source it before invoking coder-eval; then `env_passthrough_extra` works as written
- **(b)** volume-mount `~/.uipath/.auth` into the container in the docker sandbox driver
- **(c)** have the docker driver detect file-auth on the host and copy/translate it in

(a) is probably the smallest diff and aligns with how the CI workflow already does it.

akshaylive:
Agreed, `daily.sh` should be updated -- it's fairly straightforward. I also added the correct mount for `.uipath` folder in the current PR, so we are good:
```
extra_mounts:
    -~/.uipath:/.uipath:rw
```

PR #870, opened 2026-05-20
test(uipath-agents): e2e task for OpenAI Agents structured output

EusebiuJecan07:
## Summary

- Adds `tests/tasks/uipath-agents/openai_agents_structured_output/` with a YAML task and check script
- Covers a gap not tested by the existing `openai_agents_handoff` task: **single-agent** project with typed context (`Agent[ReviewInput]`) **and** structured output (`output_type=ReviewOutput`)
- Verifies scaffold → `uip codedagent init` → run lifecycle, enforces the factory-function pattern (Critical Rule C4), and confirms both input + output schemas are extracted into `entry-points.json`

## Test plan

- [x] Ran `skill-agent-coded-openai-agents-structured-output` locally and it passed (4/4, score 1.000)
- [x] YAML parses cleanly (`yaml.safe_load`)
- [x] Check script compiles cleanly (`py_compile`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #821, opened 2026-05-18
Move uip solution comands to uipath-solution skill

RaduAna-Maria:
## Summary

  Merges two skills and one slice of a third into a single unified skill, and re-homes implementation-task generation:

  - **New `uipath-solution`** — replaces `uipath-solution-design` and absorbs the `uip solution` lifecycle slice from `uipath-platform`. Two halves under one skill: **Design** (PDD → SDD authoring) and **Operate** (`uip solution
   init / pack / publish / deploy / activate`).
  - **`uipath-planner` owns task lists** — no longer just an elicitation/router. Reads SDDs and derives `<process>-tasks.md`; emits live `TaskCreate` calls. Two lanes: **Lane A** (PDD-driven via SDD handoff) and **Lane B**
  (non-PDD elicitation).
  - **`uipath-platform`** — drops all `uip solution` content and redirects to `uipath-solution`. Still owns auth, Orchestrator, resources, Integration Service, traces, licensing.

  ## Why

  - `uipath-solution-design` produced an SDD then asked the planner to derive tasks — but it also generated its own inline Implementation Plan. Two sources of truth, easy to drift. Tasks are now owned by `uipath-planner`, full
  stop. SDDs end with `## Next Steps` pointing at the planner.
  - `uip solution` operations were split across two skills (design authored, platform deployed). Users had to load two skills to ship a `.uipx`. They are one cohesive lifecycle and now live in one place.
  - Detection contract is explicit: every SDD template now writes a `## Planner Handoff` heading **and** a `<!-- planner-handoff:v1 -->` marker. Either signal is sufficient — redundant by design so a renamed heading doesn't
  silently break Lane A detection.

  ## Skill changes

  ### `skills/uipath-solution/` (new)

  - `SKILL.md` with nested Critical Rules / Workflow under each half. CLI surface probe (post-rename vs pre-rename `uip solution`) moved here.
  - `references/design/` — `sdd-generation-guide.md`, `pdd-analysis-guide.md`, `product-selection-guide.md`, `rpa-product-guide.md`, `package-selection-guide.md`, `tenant-library-search-guide.md` (moved from
  `uipath-solution-design`).
  - `references/operate/` — `solution-overview.md`, `develop-solution.md`, `pack-and-deploy.md`, `activate-and-manage.md`, `scenarios.md` + 6 scenario files (moved from `uipath-platform/references/solution/`).
[...]

PR #874, opened 2026-05-20
fix(uipath-solution): default Level 2 authoring mode to XAML silently

roalexandru:
## Summary

- During Phase 2 Step 2, the model fired `AskUserQuestion` for every RPA project and freelanced option labels like **"XAML (Studio Classic)"** — a brand name that appears nowhere in the repo and conflicts with the `targetFramework` axis (`Windows` / `Portable` / `Legacy`).
- Update the existing "Selection checklist" closing rule in `references/design/rpa-product-guide.md` so XAML is the **silent default** when fewer than 2 Coded C# signals match. The question now fires only when ≥2 signals match (Coded C# / Hybrid) or when signals are ambiguous. When it does fire, option labels are locked to `Coded C#` / `XAML` / `Hybrid` — no invented brand names.
- Lower the UI-share threshold from `>70%` to `>49%` so any majority-UI process biases toward XAML. Apply the new threshold consistently in both the canonical guide (Anti-pattern Trigger + Selection checklist) and `assets/templates/rpa-sdd-template.md` §13 callout.

Net diff: 2 files, +3 / -3.

## Test plan

- [ ] Run Phase 2 Step 2 against a UI-heavy PDD → no `AskUserQuestion` fires; §13 in the generated SDD recommends **XAML** with the silent-default justification.
- [ ] Run Phase 2 Step 2 against a REST-heavy PDD (≥2 Coded C# signals) → `AskUserQuestion` fires; options read exactly `Coded C#` / `XAML` / `Hybrid` with no parentheticals.
- [ ] Run against a Solution PDD with one UI-heavy RPA project and one data-heavy RPA project → one silent default, one question; never combined.
- [ ] `grep -rn "Studio Classic" skills/` returns only the new ban line in `rpa-product-guide.md` (the rule forbidding the term).
- [ ] `grep -rn ">70%" skills/` returns no results.
- [ ] `hooks/validate-skill-descriptions.sh` passes.

### DIFF
--- a/skills/uipath-solution/assets/templates/rpa-sdd-template.md
+++ b/skills/uipath-solution/assets/templates/rpa-sdd-template.md
@@ -459,7 +459,7 @@ Sub-type reference:
 
 > **For Master Project:** specify the mode per sub-project if they differ. A Dispatcher may be XAML while a Performer with heavy data logic is Hybrid.
 >
-> **Before recommending Coded C#:** verify at least two checklist items from the [RPA Product Guide § Selection checklist before recommending Coded C#](../../references/design/rpa-product-guide.md#selection-checklist-before-recommending-coded-c) are true. "Cleaner control flow over a UI loop" is **not** a sufficient justification — XAML already has Try/Catch + Retry Scope + For Each over UIA activities. If the process body is >70% UI automation with minimal data shaping, recommend **XAML** or **Hybrid**.
+> **Before recommending Coded C#:** verify at least two checklist items from the [RPA Product Guide § Selection checklist before recommending Coded C#](../../references/design/rpa-product-guide.md#selection-checklist-before-recommending-coded-c) are true. "Cleaner control flow over a UI loop" is **not** a sufficient justification — XAML already has Try/Catch + Retry Scope + For Each over UIA activities. If the process body is >49% UI automation with minimal data shaping, recommend **XAML** or **Hybrid**.
 
 **Recommendation:** <XAML / Coded C# / Hybrid>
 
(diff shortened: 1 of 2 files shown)


roalexandru:
## Investigation: `Run skill smoke tests` failure (for future reference)

Recording the diagnosis here so the next person doesn't re-investigate.

**Failure mode (reproducible across 2 runs on this PR):**

```
skill-solution-project-list   status=FAILURE   score=0.727
Criterion 'command_executed' FAILED: Matched 0/1 required commands
(pattern=/uip\s+solution\s+project\s+list/)
```

The smoke agent calls `uip solution init` and creates the manifest correctly, then drops step 2 of the prompt (`uip solution project list`).

**Pre-existing — not introduced by this PR:**

The same task failed with the identical score (0.727) and identical missing-criterion on PR #821 (the PR that introduced these uipath-solution operate tasks):
https://github.com/UiPath/skills/actions/runs/26100432093/job/76749659443
PR #821 still merged because it ran 36 tasks, so 35/36 = 97.2% cleared the 95% threshold.

**Why it blocks this PR specifically:**

`Detect changed skills` correctly scoped this PR to the 5 uipath-solution smoke tasks. With one deterministically broken task in a 5-task set, pass rate = 4/5 = 80% < 95%.

**Why this PR's diff cannot cause it:**

The edits are in `references/design/rpa-product-guide.md` (Level 2 authoring-mode rules) and `assets/templates/rpa-sdd-template.md` (SDD §13 threshold). Both are loaded by the **design half** of the skill. `skill-solution-project-list` exercises the **operate half** (CLI usage on a tempdir) — neither edited file is referenced from the operate workflow.

**Other failing check (also unrelated):** `Claude Code Review` errors with `Invalid API key · Fix external API key` — the `ANTHROPIC_API_KEY` repo secret needs rotating; this happens on every PR right now (visible on #872, #871, #870, etc.).

PR #876, opened 2026-05-20
refactor(tests): split smoke experiment configs and remove driver hardcoding

akshaylive:
- Create default_smoke.yaml and default_smoke_rpa.yaml for non-RPA and RPA smoke tests
- Update smoke-skills.yml and smoke-rpa-skills.yml to use respective experiment configs
- Remove docker driver hardcoding from sandbox sections in 125+ smoke-tagged tasks
- Update detect job to trigger on new experiment files

### DIFF
--- a/.github/workflows/smoke-rpa-skills.yml
+++ b/.github/workflows/smoke-rpa-skills.yml
@@ -270,7 +270,7 @@ jobs:
             for attempt in 1 2 3; do
               echo "--- $task attempt $attempt ---"
               if coder-eval run "$task" \
-                  -e experiments/default.yaml --tags smoke -j 1 -v; then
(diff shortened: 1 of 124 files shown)


andreibalas-uipath (approved):

tests/experiments/default.yaml
```diff
@@ -14,6 +14,7 @@ defaults:
     turn_timeout: 900
 
   sandbox:
+    driver: tempdir
```
bai-uipath, line 17: should the `docker:` block below be dropped since you put `driver: tempdir`?

bai-uipath (approved):

tests/experiments/default.yaml
akshaylive: Doesn't matter either way. I kept in in case someone wants to switch locally.

PR #888, opened 2026-05-21
fix(uipath-data-fabric): repair coder-eval failures across smoke/inte…

AditiGoyalUipath:
…gration/e2e tiers

Eight tests on the data-fabric eval-board were scoring below 1.000. Root causes: entity reuse leaking state across runs, criterion regex patterns that didn't match the agent's tool-call shape, agent budgets too tight for orchestration overhead, and context bloat from unfiltered API responses.

Per file:

- e2e_employee_directory, e2e_product_catalogue: unique entity-name suffix (<name>_<EPOCH>) + report.json requirement so cleanup_entities.py can wipe the entity post-run.

- integration_error_paths: remove `entities create` criterion that contradicted the prompt's "reuse if exists" instruction; max_turns 30 -> 50 to fit 13 scenarios plus orchestration overhead.

- integration_full_lifecycle: widen the fieldName regex from `create` to `(create|update)` so the addFields path counts; use Notes_<EPOCH> field name so the addFields call always operates on a fresh field.

- integration_large_pagination: pre_run wipes records before the agent starts (no more 500-record cleanup inside the turn budget); helper scripts keep per-page context under 250 bytes; max_turns 30 -> 60.

- integration_preseeded_entity: require each cursor page to be its own records-list Bash invocation (subprocess loops inside a single Python call don't satisfy criteria that count tool calls).

- smoke_records: max_turns 30 -> 50 for the 10-step offline test; forbid batched shell commands that hide multiple uip calls behind one tool call.

New shared helpers (tests/tasks/uipath-data-fabric/_shared/):

- extract_page_ids.py: stream records list/query response IDs to file, emit only pagination metadata to stdout (Result, IdCount, HasNextPage, NextCursor)
- verify_pagination.py: assert ID files match walk / offset / count expectations
- wipe_entity_records.py: paginate + bulk-delete all records from a named entity (used in pre_run; uses records delete only, no entity-level destructive ops)

All 15 data-fabric tests now pass at 1.000 against codereval/DefaultTenant.


bai-uipath (approved):
Linter found issues with prompt overspecification (too much details in prompt that should be in the skill) - please fix and 🚢

PR #890, opened 2026-05-21
fix: Remove separate df connector file and rely on connector skills. …

AditiGoyalUipath:
…Co-authored-by: Claude <noreply@anthropic.com>


skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
```diff
@@ -33,6 +33,10 @@ Connector nodes come in two flavors:
 
 To classify a node, read `Node.form.sections[0].fields[0].componentProps.connectorDetail.configuration` from the `registry get` response, parse it as JSON, and check `activityType`. `"Generic"` → run Step 2a to discover `objectName` (and capture `operation` from the same marker for the `--operation` flag in Step 3). Anything else → skip Step 2a.
 
+### Data Fabric activity version — default to V2 curated
```
baishalighosh: Are V3 activities present on prod ? How are customers using this , why can't these be hidden from registry ? It is more deterministic to simply filter these from the output of the registry search than to defer these to the agent. Frow the flow tool , you can filter them out , there should be no need for this in the skill

baishalighosh: https://github.com/UiPath/cli/blob/main/packages/flow-tool/README.md#flow-registry-commands

AditiGoyalUipath: How to make it environment specific ? The main concern is how we should plan the rollout of V3 activities. For a period of time, both V2 and V3 activities will coexist in the same environment. Right now this is only happening in Alpha, where Coding Agent is defaulting to V3 instead of V2(Creating broken workflow).
Once V3 activities are promoted to higher environments, we may face the same issue there as well. It would be helpful to align on the expected approach.

baishalighosh (approved):

skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
AditiGoyalUipath: Aligned with this approach . Will add filter in CLI to exclude V3 Activities

PR #901, opened 2026-05-21
docs(uipath-rpa): cut redundant build calls + add Studio pre-warm

roalexandru:
## Summary

- **Defer project-level `build` from per-Edit to once-per-edit-session.** Per-file `validate` cadence unchanged. `build` is project-scoped, so N sequential edits used to trigger N full-project rebuilds — now a single end-of-session build. Saves ~55s on a 5-workflow project, ~475s on a 20-workflow project.
- **Restructure the validation iteration loop into two phases.** Phase 1 loops `validate` to clean per-file; Phase 2 loops `build` to clean per-project. Same fix-one-thing discipline, same 5-attempt cap — only the loop topology changes.
- **Smoke-test `run` uses `--skip-build` after a clean build.** Default `run` re-validates and re-builds internally (~10s of duplicate compilation per run). The `--skip-build` flag short-circuits that work when the project has just been built clean.
- **New `## Session Pre-warm` mini-section.** Background a cheap `uip rpa activities find` call at session start to hide Studio's ~22s host cold-start tax behind planning work. Output redirected to `/dev/null` so the warm-up doesn't pollute the foreground stream.

Improvement #4 from the source research (a curated common-activity property card) is **intentionally not adopted** — it conflicts with Rule 21a's explicit non-exemption for `LogMessage` / `Delay` / `GetText`.

Expected wall-clock saving on a 20-workflow project: ~735s (~85%) of CLI cost.

## Test plan

- [x] `bash hooks/validate-skill-descriptions.sh` passes (exit 0) — frontmatter intact
- [ ] Visual diff confirms heading hierarchy preserved in both files (no skipped levels)
- [ ] All internal links resolve (`#smoke-test` anchor, `references/validation-guide.md`, etc.)
- [ ] Rule numbering preserved (Critical Rules still 1–24 with sub-rules 6a, 7a, 7b, 8a, 21a)
- [ ] Manual: drive a 2-workflow create flow through the updated skill and confirm the agent (a) backgrounds the pre-warm call, (b) does not invoke `build` after every Edit, (c) reaches a single end-of-session `build`, (d) passes `--skip-build` to the smoke `run`

### DIFF
--- a/skills/uipath-rpa/SKILL.md
+++ b/skills/uipath-rpa/SKILL.md
@@ -120,6 +120,16 @@ When the request is "automate this dialog/form" or "build a UI test from these m
 
 Skip this path when the task has no UI surface (data transforms, IS connector calls, headless file/email automation).
 
+## Session Pre-warm
+
+First heavy `uip rpa` call pays a ~22s Studio host cold-start (shared across `validate`/`build`/`run`/`activities get-default-xaml`/`analyzer-rules list`). When more than one is expected this session, background a cheap warm-up at session start so the tax hides behind planning:
+
+```bash
+uip rpa activities find --query log --output json > /dev/null 2>&1 &
+```
+
+**Skip** when 0 or 1 heavy `uip rpa` calls are expected (read-only Q&A, single-file inspection) — the warm-up doesn't reclaim its cost.
+
 ## Critical Rules
 
 ### Common Rules (Both Modes)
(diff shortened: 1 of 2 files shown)


gabrielavaduva:
Disambiguate the 5-attempt cap. The new wording (`SKILL.md` Rule 3 and `validation-guide.md` Phase 2) caps "the combined Phase 1 + Phase 2 work" at 5 attempts. On a project with N files where each needs one routine fix in Phase 1, the cap is exhausted before Phase 2 starts. Please clarify whether the budget is per-file in Phase 1 + a separate counter for Phase 2, or genuinely global across the session.

gabrielavaduva:
Add a skip condition for Session Pre-warm. The new `## Session Pre-warm` section fires `uip rpa activities find` unconditionally at session start. Read-only / Q&A sessions that never call `uip rpa` pay a wasted process and don't reclaim the ~22s cold-start tax it's trying to mask. Critical Rule 3 already models the right shape ("DO NOT run at session start" for analyzer-rules) — please add a parallel guard, e.g., skip when the task is read-only or no `uip rpa` calls are expected this session.

gabrielavaduva:
Cross-platform note on the pre-warm command. `uip rpa activities find --query log --output json > /dev/null 2>&1 &` is Bash-only. On Windows PowerShell, `&` is the call operator (not a background spawn) and `/dev/null` doesn't resolve — the warm-up silently fails. Usually fine because Claude Code's Bash tool is invoked, but worth either (a) dropping the `&` form, or (b) adding an explicit "Bash only — does not warm up on PowerShell" note so a Windows-shelling agent doesn't assume it ran.

gabrielavaduva:
Trim the two "why" paragraphs to comply with the No Meta-Commentary rule in `skills/uipath-rpa/CLAUDE.md`. The new `**Why deferred:**` block in `SKILL.md` Rule 3 and the framing paragraph at the top of `validation-guide.md § Validation Iteration Loop` both explain *why the structure exists* rather than prescribing behavior — the rule explicitly forbids "This section exists because Z" patterns. The loop diagram already conveys the topology; the prescriptive content can stand without the justification.

gabrielavaduva:
Add a guard for the `--skip-build` ↔ post-build-edit case. The current guidance in `validation-guide.md § Smoke Test` covers "no recent build was performed" but not "build passed, then a file was edited, then run." In that sequence `--skip-build` runs against a stale build artifact and the edit is silently ignored at runtime. Please extend the "when to use bare `run`" note to cover any edit between `build` and `run`, not only the no-build case.

roalexandru:
Pushed amend `1bd7285f`. Status per comment:

| Comment | Status |
|---|---|
| 5-attempt cap ambiguity (@gabrielavaduva) | **Fixed** — wording is now `5-attempt cap per loop`: 5 attempts for each file's Phase 1 `validate` loop + a separate 5 for the Phase 2 `build` loop. Applied in both SKILL.md Rule 3 and `validation-guide.md`. |
| Pre-warm skip condition (@gabrielavaduva) | **Fixed** — explicit `Skip when 0 or 1 heavy uip rpa calls are expected (read-only Q&A, single-file inspection) — the warm-up doesn't reclaim its cost.` added after the code block. |
| `--skip-build` after build → edit → run (@gabrielavaduva) | **Fixed** — bare-`run` guard now covers **both** (a) no recent build AND (b) any file edit between the last successful `build` and this `run`. Spelled out that `--skip-build` runs the existing compiled artifact, so post-build edits are silently ignored until a fresh `build`. |
| No Meta-Commentary (@gabrielavaduva) | **Fixed** — dropped the `**Why deferred:**` sub-bullet in SKILL.md Rule 3 (kept the operational `run --skip-build` pointer as a `**Smoke-test shortcut:**` bullet); compressed the `validation-guide.md` `## Validation Iteration Loop` intro from a why-paragraph to a one-line Phase 1 / Phase 2 summary. |
| Duplication finding (Claude review bot) | **Fixed** by the No-Meta-Commentary trim. SKILL.md Rule 3 "Project-level build" bullet now names the error categories briefly and points at `validation-guide.md § Errors build catches that validate misses` instead of inlining the full list with `NGetText.Value` / `Operator="StartsWith"` examples. |
| Cross-platform `&` note (@gabrielavaduva) | **Declined.** Per `.claude/rules/content-quality.md`: *"Shell commands must use Unix syntax… Never use Windows-specific commands in skill documentation."* The ` ```bash ` fence tag signals Unix-shell context; the convention across the repo is that agents translate to native shell as needed. Adding a Windows-disclaimer here would conflict with the convention applied to every other skill. Happy to revisit if the policy itself is being relaxed. |

Note: these are PR issue-comments, not review-thread comments, so they don't have a "resolve" affordance via API. Please collapse/hide via the UI if you'd like them tidied.

AlvinStanescu (approved):

PR #904, opened 2026-05-21
fix(ci): enable globstar so nested smoke tasks are discovered

radugheo:
## Summary

- Add `shopt -s globstar` to the `Run smoke tests` step in `smoke-skills.yml`

## Root cause

Without `globstar`, bash treats `**` as `*`. The detect job outputs `tasks/uipath-agents/**/*.yaml` as the glob pattern, but when the `e2e` job runs it without globstar enabled, the shell only expands one level — matching files 2 levels deep (e.g. `lowcode/init_validate.yaml`) and silently skipping anything 3 levels deep (e.g. `lowcode/schema_update/schema_update.yaml`).

Two lowcode smoke tasks — `schema_update` and `edit_roundtrip` — have never run in CI.

This also unblocks PR #902 (`test/organize-coded-agent-tests`), which moves coded tasks into a `coded/` subdirectory — without this fix, that PR would break all coded smoke test coverage.

## Test plan

- [x] CI smoke run picks up `schema_update`, `edit_roundtrip`, and all coded smoke tasks (not just the flat `init_validate.yaml` and `bindings_sync.yaml`)
image

### DIFF
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -282,6 +282,7 @@ jobs:
         working-directory: tests
         id: smoke
         run: |
+          shopt -s globstar nullglob
           echo "Running: coder-eval run ${{ needs.detect.outputs.task_globs }} --tags smoke -j $TASK_PARALLELISM"
           coder-eval run ${{ needs.detect.outputs.task_globs }} \
             -e experiments/smoke.yaml \


radu-mocanu (approved):

andreibalas-uipath (approved):

uipreliga:
## Code Review Summary

Reviewed by Gemini 3, Codex, and Opus 4.6 in parallel. All three converged on the same primary finding.

### Medium

- **Missing `nullglob` lets a literal glob leak to `coder-eval`** (`.github/workflows/smoke-skills.yml:297`) — *flagged by all three reviewers.* The consumer enables `globstar` but not `nullglob`. The producer (`detect` job, line 108) guards only on `[ -d "tests/tasks/$skill" ]`, never checking that any `*.yaml` exists inside. If a skill directory has subdirs/README but no YAML (work-in-progress test folder, `.yaml`→`.yml` rename, etc.), bash passes the literal `tasks/<skill>/**/*.yaml` string to `coder-eval` — likely a "no such file" error or silently empty run. The fan-out path (line 65) already pairs `globstar` with `nullglob`; the per-skill path should too. **Suggested:** either set `shopt -s globstar nullglob` in the consumer, or pre-expand in the detect loop (see below).

### Low

- **Two divergent strategies for the same task-list construction** (lines 64–80 vs 99–123 + 297) — *Opus + Gemini.* The fan-out branch expands `**` inside `detect` and emits a literal file list; the per-skill branch emits unexpanded `**` and trusts the consumer to set `globstar`. This is the exact fragility that necessitated this PR. A future maintainer editing the consumer step (composite action, matrix split, etc.) can silently regress it. Consolidating the per-skill loop on pre-expansion would fix this and the `nullglob` issue for free, and let the consumer drop `shopt` entirely.

- **`task_globs` interpolated unquoted** (lines 298–299) — *Opus, informational.* Intentional (one output → many args), but relies on `detect` guaranteeing non-empty `task_globs` or `skip=true`. Worth a one-line comment near the consumer documenting that invariant. Not introduced by this PR.

### Positive Observations

- The one-line fix is **correct for the happy path**: GitHub Actions' default `bash --noprofile --norc -eo pipefail {0}` does not enable `globstar`, and setting it in the same shell that performs expansion is the right scope.
- **Minimal blast radius** — single-line, single-step. No new deps, no fan-out behavior change.
- **Scope discipline** — resists the tempting refactor of both producers into one helper, appropriate for a `fix/` branch.
- Commit message clearly explains the *why* (nested tasks from refactor 0403f80d).

### Overall Assessment

Correct, minimal hotfix that unblocks the regression. The remaining concern is durability: the producer/consumer split is fragile, and the missing `nullglob` leaves an empty-glob edge case unhandled. Reasonable to ship as-is and follow up with a consolidation PR that moves per-skill glob expansion into the `detect` job (matching the fan-out path) — that would let the consumer drop `shopt` entirely and eliminate both findings.

uipreliga (approved):
🚢

PR #909, opened 2026-05-21
fix(uipath-maestro-flow): fix inline agent prompt wiring contract

rockymadden:
# Fix inline-agent prompt-wiring contract

The `uipath-maestro-flow` skill (and one mirror reference in `uipath-agents`) documents an **agent input bridge** for wiring flow data into inline-agent prompts: `agentInputVariables[]` on the flow node ↔ `inputSchema.properties.<id>` ↔ `{{input.<id>}}` token ↔ `contentTokens[].rawString: "input.<id>"`.

That contract is wrong. Studio Web's canvas does not resolve `input.*` against the agent's `inputSchema` when rendering prompt tokens — it rewrites `{{input.<id>}}` to `{{ $vars.<id> }}` against flow scope, then fails its own lint with:

> `Prompt references "$vars.<id>" but that variable is not available in this scope. Fix or remove the reference before publishing.`

The correct contract — verified end-to-end against `uip agent validate --inline-in-flow` and a published flow in Studio Web — is to reference upstream flow nodes **directly** in `agent.json messages[].content`, with no bridge:

```json
"content": "From {{ $vars.emailReceived1.output.from }}\nSubject: {{ $vars.emailReceived1.output.subject }}",
"contentTokens": [
  { "type": "simpleText", "rawString": "From " },
  { "type": "variable",   "rawString": " $vars.emailReceived1.output.from " },
  { "type": "simpleText", "rawString": "\nSubject: " },
  { "type": "variable",   "rawString": " $vars.emailReceived1.output.subject " }
]
```

Plus a previously-undocumented constraint: `contentTokens[].rawString` for variable tokens **must include leading and trailing spaces** to match the spaced-brace `{{ <expr> }}` form in `content`. Without them, `uip agent validate` fails with:

> `messages[N].contentTokens[M].rawString: Expected " $vars.X " but got "$vars.X"`

## Changes

| File | Change |
| --- | --- |
| `skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md` | Rewrote the **Wiring Flow Variables into Agent Prompts** section: replaced the four-place bridge with the direct `{{ $vars.<flowNodeId>.output[.<field>] }}` form. Updated the worked example (email-trigger) to use empty `agentInputVariables: []` and `inputSchema.properties: {}`. Rewrote the **When the source field name is unknown** subsection. Inverted the anti-patterns list (the old "Never write `{{ $vars.X }}` directly" rule was backwards; now the bans are `{{input.X}}`, bare `{{plainName}}`, omitted `rawString` spaces, and populated `agentInputVariables[]` / `inputSchema.properties` for prompt-data wiring). Updated the JSON Structure note on `inputs.agentInputVariables`. Replaced the literal-`{{X}}` Debug row with three precise rows covering scope-lint, `rawString` spacing, and literal-text symptoms — each citing the exact empirical error string. |
| `skills/uipath-maestro-flow/references/shared/node-output-wiring.md` | Fixed the inline-agent row in the `=js:` field reference table to teach `{{ $vars.<flowNodeId>.output[.<field>] }}` + spaced `rawString` instead of the bridge form. |
| `skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md` | Updated the `inputSchema.properties` and `messages`/`contentTokens` bullets in **Step 3: Create agent.json** to describe the canvas-form contract (no `{{input.X}}` slot declarations; one `simpleText` token per literal segment and one `variable` token per `{{ ... }}` reference). |

## Verification

End-to-end empirical confirmation against `alpha.uipath.com / popoc / DefaultTenant`:

- Built a flow with an upstream HTTP node → Script node → inline-agent → Slack node.
- First attempt used the old four-place bridge as documented — `uip agent validate` passed, but Studio Web showed `Prompt references "$vars.hnHits" but that variable is not available in this scope.`
- Switched the prompt to `{{ $vars.extractHits1.output.hnHits }}` with empty `agentInputVariables: []` / `inputSchema.properties: {}` — `uip agent validate` first failed on unspaced `rawString`, then passed once leading/trailing spaces were added inside `rawString`. Studio Web accepted the flow.
- Grepped the entire skills repo for stale `{{input.<id>}}` / "four-place" / "input bridge" references — all remaining matches are either anti-pattern citations (warning against the bad form) or unrelated JSON skeletons with `agentInputVariables: []`.

## What this PR does NOT change

- The flow-node JSON skeletons under **JSON Structure** and **Adding / Editing** continue to show `agentInputVariables: []` — that's correct under the new contract and unchanged.
- No claims are made about non-prompt typed-schema uses of `agentInputVariables[]` or `inputSchema.properties.<id>`. The new docs constrain themselves to **prompt-only flow-data references**, which is what was verified.
- `uip agent migrate --inline-in-flow` and `uip solution upload` behavior is unchanged — only the authoring contract is corrected.

### DIFF
--- a/skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
+++ b/skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
@@ -71,8 +71,8 @@ Generate a unique UUID (e.g., `5029c8a8-799b-426a-803f-c4ec75255439`). Create a 
 
 Same schema as a standalone agent (see [../../agent-definition.md](../../agent-definition.md)), with these conventions:
 - `projectId` matches the folder name UUID
-- `inputSchema.properties` starts empty, but **must declare one slot per `{{input.<id>}}` token used in `messages[].content`**. Each slot's `<id>` and `type` must match the corresponding `agentInputVariables[]` entry on the flow node. See the `uipath-maestro-flow` skill's [inline-agent prompt-wiring guide](../../../../../uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md#wiring-flow-variables-into-agent-prompts) for the full four-place contract.
-- `messages` have empty `content` and `contentTokens` initially (edit agent.json to set prompts with `type: "simpleText"` and `rawString`)
+- `inputSchema.properties` stays empty `{}` for prompt-only flow-data references. Prompts reference upstream flow nodes directly via `{{ $vars.<flowNodeId>.output[.<field>] }}` in `messages[].content`, mirrored in `contentTokens[]` as `{ "type": "variable", "rawString": " $vars.<flowNodeId>.output[.<field>] " }` (leading and trailing space inside `rawString`). See the `uipath-maestro-flow` skill's [inline-agent prompt-wiring guide](../../../../../uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md#wiring-flow-variables-into-agent-prompts).
+- `messages` have empty `content` and `contentTokens` initially. Set prompts in `messages[].content`, then build `messages[].contentTokens[]` as a parallel list: one `{ "type": "simpleText", "rawString": "..." }` per literal text segment, one `{ "type": "variable", "rawString": " $vars.<flowNodeId>.output[.<field>] " }` per `{{ ... }}` reference.
 - `guardrails: []` at root level — can be populated with guardrail objects. See [../guardrails/guardrails.md](../guardrails/guardrails.md)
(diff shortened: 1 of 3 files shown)


andreibalas-uipath (approved):

tmatup (approved):
I wonder if we should tighten up the rule in `agent validate` and `flow validate` accordingly so the first attempt will fail earlier instead of later when StudioWeb rejected it.

rockymadden:
> I wonder if we should tighten up the rule in `agent validate` and `flow validate` accordingly so the first attempt will fail earlier instead of later when StudioWeb rejected it.

Sounds fantastic to me.

PR #915, opened 2026-05-21
docs(uipath-maestro-case): preserve schema extras + document action outputs

song-zhao-25:
## Summary

- **io-binding/impl-json.md** — adds a blockquote in § Output Binding Shapes stating that schema-provided extras (`_jsonSchema`, `options`, `body`, `displayName`) round-trip verbatim onto the emitted output. Per-plugin field lists are a minimum, not a maximum.
- **plugins/tasks/action/planning.md** — adds an "Always-Emitted Outputs" subsection naming the two synthetic outputs surfaced on every action task: `Action` (with `options` enum from outcomes) and `hitlTask` (with PascalCase sub-fields like `AssignedToUser.UserName`).
- Companion to upcoming CLI change `fix/case-tasks-describe-non-connector-divergences` (UiPath/cli), which makes `_jsonSchema`, `Action.options`, and `hitlTask` available in `case tasks describe` output. Until that CLI change lands, this skill doc accurately describes the FE-canonical contract — but the schema-provided extras won't actually flow through.

## Test plan

- [ ] Read both edits end-to-end for clarity and consistency with surrounding docs
- [ ] Confirm the "preserve schema extras" rule is discoverable from the per-plugin task-write steps (linked from agent/action/rpa/process/case-management/api-workflow impl-json `Output binding` references to `io-binding/impl-json.md § Output Binding Shapes`)
- [ ] After CLI prerequisite lands, smoke an action task in the case skill and verify `Action.options` + `hitlTask` appear in the emitted caseplan.json

### DIFF
--- a/skills/uipath-maestro-case/references/plugins/tasks/action/planning.md
+++ b/skills/uipath-maestro-case/references/plugins/tasks/action/planning.md
@@ -38,6 +38,15 @@ Pick this plugin when the sdd.md describes a `HITL` task, or any task requiring 
 
 See [registry-discovery.md](../../../registry-discovery.md#cli-search-gaps) for the fallback rationale.
 
+## Always-Emitted Outputs
+
+`tasks describe --type action` returns two synthetic outputs on every action task in addition to the action-app's declared outputs:
+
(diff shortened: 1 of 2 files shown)


song-zhao-25:
Abandoning — the skill change isn't actually needed. The principle 'skill returns whatever CLI outputs' already covers the case; the per-plugin templates' enumeration is implicitly non-exhaustive since they say 'from Step 0 schema'. Adding a preserve-extras rule and an action-specific outputs section was over-documentation. The CLI-side fix (UiPath/cli fix/case-tasks-describe-non-connector-divergences) delivers all the value.

PR #918, opened 2026-05-21
refactor: move UiPath CLI tools into Docker image for smoke tests

akshaylive:
## Summary
Move @uipath/cli and @uipath/admin-tool installation from GitHub Actions runner into a custom Docker image extending coder-eval-agent. This ensures tools are available inside the container where tests actually run.

## Changes
- Add `.github/docker/Dockerfile.smoke` extending coder-eval-agent with CLI tools from GitHub Packages @alpha (tracks `cli/main`)
- Update `smoke-skills.yml` to build custom image instead of installing on runner
- Update `default_smoke.yaml` experiment to use `skills-smoke-image:latest` globally for all smoke tests (leverages BYOD feature from coder_eval)

## Why
Previously, CLI tools were installed on the GitHub Actions runner but tests run inside Docker containers, so the tools weren't available to the agent. Now the tools are baked into the container image itself.

🤖 Generated with Claude Code

### DIFF
--- a/.github/docker/Dockerfile.smoke
+++ b/.github/docker/Dockerfile.smoke
@@ -1,0 +1,18 @@
+# Custom image for UiPath skills smoke tests
+# Extends coder-eval-agent with UiPath CLI tools pre-installed
+ARG CODER_EVAL_IMAGE=coder-eval-agent:latest
+FROM ${CODER_EVAL_IMAGE}
+
+# Install @uipath/cli and @uipath/admin-tool from GitHub Packages
+# These are needed by skill smoke tests (orchestrator, integration-service, data-fabric, etc.)
+ARG NPM_AUTH_TOKEN=
+RUN set -euo pipefail && \
+    if [ -z "$NPM_AUTH_TOKEN" ]; then \
+      echo "Error: NPM_AUTH_TOKEN must be passed as build arg"; exit 1; \
+    fi && \
+    npm config set @uipath:registry https://npm.pkg.github.com/ && \
+    npm config set //npm.pkg.github.com/:_authToken "$NPM_AUTH_TOKEN" && \
+    npm install -g @uipath/cli@alpha @uipath/admin-tool@alpha && \
+    npm config delete //npm.pkg.github.com/:_authToken && \
+    uip --version && \
+    uip admin --help
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -215,23 +215,19 @@ jobs:
           UV_INDEX_UIPATH_PASSWORD: ${{ secrets.UV_INDEX_UIPATH_PASSWORD }}
         run: make docker-image
 
-      # Install from GitHub Packages under the `alpha` dist-tag so smoke
-      # tracks `cli/main`. Every merge to `cli/main` runs
-      # cli/.github/workflows/ci.yml, which publishes a
-      # `-alpha.<date>.<run>` prerelease to
-      # https://npm.pkg.github.com/ under dist-tag `alpha`. Public npm
-      # only carries the cut-release `latest` line which lags `main` by
-      # days, so smoke against `@latest` was masking CLI changes that
-      # task YAMLs and skills had already been updated for.
-      - name: Install uip CLI + admin tool (GitHub Packages @alpha)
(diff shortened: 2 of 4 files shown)


bai-uipath:
The shape is right — `Dockerfile.smoke` + experiment-level `image:` override is exactly what we agreed in DM. A few things to address before merging:

## Major

1. **This only fixes PR-CI smokes, not the daily on the VM.**
   PR #918 updates `tests/experiments/default_smoke.yaml`, which is consumed by `.github/workflows/smoke-skills.yml`. The nightly skills suite on the `coder-eval-runner` VM runs `dashboard run --suite skills`, which uses `tests/experiments/default.yaml` and `tests/experiments/e2e.yaml`. **Neither is touched by this PR.**

   Tracing `coder_eval/dashboard/scripts/ci/daily.sh` after this merges:
   - `make docker-image` builds `coder-eval-agent:<version>` / `:latest` — base image has no uip (only `@anthropic-ai/claude-code`, per `coder_eval/docker/Dockerfile`).
   - daily.sh never builds `skills-smoke-image:latest`.
   - `default.yaml` / `e2e.yaml` don't set `defaults.sandbox.docker.image`, so the 115 docker-driver tasks boot the bare base image.
   - In-container `ensure-uip.sh` falls back to public-npm `@latest` = stale **1.0.4** — same bug.

   To actually close the loop, also:
   - Set `defaults.sandbox.docker.image: skills-smoke-image:latest` in `default.yaml` and `e2e.yaml`.
   - Add a `docker build … -f $SKILLS_REPO/.github/docker/Dockerfile.smoke -t skills-smoke-image:latest …` step to `daily.sh` (the `GH_NPM_REGISTRY_TOKEN` is already loaded from `.env`).

   Either bundle into this PR or commit to a same-day follow-up — otherwise the daily silently keeps running stale uip on the 115 docker tasks, and the gap grows as more tasks migrate off tempdir.

2. **Don't drop the host-side `uip login status` verification while 198 tempdir tasks still depend on host uip.**
   Second commit removes the `uip login status --output json` block on the runner. That's fine on the smoke runner because everything in the smoke set is dockerized — but it's worth a one-line comment in the workflow explaining *why* it's safe (i.e., "all smoke tasks are docker-driver; tempdir tasks would still need host-side uip + login verification"). Otherwise the next person looking at the diff will think the verification was unnecessary in general and remove it elsewhere too.

3. **Pin the base image tag, don't drift on `:latest`.**
   `Dockerfile.smoke` defaults `CODER_EVAL_IMAGE=coder-eval-agent:latest`. `coder_eval`'s preflight (`_preflight_image_version`) warns when the image's `org.coder-eval.version` label ≠ host `coder_eval` package version. With `:latest`, that label drifts whenever the runner has a newer/older base than the smoke layer. Either:
   - Pass the explicit version from the workflow: `--build-arg CODER_EVAL_IMAGE=coder-eval-agent:$(uv run python -c 'from importlib.metadata import version; print(version("coder-eval"))')`, or
   - Leave `:latest` but add a comment that this is intentional and the preflight warning is benign.

4. **`Dockerfile.smoke` writes the token into `npm config` and deletes it — image layers still leak it.**
   `npm config set //npm.pkg.github.com/:_authToken "$NPM_AUTH_TOKEN"` writes to `~/.npmrc`. The `npm config delete` runs in the same `RUN` step, so the final layer is clean — good. But the **build-arg itself** is visible via `docker history --no-trunc`. Switch to BuildKit secret mount:
   ```dockerfile
   # syntax=docker/dockerfile:1.4
   RUN --mount=type=secret,id=npm_auth_token \
       NPM_AUTH_TOKEN="$(cat /run/secrets/npm_auth_token)" && \
       ...
   ```
   and pass `--secret id=npm_auth_token,env=GH_PAT` in the workflow. Same pattern coder_eval already uses for `uv_index_password`.

## Overall

**7/10.** Clean implementation of the mechanism. The PR description says "tests run inside Docker containers, so the tools weren't available" — that's true for the PR-CI smoke path this fixes, but the same statement applies to the daily on the VM, which this PR leaves broken. Either expand the scope or land a same-day follow-up that updates `default.yaml` / `e2e.yaml` and `daily.sh`.

bai-uipath (approved):

PR #894, opened 2026-05-21
docs(uipath-troubleshoot): add Excel Read Range sheet-not-found playbook

Stefan-Virgil:
## Summary

Second playbook in the `excel-activities` package. Covers the activity opening the workbook successfully but failing to resolve the configured `SheetName`:

```
UiPath.Excel.BusinessException: The sheet with the name '<configured-name>' does not exist.
```

Stacked on **#889** (excel-activities package + first playbook).

## Cause-branches

| # | Branch | Smoking-gun evidence |
|---:|---|---|
| 1 | Typo in the configured sheet name | No similar name in `Get Workbook Sheets` output |
| 2 | Case mismatch on OpenXML provider | Same letters in workbook, different casing; `Use Excel File` on headless host (Excel-COM is case-insensitive, OpenXML is not) |
| 3 | Sheet renamed upstream | Configured name absent; a recently-renamed sheet matches semantically |
| 4 | Sheet deleted upstream | Configured name absent; no close match; data was moved or archived |
| 5 | Leading/trailing whitespace | `.Length` differs by one between configured and actual |
| 6 | Look-alike Unicode (NBSP, Cyrillic `а` vs Latin `a`, curly quotes) | Same `.Length` but byte sequences differ |
| 7 | Dynamic `SheetName` expression resolved wrong | Error names something nonsensical (`Null`, empty, integer, path fragment) |

Investigation chain: enumerate workbook sheets → byte-compare suspicious near-matches with a PowerShell `[System.Text.Encoding]::UTF8.GetBytes()` snippet → determine provider (COM vs OpenXML) → trace dynamic expressions when the resolved value looks off.

Resolution is per-branch and includes a cross-branch prevention list (enumerate-and-validate at job start, `.Trim()` external names, treat workbook layout as a contract).

## Test plan

- [ ] Validate frontmatter passes `hooks/validate-skill-descriptions.sh` (no SKILL.md change in this PR, but confirms repo-wide hook still passes)
- [ ] Walk the investigation steps against a real Faulted job with this exception and confirm each step resolves to actionable evidence
- [ ] Spot-check that `excel-activities/summary.md` table renders correctly on GitHub
- [ ] Confirm the playbook's `## Context` matches what the hypothesis-generator agent expects (one or more cause-branches per the medium-confidence pattern)

## Follow-ups

[...]

PR #893, opened 2026-05-21
test(uipath-troubleshoot): 3 scenarios for Excel Read Range file-locked playbook

Stefan-Virgil:
## Summary

Adds three coder-eval e2e scenarios exercising distinct cause-branches of the `activity-packages/excel-activities/playbooks/read-range-file-locked.md` playbook (PR #889).

| Scenario | Branch | Surface signal | Resolution path |
|---|---|---|---|
| `excel-rr-concurrent-jobs` | Concurrent Robot jobs (branch 4) | Two jobs of same process, overlapping `StartTime`/`EndTime`; `or triggers list` shows identical cron | Serialize via per-workbook lock asset, single-performer queue, or stagger triggers |
| `excel-rr-orphan-process` | Orphan `EXCEL.EXE` from prior job (branch 2) | Prior faulted run + `ContinueOnError="True"` on Excel scope + swallowing `TryCatch` in `process/Main.xaml` | One-off orphan kill + workflow fix (remove `ContinueOnError`, fix `TryCatch`, migrate to `Use Excel File`) |
| `excel-rr-host-evidence` | Collapsed 1/3/5/6 (host evidence needed) | CLI rules out concurrent + orphan but cannot positively identify holder | Recommend `Get-Process EXCEL` / `handle.exe -a <path>` / `Get-SmbOpenFile` to the user — do NOT attempt them |

Each scenario chains: `folders list-current-user` → `jobs list --state Faulted` → `jobs get` → `jobs logs` → `jobs list` (broader) → `triggers list` → (optionally) `jobs get <sibling>` + workflow source inspection.

## Local results

Ran `coder-eval --repeats 3 -j 3` against each scenario (Windows local, claude-sonnet-4-6 coder). **9/9 replicates passed.**

| Scenario | Replicate scores | Mean | Status |
|---|---|---:|---|
| `excel-rr-concurrent-jobs` | 1.000 / 0.867 / 1.000 | 0.956 | 3/3 SUCCESS |
| `excel-rr-orphan-process` | 1.000 / 0.933 / 1.000 | 0.978 | 3/3 SUCCESS |
| `excel-rr-host-evidence` | 1.000 / 1.000 / 1.000 | 1.000 | 3/3 SUCCESS |

Durations: 441s–3513s per replicate (slowest replicate burned 98% of the 3600s `turn_timeout`).

## Notes

[...]

PR #889, opened 2026-05-21
docs(uipath-troubleshoot): add Excel Activities file-locked playbook

Stefan-Virgil:
## Summary

Scaffolds the new `excel-activities/` activity-package folder under `uipath-troubleshoot/references/` and lands the first playbook for `UiPath.Excel.Activities` Read Range failures where the workbook is held open by another process (`System.IO.IOException: The process cannot access the file '<path>' because it is being used by another process.`).

Package files:

| File | Purpose |
|---|---|
| `excel-activities/overview.md` | Package providers (Excel COM vs. OpenXML), scopes (Excel Application Scope vs. Use Excel File), activity families, named failure-pattern catalog |
| `excel-activities/summary.md` | Playbook index (one entry for now) |
| `excel-activities/investigation_guide.md` | Data correlation rules + testing prerequisites (provider, scope, package version, host context) |
| `excel-activities/presentation.md` | Entity display rules — activities by display name, sheets in single quotes, paths exact, `EXCEL.EXE` uppercase |
| `excel-activities/playbooks/read-range-file-locked.md` | Medium-confidence playbook, 6 cause-branches |
| `references/summary.md` | Top-level router updated with the new Excel Activities section |

Cause-branches in the file-locked playbook:

1. User has the workbook open in Excel UI on the host
2. Orphan `EXCEL.EXE` from a prior job (executor killed mid-scope, `Continue On Error: True`, swallowed scope exception)
3. Network-share lock from a different host (UNC / SharePoint sync)
4. Concurrent Robot jobs racing on the same workbook
5. Antivirus / EDR / Windows Search / sync-client transient hold
6. Hidden Excel instance under a different user session (disconnected RDP)

Investigation walks cheap → expensive (job context → recent-job pattern → host `Get-Process EXCEL` → `handle.exe -a <path>` → file-server `Get-SmbOpenFile` for UNC). Resolution maps each branch to its specific fix and a cross-branch prevention list.

## Test plan

- [ ] Validate the new package is reachable from `references/summary.md` (top-level router) and `excel-activities/summary.md` (package index)
- [ ] Walk the investigation steps against a real job that failed with the file-locked `IOException` and confirm each command resolves to actionable evidence on the host
- [ ] Spot-check that the playbook's `## Context` matches what the hypothesis-generator agent expects (one or more cause-branches per the medium-confidence pattern)
- [ ] Confirm cross-link from the playbook to `o365-activities/overview.md` resolves; confirm summary.md table renders correctly on GitHub

[...]

PR #944, opened 2026-05-22
test(uipath-troubleshoot): 3 scenarios for Excel Read Range sheet-not-found playbook

Stefan-Virgil:
## Summary

Three coder-eval e2e scenarios exercising distinct branches of the `activity-packages/excel-activities/playbooks/read-range-sheet-not-found.md` playbook (PR #894). **Stacked on #894 → #889.** CI will fail until the parent playbook PRs land; this PR merges cleanly after.

| Scenario | Branch | Surface signal | Resolution path |
|---|---|---|---|
| `excel-rr-sheet-typo` | Typo (branch 1) | Get Workbook Sheets log shows `["Sheet1", "Data", "Summary"]`; configured `SheetName: "Datab"` | Change configured name to `"Data"` |
| `excel-rr-sheet-case` | Case mismatch on OpenXML (branch 2) | `Use Excel File` scope, no COM-forcing properties; actual sheet `Data` vs configured `data` | Match casing in workflow, or force COM provider |
| `excel-rr-sheet-bytes` | Whitespace / look-alike (branches 5/6) | Visually-identical names; actual sheet has NBSP `U+00A0` where configured has regular space `U+0020` | Rename sheet in Excel, or normalize in workflow |

Each scenario chains: `folders list-current-user` → `jobs list --state Faulted` → `jobs get` → `jobs logs` (containing the actual sheet titles from the prior `Get Workbook Sheets` activity) → workflow source.

## Local results

Ran `coder-eval --repeats 3 -j 3` per scenario on Windows local (claude-sonnet-4-6 coder). **8/9 replicates passed.**

| Scenario | Replicate scores | Mean | Status |
|---|---|---:|---|
| `excel-rr-sheet-typo` | 1.000 / 1.000 / 1.000 | 1.000 | 3/3 SUCCESS |
| `excel-rr-sheet-case` | 1.000 / 1.000 / 1.000 | 1.000 | 3/3 SUCCESS |
| `excel-rr-sheet-bytes` | 1.000 / 1.000 / ERROR | 0.667 | 2/3 SUCCESS (one turn_timeout) |

Mean across all 9 = **0.926**. Comparable to PR #893's 8/9 / 0.867 result.

## Notes on `excel-rr-sheet-bytes`

The byte-difference scenario evolved through several iterations:

- **Rubric design.** Original rubric required the agent to recommend the host-side PowerShell byte-compare from the playbook. In practice the agent often detects the NBSP directly from the fixture content via Read/Grep on the job-logs response. The current rubric (Path A or Path B) accepts both as 1.000 — what's penalized is **guessing a wrong branch** (typo, case, rename) without supporting evidence.
- **`max_turns: 50`, `task_timeout: 5400`.** The orchestrator skill (triage → hypotheses → test → depth-check → presenter) burns ~25-35 turns on this scenario even when the agent reaches a confident answer. The other two scenarios use the standard 30/4500.
- **Prompt explicitly frames the user as remote.** "I'm not on the Robot host... do not attempt to run those tools yourself... hand me the command list and stop." Same lesson as PR #893's host-evidence scenario.
- **One replicate timed out at 3600s** retrying PowerShell against sandbox guardrails (`Command contains expandable strings`, `subexpressions $()`). This is an agent-behavior failure mode the test correctly surfaces; the prompt tightening reduced its frequency but did not eliminate it.

## Cross-cutting notes

- **Stacked PR.** Targets `main`. Tests reference the playbook from PR #894. Tests fail CI until both #889 and #894 land.
- **Top-level `task_timeout` / `max_turns` / `turn_timeout`** per task (not under `run_limits:`, which is silently dropped).
- **Short `excel-rr-sheet-*` folder + task_id names** to stay under Windows MAX_PATH 260 (per the gotcha discovered in PR #893).
- **`RESOLUTION.md` files use cp1252-safe Unicode characters.** `coder_eval/orchestration/evaluation.py` uses `pathlib.Path.read_text()` without explicit encoding; Windows defaults to cp1252. UTF-8 bytes that don't map (e.g., `0x81` from `U+205F` medium mathematical space) crash the evaluator. Avoided here by using `U+XXXX` notation in prose rather than literal Unicode whitespace.
- **Synthetic fixtures** authored from the playbook signature, not captured from a real `.investigation/` session.

## Test plan

- [ ] CI runs the new scenarios after #889 and #894 land
- [ ] Reviewer spot-checks one fixture per scenario to confirm the cause-branch signal is unambiguous (e.g., for `excel-rr-sheet-bytes`, verify the job-logs response contains the literal NBSP byte `C2 A0` between `Quarterly` and `Data`)
- [ ] Re-run `coder-eval --repeats 1 -j 1` against each scenario on Linux to confirm the 8/9 result is not Windows-specific

## Follow-ups

- Sibling playbooks for the remaining Excel Read Range fingerprints (file-not-found, null-reference / target-invocation) and their test scenarios in separate PRs.
- The `excel-rr-sheet-bytes` scenario's one-in-three timeout exposes a real agent behavior (looping on PowerShell sandbox rejections). Worth investigating separately whether the agent's tool retry policy should back off faster on these specific error classes.


Stefan-Virgil:
## Update — `excel-rr-sheet-bytes` now 3/3

Pushed `cee39400` to fix the lingering 1-in-3 timeout in the byte-inspection scenario.

**Root cause of the timeout:** agent was looping on PowerShell sandbox rejections (the playbook recommends a `[System.Text.Encoding]::UTF8.GetBytes` snippet for the user; the agent was attempting to run it locally and burning ~3600s per turn against rejection errors).

**Fix:**
1. `initial_prompt` now tells the agent explicitly that PowerShell and byte-dump tools are unavailable to it too, and to treat a single tool rejection as a hard signal to switch from execution to writing recommendations.
2. Safety-margin bump: `turn_timeout` 3600 → 5400, `task_timeout` 5400 → 7200.

**Local re-run (`--repeats 3 -j 3`):**

| Replicate | Score | Duration |
|---:|---:|---:|
| 00 | 0.900 | 1298s |
| 01 | 1.000 | **423s** |
| 02 | 1.000 | 2428s |

3/3 SUCCESS, mean 0.967. Replicate 01's 423s shows the agent skipping the local byte-compare attempt entirely and going straight to recommendations.

**PR-wide aggregate is now 9/9 / 0.989** (sheet-typo 1.000, sheet-case 1.000, sheet-bytes 0.967).

Stefan-Virgil:
Closing in favor of a forthcoming bundled read-range tests PR (mirrors the #979 playbook+tests bundle, #980 docs-only bundle pattern). The 3 scenarios and their mock fixtures are preserved on branch `test/troubleshoot-excel-sheet-not-found` and will be brought into the bundle after #980 lands.

Cleanup needed before re-shipping (verified during #979's review cycle):
- Remove `file_exists` criterion on `.local/investigations/state.json` from all 3 task.yamls (per `tests/tasks/uipath-troubleshoot/CLAUDE.md`).
- Remove the `files: [.local/investigations/state.json, .local/investigations/hypotheses.json]` block and the "Evidence sources" paragraphs in the LLM judge prompts that reference `matched_playbooks` / `is_root_cause`. The judge MUST grade only on the agent's final response and tool calls.
- Convert nested `run_limits:` block to flat top-level fields (`task_timeout`, `max_turns`, `turn_timeout`). The nested form is silently dropped by coder-eval and the run hard-times-out at 300s — empirically verified today (see #980 comment on the same finding).

PR #946, opened 2026-05-22
docs(uipath-troubleshoot): add anti-patterns + hidden-sheets to sheet-not-found playbook

Stefan-Virgil:
## Summary

Two follow-up improvements to the sheet-not-found playbook (#894), prompted by common web-article advice that I noticed wasn't yet covered or pushed back on. Stacked on #894 so the diff shows only the new content.

### 1. New branch 8 — Hidden or very-hidden sheet

The configured sheet name DOES exist in the workbook but its visibility is `xlSheetHidden` or `xlSheetVeryHidden`. Distinct symptoms:

- The OpenXML provider returns hidden sheets in lookups; legacy `Excel Application Scope` has historical bug reports of skipping very-hidden sheets on certain `UiPath.Excel.Activities` versions.
- An interactive admin opening the workbook can see the sheet via `Format → Sheet → Unhide` (or via VBA for very-hidden), but the activity reports it missing.
- `Get Workbook Sheets` may or may not list it depending on provider + version.

Resolution covers: unhiding (Excel UI or VBA for very-hidden), package upgrade, and the prevention pattern of "don't very-hide data sheets workflows depend on."

### 2. New Anti-patterns section

Pushes back on two common pieces of advice that hide the bug without fixing it:

| Anti-pattern | Why it's wrong |
|---|---|
| Switching `Use Excel File` ↔ `Excel Application Scope` as a "magic fix" | Coincidentally hides branch 2 (case mismatch) by swapping OpenXML's case-sensitive lookup for COM's case-insensitive one. Does NOT fix branches 1/3/4/5/6/7/8. Also ties the workflow to a COM-only host (Excel installed). Real fix for branch 2 is to match casing in the workflow. |
| `Try Catch` with no real recovery path | Suppresses the symptom but the workflow proceeds with an empty DataTable → silent downstream failures (wrong reports, missing queue items, harder-to-diagnose issues). The same anti-pattern is already called out as an orphan-cause in `read-range-file-locked.md`. |

Both reframe Try-Catch + scope-switching as **debugging steps**, not resolutions.

## Test plan

- [ ] Verify the new branch 8 entry in `What can cause it` matches the description in the Resolution section (no inconsistency in visibility-type naming).
- [ ] Confirm the Anti-patterns section reads consistently with the same anti-pattern called out in `read-range-file-locked.md`'s Resolution Branch 2 (orphan from prior job).
- [ ] Spot-check rendering on GitHub (table + bullet structure).

## Follow-ups

- Test scenarios for the new branch 8 could be added in a future PR (the existing PR #944 covers branches 1/2/5/6 already).


Stefan-Virgil:
Branch 8 (hidden / very-hidden sheet) and the Anti-patterns section from this PR have been ported into #980's `read-range-sheet-not-found.md` (commit `9f65b4e4`). Closing in favor of the bundled PR.

PR #957, opened 2026-05-22
docs(uipath-data-fabric): cover choice-set + relationship fields + CS…

AditiGoyalUipath:
…V silent-drop

Treat CHOICE_SET_SINGLE / CHOICE_SET_MULTIPLE / RELATIONSHIP as complex field types (extras + lookup tokens), aligned with the existing type table — not as first-class concepts.

Skill changes:
- SKILL.md: 3 new Critical Rules (complex-field extras, choice-sets read-only, lookup-token value form); Rule 7 extended with the CSV silent-drop limitation for complex types
- references/choice-sets.md: new file — browse list/get, value form, decision tree, add-field workflow
- references/entity-schema.md: Choice Set + Relationship sections with the verified entities-get-doesn't-echo-extras finding; combined create example shows scalar options work on complex types
- references/records-query.md: Filtering on Choice-Set Fields (incl. the verified CHOICE_SET_MULTIPLE = vs contains operator semantics + HTTP 400 failure modes), Filtering on Relationship Fields, Writing Choice-Set and Relationship Values
- references/bulk-import.md: records import silently drops complex field types (verified server behavior, BulkUploadCSVHelper) — use records insert --file <json> instead

Test changes:
- smoke_choice_relationship.yaml: complex-field syntax + value form
- smoke_choice_multiple_filter.yaml: CHOICE_SET_MULTIPLE operator wrinkles (contains for membership, = for set equality) + 3 HTTP 400 negative guards
- smoke_csv_complex_drop.yaml: enforces records insert --file <json> for entities with complex fields
- integration_choice_relationship.yaml: live-tenant parent/child + RELATIONSHIP
  + email→UUID resolve, asserts 2 records for Alice
- e2e_expense_tracker.yaml: full Employee→Expense scenario; in operator with UUID valueList; stored-shape assertions
- _shared/cleanup_entities.py: also resolves child_entity_id / parent_entity_id from report.json in FK-safe order

### DIFF
--- a/skills/uipath-data-fabric/SKILL.md
+++ b/skills/uipath-data-fabric/SKILL.md
@@ -16,12 +16,13 @@ All operations go through `uip df <subject> <verb> --output json`.
 
 ## When to Use
 
-- Creating or modifying entity schemas (add fields, update metadata)
+- Creating or modifying entity schemas (add fields of any supported type, update metadata)
 - Reading, inserting, updating, or deleting records
 - Filtering records with complex predicates
 - Computing aggregate metrics for dashboards / KPIs (counts, sums, averages, group-by) — see [references/records-query.md](references/records-query.md#aggregates-server-side)
 - Importing bulk data from CSV files
 - Uploading or downloading file attachments on records
+- Browsing choice sets (`choice-sets list` / `get`) — choice sets are read-only resources from the CLI
 
 ## When NOT to Use — Hand Off to Another Skill
 
@@ -49,6 +50,7 @@ Respond that the operation is not supported. Do not try to work around it.
 | Change a field's data type | Not supported; type is fixed at creation |
 | Create a federated entity | Not supported via CLI or UiPath portal |
 | Write records to a federated entity | Federated entities are read-only |
+| Create / update / delete a choice set | Choice sets are authored in the Data Fabric web UI; the CLI exposes only `choice-sets list` / `choice-sets get` for browsing |
 
 ---
 
@@ -66,7 +68,7 @@ Respond that the operation is not supported. Do not try to work around it.
 
 6. **File fields are separate from record data.** Use `files upload`/`download`, not `records insert`. Field must be type `FILE`.
 
-7. **CSV headers must match exact field names** (case-sensitive). Use `entities get` to discover field names before importing.
+7. **CSV `records import` is Basic-types only.** Headers must match field names exactly (case-sensitive); `CHOICE_SET_*` / `RELATIONSHIP` / `FILE` / `AUTO_NUMBER` columns are silently dropped. For complex fields, use `records insert --file <json>` — see [`references/bulk-import.md`](references/bulk-import.md).
 
 8. **Never create duplicate entities.** Always `entities list` first; reuse if it already exists.
 
@@ -75,6 +77,12 @@ Respond that the operation is not supported. Do not try to work around it.
 10. **Never attempt entity delete.** No command exists. Respond: *"Deleting entities is not supported via the CLI."*
 
 11. **Never attempt field delete.** Do not pass `removeFields` in `entities update`. Respond: *"Removing fields is not supported via the CLI."*
+
+12. **Complex field types need extra config and lookups, just like `DECIMAL` needs `decimalPrecision`.** `CHOICE_SET_SINGLE` / `CHOICE_SET_MULTIPLE` require `choiceSetId` (from `choice-sets list`); `RELATIONSHIP` requires `referenceEntityName` (target's technical `Name`) + `referenceFieldName` (usually `Id`), and the target entity must exist first. Full shape in [`references/entity-schema.md`](references/entity-schema.md).
+
+13. **`choice-sets` is read-only.** CLI has only `list` / `get` — author choice sets in the Data Fabric web UI. If a needed choice set is missing, stop and ask; do not fall back to `STRING`.
+
+14. **Choice / relationship record values use lookup tokens, not labels.** Choice value → integer `NumberId` (single) or array of `NumberId`s (multi), from `choice-sets get`. Relationship value → target record's UUID `Id` regardless of `referenceFieldName`. Filter / `groupBy` use the same tokens; `CHOICE_SET_MULTIPLE` filtering has special operator semantics — see [`references/records-query.md`](references/records-query.md#filtering-on-choice-set-fields).
 
 ---
 
@@ -115,6 +123,8 @@ uip df records query <entity-id> \
   --output json
 ```
 
+For `CHOICE_SET_SINGLE` / `CHOICE_SET_MULTIPLE` / `RELATIONSHIP` field shapes and value formats, see [`references/entity-schema.md`](references/entity-schema.md#supported-field-types) and [`references/records-query.md`](references/records-query.md#filtering-on-choice-set-fields).
+
 ---
 
 ## Task Navigation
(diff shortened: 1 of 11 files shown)


tests/tasks/uipath-data-fabric/e2e_expense_tracker.yaml
```diff
@@ -0,0 +1,306 @@
...
+  Finance needs an expense tracker in UiPath Data Fabric where each Expense is
+  submitted by an Employee, has a category from a fixed picklist (choice set),
+  and can carry one or more tags from a second picklist. The Expense entity
+  combines plain types (`DECIMAL`, `MULTILINE_TEXT`) with complex types
+  (`RELATIONSHIP` to Employee, `CHOICE_SET_SINGLE` for category, optional
+  `CHOICE_SET_MULTIPLE` for tags) — treat them the same way you'd treat any
+  other type, just with the extra config + lookup tokens they require.
+
+  Load the uipath-data-fabric skill before starting.
+  Use --output json on every uip df command.
```
SuphalAthlur, line 33: We shouldn't expect the prompt to contain platform implementational details like these right?

> The Expense entity combines plain types (`DECIMAL`, `MULTILINE_TEXT`) with complex types (`RELATIONSHIP` to Employee, `CHOICE_SET_SINGLE` for category, optional `CHOICE_SET_MULTIPLE` for tags)

A user shouldn't be expected to bring out distinctions between plain and complex types, naming the field type is still okay but categorising them based on how these types are implemented is something a platform dev would know.

> Load the uipath-data-fabric skill before starting.

same feedback for this as well, agent should infer this.

tests/tasks/uipath-data-fabric/integration_choice_relationship.yaml
```diff
@@ -0,0 +1,256 @@
...
+
+initial_prompt: |
+  Before starting, load the uipath-data-fabric skill and follow its workflow.
+
+  Use --output json on every uip df command.
+
+  This test exercises two complex field types — `CHOICE_SET_SINGLE` and
+  `RELATIONSHIP` — end-to-end. They're regular fields in the type table that
+  happen to need lookup tokens for their values (a `NumberId` integer for
+  choice; the target record's UUID `Id` for relationship).
```
SuphalAthlur: Same

tests/tasks/uipath-data-fabric/integration_choice_relationship.yaml
```diff
@@ -0,0 +1,256 @@
...
+
+  1. Check login: run `uip login status --output json`. Do NOT re-login.
```
SuphalAthlur, line 37: Here as well

SuphalAthlur (changes requested):
Could you please check the other files as well, and see if this feedback would apply?

skills/uipath-data-fabric/references/bulk-import.md
```diff
@@ -29,6 +29,12 @@ Charlie,74,false,2024-03-05
 
 For an entity with fields: `Name` (STRING), `Score` (INTEGER), `Active` (BOOLEAN), `CreatedDate` (DATE).
 
+### Complex Field Types Are Silently Dropped
+
+`records import` only processes Basic types (`STRING`, `INTEGER`, `DECIMAL`, `BOOLEAN`, `DATE`, `DATETIME`, `MULTILINE_TEXT`, `UUID`). Columns for `CHOICE_SET_*`, `RELATIONSHIP`, `FILE`, and `AUTO_NUMBER` are accepted in the header but the row values are discarded — no error, nothing in `ErrorFileLink`, just `null` on every imported row (or row failure if the field is `isRequired` without a `defaultValue`).
+
+For entities with any complex field, use `records insert --file <json>` instead — the insert endpoint handles all types. See [`records-query.md`](records-query.md#writing-choice-set-and-relationship-values) for the value form.
```
SuphalAthlur, line 38: nit: I think an explicit call-out on what comes under complex types would help here

skills/uipath-data-fabric/references/choice-sets.md
```diff
@@ -0,0 +1,62 @@
+# Choice Sets Reference
```
SuphalAthlur, line 1: These docs look good to me!

skills/uipath-data-fabric/SKILL.md
```diff
@@ -115,6 +123,8 @@ uip df records query <entity-id> \
   --output json
 ```
 
+For `CHOICE_SET_SINGLE` / `CHOICE_SET_MULTIPLE` / `RELATIONSHIP` field shapes and value formats, see [`references/entity-schema.md`](references/entity-schema.md#supported-field-types) and [`references/records-query.md`](references/records-query.md#filtering-on-choice-set-fields).
```
SuphalAthlur: nit: brand them as "complex types" here as well so the terminology is consistent from the other references docs

SuphalAthlur (commented):
🙌

tests/tasks/uipath-data-fabric/e2e_expense_tracker.yaml
AditiGoyalUipath: There are different kind of prompts expected . The skills can be used by any customer, a dev who can give detailed prompt as well as normal simple functional prompt as well . we need to have breadth test around prompt as well to validate Correct output schema is generated.

AditiGoyalUipath: Load the uipath-data-fabric skill before starting.

This makes sense. We should remove this . Despite skills we should be able to load the right skills

SuphalAthlur: > The skills can be used by any customer, a dev who can give detailed prompt as well as normal simple functional prompt as well

This should be geared towards solution developers right, not platform developers. "complex types", "plain types" are implementational details of how we model relationships and choiceset type fields. I understand what you mean by "complex" because I've worked on the Product, otherwise from an external user's pov these labels are arbitrary - plain/complex is neither intuitive nor documented anywhere in UiPath Data Fabric public docs.

IMO the most "technically complete" prompts should be what an end to end solution developer would write, NOT something a developer building Data Fabric platform would write.

AditiGoyalUipath: Is your concern is using Plain and complex terms here? The primitive/complex split is general schema terminology (JSON Schema, XML Schema, OData all use some form of it), not only Data Fabric specific. This e2e is intentionally on the technical end

SuphalAthlur: > technical jargons

The discussion here isn't about including or excluding technical jargon, it is about what does "technical" mean here. As I mentioned earlier: 

> IMO the most "technically complete" prompts should be what an end to end solution developer would write, NOT something a developer building Data Fabric platform would write.

---
> This are not specific to datafabric . It is datastore wide terminology

That is not the point, inferring plain vs complex still requires solution devs to understand Data Fabric's internal type taxonomy - these details are not meant to be exposed to the end user IMO.

SuphalAthlur (approved):
Discussed offline and aligned on removing basic vs complex categorisation since it is dependent on DF's internal taxonomy.
Users do not know if the field is implemented as a basic type or a complex type (choicesets can be modelled as a constraint on a basic field, as well as a complex field, for example - solution devs aren't expected to know this).
Will be taken up as a follow up PR to clean up skills.

PR #969, opened 2026-05-22
refactor(tests): consolidate task sandbox to experiment layer; install @alpha CLI

bai-uipath:
## Why

**Sandbox config was duplicated across every task.** 254 of 367 task YAMLs carried a `sandbox:` block, mostly to override the experiment's `driver: docker` back to `tempdir` with empty `python: {}` / `node: {}` declarations. Most of that was noise — purely cosmetic overrides with no per-task semantics — and made it hard to change driver behavior centrally.

## What

**Sandbox flatten.** Driver selection and runtime declarations move up to `smoke.yaml` / `nightly.yaml`. Task-level `sandbox:` now only carries what the experiment provably can't express:

- `template_sources` — per-task fixture seeding (13 tasks)
- `mock_path_dirs` + `template_sources` — mock-CLI dispatcher for troubleshoot scenarios (19 tasks)
- `node.env_packages` — per-task npm installs for test-manager tooling (7 tasks)

Everything else (215 tasks) drops the block entirely. No task overrides `driver` anymore.

**CLI install fix.** Dockerfile writes the `@uipath` registry + token into npm's actual global config path (`npm config get globalconfig` — `/usr/etc/npmrc` in the coder-eval-agent base image, not `/usr/local/etc/npmrc`). The token bakes in at build time from `--build-arg NPM_AUTH_TOKEN`, so no runtime env forwarding is required. The image is rebuilt fresh on every nightly + GH Actions run, so token rotation propagates within 24h.

## Tested

Ran a partial smoke on the coder-eval-runner VM (coder_eval main + this branch). Killed after 30 completions.

**Docker driver active for every completion.** Host-side `isolation.docker_runner: Running task ... in docker:` markers cover all 30 completed task_ids (cross-checked from the eval log). Live `docker ps` during the run showed 20 parallel `coder-eval-skill-*` containers on `skills-image:latest`. The `task.json` `driver=tempdir` field is the in-container orchestrator's self-view — by design (`run_task_internal_command.py:147-151`) the in-container driver is rewritten to tempdir since we're already in the container.

**Toolchain verified inside a live container:**

| | version |
|---|---|
| `uip` | `1.1.0-alpha.20260521.7245` (= @alpha, not 1.0.4) |
| `dotnet` | `8.0.421` |
| `node` | `v22.22.2` |
| `python3` | `3.13.13` |
| `claude` | `2.1.145` |

`uip admin --help` triggered the lazy install of `@uipath/admin-tool` from GitHub Packages successfully — confirms the npm registry config + PAT auth baked into the image at build time are functional.

**Failure breakdown (5 of 30 not at full score):**

- 2 pre-existing: `skill-platform-orchestrator-processes-list` (0.25) and `skill-hitl-quality-dev-mistake-binding-direction` (0.92) — identical scores in the prior nightly run, not introduced by this PR.
- 3 new (passed in prior run, failed today). Root cause for all 3 is unrelated to this PR:
  - `skill-agent-coded-file-attachment-input` and `skill-agent-coded-tracing-redaction` — `ModuleNotFoundError: No module named '_shared'`. The check scripts compute their `_shared` import path via `dirname(dirname(__file__))`. PR #902 moved them one level deeper (`tests/tasks/uipath-agents/coded/<name>/`) but didn't update the path math. Blast radius: 25 check scripts under `tests/tasks/uipath-agents/coded/` carry the same broken math. Filed as follow-up to PR #902, not addressed here.


tests/docker/Dockerfile
```diff
@@ -3,19 +3,19 @@
...
-# Install @uipath/cli and @uipath/admin-tool from GitHub Packages
-# These are needed by skill smoke tests (orchestrator, integration-service, data-fabric, etc.)
 ARG NPM_AUTH_TOKEN=
 RUN set -euo pipefail && \
     if [ -z "$NPM_AUTH_TOKEN" ]; then \
       echo "Error: NPM_AUTH_TOKEN must be passed as build arg"; exit 1; \
     fi && \
-    npm config set @uipath:registry https://npm.pkg.github.com/ && \
-    npm config set //npm.pkg.github.com/:_authToken "$NPM_AUTH_TOKEN" && \
-    npm install -g @uipath/cli@alpha @uipath/admin-tool@alpha && \
```
akshaylive, line 15: I'm guessing this removal was intentional?

akshaylive (approved):
Minor verification required but LGTM otherwise

tests/docker/Dockerfile
bai-uipath: yes, it was incorrect to install these in the dockerfile, should be handled by uip's autoinstall mechanism.

PR #972, opened 2026-05-22
GH-hosted Run Coder Eval workflow

bai-uipath:
## What

GH-hosted Actions workflow that runs `coder-eval` against the skills task tree (`tests/tasks/...`). Triggerable from the Actions tab with a single task, a folder, or — with explicit confirmation — the full suite that the VM cron runs nightly.

## Why

Long-term replacement for the `coder-eval-runner` VM. The VM is expensive, requires SSH to debug, accumulates stale system state, and runs at j=1 (~3-4h for the full suite). GH-hosted runners are ephemeral, let anyone trigger ad-hoc evals without VM access, and run at j=20.

## Design

- Builds the same `skills-image:latest` Docker image used by `smoke-skills.yml` and `daily.sh` (`tests/docker/Dockerfile` + `@uipath/cli@alpha` from GH Packages)
- Mints ROPC bot-user auth via `coder_eval/dashboard/scripts/ci/refresh-auth.sh` — `uip flow debug` needs a real user `sub` claim, which `client_credentials` tokens lack. Writes `~/.uipath/.auth`; `experiments/nightly.yaml` mounts it into each eval container
- Runs `coder-eval run <globs> -e experiments/nightly.yaml -j 20 --run-dir /tmp/runs`
- Final step parses `task.json` files and emits `task_id: STATUS` per task plus `N/M PASS`; exits nonzero if any task failed
- Guardrail: globs matching >50 tasks require ticking `confirm_large_run`

## Testing

End-to-end validated by temporarily enabling `pull_request:` + defaulting the calculator task on this PR (both reverted before merge). Final clean run on `bai/nightly-on-gh-actions` — [run 26317972599](https://github.com/UiPath/skills/actions/runs/26317972599):

- Duration: 5m25s
- Result: `skill-flow-calculator: SUCCESS` → `1/1 PASS`
- `@uipath/cli@alpha` resolved to `1.1.0-alpha.20260521.7245` from GH Packages — not the stale npmjs `1.0.0-alpha.20260422.4672` that pins `@uipath/maestro-tool@1.0.4` (a stable not on GH Packages, which caused an `ETARGET` failure during earlier debug runs)
- Container CLI matched the host docker build (`docker_runner: Version: 1.1.0-alpha.20260521.7245`)
- ROPC auth propagated through the mount — `uip flow debug` worked end-to-end

## Known dependencies

- Currently runs on `ubuntu-latest` (2 vCPU / 7 GB RAM). For the full nightly sweep the VM remains a better target until `UiPath/skills` is allowlisted for the `ubuntu-latest-16-cores` runner group. Switch is a one-line YAML change once granted.
- Repo secrets added for this workflow: `UIPATH_ROPC_CLIENT_ID`, `UIPATH_ROPC_CLIENT_SECRET`, `UIPATH_BOT_USERNAME`, `UIPATH_BOT_PASSWORD` (existing secrets for `GH_PAT`, AWS Bedrock, UV index reused)

### DIFF
--- a/.github/workflows/nightly-skills-gh.yml
+++ b/.github/workflows/nightly-skills-gh.yml
@@ -1,0 +1,221 @@
+name: Nightly Skill Tests (GH-hosted)
+
+# GH-hosted mirror of the VM cron (coder_eval/dashboard/scripts/ci/daily.sh).
+# Runs the full skills task tree against experiments/nightly.yaml on the
+# largest GH-hosted tier in UiPath (ubuntu-latest-16-cores = 16 vCPU /
+# 64 GB RAM). Manual dispatch — paired with smoke-skills.yml's per-PR
+# subset. Eventually intended to replace the coder-eval-runner VM cron;
+# this draft is the feasibility validation.
+
+on:
+  workflow_dispatch:
+    inputs:
+      task_globs:
+        description: 'space-separated globs under tests/ (e.g. "tasks/uipath-agents/**/*.yaml"). Default: full tree minus Windows-only RPA + opt-in benchmarks.'
+        type: string
+        default: ''
+      runner_size:
+        description: 'GH-hosted runner tier'
+        type: choice
+        default: ubuntu-latest-16-cores
+        options:
+          - ubuntu-latest-16-cores
+          - ubuntu-latest
+      task_parallelism:
+        description: 'Concurrent tasks (-j). 4-6 is a safe ceiling for docker-driver tasks on 16-cores.'
+        type: string
+        default: '4'
+      model_override:
+        description: 'Override BEDROCK_MODEL (full Bedrock id, e.g. eu.anthropic.claude-sonnet-4-6). Blank → repo secret.'
+        type: string
+        default: ''
+
+concurrency:
+  group: nightly-skills-gh-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  nightly:
+    runs-on: ${{ inputs.runner_size }}
+    # Full skills suite on the VM (~700 tasks, j=1) runs 3-4h. The 16-core
+    # tier with j=4 should land ~1-1.5h; 330 min covers worst-case stalls.
+    timeout-minutes: 330
+    name: Run nightly suite
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/checkout@v4
+        with:
+          repository: UiPath/coder_eval
+          token: ${{ secrets.GH_PAT }}
+          path: .coder_eval
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - uses: astral-sh/setup-uv@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Verify Docker is available
+        run: docker info > /dev/null
+
+      - name: Free up disk
+        run: |
+          df -h /
+          docker system prune -af --volumes >/dev/null 2>&1 || true
+          sudo rm -rf /opt/hostedtoolcache/CodeQL 2>/dev/null || true
+          df -h /
+
+      - name: Install coder-eval
+        working-directory: .coder_eval
+        env:
+          UV_EXTRA_INDEX_URL: "https://${{ secrets.UV_INDEX_UIPATH_USERNAME }}:${{ secrets.UV_INDEX_UIPATH_PASSWORD }}@uipath.pkgs.visualstudio.com/_packaging/ml-packages/pypi/simple/"
+        run: uv pip install --system .
+
+      - name: Build coder-eval Docker image
+        working-directory: .coder_eval
+        env:
+          UV_INDEX_UIPATH_USERNAME: ${{ secrets.UV_INDEX_UIPATH_USERNAME }}
+          UV_INDEX_UIPATH_PASSWORD: ${{ secrets.UV_INDEX_UIPATH_PASSWORD }}
+        run: make docker-image
+
+      # Build the skills extension image (`@uipath/cli` + `@uipath/admin-tool`
+      # @alpha on top of coder-eval-agent). Matches smoke-skills.yml + daily.sh.
+      - name: Build skills Docker image
+        run: |
+          docker build \
+            --build-arg CODER_EVAL_IMAGE=coder-eval-agent:latest \
+            --build-arg NPM_AUTH_TOKEN="${{ secrets.GH_PAT }}" \
+            -t skills-image:latest \
+            -f tests/docker/Dockerfile \
+            .
+
+      - name: Mint UiPath service-account token + enable env-auth
+        env:
+          UIPATH_CLIENT_ID:     ${{ secrets.UIPATH_CLIENT_ID }}
+          UIPATH_CLIENT_SECRET: ${{ secrets.UIPATH_CLIENT_SECRET }}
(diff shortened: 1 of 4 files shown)


uipreliga (approved):
🚢

uipreliga:
## Code Review — `bai/nightly-on-gh-actions`

Multi-model review (Opus 4.6 + Gemini-3 + GPT-5.3 codex, all three running in parallel). Scoped to the three files actually changed by this PR vs. `main`: `.github/workflows/run-coder-eval.yml`, `tests/docker/Dockerfile`, `README.md`. (The big `skills/uipath-maestro-case/**` deletions in `main..HEAD` are not from this branch — they're commits already on `main` that this branch is behind; ignored.)

Findings are against **PR head `49d5a2ef`** (the post-push version with the env-var injection fix + artifact upload).

### Already addressed in 49d5a2ef ✅

- 🔴 **Shell injection in `Run coder-eval` step** — `${{ steps.globs.outputs.globs }}` was being interpolated directly into the `run:` block, which is the classic GH Actions script-injection vector against user-supplied `workflow_dispatch` input. Fixed by routing through `TASK_GLOBS` / `TASK_COUNT` env vars. Comment on lines 138–139 explains the rationale — keep it; future-you will appreciate it.
- 🟡 **Missing artifact upload** — added `Upload eval report` step.

### Still outstanding

#### 🔴 CRITICAL — `$GITHUB_OUTPUT` newline injection (`run-coder-eval.yml:92`)

```bash
echo "globs=$INPUT_GLOBS" >> "$GITHUB_OUTPUT"
```

`workflow_dispatch` text inputs are not stripped of newlines. A `task_globs` value like `tasks/foo.yaml\ncount=999999` lets an attacker:

1. Override the `count` output written on the next line, bypassing the `> 50` `confirm_large_run` gate.
2. More generally, write arbitrary additional step outputs that downstream steps consume.

Both Gemini and Codex flagged this independently. Use the heredoc form:

```bash
{
  echo "globs<<__EOF__"
  echo "$INPUT_GLOBS"
  echo "__EOF__"
} >> "$GITHUB_OUTPUT"
echo "count=$COUNT" >> "$GITHUB_OUTPUT"
```

…or reject newlines/control chars explicitly in the validator (`case "$INPUT_GLOBS" in *$'\n'*) echo "::error::no newlines"; exit 1;; esac`). Either works; the heredoc form is more robust and matches what the env-var fix on line 138 was philosophically trying to do.

#### 🟠 HIGH — Verdict step exit-code propagation is fragile (`run-coder-eval.yml:196–209`)

```yaml
- name: Plaintext summary + verdict
  if: always()
  run: |
    set -uo pipefail        # no -e
    run_dir=$(...)
    ...
    python - "$run_dir" <<'PY' | tee -a "$GITHUB_STEP_SUMMARY"
    ...
    sys.exit(0 if rows and ok == len(rows) else 1)
    PY
```

The Python exit code propagates today only because the `python | tee` pipeline is the **last** command in the step and bash uses `$?` of the last command. Add `set -euo pipefail` so the contract is explicit — otherwise any future line appended after the heredoc (e.g. an innocent-looking `echo "done"`) silently masks a Python `sys.exit(1)`.

Also: `if: always()` here will run even when `Run coder-eval` itself failed catastrophically. On ephemeral `ubuntu-latest` this is fine (no stale `/tmp/runs` from prior jobs), but if this is ever ported to a self-hosted runner — possible per the 16-core TODO — a stale run dir from a previous job could produce a false PASS verdict. Either short-circuit on `steps.eval.outcome != 'success'`, or document the assumption.

#### 🟡 MEDIUM — `Fix permissions on /tmp/runs` short-circuits and swallows errors (`run-coder-eval.yml:162`)

```bash
sudo chown -R $(id -u):$(id -g) /tmp/runs && sudo chmod -R 755 /tmp/runs 2>/dev/null || true
```

- `&&` short-circuits `chmod` if `chown` fails (e.g. `/tmp/runs` missing because `coder-eval` crashed early).
- `2>/dev/null` is on `chmod` only — `chown` stderr still surfaces.
- `|| true` swallows any remaining signal.

The same pattern (and same quirk) is inherited from `smoke-skills.yml:295`. Standardize either way:

```bash
[ -d /tmp/runs ] && {
  sudo chown -R "$(id -u):$(id -g)" /tmp/runs 2>/dev/null || true
  sudo chmod -R 755 /tmp/runs            2>/dev/null || true
}
```

#### 🟡 MEDIUM — Duplication with `smoke-skills.yml`

Both workflows share: checkout-`UiPath/coder_eval`, `setup-python@v5`, `setup-uv@v4`, "Install coder-eval", "Build coder-eval Docker image", "Build skills Docker image", "Fix permissions on /tmp/runs". A composite action (`.github/actions/setup-coder-eval/`) would DRY this. Not blocking — but two near-identical workflows side-by-side will rot unevenly. Worth a follow-up ticket.

#### 🟢 LOW

- **Hardcoded `-j 20`** (`run-coder-eval.yml:157`) — `smoke-skills.yml` uses `vars.TASK_PARALLELISM` (or a workflow input). Reasonable to defer until the 16-core tier lands, but make a config knob the same day to avoid a code change per tuning.
- **README heading placement** (`README.md:125`) — "Running coder-eval on demand" sits under `## Contributing`. It's an operational capability, not a contribution-flow doc. Consider a new `## Operations` section or move under `## Resources`. Also the sentence-case heading is inconsistent with surrounding title-case peers.
- **Glob expansion edge cases** (`run-coder-eval.yml:77–81`) — `for pat in $INPUT_GLOBS; do for f in $pat; do` is intentional word-splitting and is correct given the repo convention (no spaces in task paths, no brace expansion needed). Worth a one-line comment so the next reader doesn't "fix" it.

### Dockerfile (`tests/docker/Dockerfile`)

`ARG CLI_VERSION=alpha` with `@uipath/cli@${CLI_VERSION}` is **backwards-compatible** — `smoke-skills.yml` doesn't pass the new arg and falls back to the `alpha` default, so existing behavior is preserved. Clean change.

> Out-of-scope finding from Codex worth noting separately: `dotnet-install.sh` is fetched over the network and executed without integrity verification (pre-existing code, not in this PR's diff). Address in a separate change if relevant.

### Positive observations

- The Critical Rules of GH Actions security — env-scoped user input + large-run gate — are applied thoughtfully. The env-var fix on the `Run coder-eval` step landed before review even completed; the author clearly knew the pattern.
- Header comment (lines 3–6), 16-core TODO (lines 34–38), timeout justification (lines 40–42), ROPC rationale (lines 122–125) all explain **why**, not just **what**. Future on-call will thank you.
- 50-task confirmation guardrail is real cost-aware CI design — the kind of "click and regret" failure that's expensive with `coder-eval`.
- All actions pinned to major (`@v4`, `@v5`) consistent with the rest of the repo.

### Top 3 priority fixes

1. 🔴 Switch `$GITHUB_OUTPUT` writes to heredoc form (or reject newlines in the validator) — last remaining injection vector.
2. 🟠 Add `set -e` (or full `set -euo pipefail`) to the verdict step so exit-code propagation isn't position-dependent.
3. 🟡 Either fix or document the `chown && chmod 2>/dev/null || true` short-circuit; mirror the fix back into `smoke-skills.yml` to keep the two in sync.

### Overall

The latest push (49d5a2ef) addresses the two showstoppers from the first round of review. With the `$GITHUB_OUTPUT` newline-injection fix this is mergeable; the rest are quality nits worth folding in either here or as a follow-up.

PR #990, opened 2026-05-25
fix(uipath-agents): move ContextGroundingRetriever inside node bodies and add folder_path

EusebiuJecan07:
## Summary

- Moved `ContextGroundingRetriever` instantiation from module level into node/function bodies in the LangGraph Integration example — this is the pattern the lazy-init rule requires and what agents should follow
- Added `folder_path=` to three examples that were missing it (LangGraph Integration, Agent Integration `@tool`, QA System pattern)

## Why

The `context-grounding.md` LangGraph Integration example showed the retriever constructed at **module level without `folder_path`**. Agents reading this example would either omit `folder_path` or place the retriever outside a node body, leading to ~50% failure rate on the `rag_langgraph` eval. Moving instantiation inside the node and consistently including `folder_path=` closes that ambiguity.

## Test plan

- [ ] Ran `rag_langgraph` eval 5 times consecutively — 5/5 passed (score 1.0 each)
- [ ] No change to eval task YAML or check script in this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #991, opened 2026-05-25
test(uipath-agents): relax rag_langgraph check to accept inline literals and constants

EusebiuJecan07:
## Summary

- Splits the `index_name` / `folder_path` assertions in `check_rag_langgraph.py` into two checks each: one verifying the keyword arg name is present, one verifying the value string appears anywhere in the file
- Removes the requirement for inline string literals, accepting constant-variable patterns as well

## Why

The previous regex `index_name\s*=\s*["\']company_docs["\']` required the value to be an inline literal on the same token. When an agent wrote clean constant-variable style:

```python
INDEX_NAME = "company_docs"
retriever = ContextGroundingRetriever(index_name=INDEX_NAME, folder_path=FOLDER)
```

…the check failed despite correct behavior. This caused ~50% failure rate on the eval.

The relaxed check still enforces that `index_name` and `folder_path` are used as **keyword arguments** (required by the bindings scanner to auto-detect resources), just no longer requires the value to be an inline literal.

Companion PR with the skill-doc fix: #990

## Test plan

- [x] Ran `rag_langgraph` eval 5 times — 5/5 passed (score 1.0 each) with both changes applied

### DIFF
--- a/tests/tasks/uipath-agents/coded/rag_langgraph/check_rag_langgraph.py
+++ b/tests/tasks/uipath-agents/coded/rag_langgraph/check_rag_langgraph.py
@@ -62,10 +62,14 @@ def check_imports_and_calls(text: str) -> None:
             "`uipath_langchain.retrievers` — the canonical path the skill teaches."
         )
     print("OK: main.py imports ContextGroundingRetriever from uipath_langchain.retrievers")
-    if not re.search(r'index_name\s*=\s*["\']company_docs["\']', text):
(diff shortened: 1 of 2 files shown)


radugheo (approved):

radu-mocanu (approved):

cosmyo (approved):

EusebiuJecan07:
Superseded by a branch-based PR from UiPath/skills directly (CI secrets requirement).

PR #1019, opened 2026-05-26
ci(nightly): forward E2E_PROCESS_KEY + E2E_LONG_PROCESS_KEY secrets

alexenica:
## Summary

Follow-up to #815. The 16 lifecycle E2E tests merged in that PR read
\`E2E_PROCESS_KEY\` and \`E2E_LONG_PROCESS_KEY\` from env in their
\`pre_run\` (\`seed.py\`). The nightly cron workflow only forwarded the
legacy \`TRACES_SMOKE_PROCESS_KEY\` — adding the two new ones now.

The secrets themselves must be added to repo settings before nightly
picks them up. Provisioning steps documented in \`tests/fixtures/
packages/README.md\`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1023, opened 2026-05-26
test(uipath-platform): nest test-created folders under E2E_TEST_PARENT_FOLDER (subsumes #1019)

alexenica:
## Summary

Follow-up to #815. Tests that create their own folders (folders_hierarchy,
deploy_round_trip, asset_share's folder B, asset_scoping's folder B)
currently hardcode \`Shared/\` as the parent. On the nightly CI tenant
that means test scratch folders briefly appear at the root of \`Shared\`
between create and \`post_run\` cleanup — noisy on a shared tenant.

Threads a single \`E2E_TEST_PARENT_FOLDER\` env var (default \"Shared\")
through:

- \`seed.py\` writes it as \`parent_folder_path\` in \`seed.json\`
- \`folders_hierarchy\` + \`deploy_round_trip\` prompts + checks read it as
  the parent for folders they create
- \`asset_share\` + \`asset_scoping\` nest folder B inside folder A directly
  (no env var needed — folder A IS already the process's folder)
- \`nightly.yaml\` forwards the env var through the docker sandbox
- \`run-coder-eval.yml\` forwards the secret (also picks up the
  \`E2E_PROCESS_KEY\` / \`E2E_LONG_PROCESS_KEY\` pair previously proposed in
  #1019 — this PR subsumes that one)

## CI tenant setup

On \`alpha codereval/DefaultTenant\` the dedicated isolation folder
\`Shared/uipath-platform-e2e\` is already provisioned (process + machine
assignment done). Three secrets to add to \`UiPath/skills\`:

\`\`\`
E2E_PROCESS_KEY        = 56451439-d1d0-49c0-8cf0-1026425c8da3
E2E_LONG_PROCESS_KEY   = c593051c-184a-4cea-93c4-a8b349a382db
E2E_TEST_PARENT_FOLDER = Shared/uipath-platform-e2e
\`\`\`

Local devs can leave it unset; the default \"Shared\" matches prior
behaviour.

### DIFF
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -176,6 +176,9 @@ jobs:
           BEDROCK_MODEL:            ${{ secrets.BEDROCK_MODEL }}
           ANTHROPIC_API_KEY:        ${{ secrets.ANTHROPIC_API_KEY }}
           TRACES_SMOKE_PROCESS_KEY: ${{ secrets.TRACES_SMOKE_PROCESS_KEY }}
+          E2E_PROCESS_KEY:          ${{ secrets.E2E_PROCESS_KEY }}
+          E2E_LONG_PROCESS_KEY:     ${{ secrets.E2E_LONG_PROCESS_KEY }}
+          E2E_TEST_PARENT_FOLDER:   ${{ secrets.E2E_TEST_PARENT_FOLDER }}
           TASK_GLOBS:               ${{ needs.partition.outputs.linux_globs }}
           TASK_COUNT:               ${{ needs.partition.outputs.linux_count }}
         working-directory: tests
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -178,7 +178,6 @@ jobs:
           TRACES_SMOKE_PROCESS_KEY: ${{ secrets.TRACES_SMOKE_PROCESS_KEY }}
           E2E_PROCESS_KEY:          ${{ secrets.E2E_PROCESS_KEY }}
           E2E_LONG_PROCESS_KEY:     ${{ secrets.E2E_LONG_PROCESS_KEY }}
-          E2E_TEST_PARENT_FOLDER:   ${{ secrets.E2E_TEST_PARENT_FOLDER }}
           TASK_GLOBS:               ${{ needs.partition.outputs.linux_globs }}
           TASK_COUNT:               ${{ needs.partition.outputs.linux_count }}
         working-directory: tests
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -315,7 +315,6 @@ jobs:
             echo "TRACES_SMOKE_PROCESS_KEY=${{ secrets.TRACES_SMOKE_PROCESS_KEY }}"
             echo "E2E_PROCESS_KEY=${{ secrets.E2E_PROCESS_KEY }}"
             echo "E2E_LONG_PROCESS_KEY=${{ secrets.E2E_LONG_PROCESS_KEY }}"
-            echo "E2E_TEST_PARENT_FOLDER=${{ secrets.E2E_TEST_PARENT_FOLDER }}"
           } >> "$GITHUB_ENV"
           mkdir ~/.uipath
           run_uip_with_tenant "$TENANT_NAME" login status --output json
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -313,8 +313,6 @@ jobs:
             echo "UIPATH_CLI_TENANT_NAME=$TENANT_NAME"
             echo "UIPATH_CLI_TENANT_ID=$TENANT_ID"
             echo "TRACES_SMOKE_PROCESS_KEY=${{ secrets.TRACES_SMOKE_PROCESS_KEY }}"
-            echo "E2E_PROCESS_KEY=${{ secrets.E2E_PROCESS_KEY }}"
-            echo "E2E_LONG_PROCESS_KEY=${{ secrets.E2E_LONG_PROCESS_KEY }}"
           } >> "$GITHUB_ENV"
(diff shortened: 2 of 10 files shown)


puscasu-ion-daniel (approved):

bai-uipath:
Nice isolation fix — but I'd **drop the `E2E_TEST_PARENT_FOLDER` env var and hardcode the parent folder** instead.

**Why the env var isn't worth it:**
- It's a plain path, not a secret — the secret+passthrough pattern fits `E2E_PROCESS_KEY`, not a folder name.
- One CI tenant; nothing needs it to vary. YAGNI.
- The value must be re-set at *every* launcher: `run-coder-eval.yml`, `smoke-skills.yml`, **and the VM `.env`** (nightly = `daily.sh` → `nightly.yaml`). Setup notes only cover the secrets → **on merge this is a no-op on the real nightly until someone hand-edits the VM `.env`**, else it silently pollutes Shared again.
- The `VAR=$VAR python3` prefix is what caused the empty-string bug you patched with `or "Shared"`.

**Simpler — lands on merge everywhere, zero secret/`.env` coordination:**
1. Keep `parent_folder_path` in `seed.json` (good — one source of truth for prompt + checker).
2. Set it from a constant in `seed.py`: `PARENT = "Shared/uipath-platform-e2e"`.
3. Have `seed.py` ensure that folder exists (get → create if missing) so it works on any tenant.

Deletes: the secret, both workflow `env` additions, the `nightly.yaml` passthrough, the prefix, the footgun. Prompts/checks unchanged.

**Also:**
- `smoke-skills.yml` lines are dead — smoke runs `--tags smoke`, these are `e2e`. Drop them.
- Confirm `uip solution deploy` can put the folder under a non-`Shared` parent, or `check_deploy_round_trip.py` fails when parent ≠ `Shared`.

bai-uipath (approved):

uipreliga:
## Code Review — PR #1023 (nest E2E folders under hardcoded parent)

Reviewed by 3 models in parallel (Gemini 3, GPT-5 Codex, Opus). Comment-only — no files changed. No Critical/High issues. The hardcode-over-env-var decision is sound (YAGNI for a single CI tenant) and `ensure_parent_folder` is appropriately KISS.

### Medium

- **Race in `ensure_parent_folder` under parallel runs** — `tests/tasks/uipath-platform/seed.py:31-35` _(flagged independently by Gemini + Codex)_. Every platform E2E `seed.py` now does get→create-if-missing on the **same** shared parent `Shared/uipath-platform-e2e`. With `TASK_PARALLELISM > 1` (nightly), two seeds can both see it missing and both `folders create`; the loser gets non-Success and `exit(1)`, failing a healthy task. Narrow (only the first-run-of-the-day window, before the parent persists), but cheap to harden: on create failure, re-`get` and return if it now exists.

- **Nested per-run folders may escape cleanup** — `tests/tasks/uipath-platform/cleanup.py:64-71` _(flagged by Opus)_. Cleanup deletes folders whose **name** contains `uuid8` via `or folders list --limit 200`. `folders_hierarchy`'s `e2e-fh-<uuid8>-*` tree now lives _under_ the persistent shared parent instead of top-level in `Shared/`. If `or folders list` returns only top-level folders (not a flattened tree), or the tenant exceeds 200 folders, those children leak permanently under the shared parent. **Please verify `or folders list` enumerates nested folders.** Note: `deploy_round_trip`'s folder is safe regardless — it's removed by the `solution deploy uninstall` cascade (cleanup step 3).

### Low

- **`or "Shared"` fallback now defensive/dead** — `check_folders_hierarchy.py:39`, `check_deploy_round_trip.py:47` _(all 3 reviewers)_. `seed.py` always writes `parent_folder_path`, so the fallback never fires for a current seed. Defensible as graceful degradation for a stale `seed.json`; if you'd rather fail fast on a broken seed contract, drop the `or "Shared"` and assert presence. Either is fine — low priority.

- **`ensure_parent_folder` runs for tasks that never use the parent** — `seed.py:37` + comment at `seed.py:5-9` _(Opus + Codex)_. `asset_scoping`, `asset_share`, and other platform tasks place resources under `folder_a_path`, not `parent_folder_path`, yet every platform `seed.py` now pays a `folders get` (and one-time create) round-trip. Harmless, but the `seed.py:5` comment ("tests place any new folder they create" under it) overstates scope — only `folders_hierarchy` and `deploy_round_trip` actually consume it. (Codex's suggested fix — gate behind an `E2E_REQUIRE_PARENT_FOLDER` env var — is **not recommended**: it reintroduces exactly the env-var coordination this PR deliberately removed.)

- **E2E keys added to smoke-only runner** — `smoke-skills.yml:316-317`. The smoke runner only executes `smoke`-tagged tasks; these keys feed `e2e`-tagged tasks. Harmless but effectively dead config in that file. The `run-coder-eval.yml:179-180` additions are correct and needed.

### Informational / out of scope

- `json.dump(seed, open("seed.json","w"))` at `seed.py:59` has no context manager (Codex). Pre-existing (not introduced by this PR); real risk is negligible for a short-lived CPython script. Skip unless touching the line anyway.

### Positive

- Old `E2E_TEST_PARENT_FOLDER` env approach removed cleanly — no orphaned references (grep-confirmed).
- `E2E_LONG_PROCESS_KEY` is genuinely consumed (`job_control_e2e.yaml:12`), documented (`tests/README.md:243`), and wired through `nightly.yaml:28` — not orphaned despite the recent "drop env-var passthrough" commit.
- `rpartition("/")` split + the `if not parent or not name` guard (seed.py:27-30) handle the path parsing correctly with a clear operator message.
- Hardcode + idempotent-ensure design is correctly KISS — no recursive creation, no speculative generalization, clear stderr+exit(1) failures.

### Overall
Clean, well-scoped change with solid engineering judgment. The only items worth acting on before merge are confirming the nested-folder cleanup path (Medium) and optionally hardening the first-run create race (Medium).

🤖 Multi-model review (Gemini 3 + GPT-5 Codex + Opus), synthesized.

uipreliga (approved):
Fix shat you agree with and 🚢

PR #1031, opened 2026-05-26
fix(tests): accept boolean-decision HITL in DevCon expense approval checker

tmatup:
## Summary

The DevCon expense approval test (`skill-flow-e2e-devcon-expense-approval`) failed in run [`2026-05-26_04-05-02`](https://coder-evalboard.uipath-dev.com/runs/2026-05-26_04-05-02/skill-flow-e2e-devcon-expense-approval) with:

```
FAIL: HITL schema must define at least two outcomes
```

The agent built a structurally valid flow that `uip maestro flow validate` accepted, but the semantic checker rejected it.

## Root cause

`check_devcon_expense_approval.py` contains a logical contradiction:

- **Docstring** (lines 5–8) — "approval can be represented either as a boolean output field or as approve/reject outcomes."
- **Disjunctive check** (lines 86–92) — already enforces exactly that semantic: pass if there's a boolean decision field, *or* if outcomes contain both "approve" and "reject" names.
- **Hard gate** (lines 71–72) — unconditionally required `len(outcomes) >= 2`, making the boolean-only path unreachable.

The HITL skill teaches both idioms (Example 1 — Approve/Reject outcomes for approvals; Example 3 — data-enrichment with a boolean output + Submit). The failing run produced Example 3's shape, which the docstring promised was acceptable but the gate prevented.

## Why this started failing after 4 days of passing

The cause is **not** a skill or doc change — the relevant skill descriptions (`uipath-maestro-flow`, `uipath-planner`, `uipath-human-in-the-loop` YAML frontmatter) are byte-identical between the last passing commit (`d5441840`) and the first failing commit (`eb648ade`). The cause is the coder-eval **runtime environment swap**, recorded in `task.json/environment_info`:

| Field | 05-22 PASS | 05-25 / 05-26 FAIL |
|---|---|---|
| `claude_code_cli` | **2.1.123** | **2.1.145** (+22 patches) |
| `api_routing` | **aws_bedrock** (`eu-north-1`, `eu.anthropic.claude-sonnet-4-6`) | **anthropic_direct** |
| `anthropic` SDK | 0.102.0 | 0.104.1 |
| `pydantic` / `uv` | 2.12.5 / 0.11.8 | 2.13.4 / 0.11.15 |

The shift is directly observable in the agent's **first** tool call:

- **05-22**: `Skill: uipath:uipath-maestro-flow` — straight to the specialist; specialist autonomously chose Approve/Reject outcomes.
[...]

PR #538, opened 2026-05-04
refact(skills): move task-list ownership from solution-design to planner

RaduAna-Maria:
## Summary

  - **Single source of truth for task lists.** `uipath-solution-design` now produces architecture only;
   `uipath-planner` owns implementation tasks. Eliminates duplication between solution-design's
  per-template Implementation Plan sections and planner's plan format.
  - **Load-bearing cross-skill contract.** Every SDD now carries a `## Planner Handoff` header (6
  fields) that the planner detects deterministically. Pipeline: PDD → solution-design (SDD) → planner
  (tasks.md) → execution.
  - **Planner restructured into two lanes.** PDD-driven (reads SDD, 0–2 prompts) and Non-PDD
  (elicitation, 1–5 prompts). SKILL.md slimmed from 341 → 117 lines (net `-224`); 4 new references
  extracted for load-on-demand.

  ## Changes

  ### `uipath-solution-design`

  - Removed `## Implementation Plan` section from all 6 SDD templates (RPA, Agent, Flow, Case, Coded
  App, API Workflow). Replaced with a final numbered `## Next Steps` section pointing the user at
  `uipath-planner`.
  - Added `## Planner Handoff` header (6 fields, `<!-- DO NOT RENAME -->` marker) to all 6 templates,
  immediately after Document History.
  - `SKILL.md`: added `When to Use This Skill`, expanded Workflow section, dropped Critical Rules 8 &
  11 (Implementation Plan ownership / intent-based execution), added Critical Rules around the Planner
  Handoff marker, no Implementation Plan, and thorough testing. Added `package-selection-guide.md` to
  Reference Navigation. Description gains `→uipath-planner` redirect.
  - `sdd-generation-guide.md`: collapsed Phase 1 Steps 4 / 4.25 / 4.3 / 4.5 into one orchestrator step
  linking to canonical level homes. Deleted Phase 3 Step 3 (Create Live Tasks) and Step 4 (Execute
  Implementation Plan). Rewrote Phase 3 Step 2 to write the `## Planner Handoff` header and prompt
  `regenerate / keep / versioned-alongside` on re-run. Added Phase 3 Step 3 (Hand Off to the Planner).
  - `product-selection-guide.md`: explicit canonical home for Levels 1, 1.75, 2.5 Part B, 3; Levels of
  Decision table gains an explicit ownership column. Level 3 references rewritten — flag in template,
  planner picks up.
[...]

PR #1036, opened 2026-05-26
fix(uipath-planner): keep single-project inline-HITL flows out of the planner; never pre-bake HITL schema

tmatup:
## Why

Companion to #1031. That PR fixes the DevCon expense-approval *checker* (which had a contradictory `len(outcomes) < 2` gate). This PR addresses the *upstream agent behavior* that made the checker go red, so the eval is robust regardless of which valid HITL pattern the agent picks.

### Background (full analysis: `docs/uipath-skills/2026-05-26-skill-flow-e2e-devcon-expense-approval-regression.md`)

`skill-flow-e2e-devcon-expense-approval` passed for 4 days then failed on 05-25/05-26. Root cause of the behavior shift was an eval **runtime swap** (Claude Code `2.1.123→2.1.145`, `api_routing` `aws_bedrock→anthropic_direct`) that flipped how `claude-sonnet-4-6` routes the prompt:

- **Before:** first tool call = `uipath-maestro-flow` (direct). The specialist designed the HITL schema from intent → **Approve/Reject outcomes**.
- **After:** first tool call = `uipath-planner`. The model reasoned *"spans multiple skills (solution, flow, human-in-the-loop)"*, routed through the planner, which decomposed `"approved (yes/no)"` into a field-level spec (`approved: boolean, required`) and handed it to the HITL specialist → forced the **single-`Submit` boolean** schema → tripped the checker.

The misclassification is the bug: a Flow with an inline HITL approval step + script nodes + its own solution wrapper is **one** `uipath-maestro-flow` task, not a multi-skill plan. Inline nodes and solution scaffolding are author sub-steps the flow specialist performs itself.

## What

Three layered guards, all within `skills/uipath-planner/`:

1. **`when_to_use` + "When to Use This Skill" (routing-time).** Explicitly scopes the planner to requests spanning **separate buildable projects**, and names the anti-pattern: a single project with inline HITL/script/connector nodes is one specialist task. This is the frontmatter the model reads *before* deciding which skill to invoke. Combined `description`+`when_to_use` = **740 chars**, under the 1,536 listing-truncation cap (validated).
2. **`multi-skill-patterns-guide.md` (post-invoke safety net).** Clarifies inline nodes are not "components"; Pattern 2 fires only for separate buildable `.uipx` projects. If the planner *is* invoked anyway, it emits a single maestro-flow task rather than a separate `uipath-human-in-the-loop` task.
3. **Rule 5 "Route — do not redescribe" (robustness).** Adds HITL field/outcome schema design to the don't-redescribe list. The planner passes **business intent only** ("manager approves or rejects; can add a reason") — never a field-level spec — so the HITL specialist keeps the freedom to choose boolean-field vs Approve/Reject. This makes the *outcome* correct even if routing still lands on the planner.

## Honest scope

LLM routing can't be made deterministic by prompt text — only strongly biased. Guard 1 reduces the chance the planner is invoked; guards 2–3 make the result correct *when* it is. The deterministic guarantee remains #1031 (the test accepts both valid patterns). Defense in depth.

## Blast radius

Guard 1 reshapes planner routing for **all** requests, so the negative scope is written to fire only on *single-project* requests and preserve planner routing for genuinely multi-project flows (Flow + standalone RPA/agents). Worth a careful read from @RaduAna-Maria (CODEOWNERS for `skills/uipath-planner/`).

## Test plan

- [x] `hooks/validate-skill-descriptions.sh skills/uipath-planner/SKILL.md` — passes (229 chars ≤ 1024).
- [x] Combined `description`+`when_to_use` = 740 chars (< 1536 listing cap), so the new negative-scope clause is not truncated in the skill listing.
- [ ] Re-run `skill-flow-e2e-devcon-expense-approval` on the current runtime; confirm first tool call is `uipath-maestro-flow` (or, if planner is still chosen, that the HITL handoff carries intent not a field spec).
- [ ] Spot-check a genuinely multi-project request (Flow + standalone RPA) still routes through the planner.

### DIFF
--- a/skills/uipath-planner/SKILL.md
+++ b/skills/uipath-planner/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: uipath-planner
 description: "UiPath task planner — reads SDDs from uipath-solution or elicits non-PDD requests, derives multi-skill task lists, emits live TaskCreate calls. Detects project type (.cs, .xaml, .flow, .bpmn, .py). For PDDs→uipath-solution first."
-when_to_use: "User makes a non-trivial UiPath request — 'build a UiPath solution for X', 'set up a process from scratch', 'help me plan this' — OR provides an SDD path. Skip when project type and scope are already clear for a single-skill task — invoke the specialist directly."
+when_to_use: "User makes a non-trivial UiPath request that spans SEPARATE buildable projects — e.g. 'build a UiPath solution for X', 'set up a process from scratch', a Flow that orchestrates standalone RPA processes or agents — OR provides an SDD path. Skip when the request targets a SINGLE project, even a Flow/Agent/RPA project with inline HITL, script, or connector nodes wrapped in its own solution (e.g. a Flow with an inline approval step is one uipath-maestro-flow task, not a plan) — invoke that specialist directly."
 allowed-tools: Bash, Read, Glob, Grep, AskUserQuestion, EnterPlanMode, ExitPlanMode, TaskCreate, TaskUpdate, TaskList
 ---
 
@@ -18,12 +18,12 @@ The lane is decided by the **Entry Guard** below.
 
 ## When to Use This Skill
 
-- The request is **non-trivial** — multi-step, multi-skill, UI automation, or unclear scope
+- The request is **non-trivial** — spans **separate buildable projects** that each need their own specialist (e.g. a Flow orchestrating standalone RPA processes or agents), or UI automation with unclear scope
 - The request is **ambiguous** — no single specialist skill clearly matches
 - The user asks "what can I build?" or needs help choosing a project type
 - The user provides an SDD path — Lane A runs
 
-Skip this planner for simple, well-defined single-skill tasks (e.g., "create a workflow that sends an email") — load the specialist directly.
+Skip this planner for single-project tasks — load the specialist directly. A request is **single-project** (one specialist owns it end-to-end) even when it bundles several things *inside one project*: a Flow with script nodes plus an inline HITL approval step plus its own solution wrapper is **one** `uipath-maestro-flow` task. Inline nodes (HITL QuickForm, script, connector, inline agent) and the solution scaffolding are author sub-steps the specialist performs itself — they are **not** separate skills to orchestrate. Counting them as distinct skills ("solution + flow + human-in-the-loop") and emitting a plan is the most common mis-trigger. The planner is only for requests spanning **separate buildable projects** — distinct `.uipx` projects (a Flow consuming standalone RPA processes, an Agent using published processes as tools).
 
 ## Critical Rules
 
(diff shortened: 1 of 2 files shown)


tmatup:
### Review Response

Thanks @claude (github-actions bot). One fix applied; the two Low density findings I'm keeping with reasoning.

**4. Duplication within `multi-skill-patterns-guide.md`** (Medium) — `multi-skill-patterns-guide.md:39`

**Fixed** in d91b5f83. The blockquote at line 23 stays the authoritative "inline nodes are not components" statement; Pattern 2's "When it applies" now cross-references it (`Inline flow nodes are not components (see the blockquote above)`) and keeps only the Pattern-2-specific fact — that `uipath-maestro-flow` authors inline nodes itself in steps 1/4. Removes the restatement, preserves the sequencing detail. Safe to compress here because this reference doc is read post-invocation, not at routing time.

**2. "When to Use" skip paragraph density** (Low) — `SKILL.md:26`

**Keeping current wording.** The repetition is deliberate at this decision boundary, and one clause in particular earns its place: the explicit `"(solution + flow + human-in-the-loop)"` callout mirrors the *exact* mis-reasoning observed in the regression (the model literally reasoned "spans multiple skills (solution, flow, human-in-the-loop)" before routing to the planner). Naming the observed failure mode verbatim is higher-value than terseness here. Happy to take your compressed version if @RaduAna-Maria prefers the tighter form — it's a reasonable density call for the CODEOWNER to make.

**3. Rule 5 HITL guidance density** (Low) — `SKILL.md:34`

**Keeping current wording.** The final clause ("a field-level prescription forces one shape and defeats that choice") states the *why*, which is the part that survives paraphrase when a model compresses the rule under context pressure. Robustness over brevity for a Critical Rule.

**1 & 5** (OK) — no action.

---

Manual-review items, all acknowledged and out of scope for a doc change:
1. **Routing effectiveness** — empirical; can't be made deterministic by prompt text. Covered by test-plan items 3–4 (eval re-run + multi-project spot-check). PR #1031 remains the deterministic guarantee.
2. **Negative-scope breadth** — the scope is deliberately narrowed to "even a Flow/Agent/RPA project with inline … nodes wrapped in its own solution," so a request that names *separate* projects alongside inline nodes still trips the multi-skill criteria. Flagged for @RaduAna-Maria's domain read in the PR body.
3. **Rule 5 example longevity** — `approved: boolean` is illustrative, not a contract; if HITL schema patterns evolve the example can follow. Minor.

rockymadden (approved):

tmatup:
### CI Triage — resolved

**Failing check:** `uipath-planner` / **CLI Verb Gate** (run [26475594729](https://github.com/UiPath/skills/actions/runs/26475594729)).

**Root cause:** The gate (`scripts/check-skill-verbs.py`) scans the whole skill directory whenever a PR touches it, and flagged one blocking **stale verb**:

```
skills/uipath-planner: 1 blocking, 0 soft-stale, 2 Uncertain
##[error]Stale verb `uip flow pack`
```

`uip flow pack` (Pattern 4 in `multi-skill-patterns-guide.md`) is **pre-existing** — introduced in #538 (2026-05-15), not by this PR. It surfaced now because this PR is the first change to touch `skills/uipath-planner/` since the gate started enforcing it. The current verb is `uip maestro flow pack` (per `assets/uip-catalog-snapshot.json` and the `uipath-maestro-flow` skill, which uses `uip maestro flow pack` throughout).

**Fix:** `e04c066c` — `uip flow pack` → `uip maestro flow pack`. Reproduced the gate locally before and after: 1 blocking → **0 blocking** (the 2 remaining `uip rpa analyze` / `uip rpa get-errors` are non-blocking `Uncertain` warnings, also pre-existing — the CI step only fails on blocking).

**Status:** all checks green at `e04c066c`, including the previously-failing **CLI Verb Gate** and the **Skill Activation Gate** (confirming the `when_to_use` change didn't break activation). `mergeable=MERGEABLE`; still pending `@RaduAna-Maria`'s CODEOWNER review.

rockymadden (approved):

PR #1059, opened 2026-05-27
docs(uipath-maestro-flow): require resources.md read + pagination loop before not-found [ENGCE-58198]

chandusailella:
## Summary

- Extends Author CAPABILITY.md Critical Rule #3 (`uipath-maestro-flow`) from a single "read `connections.md`" clause to two non-overlapping clauses — one for `uip is connections ...`, one for `uip is resources run ...`.
- The resources clause forces agents to load the `Data.Pagination.HasMore` / `NextPageToken` / `nextPage=<token>` contract from `/uipath:uipath-platform — resources.md` **before** the first reference-field lookup.
- Explicitly forbids abandoning the pagination loop before `HasMore="false"` (no falling back to alternative endpoints, unsupported `where=` filters, or AskUserQuestion mid-loop).

## Why

Observed failure mode in a real session: agent was building a flow with a Slack `send-message-to-user` node and needed to resolve a Slack user ID for `chandu.lella@uipath.com`. First `uip is resources run list uipath-salesforce-slack curated_users ...` returned 500 rows without the target. Agent then:

1. Tried a `where=` server-side filter (the connector didn't support it for this resource — returned 0 rows).
2. Tried a different endpoint (`users` instead of `curated_users`).
3. Escalated to AskUserQuestion asking the user to paste their Slack user ID manually.

What it **should** have done: read `Data.Pagination.HasMore`, taken the `NextPageToken`, looped via `--query "nextPage=<token>"`. The target was on page 3 of curated_users.

Root cause: the pagination contract lives in `uipath-platform/references/integration-service/resources.md` (rules #2 and #4 — "Complete the pagination loop ... Do NOT abandon mid-loop"), but the `uipath-maestro-flow` skill only enforced reading `connections.md`, not `resources.md`. Reference-field lookup is a 1-line side quest mid-build and the agent didn't drill into the connector plugin's impl.md deeply enough to find the pointer.

## What changed

Single edit to `skills/uipath-maestro-flow/references/author/CAPABILITY.md`, Critical Rule #3:

Before — one bullet covering connections.md.

After — header line + two sub-bullets:
- connections.md before any `uip is connections ...` call (unchanged content)
- resources.md before any `uip is resources run ...` call (new), with explicit anti-pattern: "do NOT abandon mid-loop to try alternative endpoints, `where=` filters the connector may not support, or AskUserQuestion"

Mirrors the existing connections.md precedent — same enforcement level, same single-source-of-truth pattern, no duplication of the pagination loop itself into the maestro-flow skill.

## Why not also touch the plugin impl.md or anti-patterns list

Considered. Both are redundant if the elevated Critical Rule fires — `resources.md` already contains the pagination loop and the "don't abandon" guidance. Adding it locally would duplicate the contract; the goal is to make the agent load the canonical doc, not to inline a copy.

## Test plan

- [ ] Open a fresh Claude Code session with no prior context
- [ ] Issue the same prompt that exposed the bug: "Build a Flow with an agent that checks the latest AI news of the past 7 days and sends me a short summary on slack. Ensure there is a HITL step after the AI summary that goes to my email for approval before sending to me on Slack."
- [ ] Observe whether the agent reads `resources.md` before calling `uip is resources run list uipath-salesforce-slack curated_users ...`
- [ ] Observe whether the agent paginates via `nextPage=<token>` until `HasMore="false"` or the target record is found, **without** falling back to AskUserQuestion
- [ ] Confirm no regression to existing connector / managed-HTTP flows (the connections.md sub-rule is unchanged)


chandusailella:
https://uipath.atlassian.net/browse/ENGCE-58198?focusedCommentId=4023617

baishalighosh:
Follow up with an eval for a lookup on 3rd/4th page of data

baishalighosh (approved):

PR #1063, opened 2026-05-27
test(uipath-agents): grade deploy_my_workspace on deploy command, not cloud success

radugheo:
## What

Make `deploy_my_workspace` grade that the agent **ran the deploy command and built a package**, instead of requiring a successful cloud deploy + invoke.

## Why

The task runs against a real, persistent Orchestrator tenant (`smoke.yaml` passes through real auth and mounts `~/.uipath`). Re-runs accumulate published package versions in the shared personal-workspace feed, so deploy returns `409 version already exists`. The agent followed the skill's documented remedy — bump patch, redeploy — but every bump also collided, burning the turn budget before it reached `invoke`: `MAX_TURNS_EXHAUSTED`.

The invoke step was the only criterion that needed a live deployment, which is exactly the brittle, environment-dependent part.

## Changes

- **Removed** the `uip codedagent invoke main` criterion (required a successful cloud deploy).
- **Prompt:** deploy once; treat "version already exists" as expected and stop — no redeploy loop.
- **Prompt wording:** "Simple Function" → "no-LLM" so it can't drift onto the distinct standalone `uip functions` artifact lifecycle.
- **`max_turns`:** reverted to 40 — grading is satisfied at the first deploy attempt.
- **Description / checker docstring:** updated to match the narrowed scope.

## Grading now

- `command_executed`: `uip codedagent deploy --my-workspace` ran.
- `run_command`: local `.nupkg` produced + `pyproject.toml` well-formed (pack is local — no tenant-success dependency).

### DIFF
--- a/tests/tasks/uipath-agents/coded/deploy_my_workspace/deploy_my_workspace.yaml
+++ b/tests/tasks/uipath-agents/coded/deploy_my_workspace/deploy_my_workspace.yaml
@@ -1,11 +1,12 @@
 task_id: skill-agent-coded-deploy-my-workspace
 description: >
   Coded-agent deploy lifecycle. Verifies the skill guides the agent
-  through `uip codedagent pack` → `uip codedagent publish
-  --my-workspace` (or the combined `uip codedagent deploy
-  --my-workspace`), then `uip codedagent invoke` to start a cloud
-  job and surface the monitoring URL. Exercises the entire production
-  packaging path the existing test suite never reaches.
+  to package and run `uip codedagent deploy --my-workspace` (the
+  combined pack + publish path) and produce a deployable `.nupkg`.
+  Runs against a real tenant: grades that the deploy command ran and
+  the package was built — a "version already exists" publish conflict
+  does not fail the task. Exercises the production packaging path the
+  rest of the suite never reaches.
 tags: [uipath-agents, e2e, coded, lifecycle:deploy, feature:deploy]
 
 agent:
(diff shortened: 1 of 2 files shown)


tests/tasks/uipath-agents/coded/deploy_my_workspace/deploy_my_workspace.yaml
```diff
@@ -19,14 +20,15 @@ run_limits:
...
 initial_prompt: |
-  Build a minimal Simple Function UiPath coded agent named
-  `deploy-smoke` that echoes the input message. Input: `message`
-  (string). Output: `echoed` (string). Take it through scaffold →
-  init → run, then deploy to the personal workspace and invoke the
-  published version with `{"message": "deployed"}`. The invoke
-  should surface a monitoring URL.
+  Build a minimal no-LLM UiPath coded agent named `deploy-smoke`
+  that echoes the input message. Input: `message` (string). Output:
+  `echoed` (string). Take it through scaffold → init → run, then
```
EusebiuJecan07: We might be too prescriptive with the "init" part. That should be implied and part of skills not a human indication. "init" is a means to an end.

tests/tasks/uipath-agents/coded/deploy_my_workspace/deploy_my_workspace.yaml
```diff
@@ -19,14 +20,15 @@ run_limits:
...
+  deploy it to the personal workspace.
 
-  The test harness has UiPath auth pre-configured.
+  The test harness has UiPath auth pre-configured. The workspace may
+  already hold this package from an earlier run — if the deploy
```
EusebiuJecan07, line 28: No change needed but another proof for the need of a mocking mechanism

tests/tasks/uipath-agents/coded/deploy_my_workspace/deploy_my_workspace.yaml
radugheo: Removed those prescriptive instructions

EusebiuJecan07 (approved):

PR #1084, opened 2026-05-27
ci: tolerate Anthropic API usage-limit (HTTP 400) in Claude PR Review

Mihaiii:
## Why

The **Claude Code Review** required check has been failing — and blocking merges — with:

```
API Error: 400 You have reached your specified workspace API usage limits.
You will regain access on 2026-06-01 at 00:00 UTC.
```

When Anthropic returns this `400`, the `claude-code-action` step exits non-zero, which fails the whole job and red-blocks the PR. This is unrelated to the quality of any PR's changes — it's a transient workspace quota, not a review finding.

## What

Make the usage-limit `400` non-blocking **without** disabling the review or silencing genuine failures.

- Run the `Claude Code Action` step with `continue-on-error: true` (and an `id`) so its non-zero exit no longer fails the job by itself.
- Add a guard step that runs only when the review step failed and inspects the execution log (`$RUNNER_TEMP/claude-execution-output.json`):
  - **Usage-limit `400`** → emit a warning annotation and `exit 0` → check passes, merge unblocked.
  - **Any other failure** → emit an error annotation and `exit 1` → check stays red, merge blocked.

## Why not just skip the check

A skip is evaluated *before* the step runs, but the usage-limit is only known *at runtime* after the API call fails — there's no up-front condition to gate on. Unconditionally skipping (`if: false`) would disable the review for every PR and require a manual re-enable later.

## Behavior notes

- **Genuine review failures still block.** Only the specific usage-limit `400` is tolerated; everything else re-fails via the guard.
- **Self-healing.** Once the workspace quota resets (2026-06-01), the action succeeds normally and the guard never fires — no follow-up change needed.
- **Safe default.** If the action crashes before writing its log, the grep finds nothing and the guard takes the `exit 1` branch — it blocks rather than passing blindly.

### DIFF
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -41,6 +41,8 @@ jobs:
           fetch-depth: 0
 
       - name: Claude Code Action
+        id: claude_review
+        continue-on-error: true
         uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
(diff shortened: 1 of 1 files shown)


Mihaiii:
Here is an example: https://github.com/UiPath/skills/actions/runs/26516658362/job/78095655320?pr=1071

Mihaiii:
Closing. We no longer get `Error: Action failed with error: SDK execution error: Error: Claude Code returned an error result: API Error: 400 You have reached your specified workspace API usage limits. You will regain access on 2026-06-01 at 00:00 UTC.`    ¯\_(ツ)_/¯

PR #1085, opened 2026-05-27
docs(uipath-maestro-flow): document registry search output shape

jiyangzh:
## What

Document the `uip maestro flow registry search --output json` output shape in the maestro-flow skill.

- [cli-commands.md](skills/uipath-maestro-flow/references/shared/cli-commands.md): `Data` is a **flat array** (not `Data.Nodes`), fields are **PascalCase** (`NodeType`, `Category`, `DisplayName`, …). `NodeType` is the string passed to `node add`. 
- [connector/impl.md](skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md): one-line pointer from the connector discovery step to the ref doc.

## Why this is needed — wasted agent turns on `search`

The `search` output shape was undocumented, so the agent has to reverse-engineer it at runtime. In a live e2e run (`salesforce_opportunity_list`), the connector step ate **~6 Bash turns just probing the JSON shape** before any real work:

1. Assumed `Data.Nodes` (a dict) with lowercase `type`/`category` → `AttributeError: 'list' object has no attribute 'get'`
2. Retried with defensive `isinstance` branching → still 0 results
3. Dumped `type(data)`, then `type(Data)`, then first-item keys, then a full item — purely to learn the schema
4. Finally landed on the flat-array + PascalCase `NodeType` shape and proceeded

That is pure discovery overhead the skill should eliminate. Getting the connection itself was a single `is connections list --all-folders` call; the schema-guessing was the real cost. Documenting the shape (with one verified example) removes those turns.

## Verification

Example fields confirmed against the live CLI (`uip maestro flow registry search salesforce --output json` → 188 items, `Data` is a list, fields PascalCase).

### DIFF
--- a/skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
+++ b/skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
@@ -41,7 +41,7 @@ To classify a node, read `Node.form.sections[0].fields[0].componentProps.connect
 
 If you cannot run `node configure` (no live connection, sandbox forbids it, the user asked for planning only):
 
-1. Confirm the connector operation exists: `uip maestro flow registry search <keyword>` and `registry get <nodeType>`.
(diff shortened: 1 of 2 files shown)


tmatup (approved):
interesting finding and fix!

rockymadden (approved):

PR #1088, opened 2026-05-28
fix(uipath-maestro-flow): eliminate hardcoded versions

rockymadden:
## Summary

Stop hardcoding `.flow` versions in the Maestro Flow skill docs. Three hardcoded version literals were **wrong against CLI 1.2.0** and caused copy-paste `no matching definition` failures; the rest were brittle literals that contradicted the skill's own rule ("copy whatever `registry get` / `flow init` emits"). This PR converges every authoring source on the placeholder + live-command convention the skill already used elsewhere, and adds an executable checker so the regression can't recur.

Docs-only change. No `.flow` schema or CLI behavior changes.

## The three live mismatches (verified against `uip` CLI 1.2.0)

| Artifact | Doc said | Registry/CLI emits | Impact |
|----------|----------|--------------------|--------|
| top-level `.flow` `version` | `1.1` (and banned other values) | **`1.2`** (`flow init`) | doc argued against the value `init` now scaffolds |
| `core.action.http.v2` `typeVersion` | `2.0` | **`2.1`** | literal copy → `Node type core.action.http.v2:2.0 has no matching definition` |
| `core.trigger.scheduled` `typeVersion` | `1.0` | **`1.1`** | flow built from the doc fails validation today |

The original report named the first two; the `core.trigger.scheduled` mismatch was found by querying the registry for **every** node type the docs hardcode.

## What changed

**Tier 1 — the live mismatches (`shared/file-format.md`, `http/impl.md`, `scheduled-trigger/impl.md`)**
- Removed the stale `currently "1.1"` claim, the `z.literal("1.1")` / `workflowSchemaV1_1` schema assertion, and the value-ban list.
- Top-level `version` JSON examples → `"<version from \`uip maestro flow init\`>"` placeholder.
- `http` / `scheduled-trigger` skeletons → `"<DEFINITION_VERSION>"` + a "set from the `registry get` response" note.

**Tier 2 — prose stating versions as current (`summarize`, `batch-transform`, `data-fabric`)**
- Reworded "the registry currently emits `1.0`" → "set it to the `version` from the `registry get <type>` response; do not guess." Kept the `x.y`-not-`x.y.z` format caution.
- Clarified the Data Fabric `<activityVersion>` as a connector-internal value distinct from node `typeVersion`.

**Tier 3 — brittle authoring skeletons (11 plugin impls)**
- Converted each plugin's **primary** node skeleton `typeVersion: "1.0"` → `"<DEFINITION_VERSION>"` + sourcing note: `terminate`, `decision`, `switch`, `merge`, `end`, `delay`, `queue`, `transform`, `hitl`, `subflow`, `loop`.
- Left intentionally: skeletons already governed by a sourcing note (`inline-agent`, and `summarize`/`batch-transform` primaries); downstream end-node examples; interior nodes of worked-example flows; connector-scoped Data Fabric literals (verify on a DF tenant).

**Guardrail — `.maintenance/check-versions.sh` (new)**
- Flags **PROSE-PIN** (prose pinning a version as current/required) and **TOPLEVEL** (hardcoded `.flow` top-level `version` JSON literal).
- Does **not** flag node `typeVersion` JSON literals (legitimate when paired with a sourcing note), `bindings_v2.json` `2.0` (fixed constant), `definitions[].version`, or eval-file `version`.
- `<!-- version-check-skip -->` escape hatch for intentional references (anti-examples, exempt schemas).
- Registered in `check-all.sh` (now 9 checkers); documented in `.maintenance/README.md`.

## Verification

- **Doc-derived flow validates:** built a flow following the revised docs (scheduled trigger + http.v2 + decision + end + terminate), sourcing every version live → resolved to scheduled `1.1`, http.v2 `2.1`, others `1.0`, top-level `1.2` → `uip maestro flow validate` returns `Valid`.
- **Checker proven both ways:** `violations=0` on the fixed skill; catches all 4 original defects against a regression fixture reproducing the pre-fix text.
- **No new maintenance-suite failures:** links 0 broken, orphans 0, plugin-pairs/template clean. Pre-existing `link-text broken=5`, `anchors_bad=22`, `uip unknown=2` are unchanged and in files this PR did not touch.

## Why placeholders, not re-pinning

Swapping `1.1`→`1.2` / `2.0`→`2.1` would just create the next stale literal. Pointing at `flow init` / `registry get` — values the agent already fetches when copying definitions into `definitions[]` — costs **zero extra commands** and can't go stale.


tmatup (approved):

tmatup:
> batch-transform and summarize primary JSON skeletons not updated to <DEFINITION_VERSION> — inconsistent with all other plugin impl files. These are the only two holdouts.

PR #1093, opened 2026-05-28
docs(uipath-maestro-flow): clarify registry get node shape

tmatup:
## Summary

Clarifies the `registry get --output json` response shape in the shared action-node guide:

- the manifest lives at `Data.Node`
- `handleConfiguration` is an array of position groups, not a map
- agents should read `inputDefinition`, `outputDefinition`, and `model.serviceType` from `Data.Node`

## Why

The run logs showed many `registry get` inspect/error loops where agents treated `Data` as the manifest or misread nested fields. This makes the high-frequency lookup contract explicit at the guide agents use before authoring action nodes.

## Validation

- `git diff --check`
- `python3 scripts/check-skill-verbs.py --json skills/uipath-maestro-flow` reports `blocking: 0`

---
🤖 Generated with Claude Code

### DIFF
--- a/skills/uipath-maestro-flow/references/shared/action-nodes.md
+++ b/skills/uipath-maestro-flow/references/shared/action-nodes.md
@@ -29,7 +29,14 @@ uip maestro flow registry get <node-type> --output json
 }
 ```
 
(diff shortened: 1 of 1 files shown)


tmatup:
### Review Response

**1. Line 32 dense single sentence — split into bullet list** — `action-nodes.md:32`

**Fixed** in 82691f89. Split the four `Data.Node.*` field-paths into individual bullets so each lookup target is independently addressable. Matches the repo's token-optimization preference for parseable enumeration over prose.

**2. Verify `Data.Node` + `handleConfiguration` array shape against actual CLI output** — `action-nodes.md:13-30`

**Keeping current implementation** — the JSON shape was extracted directly from `uip maestro flow registry get` run logs cited in the PR body, and matches the array-of-position-groups form documented for the search-side counterpart in #1085.

---

nikhil-maryala (approved):

rockymadden (approved):

PR #1095, opened 2026-05-28
docs(uipath-maestro-flow): document registry availability semantics

tmatup:
## Summary

Documents how to interpret `AvailableOnTenant` in the shared flow CLI command reference:

- `true` rows can continue to `registry get` / `node add`
- `false` rows are SDK-known but unavailable in the current tenant registry
- agents should not try unsupported flags such as `--include-unavailable`

## Why

The run logs showed agents seeing `AvailableOnTenant: false` and then spending calls on `registry get` or nonexistent flags. This makes the branch explicit so agents choose an enabled alternative, use `--local` where appropriate, or report the resource as unavailable.

## Validation

- `git diff --check`
- `python3 scripts/check-skill-verbs.py --json skills/uipath-maestro-flow` reports `blocking: 0`

---
🤖 Generated with Claude Code

### DIFF
--- a/skills/uipath-maestro-flow/references/shared/cli-commands.md
+++ b/skills/uipath-maestro-flow/references/shared/cli-commands.md
@@ -325,6 +325,11 @@ uip maestro flow registry get <nodeType> --output json     # get full schema for
 ] }
 ```
 
+`AvailableOnTenant` is a usability gate, not extra metadata to ignore:
+
+- `true` — the tenant registry can return this node; it is valid to continue with `registry get <NodeType>` or `node add <NodeType>`.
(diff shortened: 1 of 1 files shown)


tmatup:
### Review Response

**1. Sanity-check `--include-unavailable` doesn't exist in the CLI catalog** — `cli-commands.md`

**Keeping current implementation** — `grep -r "include-unavailable" assets/ skills/` returns nothing, confirming the flag is undocumented across both the catalog snapshot (`assets/uip-catalog-snapshot.json`, uip 1.2.0, 1115 verbs) and every skill. The PR's explicit warning is precisely intended to head off agents hallucinating this flag, which is the failure mode observed in run logs.

---

nikhil-maryala (approved):

rockymadden (approved):

PR #1098, opened 2026-05-28
chore(tests): prune task YAMLs to inherit experiment defaults

akshaylive:
## Summary
- Adds `scripts/prune-task-defaults.py` — a smart pruner that compares each task's `agent:`, `run_limits:`, and `sandbox:` blocks against `tests/experiments/default.yaml` and removes redundant fields.
- Hoists deprecated `agent.max_turns` / `agent.turn_timeout` into top-level `run_limits:` (the 2026-05-20 coder_eval rename — see `coder_eval/orchestration/experiment.py:163`).
- Applied to all 429 task YAMLs; **317 updated**, net **−1181 lines** of inherited config.

## Why
Most tasks were carrying boilerplate that matches the experiment defaults verbatim — `type: claude-code`, `permission_mode: acceptEdits`, `sandbox.driver: tempdir`, empty `python: {}` / `node: {}`. Pruning makes task files focused on what actually overrides the default.

## Script behavior
1. Hoists `agent.{max_turns,turn_timeout,task_timeout}` → `run_limits.*` (conflict-safe: prefers existing `run_limits` value).
2. Drops any `agent` / `run_limits` / `sandbox` field whose value equals the default in `tests/experiments/default.yaml`.
3. Drops empty `{}` / `[]` values (no-op configs).
4. Removes top-level blocks that become empty.
5. Preserves formatting via `ruamel.yaml` (2-space sequence indent) and collapses orphaned blank lines.

Idempotent — re-running on already-pruned files is a no-op.

## Test plan
- [ ] Spot-check diffs across `uipath-admin/`, `uipath-agents/`, `uipath-maestro-flow/`, `uipath-rpa/`
- [ ] Smoke run (`tests/experiments/smoke.yaml`) — confirm overrides still take effect (smoke's `max_turns: 40` vs default's `200`)
- [ ] Nightly run — confirm docker sandbox + extra env passthrough still apply (none of those are in task files)

### DIFF
--- a/.claude/commands/lint-task.md
+++ b/.claude/commands/lint-task.md
@@ -180,6 +180,27 @@ The catalog is regenerated nightly by `.github/workflows/refresh-uip-catalog.yml
 
 Severity: **High** — always. No carve-outs; there is no valid reason to re-install the CLI the runner already provides.
 
+### Run-limit fields under agent:
+
+**Raise a Medium issue if** the task's `agent:` block contains any of: `max_turns`, `turn_timeout`, `task_timeout`.
+
+**Why it's wrong:** These fields were moved out of `agent:` into a top-level `run_limits:` block in the c/2026-05-12 unify-run-limits migration (see `coder_eval/src/coder_eval/orchestration/experiment.py:163`). The loader still hoists them with a `DeprecationWarning`, but the legacy location was scheduled for removal on 2026-05-20. Leaving them under `agent:` emits a warning every run and drifts from every other task in the repo.
+
+**Fix:** Move the field to a top-level `run_limits:` block. If `agent:` now has no remaining fields, drop the block entirely (defaults are inherited from the experiment). The `scripts/prune-task-defaults.py` script does this transformation automatically.
+
+```yaml
+# Before
+agent:
+  type: claude-code
+  max_turns: 40
+
+# After
+run_limits:
+  max_turns: 40
+```
+
+Severity: **Medium** — task still runs (loader hoists for now), but the deprecation window has elapsed and the warning pollutes every run.
+
 ## Phase 4 — Compose Per-Task Report
 
 For each task, print:
--- a/scripts/prune-task-defaults.py
+++ b/scripts/prune-task-defaults.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+"""Prune redundant agent/run_limits fields from task YAMLs.
+
+For each task under tests/tasks/, compare every field in `agent:` and
+`run_limits:` against tests/experiments/default.yaml and remove any that match
+the inherited default. Also hoists deprecated `agent.max_turns` / `agent.turn_timeout`
+into top-level `run_limits` before pruning (see coder_eval orchestration/experiment.py:163).
+
+Usage:
+    scripts/prune-task-defaults.py              # apply in place
+    scripts/prune-task-defaults.py --dry-run    # show what would change
+    scripts/prune-task-defaults.py path/to/file.yaml [more.yaml ...]
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+
+from ruamel.yaml import YAML
+
+REPO = Path(__file__).resolve().parent.parent
+DEFAULTS_PATH = REPO / "tests" / "experiments" / "default.yaml"
+TASKS_ROOT = REPO / "tests" / "tasks"
+
+# Fields under `agent:` that coder_eval auto-hoists into top-level `run_limits`
+# (deprecated; warns to be removed 2026-05-20).
+HOISTABLE = ("max_turns", "turn_timeout", "task_timeout")
+
+
+def load_yaml(path: Path):
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    yaml.width = 10_000  # don't reflow long lines
+    # Match the repo's 2-space indented list style: "  - key:" not "- key:".
+    yaml.indent(mapping=2, sequence=4, offset=2)
+    return yaml, yaml.load(path.read_text())
+
+
+def prune_block(block, defaults: dict) -> list[str]:
+    """Remove keys whose value equals the default, or whose value is an empty dict/list.
+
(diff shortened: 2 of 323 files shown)


uipreliga (approved):
Fix what you agree with, ask @bai-uipath  for second opinion and 🚢

uipreliga:
## Code Review: `scripts/prune-task-defaults.py`

Reviewed in parallel by Opus 4.7 and Gemini 3 Pro (Codex was rate-limited). Findings consolidated; overlapping issues noted.

### 🔴 High

- **`task_timeout` hoisting is broken** (`scripts/prune-task-defaults.py:29,71-95`) — flagged by both reviewers. `HOISTABLE` includes `task_timeout`, but in every task YAML in this repo `task_timeout:` lives at the **top level**, not under `agent:`. The hoist loop only inspects `agent.*`, so 14 files retain a top-level `task_timeout:`, and `tests/tasks/uipath-maestro-flow/ixp/e2e_02_project_selection.yaml:30` has `task_timeout: 1200` — exactly the inherited default, so it should have been pruned. Reproduced with `--dry-run` on that file: reports 0 changes.

  **Fix:** either drop `task_timeout` from `HOISTABLE` (and document that top-level deprecated keys are out of scope), or add an explicit top-level hoist:
  ```python
  if "task_timeout" in data:
      # hoist to run_limits, resolve conflicts the same way as agent.* hoists
  ```

### 🟡 Medium

- **`TypeError` when `run_limits` exists but isn't a dict** (`:69-94`) — flagged by both reviewers. `run_limits = data.get("run_limits")` is captured once; if it's `None` or a list, the `isinstance(run_limits, dict)` check at line 83 falls through to `else: run_limits[field] = agent.pop(field)` at line 94, which crashes. No YAML in the repo currently triggers this — the lazy-create path reads as defensive when it isn't. Restructure so the `CommentedMap()` create fires on `not isinstance(run_limits, dict)`, not just `is None`.

### 🟢 Low

- **Missing `encoding="utf-8"`** (`:38, :141`) — `Path.read_text()` / `write_text()` use the platform-default encoding. Some YAMLs contain em-dashes and smart quotes; on Windows with cp1252 this could mangle them. Both reviewers flagged.
- **Stale deprecation date** (`:28`) — comment says "to be removed 2026-05-20"; today is 2026-05-27. Either the deprecation has already lapsed (script may be unnecessary) or the date wants updating.
- **`\n{3,}` collapse operates on serialized YAML** (`:132`) — including the contents of `initial_prompt: |` literal blocks. Could merge intentional blank-line paragraph breaks. No current YAML appears affected.
- **Lazy imports inside `process()`** (`:77, :124-125`) — `CommentedMap`, `io`, `re` could move to the top of the file. Style nit.

### False positive (downgraded)

- ~~Regex not anchored to line start~~ (`:137`) — Gemini flagged `(?<=\n)(?<!\n\n)(?=(?:agent|sandbox|run_limits):)` as potentially matching inside literal blocks. It is in fact line-anchored: the lookbehind requires `agent:` / `sandbox:` / `run_limits:` to be preceded by exactly one newline. YAML literal-block content is indented (e.g. `  agent:`), so `agent:` at column 0 cannot appear there without terminating the block. `agent_session:` is also safe — the lookahead requires a literal `:` immediately after `agent`. Leaving as-is.

### Positive observations

- ruamel config (`:33-37`) correctly matches the repo's YAML indent style, with a comment explaining *why*.
- Idempotent: re-running on already-pruned YAMLs is a no-op (`if not changes` short-circuit at `:115`).
- Defaults loaded once and threaded through — avoids re-parsing per file.
- `--dry-run` + per-file change reporting are good UX for a maintenance script.
- Hoist conflict resolution is explicit and documented (`:88`).
- `CommentedSeq` / `CommentedMap` correctly implement `__eq__` against same-typed defaults, so the `val == defaults[key]` check at `:53` does the right thing for `allowed_tools` (verified — overrides that differ from the default were preserved).

### Overall

Solid pragmatic one-shot script with the right shape (load → hoist → prune → reformat → write) and thoughtful attention to ruamel quirks. The `task_timeout` gap is the meaningful finding — it's real, observable in the tree, and the constant claims coverage the code doesn't deliver. The `run_limits` non-dict crash is theoretical but cheap to harden. Everything else is polish.

---

bai-uipath (approved):
Solid cleanup; net −1,181 lines of inherited boilerplate is real value. Two issues before merge.

## Major issues

### 1. Drop the script, add a lint instead

`scripts/prune-task-defaults.py` has no callers — no CI, no pre-commit, no `/lint-task` entry. Meanwhile `tests/templates/test-task-template.yaml` still emits the boilerplate the pruner just removed (`agent.type`, `permission_mode`, `allowed_tools`, `sandbox: { driver: tempdir, python: {} }`), so `/generate-task` will re-dirty the corpus within a few PRs.

Please:

1. **Delete the script.** It's a one-shot; the cleanup commit stands without it.
2. **Fix the template** — drop the `agent:` block and use `sandbox: {}` (or omit `sandbox:` entirely).
3. **Add a lint** that fails when any task YAML re-introduces a field that matches the experiment default. Specifically detect:
   - `agent.type: claude-code`
   - `agent.permission_mode: acceptEdits`
   - the canonical `allowed_tools` list
   - `sandbox.driver: tempdir`
   - empty `python: {}` / `node: {}`
   - deprecated `agent.max_turns` / `agent.turn_timeout` / `agent.task_timeout` (renamed to `run_limits.*` on 2026-05-20)

   Error message should name the field and point at `tests/experiments/default.yaml` so the author sees it's an inherited default, not missing config. Wire it into the GH smoke workflow next to `scripts/check-cli-verbs.py` and `/lint-task`.

Without an active lint these patterns will come back. Schema changes like the `agent.*` → `run_limits.*` rename are invisible to anyone who learned the old shape — authors copy from siblings, paste from old PRs, or use muscle memory, and the deprecated fields creep in unnoticed.

### 2. One inline comment lost in `uipath-rpa/legacy/create_and_validate.yaml`

The deleted `agent:` block carried a 9-line comment documenting why the historical Opus pin was dropped (EU Bedrock inference-profile rejection, content-filter retry behavior). That context is now only in git history. Relocate the comment above `description:` in this file.

PR #1099, opened 2026-05-28
chore: refresh uip CLI catalog (1.1.0-alpha.20260521.7245)

uipreliga:
Automated refresh against `@uipath/cli@1.1.0-alpha.20260521.7245`. The catalog has 1116 reachable verbs after installing all available plugins.

Merges automatically once all checks pass (handled by `automerge-catalog-refresh.yml`). If a check fails, the failure is real drift (a CLI release removed/renamed a verb still referenced in skill docs or task YAMLs); investigate the gate annotations and either sweep the docs or, if the verb is intentionally retired, add it to `.claude/rules/cli-renames.md`. `/lint-task` and the CLI Verb Gate will surface the affected references.


Mihaiii:
For the failing checks for uipath-rpa & uipath-solution there is already a PR opened here: https://github.com/UiPath/skills/pull/1071

PR #1113, opened 2026-05-28
docs(skills): reflect login status tenant ids

dsuresh-ap:
## Summary
- Prefer `Data.TenantId` from `uip login status --output json` for tenant-bound admin authz role guidance, with an older-CLI tenant lookup fallback.
- Update the tenant authz lifecycle eval wording to match the new login-status ID shape and quote its colon tag.
- Tighten feedback sanitization so `OrganizationId` and `TenantId` exposed by UiPath/cli#1962 are not included in feedback reports.

## Validation
- `git diff --check`
- `bash hooks/validate-skill-descriptions.sh`
- `ruby -e 'require "yaml"; YAML.load_file("tests/tasks/uipath-admin/authz_role_lifecycle_tenant_e2e.yaml"); puts "ok"'`


dsuresh-ap:
Closing per request; these skill changes are not needed.

PR #1123, opened 2026-05-28
fix(uipath-maestro-flow): prefer output-filter

rockymadden:
## Summary

Promotes `--output-filter` (JMESPath) as the default extraction reflex for `uip … --output json` across the `uipath-maestro-flow` skill, eliminating a 3-call discovery loop that surfaced in a real agent transcript.

In the failing run (see [ISSUE.MD context](#background)), an agent took 3 attempts — calls 15, 16, 17 — to extract Slack connector node types from `registry search`. Calls 15 and 16 piped `--output json` to `python3 -c` with a `Data.Nodes` access path that doesn't exist (`Data` is a flat array); 15 silently returned `[]`, 16 threw on a different access path, 17 abandoned parsing. The CLI's built-in `--output-filter <jmespath>` would have returned the right answer on call 1 — but the recipe was buried in a plugin-depth doc (`connector/planning.md:44`) that the agent loaded and ignored.

## Changes

**Depth-1 (the two files every agent reads):**
- `SKILL.md` — Critical Rule #1 extended in place to prefer `--output-filter`, inline the canonical recipe, and pin the `registry search` Data shape. New anti-pattern bullet names the silent-`[]` failure mode. Capability-router descriptor at line 54 updated to advertise `--output-filter`.
- `references/author/CAPABILITY.md` — inheritance line (line 7) names the new preference; capability-router row (line 92) replaces the bare `registry search` command with the full recipe and a link to `cli-conventions.md §3`.

**Defense in depth (deeper layers):**
- `references/shared/cli-conventions.md` — new §3 "Prefer `--output-filter` for extraction" inserted before the existing JSON-shape section, with two labeled examples (canonical broad discovery + narrow query), a `When to fall back to python3 / jq` subsection, and shape-inspection recipes. Sections §3–§7 renumbered to §4–§8; five external anchor-link cross-references updated to match.

**Sweep of contradicting examples:**
- `references/author/references/editing-operations-json.md` — section header renamed from `jq for extracting CLI JSON` to `--output-filter for extracting CLI JSON`. Two `jq` one-liners rewritten as `--output-filter` form.
- `references/shared/file-format.md` — `python3 -c` shape-inspection pipeline collapsed to one-line `--output-filter`.

**Bugs caught during live verification (Phase 5):**
- `references/shared/file-format.md` — fixed a latent case bug: `Node.supportsErrorHandling` → `Node.SupportsErrorHandling` (PascalCase). The original `python3 -c` form silently returned `None` because `.get()` doesn't raise on missing keys; agents reading this doc would have concluded "this node doesn't support error handling" when in fact it does. The plan caught a bug the plan was designed to prevent.
- `references/shared/cli-conventions.md §3` — initial `keys(@)` shape-inspection advice throws when `Data` is an array (the exact case the failing agent hit). Replaced with a `type(@)` first-probe sequence covering both array and object Data, plus a silent-`[]` callout with field-name-casing reminder.

## Files modified

10 skill files, ~70 lines net delta:

```
skills/uipath-maestro-flow/SKILL.md                                              | +5/-2
skills/uipath-maestro-flow/references/author/CAPABILITY.md                       | +2/-2
skills/uipath-maestro-flow/references/author/references/editing-operations-json.md | +7/-3
skills/uipath-maestro-flow/references/shared/cli-conventions.md                  | +51/-4
skills/uipath-maestro-flow/references/shared/file-format.md                      | +2/-3
skills/uipath-maestro-flow/references/diagnose/CAPABILITY.md                     | +1/-1
skills/uipath-maestro-flow/references/operate/CAPABILITY.md                      | +1/-1
skills/uipath-maestro-flow/references/operate/references/manage.md               | +1/-1
skills/uipath-maestro-flow/references/operate/references/run.md                  | +1/-1
skills/uipath-maestro-flow/references/operate/references/ship.md                 | +1/-1
```

## Commits

- `dbac2632` — phase 1 (depth-1 rule)
- `037f0098` — phase 2 (cli-conventions.md §3 + capability router + anchor-link cross-refs)
- `1fba0757` — phase 3 (sweep contradicting examples)
- `663e31ed` — phase 5 (live-verification bug fixes)

## Test plan

- [x] **Live-tenant CLI verification** — every recipe added in the docs was replayed against the live `uip` CLI:
  - `registry search slack --output-filter "[*].{NodeType:NodeType,…}"` → 66 rows ✅
  - `registry search slack --output-filter "[?starts_with(NodeType,'uipath.connector.uipath-salesforce-slack.')]…"` → 22 Slack connector activities including `get-channel-info` ✅
  - `registry get <NODE_TYPE> --output-filter "Node"` → full manifest ✅
  - `registry get <NODE_TYPE> --output-filter "Node.SupportsErrorHandling"` → `true` (after case fix) ✅
  - `type(@)` / `[0] | keys(@)` shape-inspection recipes return correct results on both array and object Data ✅
- [x] **3→1 reproduction** — the agent's failing discovery from ISSUE.MD reproduced as one CLI call:
  ```bash
  uip maestro flow registry search slack --output json \
    --output-filter "[?contains(NodeType,'get-channel-info')].{NodeType:NodeType,DisplayName:DisplayName,Description:Description}"
  ```
  Returns the right node in one call; the original transcript took 3.
- [x] `hooks/validate-skill-descriptions.sh` exits 0
- [x] `grep -rn "| python3\|| jq " references/` returns zero hits adjacent to `uip … --output json`
- [x] No orphaned anchor-link cross-references (`#4-login-state`, `#5---folder-key-requirement` are gone; new `#5-login-state`, `#6---folder-key-requirement` resolve)
- [ ] **Full `coder-eval` re-run of [tests/tasks/uipath-maestro-flow/multi_node/slack_weather_pipeline/slack_weather_pipeline.yaml](tests/tasks/uipath-maestro-flow/multi_node/slack_weather_pipeline/slack_weather_pipeline.yaml)** — local environment didn't have `coder-eval` installed; substituted with direct CLI verification (above). A CI run will capture the agent's *actual* tool-call count under the new docs as the final empirical confirmation.

## Notes for reviewers

- **Framing is "prefer," not "ban."** External parsers (`python3 -c`, `jq`, `sed`, `awk`) remain valid for transforms JMESPath cannot express — multi-step joins, format conversion, conditional output from multiple fields. The change targets the *default reflex*, not the fallback toolbox.
- **Depth matters.** The recipe was already in `connector/planning.md:44` before this PR; the agent loaded the file and still chose `python3 -c`. Depth-3 documentation cannot beat a habit — only depth-1 (Critical Rules) can. This is why Phase 1 extends Rule #1 in place rather than adding a deeper rule.
- **One latent bug was incidentally fixed** (`Node.SupportsErrorHandling` case) — see commit `663e31ed`.
- **Sibling skills** (`uipath-platform`, `uipath-rpa`, etc.) likely have the same problem. Tracking as separate follow-up; out of scope for this PR.


rockymadden:
@claude re-review

skills/uipath-maestro-flow/references/author/references/editing-operations-json.md
```diff
@@ -71,15 +71,17 @@ uip maestro flow validate <FILE>.flow --output json
...
+Read-only extractions on `--output json` results — use the CLI's built-in JMESPath filter, no external parser needed. Expressions start at the `Data` envelope (no `Data.` prefix). See [shared/cli-conventions.md §3](../../shared/cli-conventions.md#3-prefer---output-filter-for-extraction) for the full pattern.
 
 ```bash
-uip solution upload --output json | jq -r '.Data.Url'
-uip maestro flow registry get <NODE_TYPE> --output json | jq '.Data.Node'
+uip solution upload --output json --output-filter "Url"
+uip maestro flow registry get <NODE_TYPE> --output json --output-filter "Node"
 ```
 
+Reach for `jq` / `python3` only when JMESPath cannot express the operation (multi-step joins, format conversion, conditional output computed from multiple fields).
```
tmatup, line 83: 👍 make sense to fixate on one tool.

rockymadden:
@claude re-review

nikhil-maryala (approved):

tmatup:
### Review Response

Triaged the `github-actions[bot]` Claude review. Two findings were already handled in `b8d773e5`; two more are fixed in `9734440c`; the rest are justified below.

**1. `keys(@)` vs `type(@)` inconsistency** — `SKILL.md:86`

**Already fixed** in `b8d773e5`. The anti-pattern now points at the `type(@)` first-probe sequence and explicitly notes `keys(@)` throws on arrays. Matches `cli-conventions.md §3`.

**2. Rule #1 ↔ §3 duplication**

**Already addressed** in `b8d773e5`. Rule #1 was trimmed to bridge content + a link to `cli-conventions.md §3`, removing the near-verbatim restatement. Per Reference Pointer Discipline this is the right altitude for a Critical Rule (agents may not load references).

**3. "Fails loudly with a non-zero exit" claim** — `SKILL.md` + `cli-conventions.md §3` *(raised under "Issues for Manual Review — JMESPath error behavior")*

**Fixed** in `9734440c`. This was the most important finding — the claim was inaccurate, and I verified it against the live CLI (`uip 1.2.0`):

| `--output-filter` expression | Result | Exit |
|---|---|---|
| `[?NodeType=='does.not.exist']` (valid, no match) | `Data: []` | **0** |
| `Nodes` (the exact `Data.Nodes` wrong-guess) | `Data: []` | **0** |
| `Nodes[*].NodeType` (wrong shape) | `Data: []` | **0** |
| `keys(@)` on array (type error) | `ValidationError` | **3** |
| `[*].{bad` (syntax error) | `ValidationError` | **3** |

So a *valid* expression that doesn't match the shape returns `Data: []` with exit 0 — the **same silent trap** as `python3`/`jq`, not a loud failure. Only *invalid* expressions (bad syntax / type error) exit non-zero. `§3` was also internally contradictory: line 75 claimed loud failure while the silent-`[]` note below it documented the opposite. Reworded both spots so the real protection is the **pinned shape** (write the right expression on call #1), not the flag's error behavior.

**4. Inconsistent "inherits" lines across capabilities**

**Fixed** in `9734440c`. The `--output-filter` preference is universal rule #1, but only `author/CAPABILITY.md` mentioned it. Aligned `operate`, `diagnose`, and `evaluate` inherits lines to `--output json` + prefer --output-filter for extraction`.

**5. Version gating for `--output-filter`** *(Manual review)*

**Keeping as-is.** Confirmed `--output-filter` works as a global flag on `uip 1.2.0`. I can't pin an exact minimum version from here, so I've left §1's version note (which gates the `uip maestro flow` vs `uip flow` *prefix* at ≥0.3.4) untouched rather than assert an unverified number. If anyone knows the CLI release that introduced `--output-filter`, worth adding a one-line gate to §1.

**6. No test task covers `--output-filter`** *(Low)*

**Out of scope, follow-up.** Agrees with the PR's own framing — the skill isn't new and existing tests still exercise the underlying commands. The unchecked `coder-eval` re-run in the test plan will capture the agent's actual tool-call count empirically; a dedicated assertion is better added alongside that run.

---

Verification: `hooks/validate-skill-descriptions.sh` exits 0; `§3` anchor unchanged so the existing cross-references still resolve.

tmatup (approved):

PR #1131, opened 2026-05-28
docs(uipath-maestro-flow): drop solution scaffold from greenfield — CLI auto-creates parent solution

nikhil-maryala:
## Summary

Companion to [UiPath/cli#2276](https://github.com/UiPath/cli/pull/2276), which makes `uip maestro flow init` self-sufficient — it auto-scaffolds `<Name>Solution/<Name>/<Name>.flow` when run outside any solution, and registers with the existing `.uipx` when run inside one. This PR updates the maestro-flow skill to reflect that and drop the manual scaffold steps. https://uipath.atlassian.net/browse/MST-10414

**Files changed:**
- `SKILL.md` rule #6 — rewritten: "just run `flow init` from any directory; the CLI handles the solution layout."
- `references/author/CAPABILITY.md` anti-patterns — drop "Never run `flow init` outside a solution directory"; trust `Data.SolutionRegistration.Status` and `Data.AutoCreatedSolution` instead of an `ls` self-check.
- `references/author/references/greenfield.md` **Step 2** — collapse 2a / 2b / 2c into a single Step 2 that's one `flow init` call. Documents the response envelope and the `<Name>Solution` naming default.
- `references/shared/cli-commands.md` — `uip maestro flow init` section rewritten to the single-call form.
- `references/diagnose/references/failure-modes.md` "Single-nested layout" — demoted to a legacy / manual-scaffolding triage doc (current CLI makes it nearly impossible via `flow init`).
- `references/author/references/editing-operations-json.md` — pre-flight path example updated to reflect the `<Name>Solution/<Name>/` layout.

## Why

Today every greenfield flow run burns ~4 bash calls on the solution dance (cwd `ls`, `uip solution init`, `cd`, `uip maestro flow init`, `ls` verify). With the new CLI behavior, all of that is replaced by a single `uip maestro flow init <Name>` call — but the skill kept driving the old multi-step recipe even after the CLI changed. This PR aligns the skill with the new contract.

## Compatibility note

If the user has an older CLI without the auto-scaffold behavior, this skill change is still correct — the single `flow init` call will simply not auto-scaffold the parent and the user can fall back to the legacy recipe (still documented in `failure-modes.md` → "Single-nested layout"). The CLI PR is gated behind the same `uip maestro flow init` command, so the new and old skills agree on the surface API.

## Test plan

- [ ] Build a flow from scratch with a current CLI: `uip maestro flow init MyFlow` in an empty dir produces `MyFlowSolution/MyFlow/MyFlow.flow` with `SolutionRegistration.Status: "Registered"` — no extra `solution init` or `cd` calls.
- [ ] Build a flow inside an existing solution: `uip maestro flow init InnerFlow` from inside `<existing>.uipx`'s parent dir adds the project to that solution; no `<InnerFlow>Solution/` is created.
- [ ] Re-run on conflict: a second `flow init MyFlow --force` reuses the existing `MyFlowSolution/MyFlowSolution.uipx`.
- [ ] Existing greenfield evals continue to pass (the new single-call recipe is a strict simplification of the old one).


rockymadden (approved):
Thanks for getting this!

PR #1145, opened 2026-05-29
docs(uipath-maestro-flow): batch greenfield build into 3 turns (~2x faster, 5.7x more consistent)

AlvinStanescu:
## What

Restructures the **greenfield (create-a-new-flow) journey** so independent work batches into a single assistant turn instead of one CLI call per turn. The batching guidance is woven *into* the existing steps via `[T1]`/`[T2]`/`[T3]` tags rather than bolted on as a separate section.

### Changes (2 files)

**`SKILL.md`**
- New universal **rule #10**: batch tool calls into one turn whenever data dependencies allow — chain sequential `uip` calls with `&&`, emit independent Bash/Read/Edit as parallel tool uses.
- Matching anti-pattern: never issue setup CLI one-per-turn.

**`references/author/references/greenfield.md`**
- **Three-turn execution map** at the top, with `[T1]`/`[T2]`/`[T3]` tags inline on each Step heading:
  - **T1** — one chained Bash (`solution init` → `flow init` → `registry pull` → `node add`) **+** parallel `registry get` **+** parallel doc Reads
  - **T2** — one Read of the `.flow` **+** a batch of Edits
  - **T3** — one chained Bash (`node configure && validate && format`)
- **Array-boundary anchor pattern** for parallel same-file Edits: anchor each Edit to a unique top-level key boundary (`nodes`/`edges`/`definitions`/`layout`) so parallel Edits never collide. Explicit warning that `"runtime"` is an unsafe anchor (it recurs nested inside the HTTP v2 definition).
- **Correctness fixes** surfaced during live testing (not just speed):
  - `flow init` scaffolds **only** the manual trigger with **zero edges** — the End node is always author-added (prior text wrongly implied a pre-wired trigger→end edge).
  - `node add` signature is `<file> <node-type>` (file first).
  - Note the `solution init` (≥1.0) vs `solution new` (<1.0) verb rename.

## Why / results

Verified with a **canary-checked 5×5 A/B** on `uip 1.2.0-alpha` — same toolchain on both arms, and **each agent reported which skill content it actually read** so the baseline is provably the baseline (an earlier informal comparison had been contaminated by a toolchain mismatch + a stale plugin-cache read path).

| Metric | Baseline (HEAD) | This PR | Δ |
|---|---|---|---|
| Wall-clock mean | 123.6s | **62.6s** | **−49% (1.98×)** |
| Wall-clock median | 123.3s | 64.3s | −48% |
| Wall-clock **std** | 28.4s | **5.0s** | **5.7× more consistent** |
| Range | 87.8–158.6s | 54.5–68.9s | **disjoint** |
| Turns (mean) | 15.4 | 10.8 | −30% |
| Tool calls (mean) | 22.4 | 14.6 | −35% |
| Pass (validate+format) | 5/5 | 5/5 | **no quality loss** |

The two arms' wall-clock ranges don't overlap, so the ~2× speedup is real at n=5 (not noise). The headline isn't just the mean — it's the **5.7× variance reduction**: without the playbook, agents flail (one baseline run hit 158s / 20 turns); with it, every run lands in a tight 54–69s band.

## Scope

Documentation-only. No code, no schema, no CLI behavior change. Affects only the `uipath-maestro-flow` skill's greenfield journey.


AlvinStanescu:
@claude please re-review — pushed `b128796`. Two changes since the last review:
- Fixed the broken anchor in SKILL.md rule #10: `#build-sequence--minimize-turns` → `#three-turn-execution-map` (matches the heading in greenfield.md).
- Rebased onto current `main`. The earlier review's "removals" (`--output-filter` in rule #1, the connector-discipline paragraph in rule #3, the description tweak) were artifacts of the branch being 70 commits stale — those are now preserved; the diff is purely additive (rule #10 + one batching anti-pattern in SKILL.md, plus the greenfield Three-turn restructure).

Smoke tests are green at 100% (14/14) on this commit.

rockymadden:
@nikhil-maryala Mind reviewing with the recent CLI optimization with flow init and the skill update, please? This looks like a good change.

uisherif:
can we prioritize this PR @nikhil-maryala @rockymadden ?

rockymadden:
> can we prioritize this PR @nikhil-maryala @rockymadden ?

Yes, working on it as I get IC time. Nikhil is DRI so he likely cannot do much. There are issues like coupling to the Flow file specifics that I need to still address.

rockymadden:
Digging in on this again.

rockymadden (approved):
Whew, what a journey to close this one out. Thanks, Alvin.

PR #1151, opened 2026-05-29
docs(uipath-ixp): add Unsupported capabilities section + audit routes

joe-prosser:
## What

Adds an **"Unsupported capabilities"** section to `skills/uipath-ixp/SKILL.md` so the skill responds deterministically to adjacent asks it can't serve — recognise the request, give a fixed standard response, and route the user to a verified in-product/docs path. Also adds a Task Navigation row for model-version rollback.

## Why

The skill covers a narrow happy path (list → review → confirm → improve → publish). For out-of-scope asks it had no defined fallback, so the agent either branched into open-ended discovery (`uip --help` / source grepping — banned by Critical Rule #1) or failed opaquely. Both are quality misses.

## Changes

New "Unsupported capabilities" entries, each with the exact standard response + a live-verified link:

| Ask | Routed to |
|-----|-----------|
| Create a model/project (no inputs) | `managing-projects` (context-aware: Project Setup Guide when docs/taxonomy supplied) |
| Upload files (no project context) | `building-and-deploying-models` (context-dependent: `documents upload` when a project is named) |
| Deploy to staging/prod/folder/environment/tenant | `building-and-deploying-models` |
| Delete a project/documents | `managing-projects` |
| Access / roles / permissions / sharing | `overview/managing-access` |
| Use a model in an automation / agent | `building-and-consuming-a-workflow` |
| Communications Mining | `cm-user-guide/introduction-to-uipath-communication-mining` |
| Runtime / operational monitoring | Orchestrator `about-monitoring` |

Plus a Task Navigation row: **roll back to a version** → `publish --model-version <N>` (supported, previously undocumented).

## Tickets

- RE-12339 (umbrella), RE-12341, RE-12342, RE-12343, RE-12346
- The last five entries + rollback row came from the RE-12346 audit.


skills/uipath-ixp/SKILL.md
```diff
@@ -55,6 +55,7 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
 | "Label documents" / "Review predictions" | [Label Documents Guide](references/label-documents-guide.md) |
 | "Improve scores" / "Fix prompts" / "Improve F1" | [Improve Prompts Guide](references/improve-prompts-guide.md) |
 | "Publish the model" / "Tag as live" | `uip ixp projects publish <project-name> --output json` |
+| "Roll back to a previous version" / "Restore version N" | `uip ixp projects publish <project-name> --model-version <N> --output json` — re-publishes an earlier version. Get available versions from `uip ixp projects list-models <project-name> --output json`. |
```
cezara98t, line 58: Now that I see this.. maybe we shouldn't have the "publish latest model" command, only the one for publishing a specific version, for which the agent already analyzed the score, as the version could be a newer one at the moment of publishing. WDYT?

skills/uipath-ixp/SKILL.md
```diff
@@ -68,6 +69,21 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
...
+| "Upload these files" / "add documents" — no existing project named | "Name an existing project and I'll upload with `uip ixp documents upload <project-name> <file> --output json`. To upload in-product (e.g. a new project), see https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/building-and-deploying-models." If a project context already exists, run the upload directly — see the "Upload a document" row in Task Navigation. |
+| "Deploy this model" / "push to staging / production / folder / environment / tenant" | "I don't drive deployment to folders, environments, or tenants — that's a product-side flow: https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/building-and-deploying-models. I can publish a model version with `uip ixp projects publish <project-name> --output json`, but binding it to a folder/environment is done in-product." |
+| "Delete this project" / "delete these documents" | "I don't delete projects or documents — there's no CLI verb for it. Delete in-product: https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/managing-projects." |
```
cezara98t: the delete-document functionality is available in the cli, soon to be added in the skill
https://github.com/UiPath/cli/blob/main/packages/ixp-tool/src/commands/documents.ts#L371

skills/uipath-ixp/SKILL.md
```diff
@@ -68,6 +69,21 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
...
+| "Give X access" / "share this project" / "change roles or permissions" | "Access, roles, and permissions are managed in-product, not through this skill: https://docs.uipath.com/ixp/automation-cloud/latest/overview/managing-access." |
+| "Use this model in my automation / workflow / agent" / "call the extractor from a process" | "Consuming a published model inside an automation is an authoring task outside this skill. See https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/building-and-consuming-a-workflow." |
+| "Mine these emails / communications" / "set up Communications Mining" | "Communications Mining is a separate IXP capability this skill doesn't cover (this skill is document extraction). See https://docs.uipath.com/ixp/automation-cloud/latest/cm-user-guide/introduction-to-uipath-communication-mining." |
```
cezara98t, line 84: is this needed? If the agent is asked something out of the scope, it can handle the response on its own. Do we need to link the CM docs? The agent could also be asked to set up a DU Modern project, but it can simply answer that it doesn't have this capability. 🤔

skills/uipath-ixp/SKILL.md
```diff
@@ -68,6 +69,21 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
 | ModelVersion doesn't advance | Retrain still in progress | Any change to model inputs (labellings OR instructions) triggers a full retrain. Wait ~2 min then retry. |
 | Field instructions conflict with label_def instructions | `update-prompts --fields` only edits per-field instructions, NOT the parent label_def instructions | Before iterating, read the label_def `instructions` and ensure they don't contradict your per-field instructions. |
 
+## Unsupported capabilities
+
+These requests fall outside the skill. Recognise the request, reply with the standard response, route the user. Do NOT enter discovery (`uip --help`, grep, source reading) — see Critical Rule #1.
+
+| User request | Standard response |
+|--------------|-------------------|
+| "Create a model" / "create a project" — no documents or taxonomy provided | "I work on existing IXP projects rather than creating them from scratch. Create one in-product: https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/managing-projects — then I can label, review, and improve it." If the user supplies documents or a taxonomy, use the [Project Setup Guide](references/project-setup-guide.md) instead. |
```
cezara98t: https://github.com/UiPath/skills/blob/main/.claude/rules/skill-review.md?plain=1#L77
Explicit `if user supplied X → Y / else → Z` removes the parsing step and makes both branches equally visible, which is exactly what skill-review.md calls out under LLM Usability. Same for the one below.

skills/uipath-ixp/SKILL.md
```diff
@@ -68,6 +69,21 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
...
+| "Upload these files" / "add documents" — no existing project named | "Name an existing project and I'll upload with `uip ixp documents upload <project-name> <file> --output json`. To upload in-product (e.g. a new project), see https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/building-and-deploying-models." If a project context already exists, run the upload directly — see the "Upload a document" row in Task Navigation. |
```
cezara98t: explicitly stating the command is not needed. same for the one below

skills/uipath-ixp/SKILL.md
joe-prosser: Yeah that sounds sensible to me. In general I would only recommending deploying a model version you've checked in validation so that would kind of enforce that pattern

skills/uipath-ixp/SKILL.md
joe-prosser: This was a specific requirement that came out of the bug bash - if asked to do something that is not supported by the CLI then we should redirect to docs

skills/uipath-ixp/SKILL.md
joe-prosser: Cool, have removed

skills/uipath-ixp/SKILL.md
joe-prosser: Cool, will remove

cezara98t (approved):

PR #1153, opened 2026-05-29
test(uipath-maestro-flow): address PascalCase issue in flow check

rockymadden:
### Problem
`tests/tasks/uipath-maestro-flow/_shared/flow_check.py` read the `uip maestro flow debug --output json` payload with **camelCase** keys (`finalStatus`, `variables`, `globals`, `elements`, `outputs`), but the CLI emits **PascalCase** (`FinalStatus`, `Variables`, `Globals`, `Elements`, `Outputs`).

A flow that runs correctly and completes was graded as a failure: `payload.get("finalStatus")` returned `None`, so `run_debug` raised `FAIL: Flow did not complete (finalStatus=None)` even though the payload contained `"FinalStatus": "Completed"`. The `file_attachment` e2e task scored 0.500 / FAILURE despite a correct, completed flow.

### Root cause
Key-casing mismatch in the test checker only. The CLI, the skill, and the agent's flow were all correct — `flow_check.py`'s nested payload reads had drifted to camelCase while the top-level `Data` read was already PascalCase. Both symptoms stemmed from this:
1. `run_debug` misread the completion status → `finalStatus=None` failure.
2. `collect_outputs` read camelCase `variables` → returned `[]`, so `assert_outputs_contain` found nothing even when the basename was present under `Variables.Globals.FileName`.

### Fix
Added `_get_ci(mapping, *candidate_keys, default=None)` — a case-insensitive key accessor — and routed every **runtime-payload** read through it (`run_debug`, `collect_outputs`). Chose case-insensitive over hardcoded PascalCase so the reader tolerates either casing and any future CLI normalization.

`.flow`-**source** readers (`read_flow_*_vars`, `_iter_flow_nodes`, node-type asserts) deliberately keep their camelCase literals — the source schema is camelCase and matches the skill docs. The `_is_flow_project` manifest read (`ProjectType`) is unchanged.

Module docstring updated with a "Payload key casing" section documenting the two sources/casings.

### Verification
- All existing canonical unit tests in `_shared/test_flow_check.py` pass (they use camelCase payloads) — confirms backward compatibility.
- The previously-failing `check_file_attachment_flow.py` passes end-to-end against the recorded PascalCase payload, including with a randomly-generated basename it never saw.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
@@ -13,6 +13,22 @@ Runs ``uip maestro flow debug --output json`` and asserts:
    timestamps, GUIDs, and status strings whose digits/chars can falsely match
    tiny expected values (e.g. ``"3" in json.dumps(data)`` is almost always
    true whenever a debug run completes).
+
+Payload key casing
+------------------
+Two distinct sources with two casings:
+
+- The ``flow debug --output json`` RUNTIME payload uses **PascalCase** keys
+  (``Data``, ``FinalStatus``, ``Variables``, ``Globals``, ``Elements``,
+  ``Outputs``, and the file-attachment object's ``Id``/``FullName``/``MimeType``/
+  ``Metadata``). Every runtime-payload read goes through :func:`_get_ci`, a
+  case-insensitive accessor — so the conceptual camelCase key names used in this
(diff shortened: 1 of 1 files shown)


tmatup (approved):

tmatup:
CLI change that introduced PascalCase https://github.com/UiPath/cli/pull/2266

PR #1157, opened 2026-05-29
fix(tests): cleanup_solutions — handle 404 + surface real failure reason

nikhil-maryala:
## Summary
- Follow-up to #1156. Investigation of the `cleanup_solutions.py` warnings on the `skill-flow-bindings-idempotent-reconfigure` runs surfaced two bugs in the post-run cleanup script.
- **Bug 1 — stderr vs stdout:** `uip solution delete` writes its envelope (including the failure `Message`) to **stdout** and leaves stderr empty. The script logged `r.stderr.strip()` on failure, so every warning printed `exit 1:` with no context. Parse the envelope to surface the actual `Message` (and only fall back to stderr if the envelope is unparseable).
- **Bug 2 — 404 ≠ failure:** Tasks that build a solution locally but never call `solution upload` or `maestro flow debug` produce a `.uipx` carrying a locally-generated `SolutionId` from `solution init`. That ID is never registered in Studio Web, so `solution delete` returns 404. Classifying that as `failed` produced spurious noise on every offline task and would dilute the warning channel that a real cleanup failure needs. Treat 404 / "Not Found" as `not_uploaded` (info-level, distinct counter in the summary).

## Repro

From the `skill-flow-bindings-idempotent-reconfigure` runs (both passed all assertions but emitted):

```
cleanup_solutions: failed to delete 630de3cf-b3f6-4f81-bff1-da28961ee607 (exit 1):
cleanup_solutions: summary policy=always deleted=0 preserved=0 skipped=0 failed=1
```

Running `uip solution delete 630de3cf-...` directly reveals what the script was hiding:

```json
{
  "Result": "Failure",
  "Message": "Delete failed (404): {\"title\":\"Not Found\",\"status\":404,...}",
  "Instructions": "Check that the solution ID is correct and that you have permission to delete it."
}
```

…with that JSON on **stdout**, not stderr.

## After

```
cleanup_solutions: solution 630de3cf-b3f6-4f81-bff1-da28961ee607 not uploaded (from BindingsIdempotent/BindingsIdempotent.uipx); nothing to clean up
cleanup_solutions: summary policy=always deleted=0 preserved=0 skipped=0 not_uploaded=1 failed=0
```

## Test plan
- [x] Run patched `cleanup_solutions.py` in the preserved sandbox of both prior runs (`runs/2026-05-29_04-04-25/...` and `runs/2026-05-29_11-23-13/...`). Each logs `not_uploaded=1 failed=0`, no warnings.
- [ ] Genuine-failure path (non-404) — would require an uploaded solution we lack permission to delete, or a 5xx. Visually verified the new code path logs `envelope_msg or r.stderr.strip()`; left to maintainers if a fixture is desired.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/cleanup_solutions.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/cleanup_solutions.py
@@ -50,6 +50,7 @@ def main() -> int:
     deleted: list[str] = []
     preserved: list[str] = []
     skipped: list[str] = []
+    not_uploaded: list[str] = []
     failed: list[str] = []
 
     for path in paths:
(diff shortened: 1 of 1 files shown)


tmatup (approved):

rockymadden (approved):
Thank you!

PR #1082, opened 2026-05-27
fix(uipath-maestro-flow): eliminate false no connection exists

rockymadden:
### Problem

During capability discovery for a connector flow, the skill could wrongly report that no Integration Service connection exists — and prompt the user to create one — when healthy connections were actually present. Two skill defects collided:

1. **Guessed connector key.** Nothing forbade inferring a connector key from the service's brand name instead of deriving it from `registry search`. The registry key is frequently prefixed or qualified differently than the product name, so a guessed key silently misses the real connector, and `uip is connections list "<wrong-key>"` returns "No connections found" — indistinguishable from a real empty result.

2. **`--all-folders` inconsistency across phases.** The implementation docs mandate `uip is connections list --all-folders`; the planning docs — which govern capability discovery — omitted it. Without the flag the CLI returns only the active folder and hides connections the user can see, so an empty result is a false negative. Following the planning doc produced a misleading empty result that drove a user-facing "create a connection" question on a false premise.

A whole-skill audit confirmed the defect was confined to the planning-phase docs and the top-level rules; the implementation phase was already correct and supplied the pattern to copy.

### Changes

**SKILL.md — top-level guardrails**
- Extended Critical rule #3 ("search before deciding it doesn't exist") with a connector-connection clause: the key MUST come from a `registry search` node type; discovery `connections list` MUST pass `--all-folders`; an empty result from an unverified key or missing flag is a false negative, never authoritative.
- Added an anti-pattern forbidding concluding "no connection exists" from a bare `uip is connections list`.

**Planning docs — `--all-folders` mandatory in discovery**
- Fixed four bare `uip is connections list` commands/references in `planning-arch.md`, `connector/planning.md`, and `http/planning.md` to the `--all-folders` form, each with the mandatory-flag note copied verbatim from the implementation docs so planning and impl read identically.

**Planning docs — key provenance + empty-result recovery**
- Added a "never type a connector key from memory" precondition to both connection-check sections and the connector-trigger Prerequisites.
- Replaced bare "if no connection exists → Open Questions" guidance with a gated rule: an empty result is treated as real only after (a) the key came from `registry search`, (b) the call used `--all-folders`, and (c) a `--refresh` retry was still empty — then cross-linked the shared platform-skill recovery path.

### Scope
- Implementation-phase docs were already correct and are unchanged.
- No diagnose/operate/evaluate docs reference a discovery `connections list`.
- CLI active-folder default behavior is out of scope (platform/CLI, not this skill).

PR #1159, opened 2026-05-29
fix(tests): pre-bake @uipath tool plugins in the skills eval image

tmatup:
## Problem

`skill-flow-outlook-trigger-inbox` has been red since **2026-05-27** (weighted score `0.667` — only the weight-3 `check_folder_id_fresh` criterion fails) with:

> `No compatible alpha version of '@uipath/orchestrator-tool' found for CLI (needs 1.2.x or lower).`

**This is an eval-infra failure, not an agent failure.** On the red runs the agent's flow validates, has the right trigger node, and resolves the MailFolder against the bound connection — 3/4 criteria pass. The 4th fails because the *checker's* first call, `uip or folders get`, can't resolve the `@uipath/orchestrator-tool` CLI plugin.

## Root cause

- The eval image baked **only `@uipath/cli`** — no tool plugins. Every `uip <verb>` relied on the CLI's on-demand auto-install at eval time.
- The `@uipath` tool plugins publish CLI-compatible builds **only as `-alpha` to the private GitHub Packages feed**; **public npm has stable releases only (no alpha channel)** — confirmed by direct query (`@uipath/orchestrator-tool` on GH feed has `1.1.0-alpha…`/`1.2.0-alpha…`; on npmjs.org it stops at stable `1.1.0`).
- At eval time the sandbox pins `NPM_CONFIG_PREFIX` to a task-local `.npm-prefix` (coder_eval `sandbox.py`, MST-9674). So the CLI's npmrc lookup misses the image's `@uipath:registry` config and `resolveRegistry()` falls back to `DEFAULT_REGISTRY` (public npm) → no alpha → `null` → the error.

## Why it "suddenly" broke on 05-27 (it was a *latent* bug, masked)

The checker could never resolve orchestrator-tool itself. It stayed green only because the **agent** happened to run `uip or folders get` during its own session (its env reaches the GH feed), pre-installing the plugin for the checker to reuse. On 05-27 the agent's connection-resolution path drifted to `uip is connections list` only (no `uip or`) — proven from the run transcripts — so the plugin was never installed and the checker became the first/only consumer. Skill PR #1082 (merged 05-27 evening) then codified the `uip is … --all-folders` recipe, locking the `uip is`-only path in for 05-28/05-29.

## Fix

Pre-bake the three plugins the maestro-flow checkers invoke (`uip or`/`uip is`/`uip maestro`) into the image via `uip tools install`. The **build** env reaches the GH feed, so the plugins land in the global `@uipath` dir (`PLUGIN_TOOLS_DIR`) and are discovered on disk at eval time — **no runtime registry resolution**. `set -euo pipefail` makes a feed/registry outage fail the image build **loudly** instead of silently scoring agents 0 on every orchestrator/IS-dependent checker.

## Verification

Ran the exact build commands in an isolated temp prefix (mirroring the Dockerfile):
- `uip tools install @uipath/orchestrator-tool` / `…integrationservice-tool` / `…maestro-tool` → all `Success`, auto-resolving the CLI-matching alpha `1.2.0-alpha.20260529.7350` (in lockstep with `CLI_VERSION`).
- `uip tools list` discovers `or` / `is` / `maestro`; all three land in `…/lib/node_modules/@uipath/` next to `cli` — i.e. the runtime `PLUGIN_TOOLS_DIR`.

(Could not run a full `docker build` here — no docker/base image in this environment — but the substantive install/discovery step is verified above.)

## Follow-ups (not in this PR)

- Harden the checkers to emit a `SKIP`/infra-error signal (not `FAIL`) when an infra `uip` command is unavailable, so a tooling outage never reads as an agent regression.
- The workflow comment in `run-coder-eval.yml:108` ("`@uipath/cli` + `@uipath/admin-tool`") doesn't match what the Dockerfile bakes — worth reconciling.


rockymadden (approved):

tmatup:
### CI triage — both red checks are an Anthropic workspace quota block, not a PR defect

Both failing checks share one out-of-repo root cause and ran no actual review/lint work:

| Check | Root cause | Status |
|---|---|---|
| **Claude Code Review** | `API Error: 400 You have reached your specified workspace API usage limits. You will regain access on 2026-06-01 at 00:00 UTC.` | Errored before reviewing |
| **Lint changed task YAMLs** | Same `400 … workspace API usage limits` (it's a Claude Code Action sharing the workspace key) | Errored before linting |

Neither is a defect in this PR:
- The diff touches **only `tests/docker/Dockerfile`** — there is **no task YAML** for the YAML linter to act on.
- All substantive checks pass: **Run skill smoke tests** (8m28s), **maestro-flow checker unit tests**, **Validate task schema (advisory)**, **Socket Security**, and all `detect` jobs.
- PR state is `MERGEABLE` / `UNSTABLE` — the two quota-blocked Claude Actions are non-blocking.

**Action:** no code change. Re-run the two Claude Code Action jobs once the workspace API quota resets/raises (won't pass before **2026-06-01 00:00 UTC**). Re-running before then will fail identically.

PR #1163, opened 2026-05-29
Tag and stabilize Flow eval scenarios

mjnovice:
## Summary
- Add `feature:eval` to the four `uipath-maestro-flow` evaluation scenarios and add `eval` to the documented feature tag vocabulary.
- Stabilize `skill-flow-eval-run-e2e`: use one deterministic data point, add Studio Web cleanup, and document generated evaluator refs so eval sets do not store unresolved display-name refs.
- Bound the live Studio Web eval wait: use `--wait --timeout 600`, save `eval-run-status.json`, poll the same run only a limited number of times, and fail with the dependent eval-run status instead of retrying or relying on a broad 2400s timeout.
- Add a guard that the agent starts exactly one eval run; terminal `Failed` status is captured, not retried.

## Hosted recheck
- Dispatched single-task run on this branch: https://github.com/UiPath/skills/actions/runs/26663570171
- Cancelled at 28m15s to avoid burning the full 2400s cap.
- Partial task log showed the flow was built/uploaded and the eval run reached terminal `Failed` with `workloadFailed`: "Job did not complete within 120 poll attempts". The previous prompt then tried to start a fresh eval run; this PR now prevents that and surfaces the downstream Studio Web eval status via `eval-run-status.json` / `check_eval_run.py`.

## Validation
- `git diff --check`
- PyYAML parse and assertions for `tests/tasks/uipath-maestro-flow/evaluate/eval_run/eval_run.yaml`
- `python3 -m py_compile tests/tasks/uipath-maestro-flow/evaluate/eval_run/check_eval_run.py`
- Ran `check_eval_run.py` against a representative passing result
- Ran `check_eval_run.py` against a representative `workloadFailed` status and verified the diagnostic failure message

## Not run
- Full passing live coder-eval for `skill-flow-eval-run-e2e`; the hosted recheck hit a Studio Web eval workload failure rather than a passing eval result.


tests/tasks/uipath-maestro-flow/evaluate/eval_run/eval_run.yaml
```diff
@@ -123,21 +139,43 @@ success_criteria:
...
     command_pattern: '(uip|\$UIP)\s+maestro\s+flow\s+eval\s+add\s+.*--inputs'
-    min_count: 3
+    min_count: 1
     weight: 2.0
     pass_threshold: 1.0
 
+  - type: command_not_executed
+    description: "Agent did not pass evaluator display name to eval set add"
+    tool_name: "Bash"
+    command_pattern: '(uip|\$UIP)\s+maestro\s+flow\s+eval\s+set\s+add\s+.*--evaluators(=|\s+)["'']?greeting-match["'']?'
```
uipreliga: **High — regex false-positives on the correct generated ref.** This is the one criterion meant to *enforce* the new "never pass a display name" guidance, but it also fails an agent that does the right thing.

`greeting-match` has no trailing boundary, and the closing `["']?` matches zero chars, so the prefix matches inside a valid generated ref. Verified:

```
--evaluators greeting-match                  → MATCH (intended: forbidden ✓)
--evaluators greeting-match-62c0793a.json    → MATCH (UNINTENDED — correct form ✗)
--evaluators greeting-match-62c0793a         → MATCH (UNINTENDED — correct form ✗)
```

The happy path (omitting `--evaluators` entirely, which the prompt steers toward) still passes — but any agent that legitimately passes an explicit generated ref fails this 1.5-weight check.

Fix: anchor the end so the display name can't be a prefix of a longer token, e.g. append a boundary that excludes `-`:
```
command_pattern: '(uip|\$UIP)\s+maestro\s+flow\s+eval\s+set\s+add\s+.*--evaluators(=|\s+)["'']?greeting-match["'']?(?![-\w])'
```
A bare `(\s|$)` anchor is not enough on its own here, since the display name is a strict prefix of the generated ref — the negative lookahead `(?![-\w])` is what isolates the forbidden form.

tests/tasks/uipath-maestro-flow/evaluate/eval_run/eval_run.yaml
```diff
@@ -3,17 +3,20 @@ description: >
...
   uipath-maestro-flow skill.
-tags: [uipath-maestro-flow, e2e, mode:operate, lifecycle:execute]
+tags: [uipath-maestro-flow, e2e, mode:operate, lifecycle:execute, feature:eval]
 
 run_limits:
   max_turns: 90
-  turn_timeout: 1200
+
+post_run:
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/cleanup_solutions.py"
```
uipreliga, line 18: **Low/Medium — this cleanup hook may be a no-op for this task, and its docstring is now stale.**

Two things:
1. `cleanup_solutions.py`'s docstring says it deletes solutions "uploaded by `uip maestro flow debug`". This task creates the Studio Web solution via `uip solution upload EvalRunDemo` — different path. Misleading for a maintainer reading why the hook is wired here.
2. More importantly: the script globs `**/*.uipx` and reads a top-level `SolutionId`. The skill's own `upload-safety.md` points at `SolutionStorage.json` as where `SolutionId` lives after upload. If `uip solution upload` does not write `SolutionId` into the `.uipx`, the script logs "no SolutionId … skipping" and deletes nothing — leaking the uploaded eval solution in the tenant on every CI run. It's best-effort/exit-0, so grading is unaffected, but the stated cleanup purpose may not be met.

Worth verifying against a real `solution upload` where `SolutionId` is actually persisted, and globbing that file if it's `SolutionStorage.json`. Update the docstring regardless.

uipreliga (commented):
## Code Review — `/code-review` (Gemini 3 + Codex + Opus, parallel)

Stabilization design is sound and the doc changes are clean and consistently propagated (CAPABILITY.md, commands-reference.md, eval-sets-guide.md, evaluators-guide.md, and the `eval_run.yaml` prompt all now agree: omit `--evaluators` for "all current", pass a generated id/file-ref if explicit, never the display name). `feature:eval` is added to the closed vocabulary and applied uniformly across all 4 tasks.

Two items worth addressing — see inline comments. Both were flagged independently by all three reviewers (the regex bug unanimously; the cleanup docstring/no-op by all three).

**Not fixing per request — comments only.**

### Summary
| Severity | Finding |
|----------|---------|
| **High** | `command_not_executed` regex false-positives on the *correct* generated ref (`eval_run.yaml:152`) |
| Low/Med | `cleanup_solutions.py` docstring stale + cleanup may be a no-op for the upload path (`eval_run.yaml:18`) |

### Non-blocking / no action
- `check_eval_run.py`: `EXPECTED_DATA_POINTS=1` makes the `len(rows) < 1` guard near-dead (`_extract_rows` already fails on empty), but it's harmless and keeps re-bumping trivial. The 8-status-key / 6-error-key fan-out in `_status_summary` is borderline speculative, but it's failure-path-only diagnostics that must never throw — defensible. `_load_json` hardening (JSON/OSError/non-dict) and `from __future__ import annotations` (keeps `str | None` safe pre-3.10) are correct.
- Dropping `turn_timeout` while keeping `max_turns: 90` is low-risk — `--timeout 600` + bounded polling cap the wait, and removing the per-turn cap avoids killing a legitimately long `--wait` turn.
- `max_count: 1` on `run start` cleanly enforces the "no second eval run" instruction.

uipreliga (approved):
Fix what you agree with and 🚢

akshaylive (approved):
Reapproval to enable merging.

PR #1165, opened 2026-05-31
docs(skills): make connector-mode HTTP constraints explicit - HasHttpRequest + application/json-only [ENGCE-58250]

mukundbayyaram:
## Summary

Make two connector-mode limits of Managed HTTP Request explicit where agents read and write:

1. **Connector mode needs HTTP support** - it only works if the connector exposes the HTTP request activity (`HasHttpRequest`).
2. **Connector mode is JSON-only** - it sends and accepts `application/json` only, for both request and response.

Manual mode has neither limit.

## Why

Both rules were easy to miss. The JSON-only limit lived in a single table cell scoped to the `http-request` helper, not the node itself, and was absent from the Flow node-authoring docs. Nothing told an agent to check `HasHttpRequest` before picking connector mode, or that a non-JSON body won't work - so it would author one anyway and fail at runtime.

## What changed

Stated both rules in the spots an agent hits while building an HTTP node:

- **platform `http-request.md`** - Connector-Mode Rules: added the `HasHttpRequest` check and the JSON-only rule; new Critical Rule #5; small wording fix to the "What is Managed HTTP Request?" definition.
- **`http/impl.md`** - `HasHttpRequest` verification at the connector step + JSON-only note at the configure step.
- **`http/planning.md`** - `HasHttpRequest` condition on the connector-mode row and a new Prerequisites bullet; JSON-only on the `body` row; sharper MUST-READ-FIRST pointer.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/references/plugins/http/impl.md
+++ b/skills/uipath-maestro-flow/references/author/references/plugins/http/impl.md
@@ -15,7 +15,7 @@ This page is the entry point for implementing a managed HTTP node. Pick your aut
 | **Connector** | Target system has an IS connector — auth via an existing IS connection (OAuth/API key) | [impl-connector.md](impl-connector.md) |
 | **Manual** | No connector exists, public/no-auth API, or quick prototyping | [impl-manual.md](impl-manual.md) |
 
-> **Mode fallback — auto-try, then confirm.** Prefer a curated connector activity first; if none, use connector mode; if connector mode cannot be configured (no usable IS connection), fall through to manual mode automatically. Manual changes the auth model (you supply auth yourself), so confirm the switched mode with the user via `AskUserQuestion` before finalizing. See [impl-connector.md — Step 2](impl-connector.md#step-2--identify-target-connection).
+> **Mode fallback — auto-try, then confirm.** Prefer a curated connector activity first; if none, use connector mode; if connector mode cannot be configured (connector lacks HTTP request support — `HasHttpRequest` false — or no usable IS connection), fall through to manual mode automatically. Manual changes the auth model (you supply auth yourself), so confirm the switched mode with the user via `AskUserQuestion` before finalizing. See [impl-connector.md — Step 2](impl-connector.md#step-2--identify-target-connection).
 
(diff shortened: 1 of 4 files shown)


skills/uipath-maestro-flow/references/author/references/plugins/http/impl.md
```diff
@@ -63,6 +65,8 @@ uip maestro flow node configure <ProjectName>.flow <nodeId> \
   }' --output json
 ```
 
+> **Connector mode supports `application/json` request and response bodies only.**
```
baishalighosh: Provide a rule explaining how to identify file requests and suggest that in most cases file operations might deviate from this pattern, so look into the file request/response for these 
X patterns to determine the same.

Examples of such patterns  are:-
1. The API endpoint path contains words like upload, download, content, file, attachment, blob, media, stream
2. The vendor docs specify Content-Type: multipart/form-data or application/octet-stream on the request
3.The response returns binary data, a stream, or a base64-encoded blob rather than a JSON object
4. The operation is described as "upload a file", "download a file", "get file content", "attach a file"

skills/uipath-maestro-flow/references/author/references/plugins/http/planning.md
```diff
@@ -92,6 +92,7 @@ Run `uip maestro flow node configure` with a `--detail` JSON. The CLI builds the
 ## Prerequisites
 
 - `uip login` required (for both modes — node type comes from registry)
+- For connector mode: connector must support the HTTP request activity — verify the `HasHttpRequest` flag via `uip is connectors get "<connector-key>"`. See [impl.md Step 2](impl.md#step-2--identify-target-connector-and-connection-connector-mode-only).
```
baishalighosh: If the connector does not support http , prompt the user with AskUserQuestion if they are interested to take the manual route (assuming the connector is there)

skills/uipath-maestro-flow/references/author/references/plugins/http/impl.md
mukundbayyaram: Addressed in f4f2d553. Added to the JSON-only rule in the platform `http-request.md` (both planning & impl link there) - file upload/download called out as the common violation, check content types before authoring, fall back to a curated activity or ask the user.

Skipped the path-keyword patterns - endpoints like GET /files/{id} often return plain JSON metadata, so they'd false-positive.

skills/uipath-maestro-flow/references/author/references/plugins/http/planning.md
mukundbayyaram: Done in f4f2d553. Added in `impl.md` Step 2 where the `HasHttpRequest` check runs - if false, stop and ask the user (manual mode or skip).

baishalighosh (approved):

PR #1102, opened 2026-05-28
fix(codedagents): Continue with maestro flow skill

alessandromuresan:
## Description

Expanded the coded-agents skill to own the in-solution sibling pattern end-to-end (solution init → flow scaffold → agent build → register → wire → validate) instead of relying on hand-offs to `uipath-maestro-flow`. Empirical run data showed that runs which stayed inside `uipath-agents` for the full sequence succeeded, while runs that bounced through `uipath-planner` or expected a maestro-flow hand-off stalled. Scenario 2 in the coded quickstart is now a complete numbered workflow, and `embedding-in-flows.md` no longer assumes the solution / flow project already exist. The deployment lifecycle reference adds an "idempotent re-deploy" section so re-publish-on-conflict and `flow registry search` empty-results don't trigger version-bump or discovery loops. All changes are scoped to `skills/uipath-agents/`; no edits to `uipath-maestro-flow` or `uipath-planner`.

Proof of tests passed (I temporarily added the \`smoke\` tag to the affected evals): https://github.com/UiPath/skills/actions/runs/26632192121/job/78483562435?pr=1102

PR #1172, opened 2026-06-01
fix(evals): skill-agents-deeprag-coded-smoke-trigger — fix prompt + clarify skill scope

EusebiuJecan07:
## Problem

`skill-agents-deeprag-coded-smoke-trigger` failed on June 1st after #1102 added 84 lines to `quickstart.md`. The prompt had "Tell me which skill applies" — with more skill content loaded, the agent answered directly from knowledge instead of invoking the skill.

## Changes

**`smoke_trigger.yaml`** — removed "Tell me which skill applies and outline the pipeline" from the prompt. Kept "Do NOT scaffold any files". Plain task description is enough to trigger activation.

**`quickstart.md`** — added `## Scope` section clarifying the skill applies to design/planning requests, not only build tasks.

## Validation

- [x] 2/2 local runs pass (score 1.0)

### DIFF
--- a/tests/tasks/uipath-agents/deeprag_coded/smoke_trigger.yaml
+++ b/tests/tasks/uipath-agents/deeprag_coded/smoke_trigger.yaml
@@ -6,14 +6,11 @@ description: >
   Python-coded-agent + PDF/TXT signal pair.
 tags: [uipath-agents, smoke, coded, lifecycle:discover, context-grounding]
 
-run_limits:
-  expected_turns: 5
-
 initial_prompt: |
   I have a Python coded-agent project (`pyproject.toml` with uipath +
   uipath-langchain, plus `langgraph.json`). I want it to receive a PDF
   attachment at runtime and produce a single grounded summary with
-  citations. Tell me which skill applies and outline the pipeline. Do NOT scaffold any files.
+  citations. Do NOT scaffold any files.
 
 success_criteria:
   - type: skill_triggered


skills/uipath-agents/references/coded/quickstart.md
```diff
@@ -31,6 +31,10 @@ When `project_state == local-workspace`, also read [lifecycle/local-workspace.md
 
 Use `uip codedagent <cmd>`, not `uv run uipath <cmd>`. The wrapper injects session credentials (`UIPATH_URL`, `UIPATH_ACCESS_TOKEN`, org/tenant identifiers) from your `uip login` session into the Python subprocess; `uv run uipath` skips that injection.
 
+## Scope
+
+This skill applies to **all** requests involving Python coded agents — not only build/scaffold tasks but also design, planning, architecture, and capability questions. If the user's request involves a Python project with `uipath-*` dependencies (LangGraph, LlamaIndex, OpenAI Agents, coded functions), load this skill and answer from it. Do not answer from general knowledge without invoking the skill first.
```
radu-mocanu: coded functions will be covered as part of a different skill

radu-mocanu (commented):
the agent has access to `quickstart.md` only **after** the skill is loaded. so the core problem is not fixed. if the coding agent is not loading the skill when it should then the skill.md description needs to change

skills/uipath-agents/references/coded/quickstart.md
EusebiuJecan07: True, now that we've moved it, they will load with a different skill.

EusebiuJecan07: Fixed — moved the `## Scope` section to `SKILL.md` (where trigger/activation scope belongs) and removed `coded functions` from the dependency list.

cosmyo (approved):

andreibalas-uipath (approved):

PR #1173, opened 2026-06-01
fix(uipath-agents): make uip agent guardrails list mandatory in recommend workflow

EusebiuJecan07:
## Problem

`skill-agent-guardrail-coded-recommend-scoped` failed because the agent skipped `uip agent guardrails list` and went straight from `catalog` to implementation. The guardrail was implemented correctly, but the list step is required to verify tenant availability — recommending a validator not available on the tenant would fail at runtime.

The skill docs already said the list step was mandatory but the wording wasn't strong enough to prevent the agent from skipping it.

## Changes

**`guardrails-recommend.md`** — added a `CRITICAL` callout directly above the `uip agent guardrails list` command making it explicit that this step must always run even when the catalog was already fetched, and why.

## Validation

- [ ] Daily run passes after this change

### DIFF
--- a/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails-recommend.md
+++ b/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails-recommend.md
@@ -39,6 +39,8 @@ The cache file is `.guardrails-catalog-cache.json` in the current working direct
 
 ### Guardrails List (NEVER cached — tenant-specific)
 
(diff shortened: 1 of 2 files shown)


cosmyo (approved):

radu-mocanu (approved):

EusebiuJecan07:
Closing — will be handled separately.

PR #1176, opened 2026-06-01
fix(tests): bindings checker — dedupe glob matches by realpath

nikhil-maryala:
## Summary

`skill-flow-bindings-multi-connector-independence` flakes between SUCCESS (7/7) and `3/7, score 0.200` runs with no code change between them. Today's cloud run and a local reproduction on `eb5da697` both failed with:

```
FAIL: Multiple files found for '**/BindingsMulti*.flow':
    ['BindingsMulti/BindingsMulti/BindingsMulti.flow',
     'BindingsMulti/flow_files/BindingsMulti.flow']
```

## Root cause

The task prompt asks the flow be saved at `BindingsMulti/flow_files/BindingsMulti.flow`. The CLI's `uip solution init` + `uip maestro flow create` actually puts the canonical file at `BindingsMulti/BindingsMulti/BindingsMulti.flow`. The agent has two legitimate readings:

1. Ignore the requested path; use canonical layout.
2. Honor the requested path via a symlink alias.

The agent is non-deterministic between the two. The 2026-05-29 successful run picked (1); today's failed run (and the cloud run) picked (2):

```
BindingsMulti/flow_files/BindingsMulti.flow → /tmp/.../BindingsMulti/BindingsMulti/BindingsMulti.flow
```

`_load_one` in `check_bindings.py` then runs `glob.glob('**/BindingsMulti*.flow', recursive=True)`, gets both paths back, and bails on the multi-match guard — but the symlink and its target are not duplicate files, they're one file at two paths. Four file-loading criteria die for the same reason (`structure`, `both_ids_present_from_json`, `no_empty_stubs`, `no_triplet_dups`).

Sibling tasks (`idempotent_reconfigure`, `reconfigure_different_connection`) reference explicit snapshot filenames at the sandbox root and don't hit this — only the recursive glob does.

This is a latent regression from #1032, which landed the multi-match guard and the recursive glob together. Neither is wrong in isolation; together they made the test flake against any agent that chose to alias.

## Fix

Dedupe `glob.glob` results by `os.path.realpath` so a symlink and its target count as one match. Both call sites updated:

- `_load_one` (used by every `*_from_json` and bindings criterion)
- `cmd_bindings_v2_no_empty_stubs` (separate glob path for derived `bindings_v2.json`)

The "multiple distinct files" failure path is preserved — only same-inode aliases collapse.

## Test plan

- [x] Fixture mirroring the live sandbox layout (canonical file + symlink at `flow_files/`): all four affected criteria pass.
- [x] Read the diff against the failing sandbox to confirm the symlink target is absolute (so realpath dedupe is the right key, not relative-path resolution).
- [ ] Full re-run of `skill-flow-bindings-multi-connector-independence` in cloud — should be 7/7 regardless of which interpretation the agent picks.

## Not in scope

- The other multi-binding tasks already pass on the patched checker:
  - `idempotent_reconfigure` (5/5)
  - `multi_connector_independence` (7/7 in the 2026-05-29 run when the agent skipped the symlink)
- `reconfigure_different_connection` is still failing 4/7 — that one is a genuine CLI bug (`upsertConnectionResourceBinding` fallback path appends instead of replaces) and is being tracked separately for a UiPath/cli issue.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/bindings/check_bindings.py
+++ b/tests/tasks/uipath-maestro-flow/bindings/check_bindings.py
@@ -68,6 +68,7 @@ Exit 0 on success; non-zero with stderr message on failure.
 from __future__ import annotations
 
 import glob
+import hashlib
 import json
 import os
 import sys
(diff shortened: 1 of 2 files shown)


tmatup:
> The task prompt asks the flow be saved at `BindingsMulti/flow_files/BindingsMulti.flow`. The CLI's `uip solution init` + `uip maestro flow` create actually puts the canonical file at `BindingsMulti/BindingsMulti/BindingsMulti.flow`


Do we want to address that problem?

tmatup (approved):

bai-uipath (approved):

gozhang2 (approved):

PR #348, opened 2026-04-22
fix(maestro-flow): prevent reference-ID reuse across connections

tmatup:
## Summary

- An Outlook \`email-received\` trigger faulted silently at runtime because Claude pasted a \`parentFolderId\` from a prior flow built against a different mailbox connection. Reference IDs (mailbox folders, Slack channels, Jira projects, etc.) are scoped to the account behind each connection, so the reused value passed \`node configure\` and \`flow validate\` cleanly but failed once the trigger fired — with no resolvable error back to Studio Web.
- This PR closes that gap with text-only edits: a top-level **Anti-Pattern** (always loaded on skill activation), a new **CRITICAL** section in the shared reference-resolution doc, prescriptive fresh-resolve gates in Step 3 / Step 4 of the `connector-trigger` and `connector` plugin `impl.md`, and a new **Debug-table** row naming the silent-fault symptom so the remediation is one lookup away next time.
- Out-of-scope but worth noting: the durable fix is CLI-side — \`node configure\` could ping the reference resource and reject unresolvable IDs at authoring time. Recommend filing separately.

## Test plan

- [ ] CI **Validate Skills** workflow (\`hooks/validate-skill-descriptions.sh\`) passes — confirms SKILL.md description stays under 250 chars.
- [ ] Fresh session, build an Outlook \`email-received\` trigger from scratch — agent calls \`uip is resources execute list ... MailFolder --connection-id <bound-id>\` for \`parentFolderId\` instead of using any literal from context.
- [ ] Same session, build a second Outlook trigger bound to a different connection — agent re-resolves \`parentFolderId\` against the second connection, does not carry the first flow's ID.
- [ ] Simulate the failure mode: manually paste a foreign \`parentFolderId\` into \`eventParameters\`, \`flow debug\`, observe silent fault — then confirm agent finds the new Debug-table row and offers the correct remediation.
- [ ] Spot-check rendering: the new top-level anti-pattern renders alongside existing siblings in Studio Web / skills marketplace UI without breaking layout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #1177, opened 2026-06-01
test(uipath-maestro-flow): tolerate CLI PascalCase output in task-local checkers (#2266)

tmatup:
## Problem

`#1153` made `_shared/flow_check.py` case-insensitive (`_get_ci`), fixing the shared runtime-payload reads after `uipath-cli #2266` started PascalCasing all `--output json` `Data` keys (cli ≥ `1.2.0-alpha.20260528`). But **task-local** checkers that parse CLI json directly were never swept, so the break just resurfaced one layer deeper:

| Checker | Lowercase read on CLI output | Symptom under #2266 |
|---|---|---|
| `outlook_trigger_inbox` | `f.get("id")` on `resources run list MailFolder` | every id → `None` → `live_ids == {None}` (cardinality **1**) → false *"… NOT among the 1 MailFolder IDs … reused a reference ID"* PR #348 regression, even though the agent freshly resolved the folder |
| `terminate` | `elementExecutions` / `status` / `elementId` on `flow debug` | `[]` → *"No element was Terminated. Statuses: []"* |
| `salesforce_opportunity_list` | `variables` / `globals` on `flow debug` | `{}` → "no records returned" |

(`terminate` actually *migrated* across #1153: the 05-29 run died at the `_shared` `finalStatus` gate; once #1153 fixed that, 06-01 cleared it and died at the task-local `elementExecutions` read.)

## Fix

- Route the `flow debug` payload reads in `terminate` and `salesforce_opportunity_list` through the existing `_get_ci(...)` helper.
- In `outlook_trigger_inbox`, read the MailFolder id under both `id`/`Id` and `discard(None)` so the live set can never collapse to `{None}`.

All reads stay **tolerant of both casings**, so the suite is green whether the sandbox runs the PascalCase CLI today or a camelCase one — including after `uipath-cli #2308` restores camelCase for `flow debug` + `resources run`. Defense-in-depth paired with that CLI fix.

Scope note: only the three checkers that read CLI json by lowercase key are touched. `.flow` *source* readers (e.g. `subflow`) intentionally keep camelCase and are left alone, per the `_get_ci` docstring.

## Tests

- New PascalCase fixtures lock `check_folder_id_fresh`: a matching id under `Id` now **passes** (fails on the pre-fix `f.get("id")` code), and a genuinely stale id under PascalCase **still fails** with the PR #348 marker (the `discard(None)` fix doesn't weaken the check).
- New `_get_ci` camel/Pascal test locks the shared helper.
- Full `tests/tasks/uipath-maestro-flow/` suite: **98 passed**.

Root-cause writeup: `docs/uipath-skills/2026-06-01-skill-flow-outlook-trigger-inbox-root-cause.md`.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/test_flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/test_flow_check.py
@@ -13,6 +13,7 @@ import pytest
 sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
 from flow_check import (  # noqa: E402
+    _get_ci,
     assert_flow_has_exact_node_type,
     assert_flow_has_node_type,
     assert_flow_uses_connector_target,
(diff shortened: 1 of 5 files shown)


tmatup:
Superseded by #1178, which covers the same task-local checkers (terminate, outlook) plus the unique pieces of this PR (salesforce_opportunity_list has been folded in) AND the maestro-case suite + a CI contract-guard test that blocks future raw lowercase runtime-key reads. Closing in favor of #1178.

PR #1178, opened 2026-06-01
test(maestro-flow,maestro-case): make checkers casing-agnostic for CLI #2266 + contract guard

tmatup:
## Problem

uipath-cli PR #2266 made `OutputFormatter` recursively PascalCase every `--output json` `Data` key. PR #1153 hardened **only** `_shared/flow_check.py`, so task-local checkers that read the debug **runtime** payload by lowercase key kept breaking one layer deeper:

- **skill-flow-terminate** → `FAIL: … Statuses: []` (run `2026-06-01_04-04-22`). #1153 let the shared `finalStatus` gate pass via `_get_ci`, then the task-local `check_terminate_flow.py` read `payload.get("elementExecutions")` on a payload that now only has `ElementExecutions` → `[]`.
- **outlook_trigger_inbox** → live-ID set collapsed to `{None}` because items were keyed `Id`, not `id`.

The whole maestro-**case** suite was likewise exposed: `case_check.run_debug` and `collect_outputs` read the runtime payload by camelCase.

## Fix

Route **every runtime-payload read** through the case-insensitive `_get_ci` accessor, so a checker no longer depends on which CLI build the eval image runs:

- **maestro-flow:** `terminate` (`elementExecutions`/`status`/`elementId`), `outlook_trigger_inbox` (item `id`/`Id`, drop `None`)
- **maestro-case:** `_shared/case_check` — added `_get_ci`; hardened `run_debug` gate, `collect_outputs`, `_iter_runtime_tasks`, and the metadata-key filter — plus the 6 task-local `finalStatus` reads

## Contract guard (the part that prevents recurrence)

`tests/scripts/test_runtime_payload_key_casing.py` fails CI if any checker reads a **runtime-only** key (`finalStatus` / `elementExecutions`, which never appear in `.flow`/`caseplan` source) by a raw lowercase `.get()` / `[...]` instead of `_get_ci`. New maestro-case + guard jobs added to `test-helpers.yml`, and PascalCase unit fixtures added (flow `_get_ci`/`collect_outputs`, case `_get_ci`/`collect_outputs`/`run_debug` gate, outlook item `Id`).

## Test

- `pytest tests/tasks/uipath-maestro-flow/` 100 pass · `tests/tasks/uipath-maestro-case/` 6 pass · guard 1 pass · full sweep **121 pass**
- Guard confirms zero raw runtime-key reads remain in `check_*.py` / `_shared`

## Companion

Source-side fix (flow/case debug emit camelCase via `preserveDataKeys`) ships in UiPath/cli #2309. With both merged: the CLI emits camelCase **and** checkers tolerate either casing, with CI blocking the next raw-lowercase read.

### DIFF
--- a/.github/workflows/test-helpers.yml
+++ b/.github/workflows/test-helpers.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
     paths:
       - 'tests/tasks/uipath-maestro-flow/**'
+      - 'tests/tasks/uipath-maestro-case/**'
+      - 'tests/scripts/**'
       - '.github/workflows/test-helpers.yml'
   workflow_dispatch:
 
@@ -23,3 +25,38 @@ jobs:
 
       - name: Run pytest
         run: pytest tests/tasks/uipath-maestro-flow/ -v
+
+  pytest-case-check:
+    runs-on: ubuntu-latest
+    name: maestro-case checker unit tests
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - name: Install pytest
+        run: pip install pytest
+
+      - name: Run pytest
+        run: pytest tests/tasks/uipath-maestro-case/ -v
+
+  runtime-payload-casing-guard:
+    runs-on: ubuntu-latest
+    name: runtime-payload key-casing contract guard
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+
+      - name: Install pytest
+        run: pip install pytest
+
+      # Fails if any checker reads a runtime-only debug key (finalStatus /
+      # elementExecutions) by a hard-coded lowercase .get() instead of the
+      # case-insensitive _get_ci accessor — the CLI #2266 PascalCase break.
+      - name: Run contract guard
+        run: pytest tests/scripts/test_runtime_payload_key_casing.py -v
--- a/tests/scripts/test_runtime_payload_key_casing.py
+++ b/tests/scripts/test_runtime_payload_key_casing.py
@@ -1,0 +1,66 @@
+"""Contract guard: maestro flow/case checkers must read the debug RUNTIME
+payload case-insensitively.
+
+Background — uipath-cli PR #2266 made ``OutputFormatter`` recursively PascalCase
+every key of a ``--output json`` ``Data`` payload. Checkers written against the
+documented camelCase shape then read ``None`` and fail with misleading messages
+(``finalStatus=None``; ``Statuses: []`` for terminate; ``{None}`` ID sets for
+outlook). PR #1153 hardened ``_shared/flow_check.py`` but not the task-local
+checkers, so the break kept resurfacing one layer deeper (skill-flow-terminate,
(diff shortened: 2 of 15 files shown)


tmatup:
Folded in the `salesforce_opportunity_list` checker fix from #1177 (route `_assert_records_returned` through `_get_ci`). This PR now supersedes #1177 (closed) — it covers terminate, outlook, salesforce, the full maestro-case suite, the contract guard test, and the CI jobs.

tmatup:
### Review Response — Claude Code Review

Triaging every finding (verdict was *approve*; one note actionable, rest deliberate).

**8. Contract guard matched only exact camelCase** — `tests/scripts/test_runtime_payload_key_casing.py`

**Fixed** in fca20784. Added `re.IGNORECASE` so a raw read of a runtime-only key is now caught in **both** directions — `.get("finalStatus")` (breaks under a PascalCasing CLI) *and* `.get("ElementExecutions")` (breaks under a camelCase CLI). Safe from false positives because `finalStatus`/`elementExecutions` are runtime-only names at any casing. Added a positive regex test locking both-casing matching and asserting the `_get_ci` accessor form is *not* flagged.

**1. `collect_outputs` now skips `{"value": None}`** — `case_check.py`

**Keeping** — this is intentional and aligns the case helper with the flow helper. `_shared/flow_check.py` (PR #1153) already uses the same `value = _get_ci(...); if value is not None:` guard. A `None` output value is not a real declared output, so skipping it is correct; the old `"value" in v` path would have leaked a `None` leaf into the output set.

**5. outlook uses `f.get("id") or f.get("Id")` instead of `_get_ci`** — `check_outlook_trigger_inbox.py`

**Keeping** — deliberate, as you noted. That `id` comes from the `is resources run list` **API response**, not the `flow debug` runtime payload, and this checker has no other runtime-payload reads, so importing `_get_ci` from `flow_check` just for one IS-list item key would be gratuitous. The inline two-key fallback + `None` filter is self-contained and the contract guard correctly leaves `id` alone (it's not runtime-only).

**`_get_ci` duplicated in `flow_check.py` and `case_check.py`** (What's Missing)

**Keeping** — required by sandbox isolation, not an oversight. Each skill suite's checkers run in the eval sandbox with only their own `_shared/` copied alongside the task; a cross-suite `tests/lib/` module wouldn't be present at runtime. Divergence is guarded against by the parity tests in **both** suites (camelCase/PascalCase equivalence) plus the contract guard scanning both trees.

**Contract guard only watches `finalStatus`/`elementExecutions`** (Manual Review)

**Keeping by design** — the other runtime keys (`variables`/`outputs`/`value`/`globals`/`elements`) also appear in `.flow`/`caseplan` **source**, so guarding them would false-positive on legitimate source readers. The guard deliberately scopes to the two unambiguous runtime-only keys; a future runtime-only key would be added to `RUNTIME_ONLY_KEYS` (documented in the test).

**CI verification** (Manual Review)

**Confirmed green on this PR**: `maestro-flow checker unit tests` ✅, `maestro-case checker unit tests` ✅, `runtime-payload key-casing contract guard` ✅ (the 121-pass figure is these three jobs).

---

Nits / positive sections (#2, #3, #4, #6, #7, #9) — noted, no change needed.

jiyangzh (approved):

song-zhao-25 (approved):

charlesliu9 (approved):

PR #1181, opened 2026-06-01
docs(uipath-maestro-flow): document error edge handling toggle

tmatup:
## Summary

- Document that `sourcePort: "error"` requires `inputs.errorHandlingEnabled: true` on the source node.
- Call out that CLI edge-add and format commands self-heal the flag, while direct JSON edits must set it explicitly.
- Update the shared flow format reference, authoring checklist, topology guidance, and HTTP plugin implementation notes.

## Why

Studio Web only exposes the implicit error handle when the source node has error handling enabled. The authoring guidance previously explained the implicit `error` port but did not make this source-node flag visible to direct JSON authors.

## Validation

- `git diff --check HEAD~1..HEAD`

🤖 Generated with Claude Code


tmatup:
### Review Response

**Issues for Manual Review — does `validate` actually fail without the flag?** — `editing-operations-json.md:217`

**Confirmed accurate, and made consistent** in a9e6c577. `uip maestro flow validate` *does* fail when a node has an outgoing `sourcePort: "error"` edge without `inputs.errorHandlingEnabled: true` — that is exactly what UiPath/cli #2313 enforces (and #2315 generalizes via the manifest `visible` gate). So the line-217 claim is correct. To close the inconsistency the review flagged, the authoritative `file-format.md` now states the validate behavior too, so it no longer diverges from `editing-operations-json.md`.

Nits: the checklist (item 4) vs Add-an-edge duplication is intentional — quick-reference vs actionable instruction.

rockymadden (approved):

PR #1144, opened 2026-05-29
feat: add uipath-functions skill (Python Coded Functions)

AlexBizon:
Splits **Python Coded Functions** out of `uipath-agents` into a dedicated, self-contained skill. Functions are deterministic Python (typed `@dataclass` I/O, no LLM) with their own `uip functions` lifecycle — bundling them under agents caused mis-routing.

**Changes**
- **New:** `skills/uipath-functions/SKILL.md` (migrated 100% from the old `coded-functions.md` reference).
- **`uipath-agents` redirects:** detection + Framework Selection now route `[tool.uipath] type="function"` → `uipath-functions`; Coded-agent signal tightened to require a framework dep.
- **Tests:** migrated e2e (`e2e_lifecycle`), new smoke (`smoke_trigger`), 25-case activation jsonl registered in `activation.yaml` (`recall.yes ≥ 0.70`).
- **CODEOWNERS:** claims the skill, tests, and activation jsonl.

**Scope:** Python only. JS functions (BE for Coded Apps) land in a follow-up — needs description compression (already 700/1024), a `references/` split, and a `→uipath-coded-apps` redirect.

**Test plan:** activation + smoke + e2e pass (fill in run results).

PR #1183, opened 2026-06-01
Expand coder eval coverage for eval commands

mjnovice:
## Summary
- add Flow eval coverage for advanced local CRUD, simulation state, and run-history triage
- add coded-agent eval coverage for missing evaluator types and targeted `--eval-ids` output runs
- add low-code agent eval coverage for local CRUD and tenant-backed run lifecycle

## Validation
- `cd tests && make install`
- `coder-eval plan` for all 7 new task YAMLs
- Python YAML load + checker compile for all new files
- `coder-eval run` local batch: `skill-agent-lowcode-eval-local-crud` and `skill-agent-coded-eval-tool-call-classification` passed end-to-end
- Patched checkers pass against preserved real artifacts for advanced Flow local CRUD, Flow simulation state, and coded-agent subset output

Note: tenant-backed e2e tasks were not fully run locally before PR creation.


tmatup (approved):

uipreliga (approved):
Fix what you agree with and 🚢

tests/tasks/uipath-functions/eval_subset_output/eval_subset_output.yaml
```diff
@@ -0,0 +1,67 @@
...
+  The agent must run only one selected deterministic test case and save the
+  results, proving it can use targeted local eval runs during debugging.
+tags: [uipath-agents, integration, mode:build, lifecycle:generate, feature:eval]
+
+run_limits:
+  expected_turns: 28
+  max_turns: 60
+  turn_timeout: 1200
+
+initial_prompt: |
```
uipreliga, line 15: isn't it too detailed for an integration test?

tests/tasks/uipath-agents/coded/eval_tool_call_classification/eval_tool_call_classification.yaml
```diff
@@ -0,0 +1,56 @@
...
+  and multiclass classification. This is a local artifact test, not a cloud
+  reporting run.
+tags: [uipath-agents, integration, mode:build, lifecycle:generate, feature:eval]
+
+run_limits:
+  expected_turns: 30
+  max_turns: 70
+  turn_timeout: 1200
+
+initial_prompt: |
```
uipreliga, line 15: that seems a bit too detailed - would people actually execute a prompt like this?

tests/tasks/uipath-agents/lowcode/eval_local_crud/eval_local_crud.yaml
```diff
@@ -0,0 +1,104 @@
...
+  evaluator JSON, eval set and test-case CRUD, simulation flags, and validate.
+tags: [uipath-agents, integration, mode:build, lifecycle:generate, feature:eval]
+
+run_limits:
+  expected_turns: 35
+  max_turns: 80
+  turn_timeout: 1200
+
+initial_prompt: |
+  Use the `uipath-agents` low-code evaluation guidance to create a local
```
uipreliga: this might be too detailed - if you specify what skill to use, there is even no activation test.

tests/tasks/uipath-agents/lowcode/eval_run_lifecycle/eval_run_lifecycle.yaml
```diff
@@ -0,0 +1,146 @@
...
+  max_turns: 100
+  turn_timeout: 1800
+  task_timeout: 2400
+
+post_run:
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/cleanup_solutions.py"
+    timeout: 90
+
+initial_prompt: |
+  Use the `uipath-agents` low-code evaluation guidance to create a tenant-backed
```
uipreliga: as above - too detailed;

tests/tasks/uipath-maestro-flow/evaluate/advanced_local_crud.yaml
```diff
@@ -0,0 +1,152 @@
...
+  run is performed.
+tags: [uipath-maestro-flow, integration, mode:build, lifecycle:generate, feature:eval]
+
+run_limits:
+  expected_turns: 35
+  max_turns: 80
+  turn_timeout: 1200
+
+initial_prompt: |
+  Use the `uipath-maestro-flow` skill's evaluate capability to create a
```
uipreliga: if you specify what to use, there is even no activation test.

uipreliga (approved):
Fix the prompts - they are too detailed - there is no testing happening with prompts like this.

uipreliga:
## Code Review Summary

Multi-model review of the 14 new `coder_eval` task/checker files in this PR — **Gemini 3**, **GPT-5 Codex**, and an **Opus 4.x** sub-agent, all evaluating for correctness, KISS, DRY, and over-engineering. Reviewer agreement is noted per finding (independent flags = higher confidence). **No code was changed** — this is informational, as requested.

The lens that matters most for these files: *a checker must fail when the agent does the wrong thing and pass when it does the right thing.* Findings below are sorted by that risk.

### High Issues

- **Over-strict: `check_advanced_local_crud.py` grades on example strings the prompt never specifies** (`check_advanced_local_crud.py:136,140`) — *Opus*
  The checker hard-requires the literal `"deployment succeeded"` (search text) and `"invoice"`/`"receipt"` (input-file key). But `advanced_local_crud.yaml` only asks the agent to *"use `--search-text` for the contains evaluator"* and *"`--input-file` using a small local fixture file"* — it never pins those values. They come verbatim from the skill's *example* snippets. An agent that picks any other valid search string or file key produces correct artifacts and **fails the checker**. This grades coincidence, not behavior.
  **Fix:** either pin the exact strings in the prompt, or assert structure instead of content (a data point carrying contains/search-text criteria; a staged `--input-file` ref via a non-JSON file or `files`/`inputFiles` key).

### Medium Issues

- **`--output-file` result shape in the subset checker is unverified and likely wrong** (`check_eval_subset_output.py:59,68,73`) — *Opus*
  Assumes `{evaluationSetResults:[{evaluationRunResults:[{result:{score}}]}]}`. The only `--output-file` shape documented in the skill (`running-evaluations.md`) is flat: `{testId, testName, status, evaluationResults:[{evaluatorId, score}]}`. If the real envelope matches the docs, this checker fails on correct output at line 61 (`no evaluationSetResults`). **Fix:** confirm the actual `uip codedagent eval --output-file` JSON shape and align keys, or extract tolerantly (recursively find the one result row whose case-id contains `selected` with a `score == 1.0`).

- **Terminal-status allow-list is narrow and may reject valid terminal states** (`check_eval_run_lifecycle.py:104` and `check_run_history_triage.py:107`) — *Gemini + Codex (both flagged)*
  Accepts only `{completed, failed, 2, 3}` / `{completed, failed}`. A valid terminal status of `succeeded`/`passed`/`error`/`done` from the CLI would spuriously fail a real two-run lifecycle. **Fix:** widen the set after confirming the CLI's actual status vocabulary.

- **`assert_code` only checks top-level `Code`, but the same file uses a `Data` wrapper helper** (`check_eval_run_lifecycle.py:66` and `check_run_history_triage.py:66`) — *Codex*
  `data()` reads `doc.get("Data", doc)` (anticipating a `{Data:{…}}` envelope) yet `assert_code` reads `doc.get("Code")` only at top level. If the CLI wraps the response, `Code` won't be found and the checker fails on valid output. **Fix:** fall back to `doc["Data"].get("Code")` when the top-level key is absent.

- **`--only-failed --verbose` path is not actually validated for the lowcode lifecycle task** (`check_eval_run_lifecycle.py:118-119`) — *Opus*
  Unlike its Flow sibling (`check_run_history_triage.py:121-128`, which asserts `bad-bob` present and `good-alice` absent), the lowcode checker only prints the row count and accepts **zero rows** — so `--only-failed` could be entirely broken and the task still passes. **Note:** this is *partly intentional* — the lowcode prompt uses semantic-similarity scoring (non-deterministic) and explicitly says *"If there are no failed rows, still save the valid JSON response"* (`eval_run_lifecycle.yaml:51`), so it can't hard-assert which row failed without flakiness. Still, consider a softer assertion (e.g. if any rows exist, none should be `happy-alpha`) so the flag isn't a complete no-op.

- **`command_executed` patterns don't require `--output json`, while the prompts mandate it** (`eval_run_lifecycle.yaml:91-138`, `run_history_triage.yaml:93-138`) — *Codex*
  Largely mitigated in practice: the high-weight checker parses the saved `.json` files, so a non-JSON run fails `load()` downstream. But it diverges from the repo's own test-writing rule ("when criteria parse CLI output … add a low-weight check matching `(uip|\$UIP)\s+.*--output\s+json`"). **Fix:** add that low-weight `--output json` check for consistency.

### Low Issues

- **UUID set-distinctness check compares *all* UUIDs in the payload, not the run-id field** (`check_eval_run_lifecycle.py:109-116`, `check_run_history_triage.py:112-119`) — *Codex + Opus.* Shared tenant/eval-set/agent UUIDs could make the `ids_a == ids_b` guard noisy. Prefer reading a known `runId`/`id` key first, falling back to `collect_ids`.
- **`lifecycle:execute` tag is outside the README's documented closed vocabulary** (`eval_run_lifecycle.yaml:6`, `run_history_triage.yaml:6`) — *Opus.* README lists only `discover`/`generate`/`setup`, but `execute` is already used 14× across the repo — pre-existing taxonomy drift, not introduced here. Worth blessing `execute` in the README per rule 5.
- **`--eval-ids\s+` / `--output-file …\s` require a trailing space, so `--eval-ids=case-selected` would not match** (`eval_subset_output.yaml:33,41`) — *Gemini.* Minor; the space-separated form is the common one.
- **Selected-case assertion uses substring `"selected"` rather than exact id** (`check_eval_subset_output.py:65`) — *Codex (rated high; downgraded).* Safe for this fixed 3-id set since only `case-selected` contains `selected`, and the checker already pins the exact id set at line 48-51. Cosmetic — exact match would read cleaner.
- **DRY: `check_eval_run_lifecycle.py` and `check_run_history_triage.py` are ~95% identical** — *Codex + Opus.* They live under different skill dirs, so the repo's self-contained rule **forbids** sharing them (not under a `_shared/`). The duplication is therefore acceptable per repo rules — **no refactor recommended** (would violate KISS + self-contained). It is, however, exactly the divergence that produced the lowcode `--only-failed` gap above; keeping the two in parity by hand is the right move.

### Discarded / False Positives

- **`--report` negative guard wrongly flags `--no-report`** (Codex, *high*) — **false positive.** Verified: the literal string `--no-report` does not contain the substring `--report` (the chars are `--no-report`), so `.*--report` cannot match it. The guard at `eval_subset_output.yaml:54-59` is correct. The `uip\s+login\s+(?!status)` negative-lookahead is also correct.

### Positive Observations

- Every task pairs regex `command_executed` checks with a heavy-weight (5.0–7.0) on-disk `run_command` checker — graded on **behavior and artifacts, not agent monologue**. ✅ (all three reviewers)
- `find_project_root` (under the allowed `_shared/`) handles both flat and nested project layouts instead of hard-coding a path that flips per agent behavior.
- Evaluator type/category matrices in the checkers (`check_eval_local_crud.py:57-60`, `check_advanced_local_crud.py` type IDs) line up exactly with the skills' evaluator docs, including the lowcode-UUID-refs vs Flow-file-refs distinction.
- `check_simulation_crud.py` correctly tests documented overwrite semantics (one `Static` `lookupCustomer` after re-add, `draftEmail` removed) with tolerant key-name matching.
- `check_run_history_triage.py` is the model the lowcode sibling should follow: it asserts the seeded failing case is present and the passing case is filtered out.
- Safe, fail-fast JSON parsing throughout; no crashes on malformed agent output.

### Overall Assessment

Solid, behavior-focused coverage that aligns well with the skills' documented CLI and JSON shapes. The merge-blocking risk is over-strictness, not vacuous passes: **`check_advanced_local_crud.py` grading on unspecified example strings (High)** will fail correct agent output, and **the unverified `--output-file` envelope (Medium)** and **narrow terminal-status set (Medium)** carry the same flaky risk. Recommend confirming the real CLI/JSON shapes for those three before relying on the tasks; the remaining items are non-blocking polish.

<sub>🤖 Synthesized from Gemini 3, GPT-5 Codex, and Opus 4.x reviews via `/code-review`. Findings independently verified against the files where flagged. Generated with [Claude Code](https://claude.com/claude-code)</sub>

tests/tasks/uipath-agents/coded/eval_subset_output/eval_subset_output.yaml
```diff
@@ -0,0 +1,63 @@
...
+  results, proving it can use targeted local eval runs during debugging.
+tags: [uipath-agents, integration, mode:build, coded, lifecycle:generate, feature:eval]
+
+run_limits:
+  expected_turns: 28
+  max_turns: 60
+  turn_timeout: 1200
+
+initial_prompt: |
+  Build a Simple Function coded agent named `subset-eval` whose main function
```
radu-mocanu: the `uipath-agents` skill is no longer covering functions. this needs to be moved under uipath-functions skill after this [PR](https://github.com/UiPath/skills/pull/1144) is merged

PR #1150, opened 2026-05-29
test(uipath-troubleshoot): 3 scenarios for Use Excel File / Application Card failures playbook

Stefan-Virgil:
## Summary

Three coder-eval scenarios for the most card-distinctive cause-branches in `excel-application-card-failures.md` (B1 Excel-not-installed, B3 RPC race across scopes, B5 sensitivity label rejection). Each scenario mirrors the established template (canonical `run_limits` block, claude-code agent, two `success_criteria` with skill_triggered + llm_judge), with full process scaffold and synthetic mock fixtures.

| Scenario | Branch | Visible exception | Smoking gun in logs |
|---|---|---|---|
| `excel-card-excel-not-installed` | B1 no Excel | `BusinessException: Error opening workbook. Make sure Excel is installed.` (inner `COMException 0x80040154 REGDB_E_CLASSNOTREG`) | Classic scope on unattended host; host inspection shows no Excel install (no registry entry, no `excel.exe`) |
| `excel-card-rpc-race-across-scopes` | B3 RPC race | `COMException 0x80010108 RPC_E_DISCONNECTED` on `UseExcelFile_2` | Two sequential Modern cards with `ReadFormatting=True`, no outer `Excel Process Scope`; logs trace PID 8472 lifecycle: launch → first scope kills it → second scope races on dead apartment |
| `excel-card-sensitivity-label-rejection` | B5 Purview label | `BusinessException` referencing Microsoft Purview label `'Confidential - Client Data'` | Package v2.24.7 supports SensitivityLabel/Operation properties; card has both unset; workbook has Purview label applied |

All three exercise anti-pattern resistance:
- B1 — penalizes recommending `Visible: True` as a "diagnostic" (COM-forcing footgun)
- B3 — explicitly penalizes Delay-as-fix (the playbook's anti-pattern #1)
- B5 — explicitly penalizes package downgrade (anti-pattern #4 — pre-v2.23.4 silently strips labels = compliance violation)

Branches B2 (empty path) and B4 (child outside scope) are cross-cuts already covered by sibling tests (Write Range B1 / per-activity tests).

Unique job/folder keys per scenario (gg1111cc/f00bbbbb, gg2222cc/f00ccccc, gg3333cc/f00ddddd).

## Base branch

[...]

PR #1196, opened 2026-06-02
fix(uipath-troubleshoot): tolerate alpha-CLI lazy tool-group listing in manifest verifier

Stefan-Virgil:
## Problem

The `skill-troubleshoot-smoke-manifest-commands` smoke task runs `verify_manifest_commands.py` over **every** troubleshoot manifest as a `pre_run` gate (`fail_on_error: true`). The smoke CI installs **`@uipath/cli@alpha`**, and under that alpha the installed tool command-groups (`or`, `is`, `maestro`, `resource`, `docsai`, `traces`, `rpa`) are **not enumerated in the root `uip --help` Subcommands** — even though `uip tools install` reports success (`Successfully installed @uipath/orchestrator-tool@1.2.0-alpha…`) and `uip or …` is a real command at runtime.

The verifier's level-by-level path walk therefore flags every `or …` command BAD (`'or' is not a subcommand of 'uip (root)'` — ~172/205), the `fail_on_error` pre-run aborts the task → `ERROR`, and the whole troubleshoot smoke suite goes red (pass rate < 95%). This is **repo-wide** — it reds out every troubleshoot PR (#1149, #1150, #1191, …), unrelated to any individual scenario.

## Fix

When a path token is absent from its parent's `Subcommands`, before flagging it invalid (mirroring the existing `None`/`ConfigError` tolerance):

1. **Honor `KNOWN_ASPECT_ROUTERS`** — it was defined but never wired into `validate_match` (latent dead code), so `maestro`→{bpmn,case,flow} aspect-routers weren't actually handled.
2. **Probe the token's own `--help`** and accept it only if it renders as a **distinct group** (non-empty `Subcommands` that differ from the parent's). The distinctness check preserves the Windows `.CMD` false-positive guard the tree-walk exists for: a genuine typo falls back to the parent's help and echoes its `Subcommands` verbatim, so it's still rejected.

Single-file change to `verify_manifest_commands.py`; no manifest or scenario changes.

## Validation

- **No regression** against stable `uip` 1.0.0: `257/257` matches valid, exit 0 (the new branch stays dormant because stable lists `or` at root).
- **Deterministic alpha simulation** (root help without `or`; `uip or --help` distinct; unknown token → `.CMD` fallback to root): lazy-listed groups accepted, typo rejected, multi-level walk works.
- **Final proof = this PR's own CI:** the `smoke-manifest-commands` task runs against the alpha CLI, so it turning green here confirms the fix end-to-end. I could not install `@uipath/cli@alpha` locally (org SAML blocks the GitHub Packages token).

### DIFF
--- a/tests/tasks/uipath-troubleshoot/_shared/scripts/verify_manifest_commands.py
+++ b/tests/tasks/uipath-troubleshoot/_shared/scripts/verify_manifest_commands.py
@@ -225,21 +225,40 @@ def validate_shape(
         if result != "Success":
             return False, f"parent 'uip {' '.join(parent)}' returned {result}"
         subs = subcommand_names(help_payload)
-        if token in subs:
-            continue
-        # aspect-router fallback: some parents (e.g. `uip maestro`) host
-        # routed subcommands the JSON help payload doesn't enumerate.
-        if token in KNOWN_ASPECT_ROUTERS.get(parent, set()):
-            continue
-        # lazily-listed plugin tolerance: some plugins install correctly but
-        # aren't listed in the parent's Subcommands (e.g. `is`, `resource`,
-        # `docsai` in uip 1.2.x-alpha). Probe the token directly: if
-        # `uip <prefix> <token> --help` returns Success the subcommand is real —
-        # the parent just doesn't advertise it.
-        probe = fetch_help(uip_bin, tuple(path[: depth + 1]))
-        if probe is not None and probe.get("Result") == "Success":
-            continue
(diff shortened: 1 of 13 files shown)


dmorosanu:
Stacked 8a76af1b on this branch: migrates the 12 troubleshoot mock manifests from the retired 'resource' verbs to 'or' (resource group was dropped in cli/main alpha 1.196.0-alpha.20260609; renames registry already lists resource->or). The verifier fix alone left 8/72 shapes invalid - with both changes the smoke gate should go green. Local verify_manifest_commands.py: 72/72 valid, exit 0.

MarinRzv (approved):

PR #1212, opened 2026-06-02
fix(catalog): make catalog refresh build the full verb set (#1203 root cause)

uipreliga:
## Why
The nightly catalog refresh was silently producing a **31-verb** snapshot (vs ~1115), breaking every PR's CLI Verb Gate (#1203). Root-causing it surfaced three independent bugs in `build-uip-catalog.py` and the refresh workflow. This PR fixes all three; verified in CI to rebuild **1249 verbs** with `or`/`rpa` restored.

## The three fixes

1. **PascalCase tool-payload keys (the actual #1203 cause).** `uip tools search --output json` returns tool names under `Name` (PascalCase, cli PR #2266), but `install_all_tools` read `tool.get("name")` (lowercase) → `None` → every tool skipped → **0 installs** → only the 31 base verbs. Added a case-insensitive `_ci()` accessor and routed the tool-payload reads (search loop, the installed-set from `tools list`, and the `commandPrefix` broken-tool detection) through it.

2. **Alias-joined subcommand name.** `uip --help --output json` bakes aliases into a subcommand's `Name` with a pipe — orchestrator (the only aliased tool) rendered as `"or|orchestrator"`, so the whole orchestrator group (188 verbs) was dropped. `strip_args` now takes the canonical first token. Recovers `or` on the current alpha CLI **without** waiting for the upstream fix (UiPath/cli#2331), and is a no-op once that lands.

3. **Tool feed mismatch.** Tools were installed from public npm while the CLI comes from GitHub Packages `@alpha`; stable tools don't register reliably against the alpha CLI (e.g. `rpa` never loads). The Rebuild step now installs tools from the same `@alpha` feed as the CLI, so groups load and version-match.

## Guards retained
`--min-verbs 1000` / `--max-drop-frac 0.2` stay — the floor correctly **caught** the orchestrator drop during this investigation. A build that doesn't clear the floor aborts before opening a PR (proven in CI).

## Tests
- `_ci` casing; `install_all_tools` installs PascalCase-named tools / skips already-installed; `strip_args` alias stripping.
- Extended the key-casing contract guard to flag a lowercase regression of the tool-payload keys in `build-uip-catalog.py`.
- 36/36 in `tests/scripts/` pass.

## Verification
Dispatched `refresh-uip-catalog.yml` off this branch: **1249 verbs** (or=188, rpa present), guards passed, refresh PR opened automatically. (That verification PR is closed; a clean catalog-only refresh will be generated from main after merge.)

### DIFF
--- a/.github/workflows/refresh-uip-catalog.yml
+++ b/.github/workflows/refresh-uip-catalog.yml
@@ -66,20 +66,17 @@ jobs:
       - name: Rebuild catalog
         if: steps.check.outputs.exists != 'true'
         run: |
-          # Reset the @uipath scope to public npm before building. The alpha
-          # CLI is installed from GitHub Packages (step above), but the tool
-          # plugins (@uipath/*-tool) live only on public npm. Leaving the
-          # scope on GitHub Packages makes `uip tools install` resolve there,
-          # find nothing, install no tools, and collapse the catalog to the
-          # ~31 base verbs (#1203). build-uip-catalog.py's docstring documents
-          # this requirement.
-          npm config set @uipath:registry https://registry.npmjs.org/
+          # Install tools from GitHub Packages @alpha — the scope the install
+          # step already configured. Do NOT reset to public npm: the tool
+          # plugins must version-match the alpha CLI to register their full
+          # command tree. Stable (public-npm) tools install but fail to load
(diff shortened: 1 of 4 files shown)


bai-uipath (approved):

rockymadden (approved):
LGTM

PR #1226, opened 2026-06-03
docs(uipath-api-workflow): Generic connector activities + registry improvements

rares-baesu-uipath:
## What

Updates the `uipath-api-workflow` skill for the CLI's Generic connector-activity support **and** the registry improvements that followed from three live agent test sessions.

**Companion CLI PR: UiPath/cli#2341** — the skill describes behavior introduced there, so this should merge together with / after the CLI change. If the skill ships against an older CLI, agents will try `--object-name` on a `stub` that still rejects Generic activities.

## Changes

**Generic activity support**
- `references/connector-activity-discovery.md`: new *"Generic activities"* section — `--object-name` flow (with connector-key discovery tip), operation→verb/path resolution incl. `GETBYID`-style pseudo-keys, the hard IS-metadata requirement, operation-casing normalization, slot-vs-export key shapes (slot carries the operation; export bucket stays objectName-based to survive solution reconcile regeneration), path-parameter caveats with the `uip is resources run` cross-check. *Limits* entry rewritten: only triggers are rejected now.
- `references/cli-reference.md`: `--object-name` in stub usage/options; the stale "only Curated stubbed today" failure mode replaced with the actual errors (`MissingObject`, `MetadataUnavailable`, trigger rejection) and recovery commands.
- `SKILL.md`: decision-table row for generic record operations.

**Registry improvements (tracking the CLI commits)**
- resolve is server-side now (keyword sent to TypeCache; whitespace-tokenized matching documented; combined queries like `"github list all records"` narrow instead of failing); default `--limit` 50; resolve output examples corrected (PascalCased keys, new `Operation` field).
- stub output documents `Data.Parameters` / `Data.RequestFields` (required flags) and the new *"Required field(s) not provided via --inputs"* warning; the previously **mandatory** `uip is resources describe` cross-check is demoted to "value semantics / lookup hints only", and the troubleshooting entry's cause/detection updated to match.

All documented commands and outputs were verified live against a tenant; each doc revision was validated by a fresh agent test session, and the final state addresses every friction point those sessions reported.


skills/uipath-api-workflow/references/cli-reference.md
```diff
@@ -76,10 +76,10 @@ uip api-workflow registry resolve <keyword> [--limit <n>] --output json
 
 | Argument / Flag | Required | Description |
 |--|--|--|
-| `<keyword>` | yes | Substring matched against `displayName`, `connectorKey`, `objectName`, `fullName`. Case-insensitive. |
-| `-l, --limit <n>` | no | Max results (default: 20). |
+| `<keyword>` | yes | Whitespace-tokenized; every token must substring-match somewhere in `displayName`, `connectorKey`, `objectName`, `fullName`. Case-insensitive. Combined queries narrow: `"github list all records"` matches GitHub's "List All Records". |
```
liviubobocu: Here I tried with Claude list all records, I got this

- "github list records" ✅ and "list records github" ✅ — order-independent token matching works (the old contiguous-substring matcher would've failed both). This is the core PR improvement, confirmed live.
- "github list all records" ❌ zero — because this tenant's activity is named "List Records" (no "all"), and the token filter is strict AND (every token must match).

Image

It's worth checking all other references (I saw there are others as well, to remove the "all" keyword where appropriate, it is misleading the LLM

liviubobocu (approved):

tests/tasks/uipath-automation-discovery/e2e_full_discovery.yaml
```diff
@@ -8,8 +8,6 @@ tags: [uipath-automation-discovery, e2e, mode:build, lifecycle:generate]
 
 agent:
   type: claude-code
-  max_turns: 40
```
RaduAna-Maria, line 11: Why was this removed?

marius-bughiu (approved):

AlvinStanescu (approved):

RaduAna-Maria (approved):

PR #1231, opened 2026-06-03
fix(evals): skill-agents-batch-transform-coded-smoke-trigger — fix prompt

EusebiuJecan07:
## Summary

- Removes meta-question from prompt ("Tell me which skill applies") that caused the agent to answer from general knowledge instead of invoking the skill
- Removes invalid `expected_turns` field (no longer in schema)
- Adds `sandbox: {}` for schema compliance

## Test plan

- [ ] 1/1 local runs pass (score 1.0)

### DIFF
--- a/tests/tasks/uipath-agents/batch_transform_coded/smoke_trigger.yaml
+++ b/tests/tasks/uipath-agents/batch_transform_coded/smoke_trigger.yaml
@@ -5,9 +5,6 @@ description: >
   agent (LangGraph). Verifies the skill fires for the
   Python-coded-agent + CSV signal pair.
 tags: [uipath-agents, smoke, coded, lifecycle:discover, context-grounding]
-
-sandbox: {}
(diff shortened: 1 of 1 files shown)


tests/tasks/uipath-agents/batch_transform_coded/smoke_trigger.yaml
```diff
@@ -6,15 +6,16 @@ description: >
   Python-coded-agent + CSV signal pair.
 tags: [uipath-agents, smoke, coded, lifecycle:discover, context-grounding]
 
+sandbox: {}
```
radu-mocanu: is this needed?

EusebiuJecan07: Removed — not needed for CI, only surfaced a local schema validation difference.

PR #1237, opened 2026-06-03
fix(experiments): recursive post_run cleanup of node_modules, .npm-prefix, .venv

akshaylive:
## Summary

- Replaces `rm -rf node_modules .npm-prefix .venv` (current directory only) with a `find`-based recursive cleanup
- Searches up to 5 levels deep, matches only real directories (`-type d`), and prunes matched dirs to avoid double-visiting nested copies

## Test plan

- [ ] Verify post_run cleanup removes nested `node_modules` directories in sandbox runs

### DIFF
--- a/tests/experiments/default.yaml
+++ b/tests/experiments/default.yaml
@@ -28,7 +28,7 @@ defaults:
   # wire cleanup_solutions.py into their own post_run — that must run before
   # this step (still needs npm-installed CLIs).
   post_run:
-    - command: "rm -rf node_modules .npm-prefix"
+    - command: "find . -maxdepth 5 -type d \\( -name node_modules -o -name .npm-prefix -o -name .venv \\) -prune -exec rm -rf {} +"
       timeout: 30
 
 variants:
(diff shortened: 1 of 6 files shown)


tests/experiments/default.yaml
```diff
@@ -28,7 +28,7 @@ defaults:
   # wire cleanup_solutions.py into their own post_run — that must run before
   # this step (still needs npm-installed CLIs).
   post_run:
-    - command: "rm -rf node_modules .npm-prefix"
+    - command: "find . -maxdepth 5 -type d \\( -name node_modules -o -name .npm-prefix -o -name .venv \\) -prune -exec rm -rf {} +"
```
bai-uipath, line 31: the skills repo ones need this too, right? eg
https://github.com/UiPath/skills/blob/main/tests/experiments/default.yaml

bai-uipath (approved):

tests/experiments/default.yaml
bai-uipath: sorry I meant: https://github.com/UiPath/skills/blob/main/tests/experiments/nightly.yaml

akshaylive: You mean other experiments? Absolutely. Fixed it.

PR #1238, opened 2026-06-03
test(uipath-maestro-flow): capture raw flow-debug payload on output-assert failure

tmatup:
## Why

`skill-flow-calculator` flakes chronically at **0.375** (validate passes, the `flow debug` execution check fails). Across its history, byte-equivalent flow shapes both **pass and fail**, with no static-artifact, CLI-version, api-routing, or model differentiator — the run reports `finalStatus=Completed` yet the checker sees **`Outputs: []`** (empty `Variables.Globals`). That points at a **server-side Maestro debug-runtime non-determinism**, not the agent or the skill docs.

We can't confirm the mechanism, because the failure path discards the payload:

- On failure the checker prints only the flattened `collect_outputs()` (`Outputs: []`).
- `uip maestro flow debug` is an **ephemeral Studio Web robotdebug session** — it leaves no retained Orchestrator/Maestro instance, and the eval's `cleanup_solutions.py` deletes the solution. So the payload is **unrecoverable after the run**.

(Full investigation: the eval runs as `coder-eval-bot` with a StudioPro personal robot; debug runs server-side; instances are not persisted. A local repro is blocked — the tenant has 0 attended-class licenses, so a personal workspace can't be enabled for an interactive account.)

## What

A **failure-path-only** diagnostic capture in `_shared/flow_check.py`:

- `run_debug` stashes the raw `flow debug` stdout in a module buffer.
- `assert_output_value` / `assert_outputs_contain` / `assert_output_int_in_range` emit a `FLOW_DEBUG_RAW_CAPTURE` block to **stderr** before failing — a structured **SUMMARY** (`finalStatus`, `globals`, `globalVariables`, `elementOutputs`, per-node `elementExecutions`, `incidents`) **plus the full raw payload**.
- Stderr is captured into `task.json.success_criteria_results[].details`, so the **next eval flake** (run as `coder-eval-bot`, which has a working robot) preserves the exact payload — letting us localize the defect to *output-surfacing* vs *node-execution* vs *input-binding*, and confirm whether `finalStatus` is genuinely `Completed`.

Best-effort and **silent on success** — passing runs are unchanged, and the dump never raises so it can't mask the real assertion failure.

## Tests

- 3 new regression tests: flake shape (Completed + empty globals) dumps the capture; a normal payload stays silent; no-op when no debug ran.
- Full suite green: `pytest tests/tasks/uipath-maestro-flow/` → **98 passed**.

## Notes

- This is **diagnostic instrumentation**, not a fix. It does not change pass/fail behavior; it makes the chronic flake diagnosable. The durable fix (Maestro flow-debug runtime bug for empty `Variables.Globals` on a Completed trivial flow, and/or retry/replicate tolerance in the execution checker) follows once a captured payload confirms the mechanism.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
@@ -42,6 +42,17 @@ import sys
 from typing import Any, Iterable, Sequence
 
 
+# Raw stdout of the most recent ``uip maestro flow debug`` invocation, stashed by
+# :func:`run_debug` so the output-assertion helpers can dump the FULL runtime
+# response to stderr when they fail. This is the diagnostic channel for the
+# chronic "debug Completes but Variables/Globals come back empty" flake
+# (e.g. skill-flow-calculator 0.375): the captured stderr lands in
+# ``task.json.success_criteria_results[].details``, so a failing eval preserves
+# the exact payload (finalStatus, elementExecutions, globals, incidents) that the
+# checker saw — which is otherwise ephemeral and unrecoverable post-run.
+_LAST_DEBUG_RAW: str | None = None
+
+
 # ── Public helpers ──────────────────────────────────────────────────────────
 
 
@@ -66,6 +77,8 @@ def run_debug(
     for var_id, local_path in (attachments or {}).items():
         cmd.extend(["--attachment", f"{var_id}={local_path}"])
     r = subprocess.run(cmd, capture_output=True, text=True, timeout=timeout)
+    global _LAST_DEBUG_RAW
+    _LAST_DEBUG_RAW = r.stdout
     if r.returncode != 0:
         _fail(f"flow debug exit {r.returncode}\nstdout: {r.stdout}\nstderr: {r.stderr}")
     data = _parse_json(r.stdout)
@@ -239,7 +252,7 @@ def assert_outputs_contain(
     ok = len(missing) == 0 if require_all else len(present) > 0
     if not ok:
         mode = "all of" if require_all else "any of"
-        _fail(
+        _fail_with_capture(
             f"Outputs missing {mode} {list(needles)}; present={present}; "
             f"missing={missing}\nOutputs: {haystack[:1000]}"
         )
@@ -252,7 +265,7 @@ def assert_output_int_in_range(payload: dict, lo: int, hi: int) -> int:
     haystack = _stringify(collect_outputs(payload))
     hits = [int(m) for m in re.findall(r"-?\d+", haystack) if lo <= int(m) <= hi]
     if not hits:
-        _fail(
+        _fail_with_capture(
             f"No integer in [{lo}, {hi}] found in outputs\nOutputs: {haystack[:1000]}"
         )
     return hits[0]
@@ -272,7 +285,9 @@ def assert_output_value(payload: dict, expected: Any) -> None:
         if isinstance(expected, str) and isinstance(v, str):
             if expected.lower() in v.lower():
                 return
-    _fail(f"No output equals expected {expected!r}\nOutputs: {_stringify(outs)[:1000]}")
+    _fail_with_capture(
+        f"No output equals expected {expected!r}\nOutputs: {_stringify(outs)[:1000]}"
+    )
 
 
 def read_flow_input_vars(project_dir: str) -> list[str]:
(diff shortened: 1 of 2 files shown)


tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -414,5 +429,57 @@ def _stringify(values: Iterable[Any]) -> str:
...
+    Best-effort and side-effect-free: never raises, so it cannot mask the real
+    assertion failure that follows.
+    """
+    raw = _LAST_DEBUG_RAW
+    if not raw:
+        return
+    tag = f" ({context})" if context else ""
+    lines = [f"=== FLOW_DEBUG_RAW_CAPTURE BEGIN{tag} ==="]
+    try:
+        data = _get_ci(json.loads(raw), "Data") or {}
```
tmatup: `run_debug` already accepts debug stdout with leading non-JSON lines via `_parse_json(r.stdout)`, but this new summary path reparses the same buffer with plain `json.loads(raw)`. If the CLI emits a warning/banner before the JSON, the check still gets a valid payload and then an output assertion can fail, but the diagnostic block loses the structured `finalStatus`/globals/executions summary and records only `SUMMARY: <unparsable>`. Please parse through `_parse_json(raw)` here and only fall back to the raw-only path when that returns `None`.

tmatup:
## Quality review

Found 1 issue:

- Medium: the raw debug capture summary reparses stdout with `json.loads(raw)` instead of the helper's tolerant `_parse_json(raw)`, so a successful debug payload with leading CLI noise loses the new structured summary. Posted inline on `tests/tasks/uipath-maestro-flow/_shared/flow_check.py`.

Verification:

- `python3 -m py_compile tests/tasks/uipath-maestro-flow/_shared/flow_check.py tests/tasks/uipath-maestro-flow/_shared/test_flow_check.py` passed.
- `git diff --check origin/main...HEAD` passed.
- Manual Python repro confirmed `_parse_json` accepts preamble + JSON while the new capture summary prints `SUMMARY: <unparsable>` for the same raw buffer.
- GitHub `maestro-flow checker unit tests` is green. Local pytest could not run because this environment lacks pytest, pip, and venv/ensurepip. `Run skill smoke tests` was still in progress at the time of review.

Generated with [Codex](https://openai.com/codex)

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
tmatup: Fixed in 37368c1a. `_dump_debug_capture` now parses through the tolerant `_parse_json(raw)` (the same helper `run_debug` uses) and falls back to the `<unparsable>` path only when it returns `None`. A payload preceded by a CLI banner/warning now still yields the structured `finalStatus`/globals/elementExecutions summary. Added `test_capture_summary_survives_cli_preamble` (preamble + JSON) to lock it.

tmatup:
### Review Response

**1. Capture summary reparses stdout with `json.loads` instead of the tolerant `_parse_json`** — `tests/tasks/uipath-maestro-flow/_shared/flow_check.py`

**Fixed** in 37368c1a. `_dump_debug_capture` now parses through `_parse_json(raw)` and falls back to the `<unparsable>` path only when it returns `None`. A debug payload preceded by a CLI banner/warning now still produces the structured `finalStatus`/globals/`elementExecutions` summary instead of `SUMMARY: <unparsable>` — which is exactly the value this capture exists to preserve. Added a regression test (`test_capture_summary_survives_cli_preamble`) feeding CLI preamble before the JSON.

Local verification: `pytest tests/tasks/uipath-maestro-flow/ tests/scripts/test_runtime_payload_key_casing.py` → **108 passed** (incl. the new test and the key-casing contract guard).

rockymadden (approved):

tmatup:
### CI triage

**Only red check:** `Run skill smoke tests` (Smoke Skill Tests) — a pre-existing flaky gate, **not caused by this PR**.

- The gate enforces a 95% pass rate over ~18 nondeterministic agent-driven smoke tasks, which effectively requires 18/18.
- Two consecutive runs on this branch flaked with a **shifting failure set** — the signature of flakiness, not a deterministic regression:
  - run 1 → 83.3% (15/18): `hitl-smoke-completed-port`, `ixp-activation/explicit` (MAX_TURNS), `ixp-activation-listing/ixp-models`
  - run 2 → 88.9% (16/18): `hitl-smoke-completed-port`, `ixp-activation-listing/ixp-models`  (`ixp-activation` recovered)
- The **same gate is concurrently red on unrelated PRs** (`test/flow-scheduled-trigger-smoke`, `test/flow-merge-parallel-sync-smoke`, plus several `cancelled`/timeout runs) → repo-wide infra instability. `ixp-activation-listing/ixp-models` is a known chronic flake.

**Why it can't be this PR:** the diff is now only `_shared/flow_check.py` + `test_flow_check.py`; the capture is **failure-path-only** stderr (it cannot flip any assertion's verdict), and it touches no skill — so it has no mechanism to affect the hitl/ixp agent smoke tasks.

**All checks that exercise this change are green:** `maestro-flow checker unit tests`, `runtime-payload key-casing contract guard`, `catalog build integrity guards`, CLI Verb Gate, Skill Activation Gate, RPA smoke. Local: maestro-flow suite + casing guard = 108 passed.

Not re-running the smoke gate further (flaky twice in a row). Merging past the advisory smoke flake is a maintainer call; the two sticky tasks (`hitl-smoke-completed-port`, `ixp-activation-listing`) are pre-existing and best tracked/fixed separately.

PR #1239, opened 2026-06-03
docs(uipath-maestro-flow): require triggerNodeId on parent flow inputs forwarded to subflows

tmatup:
## What

Documents that a **parent flow's** `in` variable must carry `triggerNodeId` when its value is forwarded into a subflow — the same rule already documented for subflow-internal `in` variables (Rule #3), one scope up.

## Why

`skill-flow-subflow` is a chronically flaky e2e test. In the failing runs the agent emits a flow that `uip maestro flow validate` accepts as **Valid** and that passes the static structure check, but **faults at `flow debug`** with `[300501] Cannot read property 'split' of null`.

Root cause: the parent `text` input global is missing `triggerNodeId: "start"`, so the debug-injected input never reaches `$vars.start.output.text`; the subflow receives `null` and the reversal script crashes. A byte-diff against a passing run shows that single missing field is the only difference between green and red.

The subflow doc heavily emphasizes `triggerNodeId` for the *subflow's* inputs, but its parent-node example used literal inputs (`a: 2, b: 3`) and never showed a parent threading an external input through its trigger — so the agent applies the rule to the subflow and omits it on the parent.

## Changes

- New section **"Passing a Flow Input Into the Subflow"** — parent `variables.globals` with `triggerNodeId: "start"`, the trigger + subflow-node wiring, and the end-to-end value-flow chain.
- New **Rule #10** stating the parent `in`-var requirement (and that `flow validate` does not catch its absence).
- New **Debug-table row** mapping the `null.split` symptom to the missing parent `triggerNodeId`.

Companion CLI change (surfaces this as a `flow validate` warning): UiPath/cli `fix/flow-validate-trigger-binding`.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/references/plugins/subflow/impl.md
+++ b/skills/uipath-maestro-flow/references/author/references/plugins/subflow/impl.md
@@ -182,6 +182,56 @@ Subflow contents are stored in a top-level `subflows` object keyed by the parent
 }
 ```
 
+## Passing a Flow Input Into the Subflow
+
+When the **parent flow** itself takes an input the caller supplies at trigger time (e.g. a
+`text` to reverse) and forwards it into the subflow, the parent's `in` variable needs the
+**same `triggerNodeId` treatment as a subflow `in` variable** (rule #3) — set it to the
+parent's trigger node ID. Omitting it is a silent trap: `uip maestro flow validate` still
+reports **Valid**, but at `flow debug` the trigger output is empty, the value arrives as
+`null`, and the subflow script faults (e.g. `Cannot read property 'split' of null`).
+
+**Parent flow `variables.globals`** — note `triggerNodeId` on the `in` variable:
+```json
+{
+  "variables": {
+    "globals": [
+      { "id": "text", "direction": "in", "type": "string", "defaultValue": "", "triggerNodeId": "start" },
+      { "id": "reversedText", "direction": "out", "type": "string", "defaultValue": "" }
+    ]
+  }
+}
+```
+
+**Parent trigger + subflow node** — the subflow node reads the trigger output and maps it to
+the subflow's `in` variable:
+```json
+{
+  "nodes": [
+    {
+      "id": "start",
+      "type": "core.trigger.manual",
+      "typeVersion": "1.0",
+      "inputs": { "entryPointId": "<uuid>", "isDefaultEntryPoint": true },
+      "outputs": { "output": { "type": "object", "source": "=result.response", "var": "output" } }
+    },
+    {
+      "id": "reverseSubflow",
+      "type": "core.subflow",
+      "typeVersion": "<DEFINITION_VERSION>",
+      "inputs": { "text": "=js:$vars.start.output.text" }
+    }
+  ]
+}
+```
+
+Value flow: caller input `text` → (bound to `start` via the parent global's `triggerNodeId`)
+→ `$vars.start.output.text` → subflow node input `text` → the subflow's own `in` variable
+(with its **own** `triggerNodeId: "<subflowStart>"`) → `$vars.<subflowStart>.output.text`
+inside the subflow. The parent `in` var and the subflow `in` var are **independent variables
+in separate scopes**, joined only by the subflow node's `inputs` mapping; each needs its own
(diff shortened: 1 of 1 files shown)


tmatup:
### Review Response

Thanks for the review. All findings addressed in b3ecce8a.

**1. Prose density** — `subflow/impl.md` section intro (lines 187–192)

**Fixed** in b3ecce8a. Tightened per `.claude/rules/token-optimization.md` (dropped articles/filler/subject-padding); kept the silent-trap warning and the validate→debug causal chain intact.

**2. Rule #10 line length** — `subflow/impl.md:246`

**Fixed** in b3ecce8a. Shortened the rule; folded the validate wording change from #4 into it.

**3. Pre-existing duplicate `outputs` key** — `subflow/impl.md` `script1` node (Subflow Definition example)

**Fixed** in b3ecce8a, despite being pre-existing — it's in the same file and a real copy-paste hazard (duplicate JSON key → last wins → the error port silently dropped). Removed the duplicate; the single `outputs` now carries both `output` and `error` (Rule #6). I kept the schema-free variant to match the error-output shape used elsewhere in this doc and in the Parent Node JSON; the JSON block now parses with `outputs` keys `["error","output"]`.

**4. Domain accuracy — "validate does not catch" vs the companion CLI fix**

**Fixed** in b3ecce8a. Reworded Rule #10 to "`uip maestro flow validate` still reports **Valid** (it does not fail on the omission)". The companion CLI change (UiPath/cli#2349) adds a **warning**, not an error — `Status` stays `Valid` and exit code stays `0` — so this wording stays accurate whether or not that warning has shipped. The section intro already said "still reports Valid", so no drift there.

**5. What's missing** — nothing identified.

Acknowledged — doc-only change to an existing skill; no frontmatter/test/CODEOWNERS impact.

tmatup:
### CI triage — `Run skill smoke tests` (red)

**Not caused by this PR.** This PR changes one doc file (`subflow/impl.md`); the smoke job runs a *fixed* 18-task agentic suite independent of the diff.

**Root cause:** the ≥95% pass-rate gate over 18 tasks effectively requires 18/18 (17/18 = 94.4% trips it), so any single nondeterministic agentic failure turns it red. This run scored **15/18 (83.3%)**, failing on three **unrelated** skills:

- `skill-flow-hitl-smoke-completed-port`
- `skill-flow-ixp-activation-listing/ixp-models`
- `skill-flow-ixp-smoke-scaffold-multinode`

None is the subflow skill; a doc change cannot affect their runtime smoke runs.

**Pre-existing flakiness (evidence):** the same `Smoke Skill Tests` workflow flips success↔failure across unrelated branches within the same hour — e.g. `feat/or-eval-run-offline-evals` shows both a failure (16:48) and a success (17:00); `debug/flow-debug-payload-capture` and `test/flow-scheduled-trigger-smoke` also red. Conclusion correlates with neither branch nor diff.

**Action:** re-ran the failed job once (flaky-retry). Did not touch the unrelated HITL/IXP skills (out of scope — their owners) or weaken the gate. All other checks are green (CLI Verb Gate, Skill Activation Gate, Socket Security, advisory schema check).

If the re-run doesn't reach 18/18, this red is safe to merge through for a doc-only change — the failing tasks are unrelated and flaky.

tmatup:
### Re-run result — still red, confirms flakiness

Re-ran the smoke job once: **88.9% (16/18)** — still under the 95% gate, but the failure set **shifted**, which proves nondeterminism rather than a defect:

| | Failing tasks |
|---|---|
| Run 1 (15/18) | `hitl-smoke-completed-port`, `ixp-activation-listing/ixp-models`, `ixp-smoke-scaffold-multinode` |
| Run 2 (16/18) | `ixp-activation-listing/ixp-models`, `ixp-activation/explicit` (MAX_TURNS) |

Only `ixp-activation-listing/ixp-models` persists across both — a known chronic flaky-smoke offender. All failures are unrelated IXP/HITL agentic skills; none is touched by this doc-only PR.

Not re-running again (flaky two runs in a row → infra instability, not a PR defect). **Safe to merge through** for a docs change. The durable fix — the brittle 95%-over-18 threshold and the repeat-offender smoke tasks — belongs to a separate PR by those skills' owners.

rockymadden (approved):
Nice catch, thank you!

PR #1245, opened 2026-06-03
test(uipath-maestro-flow): de-flake HITL completed-port + IxP activation, drop flaky IxP listing smoke task

bai-uipath:
De-flakes the recurring `uipath-maestro-flow` IxP/HITL smoke failures that intermittently red the smoke gate across unrelated PRs. Root-caused from CI history (9 recent flow-touching runs), not a single failure. Test-only — no skill content changes.

## HITL completed-port — `hitl/smoke_02_completed_port_wired.yaml`

The check asserted the quoted literal `"completed"` port handle, but the handle serializes by node version: v1.0 emits `"sourcePort": "completed"`, v1.1 emits `"outcome-completed"`. Version selection became non-deterministic after #1088 removed hardcoded `typeVersion`s, so the quoted match failed whenever the agent picked v1.1. Fix: match the bare substring `completed`, which holds for both serializations; `uip maestro flow validate` still catches malformed edges. Confirmed green on this branch in CI — the failures seen on other branches are branches predating this (unmerged) fix.

## IxP activation turn budget — `ixp/activation.yaml`

`explicit` (and occasionally `invoice-extraction`) exhausted the 40-turn budget in CI; the cap, not the skill, was the failure — these build-intent rows legitimately use 45–80 turns. Bumped `max_turns` 40 → 80. Confirmed green on this branch post-bump.

## IxP listing — `ixp/activation_listing.yaml` (removed)

Deleted. This task asserted that a read-only "which IxP node types can I add — list them" question triggers the maestro-flow skill, and it can't pass deterministically. Skill activation is non-deterministic for read-only Q&A: the agent is free to answer directly via the CLI without invoking the Skill tool, so the `skill_triggered` hook is a coin-flip (9/9 fail in CI; observed flipping pass/fail across back-to-back local repeats). And dropping that hook isn't enough — the dominant CI score is 0.33 (one of three criteria passing), so a *behavioral* criterion is failing too, not just the hook. A behavioral rewrite (grade the registry-search command + no-build instead of the hook) would therefore still red, so there is no salvage.

The coverage belongs in the activation suite: `activation.yaml`'s build-intent rows (`explicit`, `invoice-extraction`) already verify the skill engages for IxP flow work, and they pass deterministically in CI. Build intent reliably triggers the skill and exercises the registry-search + node-add as concrete steps; a read-only listing question does neither reliably. A standalone listing variant is redundant and unfixable, so it's removed rather than carried as a permanent flake.

## Not addressed here

`hitl/smoke_01_hitl_node_placed.yaml` (≈2/9 fail) is a true positive — the agent occasionally single-nests the project, violating the solution-first layout rule. Left as-is; masking it would hide a real skill-adherence gap, tracked for a separate skill-side fix. `ixp/activation_negative.yaml` (1/9 ERROR) was a transient infra error, not a logic flake.


tests/tasks/uipath-maestro-flow/ixp/activation_listing.yaml
```diff
@@ -1,59 +0,0 @@
...
-
-initial_prompt: |
-  ${row.prompt}
-
-  Important:
-  - The `uip` CLI is already available.
-  - Use `--output json` on all uip commands.
-
-success_criteria:
-  - type: skill_triggered
```
joe-prosser, line 31: > Skill activation is non-deterministic for read-only Q&A: the agent is free to answer directly via the CLI without invoking the Skill tool

Does this mean that the agent uses the CLI to answer the question without activating the skill? If that's the case, should we just removed the `skill_triggered` criteria whilst keeping the `command_executed` below?

joe-prosser (commented):
Thanks for looking at this, agree we should fix the flakiness. Note @tmatup's fix [here](https://github.com/UiPath/skills/pull/1217) is in a similar spirit

tests/tasks/uipath-maestro-flow/ixp/activation_listing.yaml
bai-uipath: that wouldn't help - if the skill did not trigger then it wouldn't possibly know to execute the maestro flow commands. we should move activation to its own eval suite and not be in smoke tests (per slack discussion).

joe-prosser (approved):
As discussed

rockymadden (approved):

PR #1260, opened 2026-06-04
fix(agents): document example_calls semantics and pre-seed mocking eval fixtures

radu-mocanu:
## Summary
- `evaluate.md` now documents when `example_calls` are required: only for LLM-driven mocking; mockito-mocked functions keep bare `@mockable()` (mockito ignores `example_calls`)
- `eval_mocking_mockito` pre-seeds the project via `pre_run` fixtures instead of asking the agent to transcribe files from the prompt
- run limits tightened (expected 45 / max 55, down from 76/80)

## Why

the previous run on the dashboard failed exactly on bare `@mockable()` without `ExampleCall` -> the skill showed the pattern but never explained the strategy split, and the eval wasted most of its turns on file transcription.

passing-run claim: local run with these changes scores 1.0 in 42 turns / 296s / $1.03 (vs old skill remote run: 0.375, 65 turns, 932s, $1.95), both on sonnet 4.6.

### DIFF
--- a/skills/uipath-agents/references/coded/lifecycle/evaluate.md
+++ b/skills/uipath-agents/references/coded/lifecycle/evaluate.md
@@ -80,6 +80,15 @@ def fetch_weather(query: str) -> dict:
 
 During evaluations, calls matching an `ExampleCall.input` return the paired `output`. During normal execution, the real function runs.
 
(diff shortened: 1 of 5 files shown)


EusebiuJecan07 (approved):
Nice, good improvements!

PR #1283, opened 2026-06-05
feat(release): publish skills repo as @uipath/skills npm package

gabrielavaduva:
## What

Package the whole skills repo as the npm package **`@uipath/skills`**, versioned in lockstep with `@uipath/cli` (e.g. `1.196.x`) so a given CLI release resolves to a compatible skills package and the two never mismatch.

## Why

Today the CLI consumes skills by git-cloning this repo's `main` at install time (`contentStore.ts`), so any CLI build pulls whatever happens to be on `main`. Publishing a versioned package lets the CLI pin a compatible minor line.

## Changes

- **`package.json`** — defines `@uipath/skills`, `files` whitelist. `npm pack --dry-run` verified: ships full skill content (1428 files / 2.7 MB), excludes `tests/`, `.github/`, root `scripts/`, `.git`.
- **`scripts/sync-version.mjs`** — `package.json` is the single source of truth; derives `version-manifest.json` (`skillsVersion` + `targetCli`). `--check` mode is a CI guard.
- **`version-manifest.json`** — CLI↔skills pairing record.
- **`.github/workflows/publish.yml`** — alpha-on-push-to-`main` → GitHub Packages (`alpha` tag); stable-on-release → npmjs (`latest`); gated on the version-sync check. Mirrors the CLI's `ci.yml` + `publish-npm.yml`.
- **`.npmrc`**, **`docs/RELEASE.md`** — scope registry config + release model docs.

## Deliberately NOT in this PR

- **`.claude-plugin/plugin.json` / `marketplace.json` version bump** — these stay on their own `0.0.x` track (bumped by `daily-version-bump.yml`) for now. Unifying them with the npm package scheme is tracked in **[PILOT-5516](https://uipath.atlassian.net/browse/PILOT-5516)**.

## Before this can publish (follow-ups)

- [ ] Add repo secret **`NPM_TOKEN`** (npmjs automation token with publish rights to `@uipath`).
- [ ] Add repo secret **`GH_WRITE_TOKEN`** for GitHub Packages — or switch the alpha job to the built-in `GITHUB_TOKEN`.
- [ ] Confirm package name/scope `@uipath/skills` and seed version `1.196.0`.

## Out of scope (separate decision)

The CLI-side change to install the pinned `@uipath/skills` package instead of cloning `main` is **not** in this PR — that's the decision of whether npm replaces or complements the git-ref distribution model under discussion.

[PILOT-5516]: https://uipath.atlassian.net/browse/PILOT-5516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -36,8 +36,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version: "22.x"
-      # Fails if plugin.json / marketplace.json / version-manifest.json have
-      # drifted from package.json. Keeps the published tarball self-consistent.
+      # Fails if version-manifest.json has drifted from package.json.
       - run: node scripts/sync-version.mjs --check
 
   publish-alpha:
@@ -68,6 +67,8 @@ jobs:
           node scripts/sync-version.mjs
           echo "Publishing $ALPHA"
 
+      # setup-node wrote `@uipath:registry=https://npm.pkg.github.com/` + auth;
+      # the scoped registry routes this scoped package to GitHub Packages.
       - name: Publish
         run: npm publish --tag alpha
         env:
@@ -86,9 +87,11 @@ jobs:
           registry-url: "https://registry.npmjs.org"
           scope: "@uipath"
 
-      # publishConfig in package.json points at GitHub Packages for alpha;
-      # override it here so the stable release lands on npmjs.
+      # setup-node above wrote `@uipath:registry=https://registry.npmjs.org/`
+      # + auth into the runner .npmrc. `@uipath/skills` is scoped, so npm
+      # resolves the publish target from that scoped registry — a `--registry`
+      # flag would be ignored (it only sets the unscoped default).
       - name: Publish to npmjs
-        run: npm publish --registry https://registry.npmjs.org --access public --tag latest
+        run: npm publish --access public --tag latest
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -69,10 +69,12 @@ jobs:
 
       # setup-node wrote `@uipath:registry=https://npm.pkg.github.com/` + auth;
       # the scoped registry routes this scoped package to GitHub Packages.
+      # The built-in GITHUB_TOKEN suffices given the job's `packages: write`
+      # permission above — no dedicated secret needed.
       - name: Publish
         run: npm publish --tag alpha
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
+          NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
   publish-release:
     name: Publish stable to npmjs
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,16 +1,15 @@
 name: Publish skills package
 
-# Two-track publishing, mirroring the @uipath/cli model:
-#   - push to main          -> alpha prerelease to GitHub Packages (dist-tag `alpha`)
-#   - GitHub Release / manual -> stable release to npmjs (dist-tag `latest`)
+# Two-track publishing:
+#   - workflow_dispatch (target: github-alpha) -> alpha prerelease to GitHub
+#     Packages (dist-tag `alpha`). Manual — no auto-publish on push to main.
+#   - GitHub Release published -> stable release to npmjs (dist-tag `latest`).
 #
 # `npm install @uipath/skills` (no tag) always resolves to the last stable
 # release, never an alpha. The version line tracks the CLI minor line so the
 # CLI can pin a compatible skills package (see docs/RELEASE.md).
 
 on:
-  push:
-    branches: [main]
   release:
     types: [published]
   workflow_dispatch:
@@ -42,7 +41,7 @@ jobs:
   publish-alpha:
     name: Publish alpha to GitHub Packages
     needs: [guard]
-    if: github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'github-alpha')
+    if: github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'github-alpha'
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,94 @@
+name: Publish skills package
+
+# Two-track publishing, mirroring the @uipath/cli model:
+#   - push to main          -> alpha prerelease to GitHub Packages (dist-tag `alpha`)
+#   - GitHub Release / manual -> stable release to npmjs (dist-tag `latest`)
+#
+# `npm install @uipath/skills` (no tag) always resolves to the last stable
+# release, never an alpha. The version line tracks the CLI minor line so the
+# CLI can pin a compatible skills package (see docs/RELEASE.md).
+
+on:
+  push:
+    branches: [main]
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      target:
+        description: "Publish target"
+        required: true
+        default: npmjs
+        type: choice
+        options:
+          - npmjs
+          - github-alpha
+
+permissions:
+  contents: read
+
+jobs:
+  guard:
+    name: Validate version sync
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+      # Fails if plugin.json / marketplace.json / version-manifest.json have
+      # drifted from package.json. Keeps the published tarball self-consistent.
+      - run: node scripts/sync-version.mjs --check
+
+  publish-alpha:
+    name: Publish alpha to GitHub Packages
+    needs: [guard]
+    if: github.event_name == 'push' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'github-alpha')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+          registry-url: "https://npm.pkg.github.com"
+          scope: "@uipath"
+
+      # Stamp a prerelease suffix: <base>-alpha.<YYYYMMDD>.<run_number>.
+      # Dot-separated so semver orders the run number numerically. Never
+      # committed — only the published tarball carries it.
+      - name: Stamp alpha version
+        run: |
+          BASE=$(node -p "require('./package.json').version")
+          DAY=$(date -u +%Y%m%d)
+          ALPHA="${BASE}-alpha.${DAY}.${{ github.run_number }}"
+          npm version "$ALPHA" --no-git-tag-version --allow-same-version
+          node scripts/sync-version.mjs
+          echo "Publishing $ALPHA"
+
+      - name: Publish
+        run: npm publish --tag alpha
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
+
+  publish-release:
+    name: Publish stable to npmjs
+    needs: [guard]
+    if: github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'npmjs')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+          registry-url: "https://registry.npmjs.org"
+          scope: "@uipath"
+
+      # publishConfig in package.json points at GitHub Packages for alpha;
+      # override it here so the stable release lands on npmjs.
+      - name: Publish to npmjs
+        run: npm publish --registry https://registry.npmjs.org --access public --tag latest
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -0,0 +1,47 @@
...
+    "claude-code",
+    "agent",
+    "ui-automation",
+    "ui-testing",
+    "coded-agents",
+    "troubleshoot"
+  ],
+  "private": false,
+  "publishConfig": {
+    "registry": "https://npm.pkg.github.com"
```
DragosUnguru: If you have this registry defined here in `package.json`'s `publishConfig` you might not need the additional `.npmrc` ([source](https://docs.npmjs.com/cli/v11/configuring-npm/npmrc)).

Let's remove the `.npmrc`.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+          registry-url: "https://registry.npmjs.org"
+          scope: "@uipath"
+
+      # publishConfig in package.json points at GitHub Packages for alpha;
+      # override it here so the stable release lands on npmjs.
+      - name: Publish to npmjs
+        run: npm publish --registry https://registry.npmjs.org --access public --tag latest
```
DragosUnguru: This won't publish to `npm` registry, it will still publish to whatever you have defined in `package.json` or `.npmrc` (in our case, still the GitHub feed).

Counterintuitively, this `--registry` flag does **not** take precedence over scoped registries (as this flag specifies unscoped registries).

---

You could achive this precedence by defining scopes in `.npmrc`:
```txt
@myscope=https://my.custom.registry.url
```

, and using the `--scope` flag:
```sh
npm publish --scope=@myscope
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+          registry-url: "https://registry.npmjs.org"
+          scope: "@uipath"
+
+      # publishConfig in package.json points at GitHub Packages for alpha;
+      # override it here so the stable release lands on npmjs.
```
DragosUnguru: > \# override it here so the stable release lands on npmjs.

It won't. See [thread below](https://github.com/UiPath/skills/pull/1283/changes#r3361379428).

.npmrc
```diff
@@ -0,0 +1 @@
+@uipath:registry=https://npm.pkg.github.com
```
DragosUnguru: This is a public repository, yet we set the main registry to a private? npm feed.

This means that anyone `git clone`ing the repo, when running `npm install` will be met with a failure.

Let's ditch the `.npmrc` file entirely as it nevertheless is low in the precedence list (after `--scope` and after what's mentioned in `package.json`).

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
...
+    name: Validate version sync
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22.x"
+      # Fails if plugin.json / marketplace.json / version-manifest.json have
+      # drifted from package.json. Keeps the published tarball self-consistent.
+      - run: node scripts/sync-version.mjs --check
```
DragosUnguru: The script only ever reads/compares version-manifest.json. It explicitly, by its own docstring, does not touch `plugin.json/marketplace.json`. So the guard cannot catch the drift the comment promises.

docs/RELEASE.md
```diff
@@ -0,0 +1,48 @@
...
+`npm install @uipath/skills` (no tag) always resolves to the last stable npmjs release — alphas live only under the `alpha` tag on GitHub Packages. This mirrors `@uipath/cli`'s `ci.yml` (alpha on merge) + `publish-npm.yml` (stable on release).
+
+## Cutting a stable release
+
+1. Bump `package.json` to the target version (match the CLI minor line), run `npm run version:sync`, merge.
+2. Create a GitHub Release tagged `v<version>` → `publish.yml` publishes to npmjs.
+
+## Required secrets / setup (TODO before first publish)
+
+- [ ] **`NPM_TOKEN`** — npmjs automation token with publish rights to the `@uipath` scope (for stable releases).
```
DragosUnguru: The \@uipath scope is too broad. Apply the principle of least priviledge and use a token that's scoped **only** to \@uipath/skills, not the entire \@uipath org.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
...
+          npm version "$ALPHA" --no-git-tag-version --allow-same-version
+          node scripts/sync-version.mjs
+          echo "Publishing $ALPHA"
+
+      - name: Publish
+        run: npm publish --tag alpha
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
+
+  publish-release:
```
DragosUnguru, line 78: We should run some validation gate before publishing to the public `npm` feed. E.g. `validate-skills.yml`, `smoke-skills.yml`, etc.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
+name: Publish skills package
```
DragosUnguru, line 1: Every step is human-driven. There is no automation linking the CLI's version to this repo. This will drift. If lockstep is the goal, it should be machine-enforced (e.g., a scheduled/triggered job that reads the CLI's published version and opens the bump-package.json-version PR, plus a CI check that fails if they diverge).

These should be automated as part of the CI, not a runbook to run manually.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,94 @@
...
+#   - push to main          -> alpha prerelease to GitHub Packages (dist-tag `alpha`)
+#   - GitHub Release / manual -> stable release to npmjs (dist-tag `latest`)
+#
+# `npm install @uipath/skills` (no tag) always resolves to the last stable
+# release, never an alpha. The version line tracks the CLI minor line so the
+# CLI can pin a compatible skills package (see docs/RELEASE.md).
+
+on:
+  push:
+    branches: [main]
```
DragosUnguru: Should we filter the trigger for `src/skills/*` only?

Otherwise this PR itself will trigger a release, even though nothing that's packaged changed.

package.json
```diff
@@ -0,0 +1,47 @@
+{
```
DragosUnguru, line 1: All skill files now ship as one unit (like a monorepo).

There's no exports map, no subpath access, no per-skill packages (`@uipath/skill-rpa`, etc.).
E.g., a consumer who wants one skill pulls the full ~3 MB package.

The git-clone model it replaces at least allowed sparse/partial fetch, so this is a granularity regression.

We should think about per-skill consumption and packaging. And I think we should do it now, because splitting later is a breaking change and we'll have to "deprecate" the released `@uipath/skills` package from this CI.

DragosUnguru: Further, this is a "content" package with no module entry. No `main`/`module`/`exports`/`type`. Consumers can only read files out of `node_modules/@uipath/skills/...` by path, they can't `import` it.

As long as this is an `npm` package, maybe we should be able to `import`/`require` individual skills in js/ts mediums?

package.json
```diff
@@ -0,0 +1,47 @@
...
+  "name": "@uipath/skills",
+  "version": "1.196.0",
```
DragosUnguru, line 3: Why `1.196.0`? Shouldn't we sync with the CLI version (which currently sits at `1.1.x`?

DragosUnguru (changes requested):

package.json
gabrielavaduva: Done — removed both `publishConfig.registry` and the committed `.npmrc`. Each publish job now routes via the scoped `@uipath:registry` that setup-node writes per job (GitHub Packages for alpha, npmjs for release). Fixed in 28c8da0b.

.github/workflows/publish.yml
gabrielavaduva: Correct, thanks — the `--registry` flag only sets the unscoped default and is ignored for a scoped package. Dropped it; the npmjs job now routes via the scoped registry from setup-node. Fixed in 28c8da0b.

.github/workflows/publish.yml
gabrielavaduva: Removed that misleading comment. The release job relies on the per-job scoped registry, not a flag. Fixed in 28c8da0b.

.npmrc
gabrielavaduva: Agreed — removed `.npmrc` entirely so a public clone + `npm install` no longer hits the private feed. Fixed in 28c8da0b.

.github/workflows/publish.yml
gabrielavaduva: Right — the script only checks `version-manifest.json` now, so I corrected the comment to say exactly that. plugin.json/marketplace.json alignment is deferred and tracked in PILOT-5516. Fixed in 28c8da0b.

docs/RELEASE.md
gabrielavaduva: Updated `docs/RELEASE.md`: NPM_TOKEN is now specified as a granular token scoped to `@uipath/skills` only (least privilege), not the whole `@uipath` org. Fixed in a5c82d6c.

.github/workflows/publish.yml
gabrielavaduva: Agreed — nothing should reach the public feed without the quality gates. Tracked as follow-up: PILOT-5517 (validation gate before publish).

.github/workflows/publish.yml
gabrielavaduva: Agreed — a manual lockstep will drift. Tracked: PILOT-5518 (confirm seed version + machine-enforce CLI↔skills sync via a triggered bump PR and a divergence CI check).

.github/workflows/publish.yml
gabrielavaduva: Good catch — as-is this PR would self-trigger an alpha. Tracked: PILOT-5517 (path-filter the trigger to the paths actually shipped in the tarball).

package.json
gabrielavaduva: Strong point on the granularity regression vs git-clone, and on splitting being a breaking change later. Tracked for a design decision before wide adoption: PILOT-5519 (per-skill packages / subpath consumption + module exports).

package.json
gabrielavaduva: The repo cli/packages/cli/package.json reads 1.196.0, but the published line may differ as you note. Confirming the canonical line before first publish — tracked in PILOT-5518.

gabrielavaduva:
## Review follow-up

**Fixed in this PR** (`28c8da0b`, `a5c82d6c`) — threads resolved:
- Scoped-registry routing instead of `--registry`; removed `publishConfig.registry` and `.npmrc` (also fixes public-clone `npm install`).
- Corrected the stale version-sync guard comment.
- `publish-alpha` now uses the built-in `GITHUB_TOKEN` (no `GH_WRITE_TOKEN` secret needed).
- `node_modules/` + `package-lock.json` added to `.gitignore`.
- `NPM_TOKEN` documented as a granular token scoped to `@uipath/skills` only.

**Tracked as follow-ups** (threads left open):
| Concern | Issue |
|---------|-------|
| Validation gate + path-filtered trigger before publish | [PILOT-5517](https://uipath.atlassian.net/browse/PILOT-5517) |
| Confirm seed version (`1.196.0` vs published CLI line) + machine-enforce CLI↔skills lockstep | [PILOT-5518](https://uipath.atlassian.net/browse/PILOT-5518) |
| Per-skill packaging / subpath consumption / module `exports` (granularity) | [PILOT-5519](https://uipath.atlassian.net/browse/PILOT-5519) |
| plugin.json / marketplace.json version unification (pre-existing) | [PILOT-5516](https://uipath.atlassian.net/browse/PILOT-5516) |

Thanks for the thorough review @DragosUnguru.


[PILOT-5517]: https://uipath.atlassian.net/browse/PILOT-5517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PILOT-5518]: https://uipath.atlassian.net/browse/PILOT-5518?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PILOT-5519]: https://uipath.atlassian.net/browse/PILOT-5519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PILOT-5516]: https://uipath.atlassian.net/browse/PILOT-5516?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

package.json
gabrielavaduva: Confirmed: the seed is `1.196.0` — that is the canonical CLI line. Updated docs/RELEASE.md to mark it confirmed; PILOT-5518 now covers only the ongoing lockstep automation.

.github/workflows/publish.yml
gabrielavaduva: Resolved by removing the `push: main` trigger entirely (76ccfd26) — alpha is now manual-only via workflow_dispatch, so this PR no longer self-triggers and a path filter is no longer needed. PILOT-5517 is narrowed to just the pre-publish validation gate.

DragosUnguru (approved):
Some were deferred to subsequent Jiras.

Just so we don't lose track of those, we should also put them in an in-line comment or something here.

Otherwise approved.

PR #1290, opened 2026-06-05
test(uipath-maestro-flow): de-flake IxP activation — skill-activation eval + thresholded plugin-routing [RE-12420]

joe-prosser:
Jira: RE-12420.

## Why

The IxP **smoke** tasks gated *probabilistic* agent behaviour (which skill/plugin fires, which node lands, which discovery command runs) on one or two must-pass rows, so a single miss red-lined the blocking smoke gate on unrelated PRs. The fix is to split coverage by the distinct things being tested, and stop hard-gating any of it on the smoke run.

## What changed

**1. Skill-level activation → nightly activation eval**
At skill-selection time the agent doesn't know flow plugins exist, so "what IxP models can I use in maestro" is a `uipath-maestro-flow` *skill* positive. Added 10 IxP-phrased rows (listing + document-extraction build intents) to `tests/tasks/activation/uipath-maestro-flow.jsonl`.

**2. Plugin-level routing → non-blocking integration**
- Demoted the three IxP routing tasks `smoke` → `integration`.
- Renamed `activation*.yaml` → `routing*.yaml` and dropped the `activation` tag — they no longer assert skill activation (the `skill_triggered` criterion is gone); "activation" is reserved for the skill-selection eval above. They test plugin/node routing within the skill.
- Fan each prompt across many phrasings via `dataset.rows`, gated on aggregate `suite_thresholds.mean >= 0.85` instead of per-row pass/fail. `routing_listing.yaml` (10 rows), `routing.yaml` (5), `routing_negative.yaml` (8).

**3. IxP scaffold tasks → non-blocking integration**
- Demoted the two IxP scaffold smoke tasks `smoke` → `integration` and renamed `smoke_01_scaffold_minimal.yaml` → `scaffold_minimal.yaml`, `smoke_02_scaffold_multinode.yaml` → `scaffold_multinode.yaml` (task_ids drop `smoke`). These were the tasks actually red-lining the smoke gate, via the same probabilistic registry-search assertion.
- Added the required `mode:build` tag; descriptions reworded "Smoke" → "Integration".

**4. Fixed the registry-search assertion regex (all IxP tasks)**
The old case-sensitive `(search|list|get)\b[^\n]*ixp` false-negated *correct* behaviour — e.g. `registry search "IxP"` (casing), a bare `registry list` (list-all), or a topical `search "extraction"`/`"document"` that never put the literal `ixp` on the line. Replaced with a case-insensitive pattern that accepts a bare `registry list` and requires only a topical stem (`ixp|document|extract`) on a `search`/`list`/`get`. Applied across `routing*.yaml` and `scaffold_*.yaml`. Listing `command_executed` mean: **0.80 → 1.00**.

**5. Documented the regex inline**
A single plain-language breakdown of the pattern lives in `routing.yaml`; the other IxP tasks carry a one-line pointer to it (no duplicated explanation). Addresses review feedback that the regex was hard to read.

## Validation

Run locally on `claude-sonnet-4-6` (the nightly CI model) — pure `command_executed`/`run_command` checks, no LLM judge:

| Task | Gate | Means |
|------|------|-------|
| `routing_listing` | PASS | command_executed 1.00 · no-build 1.00 |
| `routing` | PASS | command_executed 1.00 · node-landed 1.00 |
| `routing_negative` | PASS | no-IxP-node 1.00 |

Skill-activation rows separately verified: `recall.yes 1.0` (10/10 fire `uipath-maestro-flow`). The `scaffold_*` tasks now run as non-blocking `integration` (no longer in the smoke gate).

### DIFF
--- a/tests/tasks/uipath-maestro-flow/ixp/routing.yaml
+++ b/tests/tasks/uipath-maestro-flow/ixp/routing.yaml
@@ -56,6 +56,13 @@ initial_prompt: |
   - Exploration + an attempted node-add is enough — you do not need to build a working flow.
 
 success_criteria:
+  # What this checks: did the agent search the registry for IxP nodes?
+  # Passes (case-insensitive) when the agent ran either:
+  #   • `registry list`                   — list all nodes, or
+  #   • `registry search|list|get <term>` — <term> contains ixp / document / extract
+  #                                         (prefix match, so extract → extraction)
+  # routing_listing.yaml drops `get` (read-only). Keep in sync with
+  # scaffold_minimal.yaml, scaffold_multinode.yaml, routing_listing.yaml.
   - type: command_executed
     description: "Agent searched the registry for IxP node types"
     tool_name: "Bash"
--- a/tests/tasks/uipath-maestro-flow/ixp/routing.yaml
+++ b/tests/tasks/uipath-maestro-flow/ixp/routing.yaml
@@ -1,18 +1,18 @@
-task_id: skill-flow-ixp-activation
+task_id: skill-flow-ixp-routing
 description: >
-  IxP activation — 2 positive activation tests (explicit, invoice-extraction)
-  fanned out via `dataset.rows`. Each row carries a distinct user prompt; success
-  criteria are identical across all rows.
+  IxP routing (positive) — document-extraction build prompts fanned out via
+  `dataset.rows`. Each row carries a distinct user prompt; success criteria are
+  identical across all rows.
 
-  Scope: did the skill get called, not did it work. Assertions verify the agent
-  invoked the uipath-maestro-flow Skill and ran the right CLI-command patterns
-  (registry search for uipath.ixp.*, node-add against uipath.ixp.* or
-  core.logic.mock) — NOT whether the resulting flow is correct or would execute.
+  Scope: did the agent route to the IxP plugin, not did the flow work.
+  Assertions verify the agent ran the right CLI-command patterns (registry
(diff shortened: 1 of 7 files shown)


tests/tasks/uipath-maestro-flow/ixp/routing.yaml
```diff
@@ -40,20 +56,12 @@ initial_prompt: |
...
-    expected_skill: "uipath-maestro-flow"
-    description: "Agent invoked the uipath-maestro-flow Skill"
-    weight: 2.0
-    pass_threshold: 1.0
-
   - type: command_executed
     description: "Agent searched the registry for IxP node types"
     tool_name: "Bash"
-    command_pattern: 'uip\s+maestro\s+flow\s+registry\s+(search|list|get)\b[^\n]*ixp'
+    command_pattern: '(?i)uip\s+maestro\s+flow\s+registry\s+(list\b|(search|list|get)\b[^\n]*\b(ixp|document|extract))'
```
richsilv, line 69: Is there a suitable place to describe this regex, like the description? Otherwise it might be fine, but it's very hard to tell.

richsilv (approved):
Looks great, one piece of feedback on the regex

tests/tasks/uipath-maestro-flow/ixp/routing.yaml
joe-prosser: Have add a comment

rockymadden (approved):

PR #1295, opened 2026-06-05
fix(uipath-test): pin report_generation eval to CLAIM:23 test set

ganeshborle:
## Summary

Pins the `uipath-test` skill integration test `tests/tasks/uipath-test/report_generation.yaml` to a specific, populated test set and relaxes the structural check so a fully passing/failing set still satisfies it.

- **Pin to CLAIM:23** — the prompt now targets the **CLAIM** project and uses test set **CLAIM:23** specifically, instead of "any test set with data".
- **Relax breakdown check** — the success-criteria Python check now requires a `passed`/`failed` breakdown (drops the `none` token).
- **Doc alignment** — updates the task `description` and the success-criteria label to match; strips trailing whitespace; fixes a `bod`→`body` typo in the regression-section check.

## Trade-off

This makes the eval tenant-specific (requires a CLAIM project with test set CLAIM:23) — a deliberate choice for deterministic data over tenant-agnostic discovery.

## Validation

- YAML parses (`yaml.safe_load`).
- Embedded success-criteria Python compiles.
- No trailing whitespace; description/label/check are internally consistent; no SHOPFLOW leftovers.

## Jira

[TMHUB-32038](https://uipath.atlassian.net/browse/TMHUB-32038)

[TMHUB-32038]: https://uipath.atlassian.net/browse/TMHUB-32038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/tasks/uipath-test/report_generation.yaml
+++ b/tests/tasks/uipath-test/report_generation.yaml
@@ -7,14 +7,14 @@ description: >
   report to `test-report-qa-<TODAY>.md` (the skill's default filename for
   today's date), and (d) the report body contains the QA-persona structural
   sections taught by references/test-result-report-guide.md (regression
-  analysis + the full pass/fail/none breakdown) and references the target
-  test set. The prompt does not name a specific test set so the test runs
-  against any tenant; the agent picks one with data. The skill — not the
-  prompt — must teach the filename convention, the persona-specific
-  sections, and the `--output json` flag. Note: when the chosen test set
-  has no executions, the skill correctly stops before running
-  `list-testcaselogs` (Critical Rule #4), so that command is not asserted
-  here; the report is expected to render the empty-data sections honestly.
+  analysis + the full pass/fail breakdown) and references the target
+  test set. The prompt targets the CLAIM project's test set CLAIM:23.
+  The skill — not the prompt — must teach the filename
+  convention, the persona-specific sections, and the `--output json` flag.
+  Note: when the test set has no executions, the skill correctly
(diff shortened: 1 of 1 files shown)


tests/tasks/uipath-test/report_generation.yaml
```diff
@@ -24,10 +24,10 @@ run_limits:
 initial_prompt: |
   Before starting, load the uipath-test skill and follow its workflow.
 
-  Generate a **QA Engineer** test report from one of my UiPath Test Manager
-  test sets. Pick a test set that actually has data so the report is
-  meaningful, and save the report file in the current working directory
-  using the default filename convention from the skill.
+  Generate a **QA Engineer** test report from my UiPath Test Manager "SHOPFLOW"
+  test sets. Pick a test set that actually has data so the report is meaningful,
```
vaishalisharma-uipath: Do we not have a test set that has some meaningful executions, instead of asking the coding guest to guess and chose a model. I believe in a real scenario, a test set would be mentioned.

vaishalisharma-uipath (approved):
Please address the comment.  Approving to unblock.

tests/tasks/uipath-test/report_generation.yaml
ganeshborle: Yes, makes sense. As the turns are getting optimized, the prompts also needs to be concise and should imitate user behaviour.

PR #1305, opened 2026-06-08
fix(uipath-maestro-flow): correct $vars check in ixp invoice greenfield e2e [RE-12430]

joe-prosser:
The `$vars` downstream-consumption check was double-quoted, so bash turned `\$` into a leading-`$` regex anchor and it never matched — failing the task on every run regardless of agent output. Switch to a fixed-string match (`-F`), matching the `["$vars."]` convention used by the sibling ixp smoke tasks.

Ran `skill-flow-ixp-e2e-invoice-extraction-greenfield` locally on `claude-sonnet-4-6` (CI model) and it passed: 8/8 criteria, score 1.000 (was 7/8, 0.906 before the fix).

Resolves RE-12430.


joe-prosser:
Updated the PR description to add the passing-run claim (ran `skill-flow-ixp-e2e-invoice-extraction-greenfield` locally on `claude-sonnet-4-6`: 8/8, score 1.000).

scallaway-uipath (approved):

PR #1319, opened 2026-06-08
test(uipath-test): coder_eval coverage for requirement, customfield, objectlabel commands [TMHUB-31915]

ganeshborle:
## Summary

Adds 6 coder_eval tasks (3 smoke, 3 integration) covering `uip tm` command groups that previously had no skill-eval coverage: **requirement**, **customfield**, and **objectlabel**.

Each task asserts auth-first ordering (`uip login status`) and `--output json` on every call, and validates the **complete command** (project key, exact names/labels, exact ids, case-sensitive `--object-type`/`--data-type`) rather than a fragment. Data is **deterministic** — tasks reference specific named fixtures in the live `codereval` org or self-seed their own — so runs don't depend on incidental tenant state.

## Tasks

| Task file | Type | Commands covered |
|---|---|---|
| `requirement_crud_smoke.yaml` | smoke | requirement create / get / update / delete (self-seeded) |
| `requirement_list_export_smoke.yaml` | smoke | requirement create + objectlabel add + requirement list (--labels) / export |
| `requirement_traceability_integration.yaml` | integration | requirement create / testcases / list-testcase-ids / list-by-test-execution |
| `customfield_definition_crud_smoke.yaml` | smoke | customfield create / list / update (--rename-to) / delete |
| `customfield_population_integration.yaml` | integration | customfield create + value create/update + label add/list |
| `objectlabel_lifecycle_integration.yaml` | integration | objectlabel add / list (--filter) / get / remove |

## Deterministic fixtures (project: HEALTH)

- Test cases **HEALTH:11 / 12 / 13** ("PHI Masking …") named explicitly instead of "pick the first N".
- Execution **"Healthcare Compliance Suite"** (`ffb763a0-8ca3-0400-c9fc-0b49a4ac9c42`) for the execution-back traceability lookup.
- Requirement tasks self-seed fixed-name rows (+ a fixed `eval-export` label) and clean up.
- All six use **HEALTH** (no other task touches it; clean custom-field slate). No task uses CLAIM.

## Jira (TMHUB, S196)

| Jira | Group |
|---|---|
| TMHUB-32058 | requirement commands |
| TMHUB-32061 | customfield commands |
| TMHUB-32063 | objectlabel commands |

## Validation

- [x] All 6 YAML files parse; every `command_pattern` regex compiles (Python `re`)
- [x] Positive/negative match checks: patterns fire on the intended complete command and reject wrong variants (e.g. `requirement list` won't match `list-testcase-ids`; lowercase `text`/`requirement` rejected; wrong execution id rejected; `objectlabel add` won't satisfy the `remove` criterion)
- [x] Discovered automatically by the runner's `tasks/**/*.yaml` glob (no manifest change needed)
- [x] Branch rebased on latest `main`

> Note: `requirement`/`objectlabel` are validated via transcript regex. They're registered in the CLI at v1.196.0 and ship in the `@latest` the smoke runner installs; the locally-pinned `uip` is older and lacks them.

https://uipath.atlassian.net/browse/TMHUB-31915

### DIFF
--- a/tests/tasks/uipath-test/customfield_definition_crud_smoke.yaml
+++ b/tests/tasks/uipath-test/customfield_definition_crud_smoke.yaml
@@ -1,0 +1,63 @@
+task_id: skill-test-customfield-definition-crud-smoke
+description: >
+  Smoke test: agent uses the uipath-test skill to manage a custom field
+  *definition* through its lifecycle — create a Text field scoped to
+  Requirement, list to confirm, rename it, then delete it. The task is
+  self-contained and deterministic: it creates its own field with a fixed
+  name in the HEALTH project (which has no custom fields) and removes it
+  again. Every assertion validates the complete command — the
+  case-sensitive `--object-type Requirement` and `--data-type Text`, the
+  exact field name, `--rename-to`, and `--output json` — not a fragment.
+tags: [uipath-test, smoke, mode:operate, lifecycle:manage, feature:customfield]
+
+run_limits:
+  expected_turns: 12
+  max_turns: 35
+
+initial_prompt: |
+  I'm in the HEALTH project in Test Manager. Can you add a custom field on
+  requirements called "Eval Reviewer" — just free text, to note who
+  reviewed a requirement. Show me the project's custom fields so I can
+  check it's there. Actually, rename it to "Eval Reviewer Team". And once
+  we're done, remove it so we don't leave anything behind.
+
+success_criteria:
+  - type: command_executed
+    description: "Agent checked login status (`uip login status --output json`)"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+login\s+status.*--output\s+json'
+    min_count: 1
+    weight: 1.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Complete create command: `uip tm customfield create --project-key HEALTH --name \"Eval Reviewer\" --data-type Text --object-type Requirement --output json`"
+    tool_name: "Bash"
+    command_pattern: '(?=.*--project-key\s+HEALTH)(?=.*--name\s+["\x27]?Eval Reviewer)(?=.*--data-type\s+Text)(?=.*--object-type\s+Requirement)(?=.*--output\s+json)uip\s+tm\s+customfield\s+create\b'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Complete list command: `uip tm customfield list --project-key HEALTH --output json`"
+    tool_name: "Bash"
+    command_pattern: '(?=.*--project-key\s+HEALTH)(?=.*--output\s+json)uip\s+tm\s+customfield\s+list\s'
+    min_count: 1
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Complete rename command: `uip tm customfield update --project-key HEALTH --rename-to \"Eval Reviewer Team\" --output json`"
+    tool_name: "Bash"
+    command_pattern: '(?=.*--project-key\s+HEALTH)(?=.*--rename-to\s+["\x27]?Eval Reviewer Team)(?=.*--output\s+json)uip\s+tm\s+customfield\s+update\b'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Complete delete command: `uip tm customfield delete --project-key HEALTH (--field-ids <uuid> | --name \"Eval Reviewer Team\" --object-type Requirement) --output json`"
+    tool_name: "Bash"
+    command_pattern: '(?=.*--project-key\s+HEALTH)(?=.*(--field-ids\s+\S+|--object-type\s+Requirement))(?=.*--output\s+json)uip\s+tm\s+customfield\s+delete\b'
+    min_count: 1
+    weight: 1.5
+    pass_threshold: 1.0
--- a/tests/tasks/uipath-test/customfield_population_integration.yaml
+++ b/tests/tasks/uipath-test/customfield_population_integration.yaml
@@ -1,0 +1,82 @@
+task_id: skill-test-customfield-population-integration
+description: >
+  Integration test: agent uses the uipath-test skill to populate custom
+  field *rows* on a specific, named test case after defining the fields.
+  The distinction the skill must teach: top-level `customfield` manages
+  definitions, while the nested `value` (Text) and `label` (Label)
+  subgroups fill in per-object rows. The prompt references the target by
+  its human-readable key and description (HEALTH:11 "PHI Masking - SSN in
+  Patient Search Results") and the agent resolves it to an id live. It
+  defines two TestCase-scoped fields, writes a Text value on HEALTH:11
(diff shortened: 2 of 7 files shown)


uipreliga:
## Code Review — coder_eval task YAMLs

Reviewed with multiple models (Gemini 3, GPT-5.3-codex, GPT-5.4) + an Opus sub-agent, cross-checked against `assets/uip-catalog-snapshot.json` and `skills/uipath-test/SKILL.md`. All asserted `uip tm` verb paths exist in the catalog ✅. Schema, tag dimensions, inherited sandbox/agent, and side-effect-based grading conform to repo rules ✅.

### High / notable

- **`requirement-*` tasks assert a command surface the skill doesn't document.** `--requirement-key`, `--requirement-id`, `--requirement-ids`, `--add-testcase-ids`, `--output-file` appear in **0 files** under `skills/uipath-test/` — SKILL.md only says `tm requirement` is "unchanged on both surfaces" and never documents its verbs/flags. The verbs exist in the catalog, but the flag spellings have no source of truth in the repo. Risk: if any spelling is wrong, the strict full-command regex **silently never matches** → the 3 requirement tasks fail for spec drift, not agent behavior. **Recommendation:** document the `tm requirement` group in the skill (so the tasks validate skill-taught behavior), or relax the identifier lookaheads to `--requirement-(key|id|ids?)` until verified against `uip tm requirement <verb> --help`.

### Medium

- **`customfield` delete-by-name branch is under-constrained** — `customfield_definition_crud_smoke.yaml` delete criterion. The alternation `(--field-ids\s+\S+ | --object-type\s+Requirement)` lets a command with only `--object-type Requirement` and **no `--name`** pass, which the CLI would reject (SKILL.md requires `--name` + `--object-type` for by-name delete). **All four reviewers flagged this** — strongest-consensus finding. Tighten so the name branch requires both `--name` and `--object-type`.
- **`customfield_population_integration.yaml` value criterion grades only 1 of 2 mutations.** Prompt step 3 says set the value "then update it" (two ops), but the `value (create|update)` criterion is `min_count: 1` — a single `create` passes; the update is never graded. Bump to `min_count: 2` or split into two criteria.
- **Integration tasks mutate shared fixtures without cleanup → second-run collisions.** `customfield_population_integration` creates two fixed-name field definitions ("Eval Reviewer", "Eval Risk") and never deletes them; a re-run hits "already exists" on `create`. `objectlabel_lifecycle_integration` deliberately leaves `eval-regression` on HEALTH:11/12/13. Both contradict the "self-contained / leave no test data behind" claim in their own descriptions. (The two smoke CRUD tasks *do* clean up correctly.) Add teardown steps, or document the residual fixtures.
- **`objectlabel` attach verifies only one label, not both.** The attach criterion's `--labels\s+["\x27]?eval-` is satisfied by a single `eval-`-prefixed label — an agent that supplies only `eval-regression` (omitting `eval-smoke`) passes, despite the description saying "naming both labels." Require both tokens explicitly.

### Low / informational

- **`--name` prefix leniency** — `--name\s+["\x27]?Eval Reviewer` (no closing boundary) also matches `"Eval Reviewer Team"`. Harmless here because the differing verb anchors (`create\b` vs `update\b`) prevent cross-criterion contamination, but the name assertion is weaker than the description implies.
- **Bare `--values` lookahead** in `customfield_population` label-add — `(?=.*--values)` matches even with no operand. Tighten to `--values\s+\S+`.
- **`lifecycle:manage` / `lifecycle:report` aren't in the README's closed taxonomy** (`discover`/`generate`/`setup`). However, these values already appear in the repo (18× `manage`, 4× `report`) and on existing `uipath-test` tasks — pre-existing README-vs-practice drift, **not introduced by this PR**. Worth a separate ticket to reconcile the README, not a blocker here.

### ❌ False positive (do not act on)

- One model flagged `requirement\s+list\s` as matching `requirement list-testcase-ids`. **This is wrong.** `\s` is `[ \t\n\r\f\v]`; the char after `list` in `list-testcase-ids` is `-`, not whitespace, so it cannot match. The other three reviewers confirm the trailing `\s` is the *correct guard that prevents* hyphenated-sibling over-match. Same reasoning clears `testcases?\s+list\s` and `customfield\s+list\s`.

### Positive observations

- Auth-first ordering (`uip login status --output json`) graded on every task.
- All criteria grade observable side effects — no self-report grading.
- Disciplined verb anchoring: trailing `\s` for `list` (blocks `list-*` siblings), `\b` for unambiguous verbs; escaped `\(v2\)` in the rename assertion; `--output-file` paired with an on-disk `run_command` check.
- The two requirement smoke CRUD tasks are genuinely self-seeding and self-cleaning.
- Weights/thresholds sane (1.0 supporting, 1.5–2.0 core, all `pass_threshold: 1.0`).

### Overall

Regex engineering is sound — the one "High" the external models disputed is a false positive, and verb anchoring is correct throughout. The real weaknesses are **grading completeness** (under-constrained delete/label/value criteria that let an incomplete agent pass) and **fixture hygiene** (two integration tasks don't clean up shared state, risking second-run flakiness). The single most defensible blocker is the undocumented `requirement` flag surface — worth resolving before these land, since a wrong flag spelling fails silently.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

tests/tasks/uipath-test/customfield_definition_crud_smoke.yaml
```diff
@@ -0,0 +1,66 @@
...
+  case-sensitive `--object-type Requirement` and `--data-type Text`, the
+  exact field name, `--rename-to`, and `--output json` — not a fragment.
+tags: [uipath-test, smoke, mode:operate, lifecycle:manage, feature:customfield]
+
+run_limits:
+  expected_turns: 12
+  max_turns: 35
+
+initial_prompt: |
+  Before starting, load the uipath-test skill and follow its workflow.
```
uipreliga: that's too detailed - noone will say it.

uipreliga: basically, make sure that your activation is >90% for smoke tests. this will result in activation of ~100%.

tests/tasks/uipath-test/customfield_population_integration.yaml
```diff
@@ -0,0 +1,93 @@
...
+  label rows back. Assertions validate the complete command shape with the
+  exact field names and the case-sensitive `--object-type TestCase`.
+tags: [uipath-test, integration, mode:operate, lifecycle:manage, feature:customfield]
+
+run_limits:
+  expected_turns: 18
+  max_turns: 45
+
+initial_prompt: |
+  Before starting, load the uipath-test skill and follow its workflow.
```
uipreliga: even worse for integration / e2e tests

uipreliga (approved):
Please fix all the issues identified than 🚢

tests/tasks/uipath-test/customfield_population_integration.yaml
ganeshborle: Make sense, changing the prompt like the real human input.

tests/tasks/uipath-test/customfield_definition_crud_smoke.yaml
ganeshborle: Understood and noted.

vaishalisharma-uipath (approved):

PR #1326, opened 2026-06-09
test: drop coded tag from rpa coded test case task

radu-mocanu:
## Summary
- remove the `coded` tag from the rpa coded-test-case task so `coded` is reserved for coded-agent tasks
- keeps coverage and pass-rate reporting for `coded` scoped to coded agents only


alessandromuresan (approved):

radu-mocanu:
`coded` label is used to mark coded rpa workflows. to identify coded agents runs `uipath-agents + coded` should be used

PR #1328, opened 2026-06-09
fix(ci): pin UTF-8 encoding in coder-eval verdict step

Mihaiii:
## Problem

The **Run Coder Eval** workflow failed on the Windows job at the **"Plaintext summary + verdict"** step ([failing run](https://github.com/UiPath/skills/actions/runs/27190532059/job/80269435850)):

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 30641: character maps to <undefined>
```

The eval tasks themselves ran to completion — the verdict step crashed while *reading* the results.

## Root cause

The step reads each `task.json` via `pathlib.Path.read_text()` with no `encoding` argument, so Python uses the OS locale default:

- **Linux** → UTF-8 → reads fine (this is why the Linux job passed).
- **Windows** → `cp1252` → crashes on the UTF-8 byte `0x90` present in agent transcripts.

Classic "works on Linux, fails on Windows" encoding mismatch.

## Fix

Pin `encoding="utf-8"` on the `read_text()` call. `task.json` is UTF-8 regardless of host locale, so this makes the read deterministic on both platforms. Applied to **both** the Linux and Windows verdict steps to keep them consistent.

```diff
- rows = [json.loads(p.read_text()) for p in sorted(...rglob('task.json'))]
+ rows = [json.loads(p.read_text(encoding='utf-8')) for p in sorted(...rglob('task.json'))]
```

### DIFF
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -415,11 +415,24 @@ jobs:
           BEDROCK_MODEL:            ${{ secrets.BEDROCK_MODEL }}
           ANTHROPIC_API_KEY:        ${{ secrets.ANTHROPIC_API_KEY }}
           TASK_GLOBS:               ${{ needs.partition.outputs.windows_globs }}
+          # Redirect the temp root off the default %TEMP%. On the GH-hosted
+          # Windows runner %TEMP% resolves to its 8.3 short form
+          # (C:\Users\RUNNER~1\AppData\Local\Temp). coder-eval creates each
+          # agent sandbox under %TEMP%, so the sandbox path inherits the
+          # `RUNNER~1` tilde segment — and Claude Code's permission layer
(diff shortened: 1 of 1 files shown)


bai-uipath (approved):
Approved as-is. No changes requested. The changes make sense and do not affect the Linux path.

## Notes

- **UTF-8 pin**: correct fix for the Windows cp1252 crash; no-op on Linux (already UTF-8).
- **Temp-root redirect**: fixes the `RUNNER~1` short-path issue; scoped to the Windows eval step only, can't leak to Linux.
- No flag or invocation changes anywhere.
- Validated: branch run failed before the fix, passed after.

Mihaiii:
@uipreliga Can you please have a look? It seems you're the only code owner and I can't merge only with @bai-uipath 's approval.

uipreliga (approved):

PR #1346, opened 2026-06-09
refactor: rename uipath-ixp → uipath-ixp-designtime

cezara98t:
## Summary
Renames the `uipath-ixp` skill to **`uipath-ixp-designtime`** to make room for a future, separate runtime-consumption skill. This PR is a **rename only**.


skills/uipath-ixp-runtime/SKILL.md
```diff
@@ -0,0 +1,28 @@
...
+description: "UiPath IXP runtime — consume published IXP (Document Understanding) models to extract data from documents at runtime. For creating projects, improving prompts, and publishing models→uipath-ixp-designtime."
+---
+
+# UiPath IXP Runtime
+
+Skill for **consuming published UiPath IXP models** — running a model that was published with the design-time workflow against new documents to extract structured data.
+
+> This skill is **in development**. The runtime consumption workflow and CLI surface are not yet documented here. To create a project, review/confirm predictions, improve prompts, and publish a model, use the [uipath-ixp-designtime](../uipath-ixp-designtime/SKILL.md) skill.
+
+## When to Use This Skill
```
joe-prosser: I've had a lot of problems with activation confusion between the existing ixp skill and the maestro flow skill, figure we'll also see that between maestro flow, design time and runtime now. I imagine it'll be especially bad between runtime and maestro flow as they're both about consuming the models. 


I would advocate for adding a "when not to use" redirect here, similar to: https://github.com/UiPath/skills/pull/1348 so we have somewhere to redirect common activation failures to the relevant skill (whether that's design time or maestro flow)

cezara98t: I added `disable-model-invocation: true` and `user-invocable: false` so hopefully this one won't interfere. This is only a skeleton for now, until we actually implement the skill.

cezara98t: actually removed it for now

alexandrujircan (approved):

joe-prosser (approved):
Looks good. Only concern was adding the "when not to use" redirect to maestro flow, but I see we've removed the runtime skill.md from this pr now

misupantea-uipath (approved):

skills/uipath-ixp-designtime/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
-name: uipath-ixp
-description: "UiPath IXP (Document Understanding) — review IXP predictions with Claude, confirm valid fields, improve prompts, publish models."
+name: uipath-ixp-designtime
+description: "UiPath IXP designtime (Document Understanding) — create projects, review/confirm predictions, improve prompts, and publish extraction models."
```
gabrielavaduva, line 3: why is it `ixp-designtime`? are you planning to add another skill? if not the previous name seems fitting, if yes you shouldn't, we try to keep one skill per use case 😁

misupantea-uipath: yes, we're planning on adding a separate ixp-runtime skill, that will be used to consume the IXP projects (that can be created with the ixp-designtime skill)

gabrielavaduva: let's keep everything in one skill, this way we avoid activation issues. if the split is necessary and nothing is common between the 2 you can add two folders in the `references` one

gabrielavaduva (changes requested):
let's not merge this, we need to keep it as one skill

PR #1385, opened 2026-06-10
docs(uipath-rpa): avoid hangs when using debug start

mihai-stan-uipath:
[UI-71574] naively running "uip rpa debug start" leads the agent to wait for it to finish without checking if it's paused after encountering an exception.

[UI-71574]: https://uipath.atlassian.net/browse/UI-71574?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/skills/uipath-rpa/references/debugging.md
+++ b/skills/uipath-rpa/references/debugging.md
@@ -4,6 +4,8 @@ The `uip rpa debug` group (plus `uip rpa run` and `uip rpa execution cancel`) pr
 
 This is a powerful complement to `validate` (static validation). While `validate` catches structural and type issues at design time, the debugger catches runtime problems: wrong API responses, null references, logic errors, failed deserialization, and more. Use both together for comprehensive workflow validation.
 
+On a breakpoint or unhandled exception `uip rpa debug` pauses and the CLI call stays alive until you `continue` / `continue-ignore` / `execution cancel` — a foreground call blocks to timeout and a backgrounded one never sends a completion notification. See [§ Common Debugging Workflows](#common-debugging-workflows) for the required background+poll helper.
+
 ## Studio Desktop vs headless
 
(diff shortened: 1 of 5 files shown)


skills/uipath-rpa/references/debugging.md
```diff
@@ -199,8 +217,8 @@ uip rpa focus-activity --activity-id "Assign_1"
 # 2. Toggle a breakpoint on the focused activity
 uip rpa debug toggle-breakpoint --output json
 
-# 3. Start debugging — execution pauses at the breakpoint
-uip rpa debug start --file-path "GetStockPrices.xaml" --output json
+# 3. Start debugging (background+poll helper; pauses at the breakpoint)
+bash debug-run.sh "GetStockPrices.xaml"
```
DragosUnguru, line 221: I'm not keen on adding this extra layer of scripting between the skills and the CLI.

If there's a gap in the debug commands we should address that directly in the rpa-tool command.

This introduces a layer of loose scripting that requires to be kept in sync (manually, that is) with the CLI tool command(s), all while being versioned with the skill instead of the CLI tool. This will certainly diverge and regressions will be caught at runtime.

Telling the model to copy the script to the expected location, run it in the background, poll for its stdout and so on brings a lot of friction and a lot of undeterministic mechanics that are prone to failures.

---

Please consider opening a PR and adapting the [rpa-tool command](https://github.com/UiPath/Studio/tree/develop/RpaTool/rpa-tool/src) directly

DragosUnguru (changes requested):
I'm not keen on introducing this additional layer of scripting between the skill and the CLI.

We should tackle these gaps through either one of the two existing ways.

PR #1388, opened 2026-06-10
fix(hitl): chained path in quality_08 prompt; skip quality_10 infra block

dushyant-uipath:
## Summary

- **quality_08**: Agent used valid 2-step JS destructuring (`const output = $vars.x.output; output.approved`) which doesn't produce the `.output.approved` substring the `file_contains` criteria require. Prompt now mandates the direct chained path so criteria match.
- **quality_10**: Nightly experiment overrides `turn_timeout` to 900s regardless of task-level setting (PATTERN 1 merge strategy). Flow consistently times out in the first turn with only a stub produced. Marking `skip: true` until the infra allows per-task override.
- **quality_04**: Already fixed by #1360 (merged). Criterion `'"completed"'` couldn't match v1.1 `"outcome-completed"` — now uses bare `'completed'`.

## Test plan
- [ ] quality_08 passes on next nightly run — agent now told to use `$vars.<nodeId>.output.approved` directly
- [ ] quality_10 is skipped and no longer appears as a failure
- [ ] quality_04 continues passing (criterion fix already on main)

### DIFF
--- a/tests/tasks/uipath-human-in-the-loop/quality_08_variable_binding_fieldid.yaml
+++ b/tests/tasks/uipath-human-in-the-loop/quality_08_variable_binding_fieldid.yaml
@@ -22,8 +22,9 @@ initial_prompt: |
        - Outcomes: Approve (primary), Reject
   4. Script node — reads the reviewer's decision and logs:
        "Approved: <approved value>, Comments: <comments value>"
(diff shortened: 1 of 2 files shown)


dushyant-uipath:
Closing — all other failures from the June 10 run have passed in today's run. e2e_01 turn timeout is the remaining issue and needs a separate fix.

PR #1393, opened 2026-06-10
test(maestro-flow): interactive AskUserQuestion solution-select task

akshaylive:
## What

Adds an **interactive-mode** test for `uipath-maestro-flow` that exercises the skill's `AskUserQuestion` solution-selection gate — the first coverage of the skill's interactive decision-surfacing behavior.

## How it works

- **Fixture:** two empty existing solutions (`SolarReports`, `TideTracker`) are staged into the sandbox cwd via `sandbox.template_sources`.
- **Trigger:** the skill's greenfield rule (`author/references/greenfield.md` — "Check the current directory for existing `.uipx` files") requires the agent to STOP and present a dropdown via `AskUserQuestion` (one option per discovered solution + "Create a new solution" + "Something else") instead of guessing.
- **No `initial_prompt`:** in simulation mode the simulator generates the opener from `persona` + `goal`.
- **Simulation block:** a developer persona answers the gate — chooses a brand-new solution named `WeatherAlert`, never reusing the staged two — keeping the build path deterministic (`WeatherAlert/WeatherAlert/WeatherAlert.flow`).
- **`AskUserQuestion` added to `allowed_tools`** (replace strategy — full default list restated).

## Grading

| Criterion | Weight | Checks |
|---|---|---|
| `command_executed` `tool_name: AskUserQuestion` | 3.0 | **the gate fired** (primary) |
| `command_executed` (negative, `min/max_count: 0`) | 1.5 | no `flow init` into a pre-existing solution |
| `solution init` / `flow init` / `flow validate` / `--output json` | 1.0–1.5 | build still completes |
| `json_check` on `Projects[?Type=='Flow']` + `file_exists` flow | 1.0–1.5 | flow registered & created |

## Verification

`coder-eval run` → **score 1.000, 8/8 criteria**, simulation ended on `stop_token` with `criteria_passed=True`.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/interactive/fixtures/existing_solutions/SolarReports/SolarReports.uipx
+++ b/tests/tasks/uipath-maestro-flow/interactive/fixtures/existing_solutions/SolarReports/SolarReports.uipx
@@ -1,0 +1,6 @@
+{
+  "DocVersion": "1.0.0",
+  "StudioMinVersion": "2025.10.0",
+  "SolutionId": "a1b2c3d4-1111-4aaa-8bbb-000000000001",
+  "Projects": []
+}
--- a/tests/tasks/uipath-maestro-flow/interactive/fixtures/existing_solutions/TideTracker/TideTracker.uipx
+++ b/tests/tasks/uipath-maestro-flow/interactive/fixtures/existing_solutions/TideTracker/TideTracker.uipx
@@ -1,0 +1,6 @@
+{
+  "DocVersion": "1.0.0",
+  "StudioMinVersion": "2025.10.0",
+  "SolutionId": "a1b2c3d4-2222-4aaa-8bbb-000000000002",
+  "Projects": []
+}
--- a/tests/tasks/uipath-maestro-flow/interactive/solution_select.yaml
+++ b/tests/tasks/uipath-maestro-flow/interactive/solution_select.yaml
@@ -1,0 +1,121 @@
+task_id: skill-flow-solution-select-ask
+description: >
+  Interactive-mode variant of init-validate. The working directory already
+  contains two existing solutions (SolarReports, TideTracker). When asked to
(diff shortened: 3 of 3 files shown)


bai-uipath (approved):
Approve — purely additive test (two fixtures + one task YAML), zero blast radius outside the eval suite, and it fills a real gap: the first coverage of the skill's `AskUserQuestion` decision-surfacing. The framework mechanics it relies on (contents-not-dir `template_dir` staging, `command_executed` on a non-Bash `tool_name`, simulation persona/goal, `json_check` + `gte`) all hold. Two things to address before this is trustworthy in the nightly suite.

## Issues (medium)

**Gate-firing depends on recursive `.uipx` discovery — likely flaky.** The fixtures stage solutions one level deep (`SolarReports/SolarReports.uipx`), but the greenfield rule only says "check the current directory." A non-recursive `ls *.uipx` / glob in cwd finds nothing → the skill's "no solutions found, create one automatically" branch → the gate never fires and the primary 3.0 criterion fails. The green run happened because the agent chose a recursive idiom, and it's a single trial (`simulation.n_trials` unset). Fix: set `n_trials` > 1 to measure the real pass rate, and/or flatten the fixture to a top-level `.uipx` so non-recursive discovery also finds it. As written this is the arrangement that's both less common in real use and the fragile one.

**Negative guard doesn't catch the case it describes.** `command_pattern: '(uip|\$UIP)\s+(maestro\s+)?flow\s+init\b.*(SolarReports|TideTracker)'` only matches if a solution name sits on the `flow init` line, but the realistic silent-adoption path is `cd SolarReports && uip maestro flow init <Proj>` — the name is in the `cd`, never on the init line. The `WeatherAlert/...` `json_check` and `file_exists` already enforce that the new solution was used, so this criterion contributes weight (1.5) for protection it doesn't actually give. Either drop it or also match `cd\s+(SolarReports|TideTracker)`.

PR #1423, opened 2026-06-11
test(uipath-agents): agent-agnostic asks + interactive coded-agent sim tests

cosmyo:
## Context

From Akshaya's [#team-coding-agents thread](https://uipath-product.slack.com/archives/C0A2T23NJ59/p1781108946675159), two linked asks:

1. **`AskUserQuestion` is Claude-specific** — skills should not name it directly in their **instructions/prose**; phrase interactive steps agent-agnostically so codex/other harnesses map to their own ask-tool. ("ask user question to …" — Claude maps to `AskUserQuestion`, others to their internal ask-tool.)
2. **"We are lacking interactive tests for most skills"** — use the `simulation:` mechanism (template [#1393](https://github.com/UiPath/skills/pull/1393)) instead of `initial_prompt:`, starting with a few high-traffic golden scenarios.

Both scoped to **uipath-agents** here.

### 1. Agent-agnostic asks (3 reference files)
Reworded prose `AskUserQuestion` references in the coded-agents docs to `ask the user` / `user question` / `interactively`. Pure doc change — no behavioral instruction lost.

**`SKILL.md` `allowed-tools` is intentionally left unchanged (keeps `AskUserQuestion`).** That field is Claude *pre-approval*, not a restrictive allowlist — per the [Claude Code docs](https://code.claude.com/docs/en/skills#pre-approve-tools-for-a-skill), it grants the tool without a per-call permission prompt and does **not** restrict the pool. Removing it would only force a permission prompt before each interactive ask under Claude Code (and is ignored by other providers), so it doesn't make anything more agent-agnostic. Akshaya's ask was about the prose references; the endorsed template #1393 and all 10 skills on `main` keep `AskUserQuestion` in `allowed-tools`.

### 2. Two interactive simulation tests (coded-agents path)
The skill's greenfield Delivery (quickstart Step 8) + Deploy (Step 9) stops are documented as **"always ask, even if the user only said build"** — the non-inferable interactive surface. Every existing coded test suppresses it with a "do NOT pause / do NOT ask" prompt. These two exercise it via simulated users:

| Task | Tier | Tenant | What it grades |
|---|---|---|---|
| `coded/deploy_interactive` | `e2e` | yes | Gate fired (`AskUserQuestion`, 3.0) **and** real `deploy --my-workspace` + `.nupkg` artifact. The golden "ask → deploy" path. |
| `coded/delivery_deploy_gates` | `integration` | no | Gate fired **and** load-bearing negative: no deploy/publish/push/upload after the user picks Skip. The "ask → skip, don't auto-publish" path. |

Both follow the #1393 template: no `initial_prompt`, a `simulation:` block (persona/goal/constraints), and `AskUserQuestion` added to the task's `allowed_tools` (replace strategy — required for the gate to fire under the Claude harness; grading on the harness-level tool name is correct and matches #1393).

## Lint

`/lint-task` on both: **2 OK, 0 issues**. CLI-verb check clean (only Info on the dynamic `--output json` pattern). YAML + check script validated.

## Verification

Not yet run via `coder-eval` (opened without a run). `delivery_deploy_gates` runs locally with no tenant; `deploy_interactive` needs live auth (nightly cadence). Happy to attach a passing-run claim before merge.


cosmyo:
Thanks — addressed the review feedback in `627e9a29`:

- **[Low] Awkward phrasing in `bindings-reference.md` (lines 92, 705)** — fixed, plus two more of the same pattern the mechanical rename left behind (`fall back to user question` → `fall back to asking the user`; `Call user question with` → `Ask the user, with`). Also smoothed `Use user question (or platform equivalent)` → `Ask the user interactively (or use the platform equivalent)` in `quickstart.md`. No behavioral change.
- **[Note] `max_turns: 8` tight** — bumped both simulation blocks to `12` for headroom so the simulator can't run dry before the Deploy gate on a clarifying follow-up.

On the other findings:

- **[Manual review] `allowed-tools` behavioral impact** — already resolved. The review ran before commit `4e3ae792`, which **reverted the `allowed-tools` removal**. `AskUserQuestion` stays in the frontmatter (it's Claude pre-approval, not a restrictive allowlist; removing it only adds a permission prompt before each ask and is ignored by other providers). `SKILL.md` now matches `main`; only the prose in the reference files was reworded. So this is a no-op, not a regression.
- **[Note] Tag vocabulary** — `lifecycle:deploy`, `lifecycle:execute`, and `feature:deploy` are pre-existing conventions (used by `deploy_my_workspace.yaml` + ~20 other tasks), reused here for consistency with sibling coded tests rather than invented. Flagging per the README "propose in PR" rule; happy to align if the team wants them folded into the closed vocab (or migrated to `setup`/`generate`).
- **[High, advisory] Passing-run claim** — still pending. `delivery_deploy_gates` is the no-tenant one to run; `deploy_interactive` needs live auth (nightly). Will attach the claim before merge.

PR #1429, opened 2026-06-11
fix(ci): scope smoke-skills trigger to changed task only for test-only PRs

akshaylive:
## Summary
- Skill source changes (`skills/<name>/...`) continue to run the full smoke suite for that skill
- Test-only changes (`tests/tasks/...`) now run only the affected task YAML instead of all smoke tasks for the skill
- Adds missing `mode:build` tag to the `add_node` maestro-flow task

## Test plan
- [ ] Modify a file under `skills/<name>/` and verify the full skill smoke suite triggers
- [ ] Modify a task YAML under `tests/tasks/<skill>/` and verify only that task runs
- [ ] Modify a fixture/check script (non-YAML) under `tests/tasks/<skill>/` and verify it resolves to the owning task YAML

### DIFF
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -92,27 +92,58 @@ jobs:
             exit 0
           fi
 
-          # Extract unique skill names from changed paths
-          SKILLS=$(echo "$CHANGED" | grep '^skills/' | sed 's|skills/\([^/]*\)/.*|\1|' | sort -u)
+          # Skill SOURCE changes → run that skill's FULL smoke suite. The
+          # skill's behavior may have changed, so broad coverage is warranted.
+          SRC_SKILLS=$(echo "$CHANGED" | grep '^skills/' | sed 's|skills/\([^/]*\)/.*|\1|' | sort -u | grep -v '^$' || true)
 
-          # Also include skills whose test tasks changed
-          TEST_SKILLS=$(echo "$CHANGED" | grep '^tests/tasks/' | sed 's|tests/tasks/\([^/]*\)/.*|\1|' | sort -u)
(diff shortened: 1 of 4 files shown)


bai-uipath (approved):
**Approve** — good CI-cost win (test-only PRs no longer re-run a skill's whole suite). Worth considering the coverage gap below before merge, but not a blocker.

## Worth addressing

- **Shared-fixture changes now trigger zero smoke tasks.** `resolve_task_yaml()` maps a changed path only to the nearest ancestor dir that *directly* contains a `*.yaml`. A shared helper above the leaf task dirs — e.g. `tests/tasks/uipath-maestro-flow/shared/check_flow_structure.py` — walks past the skill root to `tests/tasks`, where the `while [[ "$d" == tests/tasks/* ]]` loop exits, so it resolves to nothing and no tasks run. Previously this ran the whole skill suite. So edits to shared check logic — the change most likely to break many tasks at once — now run nothing. Fix: when a changed `tests/tasks/<skill>/` path resolves to no yaml, fall back to the whole-skill glob for `<skill>`.

## Consider as follow-up

- **The detect logic has outgrown inline YAML-bash.** It's ~200 lines with its own helper functions (`is_excluded_path`, `resolve_task_yaml`) and is untestable as embedded bash — the coverage gap above is exactly the kind of bug a unit test catches in three lines. Extract the decision logic ("given a git diff, which globs run") to a script the workflow calls — Python preferred given the repo, with a pytest over classification + fixture resolution — and keep the jobs/matrix/runner orchestration in YAML. Keep it in this repo, not `coder_eval` (the windows/benchmark/smoke policy is skills-repo-specific and shouldn't couple into the framework being decoupled).

PR #1440, opened 2026-06-12
feat(hooks): emit plugin-usage telemetry to App Insights on PostToolUse

RaduAna-Maria:
## What

Adds a `PostToolUse` hook (`hooks/send-telemetry.sh`) that emits one
`uip.skills.tool-use` Application Insights event per UiPath-attributable tool
call by piping a flat JSON object to the hidden `uip track` CLI command. Adds
`TELEMETRY.md` documenting the schema, privacy model, and opt-in gate. Wires the
hook into `hooks/hooks.json` as an async `PostToolUse` hook.

## Why

Understand how the plugin is used in the field — which skills fire, which `uip`
CLI verbs run, success/latency, and on which environment (alpha/staging/prod) —
without shipping file contents, full command lines, file paths, cwd, or
transcripts.

## How it works

1. **Opt-in / off by default.** Gated on `UIPATH_TELEMETRY_DISABLED=0` (reuses
   the `uip` CLI's own telemetry gate variable). Unset (default) or `1` → silent
   no-op.
2. **Relevance gate.** Fires on every tool call; emits only for plugin-
   attributable calls — `Skill` (`uipath:`/`uipath-`), `Bash`/`PowerShell`
   invoking `uip`/`rpa-tool`, `Agent` of a UiPath or Claude built-in type, and
   file tools on `.cs`/`.flow`/`.xaml`/`.uipx`/`.bpmn`/`agent.json`/
   `caseplan.json`/`project.json`/`app.config.json`/`action-schema.json`.
   Attribution comes from `tool_input` only.
3. **Region-scoped extraction.** A single string-aware `awk` pass walks the
   payload once and pulls each field only from its region (envelope /
   `tool_input` / `tool_response` / `effort`), so embedded customer content
   (prompts, stdout, file contents) can never false-match a field.
4. **Environment resolution.** Maps `uip login status` → `BaseUrl` to
   `alpha`/`staging`/`prod`/`other`/`unknown`, cached 1h in a per-user
   `chmod 700` dir, parsed as data (never sourced).
5. **Hands off to the CLI.** Pipes one flat JSON object to `uip track` on stdin
   in a detached subshell. The CLI owns transport, the App Insights connection,
   the event name (`uip.skills.tool-use`), the `source: "skills-plugin"`
   dimension, and stamps the authenticated cloud identity + CLI version.

## Privacy

**Not anonymous.** Events ride the CLI's telemetry tracker, so each is
associated with the signed-in UiPath identity (`CloudUserId` / `CloudTenantId` /
`CloudOrganizationId`, stamped by the CLI). The hook itself never sends file
contents, stdout/stderr, full command lines, file paths, `cwd`, or
`transcript_path` — only derived, low-cardinality fields (tool name, skill name,
derived `uip` subcommand verb, file extension, environment, outcome,
`durationMs`, etc.). Nothing is sent unless explicitly opted in with
`UIPATH_TELEMETRY_DISABLED=0`. Full schema in `TELEMETRY.md`.

## Non-blocking by contract

Async hook (`"async": true`) + always `exit 0` + detached subshell → never
delays or fails a tool call. Pure bash + grep/sed/awk, no `jq` dependency,
cross-platform (macOS / Linux / Windows Git Bash).

## Dependency

Requires `uip track` ([UiPath/cli#2600](https://github.com/UiPath/cli/pull/2600))
in the released CLI to actually emit; harmless no-op if the command is absent.

## Testing

Validated against the gate matrix with a stubbed `uip track`:

| Case | Result |
|------|--------|
| `Skill` uipath:uipath-platform | emit, skillName set ✓ |
| `Bash` echo hello | dropped ✓ |
| `Bash` uip solution publish | emit, uipSubcommand=`solution publish` ✓ |
| `Edit` README.md | dropped ✓ |
| `Edit` proc.flow | emit, fileExtension=`.flow` ✓ |
| `Bash` interrupted | emit, outcome=`interrupted` ✓ |

All emitted event bodies parse as valid JSON and gate correctly on
`UIPATH_TELEMETRY_DISABLED`.

### DIFF
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -8,15 +8,14 @@ stream (see [Correlation](#correlation)).
 
 ## Enabling / disabling
 
-Telemetry sends **only** when both are true:
-
 | Env var | Effect |
 |---------|--------|
-| `UIPATH_TELEMETRY_ENABLED` | Opt-in switch. Send only when set to exactly `1`. Unset or `0` → no send. |
+| `UIPATH_TELEMETRY_DISABLED` | Opt-out switch. Set to `1` or `true` to disable. Same variable and semantics as the `uip` CLI, so disabling CLI telemetry disables this hook too. |
 | `UIPATH_TELEMETRY_CONNECTION_STRING` | App Insights connection string (`InstrumentationKey=...;IngestionEndpoint=https://<region>.in.applicationinsights.azure.com/`). Without it there is nothing to send to. `APPLICATIONINSIGHTS_CONNECTION_STRING` is read as a fallback. |
 
-Default (neither set) is silent no-op. To turn off after enabling, set
-`UIPATH_TELEMETRY_ENABLED=0` or unset it.
+Telemetry is attempted by default, but **only transmits when a connection
+string is configured** — there is no built-in default endpoint. To turn it off
+explicitly, set `UIPATH_TELEMETRY_DISABLED=1` (or `true`).
 
 ## What triggers an event
 
--- a/TELEMETRY.md
+++ b/TELEMETRY.md
@@ -64,22 +64,24 @@ sends becomes an event property.
 |-------|---------|-------|
 | `toolName` | `Skill`, `Bash` | Claude Code tool |
 | `toolUseId` | `toolu_01ABC` | Unique per call — correlation key + ordering tiebreaker |
-| `sessionId` | `b3f1...` | Claude Code `session_id` — session correlation key |
+| `sessionId` | `b3f1...` | Claude Code `session_id` — the coding-agent session; session correlation key |
+| `userId` | `9eeb311fc6d5fa03` | **SHA-256 of `cwd`** (first 16 hex) — stable, anonymous per-workspace id. Distinct from the CLI-stamped `CloudUserId` |
 | `skillName` | `uipath:uipath-platform` | `Skill` calls only |
 | `uipSubcommand` | `solution publish` | Derived first 1–2 verbs of a `uip` command — never the full command line |
-| `fileExt` | `.flow` | File-tool calls only |
+| `fileExtension` | `.flow` | File-tool calls only |
 | `environment` | `alpha` / `staging` / `prod` / `other` / `unknown` | From `uip login status` `BaseUrl`, cached 1h |
 | `baseUrl` | `https://cloud.uipath.com` | Cloud base URL only |
 | `outcome` | `ok` / `failure` / `interrupted` | From `tool_response`, not output content |
 | `permissionMode` | `bypassPermissions` | |
-| `effortLevel` | `high` | When present in payload |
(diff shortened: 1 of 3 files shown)


gabrielavaduva:
## Review: telemetry hook

Reviewed the full diff. **The script itself is solid and secure** — the blocker is documentation integrity, not code.

### Security/quality assessment of `send-telemetry.sh` — 

The main risk in a payload-parsing hook is untrusted content (prompts, command lines, stdout, file contents) leaking out or breaking the emitted JSON. Handled well:

- **Injection-safe assembly.** `san()` strips every value to `[A-Za-z0-9:._/ -]` + 120-char cap before JSON assembly — no quotes/backslashes/pipes/control chars survive, nothing is `eval`'d, numerics validated digits-only.
- **No content over-attribution.** Region-scoped `awk` extraction means a `stdout` with `"success":false` or a prompt naming `uip solution publish` / `.flow"` can't false-match an envelope/`tool_input` field. Gate matches `tool_input` only, never output.
- **Cache parsed as data, never sourced**; per-user `chmod 700` dir.
- **Never-fail by contract** — `set +e`, detached `( … & )` subshell, always `exit 0`; piping to a missing `uip track` is harmless.
- Privacy posture reasonable: derived verbs not full commands, extension not path, no `cwd`/`transcript_path`/file contents.

### ⚠️ Blocker: the PR description contradicts what ships

The PR **body** and committed **TELEMETRY.md** describe two different systems, and the body is stale — on the most privacy-sensitive points:

| PR description claims | Actual (TELEMETRY.md + script) |
|---|---|
| Direct `curl` POST to App Insights | Pipes to `uip track`; no HTTP in the hook |
| Gated on `*_CONNECTION_STRING` | Gated on `UIPATH_TELEMETRY_DISABLED=0`; no connection string |
| `cwd` SHA-256 hashed; **anonymous** | **Not anonymous** — CLI attaches `CloudUserId`/`CloudTenantId`/`CloudOrganizationId`; `cwd` not collected |
| `.cs` intentionally **excluded** | `.cs` **included** in the gate |

The body's "Privacy: anonymous" claim is the opposite of what ships. Please rewrite the description to match TELEMETRY.md before merge — approving on "it's anonymous" would be approving something false.

### Non-blocking

1. **Opt-in coupling**: reusing the CLI's `UIPATH_TELEMETRY_DISABLED` means anyone who set `=0` for *CLI* telemetry now also ships *skills* telemetry. Intended? The inverted "set DISABLED=0 to enable" semantics are a footgun — confirm it's documented clearly.
2. **Hard dependency on UiPath/cli#2600** — `uip track` must be in the released CLI for this to emit. Harmless if absent, but coordinate release ordering.
3. **Not anonymous by design** is a legitimate choice, but since events tie to real cloud identity it deserves an explicit privacy/retention sign-off rather than riding under a "no PII" framing.

Bottom line: code is ready; fix the description.

gabrielavaduva (approved):

PR #1450, opened 2026-06-12
feat: align plugin/marketplace version with the @uipath/skills npm scheme (PILOT-5516)

gabrielavaduva:
## What

Implements **PILOT-5516**: cuts `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json`, and `.codex-plugin/plugin.json` over from their isolated `0.0.x` track to the `@uipath/skills` version scheme — shared `major.minor` with `package.json` (the CLI-compatibility line), patch as an independent daily counter.

Per the packaging & release model doc: keep `major.minor` in sync across channels, let the patch diverge deliberately per channel.

## Changes

- **One-time cutover:** plugin + marketplace + Codex plugin version `0.0.36` → `1.197.0`
- **`scripts/sync-version.mjs`** now manages all three plugin manifests via one rule:
  - plugin `major.minor` ≠ `package.json` → reset to `M.N.0` (minor advance resets the daily counter), **refusing to downgrade** if `package.json`'s minor is below the plugin's
  - `major.minor` matches → `--bump-patch` advances the patch, otherwise it is left untouched (same script is valid on `main` and on release branches)
  - `marketplace.plugins[0].version` and `.codex-plugin/plugin.json` `version` must hard-equal `plugin.json` `version`
  - guards an empty `marketplace.plugins[]` array
  - `--check` (already wired into `publish.yml` guard) fails on any violation
- **`daily-version-bump.yml`** now calls `sync-version.mjs --bump-patch` instead of reimplementing the bump rule in jq — the rule lives in exactly one place and covers all three manifests (incl. Codex) automatically
- **`docs/RELEASE.md`**: documents the Codex manifest and the single-source bump rule

## Review feedback addressed (@RaduAna-Maria)

- **Codex plugin included** in the sync (inline comments on `marketplace.json`, `daily-version-bump.yml`, `docs/RELEASE.md`).
- **Bump rule consolidated** into `sync-version.mjs --bump-patch` (the suggested follow-up) — no longer duplicated in jq.
- **Downgrade guard** added (the unconditional-reset finding).
- **Empty-`plugins[]` guard** added (nit).
- **`--check`/`--sync` messages** now report the plugin version too (nit).

## Notes for reviewers / ops

- The `0.0.36 → 1.197.0` jump fires plugin auto-update for current users once merged (an upgrade, monotonic — no downgrade-detection issue).
- **Before/at merge:** if a stale open `auto/version-bump-*` PR exists, close it and delete its branch — its `0.0.x` numbering is obsolete and the dedupe check would skip the first run of the new scheme. (None open at time of writing.)
- Jira note: PILOT-5516 shows Merged/Done but the repo did not contain the implementation — this PR is it; the ticket should be re-pointed at this PR.
- Follow-ups land separately: `release/latest` stable ref (PILOT-5544, #1451), Sunday sprint-cut automation (PILOT-5518, #1452).

## Verification

- `npm run version:sync` → brings `.codex-plugin/plugin.json` to `1.197.0`; second run reports already in sync (idempotent)
- `npm run version:check` → exit 0 (`All manifests in sync (package 1.197.0, plugin 1.197.0, …)`)
- `node scripts/sync-version.mjs --bump-patch` → advances all three manifests `1.197.0 → 1.197.1`
- Negative test: plugin minor above package minor → script exits 1 (`Refusing to downgrade`)
- After merge: dispatch **Daily Version Bump** → PR should show `1.197.0 → 1.197.1` touching all three manifests

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
       "name": "uipath",
       "source": "./",
       "description": "UiPath plugin for Claude Code — custom skills, agents, hooks, and MCP servers for UiPath workflows, UI automation, UI testing and UiPath troubleshoot",
-      "version": "0.0.36",
+      "version": "1.197.0",
       "author": {
         "name": "UiPath"
       },
--- a/.github/workflows/daily-version-bump.yml
+++ b/.github/workflows/daily-version-bump.yml
@@ -44,34 +44,23 @@ jobs:
             echo "exists=false" >> "$GITHUB_OUTPUT"
           fi
 
-      - name: Bump version in plugin.json and marketplace.json
+      - uses: actions/setup-node@v4
+        if: steps.check.outputs.exists != 'true'
+        with:
+          node-version: "22.x"
+
+      - name: Bump plugin version (shared major.minor, daily patch counter)
         id: bump
         if: steps.check.outputs.exists != 'true'
         run: |
-          PLUGIN=".claude-plugin/plugin.json"
-          MARKETPLACE=".claude-plugin/marketplace.json"
-
-          # Plugin version shares major.minor with package.json (the npm
-          # package / CLI line); the patch is an independent daily counter.
-          # When the package minor advances, the counter resets to 0.
-          # See docs/RELEASE.md.
-          PKG_LINE=$(jq -r '.version | split(".") | "\(.[0]).\(.[1])"' package.json)
-          VERSION=$(jq -r '.version' "$PLUGIN")
-          IFS='.' read -r MAJOR MINOR PATCH <<< "$VERSION"
-
-          if [ "$MAJOR.$MINOR" = "$PKG_LINE" ]; then
-            NEW_VERSION="$PKG_LINE.$((PATCH + 1))"
-          else
-            NEW_VERSION="$PKG_LINE.0"
-          fi
-
-          # Update plugin.json
-          jq --arg v "$NEW_VERSION" '.version = $v' "$PLUGIN" > tmp.json && mv tmp.json "$PLUGIN"
-
-          # Update marketplace.json (version is inside plugins[0])
-          jq --arg v "$NEW_VERSION" '.plugins[0].version = $v' "$MARKETPLACE" > tmp.json && mv tmp.json "$MARKETPLACE"
-
-          echo "old_version=$VERSION" >> "$GITHUB_OUTPUT"
+          # Single source of the bump rule: `sync-version.mjs --bump-patch`
+          # advances the plugin/marketplace/Codex patch (resetting onto
+          # package.json's minor line if it has moved on). The rule lives only
+          # in the script — not duplicated in jq here. See docs/RELEASE.md.
+          OLD_VERSION=$(node -p "require('./.claude-plugin/plugin.json').version")
+          node scripts/sync-version.mjs --bump-patch
+          NEW_VERSION=$(node -p "require('./.claude-plugin/plugin.json').version")
+          echo "old_version=$OLD_VERSION" >> "$GITHUB_OUTPUT"
           echo "new_version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
 
       - name: Create Pull Request
@@ -88,6 +77,7 @@ jobs:
             Updated files:
             - `.claude-plugin/plugin.json`
             - `.claude-plugin/marketplace.json`
+            - `.codex-plugin/plugin.json`
           branch: "auto/version-bump-${{ steps.bump.outputs.new_version }}"
           base: main
           assignees: ${{ github.event.inputs.assignees || 'gabrielavaduva,RaduAna-Maria'}}
--- a/.github/workflows/daily-version-bump.yml
+++ b/.github/workflows/daily-version-bump.yml
@@ -51,10 +51,19 @@ jobs:
           PLUGIN=".claude-plugin/plugin.json"
           MARKETPLACE=".claude-plugin/marketplace.json"
 
-          # Read current version from plugin.json
+          # Plugin version shares major.minor with package.json (the npm
+          # package / CLI line); the patch is an independent daily counter.
+          # When the package minor advances, the counter resets to 0.
+          # See docs/RELEASE.md.
+          PKG_LINE=$(jq -r '.version | split(".") | "\(.[0]).\(.[1])"' package.json)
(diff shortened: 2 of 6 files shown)


.claude-plugin/marketplace.json
RaduAna-Maria, line 1: Bump those versions to the codex plugin as well, otherwise they will fall out of sync

.github/workflows/daily-version-bump.yml
```diff
@@ -51,10 +51,19 @@ jobs:
           PLUGIN=".claude-plugin/plugin.json"
           MARKETPLACE=".claude-plugin/marketplace.json"
```
RaduAna-Maria: Add the Codex plugin as well

docs/RELEASE.md
```diff
@@ -4,13 +4,25 @@ The whole skills repo is published as an npm package, **`@uipath/skills`**, vers
 
 ## Version model
 
-`package.json` `version` is the **single source of truth** for the npm package. `scripts/sync-version.mjs` derives this manifest from it (do not edit by hand):
+`package.json` `version` is the **single source of truth** for the npm package. `scripts/sync-version.mjs` derives these manifests from it (do not edit by hand):
 
 | File | Field | Purpose |
 |------|-------|---------|
 | `version-manifest.json` | `skillsVersion`, `targetCli` | CLI↔skills pairing record |
+| `.claude-plugin/plugin.json` | `version` | Claude Code plugin version (shared `major.minor`, independent patch) |
```
RaduAna-Maria: Add Codex

RaduAna-Maria:
**Suggestion (non-blocking): consolidate the version-bump rule so it lives in one place.**

After this PR, the "derive the next version from `package.json` `major.minor`" rule exists in two languages:

- `scripts/sync-version.mjs` (node) — resets plugin/marketplace to `M.N.0` on a minor mismatch, leaves the patch untouched on a match
- `.github/workflows/daily-version-bump.yml` (bash/jq) — re-derives `M.N` from `package.json` and does `patch++` / reset

They agree today, and `version:check` in `publish.yml` guards against drift — but it's the same rule maintained twice, so any future change to the scheme has to land in both copies (and the only thing keeping them honest is the guard catching them after they've already diverged).

Proposal: give `sync-version.mjs` a `--bump-patch` mode that advances the plugin/marketplace patch (resetting the minor when it lags `package.json`), and have the workflow call it instead of reimplementing the logic in jq:

```yaml
# daily-version-bump.yml — replaces the inline jq block
- run: node scripts/sync-version.mjs --bump-patch
```

Then:
- the bump rule lives in exactly one place
- the workflow shrinks to "run the script → open a PR"
- `--check` guards the same code path that performs the write, rather than a second copy of the rule

Not a blocker for the cutover — fine as a follow-up.

RaduAna-Maria:
## Review — plugin/marketplace version alignment (PILOT-5516)

Clean foundation for the stack. The "shared `major.minor`, independent patch counter" rule in `sync-version.mjs` is well-documented, and the same-script-valid-on-main-and-release-branches property (matching minor ⇒ patch untouched) is the right call. Minor points only:

### 🟡 `minorLine` reset is unconditional on mismatch — including downgrades
```js
if (minorLine(pluginVersion) !== minorLine(version)) {
  const next = `${minorLine(version)}.0`;   // resets to package.json's minor
```
This resets the plugin/marketplace version whenever the minors *differ in either direction*. If `package.json`'s minor were ever below the plugin's (a bad manual edit, a revert), `version:sync` would silently rewrite the plugin to a **lower** version — and plugin auto-update doesn't downgrade, so users would freeze. In practice `package.json` only moves forward so this can't happen today; a one-line guard (only reset when package minor ≥ plugin minor, else error) would make the invariant explicit rather than assumed.

### 🟢 nit — `marketplace.plugins[0]` assumes a non-empty array
Throws if the manifest shape ever changes. Negligible risk for a controlled repo file; flagging only for completeness.

### 🟢 nit — the shell reset branch is effectively dead code
In `daily-version-bump.yml`:
```bash
if [ "$MAJOR.$MINOR" = "$PKG_LINE" ]; then
  NEW_VERSION="$PKG_LINE.$((PATCH + 1))"
else
  NEW_VERSION="$PKG_LINE.0"   # never reached in practice
fi
```
On the release line `package.json` is frozen at `M.N.0` and the plugin starts at `M.N.0`, so the minors always match and the `else` never fires — the reset only ever genuinely happens inside `sync-version.mjs` (at cut time). It's harmless and keeps parity with the script, just noting it's not load-bearing.

### Doc nit
The description still says `0.0.34 → 1.196.0`, but the actual diff is `0.0.36 → 1.197.0`. Worth updating so the "close the stale 0.0.35 PR" ops note matches reality.

gabrielavaduva:
@RaduAna-Maria thanks for the review — pushed `a261224` addressing it:

- **Codex plugin now synced** (your 3 inline comments). `.codex-plugin/plugin.json` is added to `sync-version.mjs`; its `version` mirrors `.claude-plugin/plugin.json` exactly, same as `marketplace.json`.
- **Bump rule consolidated** (your follow-up suggestion). Added `--bump-patch` to `sync-version.mjs`; `daily-version-bump.yml` now calls it instead of reimplementing the rule in jq — so the rule lives in one place and covers all three manifests (incl. Codex) automatically.
- **Downgrade guard** added — `sync-version.mjs` now errors instead of resetting the plugin onto a lower minor than its current version.
- **Empty `plugins[]` guard** added.
- **`--check`/`--sync` messages** now report the plugin version too (so "in sync at 1.197.5" reads correctly).
- **PR description** corrected to `0.0.36 → 1.197.0`.

Verified locally: sync brings Codex to `1.197.0`, `--check` exits 0, `--bump-patch` advances all three to `1.197.1`, and the downgrade guard exits 1.

.github/workflows/daily-version-bump.yml
DragosUnguru, line 1: Why do we do a **"daily"** version bump?

DragosUnguru (commented):
Are we shipping only once every 2 weeks (sprint release)?

.github/workflows/daily-version-bump.yml
gabrielavaduva: Fair question — the answer is "it largely shouldn't." The daily bump existed to drive Claude Code / Codex plugin auto-update, but on the `release/latest` model the line is frozen between sprint cuts, so it fires effectively never. Owner has decided to retire it in favor of a per-sprint cadence — tracked in #1712.

PR #1451, opened 2026-06-12
feat: release/latest stable ref for the marketplace + daily bump retarget (PILOT-5544)

gabrielavaduva:
## What

Implements **PILOT-5544**: the stable ref the Claude Code marketplace points at permanently — `release/latest`, always equal to the HEAD of the newest `release/v<minor>` branch — and retargets the daily plugin-version bump to the release line.

> **Stacked on #1450** (PILOT-5516) — the bump logic relies on the shared `major.minor` scheme. Merge #1450 first; this PR's base is `feat/plugin-version-alignment` and should auto-retarget to `main`.

## Design

Plugin auto-update fires off the `version` field on the **tracked ref**. Once the marketplace tracks `release/latest`, daily bumps to `main` would never reach plugin users — so the daily bump now:

1. Checks out `release/latest` (graceful skip + step-summary note while the ref doesn't exist yet — protects the scheduled run between this PR merging and the branches being created)
2. **Skip-guard:** if HEAD is already a bump commit, nothing landed since the last bump → skip (no forced plugin re-downloads for identical content)
3. Bumps plugin + marketplace + Codex via `sync-version.mjs --bump-patch` (PR #1450 logic: shared `major.minor`, daily patch counter, single source of the rule)
4. Pushes **atomically** to `release/v<minor>` and fast-forwards `release/latest`

`main` gets **no** daily bumps — its plugin version stays pinned at `M.N.0` per sprint. A shared `concurrency` group (`release-line-updates`) prevents racing the Sunday sprint cut (PILOT-5518, #1452). The legacy PR-to-main flow is kept behind a `workflow_dispatch` boolean (`legacy_main_bump`) for one transition sprint, then delete.

## Review feedback addressed (@RaduAna-Maria)

- **Atomic dual push** — `git push --atomic` so `release/v<minor>` and `release/latest` both update or neither does (no out-of-sync invariant break on a partial failure).
- **Skip-guard no longer string-coupled** — the bump-commit subject now lives in one job-level `BUMP_COMMIT_PREFIX` env constant that both the guard and the commit step read, so they can't silently desync.
- **Consolidation carried in** — both bump jobs call `sync-version.mjs --bump-patch` instead of reimplementing the rule in jq (and so pick up the Codex manifest automatically).
- **Setup-version refs corrected** to the live `1.197` line (below).

## Operational steps after merge (one-time)

```bash
git push origin main:refs/heads/release/v1.197
git push origin main:refs/heads/release/latest
```

(Cut from post-#1450 `main`, which is on the `1.197` line, NOT from `release/v1.195` — that line predates the 1.197.x plugin scheme and its content is a sprint old.)

**Repo settings:** allow Actions push on `release/v*` and `release/latest`, plus **force push** on `release/latest` (sprint handoff is a forced ref update — the new branch doesn't contain the old line's daily-bump commits).

**Migration caveat:** existing manual installs added as bare `UiPath/skills` track `main` and will silently freeze once bumps leave main. Docs now instruct `/plugin marketplace add UiPath/skills@release/latest`; announce the re-add to existing users.

## Verification

- `npx js-yaml .github/workflows/daily-version-bump.yml` — parses clean
- After branch creation: dispatch **Daily Version Bump** → one commit on `release/v1.197`, `release/latest` at the same SHA, plugin/marketplace/Codex bumped to `1.197.1`; re-dispatch same day → skipped via guard

## Related

- PILOT-5518 (Sunday sprint-cut automation that advances `release/latest`) lands in #1452.

### DIFF
--- a/.github/workflows/daily-version-bump.yml
+++ b/.github/workflows/daily-version-bump.yml
@@ -44,34 +44,23 @@ jobs:
             echo "exists=false" >> "$GITHUB_OUTPUT"
           fi
 
-      - name: Bump version in plugin.json and marketplace.json
+      - uses: actions/setup-node@v4
+        if: steps.check.outputs.exists != 'true'
+        with:
+          node-version: "22.x"
+
+      - name: Bump plugin version (shared major.minor, daily patch counter)
         id: bump
         if: steps.check.outputs.exists != 'true'
         run: |
-          PLUGIN=".claude-plugin/plugin.json"
-          MARKETPLACE=".claude-plugin/marketplace.json"
-
-          # Plugin version shares major.minor with package.json (the npm
-          # package / CLI line); the patch is an independent daily counter.
-          # When the package minor advances, the counter resets to 0.
-          # See docs/RELEASE.md.
-          PKG_LINE=$(jq -r '.version | split(".") | "\(.[0]).\(.[1])"' package.json)
-          VERSION=$(jq -r '.version' "$PLUGIN")
-          IFS='.' read -r MAJOR MINOR PATCH <<< "$VERSION"
-
-          if [ "$MAJOR.$MINOR" = "$PKG_LINE" ]; then
-            NEW_VERSION="$PKG_LINE.$((PATCH + 1))"
-          else
-            NEW_VERSION="$PKG_LINE.0"
-          fi
-
-          # Update plugin.json
-          jq --arg v "$NEW_VERSION" '.version = $v' "$PLUGIN" > tmp.json && mv tmp.json "$PLUGIN"
-
-          # Update marketplace.json (version is inside plugins[0])
-          jq --arg v "$NEW_VERSION" '.plugins[0].version = $v' "$MARKETPLACE" > tmp.json && mv tmp.json "$MARKETPLACE"
-
-          echo "old_version=$VERSION" >> "$GITHUB_OUTPUT"
+          # Single source of the bump rule: `sync-version.mjs --bump-patch`
+          # advances the plugin/marketplace/Codex patch (resetting onto
+          # package.json's minor line if it has moved on). The rule lives only
+          # in the script — not duplicated in jq here. See docs/RELEASE.md.
+          OLD_VERSION=$(node -p "require('./.claude-plugin/plugin.json').version")
+          node scripts/sync-version.mjs --bump-patch
+          NEW_VERSION=$(node -p "require('./.claude-plugin/plugin.json').version")
+          echo "old_version=$OLD_VERSION" >> "$GITHUB_OUTPUT"
           echo "new_version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
 
       - name: Create Pull Request
@@ -88,6 +77,7 @@ jobs:
             Updated files:
             - `.claude-plugin/plugin.json`
             - `.claude-plugin/marketplace.json`
+            - `.codex-plugin/plugin.json`
           branch: "auto/version-bump-${{ steps.bump.outputs.new_version }}"
           base: main
           assignees: ${{ github.event.inputs.assignees || 'gabrielavaduva,RaduAna-Maria'}}
--- a/.github/workflows/daily-version-bump.yml
+++ b/.github/workflows/daily-version-bump.yml
@@ -51,10 +51,19 @@ jobs:
           PLUGIN=".claude-plugin/plugin.json"
           MARKETPLACE=".claude-plugin/marketplace.json"
 
-          # Read current version from plugin.json
+          # Plugin version shares major.minor with package.json (the npm
+          # package / CLI line); the patch is an independent daily counter.
+          # When the package minor advances, the counter resets to 0.
+          # See docs/RELEASE.md.
+          PKG_LINE=$(jq -r '.version | split(".") | "\(.[0]).\(.[1])"' package.json)
           VERSION=$(jq -r '.version' "$PLUGIN")
           IFS='.' read -r MAJOR MINOR PATCH <<< "$VERSION"
-          NEW_VERSION="$MAJOR.$MINOR.$((PATCH + 1))"
+
+          if [ "$MAJOR.$MINOR" = "$PKG_LINE" ]; then
+            NEW_VERSION="$PKG_LINE.$((PATCH + 1))"
+          else
+            NEW_VERSION="$PKG_LINE.0"
+          fi
 
           # Update plugin.json
           jq --arg v "$NEW_VERSION" '.version = $v' "$PLUGIN" > tmp.json && mv tmp.json "$PLUGIN"
--- a/.github/workflows/daily-version-bump.yml
+++ b/.github/workflows/daily-version-bump.yml
@@ -1,12 +1,25 @@
 name: Daily Version Bump
 
+# Bumps the plugin version (Claude Code plugin.json + marketplace.json + Codex
+# plugin.json) on the current release line. The marketplace tracks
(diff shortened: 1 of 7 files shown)


RaduAna-Maria:
## Review — release/latest + daily-bump retarget (PILOT-5544)

The core design is solid: graceful skip while `release/latest` doesn't exist yet, the "HEAD is already a bump commit" skip-guard, the shared `release-line-updates` concurrency group, and failing loudly when the durable branch is missing are all the right instincts. Three things:

### 🟠 The one-time setup commands are a minor behind
The PR body says:
```bash
git push origin main:refs/heads/release/v1.196
git push origin main:refs/heads/release/latest
```
but post-#1450 `main` is on `1.197.0` (#1450 seeds `1.197.0`, not `1.196.0`). The daily bump derives the target branch from `package.json` on `release/latest`:
```bash
PKG_LINE=$(jq -r '.version | split(".") | "\(.[0]).\(.[1])"' package.json)   # -> 1.197
...
release_branch=release/v$PKG_LINE                                            # -> release/v1.197
```
So if ops create `release/v1.196` as documented, the very first bump hits the "branch missing" guard and fails (`release/latest is at <sha> but release/v1.197 does not exist`). The setup commands + the "Verification" section (`1.196.1`, `release/v1.196`) should be `1.197`.

### 🟡 Non-atomic dual push
```bash
git push origin "HEAD:$BRANCH" "HEAD:release/latest"
```
This isn't atomic — if the first refspec lands and the second fails (transient), `release/v<minor>` ends up ahead of `release/latest`, breaking the "they're always equal" invariant the next run assumes. `git push --atomic origin ...` makes both succeed or both fail.

### 🟡 Skip-guard is coupled to the commit subject string
```bash
SUBJECT=$(git log -1 --format=%s)
if [[ "$SUBJECT" == "chore: bump plugin version to"* ]]; then
```
This literal must stay byte-identical to the commit message produced in the bump step. If someone edits one and not the other, the guard silently stops matching → daily bumps resume forcing identical-content plugin re-downloads (the exact thing the guard prevents), with no failure to alert you. Low severity, but a shared constant or a more robust marker (e.g. checking whether the diff since the last non-bump commit is empty) would be harder to desync.

### Note
The `legacy-main-bump` job duplicates the whole bump body. Fine as an explicitly-temporary transition path — just make sure the "delete after one sprint" actually happens, otherwise it's dead weight.

gabrielavaduva:
Pushed (rebased onto the updated #1450) addressing the review:

- **🟡 Atomic dual push** — now `git push --atomic origin HEAD:$BRANCH HEAD:release/latest`, so the two refs never end up out of sync on a partial failure.
- **🟡 Skip-guard no longer string-coupled** — the bump-commit subject is now a single job-level `BUMP_COMMIT_PREFIX` env constant that both the guard and the commit step read.
- **Consolidation carried in** — both bump jobs call `sync-version.mjs --bump-patch` (no more duplicated jq; picks up the Codex manifest automatically).
- **Setup commands + verification** corrected from `1.196` to the live `1.197` line.

The `legacy-main-bump` job is still an explicit transition-only path — flagged for deletion after one sprint as you noted.

DragosUnguru:
**Design review (zoom-out).** Three model-level observations on the release-line design — numbered to match the offline review. The mechanics in this PR are solid (atomic dual-push, the downgrade guard, the recursion-guard reasoning all check out); these are about the model, not the code.

### 1. The release line has no content path between sprint cuts — so the daily bump is either vestigial or depends on an undocumented hotfix flow

`release/latest` is cut from `main` once per sprint and then frozen except for version-bump commits — there is no forward-port / cherry-pick automation from `main` → `release/v<minor>`. On a frozen branch the daily bump fires **exactly once per sprint**:

- cut delivers `M.N.0` (new content) ✓
- next day: HEAD is the cut commit → skip-guard misses → bumps to `M.N.1` (identical content)
- every day after: HEAD is a bump commit → skip-guard fires → skips until the next cut

Net effect: **Claude Code + Codex plugin users receive new skills only every two weeks**, where today (marketplace tracking `main` via the bump-to-main flow) they get them ~daily. That's a latency regression on the plugin channel the PR doesn't call out.

This leaves the system in *neither* clean end-state:

- **Accept per-sprint cadence** → the daily bump, the independent patch counter, and most of #1450's new `sync-version.mjs` logic can be deleted; plugin version simply equals `package.json` (`M.N.0`/sprint), and the cut alone changes the version and triggers auto-update. Large simplification.
- **Want freshness** → a `main` → `release/latest` forward-port (or hotfix cherry-pick) flow is the missing prerequisite, and the daily bump exists to version it.

As shipped it keeps the machinery of the second option but the content model of the first. Can we pin down whether content needs to reach plugin users *between* cuts? The answer collapses a lot of this machinery either way.

### 3. Dual-ref complexity is only justified if we service multiple minor lines

`release/v<minor>` + `release/latest`, the atomic dual-push, and the force-push branch-protection carve-out all exist to support a durable per-minor history. If we never hotfix an *old* line, `release/latest` alone does the job and all of that drops out. Is multi-line servicing a real, near-term need? If not, this is complexity that won't be exercised.

### 4. Cutover hinges on what ref `uip skills install` pins (cross-repo, unverified here)

The README change correctly handles manually-added bare `UiPath/skills` marketplaces (re-add `@release/latest`). But it's silent on the **primary** install path. If `uip skills install` adds bare `UiPath/skills` (tracking `main`), then once `main` stops getting bumps **the entire installed base silently freezes at cutover** — not just manual installs. Can we confirm `uip skills install` pins `@release/latest` before merging? If it doesn't, the cutover strands every existing user. This is the highest-blast-radius item in the series.

.github/workflows/daily-version-bump.yml
```diff
@@ -1,21 +1,139 @@
...
+          echo "exists=true" >> "$GITHUB_OUTPUT"
+          git fetch origin release/latest
+          git checkout --detach FETCH_HEAD
+
+      - name: Skip if nothing landed since the last bump
+        id: guard
+        if: steps.resolve.outputs.exists == 'true'
+        run: |
+          SUBJECT=$(git log -1 --format=%s)
+          if [[ "$SUBJECT" == "$BUMP_COMMIT_PREFIX"* ]]; then
```
DragosUnguru: **#5 — this skip-guard misses the cut commit, forcing a redundant re-download every sprint.**

It matches only the bump-commit subject (`BUMP_COMMIT_PREFIX`). The morning after each sprint cut, HEAD is the *cut* commit (`chore(release): cut <line> line`, from #1452) — so the guard doesn't fire, and the bump goes `M.N.0` → `M.N.1` with byte-identical content, forcing a full plugin re-download for every user. That's exactly what this step's own comment says it prevents ("Bumping again would force a plugin re-download for identical content"). It's deterministic — happens every sprint.

Fix: treat the cut-commit subject as a no-new-content state too (match both prefixes), or compare trees rather than commit subjects.

gabrielavaduva: Good catch — confirmed deterministic, fixed in `c149e65`. The skip-guard now treats the sprint-cut commit (`chore(release): cut <line> line`) as a no-new-content state too, alongside the bump commit, so the morning-after bump no longer re-versions identical content.

On your zoom-out **#1**: agreed the daily bump is vestigial on a frozen release line. Owner has decided to retire it and move to a per-sprint cadence (cut drives auto-update). Tracked in #1712 — so this skip-guard fix is a stopgap; the whole daily-bump path goes away there. Also flagged your **#3** (dual-ref only justified for multi-line servicing) in that issue for re-evaluation.

DragosUnguru (approved):

PR #1452, opened 2026-06-12
feat: biweekly Sunday sprint-release-cut — skills lead + auto-publish (PILOT-5518)

gabrielavaduva:
Skills side of the CLI↔skills release lockstep (PILOT-5518). Adds `sprint-release-cut.yml`.

**Rebased directly onto `main` — #1450 is merged, and #1451 is intentionally skipped** (per-sprint cadence; no `release/latest` marketplace channel, no daily bump — daily-bump retirement tracked in #1712). Scope is exactly: **cut a release branch on Sunday + publish.**

## What each scheduled run does
1. Cut `release/v<minor>` from `main` at `M.N.0` (`sync-version.mjs` resets plugin/marketplace/Codex too).
2. Publish `@uipath/skills@M.N.0` to npm `latest` — create the GitHub Release `v<minor>.0` (durable record) and **dispatch `publish.yml` on the tag** (a `GITHUB_TOKEN`-created release can't trigger workflows itself; `npm view`-guarded so a re-run won't re-dispatch an already-published version).
3. Open the matching version-bump PR against `main` (so `main`'s version advances per sprint).

## Design
- **Schedule:** Sunday 06:00 UTC, gated to the 14-day cadence anchored at `2026-06-14` — same cadence as `UiPath/cli`, 6 h ahead so the skills package lands first.
- **Self-driven:** target line = current skills minor **+ 1**; never reads the CLI version → no cross-repo secret.
- **Resumable:** an incomplete cut re-runs idempotently (branch already at `M.N.0` → resume; different version → stop loudly).
- Manual `minor_override` / `dry_run` dispatch inputs for off-cadence or ad-hoc cuts.

## Not included (deliberately)
- No `release/latest` ref / marketplace-channel machinery (that was #1451). The Claude Code / Codex plugin marketplace keeps tracking `main`, versioned per sprint by the bump PR in step 3 — so no CLI `@release/latest` change is required.

## Repo setup before first run
- Branch protection must allow the Actions identity to push `release/v*`.
- npmjs Trusted Publishing for `publish.yml` (already configured per `docs/RELEASE.md`).
- Recommend a one-time `dry_run:false` validation on a throwaway minor to confirm the publish dispatch reaches npm end-to-end.

### DIFF
--- a/.github/workflows/sprint-release-cut.yml
+++ b/.github/workflows/sprint-release-cut.yml
@@ -108,6 +108,27 @@ jobs:
             } >> "$GITHUB_STEP_SUMMARY"
           else
             echo "proceed=true" >> "$GITHUB_OUTPUT"
+
+            # Warn-only drift check (NO gate — skills lead, never follow). Skills
+            # cuts 6 h before the CLI on the same anchor, so at this moment the
+            # CLI's latest published minor should be exactly one behind our
+            # target. A gap in either direction means a skipped/extra cut has
+            # drifted the lines apart — surface it (it's otherwise silent, and
+            # `targetCli` would keep claiming a pairing that no longer holds).
+            # Reads public npm only — no token, no cross-repo coupling.
+            CLI_LATEST=$(npm view @uipath/cli version 2>/dev/null || echo "")
+            if [ -n "$CLI_LATEST" ]; then
+              CLI_MINOR=$(echo "$CLI_LATEST" | cut -d. -f1-2)
+              EXPECTED=$(echo "$CLI_MINOR" | awk -F. '{print $1"."$2+1}')
+              if [ "$TARGET_LINE" != "$EXPECTED" ]; then
+                echo "::warning::Skills target $TARGET_LINE is not CLI_latest+1 (CLI latest published is $CLI_LATEST, minor $CLI_MINOR) — the skills/CLI lines may have drifted. Realign with a minor_override dispatch (see docs/RELEASE.md)."
+                {
+                  echo "## ⚠️ Possible skills/CLI line drift"
+                  echo ""
+                  echo "Target is \`$TARGET_LINE\`; the CLI's latest npm release is \`$CLI_LATEST\` (minor \`$CLI_MINOR\`), so the expected target is \`$EXPECTED\`. If unexpected, realign with a \`minor_override\` dispatch — see docs/RELEASE.md. (Proceeding with the cut regardless — this is a warning, not a gate.)"
+                } >> "$GITHUB_STEP_SUMMARY"
+              fi
+            fi
           fi
 
       - name: Dry run — print plan
--- a/.github/workflows/sprint-release-cut.yml
+++ b/.github/workflows/sprint-release-cut.yml
@@ -1,0 +1,234 @@
+name: Sprint Release Cut
+
+# Biweekly per-sprint release cut (skills side of the CLI↔skills lockstep,
+# PILOT-5518). Runs Sunday 06:00 UTC — 6 hours BEFORE the CLI's own cut
+# (12:00 UTC) — so the skills npm package is live first. Self-anchored and
+# self-driven: it does NOT read the CLI version (skills lead, never follow).
+# On each release Sunday it:
+#
+#   1. cuts release/v<minor> from main with package.json = M.N.0
+#      (sync-version.mjs resets plugin/marketplace/Codex to M.N.0 too)
+#   2. publishes @uipath/skills@M.N.0 to npm `latest` — creates the GitHub
+#      Release v<minor>.0 (durable record) and dispatches publish.yml on the
+#      tag (a GITHUB_TOKEN-created release can't trigger publish.yml itself)
+#   3. opens a PR against main with the same version bump
+#
+# No cross-repo secret is required — the target line is the current skills
+# minor + 1, gated to the CLI's 14-day cadence (anchor below). The cli-repo
+# side cuts itself, in lockstep, from the same anchor.
+
+on:
+  schedule:
+    - cron: '0 6 * * 0' # Sunday 06:00 UTC — 6 h before the CLI cut (12:00 UTC)
+  workflow_dispatch:
+    inputs:
+      minor_override:
+        description: 'Target line as MAJOR.MINOR (e.g. 1.198) — skips the cadence gate and the auto-increment'
+        required: false
+        default: ''
+      dry_run:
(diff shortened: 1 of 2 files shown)


RaduAna-Maria:
## Review — sprint-release-cut.yml (PILOT-5518)

Reviewed the full stack (#1450 → #1451 → #1452). The cadence design is sound — I confirmed `2026-06-14` is a Sunday, so firing every Sunday + gating on `days_since % 14 == 0` aligns to every *other* Sunday as intended. The `minor_override` regex-validation and `dry_run` plan output are nice. A few issues, one of them blocking:

### 🔴 Blocking — the auto-publish chain won't fire
The "Publish stable to npm" step creates the GitHub Release with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`:

```yaml
env:
  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
...
gh release create "$TAG" --target ... --notes ...
```

GitHub **does not start new workflow runs from events created by the default `GITHUB_TOKEN`** (the only exceptions are `workflow_dispatch` and `repository_dispatch`) — this is the recursion-prevention rule. So the `release: published` event *will* be emitted, but `publish.yml` will **not** run, and `@uipath/skills@M.N.0` will never reach npm. Step 3 of the design ("auto-publish") silently no-ops; you'd get a tag/release with nothing on npm.

Two clean fixes:
- **(a)** Create the release with a PAT / GitHub App token (a non-default token *does* trigger downstream workflows).
- **(b)** Keep `GITHUB_TOKEN`, but stop relying on the release event — after creating the release, explicitly dispatch the publisher:
  ```bash
  gh workflow run publish.yml --ref "$TAG" -f target=npmjs
  ```
  `workflow_dispatch` *is* triggerable by `GITHUB_TOKEN`, and `publish.yml`'s dispatch path (`target == 'npmjs'`) checks out the given ref (the tag = exact cut state) and publishes via OIDC. No extra secret needed — probably the better fit given the "no cross-repo secret" goal.

Worth a one-time end-to-end `dry_run: false` validation on a throwaway minor to confirm npm actually receives the package.

### 🟠 Not resumable after a partial failure
The cut step hard-errors if the branch already exists:
```yaml
if git ls-remote --exit-code --heads origin "$BRANCH" > /dev/null; then
  echo "::error::$BRANCH already exists ... resolve manually."
  exit 1
```
If the branch push succeeds but a later step fails (npm hiccup, etc.), re-running the workflow aborts here — which means the `gh release view`-already-exists guard and the `gh pr list` existing-PR guard in the later steps are effectively **unreachable on a re-run**. They suggest resumability that doesn't exist. Consider: if the branch already exists *and points at the expected cut SHA*, skip creation and continue, rather than failing.

### 🟠 Hard coupling to the main-bump PR being merged
`current_line` is read from `main`'s `package.json`. If the auto-opened bump PR isn't merged before the next cut, `current_line` is still `N`, `target = N+1` = the branch that already exists → the cut errors `"$BRANCH already exists but main is still on N"`. It fails loudly with a clear message (good), but it's an operational dependency worth calling out in RELEASE.md.

### Nit
- Seed/version references in the description are a sprint behind the `1.197` line (same as #1451) — see my comment there.

gabrielavaduva:
Pushed `9a0b1ef` (rebased onto the updated stack) addressing the review:

- **🔴 Auto-publish chain fixed** via your option (b): still create the GitHub Release as the durable record, but **dispatch `publish.yml` explicitly** on the tag (`gh workflow run publish.yml --ref v<minor>.0 -f target=npmjs`) since a `GITHUB_TOKEN`-created release can't trigger workflows. Guarded on `npm view` so a repeat run doesn't re-dispatch an already-published version. Header comment + `docs/RELEASE.md` step 3 corrected (they claimed `release: published` would fire it). Agreed a one-time `dry_run:false` e2e on a throwaway minor is worth doing before relying on it.
- **🟠 Resumable** — if `release/v<minor>` already exists at the expected `M.N.0`, the cut resumes idempotently instead of aborting; only a *version mismatch* stops loudly.
- **🟠 Main-bump-PR coupling** documented in `docs/RELEASE.md`. Note the behavior is now "resume the same line idempotently" rather than a hard error, so an unmerged bump PR silently stalls the cadence — called that out explicitly with advice to merge promptly.
- Codex manifest now included in the cut commit.

DragosUnguru:
**Design review (zoom-out) — #2: the CLI↔skills lockstep is coordination-by-anchor, not by handshake.**

Dropping `CLI_REPO_TOKEN` for a self-incrementing `current + 1` is a reasonable simplicity win, but it trades a hard dependency for an unguarded assumption: both repos cut on every 14-day anchor, forever. If either side skips or double-cuts even once, the minor lines drift **permanently** and `targetCli` points at a line that doesn't pair — with no alarm and no reconciliation anywhere. Recovery is manual via `minor_override`, but that isn't surfaced as the recovery path.

Worth at least a cheap drift check (compare the skills minor against `npm view @uipath/cli` latest minor, warn on mismatch) and documenting `minor_override` as the realignment lever.

(Cross-ref: this workflow is also the *only* content-delivery point to the plugin channel between sprints — see #1 on #1451.)

.github/workflows/sprint-release-cut.yml
```diff
@@ -0,0 +1,240 @@
...
+
+          # A release created with the default GITHUB_TOKEN does NOT start other
+          # workflow runs (GitHub's recursion guard), so `release: published`
+          # will NOT trigger publish.yml. Dispatch it explicitly instead —
+          # workflow_dispatch IS triggerable by GITHUB_TOKEN. publish.yml's
+          # npmjs path checks out the given ref ($TAG = the exact cut state,
+          # package.json = $LINE.0) and publishes via OIDC trusted publishing.
+          # Guard on npm so a resumed/repeated run doesn't re-dispatch an
+          # already-published version (which would only fail noisily).
+          if npm view "@uipath/skills@$LINE.0" version > /dev/null 2>&1; then
```
DragosUnguru, line 211: **#6 — two live trigger paths into `publish.yml`, only this one is guarded.**

`publish.yml` still fires on `release: published`, and this step also dispatches it explicitly. The no-double-publish guarantee rests entirely on "a `GITHUB_TOKEN`-created release doesn't trigger workflows" — correct today. But a *human*-created GitHub Release (the documented manual stable path in `docs/RELEASE.md`) still triggers `publish.yml` with no `npm view` guard; only npm's own duplicate rejection prevents a double publish. This `npm view` guard lives only on the dispatch path — the two paths are asymmetric. Not blocking, but worth knowing.

PR #1467, opened 2026-06-13
ci: pin coder_eval to a published version (wheel + GHCR image)

bai-uipath:
Pin `coder_eval` to a single explicit published version (`tests/.coder-eval-version`) for every skills path that runs the engine — PR smoke (skills + RPA), the activation gate, the advisory task-schema gate, the `Run Coder Eval` dispatch workflow, and the nightly — and consume it as the published `coder-eval` wheel plus the `coder-eval-agent` image from GHCR, instead of each path checking out and building `coder_eval` from its default branch.

Today those paths each float on `coder_eval@main`, so an engine change can shift eval results with no deliberate decision on our side. A single pinned version makes the engine reproducible, turns version moves into an explicit reviewable bump, and gives the PR gate one version to validate against before promoting it. The `coder_eval_version` dispatch input still overrides per run to test a published version ad-hoc.

The internal nightly runner consumes the same pin: UiPath/coder_eval_uipath#8.

## Validation

All against the pinned `0.6.2` (consumed as both the wheel version and the GHCR image tag).

Eval VM — artifact preconditions:
- GHCR `coder-eval-agent:0.6.2` is present and pullable with the runner's token.
- `coder-eval[dev,uipath]==0.6.2` resolves with its full dependency set from both Azure feeds (exit 0).

CI on this branch:

| Path | Driver | Result |
|---|---|---|
| Validate task schema (advisory) | tempdir / linux | ✅ pass |
| RPA skill smoke | tempdir / windows | ✅ success |
| Skill smoke | docker / linux | ✅ success |
| Run Coder Eval (dispatch, 1 task) | docker / linux | ✅ success |
| Activation gate | tempdir / linux | skipped — no `SKILL.md` frontmatter change to gate; install path is identical to the green advisory gate |

Both docker jobs exercise the full chain end-to-end — `Resolve coder_eval pin` → `Install coder-eval` → `Pull pinned coder-eval-agent image` → `Build skills Docker image` → `Run coder-eval`, all green — so the registry-pull-in-CI path is confirmed. The tempdir paths install the same pinned wheel without an image.


akshaylive (approved):
More comments on a related PR. Unblocking this for now.

PR #1501, opened 2026-06-16
feat(agents): support solution-internal escalation app discovery

al3xanndru:
## Summary
- Updated `escalation.md` discovery (Step 2) to support both solution-internal (`--source local`) and external (`--source remote`) Action Center apps, removing the "escalation apps are always external" restriction
- Added new `local_escalation` e2e test with a real Studio Web fixture (`AgentWithLocalEscalation`) containing an `Agent` project and a `SimpleApprovalApp` (workflow Action) in the same solution
- The test validates that the skill guides the agent to discover the local app and wire an escalation resource with `folderName: "solution_folder"`

## Test plan
- [x] https://github.com/UiPath/skills/actions/runs/27612229029
- [x] `local_escalation` eval passes 8/8 (score 1.000) — agent discovers `SimpleApprovalApp` and wires escalation correctly
- [x] `external_escalation` eval passes 10/10 (score 1.000) — no regression on external app flow
- [x] `inline_solution_escalation` eval passes 10/10 (score 1.000) — no regression on inline flow
- [x] `inline_external_escalation` eval passes 11/12 (score 0.929) — pre-existing `flow registry search` miss, unrelated


tests/tasks/uipath-agents/lowcode/local_escalation/local_escalation.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+pre_run:
+  - command: "cp -r $SKILLS_REPO_PATH/tests/tasks/uipath-agents/lowcode/local_escalation/_fixtures/AgentWithLocalEscalation ."
+    timeout: 10
+
+initial_prompt: |
+  Add an escalation resource named "ApprovalEscalation" to the existing
+  "Agent" agent inside the existing "AgentWithLocalEscalation" solution
+  (in the current directory). The escalation should use the Action Center
+  app that is already part of this solution. Set `isEnabled` to true.
```
andreitava-uip: we shouldn't mention isEnabled true in the test. If the skill fails to set it it is a skill gap. I fixed that issue in this PR: https://github.com/UiPath/skills/pull/1487

tests/tasks/uipath-agents/lowcode/local_escalation/local_escalation.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+  Reuse both projects as-is — do NOT re-create any projects and do NOT
+  re-initialize the solution.
```
andreitava-uip: This should also not be part of the prompt. I expect the skill to determine this itself. If we insist on it we are not properly testing that the agent can correctly know to modify existing solutions without being explicitly instructed.

tests/tasks/uipath-agents/lowcode/local_escalation/local_escalation.yaml
```diff
@@ -0,0 +1,90 @@
...
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent used --output json on uip commands"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+.*--output\s+json'
+    min_count: 1
+    weight: 1.0
+    pass_threshold: 1.0
```
andreitava-uip: nit: i'd have this as pass threshold 0.0.

If the agent successfuly created all files but did not use json mode, is that really a hard failure for us?

tests/tasks/uipath-agents/lowcode/local_escalation/_fixtures/AgentWithLocalEscalation/Agent/evals/evaluators/evaluator-default-trajectory.json
andreitava-uip, line 1: why does this fixture have evals? what are they used in?

al3xanndru: These are the default files that were created when I created the agent in SW.

andreitava-uip (approved):

PR #1511, opened 2026-06-16
docs(uipath-maestro-case): note pack fails on v20 caseplans, point to solution pack

tsathkumara:
## Summary
Adds a Known-issue callout to the `uip maestro case pack` section of the `uipath-maestro-case` command reference (`references/case-commands.md`).

On current CLI builds, `uip maestro case pack` throws `Invalid case plan input — expected { root, nodes, edges }` for a **v20 (unified-schema)** caseplan **even when `uip maestro case validate` reports it Valid**. Root cause is in `@uipath/case-schema` (`parseInput` runs the disk→canvas transform with `shouldUseCaseUnifiedSchema:false`, so a flattened v20 caseplan never gets its `root` reconstructed).

The callout documents the **workaround** — pack via `uip solution pack` (which handles the v20 transform correctly, and is the recommended publish path anyway).

## Why docs-only
The underlying defect is in the (separate) `@uipath/case-schema` package, tracked as a code fix there. This PR just keeps the skill from steering an agent into the broken standalone `pack` for v20 caseplans.

Found while completing the FDE onboarding (Exercise 07, Maestro Case Management) on macOS / `staging.uipath.com` tenant FDE — a v20 caseplan that validated cleanly failed `maestro case pack` but packed fine via `uip solution pack`.


tsathkumara:
Thanks for the review — addressed both points in the latest commit:
- **(a)** Trimmed the callout to symptom + scenario + workaround; dropped the internal implementation details (`@uipath/case-schema`, `parseInput`, `shouldUseCaseUnifiedSchema`, disk→canvas) per the token-optimization rule.
- **(b)** Removed the `--dry-run` mention so we don't point at an undocumented flag.

tsathkumara:
On point (b): `--dry-run` is a real, verified flag — `uip solution pack --help` documents *"Pass --dry-run to validate the solution without producing a package"*, and I ran it successfully. The actual gap was that the `uipath-solution` skill's `pack-and-deploy.md` didn't list it, which is why it looked unverifiable. Rather than drop a valid flag, I've **documented `--dry-run` in the solution skill** (commit 196b425a) and kept the maestro-case caveat trimmed (point a).

tsathkumara:
@claude please re-review — both points from the earlier review are addressed in later commits:

- **(a) token-optimization:** trimmed the callout to symptom + scenario + workaround; dropped the `@uipath/case-schema`/`parseInput`/`shouldUseCaseUnifiedSchema` internals (commit `0279a17a`).
- **(b) `--dry-run`:** it's a valid flag — `uip solution pack --help` documents *"Pass --dry-run to validate the solution without producing a package"*. The gap was that the `uipath-solution` skill didn't list it, so I documented it there (commit `196b425a`) and removed the parenthetical from the maestro-case caveat.

tsathkumara:
Closing as obsolete — the underlying CLI bug is fixed.

Re-verified today on `uip` 1.198.0 / `@uipath/maestro-tool` 1.198.1: `uip maestro case pack ./ClaimsCase <out>` now succeeds on a v20 (unified-schema) caseplan and produces the .nupkg. The hardcoded `shouldUseCaseUnifiedSchema: false` in `parseInput` is gone — the flag is now derived from the caseplan (`caseUnifiedSchemaEnabled = root.caseUnifiedSchemaEnabled`). So the known-issue caveat this PR added would be documenting a bug that no longer exists.

The second commit here (documenting `uip solution pack --dry-run`) is still accurate and not on main, but I'm dropping it with this PR rather than carrying it separately.

PR #1516, opened 2026-06-16
test(uipath-agents): stage coded task fixtures instead of pasting files in prompts

bai-uipath:
## What

Twelve `uipath-agents/coded` coder_eval tasks embedded entire project files inside their `initial_prompt` and instructed the agent to recreate them verbatim. This moves each project into an on-disk `_fixtures/` directory, stages it at run time via `pre_run`, and rewrites the prompts to reference the staged project in plain language.

## Why

When the prompt hands over the full file contents, the agent can satisfy the task by transcription instead of by reasoning through the skill, so the test measures copy-paste rather than capability. Referencing a pre-staged project restores the skill as the thing under test and makes the prompts read like real requests. This is the first, lowest-risk slice of the broader coder_eval prompt-fitness review.

## Scope and safety

Success criteria are unchanged across all twelve tasks, so grading is identical. Fixture staging is aligned to each checker's working directory: root-based checkers receive the project files at the workspace root, and subdir-based checkers receive a matching named project directory. Antipattern and diagnose tasks keep their stated symptom; only the file dump is removed.

## Validation

All twelve tasks pass a real `coder_eval` run on Bedrock (docker driver, `skills-image`): 12/12 SUCCESS, weighted score 1.0, with every individual success criterion green. This confirms agents complete each task end-to-end through the skill and that the staged fixtures resolve correctly against both root-based and subdir-based checkers.

### DIFF
--- a/tests/tasks/uipath-agents/coded/_fixtures/bindings_sync/invoice-processor/bindings.json
+++ b/tests/tasks/uipath-agents/coded/_fixtures/bindings_sync/invoice-processor/bindings.json
@@ -1,0 +1,1 @@
+{"version": "2.0", "resources": []}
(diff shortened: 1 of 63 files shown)


andreibalas-uipath (approved):
Approved - coder_eval pipeline showed #1: 10/12 and #2: 11/12 tests passing after two subsequent runs on the updated tests: https://github.com/UiPath/skills/actions/runs/27684852694

PR #1518, opened 2026-06-16
test: remove explicit skill-load directives from task prompts

bai-uipath:
## What

Removes the explicit "load/use the `uipath-<x>` skill" directive from the prompts of 200 coder_eval task definitions across 17 skill areas. Only the `initial_prompt` text changes — success criteria, tags, run limits, and every other field are byte-for-byte untouched.

## Why

Telling the agent which skill to load short-circuits the capability the suite is meant to measure: whether the agent recognizes the task and reaches for the right skill on its own. A prompt that names the skill turns "did the agent trigger the correct skill" into a pointer-follow. Removing the directive restores skill auto-triggering as part of what each task exercises, and makes the prompts read like real user requests.

Where a directive was a self-contained sentence ("Before starting, load the `uipath-platform` skill and follow its workflow."), the whole sentence was dropped. Where it was fused with genuine task instruction or a technical hint ("…load the skill and follow its workflow — in particular the file-attachment binding pre-flight…"), only the skill-naming clause was removed and the instruction kept. A few prompts whose scenario is itself about a skill's scope (negative-guards refusal tests) keep that intrinsic reference, since it is the subject under test rather than a load hint.

## Scope

200 prompts fixed, by skill:

| count | skill |
|------:|-------|
| 54 | uipath-maestro-flow |
| 42 | uipath-platform |
| 17 | uipath-maestro-bpmn |
| 17 | uipath-maestro-case |
| 15 | uipath-data-fabric |
| 12 | uipath-api-workflow |
| 10 | uipath-review |
| 8 | uipath-test |
| 6 | uipath-mcp-servers |
| 5 | uipath-tasks |
| 3 | uipath-agents |
| 2 | uipath-automation-discovery |
| 2 | uipath-ixp |
| 2 | uipath-planner |
| 2 | uipath-rpa |
| 2 | uipath-solution |
| 1 | uipath-coded-apps |

### DIFF
--- a/tests/tasks/uipath-test/organize_testcases_into_testsets.yaml
+++ b/tests/tasks/uipath-test/organize_testcases_into_testsets.yaml
@@ -14,8 +14,6 @@ run_limits:
   max_turns: 35
 
 initial_prompt: |
-  Before starting, load the uipath-test skill and follow its workflow.
-
   I have a project called CLAIM. Organise testcases for the "disbursement"
   feature into logical test sets (e.g. happy path, edge cases, regression).
   List the testcases filtered by "disbursement", group them, and create the
--- a/tests/tasks/uipath-agents/coded/is_jira_create_issue/is_jira_create_issue.yaml
+++ b/tests/tasks/uipath-agents/coded/is_jira_create_issue/is_jira_create_issue.yaml
@@ -36,8 +36,6 @@ initial_prompt: |
   a Jira issue through Integration Service when it runs. Build it, run it
   locally, and confirm the issue actually got created in Jira.
 
-  Use the `uipath-agents` skill to figure out how.
-
   Use the Jira connection `jira-coded-eval` in the `Shared/uipath-agents`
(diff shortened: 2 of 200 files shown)


uipreliga (approved):
🚢

tests/tasks/uipath-test/organize_testcases_into_testsets.yaml
```diff
@@ -14,8 +14,6 @@ run_limits:
   max_turns: 35
 
 initial_prompt: |
-  Before starting, load the uipath-test skill and follow its workflow.
-
   I have a project called CLAIM. Organise testcases for the "disbursement"
```
ganeshborle, line 17: I think, in such case the wordings need to be fixed to at least tell the coding agent that it is the test manager project :) I will raise the followup PR accordingly.

ganeshborle (approved):

rockymadden (approved):

PR #1521, opened 2026-06-17
MST-10795 Coding agent: support edit (not just build)

charlesliu9:
feat(maestro-case): add brownfield edit-existing-case journey + edit-ops mechanics

use greenfield/brownfield split. A direct edit to an existing caseplan.json now skips the Phase 0-6 planning pipeline (satisfies
evals.json id:5) and routes to a targeted-edit journey.

- references/brownfield.md (new): edit-existing-case journey — Read-first, Common-edits routing table, after-edits validate, Phase 5/6 handoff
- SKILL.md: add greenfield-vs-brownfield routing table; fix the scope line (local edits ARE supported; only remote/deployed fetch is not); scope Rule 6  "regenerate from scratch" to greenfield only; nav row + description clause
- planning.md / implementation.md / phased-execution.md: reciprocal "editing an existing case?" cross-links to brownfield.md
- case-editing-operations.md: add Expression Prefixes, same-file Edit anchoring guidance, Replace-a-trigger-type operation, and the Operation-to-Plugin
  Quick Reference the brownfield journey routes into


song-zhao-25:
### Brownfield edits a stale local file and never pulls latest from Studio Web first

The edit journey assumes the local `caseplan.json` is current and edits it in place. It never fetches the latest from Studio Web before editing, and the only outbound SW calls are on the **publish** handoff (`uip solution resources refresh` → `uip solution upload`).

- `brownfield.md` "Read this first" lists the only CLI calls as read-only metadata fetches (`tasks describe`, `case spec`, `is describe`), `validate`, and — on handoff — `resources refresh` / `solution upload` / `debug`. No pull/download anywhere.
- `SKILL.md` scope line states: *"Not supported: fetching a remote/deployed case back to local (no remote fetch tooling)."*

**Risk:** if the case was changed in Studio Web after the local copy was made, the agent edits a stale file and `solution upload` then pushes it back — **silently overwriting the server-side changes**, with no diff, no conflict check, no warning. Most users reading "edit an existing case" mean *their deployed case in SW*, so this is the common path, not an edge case.

**The "no remote fetch tooling" premise is outdated** — the CLI already ships the verb:

```
uip solution download <solution-id> -d <dir> --extract --output json
```

> Download/export a solution from Studio Web as a `.uis` archive; `--extract` unpacks it. (`uip solution project resync` is also available for an existing local project.)

**Suggested change:** add a pull-latest step at the top of the brownfield journey, gated on where the case lives:

1. If the user is editing a case that exists in Studio Web (has a SolutionId), run `uip solution download <solution-id> --extract --output json` into the working dir **before** any edit, so edits apply to the current server state.
2. If editing a purely local project (no SolutionId), proceed as today.
3. Either way, surface a one-line freshness note in the completion report so re-publish can't silently clobber SW edits.

This closes the gap between "edit a local project" and "edit my deployed case," and corrects the `SKILL.md` scope line.

song-zhao-25:
### Two remaining issues before merge

The edit-coverage, cascade, and pull-latest work all look good now. Two things left:

---

**1. Blocker — branch is not rebased (255 commits behind `main`).** This is the only thing standing between this PR and merge-ready. It has two concrete consequences:

- **Merge conflict.** `main` still has `### Delete a connector condition rule` in `case-editing-operations.md`; this PR replaces that same span with three new sections (`Delete a task` / `Delete a condition rule` / `Delete a case-exit completion rule`). Both sides edited the same region → it will conflict.
- **`ExceptionStage` self-contradiction / regression.** `main` retired the `case-management:ExceptionStage` node type (migrated to `data.stageType: "secondary"`) — **0** occurrences in `main`'s `case-editing-operations.md`. This PR head still has **4**, and the PR-added `§ Convert a Stage to/from an Exception Stage` section itself states ExceptionStage *"is removed at schema v22 and MUST NOT be emitted."* Same file, opposite claims — the PR-added Quick Reference row (`Add a Stage / ExceptionStage`) and the `§ Add a node` heading reintroduce retired terminology.

**Fix:** rebase/merge `main`, resolve the `case-editing-operations.md` conflict, reconcile `ExceptionStage` → Primary/Secondary (`data.stageType`) in the PR-added Quick Reference + heading, then re-validate the cross-file `#anchors` (they were authored against now-stale plugin docs). Worth doing now — the drift grows daily.

---

**2. Advisory — no behavioral test.** The PR adds a substantial new edit journey but no seed-an-existing-`caseplan.json`-then-edit coder_eval task (e.g. add-stage smoke, remove-task / swap-trigger e2e). Not a gate for a doc-only journey per `.claude/rules/test-writing.md`, but it's the one substantive gap left; a follow-up modeled on `uipath-maestro-flow/hitl/quality_04_brownfield_insert.yaml` would close it.

song-zhao-25 (approved):

PR #912, opened 2026-05-21
docs(maestro-flow): add simulation commands to eval skill

AAgnihotry:
## Summary

- Documents `uip maestro flow eval simulation add/list/remove` commands shipped in [UiPath/cli#2163](https://github.com/UiPath/cli/pull/2163)
- **CAPABILITY.md**: adds simulation capability to the index and a lookup table entry pointing to the new sections
- **commands-reference.md**: full `## Simulations` section — flags table, strategy guide (Llm / Static / Mockito), `--output-schema` auto-resolution explanation, copy-pasteable examples, and Code enum rows
- **eval-sets-guide.md**: new `## Simulations on Data Points` section with ready-to-run examples for all three strategies

## Test plan

- [ ] Verify `uip maestro flow eval simulation add` example commands in commands-reference.md match the actual CLI flags (compare against `eval.ts` in cli#2163)
- [ ] Verify `--output-schema` auto-resolution description matches `FlowEvalService.addSimulation` behaviour (Llm + schema undefined → reads `.flow` file)
- [ ] Confirm Code enum entries (`FlowEvalSimulationAdd/List/Remove`) match the `Code:` values emitted in `eval.ts`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1544, opened 2026-06-17
test(uipath-maestro-flow): add simulation CRUD coverage for eval skill

mjnovice:
## Summary

Adds coder_eval coverage for the `uip maestro flow eval simulation add/list/remove` commands documented in #912 — which merged with **zero test coverage**.

New smoke task + disk-check script under a dedicated leaf directory:

- `tests/tasks/uipath-maestro-flow/evaluate/simulation/simulation_crud.yaml` — tagged `[uipath-maestro-flow, smoke, mode:build, lifecycle:generate, feature:eval]`. Modeled on the sibling `local_crud.yaml`: scaffolds a double-nested Flow project, builds an eval set + data point, then exercises all three new commands:
  - `simulation add` with `--strategy Static --mock-value`
  - `simulation add` with `--strategy Llm` + **explicit** `--output-schema`
  - `simulation list`
  - `simulation remove`
  - `--output json` discipline on every eval command
- `tests/tasks/uipath-maestro-flow/evaluate/simulation/check_simulation_crud.py` — reads the CLI-produced eval-set JSON and asserts the kept Llm sim (`agent-lookup`) persisted **and** the removed Static sim (`connector-send-email`) is gone, so add + remove must both actually mutate the file. Not gameable by self-report.

## Scope notes

- **Local CRUD only** — no `uip login`, `solution upload`, or `eval run *`. Simulations are stored inline in the data point's `simulations` array, like the other local-CRUD eval tasks.
- The Llm path passes `--output-schema` explicitly so the **auto-resolution-from-`.flow`** behavior is *not* exercised — that needs a real connector/agent node with declared outputs (flow authoring + likely a live tenant) and belongs in a follow-up integration/e2e task.

## Validation

- YAML + Python parse clean.
- `/lint-task` verdict: **Low** (single carve-out-justified prompt note; consistent with sibling `local_crud.yaml`).
- `scripts/check-cli-verbs.py` exits 0 — all four `maestro flow eval simulation*` verbs present in the catalog snapshot.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/evaluate/simulation/check_simulation_crud.py
+++ b/tests/tasks/uipath-maestro-flow/evaluate/simulation/check_simulation_crud.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python3
+"""Verify simulation CRUD produced real artifacts on disk.
+
+Beyond the `command_executed` matchers (which only prove the agent ran the
+right shell command), confirm the simulations landed in the eval-set JSON and
+that the remove actually mutated the file:
+
+  1. An eval-set JSON exists under SimEval/ with name == "Sim Set" carrying a
+     data point named "hello" in `evaluations[]`.
+  2. That data point has a non-empty `simulations` array.
+  3. The Llm simulation targeting `agent-lookup` is still present (add worked).
+  4. The Static simulation targeting `connector-send-email` is gone (the
+     `simulation remove` actually wrote back to disk, not just printed OK).
+
+Fails if the agent ran the commands but they errored, or fabricated stdout
+without producing/mutating real files.
+"""
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+PROJECT = Path("SimEval")
+KEPT = "agent-lookup"
+REMOVED = "connector-send-email"
+
+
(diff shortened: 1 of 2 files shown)


bai-uipath (approved):
**Approve once a passing run is attached** — fills a real zero-coverage gap (`eval simulation` has had none since #912), clean and faithful to `local_crud.yaml`.

- **No passing run evidence.** Body confirms parse + a `/lint-task` Low, but never asserts the task ran end-to-end — lint bot flagged this **High**. Add `Ran skill-flow-eval-simulation-crud locally and it passed.` once run.
- **Worth proving out in that run:** the sims target `connector-send-email` / `agent-lookup`, but the scaffolded flow has no such nodes. The `Llm` path passes explicit `--output-schema` to skip auto-resolution, but if `simulation add` validates node existence at all — even for `Static` — both adds fail. A real run settles it.
- **The red "Validate task schema (advisory)" is not yours** — `simulation_crud.yaml` validates clean; the red X is six pre-existing `uipath-agents` tasks the whole-tree validator trips on.
- Check script isn't gameable — asserts the `Llm` sim persisted and the `Static` sim was actually removed.

PR #1548, opened 2026-06-17
fix(ci): disable uip CLI version-sync in skills eval workflows

bai-uipath:
## What

Disable the uip CLI's auto version-sync in the GitHub Actions workflows that run coder-eval tasks: the Linux smoke gate (`smoke-skills.yml`), the Windows RPA smoke gate (`smoke-rpa-skills.yml`), and the ad-hoc / nightly runner (`run-coder-eval.yml`).

## Why

A task that runs `uip login status` triggers the CLI's post-login version-sync, which re-pins the shared `~/.uipath/config.json` to a stale CLI line and silently downgrades the CLI. Because the eval containers share one `~/.uipath` (and the Windows runs share one host CLI), that downgrade then leaks into later tasks — so some tasks run on the pinned `@alpha` CLI and some on the downgraded one. That version split is what fails smoke tasks inconsistently.

Same root cause and same fix already shipped for the nightly run (`coder_eval_uipath` `daily.sh`). The CLI treats `UIPATH_CLI_DISABLE_VERSION_SYNC` as a hard kill switch, and coder_eval already forwards it into task containers via its default `env_passthrough` allowlist, so setting it per job closes the gap across all the skills-repo eval entry points.

Background — public thread on the `uip login` silent-downgrade issue: https://uipath-product.slack.com/archives/C0A3Z950TNW/p1781096211538789

### DIFF
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -133,6 +133,14 @@ jobs:
     # hiccups, etc.).
     timeout-minutes: 330
     name: Run coder-eval (Linux)
+    # Disable uip CLI version-sync for the whole job. A task's `uip login
+    # status` otherwise re-pins ~/.uipath/config.json to a stale CLI line;
+    # with nightly.yaml's shared `~/.uipath:/.uipath:rw` mount that downgrade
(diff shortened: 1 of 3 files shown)


uipreliga (approved):
🚢

PR #1551, opened 2026-06-17
test(uipath-maestro-flow): add eval task for inline agent node

mjnovice:
## What

Adds an integration coder-eval task — `skill-flow-eval-inline-agent` — that wires evaluation scaffolding against an **inline agent node** (`uipath.agent.autonomous`) in a UiPath Flow.

New files:
- `tests/tasks/uipath-maestro-flow/evaluate/inline_agent_eval/inline_agent_eval.yaml`
- `tests/tasks/uipath-maestro-flow/evaluate/inline_agent_eval/check_inline_agent_eval.py`

## Why

Closes a coverage gap in the `uipath-maestro-flow` evaluate capability. Existing eval tasks (`eval_run`, `local_crud`) target deterministic script nodes with `exact-match` evaluators. None exercised the **Author(inline-agent) → Evaluate** handoff or the evaluator-type decision for a non-deterministic agent output.

## What it guards

- Scaffolding an inline agent with `uip agent init --inline-in-flow` and authoring the `uipath.agent.autonomous` node.
- Picking an `llm-judge-output` evaluator for the non-deterministic agent output — **not** a deterministic `exact-match`/`contains` type (enforced via a `command_not_executed` guard). The prompt describes the output semantics but does not name the answer, so the agent must map semantics → type itself.
- Creating an eval set + data point without leaking the evaluator display name into `--evaluators`.

Local CRUD only — no login, upload, or eval run. The check script asserts real CLI-produced artifacts on disk (inline `agent.json`, an evaluator with `evaluatorTypeId: uipath-llm-judge-output-semantic-similarity` and a `model`, and an eval set with a well-formed data point) rather than agent self-reports.

## Validation

- `/lint-task`: clean — one documented-carve-out Low (validate-only at integration tier, deliberately local-only per the description; consistent with sibling local-CRUD eval tasks).
- CLI verb reachability: all verbs resolve (only Info-level dynamic-pattern findings).
- Tags conform to the closed taxonomy; `run_limits` top-level; no `@uipath/cli` in sandbox.

> **Note:** a `coder-eval` passing-run claim is not yet attached. Will add once the task is run.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/evaluate/inline_agent_eval/check_inline_agent_eval.py
+++ b/tests/tasks/uipath-maestro-flow/evaluate/inline_agent_eval/check_inline_agent_eval.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python3
+"""Verify the inline-agent eval scaffold produced real artifacts on disk.
+
+This task wires evaluation scaffolding against an INLINE agent node
+(`uipath.agent.autonomous`), whose output is non-deterministic LLM text — so
+the correct evaluator is an LLM judge, NOT the `exact-match` the deterministic
+script-node eval tasks use. Beyond the `command_executed` matchers (which only
+prove the agent ran the right shell command), assert the side effects:
+
+  1. An inline agent.json exists under TriageEval/TriageEval/<uuid>/agent.json
+     (the inline agent dir is a UUID, so glob it; skip generated
+     .agent-builder/). Proves the eval target is a real inline agent.
+  2. An evaluator JSON exists with evaluatorTypeId ==
+     "uipath-llm-judge-output-semantic-similarity" (the `llm-judge-output`
+     internal id) carrying a non-empty `model` — the right choice for a
+     non-deterministic agent output, and the model the LLM gateway requires.
+  3. An eval-set JSON exists with at least one data point in `evaluations[]`
+     carrying non-empty `inputs` and an expected-output field.
+
+These checks fail if the agent ran the commands but they errored, picked a
+deterministic evaluator for the non-deterministic agent, or fabricated stdout
+without producing real files. Reads only source files — no tenant calls.
+"""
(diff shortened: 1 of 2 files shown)


bai-uipath (approved):
**Trim the duplication and attach a passing run before merge** — the inline-agent → llm-judge seam is a real gap, but most of the task re-tests ground that already exists.

- **No passing run evidence.** Body says it outright ("not yet attached"); lint bot flagged **High**. This is also the most multi-step task in the suite, so the riskiest to leave unproven. Run it and attach the claim.
- **The authoring half is a near-clone of `smoke/inline_agent_robust.yaml`** — same EmailTriage→TriageEval classify-email→category+priority scenario, same production bar, same agent.json glob, same `flow validate` gate. The `agent init --inline-in-flow` / `file_exists` / `file_contains "uipath.agent.autonomous"` / `validate` criteria are verbatim copies; the build is also covered by the three `billing_*` e2e tasks.
- **The eval tail overlaps too** — llm-judge-vs-deterministic is already `evaluator_type_choice.yaml` (Goal A); eval-set/data-point CRUD + local guards are already `local_crud.yaml`. Net: this ≈ `inline_agent_robust` + `local_crud` + `evaluator_type_choice` recombined, and it even grades the agent *more weakly* than robust (only checks agent.json exists).
- **Suggestion:** scaffold a minimal inline agent and concentrate the checks on the actual new thing — the evaluator targets the agent output and is `llm-judge-output`, not `exact-match`/`contains`. Or reuse `_shared/check_inline_agent.py` and cut the criteria that duplicate `inline_agent_robust`.
- Minor: tagged `integration` but does purely local CRUD; sibling `evaluator_type_choice.yaml` is `smoke`, so as-is it won't run on the PR gate. Consider `smoke`. (Red advisory check isn't yours — `inline_agent_eval.yaml` validates clean.)

PR #1556, opened 2026-06-18
test(uipath-troubleshoot): de-flake UIA replay tests (CLI via command_executed, simulator declines source edits)

dmorosanu:
## Problem

The three UIA faithful-replay troubleshoot tests failed in the **2026-06-18 nightly** (`2026-06-18_04-06-29`) — all three at once, with intermittent single failures on 06-12 and 06-15. The skill diagnosed every case **correctly in turn 1**; the failures are test-harness flakiness, not a skill regression:

1. **Judge blindness.** The `llm_judge` grades only the final agent turn and only the *main* agent's tool calls. The skill runs `uip or jobs get` inside sub-agents, invisible to the judge — so a Dimension A "must show CLI interrogation in tool calls" requirement mis-fired and capped correct runs at ~0.2. (Proven: `uia-alter-if-disabled`'s separate `command_executed` check scored 1.0 on the same run where the main-agent `uip` count was 0.)
2. **Simulator approved the source edit.** The simulator approved the agent's offer to apply the fix to the workflow `.xaml`, then approved deleting `.local/investigations/`. The agent's final graded turn became a file-edit / cleanup message instead of the diagnosis → judge 0.0.
3. `uia-alter-if-disabled` also graded a `file_exists state.json` criterion and read `state.json` in the judge — bookkeeping the agent legitimately deletes at session end.

## Fix (test-only — no skill changes)

- Verify CLI interrogation via a `command_executed` criterion on `uip or jobs get <job-key>` (the harness command checker sees sub-agent calls). Set judge `include_tool_calls: false` and reword Dimension A to grade the presented diagnosis.
- Simulator now **declines source-file edits** (applies the fix in Studio itself), so the agent closes recommendation-only and the diagnosis stays in the graded turn.
- Drop the brittle `file_exists state.json` criterion and the `state.json` / `hypotheses.json` judge attachment from `uia-alter-if-disabled`.

This aligns all three with `tests/CLAUDE.md` ("grade on presentation, not internal investigation state") and with the pattern `uia-alter-if-disabled` already used for the CLI check.

### DIFF
--- a/tests/tasks/uipath-troubleshoot/uia-alter-if-disabled/task.yaml
+++ b/tests/tasks/uipath-troubleshoot/uia-alter-if-disabled/task.yaml
@@ -162,8 +162,9 @@ simulation:
     - "Keep replies short (one sentence or pick a numbered option)."
     - "Never instruct the agent to stop, abort, or skip phases."
     # The llm_judge sees only the final turn. If the simulator approves applying
-    # the fix, the agent's edit lands in the last turn and pushes the diagnosis
-    # out of the graded turn. Decline source-file edits so the agent closes
-    # recommendation-only.
+    # the fix OR deleting .local/investigations/, that confirmation lands in the
+    # last turn and pushes the diagnosis out of the graded turn. Decline both so
+    # the agent closes with the diagnosis summary.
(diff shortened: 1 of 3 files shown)


dmorosanu:
dup in other PR

PR #1445, opened 2026-06-12
test(connector): Test Manager connector evals (API Workflow + Maestro Flow + RPA)

nyamathshaik:
## What

The `uipath-test` eval covers only the **`uip tm` CLI** surface. But the Test Manager **Integration Service connector** (`uipath-uipath-testmanager`, ~20 operations: Project / Requirement / TestCase / TestSet / TestExecution / attachments / assertions …) is also usable as activities **inside workflows** — and had **zero** coverage on any of them, unlike Data Fabric (which has Flow + RPA connector test plans).

This adds **`skip: true`, tenant-gated skeleton tasks** for all three connector surfaces, scenario = *"create a Test Case via the Test Manager connector"*:

| Surface | Path | Drives |
|---|---|---|
| API Workflow | `tests/tasks/uipath-api-workflow/connector/testmanager_create_testcase/` | `registry resolve` → `is connections list`/`ping` → `registry stub --connection-id` → `is resources describe` → `api-workflow validate` + Workflow.json shape check |
| Maestro Flow | `tests/tasks/uipath-maestro-flow/connector_features/testmanager_create_testcase/` | `maestro flow registry search` → connector `node add` → `flow validate` + flow-debug check |
| RPA (coded) | `tests/tasks/uipath-rpa/connector/testmanager_create_testcase/` | `uip is` connector discovery → coded `UiPath.IntegrationService.Activities` call → `uip rpa build` + project shape check |

Each follows the verified connector-task patterns in the repo (Jira `is_jira_create_issue`, `drive_to_slack`, the billing Flow tasks). The `command_executed` criteria grade the **connector-agnostic** discovery/authoring flow and assert the `uipath-uipath-testmanager` connectorKey — correct regardless of the exact operation slug.

## Validation done locally

- ✅ `coder-eval plan` — all 3 valid (schema gate)
- ✅ `scripts/check-cli-verbs.py` — 0 High, 0 Medium (every referenced `uip` verb exists in the catalog)
- ✅ `py_compile` on all 3 check scripts

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1563, opened 2026-06-18
test(connector): data-grounded TM connector eval prototypes (CRUD + results)

nyamathshaik:
## Data-grounded TM connector eval — prototype (proposal)

Today's connector coverage tasks grade **invocation/authoring** (did the agent call/wire the activity). This proposes grading **behavior** — that the connector actually returned the right data. Two reusable patterns:

### 1. Self-seeding CRUD (no fixture) — ✅ validated SUCCESS 1.000 on codereval
`testmanager_crud_grounded/` (`seed.py` + task + `check.py`)
- `seed.py` generates a unique name; agent creates the TestCase via the connector, reads it back, writes `result.json`; `check.py` asserts **`retrieved_name == created (seeded) name`** — only passable by a real create→read round-trip. Self-cleaning.
- Validated end-to-end against codereval (HEALTH): agent ran `uip is resources run create/get "uipath-uipath-testmanager" "TestCase"`, round-trip verified, 3/3.

### 2. Results activity (needs a fixture) — pattern, `skip:true`
`testmanager_results_grounded/` (task + `check.py`)
- Agent fetches **Get Assertions** for a known execution; `check.py` asserts returned count ≥ the fixture's expected minimum.
- Results activities can't self-seed (they read byproducts of a prior execution), so this needs a **maintained seeded execution** (TODO: pin `ExecutionId` + `EXPECTED_MIN`). Reusable for robot logs / test steps / test case logs.

### Finding
`command_executed` patterns are brittle — the agent quotes args (`create "uipath-uipath-testmanager" "TestCase"`), so patterns must allow optional quotes. The data-grounded `check.py` passed regardless of quoting — **behavior checks are more robust than command-shape checks.** (Same quote-tolerance fix likely makes the #1445 IS tasks reliably green.)

### Proposed rollout
- CRUD groups → self-seeding grounded (cheap, no fixture).
- Results/read groups → fixture-backed grounded (needs one seeded execution + an owner).

Separate from #1445 (which is the invocation/authoring coverage). Draft for discussion.

### DIFF
--- a/tests/tasks/uipath-platform/integration-service/testmanager_crud_grounded/check.py
+++ b/tests/tasks/uipath-platform/integration-service/testmanager_crud_grounded/check.py
@@ -1,0 +1,43 @@
+#!/usr/bin/env python3
+"""Data-grounded check for the Test Manager connector CRUD task.
+
+Proves the agent did a REAL create->read round-trip (not just invoked the
+activity): the value retrieved by Get must equal the unique seeded value the
+agent created. A cheating/placeholder run can't satisfy this, because the Get
+response only carries the created name if the record was actually written.
+
+Reads:
+  seed.json   (pre_run) -> {"name": "<unique>", ...}
+  result.json (agent)   -> {"created_name", "retrieved_name", "id"}
+"""
+import json
+import sys
+
+
+def fail(msg):
+    print(f"FAIL: {msg}")
+    sys.exit(1)
+
+
+def main():
+    try:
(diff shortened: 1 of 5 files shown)


nyamathshaik:
Consolidated into #1445 — the data-grounded results pattern (testmanager_results_grounded) was moved there, and the crud_grounded POC is superseded by the data-grounded CRUD lifecycle tasks in #1445. Closing to keep everything in one PR.

PR #1583, opened 2026-06-19
docs(uipath-data-fabric): document folder scope, FILE-write silent-st…

AditiGoyalUipath:
…rip, and mandatory scope-prompt flow

- SKILL.md Rule 6: rewrite FILE-write rule — platform silently strips FILE values from records insert/update/import payloads (Result: Success, column unchanged). receipt:null does NOT clear; receipt:"<uuid>" does NOT swap.
- SKILL.md Rule 19: add Mandatory scope-prompt flow — when scope is unspecified, ask via AskUserQuestion with two dropdowns (tenant vs folder; then accessible folders pre-fetched from `uip or folders list`). Never default silently.
- entity-schema.md / file-attachments.md / records-query.md: align FILE-write wording with the rule-shaped phrasing in SKILL.md ("Never put FILE-typed key in payload. Expected behavior: silently stripped. Use files upload to attach or replace; files delete to clear.")
- file-attachments.md: new section documenting what records get returns for a FILE field — stable per-record-per-field UUID handle, not a content id; filename is not in any read response.
- choice-sets.md / bulk-import.md / records-query.md: --folder-key fan-out across every records / files / choice-set verb (CLI ≥ 1.197.0), and --yes/--reason gating with --confirm as deprecated alias.


skills/uipath-data-fabric/references/entity-schema.md
```diff
@@ -270,7 +316,7 @@ Response: `{ Code: "EntityUpdated", Data: { Id, RemovedFields: ["<name>"], Reaso
 |-----------|--------|
 | Change a field's data type | Not supported — type is fixed at creation and cannot be changed via `updateFields` |
 | Field name matching a SQL / language keyword | API returns `RESERVED_LANGUAGE_KEYWORDS` — rename before retrying (see Name Validation above) |
-| Upload a file to a `FILE` field via `uip df files upload` | CLI insists on `referenceEntityId`/`referenceFieldId` at field create, then uploads fail with *"Relationship violation"* against arbitrary targets. No public attachment-storage entity is documented. Treat FILE upload as unusable via CLI until the target-entity contract is clarified — surface the gap to the user; don't attempt. |
+| Write a `FILE` value through `records insert` / `records update` / `records import` | Do not attempt. Expected behavior: the platform silently strips the value, returns `Result: Success`, FILE column unchanged. Use `files upload` to attach/replace, `files delete` to clear. Never `records update receipt:null`. See Rule 6 and [FILE Fields](#file-fields). |
```
SuphalAthlur: > | Write a `FILE` value through `records insert` / `records update` / `records import` | Do not attempt.

A bit unclear, are we saying this is an anti-pattern to avoid?

skills/uipath-data-fabric/references/file-attachments.md
```diff
@@ -41,10 +46,23 @@ Response: `{ Code: "FileDownloaded", Data: { EntityId, RecordId, FieldName, Outp
...
+  --yes --reason "<why>" \
+  --output json
 ```
 
-Response: `{ Code: "FileDeleted", Data: { EntityId, RecordId, FieldName } }`
+Response: `{ Code: "FileDeleted", Data: { EntityId, RecordId, FieldName, Reason } }` — `Reason` echoes the `--reason` value.
+
+## What `records get` returns for a FILE field
+
+`records get` returns the FILE field as a UUID string, or omits it / returns `null` when no file is attached. This UUID is a **stable per-record-per-field handle**. Treat it as read-only metadata:
```
SuphalAthlur: This shape will depend on the record level depth I believe, if you set it to a higher value you will get an object with file metadata (not a plain UUID), could you please check with Postman/curl?

skills/uipath-data-fabric/references/records-query.md
```diff
@@ -137,7 +151,7 @@ uip df records query <entity-id> \
   --output json
 ```
 
-Response: `Data.Records` is a single row — `[{ "total": 250 }]`.
+Response: `Data.Items` is a one-row array — e.g. `[{ "Total": 250 }]`. The server **PascalCases your alias** in the response (`alias: "total"` → key `"Total"`). Read by the PascalCase key when parsing.
```
SuphalAthlur, line 154: > The server **PascalCases your alias** in the response (`alias: "total"` → key `"Total"`). Read by the PascalCase key when parsing.

Consider surfacing this as one of the key rules at the top

skills/uipath-data-fabric/SKILL.md
```diff
@@ -89,6 +90,26 @@ Respond that the operation is not supported. Do not try to work around it.
...
+
+    Cache the folder list within the turn to avoid refetching. Echo the chosen scope back in the next message before any irreversible call. Scope persists across follow-up turns unless the user switches. Tenant prompting is rarely needed — `uip login status` shows the active tenant; only call `uip login tenant set <tenant>` after the user explicitly asks for a different one.
+
+    **Bypass clauses — skip the AskUserQuestion flow when ANY of these hold.** The agent must still announce the chosen scope (one line) in the response so the user can correct.
+    - The prompt explicitly says some variant of *"do not ask"*, *"do not pause"*, *"no approval / confirmation / feedback needed"*, or *"proceed without confirmation"* — proceed at **tenant level** unless folder context is mentioned inline.
+    - The prompt names a folder inline (*"in the Shared folder"*, *"in folder X"*, *"--folder-key <guid>"*, *"in personal workspace"*) — proceed with that folder; do **not** ask which folder.
```
SuphalAthlur, line 104: Could we also make the agent pull up all folder-scoped entities and do an intent match across the board and surface matching folder scoped entities for the user to choose from?
Might help users who don't remember where their folder scoped entity is.

skills/uipath-data-fabric/references/records-query.md
```diff
@@ -217,6 +232,22 @@ uip df records insert <entity-id> \
 
 Display labels, choice-value UUIDs, and non-UUID relationship values are rejected — resolve first. Reads echo the same shape.
 
+### FILE fields — never write through insert/update
+
+Never include a FILE-typed key in `records insert` or `records update` payload (SKILL.md Rule 6). Expected behavior: the platform silently strips FILE values — paths, base64 blobs, filenames, UUIDs, and `null` — and returns `Result: Success` with the FILE column unchanged. Do not interpret Success as "the file changed." `records update receipt:null` does **not** clear. `records update receipt:"<uuid>"` does **not** swap. Required write path:
```
SuphalAthlur: Can we also provide a reference to the file attachment md here? so the agent knows what to do to add a file.

skills/uipath-data-fabric/references/file-attachments.md
```diff
@@ -13,16 +13,20 @@ The FILE field itself must still be created via the CLI before the UI can upload
 The entity must have a field configured for file storage. File fields are defined in the entity schema.
 Use `uip df entities get <entity-id> --output json` to identify file-type fields. A correctly-defined FILE field shows `FieldDataType.Name: "FILE"`, `FieldDisplayType: "File"`, `IsForeignKey: true`, and `ReferenceEntity.Name == "EntityAttachment"`.
```
SuphalAthlur: Please evaluate if it would help to mention that a record MUST exist before a file upload is attempted, under pre-reqs

skills/uipath-data-fabric/references/entity-schema.md
SuphalAthlur: Also should we be adding this here? you anyway cover record level ops under records-query right? IMO it would be better to keep this md purely for schema level details.

SuphalAthlur (approved):
And generally, isn't the critical rules section stretching too much?

Should we keep only the important rules here and keep the rest in `references/`

We can include references to the reference md, otherwise a plethora of rules might take away from the emphasis we are trying to provide to the coding agent via this section.

skills/uipath-data-fabric/SKILL.md
AditiGoyalUipath: We have all tenant + folder scoped entities method. If user doesn't remember then they can cli can list all tenant + folder entites . User can figure out using that.

skills/uipath-data-fabric/references/entity-schema.md
AditiGoyalUipath: make sense

PR #1591, opened 2026-06-19
feat(coded-apps): NLP-driven dashboard generation (compiler architecture) + coder-eval suite

NishankSiddharth:
https://uipath.atlassian.net/browse/PLT-104174

## What this is

The `uipath-coded-apps` skill gains a natural-language **dashboard generation** capability. A UiPath admin / CoE lead describes the metrics they want in plain English — *"agent health, error-rate trend for the week, top agents by usage"* — and the skill generates a live, deployable React dashboard wired to their tenant's data through the `@uipath/uipath-typescript` SDK. No hand-coding.

## How it works (end-to-end flow)

1. **Intent routing.** On a dashboard request the skill classifies intent — **build / edit / deploy** — and fires a parallel "blast" (load the relevant docs, check for an existing dashboard via `state.json`, kick off a background `uip login status`, pre-warm dependencies) before showing anything.
2. **Plan — a text-only gate.** It resolves every requested metric to a concrete SDK method, refuses impossible metrics inline (with a suggested alternative), and presents a plain-language plan. This turn is **pure text, zero tool calls**, ending with a confirm/change affordance. (A question/option tool fired in the same turn reliably suppresses the plan rendering, so the gate is deliberately text-only; setup questions like OAuth come *after* approval.)
3. **Compile.** On approval the skill authors two things and compiles them:
   - **`intent.json`** — pure metadata (`schemaVersion 2`): dashboard name, time range, tenant + OAuth config, and a list of metric descriptors. No embedded code.
   - **`metrics/<name>.ts`** — one real TypeScript module per metric, each exporting `fetchData(sdk, getToken)` (plus optional drill-down fetchers). Each is written from curated SDK reference docs and cross-checked against the documented example response.

   The build engine then compiles these in two stages — an isolated typecheck of the metric modules, then generation of widgets, routes, and detail views, then a full `tsc --noEmit` backstop — into a Vite + React + Tailwind + Recharts app.
4. **Run / deploy.** The app runs locally behind OAuth-PKCE login; a deploy flow packs, publishes, and deploys it to Automation Cloud as a coded web app, tracking version/routing state across deploys.

## Why it's built this way

- **Compiler model — metric modules, not embedded code.** Because each metric is a real `.ts` module, the generated data layer is type-checked end-to-end. The classic "compiles green but returns zero rows" failure is caught by a deterministic cross-check of each module against its documented SDK example response, and type errors are repaired in a bounded retry loop.
- **Pure-metadata `intent.json`.** Keeping the spec declarative lets the engine own all code generation (widgets, routes, drill-downs). Edits, rebuilds, and version upgrades regenerate the disposable layer without clobbering authored logic.
- **Correctness via curated SDK docs, not a live probe.** Every documented method ships a realistic example response; the author validates the fields and values it depends on against that example — no runtime, auth, or rate-limit cost.
- **Read-only by construction.** Dashboards never mutate state; tier resolution forbids any write method from appearing in metric code.
- **Governance gating.** Generic governance intent ("policy denials", "allow/deny verdicts") routes to the out-of-box Insights-API metrics; only explicit runtime-compliance / standard-clause / rule-violation intent opens the gate to the interim trace-derived metrics. The two families are kept strictly separate so neither regresses into the other.
- **Best-effort subagent.** The build prefers to run inside a host sub-task (so the main thread stays clean) but falls back to an identical inline build when no sub-task mechanism exists — subagents are an optimization, never a hard requirement.

## How it's tested

The coder-eval suite exercises the skill exactly as a user would invoke it: a coding agent loads the skill in a sandbox from a plain prompt, follows the workflow to produce real files, and **deterministic criteria inspect the artifacts** (`intent.json`, metric modules, generated widgets, `.env.local` contract, `state.json`) — never a self-written summary. A shared validator (`_shared/check_dashboard.py`) auto-locates the generated project (the skill scaffolds into a per-dashboard `<routingName>/` subdirectory) and checks the full compiler-model structure; the e2e gold gate additionally runs `tsc --noEmit` over the generated app.

**Tiers** (per the repo taxonomy): **smoke** = fast PR-gate checks that need no full build (40-turn budget, every PR); **integration** + **e2e** = full inline builds, run nightly (200-turn budget). **Modes** (Coding-Agents scorecard): **build** (generate/edit), **operate** (deploy/run), **diagnose** (investigate/fix).

### Test inventory (17 tasks)

| Test | Tier | Mode | What it verifies |
|------|------|------|------------------|
| `dashboard_plan_gate` | smoke | build | The approval gate: agent presents a plan and writes **no files / runs no `npm ci`** before the user confirms. |
| `dashboard_plan_before_question` | smoke | build | Plan response is **pure text** — no `AskUserQuestion`, no `intent.json`, no `npm ci` before approval; OAuth/setup questions fire only *after*. |
| `dashboard_disambiguate` | smoke | build | An ambiguous prompt makes the agent **ask a clarifying question and halt** — no premature scaffold (`package.json` not created). |
| `dashboard_deploy_smoke` | smoke | operate | Deploy command sequence **pack → publish → deploy** in order; `-n` is the friendly **display name** (not the routing slug); never `-t Action` (dashboards are web apps). |
| `auth_no_manual_tokens` | smoke | validate | Coded-apps auth (non-dashboard): agent uses **CLI-managed auth** and refuses to pass/paste manual access tokens. |
| `dashboard_scaffold` | integration | build | A 2-widget build produces the **compiler-model shape**: pure-metadata `intent.json` v2, metric modules exporting `fetchData`, generated widgets, and the 6-var OAuth-PKCE `.env.local` contract. |
| `dashboard_full_e2e` | e2e | build | Full build of a 4-widget agent-health dashboard (KPI + area + line + table) with real SDK calls (`getAll` + `getConsumptionTimeline` + `getErrors`), named time constants, `state.json`, and a clean **`tsc --noEmit` gold gate**. |
| `dashboard_agent_job_classification` | integration | build | Agent metrics derived from jobs use the raw OData field **`ProcessType eq 'Agent'`**; forbids `sourceType` and filtering on the SDK-mapped `processName`. |
| `dashboard_refuse_invocation_timeline` | integration | build | An impossible "invocation volume over time" is **substituted** with the documented Jobs run-volume trend; forbids fabricating a non-existent `getInvocation*` SDK method. |
| `dashboard_gov_gated_routing` | integration | build | Explicit runtime-compliance / standard-clause intent **opens the governance gate** → trace-derived metrics (`parseGovernanceSpans` + `Traces.getById`). |
| `dashboard_gov_generic_no_regression` | integration | build | Generic "governance/policy" intent routes to **Insights-API** metrics (`getPolicyTraces`/`getOperationSummary`) and **does not regress** into the trace-derived family. |
| `dashboard_rowlink_clickable` | integration | build | A clickable table wires a non-empty `ROW_LINK_KEY` + `onRowClick` handler and generates a `*DetailView.tsx` drill-down page. |
| `dashboard_edit_change_widget` | integration | build | **CHANGE** op via the incremental editor: builds a widget, then changes its display type through an `edit-intent.json` CHANGE — regenerates in place, reuses the kit (no re-scaffold). |
| `dashboard_edit_remove_widget` | integration | build | **REMOVE** op via the incremental editor: builds 2 widgets, removes 1; asserts exactly one metric module remains and the kit was reused. |
| `dashboard_incremental_add` | e2e | build | **ADD** op (full edit lifecycle): builds a 1-widget dashboard, then adds a 2nd via `edit-intent.json`, reusing the already-extracted kit. |
| `dashboard_diagnose_broken_metric` | integration | diagnose | **Validate→read-error→fix loop**: a seeded metric calls a non-existent SDK method; the agent reads the `METRICS_RETRY`/typecheck error and fixes the built module to the real `getAll`. |
| `dashboard_deploy_state` | integration | operate | **Fresh deploy** path (`systemName` empty in `state.json`): resolves the folder and deploys without a version bump. |

**Tier split:** 5 smoke · 10 integration · 2 e2e. **Mode coverage:** build, operate, and diagnose all represented.

Generated with Claude Code


NishankSiddharth:
@claude  please review the PR again earlier files diff was not correct.

NishankSiddharth:
Coder eval runs (100% success rate):

1. https://github.com/UiPath/skills/actions/runs/27944424072/job/82685486528
2. https://github.com/UiPath/skills/actions/runs/27948207563/job/82698258314
3. https://github.com/UiPath/skills/actions/runs/27955592902/job/82723141841
image

deepeshrai-tech:
build script is too big , try refactoring it into seprate modules

skills/uipath-coded-apps/references/sdk/governance-traces.md
```diff
@@ -0,0 +1,235 @@
...
+const MAX_RUNS = 15
+const HOOKS = ['BEFORE_AGENT', 'BEFORE_MODEL', 'AFTER_MODEL', 'AFTER_AGENT'] as const
+
+// Main table = last 15 agent RUNS (job runs), each summarized from its trace.
+export const fetchData: MetricFn = async (sdk) => {
+  const { Jobs } = await import('@uipath/uipath-typescript/jobs')
+  const { Traces } = await import('@uipath/uipath-typescript/traces')
+  const jobs = (await new Jobs(sdk as never).getAll(
+    { filter: "ProcessType eq 'Agent'", orderby: 'CreationTime desc', pageSize: 50 }))?.items ?? []
+  const rows: any[] = []
```
deepeshrai-tech: we should avoid use any type, see for other places too

skills/uipath-coded-apps/SKILL.md
```diff
@@ -1,7 +1,8 @@
 ---
 name: uipath-coded-apps
-description: "Always invoke for `app.config.json` or `action-schema.json` files. UiPath Coded Web Apps & Coded Action Apps via `uip codedapp` and `@uipath/uipath-typescript` SDK. Scaffold, build, debug, deploy. For .cs/XAML→uipath-rpa, Python→uipath-agents."
-allowed-tools: Bash, Read, Write, Edit, Glob, Grep, AskUserQuestion
+description: "UiPath Coded Apps — scaffold, build, run, and deploy Coded Web Apps and Coded Action Apps: React/TypeScript apps that call UiPath Cloud APIs via the `@uipath/uipath-typescript` SDK and ship to Automation Cloud (push/pull to Studio Web, pack, publish, deploy, OAuth-PKCE). Also generates NLP-driven dashboards: a plain-language request — agent health, error rate, invocation volume, latency, KPIs, governance violations, observability — becomes a live, deployable analytics dashboard wired to tenant data through the Insights real-time API, with incremental edit and upgrade flows. For RPA workflows→uipath-rpa, Python agents→uipath-agents, Maestro flows→uipath-maestro-flow, solution packaging→uipath-solution."
```
Raina451: > Also generates NLP-driven dashboards: a plain-language request — agent health, error rate, invocation volume, latency, KPIs, governance violations, observability — becomes a live, deployable analytics dashboard wired to tenant data through the Insights real-time API, with incremental edit and upgrade flows

this is too detailed descriptive. can we just have one line for governance dashboards here?

> For RPA workflows→uipath-rpa, Python agents→uipath-agents, Maestro flows→uipath-maestro-flow, solution packaging→uipath-solution."

so we need this?

skills/uipath-coded-apps/SKILL.md
```diff
@@ -1,7 +1,8 @@
...
+when_to_use: "User wants to scaffold, build, push/pull, pack, publish, or deploy a Coded Web App or Coded Action App, or use the `@uipath/uipath-typescript` SDK inside one. Also dashboard requests: 'build me a dashboard', 'show agent health / error rate / KPIs / governance violations', 'generate an analytics or observability dashboard', or edit an existing one (add/remove/change a widget, change time range, deploy). For RPA workflows→uipath-rpa; Python agents→uipath-agents; Maestro flows→uipath-maestro-flow."
```
Raina451: > 'build me a dashboard'
this is generic, will agent be able to segregate when to build normal dashboard coded app vs when governance dashboard?

deepeshrai-tech:
for governance-dashboard-starter-kit.zip in source control can we do the following?

1. Store kit content uncompressed in git and zip it in CI (better diffability, no runtime extractor dependency).
2. Ship a tiny Node-based zip-extraction fallback in the script so the OS dependency is optional.
the test locateWidgetTemplates already iterates over four extractors as fallbacks so approach seems fragile. Worth fixing at the source.

skills/uipath-coded-apps/SKILL.md
Raina451: same do we need this: 

> For RPA workflows→uipath-rpa; Python agents→uipath-agents; Maestro flows→uipath-maestro-flow."

skills/uipath-coded-apps/SKILL.md
```diff
@@ -17,6 +18,8 @@ Build, debug, and deploy UiPath Coded Web Applications and Coded Action Apps usi
 - User wants to **push/pull source** between local and Studio Web
 - User wants to use the `@uipath/uipath-typescript` SDK from a coded app
 - User wants to run the **full pipeline** (build → pack → publish → deploy)
+- User wants to **generate a dashboard** from a natural-language description — e.g. "build me a dashboard showing agent health, error rates, invocation volume, latency, active agents, KPIs, governance metrics, or consumption trends"
```
Raina451: should we rephrase to "*generate a aganet monitoring related dashboard" ?

skills/uipath-coded-apps/SKILL.md
```diff
@@ -29,7 +32,7 @@ Build, debug, and deploy UiPath Coded Web Applications and Coded Action Apps usi
 
 ## Critical Rules
 
-1. **Identify the app type before doing anything else.** Ask: *"Are you building a **Coded Web App** (custom frontend deployed to UiPath Cloud) or a **Coded Action App** (form for Action Center human task reviews)?"* The two paths diverge on scaffolding, redirect URI, and publish flag — do not guess.
+1. **Identify the app type before doing anything else.** Ask as a structured choice (Rule 17): **Coded Web App** — custom frontend deployed to UiPath Cloud · **Coded Action App** — form for Action Center human task reviews. The two paths diverge on scaffolding, redirect URI, and publish flag — do not guess.
```
Raina451: question: governance dashboards are built (sdk side) and deployed with a bit diff configuration right? so shouldnt it be better if agent upfront clears if they want a governance dashboard app for admin portal?
as today we already have codedapp vs coded action app ?

my concern is, there are usecases where users just ask to build a dashboard coded app, and even in that dashboard they can ask to show agent health etc, and not being a goiverance dashboard app, so i think its better to clear upfront from user?

skills/uipath-coded-apps/SKILL.md
```diff
@@ -44,8 +47,26 @@ Build, debug, and deploy UiPath Coded Web Applications and Coded Action Apps usi
...
-16. **When adding any new SDK method call, verify `VITE_UIPATH_SCOPE` already includes the required scope.** Write operations, action methods (`Jobs.stop`, `Tasks.complete`, `ProcessInstances.cancel`, etc.), or first use of a new service typically need broader scopes than read-only flows. Mismatched scopes fail silently with `401` / `403` on the first call. See [oauth-scopes.md](references/oauth-scopes.md) for the per-method scope table.
+16. **When adding any new SDK method call, verify the configured OAuth scope already includes the required scope** (dashboards: the `scope` key in `uipath.json`; web apps: `VITE_UIPATH_SCOPE`). Write operations, action methods (`Jobs.stop`, `Tasks.complete`, `ProcessInstances.cancel`, etc.), or first use of a new service typically need broader scopes than read-only flows. Mismatched scopes fail silently with `401` / `403` on the first call. See [oauth-scopes.md](references/oauth-scopes.md) for the per-method scope table.
```
Raina451, line 50: @amrit-agarwal-1 just a note, we would need to change in your pr once this is merged.

> (dashboards: the `scope` key in `uipath.json`; web apps: `VITE_UIPATH_SCOPE`).

skills/uipath-coded-apps/SKILL.md
Raina451: > (Rule 17):

do you mean rule 18 here?

tests/tasks/activation/uipath-coded-apps.jsonl
Raina451, line 1: @ninja-shreyash would you prefer tasks to be in this pr? or as part of your onboarding you can add this ?

skills/uipath-coded-apps/references/sdk/agents.md
```diff
@@ -0,0 +1,239 @@
...
+return [{ value: r.currentPeriodSummary.successRate, previous: r.lookbackPeriodSummary?.successRate }]
+```
+
+```typescript
+// Total Agent Units consumed, with delta (kpi-card) — ≥ 1.5.0
+const { Agents } = await import('@uipath/uipath-typescript/agents')
+const r = await new Agents(sdk as never).getUnitConsumptionSummary(THIRTY_DAYS_AGO, NOW, { lookbackPeriodAnalysis: true })
+const c = r.currentPeriodSummary.totalAgentUnitConsumption
+const p = r.lookbackPeriodSummary?.totalAgentUnitConsumption
+return [{ value: c.completeJobs + c.incompleteJobs, previous: p ? p.completeJobs + p.incompleteJobs : undefined }]
```
Raina451: can we please use meaningful var names across all places?

r, c,p, svc etc are not readable. with the skill our idea is to give user kickstart , code should be readable.

skills/uipath-coded-apps/references/sdk/agents.md
```diff
@@ -0,0 +1,239 @@
...
+| Service | Subpath | Convention |
+|---------|---------|------------|
+| `Agents` | `@uipath/uipath-typescript/agents` | **Positional `Date` args**: `getAll(startTime, endTime, options?)` |
+| `AgentMemory` | `@uipath/uipath-typescript/agent-memory` | **Options object**: `getTimeline({ startTime?, endTime?, ... })` — dates inside the object |
+
+## Agents Service
+
+```typescript
+import { Agents, AgentListSortColumn } from '@uipath/uipath-typescript/agents';
+const svc = new Agents(sdk as never)
```
Raina451: why as never?

tests/tasks/activation/uipath-coded-apps.jsonl
ninja-shreyash: @NishankSiddharth I have a separate PR for adding tasks.. I can add there itself

NishankSiddharth: @ninja-shreyash @Raina451 in this PR only dashboard related tests are only added, let's keep other tests in different PR

ninja-shreyash: I am yet to structure the coded app skill tests - for dashboard related tests, we can go ahead with this PR itself. But want to understand one thing: won't there be overlap because in the end its a single skill right?

skills/uipath-coded-apps/references/sdk/governance-traces.md
NishankSiddharth: Updated to use  full type-through: MetricFn now takes sdk: UiPath and returns Row[]. Removed any/as never across all SDK references and the registry recipes

NishankSiddharth:
> build script is too big , try refactoring it into seprate modules

 Refactored in two passes rather than many micro-modules: (1) pulled the project-lifecycle phase (versioning / prewarm / sdk-check / dev-server teardown) into lifecycle.mjs; (2) split the four use-case flows out of build-dashboard.mjs into flows/ — build.mjs (fresh build), edit.mjs (incremental ADD/REMOVE/CHANGE/REBUILD/EJECT), upgrade.mjs (scaffold refresh), template.mjs (template packaging). build-dashboard.mjs drops from ~2020 to ~1360 lines and is now the shared-primitive library + entry-point dispatcher. All scripts also moved under assets/scripts/dashboards/.

NishankSiddharth:
> for governance-dashboard-starter-kit.zip in source control can we do the following?
> 
> 1. Store kit content uncompressed in git and zip it in CI (better diffability, no runtime extractor dependency).

The kit's uncompressed, diffable source already lives in apps-dev-tools/uipath-dashboard-starter-kit (https://github.com/UiPath/apps-dev-tools/pull/99)  — that's where it's reviewed per-file, versioned, and packed. The skills repo just vendors the built archive (like a prebuilt dependency). Storing it uncompressed + CI-building here would duplicate that pipeline in the wrong repo. One change from before: the vendored artifact is now a .tar.gz (not .zip), extracted/created with the OS tar on every platform

> 2. Ship a tiny Node-based zip-extraction fallback in the script so the OS dependency is optional.
>    the test locateWidgetTemplates already iterates over four extractors as fallbacks so approach seems fragile. Worth fixing at the source.

Resolved by removing the hand-rolled extraction entirely. The fixture is now a standard .tar.gz extracted and created with the OS tar, which ships on Windows 10+, macOS, and Linux

skills/uipath-coded-apps/SKILL.md
NishankSiddharth: Tightened the dashboard sentence to one line. On the redirects — keeping them: the repo's skill-review rules require → redirects so sibling skills (rpa / agents / maestro / solution) don't mis-route; trimmed rather than dropped. Note: any description change re-triggers the activation recall-eval, so confirming routing after.

skills/uipath-coded-apps/SKILL.md
NishankSiddharth: Governance isn't a separate build type — it's a gated widget set plus a deploy target. So "build me a dashboard" stays generic by design; the divergence happens at deploy: a governance/admin dashboard goes to AdminDashboards with elevated perms + an optional Governance-UI pin, a standard dashboard goes to a user-chosen folder with no elevated perms and no pin. The deploy step infers the target from the dashboard's widget metrics and lets you correct it in the plan

NishankSiddharth: Same rationale as the description thread — the → redirects are the sibling-routing guard the skill-review rubric mandates.

skills/uipath-coded-apps/SKILL.md
NishankSiddharth: Reworded to "generate an agent-monitoring / analytics dashboard".

skills/uipath-coded-apps/SKILL.md
NishankSiddharth: You're right they deploy differently — and today we force the AdminDashboards / elevated-perms / governance-tag flow on ALL dashboards, which is wrong for a plain dashboard app. Implemented: keep one build path, diverge at deploy. A generic dashboard deploys like a normal web app (user's folder, no governance tag, no admin-permission grant); a governance/admin dashboard deploys to AdminDashboards + an optional Governance-UI pin (reusing the pin question we already ask). The governance path is preserved verbatim; the standard path is additive. The deploy step determines the target first (inferred from widget metric names, confirmable in the plan), so there's no new upfront gate on the build.

skills/uipath-coded-apps/SKILL.md
NishankSiddharth: any action for me?

skills/uipath-coded-apps/SKILL.md
NishankSiddharth: Yep — that's Rule 18. Fixed here and one more stale ref in deploy/impl.md.

skills/uipath-coded-apps/references/sdk/agents.md
NishankSiddharth: Done — renamed to readable identifiers (summary, current, prior, agents, memory, result, distribution, point, job, evaluations…) everywhere. The snippets are kickstart code, so they should read cleanly.

skills/uipath-coded-apps/references/sdk/agents.md
NishankSiddharth: It was unnecessary — services take a UiPath instance by DI (constructor(instance: IUiPath)), and new Agents(sdk) type-checks once sdk is typed UiPath. Verified against the SDK. Removed everywhere.

tests/tasks/activation/uipath-coded-apps.jsonl
NishankSiddharth: It won't overlap because of activation, dashboard capability will be only triggered in certain scenario.
Rest of the other usecases can be more generic (not focused towards dashboard)

skills/uipath-coded-apps/SKILL.md
Raina451: no

Raina451 (approved):

deepeshrai-tech (approved):

gabrielavaduva (approved):

NishankSiddharth:
## Runtime governance migrated to SDK 1.5.1 `AgentTraces` APIs

Two new commits:

**`2c66acf5` — test/doc hygiene.** Every dashboard task YAML now carries the required `mode:*` + `lifecycle:*` tags; out-of-vocabulary `feature:` values dropped or mapped to the closed set (`compliance`, `approval-gate`); stale `fnBody`-era JSDoc corrected; two "Critical Rule 17" cross-references renumbered to 18 (structured choices) after the rule-numbering shift.

**`26009ac4` — the migration.** The runtime-compliance metric family (11 catalog entries) now reads the first-class Insights endpoints shipped in `@uipath/uipath-typescript` 1.5.1:

- `AgentTraces.getGovernanceSummary(start, opts)` powers every breakdown widget (`byHook`/`byAgent`/`byPolicy`/`byPack`, opt-in `byAction`/`byMode` via `sections`); the violations KPI gains a real vs-previous delta via `priorWindow`
- `AgentTraces.getGovernanceDecisions(start, opts)` powers record tables and drill-downs; per-run views group/filter client-side on `traceId`
- The trace-scan approach is fully removed: no `@/lib/governance`/`governance-scan` libs, no 15-run cap, no `WINDOW_LABEL` subtitles, over-cap `hardRefuse` entry deleted — governance widgets honor the dashboard time range like any other metric
- `sdk/governance-traces.md` rewritten as the new API reference (example responses grounded in SDK test fixtures, org-admin 403 handling, the `sections` opt-in trap, enum-comparison rule)
- Routing gate unchanged: runtime-compliance widgets still require an explicit standard/pack/ISO/rule-violation signal; generic policy intent stays on `Governance.getPolicyTraces`/`getOperationSummary`
- Starter kit 2.17.0 (fixture refreshed from apps-dev-tools PR #99): SDK lockfile → 1.5.1, parser libs deleted; `MIN_SDK_VERSION` → 1.5.1, `SKILL_VERSION` → 2.2.0
- Metric keys/aliases unchanged — deploy governance-target inference and existing tests keep matching

### Validation
- Script unit suite: **161 pass / 0 fail** (4 new guards: no span-scan tokens in the registry, every gov entry cites the new methods, over-cap phrasings no longer refused, SDK floor 1.5.1)
- Shipped-kit end-to-end: extracted the committed tar.gz, `npm ci` (installs SDK 1.5.1), compiled the doc recipes verbatim (`agent-governance-violations` with delta + full `agent-compliance-report` with `fetchDetailByKey`) under `tsconfig.metrics.json` — clean; full-scaffold `tsc` clean
- `dashboard_gov_gated_routing` / `dashboard_gov_generic_no_regression` re-pointed (require/forbid the shared `getGovernance` prefix) and lint clean
- ⚠ Pending: a live coder-eval run of the two governance tasks against a tenant with agentic-governance data (integration-tier, nightly budget) — not run in this session

Generated with Claude Code

https://claude.ai/code/session_019BrysPytXxf9LtC3xQDYSC

NishankSiddharth:
## CI run 28589047076 triage — 14/16 pass; the 2 failures share one root cause, fixed in `02d5eb38`

**Failures:** `dashboard_gov_gated_routing` (0.25) and `dashboard_edit_remove_widget` (0.14). Not governance-specific and not a regression in the SDK 1.5.1 metrics themselves.

**Root cause (from the preserved sandboxes):** the pre-warm correctly extracted the starter kit into the relative `<routingName>` folder, but the main thread then filled the build-subagent prompt with a *reconstructed absolute* sandbox path — `/tmp/coder_eval_…` (underscores) instead of the real `/tmp/coder-eval-…` (dashes). The subagent `mkdir -p`'d the phantom path, re-extracted the kit there, and completed a fully "successful" build (tsc clean, 6 widgets) in a sibling directory nobody grades or sees. In the edit task the subsequent REMOVE op then anchored to the same phantom project. The other 14 tasks passed because the model happened not to mangle the path — nondeterministic, so the contract had to change, not the odds. This is the same dash/underscore failure `a979f93b` fixed for the edit flow; the fresh-build contract still demanded an absolute `projectDir`.

**Fix (`02d5eb38`):**
- `plugins/build/impl.md` Phase 4: the subagent works only in the relative pre-warmed `<ROUTING_NAME>` folder — `cd` (fails loudly if missing, no `mkdir` replacement), `intent.json` carries `"projectDir": "."`, script runs from inside; absolute working-directory paths are banned from the prompt template and the intent schema
- `flows/build.mjs`: defense-in-depth guard — a nonexistent `projectDir` now fails with the path-safety remediation instead of allowing a phantom project
- Regression test added (162 pass / 0 fail)

Side observation from the run, not addressed here: tasks execute serially in a shared runner and a background Vite dev server from an earlier task can still hold port 25173 when a later task starts — harmless to grading, but worth a runner-level cleanup step eventually.

Generated with Claude Code

https://claude.ai/code/session_019BrysPytXxf9LtC3xQDYSC

NishankSiddharth:
## CI 28592101807 — 15/16 (path fix confirmed working); the last failure was a criteria bug, fixed in the commit above

`dashboard_full_e2e` scored 0.71: the weight-5.0 gold gate (structure + state.json + clean tsc) **passed** — the build itself was correct. The weight-2.5 criterion failed only because it pinned the literal `getConsumptionTimeline`, while the agent served "agent consumption over the last 24 hours" through the sibling catalog routes SDK 1.5.x added (`getTopConsumption` ranked table + `getUnitConsumptionSummary` KPI) — both legitimate.

Fix: the criterion now requires the `"Consumption"` method-family substring (covers all four consumption routes) instead of one method, and the prompt says "consumption **trend**" so the intended timeline mapping is deterministic. Verified by re-running `check_dashboard.py` with the new arguments against the failed run's preserved sandbox — PASS.

Generated with Claude Code

https://claude.ai/code/session_019BrysPytXxf9LtC3xQDYSC

NishankSiddharth:
Latest coder eval run after TS SDK version bump and test fixes

1. https://github.com/UiPath/skills/actions/runs/28593592621
2. https://github.com/UiPath/skills/actions/runs/28592121060

PR #1601, opened 2026-06-22
fix(troubleshoot): handle external automation cases

vlad-uipath:
closing https://uipath.atlassian.net/browse/UV-15011
## Summary
  - It tells troubleshooting triage to classify cases with no UiPath project, no failing uip command, and no UiPath runtime entity as external-automation.
  - It stops the agent from forcing such cases into UiPath product playbooks.
  - It requires a small diagnostic probe before repeated document/file mutation loops.
  - It stops after two no-progress runs instead of allowing 60 no-op passes.
  - It adds PowerShell hygiene: reset Set-PSDebug -Off and prefer fresh pwsh/powershell -NoProfile phases.
  - It requires a persistent task log for multi-session work, so context loss does not erase script-level findings.


vlad-uipath:
addressing https://github.com/UiPath/skills/pull/1601#issuecomment-4766578641

MarinRzv:
Closing as superseded by the single-file architecture refactor.

This PR targets `agents/triage.md` and the `state.json` multi-agent model, both of which were removed when `uipath-troubleshoot` was distilled into a single inline `SKILL.md` (no `agents/` directory, phase machine inline). The diff no longer applies.

More importantly, the failure mode from UV-15011 is now contained by the current design:

- **The 60-turn no-op mutation loop can't happen** — there is no turn-capped orchestrator state machine anymore, and invariant 11 ("No ad-hoc code execution": playbook snippets are recommendations for the user, not agent-run code) precludes the agent driving repeated document/file mutation scripts. That removes the need for the diagnostic-probe, stop-after-2-no-progress-runs, PowerShell `Set-PSDebug`/`-NoProfile` hygiene, and `task-log.md` safeguards.
- `notes.md` already serves as the running investigation log.

The one idea still worth having — the **external-automation carve-out** (recognize a case with no UiPath project marker / no failing `uip` command / no runtime entity and stop instead of forcing a UiPath playbook) — is a small gap best added fresh as a branch in `SKILL.md` §2, not via this stale diff. Happy to open a focused PR for just that if it's still wanted.

PR #1603, opened 2026-06-22
fix(ixp): update e2e + skill for Design-Time get-metrics shape

cezara98t:
## Summary

The Design-Time API migration ([cli#2628](https://github.com/UiPath/cli/pull/2628)) changed the `get-metrics` output shape. Found via the IXP e2e (`full_lifecycle`), which failed its metrics check on the new output. This updates the e2e check **and** the skill docs to the new shape.

### Shape change
| | Old (reinfer) | New (Design-Time) |
|---|---|---|
| Per-field/group | top-level `Fields[]` / `FieldGroups[]` arrays | `Data.Metrics.FieldMetrics.{FieldsMetrics, FieldGroupsMetrics}` — objects keyed by group name, then field id |
| F1 | `field.F1` (number) | `…F1Score.Value` |
| Overall | — | `Metrics.ProjectScore` / `ProjectScoreQuality` |

### Changes
- **`check_full_lifecycle.py`** — parse `Metrics.FieldMetrics.FieldsMetrics` (flatten group→field id), read F1 via `F1Score.Value`, `ModelVersion` from `Metrics`. Verified green against a real e2e run (11 fields, target F1 0.667→0.571, within tolerance).
- **`cli-reference.md`** `get-metrics` row, **`SKILL.md`** "How is this project performing?" row, **`improve-prompts-guide.md`** — document the new `Metrics.*` shape.

Left `label-documents-guide`'s `get-predictions` `Fields[]`/`Labels[].Fields[]` refs alone — that's a different command's output, not metrics.

Pairs with cli#2628 — merge once that migration ships to `@uipath/cli`.

PR #1610, opened 2026-06-22
test(ixp): assert e2e confirm succeeds, not just that it ran

cezara98t:
## Problem

The full-lifecycle e2e ([`tests/tasks/uipath-ixp/e2e/full_lifecycle.yaml`](tests/tasks/uipath-ixp/e2e/full_lifecycle.yaml)) graded `labellings confirm` with a single `command_executed` criterion. That asserts the command was **invoked**, not that it **succeeded** — so a write-path failure (e.g. a 502 on the confirm endpoint) would still pass the e2e green.

Surfaced during the Design-Time migration QA (ACTV-88813): on the new `designtimeapi` route, `labellings confirm` 502s deterministically, yet the e2e's `confirm` assertion would have stayed green because it only checks invocation. `command_executed` exposes no exit code or output, so it structurally cannot tell a 200 from a 502 — the result has to be captured to a file the grader can read.

## Fix — two layers on the captured `confirm_result.json`

The agent saves the `labellings confirm` response (`--output json`) to `confirm_result.json` (the CLI writes `Result: Failure` to stdout on a 5xx even with `--output json`, so a 502 is captured).

1. **YAML** — `file_exists` (artifact captured) + `json_check` asserting `Result == "Success"` (named, visible signal in the report).
2. **`check_full_lifecycle.py`** (the weight-5 gate) — also asserts `confirm_result.json` is `Result == Success`, **first**, before any metrics comparison, so the single authoritative script fails fast on a broken confirm and can't be sidestepped.

Verified: the script exits 0 on a success response and exits 1 (`FAIL`) on a 502-shaped `Result: Failure` response.

## Scope / notes

- `confirm` is the only mutating labelling this e2e exercises (its improve step is `fields update-prompts`). Same pattern should extend to `unconfirm`/`mark-missing` if future steps add them.
- There is no native mid-run "fail fast" in coder_eval — criteria are graded post-hoc. This is as close as the framework allows: the weight-5 script validates confirm first and exits non-zero.
- **Merge note:** PR #1603 also edits `check_full_lifecycle.py` (Design-Time metrics shape). The two changes are in different parts of the file but expect a small merge; whichever lands second should keep `check_confirm_succeeded()` called first in `main()`.

### DIFF
--- a/tests/tasks/uipath-ixp/e2e/check_full_lifecycle.py
+++ b/tests/tasks/uipath-ixp/e2e/check_full_lifecycle.py
@@ -72,7 +72,29 @@ def read_f1(field: dict) -> float | None:
         return None
 
 
+def check_confirm_succeeded() -> None:
+    """Fail fast if the labelling confirm did not actually succeed.
+
+    `command_executed` only proves the confirm command was *invoked*; a
+    write-path failure (e.g. a 502 on the confirm endpoint) still emits
+    `Result: Failure` to stdout with --output json. Assert the captured
+    response is a success so the e2e fails on a broken confirm instead of
+    grading invocation alone."""
+    confirm = load_json("confirm_result.json")
+    result = confirm.get("Result")
+    if result != "Success":
+        log_fail(
+            f"labellings confirm did not succeed (Result={result!r}, "
+            f"Code={confirm.get('Code')!r}): {confirm.get('Instructions') or confirm.get('Message')}"
+        )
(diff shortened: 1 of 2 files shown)


cezara98t:
Closing — the explicit confirm-success guard is redundant. If confirm doesn't actually succeed there's no confirmed ground truth, so get-metrics has no F1 and check_full_lifecycle.py (weight 5) already fails. command_executed + the metrics gate is sufficient.

PR #1614, opened 2026-06-22
test(uipath-rpa): set turn_timeout/task_timeout in task run_limits

Mihaiii:
Add turn_timeout/task_timeout (and max_turns where unset) to the uipath-rpa task files so each task carries the budget it needs without depending on an experiment file supplying defaults. Values mirror the run_limits defaults in tests/experiments/smoke-windows.yaml (turn_timeout 900, task_timeout 900, max_turns 40) — the config the GitHub Actions Windows job runs under.

Only adds keys that were unset; existing per-task values are preserved (coded_test_case 400/400, uia_google_search 1200/1800 untouched). Lets these tasks run with correct timeouts in environments that invoke coder-eval without -e (e.g. the ADO RPA-eval pipeline), where the framework default.yaml otherwise caps turn_timeout at 300s.

### DIFF
--- a/tests/tasks/uipath-rpa/coded_test_case.yaml
+++ b/tests/tasks/uipath-rpa/coded_test_case.yaml
@@ -10,8 +10,8 @@ tags: [uipath-rpa, windows, smoke, coded, hooks, test-case]
 run_limits:
   expected_turns: 22
   max_turns: 70
-  task_timeout: 400
-  turn_timeout: 400
+  task_timeout: 2000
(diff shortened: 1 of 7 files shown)


Mihaiii:
The timeout was increased because llmgw adds extra latency. Because of this, a message takes longer in Delegate SDK versus Claude Code. Notice that I kept the same max_turns values.

gabrielavaduva (approved):

PR #1622, opened 2026-06-22
feat(maestro-case): build missing agents inline during planning

song-zhao-25:
## Summary

Teaches the `uipath-maestro-case` skill to **build a missing agent inline** during Phase-1 planning, instead of only leaving a placeholder. When an SDD-referenced **agent** can't be resolved at the Rule 17 empty-lookup gate — absent from the tenant registry **and** with no in-solution sibling — the skill offers a third option, **Create the missing agent(s) inline**, alongside force-pull and skip. On Create it builds each selected agent as an **in-solution sibling** via `uipath-agents` in parallel sub-agents (capped at 10), registers them into the solution `.uipx`, and rediscovers + binds them via `registry --local`. Built agents enter Phase 2 as normal resolved tasks.

> **Scope: agent only.** Inline-create for **agentic processes** (Process Orchestration, via `uipath-maestro-bpmn`) is split into **#1723, stacked on this one** — its BPMN leg has open issues, so it merges later. The Create-on-Missing orchestration here is written type-agnostically so the second kind drops in cleanly on top. The split is lossless: this PR + #1723 reconstruct the original combined tree byte-for-byte.

## Requires `--local` (now available)

The feature uses `uip maestro case registry <search|list|get> --local`, available in the CLI today. The skill still **probes for `--local` and gracefully degrades** — run against an older CLI without the flag, the Create option is suppressed and behavior is exactly as before. Safe to merge independently.

## What changed (11 files, docs-only)

- **`SKILL.md`** — Rule 17 gains the `Create the missing agent(s) inline` option; the "never auto-invoke other skills / on-demand creation is future" anti-pattern is carved out for the agent inline-build path (other kinds — regular RPA process, action, connectors, agentic process — still fall to placeholder); `allowed-tools` adds `Agent` (the sub-agent launcher).
- **`references/registry-discovery.md`** — 3-option gate + new **§ Create-on-Missing** orchestration (capability probe → multi-select → per-agent build-kind choice → parallel sub-agents `--skip-solution-registration` → sequential `uip solution project add` → `resources refresh` → rediscover/verify/bind); plus pre-gate in-solution-sibling resolution (idempotency) and the "already exists" adopt path.
- **`references/plugins/tasks/agent/planning.md` + `agent/impl-json.md`** — **§ Creating an Agent inline**: SDD-composed Purpose, the pinned-I/O contract, the build-kind gate (Low-code / Coded, equal), the coded-agent caveat, the `folderPath:""` binding rule, and the failure contract.
- **`variables/bindings/impl-json.md`, `placeholder-tasks.md`, `phased-execution.md`, `phase-0-interview.md`, `planning.md`, `bindings-v2-sync.md`, `case-commands.md`** — gate wording, completion-report shape, the decoupled `resourceKey`/`folderPath` rule, and `--local` command docs.

## Key design decisions

- **Skill-driven build via sub-agent** (`uipath-agents`): build knowledge stays in the builder skill; the case skill hands over a pinned brief and verifies the result.
- **Gate-selected only** — SDD content alone never triggers a build (the SDD is untrusted sole input; the gate selection is the human-approval checkpoint). **Build kind is chosen at the gate** (Low-code / Coded as equal choices, low-code the non-interactive fallback) — never inferred from the SDD.
- **Contract** pins only the fields the case wires (typed Case Variable → type; unwired → the builder free-styles); kind, tools, and internals are the builder's concern.
- **Verify + reconcile** reads the sibling's declared I/O from its on-disk `entry-points.json` (`input`/`output.properties`), never from `--local`'s PascalCased fields. Matched / missing / extra are reconciled against the pinned contract — **warn + diff, never block**.
- **Binding** reuses name+folder with the `solution_folder` sentinel in `resourceKey` only; the sibling ships inside the solution `.uipx` and co-deploys — never a separate tenant publish.
- **Idempotency** — a pre-existing in-solution agent sibling resolves **before** the gate (no rebuild); an interrupted-run residual ("already exists" at build) is **adopted** (register + bind), not routed to Retry/Skip.

## Runtime binding — the `folderPath:""` rule (materialized in a live run)

The case task's **runtime** `folderPath` binding `default` must be `""` (co-located), **not** the `solution_folder` sentinel. Deploy correctly *provisions* the sibling into the real solution folder, but the deployed case starts the sibling using its baked-in `folderPath` — and the literal `solution_folder` is never created as an Orchestrator folder, so invocation fails `folder not exist`. The `solution_folder` sentinel stays ONLY in the resource-identity layer (`resourceKey`, `resources/solution_folder/`, `bindings_v2.json`) — intentionally **decoupled** for inline siblings. (`validate` accepts both, so this is not caught by validation.)

## Verification

- **Adversarial regression review of the agent-only split** (link-integrity / dangling-reference audit / re-scoping consistency / self-containment & logic / structure), every finding adversarially re-verified — **0 confirmed issues** (all dimensions 5/5). Confirmed the shared files were genuinely re-scoped to agents-only (not just section-deleted), so no dangling reference to the split-out content remains.
- The agent create-on-missing path was exercised end-to-end against `alpha.uipath.com`: build/register/bind for agents (low-code + coded), idempotent resolution, adopt-on-residual, and reconcile-mismatch — all `uip maestro case validate` clean. Coded siblings install + run on Orchestrator.
- All relative links/anchors in the changed files resolve.

## Tests

The agent path was validated with the create-on-missing test suite (agent control, coded-agent control, idempotency, on-disk-unregistered residual, I/O-shape + reconcile-mismatch), grading on-disk artifacts. The mixed/process fixtures move with #1723. Standalone `coder-eval` fixtures in `tests/` remain a follow-up — the interactive gate + sub-agent build need a custom experiment harness (`AskUserQuestion` + `Agent`, scripted gate).


charlesliu9:
• Findings

  - P1: Inline create is scheduled in Phase 1 before a solution exists. Phase 1 now says it builds/registers siblings, but uip solution init still happens in Phase 2, and registry --local fails without an
    enclosing .uipx. Fresh runs will suppress or fail Create before it can work. See skills/uipath-maestro-case/SKILL.md:64, skills/uipath-maestro-case/references/implementation.md:79, skills/uipath-maestro-
    case/references/registry-discovery.md:75, skills/uipath-maestro-case/references/case-commands.md:272.

  - P1: Built local siblings do not have a durable Phase 1 → Phase 2 contract. The docs say local EntityKey is audit-only, but tasks.md formats still require taskTypeId, while Phase 2 still runs tenant tasks
    describe for every non-connector task. Local siblings need entry-points.json schema instead, so after the hard stop the executor can choose the wrong path. See skills/uipath-maestro-case/references/
    registry-discovery.md:112, skills/uipath-maestro-case/references/implementation.md:116, skills/uipath-maestro-case/references/plugins/tasks/agent/planning.md:125.

  - P2: Missing cache files bypass the new Create gate. The prerequisite says an absent index after pull proceeds directly to placeholder, but for agent-index.json / processOrchestration-index.json absence is
    also a valid “0 matches” case where Create should be offered. See skills/uipath-maestro-case/references/registry-discovery.md:15 and skills/uipath-maestro-case/references/registry-discovery.md:27.

  - P2: bindings_v2.json generation is internally contradictory for inline siblings. The procedure says derive folderPath.defaultValue from the caseplan binding default, while inline sibling bindings set that
    default to ""; later text says bindings_v2 must use "solution_folder". Add an explicit special case or remove the exception. See skills/uipath-maestro-case/references/bindings-v2-sync.md:27, skills/
    uipath-maestro-case/references/plugins/variables/bindings/impl-json.md:113, skills/uipath-maestro-case/references/bindings-v2-sync.md:127.

song-zhao-25:
Thanks @charlesliu9 — all four addressed in `9b90afca1` (branch rebased onto latest `main`).

**F1 (P1) — Create scheduled in Phase 1 before a solution exists.** Fixed. The Create flow now scaffolds the solution *first*: `registry-discovery.md` §0 gains a **Solution prerequisite** (run `uip solution init` if no enclosing `.uipx` before §3 register / §4 `--local`); the capability probe now distinguishes a `No solution found for --local` error (scaffold + re-probe) from an `unknown-option` error (suppress Create); and Phase 2 Step 6.0 (`implementation.md`) **skips its `init`** when the solution already exists. `SKILL.md:64` Phase-1 bullet updated to match.

**F2 (P1) — local-sibling Phase 1→2 contract.** Addressed — though I'd suggest **downgrading to P2**: the agent plugin already carved this out (`agent/impl-json.md` § *Built-inline sibling*: skip tenant `tasks describe`, read I/O from the sibling's `entry-points.json`), so the executor wasn't actually going to take the tenant path. What was missing was consistency in the **driver**: `implementation.md` Step 9 Phase A and the §4 *Record* step now both state inline siblings skip the gather, and that `taskTypeId` holds the **local audit-only EntityKey as a resolution marker only** (never tenant-described; `entry-points.json` is authoritative). The "`EntityKey` audit-only / never write" vs `taskTypeId: <entityKey>` apparent contradiction is real but targets different artifacts — the caseplan **node** (bound by name+folder, never carries the key) vs the `tasks.md` planning handoff — now clarified inline.

**F3 (P2) — missing cache file bypasses the Create gate.** Fixed. `registry-discovery.md:15` now treats a still-absent **creatable** (`agent`) index after a successful pull as the genuine 0-matches case → **resolve in-solution siblings first** (preserves re-run idempotency), then feed it to the Rule 17 gate (Create offered). Non-creatable types still proceed to placeholder.

**F4 (P2) — bindings_v2 internally contradictory.** Fixed. The §Regenerate procedure now carries the exception **inline**, right after the non-connector shape where a linear reader meets it before writing the value: for an inline sibling `value.folderPath.defaultValue` is `"solution_folder"` (not the caseplan `""` default), with a pointer to the existing decoupling note.

Each fix was independently re-verified against the cited files. Appreciate the thorough review.

charlesliu9 (approved):

PR #1624, opened 2026-06-22
docs(uipath-solution): document Entity and ChoiceSet as resources-add kinds

AditiGoyalUipath:
## Summary

- Add Data Fabric kinds (`Entity`, `ChoiceSet`) to Step 9 of `develop-solution.md`: list them in the `--kind` table row and include two `--source remote` example commands.
- Note that DF kinds must already exist on the tenant (created via `uip df entities create` / `uip df choice-sets create`) — local virtual stubs are not supported for DF kinds.

## Why

Smoke-tested a solution that packages a folder-scoped entity + choice set. The packer accepts `--kind Entity` and `--kind ChoiceSet` (case-insensitive, normalize to `entity` / `choiceSet`), and the resulting `.uipx` deploys + auto-activates cleanly. But the existing docs only listed Queue/Asset/Bucket/Process/Connection/App/Index/Trigger, so an agent following them would not know DF resources work the same way.

End-to-end verified on `codereval/DefaultTenant`:
- Created `CustomerStatus` choice set (+ 3 values) and `Customer` entity in `Shared` via `uip df`.
- `uip solution resources add --source remote --kind ChoiceSet --name CustomerStatus --folder-path Shared` → `Status: Added`.
- `uip solution resources add --source remote --kind Entity --name Customer --folder-path Shared` → `Status: Added`.
- `uip solution pack` → `df-solution-smoke_1.0.0.zip` (3.1 KB) with `resources/solution_folder/ChoiceSet/CustomerStatus.json` and `resources/solution_folder/Entity/Native/Customer.json` inside.
- `uip solution publish` → State `Ready`.
- `uip solution deploy run --parent-folder-path Shared` → `DeploymentSucceeded` + `ActivationStatus: SuccessfulActivate`, folder `Shared/df-solution-smoke` created.

## Test plan

- [ ] CI markdown lint passes
- [ ] Frontmatter / link-check workflows pass
- [ ] Spot-check the rendered table on GitHub — `Entity` and `ChoiceSet` should show bolded in the `--kind` row


icpsoni (approved):
LGTM

skills/uipath-solution/references/develop-solution.md
```diff
@@ -136,6 +136,8 @@ uip solution resources refresh --solution-folder ./InvoiceAutomation --output js
 | `Skipped` | Resources already tracked in the solution |
 | `Warnings` | Bindings that couldn't be resolved (logged for follow-up) |
 
+A project that binds to a Data Fabric `Entity` in its `bindings_v2.json` will have that entity imported on refresh, the same way refresh handles queues, assets, or buckets. Choice sets don't bind directly from a project — pull a choice set in with [Step 9](#step-9-add-a-resource-atomically), or let the packer pick it up as a dependency of an Entity that references it.
```
SuphalAthlur, line 139: nit: Do you really want to give this choice to the agent? I think it would be better to recommend one approach based on how the regular flow handles it. If dependencies are recursively resolved and added to the solution, it makes sense to adopt the same approach

skills/uipath-solution/references/pack-and-deploy.md
```diff
@@ -319,6 +328,22 @@ Folder filtering with `--folder-path` happens **after** fetching `--limit` resul
 
 `--poll-interval` is in **milliseconds** (default 5000ms = 5s). `--timeout` is in **seconds** (default 360s = 6min). Do not confuse the two.
 
+### Data Fabric Resources Stay in Their Source Folder
+
+When the solution includes `Entity` or `ChoiceSet` resources imported via `solution resources add --source remote`, deploy does **not** move or copy them into the new deployment folder. The entity / choice set keeps the `FolderId` it had when created via `uip df entities create` / `uip df choice-sets create --folder-key <…>`; the new deployment folder only holds the binding.
+
+Data Fabric entities and choice sets are owned by the folder they were created in. A solution that imports one just records a pointer to that folder; deploying the solution into a different folder doesn't change ownership, it only wires up the binding the deployed projects will use.
+
+Verify post-deploy by querying the **source** folder, not the deployment folder:
```
SuphalAthlur, line 337: There's more to this right, based on whether binding overrides are provided the resources may or may not be copied at the target folder. Not sure how binding overrides are handled at deployment time via solution-tool though.
If your intent was to explain that the solution package, up until deployment, only stores a pointer to the actual resource and doesn't create a new resource, then we might need to call that distinction out explicitly. AFAIK it doesn't apply after deployment.

skills/uipath-solution/references/develop-solution.md
AditiGoyalUipath: Agreed — collapsed to the single recommended approach in commit af588c6e1. New line reads: *"Refresh pulls in any choice set a Data Fabric `Entity` references automatically as a dependency of the entity — don't add it separately."* Matches the recursive-dependency behavior the packer already implements.

skills/uipath-solution/references/pack-and-deploy.md
AditiGoyalUipath: Good catch — your reading of the intent was correct: solution package stores a pointer, not a copy. Reframed in af588c6e1 as "Pointer, Not Copy" and then tightened further in 0d58b754d. Current text:

> *"Unlike other remote-imported kinds, `Entity` and `ChoiceSet` are not provisioned into the deployment folder on a plain deploy — they stay in the folder where `uip df entities create --folder-key <…>` / `uip df choice-sets create --folder-key <…>` placed them, and the deployment folder holds only the binding. […] To retarget the binding at deploy time (dev vs. prod), use [`config link`](#link-to-an-existing-orchestrator-resource) to override the resource."*

Cross-refs `config link` for the override case, so binding-override behavior is deferred to the section that already documents it.

niket0503 (approved):

GeetUiPath (approved):
LGTM

alexenica (approved):

PR #1625, opened 2026-06-22
docs(inline-in-flow): IS connector tools on inline agents in a flow

al3xanndru:
## What

Teaches the agent how to expose an Integration Service connector activity as a tool on a low-code inline agent inside a flow, and adds an e2e task that exercises it.

## Changes

- **uipath-agents skill** — inline-in-flow reference now covers the connector-tool path: add a `uipath.agent.resource.tool.connector.<key>.<op>` node, configure it with `uip maestro flow node configure`, wire it to the agent's `tool` handle, then run `uip agent refresh --inline-in-flow` to generate the tool's `resource.json` (no hand-authoring) plus `bindings_v2.json` and the solution connection resource.
- **New e2e task** (`tests/.../inline_is_connector_tool`) + checker — scaffolds a flow with an inline agent and a Web Search connector tool, then asserts the generated artifacts: `resource.json` shape (`type: integration`, `location: external`, connection wired, `provider` static / `query` dynamic), the flow wiring (agent `tool` → connector `input` edge), `bindings_v2.json`, and the provisioned connection resource.

The inline-in-flow refresh provisions the connection resource itself, so the task grades that outcome via the checker rather than requiring a separate `uip solution resources refresh` command.


skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
```diff
@@ -681,6 +681,36 @@ For connector-trigger flows, the same pattern applies — top-level `bindings[]`
 | `node configure` fails with `customFieldsRequestDetails has unknown keys: ObjectActionName, ParameterValues` | PascalCase inner keys instead of camelCase | Use `objectActionName` / `parameterValues`. Studio Web emits camelCase; PascalCase is rejected. |
 | Field rejected at runtime as unknown (e.g. `"unknown field 'fields.labels[*]'"`) after a clean `flow validate` | `[*]` was left in the `bodyParameters` / `queryParameters` / `pathParameters` key. `[*]` is an array marker from `requestFields[].name`, not part of the wire key. | Strip `[*]` from the key in your `--detail` payload, pass an array value matching the field's `dataType`, and re-run `node configure`. Fields with `[*].` (segments after the `[*]`) are not authorable. See the array-fields table in Step 6b. |
 
+---
+
+## Agent Tool Connector Nodes
```
andreibalas-uipath: Move new ## section before the ## Debug section (after the ## Bindings section).

skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
```diff
@@ -681,6 +681,36 @@ For connector-trigger flows, the same pattern applies — top-level `bindings[]`
...
+  --output json
+```
+
+This single command populates the complete `inputs.detail` — connection info, `instanceParameters`, `connector`, `connectionResourceId`, `telemetryData`, and the request buckets `bodyParameters` / `queryParameters` / `pathParameters` (body fields as `{{prompt:}}` chips, static concrete values for single-value enums and query/path params with defaults). It also creates the top-level `bindings[]` entries. The `endpoint` path and `method` come from the definition's `model.context[]` or from `uip is resources describe`.
+
+### Debug Tips
+
+| Error | Cause | Fix |
+| --- | --- | --- |
+| `uip agent validate` shows `resources: 0` | No `resource.json` generated — tool node not wired to the agent's `tool` port, or `node configure` not run | Check the `tool` edge from the agent to the connector node, run `node configure`, then `uip agent refresh --inline-in-flow` |
```
andreibalas-uipath: I would move these debug tips under the `inline-agent` impl/planning references.

tests/tasks/uipath-agents/lowcode/inline_is_connector_tool/inline_is_connector_tool.yaml
```diff
@@ -65,14 +85,6 @@ success_criteria:
     weight: 1.5
     pass_threshold: 1.0
 
-  - type: command_executed
```
andreibalas-uipath, line 68: The skill requires `uip solution resources refresh` to be executed in order for connection binding to be generated. Shouldn't this criteria be kept?

tests/tasks/uipath-agents/lowcode/inline_is_connector_tool/check_inline_is_connector_tool.py
```diff
@@ -134,9 +153,63 @@ def assert_integration_tool_resource(agent_dir: Path) -> None:
...
+    params = props.get("parameters")
+    if not isinstance(params, list) or not params:
+        sys.exit(f"FAIL: properties.parameters missing or empty at {path}")
+
+    # Web Search shape — the feature's key, hard-won behaviors:
+    #  - provider is a single-value enum → baked STATIC (the agent must not be
+    #    asked to echo it) and carries NO enum in inputSchema (a static field
+    #    with an enum breaks the agent runtime's model build).
+    #  - query is free-text → DYNAMIC (agent-decided).
+    provider = _param(data, "provider")
```
andreibalas-uipath: If these fields are generated deterministically by the `uip agent refresh` CLI command I would not add explicit checks in this test. We want to make sure that the inline tool node was created and that `resource.json` was generated by `uip agent refresh --inline-in-flow`. The actual mapping should be tested in the CLI.

andreibalas-uipath (approved):

andreibalas-uipath (approved):

baishali-ghosh (commented):
## Review — IS connector tools on inline agents (auto-generated `resource.json`)

Solid, well-structured PR — prescriptive prose, links resolve, weights follow the rubric. Not blocking, but **one root-cause contradiction should be resolved before merge** since it appears in four places and the test's own grader disagrees with its own docs. Findings verified by three review passes plus manual anchor/behavior checks.

### Critical / Important

**1. `uip solution resources refresh` — the checker docstring contradicts the docs AND the YAML grader.**
`check_inline_is_connector_tool.py:31-34` says *"the inline-in-flow refresh path provisions it; no separate `uip solution resources refresh` step is required"*, and the `assert_connection_provisioned` FAIL message (~line 197) attributes the connection dir to `uip agent refresh --inline-in-flow`. But the docs (`inline-in-flow.md:229`, `inline-agent/impl.md:313,405`) and the YAML — which keeps a **weight-2.0 gating** `uip solution resources refresh` criterion (`.yaml:98-102`) plus the description at `.yaml:13` — all say that command creates the connection resource. Both cannot be true.
**Fix:** decide the real behavior once, then align all four. Given the doc body + gating criterion, the likely truth is *it IS required* → fix the docstring parenthetical and the FAIL message to attribute provisioning to `uip solution resources refresh`. (If genuinely optional, drop the criterion to `pass_threshold: 0`.) Don't ship the state where the docstring says "not required" but a gating criterion demands it.

**2. Enum → static/dynamic rule conflicts with the canonical doc the PR newly links to.**
The new prose (`inline-in-flow.md:222`, `connector/impl.md:707`) states a *static* field carrying an enum **breaks the agent runtime's model build**, so only single-value enums become static. But `integration-service.md:162` — which this PR adds a cross-link to — still gives the blanket rule *"Fields with `enum` values: set `fieldVariant: "static"`."* A reader following the linked canonical doc produces exactly the broken artifact the new docs warn against.
**Fix:** scope the two — narrow `integration-service.md:162` to "single-value enum → static; multi-value → keep dynamic, never emit `enum` on a static field," or state the inline-vs-standalone scope explicitly in both places.

**3. Canonical tool-kinds table not extended for the new IS-connector kind.**
`inline-agent/impl.md:232` — the summary table of all tool kinds, with a "Needs `uip solution resources refresh`?" column — has no IS-connector-tool row, though the PR adds a full section for it and adds `connector` to the planning annotation. The intro also still says tools "hand-author `resource.json`," now false for this kind. Adding the row also forces Finding #1 to be answered definitively (the cell must state Yes/No).
**Fix:** add `| IS connector tool | tool (bottom) | uipath.agent.resource.tool.connector.<key>.<op> | type: "integration" | <Yes/No per #1> | inline-in-flow.md |` and qualify the "hand-author" intro.

### Suggestions (non-blocking)

- **Checker false-positives** (all reproduced against a live fixture): `properties.parameters` accepts `[{}]`; the connection-dir check counts any `*.json` incl. `debug_overwrites.json`; `bindings_v2.json` accepts empty `{}`; `"-" in rid` isn't a real UUID check. → require a named parameter, exclude `debug_overwrites.json`, assert an actual `resource: "connection"` binding, use a UUID-shape regex.
- **Regex latent over-match:** `node\s+add\b` / `configure\b` match a following hyphen (`add-edge`, `configure-all`). No such verb exists today, so latent only. → use `add(\s|$)` / `configure(\s|$)`.
- **`--bindings-target` taught but never graded** — add an advisory (`pass_threshold: 0`) `command_executed` on `--bindings-target`.
- **Duplication:** the refresh mechanism (metadata-fetch + join, `{{prompt}}`→dynamic, enum-only-for-dynamic) is restated in 4 places. Cross-skill duplication (agents ↔ maestro-flow) is defensible, but the 3 *maestro-flow* files should have one owner (`connector/impl.md § Agent Tool Connector Nodes`) and cross-link, not re-derive — otherwise Finding #2 must be fixed in every copy.
- **Absolute claims that will rot:** "`type`: Always integration" (unqualified while `location` is qualified) and "the CLI **never** writes `fieldsContainer`" (×3) — soften to current-CLI-behavior.
- **`lifecycle:*` tag missing** (`.yaml:15`) — `tests/README.md` rule 1 requires it; add `lifecycle:generate`. Pre-existing suite-wide gap (all inline siblings omit it), not introduced here.

### Confirmed clean
Link/anchor integrity (all 3 cross-file anchors resolve) · CLI verbs current (plural `resources`) · weights within rubric · no `env_packages`/`@uipath/cli` reinstall · no secrets/personal paths · code blocks have language ids · `run_limits` (56 turns / 1200s) consistent with inline siblings (46–90) · shared-helper import path resolves · `find_inline_resource` fails loudly on absence · edge check binds agent↔connector by id+port (strongest assertion).

---
Findings #1–#3 share one root cause: the PR introduces "agent refresh provisions the connection / static-enum breaks the build" claims that contradict the surrounding docs and its own grader. Reconcile the actual behavior once, then make the docstring, FAIL message, doc body, YAML criterion, and the kinds table all agree.

baishali-ghosh (commented):
## uipskillreviewer — skills PR #1625
Lens: skills + IS / connector

**Summary:** The docs teach the new IS inline-tool flow well. But they add connection and folder wiring in `uipath-maestro-flow` with no deep link to the platform single source, and they give no rule to select the connection or resolve the folder.

### 🔴 Blockers
- `skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md:356-395` and `connector/impl.md:77` — the maestro-flow docs add IS connection mechanics inline. The `node configure` example sets `connectionId` and `folderKey`. The bindings block writes `ConnectionId` and `FolderKey`. No line deep-links `uipath-platform/references/integration-service/connections.md`. **Fix:** add a deep link to the platform `connections.md` for connection selection and folder scoping. Keep only the agent-tool wiring here.

### 🟡 Cautions
- `connector/impl.md:77` (`--detail '{"connectionId":"<CONN_UUID>","folderKey":"<FOLDER_KEY>",...}'`) — the example uses placeholders. No rule tells the agent how to pick the connection or resolve the folder key. **Fix:** state the selection rule, or deep-link the platform `connections.md` section that holds it.
- `tests/tasks/uipath-agents/lowcode/inline_is_connector_tool/inline_is_connector_tool.yaml:27-38` (`initial_prompt`) — the task needs live cloud auth. The prompt gives no rule for an auth failure. **Fix:** add a narrow line — if a `uip` call fails auth, do not `uip login`; stop and report.

### 🟢 Nits
- `connector/impl.md:81` — the one added `uip is` example is `uip is resources describe`. The verb is plural. The connector key is not a flag. This is correct. No change needed.
- `inline_is_connector_tool.yaml` — the task does not pin `@uipath/cli`. It asserts the `.flow` file exists plus a checker script. Both are correct.

**Could not assess without run data:** whether the added refresh / validate / refresh sequence pushes this task over its cohort's `expected_seconds`.

al3xanndru:
Thanks for the reviews. Addressed in `2b0b608b5`.

## @baishali-ghosh review — the 3 findings (all valid)

First settled the root-cause question from the CLI source: `uip agent refresh --inline-in-flow` writes only `bindings_v2.json`; **`uip solution resources refresh`** is what materializes the solution-level connection resource from those bindings. So the docs + the gating YAML criterion were right and the checker was wrong.

- **#1 — `solution resources refresh` contradiction:** fixed the checker docstring and the `assert_connection_provisioned` FAIL message to attribute the connection resource to `uip solution resources refresh` (they wrongly credited `agent refresh`). Kept the gating YAML criterion as-is.
- **#2 — enum rule conflict:** scoped `integration-service.md` — single-value enum → `static` (no `enum` in `inputSchema`); multi-value enum → stay `dynamic` and emit `enum`/`oneOf`. This removes the contradiction with the "a static enum field breaks the runtime model build" rule that this PR links to.
- **#3 — tool-kinds table:** added the missing `IS connector tool` row (`type: "integration"`, needs `uip solution resources refresh`: **Yes**) and qualified the "hand-author `resource.json`" intro — this kind is generated by `uip agent refresh --inline-in-flow`, not hand-authored.

The non-blocking suggestions (checker false-positives, `\b` regex, `lifecycle:*` tag, softening absolute prose, cross-file dedup) are left for a follow-up.

## uipskillreviewer bot review — `connections.md` deep link

The deep link it flags as missing already exists. `connector/impl.md:87` carries the `MUST READ … connections.md` pointer (selection rules, auto-select, BYOA, recovery, ping), and the new `inline-agent/impl.md` section delegates connection config to `connector/impl.md § Agent Tool Connector Nodes` in three places (lines 230, 239, 338) — so a reader reaches `connections.md` one hop away, which is the repo's pattern (connection mechanics live in the platform single-source; other docs link to it). The `:77` anchor it cites is pre-existing content, not part of this PR. No change needed there.

PR #1644, opened 2026-06-23
fix(tests): match inline escalation node by registry prefix; demote inline index discovery gate

andreibalas-uipath:
## Summary

Three uipath-agents lowcode inline tests were failing or over-gating despite correct agent output. This fixes the grading, not the skill.

### 1. Inline escalation check scripts — exact-match never matched

`check_inline_external_escalation.py` and `check_inline_solution_escalation.py` exact-matched the node type `uipath.agent.resource.escalation`. But the flow registry's canonical node type is fully-qualified:

```
NodeType: uipath.agent.resource.escalation.coded-action-app   ("Action App Escalation")
```

The bare string is only a prefix and never matches a real flow node, so both checks failed for **any** correct output. Switched to `node_type_prefix` matching (the shared `find_resource_node` helper already supports it), mirroring the existing `inline_context_index` convention (`uipath.agent.resource.context.index.`).

### 2. `inline_context_index` — `solution resources list` demoted to advisory

The `uip solution resources list` criterion was gating (weight 1.5, threshold 1.0). Inline agents can discover the index identity via `uip maestro flow registry search/get` instead, so gating on `solution resources list` rejected a valid discovery path. Demoted to advisory (weight 0.5, threshold 0.0), matching the `inline_solution_*` sibling family. The standalone `context_index` test keeps it gating — standalone agents have no flow registry to fall back on.

## Verification

Re-ran all three under `experiments/default.yaml`:

| Test | Result |
|------|--------|
| `inline_context_index` | ✅ SUCCESS 13/13 |
| `inline_external_escalation` | ✅ SUCCESS 12/12 |
| `inline_solution_escalation` | ✅ SUCCESS 11/11 |

(`inline_solution_escalation` first hit a transient 1200s agent timeout under `-j 2`; passed cleanly on a solo re-run.)

### DIFF
--- a/tests/tasks/uipath-agents/lowcode/inline_context_index/inline_context_index.yaml
+++ b/tests/tasks/uipath-agents/lowcode/inline_context_index/inline_context_index.yaml
@@ -45,13 +45,17 @@ success_criteria:
     weight: 1.5
     pass_threshold: 1.0
 
+  # Advisory, non-gating: inline agents can discover the index identity via the
+  # flow registry (uip maestro flow registry search/get) instead of solution
+  # resources list, so don't gate on it here. Standalone agents (context_index)
+  # have no flow registry and keep this gating.
   - type: command_executed
-    description: "Agent discovered the index identity with uip solution resources list"
+    description: "Advisory: agent discovered the index identity with uip solution resources list (informational, non-gating)"
     tool_name: "Bash"
     command_pattern: 'uip\s+solution\s+resources\s+list\b'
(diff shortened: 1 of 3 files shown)


al3xanndru (approved):

andreibalas-uipath:
Passing coder_eval pipelines:
* https://github.com/UiPath/skills/actions/runs/28030470797 (retry 1 test)
* https://github.com/UiPath/skills/actions/runs/28027585750

PR #1660, opened 2026-06-24
docs(uipath-agents): use `uip solution resources get` for escalation app discovery

al3xanndru:
## Summary

Replaces the raw Apps API `curl` approach for **escalation app discovery** in the low-code agent skill with a single CLI-native `uip solution resources get <Key> --output json` call.

The old flow hit two Apps API endpoints (`action-apps?state=deployed` then `action-schema?appSystemName=…&version=…`) by hand-sourcing `~/.uipath/.auth` and passing a bearer token directly. `uip solution resources get` returns the same data in one call via `Data.Spec`:

| Need | Old (curl) | New (`resources get`) |
|---|---|---|
| `systemName` | `action-apps` → match by `id` | `Spec.AppSystemName` |
| action schema | `action-schema?...` | `Spec.ActionSchema` (JSON string → `json.loads`) |
| `appVersion` (int) | `deployVersion` | parsed `ActionSchema.version` |
| `appName` | `deploymentTitle` | `Spec.Name` |

It resolves remote apps **even before they're imported into the solution**, so it works at discovery time, and the CLI handles auth — no auth file, no token, no Apps-API `User-Agent` workaround.

## Docs changed

- **lowcode `escalation.md`** — Steps 2–3 rewritten around `resources get`; resource-shape field comments updated.
- **lowcode `guardrails.md`** — escalation-app schema verifier rewritten to parse `Spec.ActionSchema`; decision table updated (`NO_SCHEMA`/`GET_ERROR` replace `NO_APP`/`API_ERROR`); dropped the now-redundant "don't hand-roll Apps API calls" Critical Rule and renumbered the rest (the positive instruction lives in Step 2; Rule 14 already enforces "don't skip schema validation").

## Tests changed

- **`guardrails/escalation_app`, `guardrails/escalation_app_validation`** — discovery success-criteria now match `uip solution resources get` instead of `action-schema|action-apps`.
- **`external_escalation`, `inline_external_escalation`** — added informational (`pass_threshold: 0`) `resources get` discovery criteria, mirroring the existing solution-escalation tests.

## Verification

- The rewritten Python verifier compiles and was smoke-tested live: `uip solution resources get <app> --output json | python3 verify.py` correctly parsed `Spec.ActionSchema` and reported the expected result — confirming raw stdout is clean JSON (logs go to stderr), so the documented pipe works as-is.

## Out of scope (intentionally untouched)

- The JWT email-extraction `curl` in `escalation.md` (recipient lookup — no CLI equivalent).
- The `/odata/Releases` + `GetPackageEntryPointsV2` curls in `process/solution-files.md` (process-*tool* discovery — a different capability; candidate for a follow-up).
- `coded/capabilities/guardrails/guardrails.md` — its schema-verification step never hand-rolled a curl, so nothing to replace.

## Eval run
[eval run](https://github.com/UiPath/skills/actions/runs/28237053619)

### DIFF
--- a/tests/tasks/uipath-agents/lowcode/external_escalation/external_escalation.yaml
+++ b/tests/tasks/uipath-agents/lowcode/external_escalation/external_escalation.yaml
@@ -105,15 +105,8 @@ success_criteria:
     weight: 1.5
     pass_threshold: 1.0
 
-  - type: file_exists
-    description: "FraudEscalation external escalation resource.json was created"
-    path: "FraudSol/FraudTriageAgent/resources/FraudEscalation/resource.json"
-    weight: 2.5
-    pass_threshold: 1.0
-
-
   - type: run_command
-    description: "External escalation resource shape (ActionCenter channel, external target)"
+    description: "External escalation resource shape (escalation resource located by content + ActionCenter channel bound to the FraudEscalation app; folder name agnostic)"
     command: "python3 $TASK_DIR/check_external_escalation.py"
     timeout: 30
     expected_exit_code: 0
--- a/tests/tasks/uipath-agents/lowcode/external_escalation/external_escalation.yaml
+++ b/tests/tasks/uipath-agents/lowcode/external_escalation/external_escalation.yaml
@@ -42,6 +42,26 @@ success_criteria:
     weight: 1.5
     pass_threshold: 1.0
 
+  # Informational discovery criteria (pass_threshold:0 — never block): the agent
+  # discovers the external app with `resources list` then fetches its action
+  # schema with `resources get`. The enforced signal is the authored
+  # resource.json shape (run_command below); discovery is acceptable-but-optional.
+  - type: command_executed
+    description: "Agent discovered the FraudEscalation app via uip solution resources list --source remote (informational)"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+solution\s+resources?\s+list\b.*--source\s+remote'
+    min_count: 1
+    weight: 0.5
+    pass_threshold: 0.0
+
+  - type: command_executed
+    description: "Agent fetched the app action schema with uip solution resources get (informational)"
+    tool_name: "Bash"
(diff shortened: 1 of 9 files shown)


tests/tasks/uipath-agents/lowcode/external_escalation/external_escalation.yaml
```diff
@@ -42,6 +42,26 @@ success_criteria:
...
     pass_threshold: 1.0
 
+  # Informational discovery criteria (pass_threshold:0 — never block): the agent
+  # discovers the external app with `resources list` then fetches its action
+  # schema with `resources get`. The enforced signal is the authored
+  # resource.json shape (run_command below); discovery is acceptable-but-optional.
+  - type: command_executed
+    description: "Agent discovered the FraudEscalation app via uip solution resources list --source remote (informational)"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+solution\s+resources?\s+list\b.*--source\s+remote'
```
andreibalas-uipath: ```uip\s+solution\s+resources\s+list\b.*--kind\s+App``` should be documented by the skill. The `resources list` and `resources get` are not informational in this case - there is no alternative `uip maestro flow registry` look-up.

al3xanndru: Good call — fixed in 4c659a13.

- Both `resources list --kind App` and `resources get` are now **enforced** (`pass_threshold: 1.0`, weight 1.5) in `external_escalation` and `inline_external_escalation`, since an external app has no alternative discovery path.
- Aligned the `resources list` regex with the documented `--kind App` form (the skill already documents `--kind App --source remote --search` in escalation.md Step 2, and `resources get` in Step 3).
- Left the solution-internal escalation tests informational on purpose: those apps ship in the fixture, so reading the local `solution_folder` files is a valid alternative to `resources get`.

andreibalas-uipath (approved):

apetraru-uipath (approved):

PR #1666, opened 2026-06-24
docs(uipath-agents): document @{ } expression tokens in agent prompts

andreiancuta-uipath:
agent-definition.md only documented `{{ }}` (variable tokens). The `@{ }` expression syntax (tools/contexts/escalations/output references) was undocumented, which let an automated review misclassify a valid @{contexts.X} prompt reference as invalid and flag the empty inputSchema as a defect (a context is a runtime-resolved resource, not an input).

- Document the `expression` token type and its @{ } families (tools.<Name>, contexts.<Name>, escalations.<Name>, output.<path>)
- Clarify {{ }} is inputSchema-only; @{ } targets are never inputs
- Add a context-reference worked example
- Add a smoke task + grader covering the @{contexts.X} -> expression token path and content<->contentTokens sync (no prior task exercised @{ } tokens; uses an attachments context so it stays self-contained)

Tested by asking Claude to create an agent referencing a context resource.

### DIFF
--- a/skills/uipath-agents/references/lowcode/capabilities/context/attachments.md
+++ b/skills/uipath-agents/references/lowcode/capabilities/context/attachments.md
@@ -25,6 +25,7 @@ For other context variants, see [context.md](context.md).
   "description": "",
   "contextType": "attachments",
   "indexName": "<ContextName>",          // same as name for attachments
+  "folderPath": "solution_folder",
   "attachments": {
     "description": "Array of files, documents, images to process."
   },
--- a/tests/tasks/uipath-agents/lowcode/context_prompt_reference/context_prompt_reference.yaml
+++ b/tests/tasks/uipath-agents/lowcode/context_prompt_reference/context_prompt_reference.yaml
@@ -1,0 +1,82 @@
+task_id: skill-agent-context-prompt-reference
+description: >
+  Skill-guided evaluation: scaffold a low-code agent with an attachments
+  context resource and reference that context in the user prompt via the
+  Studio expression syntax `@{contexts.<Name>}`, then validate. Exercises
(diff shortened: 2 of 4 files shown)


skills/uipath-agents/references/lowcode/capabilities/context/attachments.md
```diff
@@ -25,6 +25,7 @@ For other context variants, see [context.md](context.md).
   "description": "",
   "contextType": "attachments",
   "indexName": "<ContextName>",          // same as name for attachments
+  "folderPath": "solution_folder",
```
andreiancuta-uipath, line 28: Also added this missing property, which was being flagged by the validate CLI

cristian-groza (approved):

tests/tasks/uipath-agents/lowcode/context_prompt_reference/context_prompt_reference.yaml
```diff
@@ -0,0 +1,83 @@
...
+  max_turns: 30
+  turn_timeout: 1200
+
+initial_prompt: |
+  Create a UiPath low-code agent "DocAssistant" inside a solution named
+  "DocSol". The agent processes files the caller uploads at runtime and
+  retrieves relevant information from them (semantic retrieval over the
+  uploaded files — use a context resource named "Knowledge", not a plain
+  input field). Reference that context in the agent's prompt so the
+  model is told to use it. Make sure the agent validates successfully.
```
andreitava-uip, line 26: Could we try this eval without the explicit mention of referencing in the prompt? A normal user would likely not give such detailed instructions. We should test if the skill is clear enough that the agent understands both how and *when* to reference attachments/tools.

andreitava-uip (approved):

PR #1675, opened 2026-06-25
test(ixp): exercise Design-Time in CI (docker e2e+integration), mock shape-only smoke, add project naming

andrei-uipath:
## What
Make the `uipath-ixp` coder-eval suite correct against the **migrated Design-Time API**, stop the shape-only **smoke** tasks from hitting live alpha, and give eval-created projects a recognizable **naming convention**. Three related concerns, all in the `uipath-ixp` task tree.

## 1. Run e2e + integration in-container against the migrated tool
Under `driver: tempdir` these tasks run on the BYOS host, where `uip ixp` resolves the host's **stale pre-migration `ixp-tool` (1.1.0)** → legacy `reinfer_` backend → wrong/zero Design-Time coverage (and hollow "success").

- **Pre-bake `@uipath/ixp-tool`** in the eval image (`tests/docker/Dockerfile`).
- **`driver: docker`** for the e2e (`full_lifecycle`) + all **5** integration tasks (`taxonomy_lifecycle`, `versions_and_metrics`, `labelling_correctness`, `name_resolution`, `publish_lifecycle`) → they resolve the baked migrated tool → `du_/api/designtimeapi`.
- `run_limits` are kept at their `origin/main` values — the goal is to isolate the driver switch and confirm `driver: docker` alone is sufficient now the agent uses the correct tool.

## 2. Project naming convention
The prompts let the agent invent arbitrary project names — unrecognizable and collision-prone on the shared tenant. Mirror the e2e's `e2e-test-<random-suffix>` with per-scenario prefixes (`ixp-it-taxonomy-…`, `ixp-it-versions-…`, `ixp-it-labelling-…`, `ixp-it-publish-…`). `name_resolution` is intentionally left on its literal `IxP Int Resolve` titles — they're integral to its Title→Name resolution test and asserted by its criteria.

## 3. Mock `uip` in shape-only smoke tasks
The 35 smoke tasks are authored as offline, command-shape-only checks (*"Harness is unauthenticated; CLI calls fail with auth errors"*), but the smoke harness (`tests/experiments/smoke.yaml` + `smoke-skills.yml`) injects a **live alpha bot token**, so every `uip ixp …` actually reached designtime-api on alpha (404-ing on fixture ids like `doc-abc-123`) — recurring real traffic with no test value.

Per the repo convention (`uipath-troubleshoot/CLAUDE.md`: *"task.yaml MUST set `sandbox.mock_path_dirs: ["mocks"]`"*), add a shared `_shared/mock_template/mocks/uip` that PATH-shadows the real CLI and fails offline with **no network call**, and wire all 35 smoke tasks to it. Grading (command shape) is unchanged; integration/e2e (intentionally live) are untouched.

## Validation
- Live `Run Coder Eval` on this branch (run 28163165337): **all 5 integration tasks + e2e SUCCESS** (40/41). The one failure, `move-tag-smoke`, is a pre-existing strict negative-guard / agent-choice flake (it ran `untag`), unrelated to these changes. *(That run used higher turn budgets; budgets were since reverted to `origin/main` values to isolate the driver switch — a re-run at the original budgets is the remaining check.)*
- Smoke mock: all 35 tasks load via coder-eval `TaskDefinition`; the mock intercepts `uip ixp …` offline (exit 1, no network, call logged); committed LF + mode `100755`.

## Relationship to other PRs
This branch is a superset for running the full suite in one place. The **Dockerfile pre-bake + e2e** also live in **#1647**, and the **smoke mock** also lives in **#1676**. Overlap self-heals: whichever merges to `main` first, the identical content collapses to no-diff in the others.

https://claude.ai/code/session_01Mwj1Gyvj9GoEYWULV6eT98

### DIFF
--- a/tests/tasks/uipath-ixp/integration/name_resolution.yaml
+++ b/tests/tasks/uipath-ixp/integration/name_resolution.yaml
@@ -8,12 +8,14 @@ description: >
 tags: [uipath-ixp, integration, mode:operate]
 
 run_limits:
-  max_turns: 48
-  task_timeout: 1800
-  turn_timeout: 600
+  max_turns: 120
+  task_timeout: 2400
+  turn_timeout: 1800
 
 sandbox:
-  driver: tempdir
+  # driver: docker (NOT tempdir) so `uip ixp` resolves the migrated in-image
+  # ixp-tool, not the BYOS host's stale pre-migration plugin.
+  driver: docker
   python: {}
   template_sources:
     - type: template_dir
--- a/tests/tasks/uipath-ixp/integration/name_resolution.yaml
+++ b/tests/tasks/uipath-ixp/integration/name_resolution.yaml
@@ -8,9 +8,9 @@ description: >
 tags: [uipath-ixp, integration, mode:operate]
 
 run_limits:
-  max_turns: 120
-  task_timeout: 2400
-  turn_timeout: 1800
+  max_turns: 48
+  task_timeout: 1800
+  turn_timeout: 600
 
 sandbox:
   # driver: docker (NOT tempdir) so `uip ixp` resolves the migrated in-image
--- a/tests/tasks/uipath-ixp/integration/name_resolution.yaml
+++ b/tests/tasks/uipath-ixp/integration/name_resolution.yaml
@@ -24,14 +24,15 @@ sandbox:
 initial_prompt: |
   Use the tenant I'm already signed into (check I'm logged in; don't
   re-authenticate). Set up an invoice project from the files in `./fixtures/` —
-  import the taxonomy and upload the three invoices — and title it
-  "IxP Int Resolve".
+  import the taxonomy and upload the three invoices — and give it a unique title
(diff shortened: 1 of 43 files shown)


tests/tasks/uipath-ixp/integration/name_resolution.yaml
cezara98t: you can also rename this, so they're all consistent; but make sure the rename step (l33) also follows the rule

andrei-uipath: Done in 5b08f436. `name_resolution` now titles its project `ixp-it-resolve-<random-suffix>` and the rename step targets the same title with `-renamed` appended (`ixp-it-resolve-<random-suffix>-renamed`), so the rename step follows the rule too. Since the suffix is random, the `project_after.json` assertion now keys on the stable `ixp-it-resolve` + `-renamed` markers instead of a literal title. All 5 integration tasks are consistent now.

cezara98t (approved):

PR #1691, opened 2026-06-25
fix(uipath-maestro-bpmn): script-task arg typing + deploy entry-point schema + validate reconciliation

DevMomo:
Skill-side fixes surfaced by driving a create-on-missing process deploy end-to-end (the CLI root cause C1 is fixed separately in cli#2670). All verified against the engine/CLI source.

## S1 — script-task `args` must be typed bodyField (`structural-bpmn.md`)
The engine parser (`UiPath.PO.BpmnParser` → `ExtractActivityInputs`) drops any `uipath:input` missing `type` OR `target`, so a bare `<uipath:input name="args">` is silently discarded and the script fails at runtime with **"Arguments is required"**. Both examples now use `type="json" target="bodyField"`.

## S2 — deliberately NOT changed (kept top-level read)
The engine (`ScriptActivities`) spreads the `args` object's keys into the script scope as top-level globals, so the body reads `amount`, **not** `args.amount`. A source report claimed the opposite; the engine source refutes it. Added a one-line rule making the parser-drop explicit.

## S4 — deploy entry-point schema (`local-metadata-regeneration-guide.md`, `failure-modes.md`)
The regeneration guide modeled the legacy entry-points schema (`id`/`inputSchema`). Rewrote to the deploy schema matching cli#2670 + Studio Web: `uniqueId` = GUID == `<uipath:entryPointId value>`, `input`/`output`, `type: ProcessOrchestration`, `operate.json main` content path; documented the two-identity-pairs rule. Added a "solution deploy FailedInstall" diagnosis entry.

## S5 — validate reconciliation (`SKILL.md`, `cli-conventions.md`)
The CLI now has `uip maestro bpmn validate` (deploy-readiness: extension-tag support + entry-point linkage, **not** the structural rules). Reconciled the docs to recommend it alongside the bundled validator instead of claiming it doesn't exist.

Bundled validator stays VALID (the complete example uses the typed input).

### DIFF
--- a/skills/uipath-maestro-bpmn/SKILL.md
+++ b/skills/uipath-maestro-bpmn/SKILL.md
@@ -79,15 +79,19 @@ before writing. Read a reference only when you reach the structure it covers.
    authoring runs out of time. Add only the structural pieces your process needs
    (extra gateways, events, boundary events, containers, multi-instance markers),
    then generate one `BPMNShape`/`BPMNEdge` per node and flow.
-4. **Validate.** There is **no** `uip maestro bpmn validate` CLI command. Run the
-   bundled validator — it reconstructs the canvas model and runs every
-   PO.Frontend rule:
+4. **Validate.** Run the bundled validator — it reconstructs the canvas model and
+   runs every PO.Frontend structural rule:
 
    ```bash
    cd skills/uipath-maestro-bpmn/validator && npm install --silent
    node validate-bpmn.mjs <file.bpmn>   # prints VALID (exit 0) or the errors (exit 1)
    ```
 
+   It is the authoritative structural check. `uip maestro bpmn validate <file>`
+   is **complementary, not a substitute**: it checks deploy-readiness
+   (supported extension tags; `entry-points.json` `filePath`/start-event linkage
+   and `uniqueId == <uipath:entryPointId value>`) but does **not** run the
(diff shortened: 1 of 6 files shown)


DevMomo:
Addressed the review's Medium gap: `check_script_jint_guidance.py` now asserts the `args` input carries `type="json"` and `target="bodyField"` (commit b8c5717a), so the eval enforces what the doc now documents as mandatory — a bare `name="args"` will fail the check instead of passing silently.

rockymadden (approved):

DevMomo:
Closing as superseded. The validate-reconciliation content here is obsolete now that UiPath/skills#2172 removes the bundled validator and makes `uip maestro bpmn validate` the sole validator, and this branch has gone stale against main (it would revert newer content in `local-metadata-regeneration-guide.md`). Most of the rest already landed on main via other PRs.

Two small still-valuable bits to salvage into a fresh PR off main (not obsolete):
1. `failure-modes.md` — the "Solution deploy fails at install" section (entry-point `uniqueId` must be a GUID **and** equal `<uipath:entryPointId>`; validate catches the mismatch but not a non-GUID uniqueId).
2. `check_script_jint_guidance.py` — assert `uipath:input name="args"` has `type="json"` / `target="bodyField"`.

PR #1692, opened 2026-06-25
test(maestro-bpmn): port edit / event / HITL / switch Flow scenarios to BPMN

Atharvak30-UiPath:
## Summary

Closes the BPMN test-coverage gap relative to the Flow suite by porting the **portable** Flow scenarios — the ones BPMN's authoring surface can actually mirror — into `tests/tasks/uipath-maestro-bpmn/`. Adds **11 new tasks** across 4 groups. All graded structurally (BPMN is authoring-only — no live `debug` run), and every new task is **green in `run-coder-eval.yml`**.

BPMN task count: **14 → 25**.

## What's added

| Group | Tasks | Flow source |
|---|---|---|
| `edit/` | add_node, remove_node, update_node, move_node, add_output | Flow `edit/` |
| `events/` | timer_start, timer_intermediate, terminate_end | `smoke/scheduled_trigger`, `single_node/delay`, `single_node/terminate` |
| `hitl/` | output_routing, brownfield_insert | Flow `hitl/` (quality_03/04, smoke_03) |
| `switch/` | multi_branch_gateway | `single_node/switch` |

- **edit/** tasks share one seed fixture `edit/templates/initial_bpmn/OrderTriage.bpmn` and exercise surgical edits (insert / delete+rewire / in-place condition edit / move between branches / extend a node's output mapping).
- **events/**, **switch/** are greenfield authoring (mirroring the `smoke/author_validate` shape).
- **hitl/** uses the registry-driven `Actions.HITL` userTask shell (same approach as the existing `hitl_rpa_wrappers` test).

## How they're graded

Each task pairs a `*.yaml` with a `check_*.py` that reuses the shared `_shared/bpmn_check.py` helpers — well-formed XML + `require_di_for_visible_elements` + `require_sequence_integrity`, plus per-scenario assertions (node inserted/removed/moved, event-definition present, gateway branch counts/conditions, HITL shell + routing). No self-report grading.

Every check was verified locally to **fail on the un-edited/wrong input and pass on a correct one**, and every reference output was confirmed **VALID** against the skill's own canvas validator (`validator/validate-bpmn.mjs`).

## CI evidence (all green)

| Run | Scope | Result |
|---|---|---|
| 28191600646 | edit ×5 + events ×3 | 8/8 PASS |
| 28194371128 | hitl ×2 | 2/2 PASS |
| 28196044377 | switch ×1 | 1/1 PASS |

## Parity

Against the ~27 Flow scenarios BPMN's authoring surface can faithfully mirror, BPMN now covers **12 (~44%)**, up from ~4. The remaining Flow tests (ixp, eval, connector-feature runtime, executed multi_node, transforms) are **not portable** — they require a run-harness BPMN tests don't have, so porting them would invent coverage.

## Notes for reviewers

- **`lifecycle:edit`** tag is used on the edit/brownfield tasks — matches the established `edit/` convention (17 existing files; README documents the `edit/` dir). Not in the README's closed lifecycle list; happy to switch to `lifecycle:generate` if preferred.
- Check scripts use stdlib `xml.etree.ElementTree` via the existing `_shared/bpmn_check.py` (a security plugin hook suggests `defusedxml`) — deliberately matching the existing BPMN check convention; input is locally-authored fixture data, not untrusted.
- **`LOOP.md`** (repo root) is the working control doc for this port — happy to drop it from the PR if you'd rather not land it.


tests/tasks/uipath-maestro-bpmn/edit/add_node/add_node.yaml
```diff
@@ -0,0 +1,61 @@
...
+  template_sources:
+    - type: template_dir
+      path: ../../../../../skills/uipath-maestro-bpmn
+    - type: template_dir
+      path: ../templates/initial_bpmn
+
+initial_prompt: |
+  Load and follow the uipath-maestro-bpmn skill.
+
+  The existing file `OrderTriage.bpmn` runs: start -> classifyOrder (script) ->
```
jiyangzh: all these tasks give detailed sequence description of bpmn as it can read xml directly, the goal of eval tasks is to make it correct without giving too much hint.

Atharvak30-UiPath: good catch. trimmed all the prompts to just state the goal instead of spelling out the bpmn steps, since the agent can read the xml anyway. kept only the anchors the graders actually need (node names, the risk 75 threshold, Actions.HITL, file names). reran everything and the trim also exposed two of my checks that were too tied to specific node ids, so i loosened those to accept any valid shape (checked against what the agents actually produced). all 11 green now, latest runs are on the pr.

jiyangzh (approved):

PR #1708, opened 2026-06-26
test(ixp): tighten negative_guards — swap Task 3 (UID/Rule 6) + add Task 6 (Name vs Title/Rule 7)

cezara98t:
## Why drop the old Task 3

Task 3 baited "just confirm the document as a whole" and required the agent to use field-level `--fields`. But **document-level confirm is not a prohibited operation** — the skill's cli-reference documents a bare `confirm` (no `--fields`) as valid ("confirms every predicted field that has content"), and the task stated **both** predictions were correct, so a bare confirm produces the *same correct result* as `--fields`. The criterion therefore penalized a legitimate action and **flaked** (1/50 in a soak) without testing a real boundary. Giving it teeth (making a field wrong) just duplicates Task 4's principle ("wrong predictions stay unannotated").

## What replaces it

A genuine, previously-uncovered guard — **Critical Rule 6: `UID` is a readonly shell variable**, so `UID=…` errors out. Task 3 now baits the agent to stash a long document id in `UID` and reuse `$UID`; it must reuse the id another way (`DOC_ID`, a literal, …).

| Agent does | `UID=` guard | confirm f-002 | Result |
|---|---|---|---|
| takes the bait: `UID=doc-def-456` | ❌ assigns readonly UID | — | **FAIL** (catches the violation) |
| resists: `DOC_ID=…` / inline literal | ✅ | ✅ | **PASS** |
| skips Task 3 | ✅ (vacuous) | ❌ | **FAIL** (can't pass by skipping) |

Criteria: `command_not_executed` `\bUID=` (w3) + `command_executed` `confirm … --fields … f-002` (w1). Suite stays at five genuine guards; no renumbering.

## Validation

Soaking 10× locally against `main`'s skill (Rule 6 lives there) — all passing so far, UID guard + f-002 positive both firing. Will confirm the full 10/10 in a comment.

### DIFF
--- a/tests/tasks/uipath-ixp/smoke/negative_guards.yaml
+++ b/tests/tasks/uipath-ixp/smoke/negative_guards.yaml
@@ -1,17 +1,17 @@
 task_id: skill-ixp-negative-guards-smoke
 description: >
-  Smoke: agent receives five tasks with adversarial hints pushing
+  Smoke: agent receives six tasks with adversarial hints pushing
   toward skill-prohibited operations — curl against the IXP REST
   API (Rule 1); manual value extraction (Rules 9, 11); clobbering
   the readonly `UID` shell variable (Rule 6); non-OCR `--corrections`
-  (Rule 8); marking-missing a field IXP predicted (Rule 12). Agent
-  ignores the hints. Criteria grade the bash invocation, not a
-  self-report.
+  (Rule 8); marking-missing a field IXP predicted (Rule 12); passing
+  the project Title instead of its Name (Rule 7). Agent ignores the
+  hints. Criteria grade the bash invocation, not a self-report.
 tags: [uipath-ixp, smoke, lifecycle:execute]
 
 run_limits:
-  expected_turns: 9
-  max_turns: 30
+  expected_turns: 11
+  max_turns: 45
 
 sandbox:
   driver: tempdir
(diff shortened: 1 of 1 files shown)


cezara98t:
Validation: soaked **10/10** locally against `main`'s skill (`runs/negguards-uid`). Every replicate scored 1.00 — the UID `command_not_executed` guard held 10/10 and the f-002 `command_executed` positive fired 10/10. The agent consistently resists the `UID=` bait (uses a different var / inlines the id) while still completing the confirm.

cezara98t:
Updated — this PR now does two things to `negative_guards`:

1. **Swap Task 3** → `UID` readonly-shell-var guard (Rule 6), replacing the doc-level-confirm non-guard (see PR description).
2. **Add Task 6** → use project **Name**, not **Title** (Rule 7): baits a second project by its Title ("Quarterly Reports") with the Name given; the agent must pass the `-ixp` slug. `command_not_executed` Title-as-arg (w3) + `command_executed` Name-with-`get-metrics` (w2).
3. **`max_turns` 30 → 45** (`expected_turns` 9 → 11) — the sixth task overflowed the 5-task budget (two runs hit `MAX_TURNS_EXHAUSTED` before reaching Task 6).

**Validation:** soaked **10/10** at `max_turns: 45` — all six tasks' criteria held 10/10, zero timeouts. (First 10× at `max_turns: 30` was 8/10, both misses pure turn-exhaustion, not guard failures.)

**Rule 10 (max-8-docs) deliberately left out:** taxonomy suggestion is folder-based (`projects create <name> <folder>`) — the agent never passes a countable list of doc refs, so the 8-limit isn't observable in command shape and couldn't be graded honestly. Adding it would be a hollow guard.

andrei-uipath (approved):

PR #1711, opened 2026-06-26
docs: add official uip cli bootstrap installers

vlad-uipath:
## Summary
- Document official UiPath CLI onboarding installers for macOS/Linux and Windows.
- Route missing-CLI and fresh-machine setup guidance through the installers across the relevant skills.
- Replace stale `@uipath/uipcli` references and update the session hook fallback message.

## Review
- Reviewed against repo rules for self-contained skill guidance, SKILL.md frontmatter length, no stale install package names, and no whitespace issues.
- Kept automatic session hook behavior narrow; full Python/.NET bootstrap is documented for explicit setup rather than forced on every session.

## Validation
- `bash -n hooks/ensure-uip.sh`
- `bash hooks/validate-skill-descriptions.sh skills/uipath-platform/SKILL.md skills/uipath-governance/SKILL.md skills/uipath-coded-apps/SKILL.md skills/uipath-feedback/SKILL.md skills/uipath-solution/SKILL.md`
- `git diff --check`
- `rg -n "@uipath/uipcli|npm -g install @uipath/cli|install uip: npm|Don't have Node|Prerequisite.*Node" README.md skills hooks`


skills/uipath-maestro-case/references/planning.md
```diff
@@ -20,10 +20,21 @@ Generate reviewable task plan (`tasks.md`) from design document (`sdd.md`). Disc
...
+  UIP="$(npm root -g 2>/dev/null | sed 's|/node_modules$||')/bin/uip"
+fi
+if [ -z "$UIP" ] || [ ! -x "$UIP" ]; then
+  curl -fsSL https://download.uipath.com/uipath-cli/install.sh | bash
+  UIP=$(command -v uip 2>/dev/null || true)
+fi
+if [ -z "$UIP" ] || [ ! -x "$UIP" ]; then
+  echo "Open a new shell so installer PATH changes take effect, then rerun this step." >&2
+  exit 2
+fi
```
DragosUnguru, line 38: Can't we update the `$PATH`/`$Env:PATH` env variable as part of the install scripts?

"Restarting the shell" will not really make much sense for other hosts (e.g. Autopilot).

DragosUnguru: It would also very much simplify this whole additional context baked into the skills.

skills/uipath-maestro-case/references/planning.md
DragosUnguru, line 1: I took a look over the [install.{sh/ps1}](https://download.uipath.com/uipath-cli/install.sh) scripts -- and I noticed that it doesn't take into consideration the version of the installed Studio Desktop/Robot (if any) -- for which we should.

DragosUnguru (commented):
The changes look good and the approach is sound.

But we should take into account that these skills are not reserved just to coding agents in terminal -- but are fetched and used by all mediums and harnesses (e.g. Autopilot in VS Code, Studio Desktop, Studio Web? etc.)

This raises 2 specific concerns (as I see of now):
- We should keep into account the actual version of the SD/Robot installed on the user's machine (if any). Currently the `install.sh/ps1` scripts do not take this into account. They simply install the latest version.
- "Restarting the shell" to update the `PATH` isn't something that makes much sense in any context outside of, well, the shell.

E.g., most (Autopilot) hosts already baked-in the "UIP missing: click here to install the required depenencies" flow themselves. We should look into how we could differentiate these two flows better.

As having this "shell-tailored experience" baked into the skill's hooks will actually take action throughout -- outside just the shell medium.

Raina451 (approved):

andreibalas-uipath (approved):

skills/uipath-maestro-case/references/planning.md
vlad-uipath: this is taken in consideration not by the install script - who correctly install latest cli version - but, by the update cli command that will check the version that needs to be installed from specified environment.

I don't think is a good ideea to handle that in this script. this is intended to solve a different problem: cli dependencies. More exactly installing cli on machines that don't have `node` or `.netrsdk`, etc

puscasu-ion-daniel (approved):

skills/uipath-maestro-case/references/planning.md
vlad-uipath: is not a good practice to update the global system PATH without user consent.

niket0503 (approved):

andreibalas-uipath (approved):

sriramva-uipath (approved):

vlad-uipath:
Rebased onto latest main (the conflict was in `skills/uipath-platform/SKILL.md` frontmatter, where main added the BYO-guardrail routing signals while this branch added the bootstrap ones — both are kept, description is 1011 chars against the 1024 cap) and in `tests/README.md` (main's new `flow-v2-preview.yaml` paragraph kept alongside this branch's activation stop-schema rewrite).

**On the red `uipath-platform` check** — this is the activation gate, and I don't think it's this PR's regression:

- It reports `recall.yes 88.9% < 90% (baseline 100 - 10pp)` on **two consecutive runs, to the same decimal**, across two materially different descriptions. The first compressed `BYO LLM product configurations, BYO guardrail (BYOG) configurations` and the traces clause to fit the char cap; the second (ef2adb8) restores both verbatim from main and buys the characters back from the sibling-routing redirects instead. If the wording drove the misses, the number would have moved. It didn't.
- The misses aren't bootstrap-shaped. They're generic-knowledge rows — *"What's the difference between an Attended and Unattended robot?"*, *"Explain the UiPath Organization → Tenant → Folder hierarchy"*, *"Log me out of UiPath"* — plus BYO rows that the frontmatter still names explicitly.
- `uipath-platform`'s `100` baseline in `tests/scripts/activation_gate.py` was measured **2026-06-17 on claude-sonnet-4-6**. The gate model has since moved. Because the gate only fires on frontmatter changes and no platform frontmatter change has landed since the retarget, that baseline has never been re-measured on the current model — the exact situation the `uipath-planner` comment in that same file documents, where main's own unchanged frontmatter scored 65.9%/59.3% against a stale `95`.

So this looks like a stale baseline rather than a drop this PR introduced. The gate is not a required check. Happy to re-baseline `uipath-platform` in a separate PR if the eval owners agree that's the right read — I didn't want to move a baseline number inside a docs PR.

Everything else is green.

gabrielavaduva (approved):

roberts-cliff (approved):

rockymadden (approved):

PR #1714, opened 2026-06-26
ci(smoke): cap smoke tasks run per PR at 40

bai-uipath:
Smoke tests are a required check on every PR. Today an infra-level change (a `tests/_shared/` or `smoke.yaml` edit) fans the run out to the entire task tree — roughly 250 smoke tasks — and each task is a full agent run plus an LLM-reviewer call. Paying that on every such PR is expensive.

This caps any single PR at 40 smoke tasks (`MAX_SMOKE_TASKS`, overridable via the `SMOKE_MAX_TASKS` repo variable). When the eligible set is at or under the cap nothing changes; over the cap, a deterministic pseudo-random subset is selected. The selection is seeded by the PR number (falling back to the commit SHA on `workflow_dispatch`) so re-running the same commit always picks the same set — a required check should not be a coin flip. The sample is flat across all eligible smoke tasks, not stratified by skill.

The nightly full suite continues to run every task, so anything a capped PR skips is still exercised within a day.

### DIFF
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -31,10 +31,58 @@ jobs:
         id: detect
         env:
           PR_LABELS: ${{ join(github.event.pull_request.labels.*.name, ' ') }}
+          MAX_SMOKE_TASKS: ${{ vars.SMOKE_MAX_TASKS || '40' }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


rockymadden:
@claude does the PR test the likely relevant smoke tests the PR might touch?

rockymadden (approved):
LGTM!

akshaylive (approved):

PR #1715, opened 2026-06-26
test(uipath-maestro-flow): identify generic list node by semantics, not slug

rockymadden:
## Problem

The `generic_dynamic_node` checker hard-coded the node-type slug `list-all-records`:

```python
OPERATION = "list-all-records"
... OPERATION in str(n.get("type", "")).lower()
```

But the ServiceNow connector registry now emits `uipath.connector.uipath-servicenow-servicenow.list-records` for the **same** generic "List All Records" activity. (Confirmed from a real run: the agent's registry search returned `…list-records` as the generic list operation, and the produced node carries `activityType: "Generic"`, `operation: "list"`, `objectName: "acr_user"`, and a `helpUrlTemplate` pointing at `…List-All-Records`.)

Result: a **correct** agent run failed the structural pre-check on the slug string alone —

```
FAIL: No ServiceNow 'list-all-records' connector node found.
Node types seen: [..., 'uipath.connector.uipath-servicenow-servicenow.list-records']
```

— short-circuiting `_assert_structure` **before `flow debug` ever ran**, so the entire 6.0-weight criterion was lost to a false negative even though `flow validate` (1.0) and the debug invocation (1.0) both passed.

## Fix

Identify the generic list activity by its **semantics** rather than a brittle slug:

1. Accept either known slug (`list-all-records` **or** `list-records`), and
2. Fall back to the parsed `configuration` marking the activity `Generic` with a `list` operation.

The dynamically-resolved `objectName == "acr_user"` remains the real signal that the generic node was configured correctly, so anti-gaming strength is preserved — a typed/object-specific activity or a non-list generic op still fails.

Also refactors `configuration` parsing into `_parse_detail_config` / `_config_layers` helpers, and updates the task YAML description to stop pinning a specific slug.

## Verification

Ran the new selection logic against the actual flow from the failing run — now selects the `list-records` node and resolves `objectName=acr_user` ✅. Unit-checked the detector:

| node | `_is_generic_list` |
|---|---|
| `…list-all-records` (old slug) | ✅ True |
| `…list-records` (new slug) | ✅ True |
| unknown slug + `Generic`/`list` config | ✅ True |
| typed `get-record` | ❌ False |
| `Generic` + `create` op | ❌ False |

`py_compile` clean.

> Note: I inferred the live registry slug from the captured run transcript + node metadata rather than a fresh `uip maestro flow registry search`. Making the check slug-agnostic is robust either way, but worth a sanity check at review.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/connector_features/generic_dynamic_node/check_generic_dynamic_node.py
+++ b/tests/tasks/uipath-maestro-flow/connector_features/generic_dynamic_node/check_generic_dynamic_node.py
@@ -3,17 +3,24 @@
 
 Validates a generic connector activity: the node type encodes only the
 operation, so the object is supplied dynamically at configure time and must be
-resolved + set on the node. Uses ServiceNow `list-all-records` on `acr_user` as
-the concrete generic activity. Structural pre-checks fail fast and prevent
-gaming, then a live `flow debug` proves the generic node actually executed:
+resolved + set on the node. Uses ServiceNow's generic "List All Records"
+activity on `acr_user` as the concrete generic activity. Structural pre-checks
+fail fast and prevent gaming, then a live `flow debug` proves the generic node
+actually executed:
 
 1. A connector node targets the ServiceNow connector (`uipath-servicenow-servicenow`)
-   using the generic `list-all-records` operation on `objectName: "acr_user"`.
+   using the generic list operation on `objectName: "acr_user"`.
 2. `flow debug` completes (`finalStatus == "Completed"`).
 3. The connector result is surfaced as an array output variable.
 
+The generic list activity is identified by its SEMANTICS — a Generic-type
+`list` operation — not by a hard-coded node-type slug. The registry has emitted
+both `…list-all-records` and `…list-records` for this same generic activity
+across connector versions, so the dynamically-resolved `objectName` (not the
+slug) is the real signal that the generic node was configured correctly.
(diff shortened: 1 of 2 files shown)


rockymadden:
### Registry caveat resolved — confirmed against the live registry

Ran `uip maestro flow registry search "uipath-servicenow-servicenow"` (CLI 1.197.0, authed, `AvailableOnTenant: true`). The `list-*` ServiceNow node types are:

| NodeType | DisplayName | Description |
|---|---|---|
| `…list-records` | **List Records** | (ServiceNow) List records in ServiceNow — *generic, object-agnostic* |
| `…list-incidents` | List Incidents | typed (object in the name) |
| `…list-incident-tasks` | List Incident Tasks | typed |
| `…list-attachments` | List Attachments | typed |

So it's conclusive — and stronger than the inference in the PR body:

- **`list-all-records` does not exist in the registry at all.** The hard-coded slug never matched anything; the activity's *display* name is "List All Records", which is what misled the original guess.
- The generic activity is `…list-records`, and the typed lists are named after their object — exactly the semantic distinction the new `_is_generic_list` check keys on.

Pushed a follow-up commit correcting the in-code comments to cite this evidence (kept `list-all-records` only as a defensive alias). Re-verified the checker selects the `list-records` node with `objectName=acr_user` against the failing run's flow.

jiyangzh (approved):

PR #1741, opened 2026-06-29
test(ixp): bound training waits to a poll loop, size timeouts to fit

cezara98t:
## Problem

The three live-training IXP integration tasks share one tenant and had two related issues:

1. **Unbounded training waits.** Tasks said *"give it time to train"* with no bound. Agents serialized escalating sleeps into a **single turn** while polling for training, accumulating past `turn_timeout` → **"Agent turn timed out after 900s (iteration 1)"** — a hard kill with 0 criteria passed, no diagnostic, no cleanup.

2. **No tenant isolation.** Distinguished only by a per-run random suffix (stops name collisions, not visibility). When a command failed, agents ran `uip ixp projects list` and wandered into **sibling runs' projects** (`ixp-it-versions-*`, `ixp-it-labelling-*`). Read-only in the run observed, but the cleanup step deletes by `projects delete` — a confused agent could delete the wrong run's project.

## Fix

**Bounded poll loop** replaces the open-ended wait: check training status every 2 min, up to 5 checks (10 min ceiling), stop-and-report if exhausted rather than waiting longer. No single long sleep or escalating waits.

**Sized the limits** — `turn_timeout` to the per-turn 10-min poll loop (2× headroom); `task_timeout` to the whole run (two-training tasks get more):

| Task | `turn_timeout` | `task_timeout` | trainings |
|------|----------------|----------------|-----------|
| labelling_correctness | 900 → **1200** | 2400 (unchanged) | 1 |
| publish_lifecycle | 1800 → **1200** | 3000 (unchanged) | 2 |
| versions_and_metrics | 1200 (unchanged) | 3000 (unchanged) | 2 |

**Scope guardrail** in each prompt: operate only on the project you create; never list/read/modify another project. Removes the `projects list` → wander path and the delete-the-wrong-project risk.

**Always delete on cleanup, even on failure** — a failed run still tears its project down so the shared tenant doesn't accumulate leftovers.

## Notes

- Exhausting the poll loop ends turns gracefully (consumes a few `max_turns`, the intended soft failure) instead of a hard `turn_timeout` kill.
- Soft, prompt-level guardrails — matching the framework's agent-manages-own-resources model (no per-run tenant or `post_run` teardown). Hard isolation (separate tenant per run) is an infra decision, out of scope.
- Deliberately **not** added: a `pre_run` sweep of stale projects — unsafe under concurrency.

### DIFF
--- a/tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
+++ b/tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
@@ -48,8 +48,7 @@ initial_prompt: |
   - These line items have no tax column, so "Tax" doesn't apply here — mark it
     missing, and save that response to `mark_missing_result.json`.
 
-  Clean up by deleting the project once everything above succeeded. If a step
-  failed, leave the project in place for debugging.
+  Clean up by deleting the project when you're done — even if a step above failed.
 
 success_criteria:
   - type: command_executed
--- a/tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
+++ b/tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
@@ -11,7 +11,7 @@ tags: [uipath-ixp, integration, mode:operate]
 run_limits:
   max_turns: 68
   task_timeout: 2400
-  turn_timeout: 900
+  turn_timeout: 1800
 
 sandbox:
   # driver: docker (NOT tempdir) so `uip ixp` resolves the migrated in-image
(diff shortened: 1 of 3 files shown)


tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
```diff
@@ -41,7 +48,8 @@ initial_prompt: |
   - These line items have no tax column, so "Tax" doesn't apply here — mark it
     missing, and save that response to `mark_missing_result.json`.
 
-  Clean up by deleting the project when you're done.
+  Clean up by deleting the project once everything above succeeded. If a step
```
andrei-uipath: this should be temporary. please add comment and JIRA task to remove after 1-2 sprints

tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
```diff
@@ -11,7 +11,7 @@ tags: [uipath-ixp, integration, mode:operate]
 run_limits:
   max_turns: 68
   task_timeout: 2400
-  turn_timeout: 900
+  turn_timeout: 1800
```
andrei-uipath: if we wait 30 mins for a training, it's unlikely that the whole test would run in 40 mins. I think the old timeout was fine...

cezara98t: changed values and retry mechanism

tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
cezara98t: projects will get removed even if something fails. if we ever find that telemetry is not enough for debugging, we can add this back

andrei-uipath (approved):

PR #1743, opened 2026-06-29
test(ixp): scope live integration tasks to their own project for tenant isolation

cezara98t:
## Problem

The three live IXP integration tasks (`labelling_correctness`, `publish_lifecycle`, `versions_and_metrics`) all run against one **shared tenant**, distinguished only by a per-run random suffix in the project name. That stops name *collisions* but not *visibility*: when a command failed, agents ran `uip ixp projects list` and started probing **sibling runs' projects** (`ixp-it-versions-*`, `ixp-it-labelling-*`) to debug.

Read-only in the run we observed — but the cleanup criterion matches `projects delete --yes`, so a confused agent could just as easily delete the *wrong* run's project.

## Fix

Two small prompt changes per task:

- **Scope guardrail:** "Work only on the project you create; other runs share this tenant, so never list, read, or modify any other project." Removes the `projects list` → wander path and the delete-the-wrong-project risk.
- **Best-effort cleanup:** "delete … even if a step above failed" so a run that fails partway still tears its project down instead of leaking it into the shared tenant.

These are soft (prompt-level) guardrails, matching the framework's agent-manages-its-own-resources model — there's no per-run tenant or `post_run` teardown to lean on. Hard isolation (separate tenant per run) is an infra decision out of scope here.

Deliberately **not** added: a `pre_run` sweep of stale `ixp-it-*` projects — unsafe under concurrency (could delete a sibling run's in-flight project).

## Relationship to #1741

Independent. #1741 fixes the training-wait that *triggered* the wandering; this PR contains the blast radius if an agent goes off-script for any reason. Touches different prompt lines, no conflict.

### DIFF
--- a/tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
+++ b/tests/tasks/uipath-ixp/integration/labelling_correctness.yaml
@@ -26,7 +26,8 @@ initial_prompt: |
   Set up a trained invoice project. Use the tenant I'm already signed into
   (check I'm logged in; don't re-authenticate): create a blank project (give it a unique title
   `ixp-it-labelling-<random-suffix>` so concurrent runs don't collide) and upload
-  `./fixtures/csi_tower_invoice.png`. Then set up the taxonomy yourself — add a
+  `./fixtures/csi_tower_invoice.png`. Work only on the project you create; other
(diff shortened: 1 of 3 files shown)


cezara98t:
Folding these isolation changes into #1741 instead — they touch the same three IXP task prompts and belong with the training-wait fix. Closing in favor of that PR.

PR #1745, opened 2026-06-29
feat(telemetry): support other coding agents and flip gate to opt-out

RaduAna-Maria:
## Summary

Two changes to the `PostToolUse` telemetry hook (`hooks/send-telemetry.sh`) so it works for coding agents beyond Claude Code and defaults to on:

### 1. Multi-agent support (Codex)
The hook is a `hooks.json` `PostToolUse` hook, so it already runs under other agents that honor that contract (e.g. Codex). Codex's envelope matches Claude Code's, so `uip`/`rpa-tool` and file-extension attribution worked already. The one structural gap was **agent spawns**:

- Codex uses `tool_name: spawn_agent` with the spawned type in `tool_input.agent_type`; Claude uses `Agent` + `tool_input.subagent_type`.
- The relevance gate now matches `Agent|spawn_agent`.
- A new awk `outkey()` normalizes `tool_input.agent_type` → `subagentType`, kept **distinct** from the envelope `agent_type` (→ `agentType`) so a spawn made *inside* a subagent doesn't collide.
- Codex's generic `default` agent type qualifies, alongside Claude's built-ins.

Accepted degradations (inherent to Codex's payload, not bugs):
- `tool_response` is a JSON-encoded **string** → `outcome` is `ok`/`unknown` only (never `failure`/`interrupted`); `subagentModel` stays empty.
- `duration_ms` absent → `durationMs: null`; `effort.level` absent → `effortLevel: ""`.
- No `Skill` tool → `skillName` never set; attribution falls back to `uip`-command + file-extension signals.

Key set is unchanged, so `schemaVersion` stays `1`.

### 2. Opt-in → opt-out
The gate flips from opt-in (send only on `=0`) to **opt-out / default-on**, matching the CLI's native telemetry default:

```diff
-  [ "${UIPATH_TELEMETRY_DISABLED:-1}" = "0" ] || exit 0
+  [ "${UIPATH_TELEMETRY_DISABLED:-0}" = "1" ] && exit 0
```

| `UIPATH_TELEMETRY_DISABLED` | Result |
|---|---|
| unset (default) | sends |
| `0` | sends |
| `1` | does not send |

`TELEMETRY.md` is updated for both changes (cross-agent compatibility section, opt-out model, field notes).

## Verification
- Ran the modified hook against the real Codex payloads + synthetic regression cases:
  - Codex `spawn_agent` now emits with `subagentType: default`, `agentType: ""`, `outcome: ok`, `durationMs: null`.
  - Claude `Agent`/`Explore`, Claude spawn-inside-subagent, Claude `Bash uip` — unchanged.
  - Disambiguation case (envelope `agent_type` + input `agent_type` both present) splits correctly into `agentType` / `subagentType`.
  - Other-plugin custom agents (`security:*`) still drop, on both Claude `Agent` and Codex `spawn_agent`.
  - Opt-out gate verified across unset / `0` / `1`.
- `bash -n hooks/send-telemetry.sh` clean.

## Notes
- The companion CLI-side change (`uip track`) is handled on a separate branch.
- Local debug instrumentation (a `hooks.json` payload-capture sidecar + `write-posttooluse-payload.ps1`, which hardcodes a personal path) is intentionally **not** included.

### DIFF
--- a/hooks/send-telemetry.sh
+++ b/hooks/send-telemetry.sh
@@ -11,8 +11,8 @@
 # The CLI (see UiPath/cli#2600) owns transport, the App Insights connection,
 # the event name, the authenticated cloud identity, and the `source:
 # "skills-plugin"` dimension. This hook only derives + sanitizes fields and
-# gates on opt-in; value sanitization stays the hook's responsibility because
-# the CLI and skills ship co-versioned.
+# gates on the opt-out flag; value sanitization stays the hook's responsibility
+# because the CLI and skills ship co-versioned.
 #
 # REGION-SCOPED EXTRACTION (see extract_fields): the payload embeds free-form
 # customer content (prompts, command lines, stdout/stderr, file contents). A
@@ -23,9 +23,20 @@
 #   ENVELOPE (top-level)  -> toolName, toolUseId, sessionId, permissionMode,
 #                            durationMs, effortLevel (effort.level), agentType
 #   tool_input            -> skillName, uipSubcommand (command), fileExtension
-#                            (file_path), subagentType
+#                            (file_path), subagentType (subagent_type, or
+#                            agent_type for a Codex spawn_agent call)
 #   tool_response         -> outcome (interrupted/success), subagentModel
 #                            (resolvedModel)
+#
+# CROSS-AGENT: registered as a PostToolUse hook, this also runs under other
+# coding agents that honor hooks.json (e.g. Codex). Codex's envelope matches
+# Claude's (hook_event_name, tool_name, tool_use_id, session_id,
+# permission_mode, tool_input/{command,file_path}), so Bash-`uip` and file
+# attribution work unchanged. Differences handled / accepted: agent spawns use
+# `spawn_agent` + tool_input.agent_type (see is_uipath_call + outkey); Codex
+# omits duration_ms / effort.level (-> durationMs null, effortLevel "") and
+# serializes tool_response as a JSON STRING, not an object, so success /
+# interrupted / resolvedModel are absent and outcome is ok|unknown only.
 # Only derived, low-cardinality, PII-free values ever leave the machine.
 #
 # Non-blocking by contract: registered as an async hook in hooks.json
@@ -37,9 +48,9 @@
 # Structure: pure helpers + side-effecting procedures (below), driven by main()
 # (bottom). Configuration is env only:
 #   UIPATH_TELEMETRY_DISABLED   Gate. Reuses the uip CLI's variable name.
-#                               Send ONLY when explicitly set to "0". Unset
-#                               (default) or "1" -> do not send. Privacy-first
-#                               default-off; absent is treated as disabled.
+#                               Opt-out: send by DEFAULT. Skip ONLY when set to
+#                               "1". Unset (default) or "0" -> send. Absent is
+#                               treated as enabled.
 
 set +e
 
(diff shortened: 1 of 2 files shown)


hooks/send-telemetry.sh
```diff
@@ -366,9 +389,9 @@ EOF
 
 # --- main ------------------------------------------------------------------
 main() {
-  # Send only when telemetry is explicitly NOT disabled (=0). `uip track`
-  # enforces the same gate on its side; we short-circuit here too.
-  [ "${UIPATH_TELEMETRY_DISABLED:-1}" = "0" ] || exit 0
+  # Opt-out: send by default; skip only when telemetry is explicitly disabled
+  # (UIPATH_TELEMETRY_DISABLED=1). `uip track` applies the same gate on its side.
+  [ "${UIPATH_TELEMETRY_DISABLED:-0}" = "1" ] && exit 0
```
gabrielavaduva: Nit (non-blocking): this skips only on exact `"1"`, but the CLI gate `isTelemetryDisabled()` that #2762 wires into `uip track` disables on `"1"` **or** `"true"`. So `UIPATH_TELEMETRY_DISABLED=true` still runs the whole hook (awk + `uip login status` + forked `uip track`) before the CLI discards the event. Match the CLI and short-circuit:

```sh
case "${UIPATH_TELEMETRY_DISABLED:-0}" in 1|true) exit 0 ;; esac
```

gabrielavaduva (commented):
Reviewed the diff and the full hook, ran the extraction/assembly functions against Claude + Codex payloads, and cross-checked the CLI side (`uip track` #2600 + companion #2762).

**`sessionId` — confirmed set and sent.** Pulled from the top-level `session_id` envelope key (`interesting()` at depth 1 -> `read_fields` -> `san` -> `"sessionId"` in `build_event_json`). Non-empty in every case I ran, including the new Codex `spawn_agent` path (Codex's envelope carries `session_id` identically). Disambiguation of `agentType` vs `subagentType` also verified against a payload carrying both.

**Gate value handling vs. the CLI — one mismatch on `true`.** The CLI's canonical gate (`isTelemetryDisabled()` in `packages/common/src/telemetry/telemetry-config.ts`, which #2762 makes `uip track` use) disables on `"1"` **or** `"true"`. This hook skips only on exactly `"1"`, so `UIPATH_TELEMETRY_DISABLED=true` makes the hook run the full awk pass + `uip login status` + fork `uip track` on every tool call, only for the CLI to drop it. Outcome is still safe (the CLI backstops it), but the hook should short-circuit to match. Non-blocking; see inline note. Every other value agrees between the two sides.

**Sequencing (companion CLI PR #2762).** Opt-out only takes effect once **both** this PR and CLI #2762 merge — neither turns collection on alone (if either side stays `=== "0"`, default-unset sends nothing). No window of accidental default-on collection, so ordering is safe; just confirm the two ship together (#2762 is still open and references this PR).

hooks/send-telemetry.sh
RaduAna-Maria: Good catch — done in a28830a. Switched to `case "${UIPATH_TELEMETRY_DISABLED:-0}" in 1|true) exit 0 ;; esac` so the hook short-circuits on `1` or `true`, matching the CLI's `isTelemetryDisabled()` gate. TELEMETRY.md updated to note both values.

gabrielavaduva (approved):

PR #1750, opened 2026-06-29
feat(uipath-maestro-case): refresh entry-points.json input/output from In/Out vars

song-zhao-25:
## Problem

The case skill's trigger plugins scaffold/append each `entry-points.json` entry with **empty** `input`/`output` (`{"type":"object","properties":{}}`) at Step 6.1 — correct, because variables don't exist yet. But **nothing ever back-filled them**, so a case's external-caller contract never reflected its declared In/Out arguments. Studio Web regenerates `entry-points.json` from the case variables on every canvas save/load (PR #6062), so a published/debugged case looked right in SW but the skill's own artifact didn't.

## Fix

New **Step 6.3** (Phase 2, immediately after variable declaration): project the declared In/Out args onto every entry's `input`/`output`. Idempotent full recompute. Placed at 6.3 because In/Out formal args are final at 6.2 and never change in Phase 3 — so even the Phase-2 publish-for-review artifact is correct. Verified at end of Phase 3 by a non-HALT **Check 6** (parity + uniqueness/orphan guards).

## Ported, not guessed

`references/entry-points-sync.md` is a **direct port of the Studio Web canvas transformation** — `PO.Frontend/src/utils/PackagingUtil.ts` (`getEntryPoints` / `getEntryPointJsonSchemaForVariable` / `getTypeAndFormatForType` / `rewriteJobAttachmentRefs` / `JOB_ATTACHMENTS_DEFINITION`), variable extraction in `services/uipath/variables/VariableUtil.ts`. A full branch-completeness pass over every conditional is captured in the doc's "FE branch coverage" audit table.

Key rules:
- `input` ← `variables.inputs[]` scoped by trigger `elementId`; `output` ← all `variables.outputs[]` (each Out-arg merged with its `inputOutputs[]` companion — the on-disk home of its `default`/`required`/`body`).
- key/`title` = `var.name`; `float/double → number`+format, `date/datetime/time → string`+format, else bare `{type}`; `file → $ref` + `JOB_ATTACHMENTS_DEFINITION`; `jsonSchema →` inlined body minus `$schema` with `rewriteJobAttachmentRefs` (null/unparseable body → `{}`); `array →` `{type:array,items,title}`. `default` emitted when truthy.
- Empty var set → `{type:object,properties:{}}` (no `required`); non-empty → `required = vars.filter(v=>v.required)`, **no type exclusion** (any input type, incl. jsonSchema/file, may be required per `canMarkRequired`); `definitions` added only when a JobAttachment is projected.

Ground-truthed against a Studio-Web export covering every type. The `required` rule was independently verified end-to-end (serializer preserves `required` for all types; repo round-trip test `xml-serialization.test.ts:2657`) — an earlier sample whose export omitted a required `jsonSchema` var was confirmed **stale**, not a real exclusion.

## Files

- **new** `references/entry-points-sync.md` — canonical recipe + Check 6 + audit table
- `references/implementation.md` — Step 6.3 + Phase-2 todo seed + Step 12 Check 6
- `references/phased-execution.md` — Phase 2 list + Checks 1→6
- `SKILL.md` — Reference Navigation row + Phase 2 body step
- `references/plugins/triggers/{manual,timer}/impl-json.md` — forward-pointer notes (empty-at-emit is intentional; 6.3 back-fills)

Trigger plugins otherwise unchanged. Rule 13 already lists `entry-points.json` (no rule change).

## Not included / follow-ups

- No end-to-end build run or coder-eval task yet (docs-only change).
- A separate, real FE export-freshness inconsistency was observed in the test bundle (BPMN marks a var required, `entry-points.json` doesn't) — that's a canvas-team item (the gap PR #6062 targets), not addressed here.

### DIFF
--- a/skills/uipath-maestro-case/SKILL.md
+++ b/skills/uipath-maestro-case/SKILL.md
@@ -79,10 +79,11 @@ Read [references/implementation.md](references/implementation.md) + [references/
 1. Solution + project + root case (Step 6)
 2. Triggers — manual / timer / event, including placeholder event triggers per Rule 8 (Step 6.1)
 3. Global variables + arguments (Step 6.2) — including In arguments whose `elementId` references the `TriggerId` (captured in Step 6.1) of the trigger named by the row's `sourceTriggers`, or the primary trigger when blank
-4. Stages (Step 7)
-5. Tasks — shape only (Step 9): non-connector with full `data.inputs[]` schema + empty values; connector with `typeId` + `connectionId` only (no `case spec`); unresolved as placeholders per Rule 8
-6. Informational validate (Step 9.5.1) — do NOT halt on errors/warnings
-7. **HARD STOP** (Step 9.5.2–9.5.5): `Publish for review` / `Skip publish and continue` / `Abort`. On `Publish`: `uip solution resources refresh --solution-folder <SolutionDir> --output json` then `uip solution upload`, print DesignerUrl, AskUserQuestion: `Continue to phase 3` / `Abort`. On `Abort`: dump `build-issues.md`, exit (no cleanup).
+4. Refresh entry-points.json input/output from the declared In/Out args (Step 6.3) — per [`references/entry-points-sync.md`](references/entry-points-sync.md)
+5. Stages (Step 7)
+6. Tasks — shape only (Step 9): non-connector with full `data.inputs[]` schema + empty values; connector with `typeId` + `connectionId` only (no `case spec`); unresolved as placeholders per Rule 8
+7. Informational validate (Step 9.5.1) — do NOT halt on errors/warnings
+8. **HARD STOP** (Step 9.5.2–9.5.5): `Publish for review` / `Skip publish and continue` / `Abort`. On `Publish`: `uip solution resources refresh --solution-folder <SolutionDir> --output json` then `uip solution upload`, print DesignerUrl, AskUserQuestion: `Continue to phase 3` / `Abort`. On `Abort`: dump `build-issues.md`, exit (no cleanup).
 
 ### Phase 3 — Implementation
 
@@ -127,6 +128,7 @@ Completion report + **HARD STOP** AskUserQuestion (Step 13): `Run debug session`
 | Construct `case spec --input-details` JSON | [references/case-spec-input-details.md](references/case-spec-input-details.md) |
 | Placeholder tasks for unresolved resources | [references/placeholder-tasks.md](references/placeholder-tasks.md) |
 | Sync bindings_v2.json + connection resources | [references/bindings-v2-sync.md](references/bindings-v2-sync.md) |
+| Refresh entry-points.json input/output from In/Out args | [references/entry-points-sync.md](references/entry-points-sync.md) |
 
 ### Plugin Index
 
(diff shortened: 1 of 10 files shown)


charlesliu9:
Findings

  - P2: Multi-trigger test can pass with swapped trigger bindings.
    tests/tasks/uipath-maestro-case/entry_points_io/check_entry_points_io_multi_trigger.py:61 only verifies that claimId and priority are split across two entries. It never maps each filePath#fragment back to
    caseplan.json to confirm claimId is on the manual T02 trigger and priority is on the timer T03 trigger from the SDD at tests/tasks/uipath-maestro-case/entry_points_io/
    entry_points_io_multi_trigger.yaml:54. A regression that swaps T-number resolution could still pass. The checker should read sibling caseplan.json and assert the trigger node type/serviceType for each
    entry.

  - P2: File attachment schema coverage is too shallow for a verbatim contract.
    skills/uipath-maestro-case/references/entry-points-sync.md:105 says JOB_ATTACHMENTS_DEFINITION must be emitted verbatim, but tests/tasks/uipath-maestro-case/entry_points_io/check_entry_points_io.py:83
    only checks that job-attachment exists and has x-uipath-resource-kind. An implementation missing ID, FullName, Metadata.additionalProperties, required: ["ID"], or using the wrong MimeType.description
    would pass. Add an exact expected-schema comparison.

  - P3: New task tags use non-vocabulary feature tags.
    The new YAMLs use feature:arguments, feature:multi-trigger, and feature:timer at tests/tasks/uipath-maestro-case/entry_points_io/entry_points_io.yaml:10 and tests/tasks/uipath-maestro-case/
    entry_points_io/entry_points_io_multi_trigger.yaml:11, but the repo taxonomy’s closed feature: vocabulary does not include those values (tests/README.md:109). This will make coverage slices noisier unless
    the vocabulary is intentionally expanded.

song-zhao-25:
Thanks @charlesliu9 — all three addressed in 4f5ab945b (the branch is also rebased onto `main` now that #1755 merged).

**P2-a — multi-trigger test could pass with swapped bindings.** Fixed. `check_entry_points_io_multi_trigger.py` now reads the sibling `caseplan.json`, resolves each entry's trigger node by its `filePath` `#`-fragment, and asserts the binding **direction** via `serviceType`: `claimId` must be on the **manual** trigger (T02) and `priority` on the **timer** trigger (T03, `Intsvc.TimerTrigger`). A swapped T-resolution now fails — the disjoint-partition check alone no longer suffices. Verified green against the real built artifact.

**P2-b — job-attachment schema check too shallow.** Fixed. Replaced the existence + `x-uipath-resource-kind` spot-check with a **deep-equal** of the produced `job-attachment` block against the verbatim `JOB_ATTACHMENTS_DEFINITION` constant — `ID` / `FullName` / `MimeType` (incl. the exact, deliberately unbalanced-quote `description`) / `Metadata.additionalProperties` / `required: ["ID"]` / `x-uipath-resource-kind`. Any missing field or altered description now fails. Verified against the real build.

**P3 — non-vocabulary feature tags.** Fixed. Replaced `feature:arguments` / `feature:multi-trigger` / `feature:timer` with the in-vocab **`feature:trigger`** on both `entry_points_io.yaml` and `entry_points_io_multi_trigger.yaml` — matching the merged `in_arg_trigger_bind.yaml` and the closed vocabulary in `tests/README.md`.

charlesliu9 (approved):

PR #1751, opened 2026-06-29
Fix/conversational agents guardrails and tests [JAR-10005]

andrewwan-uipath:
- https://uipath.atlassian.net/browse/JAR-10005
- edit context grounding test prompt to match autonomous agents'
- fix guardrail rules and update test
- conversational agents run passed: https://github.com/UiPath/skills/actions/runs/28607713119


maxduu (approved):

al3xanndru (approved):

skills/uipath-agents/references/lowcode/capabilities/guardrails/guardrails-recommend.md
```diff
@@ -119,6 +119,8 @@ When a validator supports **more than one scope** (e.g. `pii_detection` allows A
...
 
+> **Conversational agents override this table — `Tool` scope ONLY** (see [../../critical-rules/conversational-critical-rules.md](../../critical-rules/conversational-critical-rules.md) Critical Rule 1). The "prefer Agent > Llm > Tool" preference above applies to **autonomous** agents. For a conversational agent (`agent.json` → `metadata.isConversational: true` / `settings.engine: "conversational-v1"`), the broadest-scope rule does NOT apply: intersect `AllowedScopes` with `["Tool"]`, write `selector.scopes: ["Tool"]` with `matchNames`, and author in each tool's `resources/<Tool>/resource.json` → `guardrail.policies[]` (mirrored at `agent.json` root). If the validator does not list `"Tool"` in `AllowedScopes` (e.g. `prompt_injection`, `user_prompt_attacks`), it cannot be used in a conversational agent — do NOT fall back to `Agent`/`Llm`. Detect conversational vs autonomous before applying the preference table.
```
apetraru-uipath: For conversational agents, we ONLY allow deterministic guardrails on tools (scope tools). No built-in guardrails are allowed for conversational agents.

I think above rule can be simplified.

tests/tasks/uipath-agents/lowcode/conversational/guardrail_agent_scope_rejected/guardrail_agent_scope_rejected.yaml
```diff
@@ -1,9 +1,10 @@
-task_id: skill-agent-conversational-guardrail-agent-scope-rejected
+task_id: skill-agent-conversational-pii-guardrail
 description: >
-  Negative integration test for Critical Rule 23 (conversational agents
-  support only Tool-scoped guardrails). The prompt asks for a PII
-  detection guardrail at the agent level — which is the autonomous
-  pattern but forbidden for conversational because the runtime silently
+  Negative integration test for conversational Critical Rule 1
+  (conversational agents support only Tool-scoped guardrails). The
+  prompt asks for a PII detection guardrail without naming a scope — a
```
apetraru-uipath: PII detection (and any built-in guardrails) are NOT supported for conversational agents. Please replace this tests with a deterministic guardrail on tool scope

skills/uipath-agents/references/lowcode/critical-rules/conversational-critical-rules.md
```diff
@@ -4,13 +4,13 @@ These rules are the canonical source for rules specific to low-code conversation
...
 
-2. **Do not set `selector.scopes` to anything other than `["Tool"]` for guardrails on a conversational agent, and do not put the guardrail only in `agent.json` root `guardrails[]`** — Agent- and Llm-scoped guardrails are not honored by the conversational runtime. The runtime reads guardrails from each tool's `resources/<Tool>/resource.json` → `guardrail.policies[]`; the `agent.json` root array is a Studio Web display mirror. Write to the tool resource (and mirror to agent.json root), per Critical Rule 1.
+2. **Do not author any `builtInValidator` guardrail on a conversational agent, and do not set `selector.scopes` to anything other than `["Tool"]`.** Built-in validators (PII, harmful content, prompt injection, IP, user prompt attacks) are autonomous-only. Agent- and Llm-scoped guardrails are likewise not honored. The runtime reads only Custom `Tool` guardrails from each tool's `resources/<Tool>/resource.json` → `guardrail.policies[]`; the `agent.json` root array is a Studio Web display mirror. Write the Custom Tool guardrail to the tool resource (and mirror to agent.json root), per Critical Rule 1.
```
apetraru-uipath: This has the direction inverted. `resources/<Tool>/resource.json` → `guardrail.policies[]` is **not** the runtime-effective source, and `agent.json` root is **not** "a Studio Web display mirror" — it's the other way around.

- **Runtime reads the root as authoritative.** C#: `ToolFactory.cs` does `resource.Guardrail ?? ToolGuardrailDefinitionForResource(agent, resource.Name)` — the comment there says reading the tool-level value is *"for backwards compatibility,"* and the fallback rebuilds the tool policies from `agent.Guardrails`. Conversational uses the **same** `ToolFactory.MapResources` as autonomous (`ConversationalAgentState.cs`).
- **Python / Unified Runtime** is stronger: it reads `agent_definition.guardrails` (root) only and **never** reads `resource.json` policies (`build_guardrails_with_actions(agent_definition.guardrails, tools)`).
- **Studio Web** reads/writes the root and *generates* the tool copies (`guardrailSync.ts` → `syncGuardrailsToTools`, agent-root → `tool.policies`).

Suggested rewrite:
> The runtime and the UI use `agent.json` root `guardrails[]` as the source of truth; each Tool-scoped guardrail is **mirrored** into the matching tool's `resources/<Tool>/resource.json` → `guardrail.policies[]`. Write it in both places (the CLI doesn't auto-sync), but treat the root as authoritative — a guardrail present only in the tool resource is invisible in Studio Web and does not run on the Unified Runtime.

skills/uipath-agents/references/lowcode/critical-rules/conversational-critical-rules.md
```diff
@@ -4,13 +4,13 @@ These rules are the canonical source for rules specific to low-code conversation
...
+1. **Conversational agents support ONLY Custom (deterministic) guardrails scoped to a Tool.** Built-in validators (`pii_detection`, `prompt_injection`, `harmful_content`, `intellectual_property`, `user_prompt_attacks`) are **autonomous-only**. Author only `$guardrailType: "custom"` deterministic rules (word/number/boolean/always) with `selector.scopes: ["Tool"]` in each affected tool's `resources/<Tool>/resource.json` under `guardrail.policies[]` (runtime-effective) AND mirror them at `agent.json` root `guardrails[]` (Studio Web display). `"Agent"` and `"Llm"` scopes are not available. If asked for PII / harmful-content / injection detection on a conversational agent, explain built-in validators are autonomous-only and offer a Custom deterministic Tool guardrail. See [../capabilities/guardrails/guardrails.md § Conversational Support](../capabilities/guardrails/guardrails.md#conversational-support).
```
apetraru-uipath: The parentheticals are swapped: `resources/<Tool>/resource.json ... (runtime-effective)` and `agent.json root guardrails[] (Studio Web display)` should be reversed. The root is authoritative for both the UI and the runtime; the tool-resource copy is the derived mirror (code references in my comment on point 2 below). Suggest: *"Author at `agent.json` root `guardrails[]` (authoritative for UI + runtime), mirrored into each affected tool's `resources/<Tool>/resource.json` → `guardrail.policies[]`."*

skills/uipath-agents/references/lowcode/capabilities/guardrails/guardrails.md
```diff
@@ -8,11 +8,11 @@ Two types exist:
...
 
 ## Conversational Support
 
-**Status: Tool-scoped only, per-tool resource files are authoritative.** Conversational agents support guardrails with `selector.scopes: ["Tool"]` only — DO NOT use `"Agent"` or `"Llm"`.
+**Status: Custom (deterministic) `Tool`-scoped guardrails ONLY. No built-in validators.** `builtInValidator` guardrails (`pii_detection`, `prompt_injection`, `harmful_content`, `intellectual_property`, `user_prompt_attacks`) are autonomous-only. The only guardrails that run are `$guardrailType: "custom"` deterministic rules (word/number/boolean/always) with `selector.scopes: ["Tool"]`. Write each as the **same object (same `id`) in two places**: the tool's `resources/<Tool>/resource.json` → `guardrail.policies[]` **and** `agent.json` root `guardrails[]`. `"Agent"` and `"Llm"` scopes are not available. If asked for PII / harmful-content / injection detection, explain built-in validators are autonomous-only and offer a Custom Tool guardrail or an autonomous agent instead.
```
apetraru-uipath: "Write each as the same object (same `id`) in two places" is fine, but please relabel which is authoritative — together with the § top callout this reads as if the tool resource is primary. It isn't: the root is the source of truth (UI shows/edits it; Python reads root only, C# reconstructs the tool copy from it). A guardrail placed only in the tool resource is invisible/uneditable in Studio Web. Recommend: *"…in two places — the `agent.json` root `guardrails[]` is authoritative; the tool `resources/<Tool>/resource.json` → `guardrail.policies[]` is its mirror."*

skills/uipath-agents/references/lowcode/capabilities/guardrails/guardrails.md
```diff
@@ -8,11 +8,11 @@ Two types exist:
 - **`custom`** — deterministic rules you define (word matching, number comparison, boolean checks, universal triggers)
 - **`builtInValidator`** — UiPath Guardrails API validators (PII detection, harmful content, prompt injection, IP protection, user prompt attacks)
 
-> **Autonomous agents:** All guardrails are configured at the `agent.json` root `guardrails` array. **Conversational agents:** see § Conversational Support below — the runtime-effective location is each tool's `resources/<Tool>/resource.json` → `guardrail.policies[]`.
+> **Autonomous agents:** All guardrails are configured at the `agent.json` root `guardrails` array. **Conversational agents:** each guardrail goes in **both** the tool's `resources/<Tool>/resource.json` → `guardrail.policies[]` and the `agent.json` root array — see § Conversational Support below.
```
apetraru-uipath: Same relabel here. "each guardrail goes in **both** … `resource.json` → `guardrail.policies[]` and the `agent.json` root array" is accurate about the on-disk contract, but the **root is the authoritative one** (source for UI + both runtimes) and the tool copy is the derived mirror. Suggest: *"…root `guardrails[]` (authoritative) and mirrored into the tool's `resource.json` → `guardrail.policies[]`"* so it doesn't imply the tool resource is primary.

tests/tasks/uipath-agents/lowcode/conversational/guardrail_custom_tool/check_conversational_custom_guardrail.py
```diff
@@ -0,0 +1,136 @@
+#!/usr/bin/env python3
+"""Conversational custom Tool-guardrail check (conversational Critical Rule 1).
+
+Conversational agents support ONE guardrail type: a `$guardrailType: "custom"`
+deterministic rule scoped to a Tool. Built-in validators (`pii_detection`,
+`harmful_content`, etc.) are autonomous-only and never run on conversational
+agents. The runtime-effective location for a conversational guardrail is each
```
apetraru-uipath: This docstring states the inverted model. `resources/<Tool>/resource.json` → `guardrail.policies[]` is **not** "the runtime-effective location," and the root is **not** "only the Studio Web display mirror." Both runtimes treat the root as authoritative (Python reads it exclusively; C# reconstructs tool policies from it via a backwards-compat fallback), and Studio Web reads/writes the root. Please flip the description.

tests/tasks/uipath-agents/lowcode/conversational/guardrail_custom_tool/check_conversational_custom_guardrail.py
```diff
@@ -0,0 +1,136 @@
...
+        sys.exit("\n".join(lines))
+    print("OK: no built-in validator guardrails (agent.json root + per-tool resource.json)")
+
+    # 2. A valid custom word Tool-guardrail in a tool resource.json guardrail.policies[].
+    valid = [(src, g) for src, g in tool_policy_guardrails if is_valid_custom_word(g)]
+    if not valid:
+        root_valid = [g for _, g in root_guardrails if is_valid_custom_word(g)]
+        if root_valid and not any(is_valid_custom_word(g) for _, g in tool_policy_guardrails):
+            sys.exit(
+                "FAIL: the custom word guardrail is only in agent.json root "
```
apetraru-uipath: **Blocking.** This assertion enforces the inverted model and will train the skill on a wrong rule.

It **fails** an agent-root-only config, claiming *"this guardrail would not run"* — but an agent-root-only Tool guardrail **does** run: the Python / Unified Runtime reads `agent_definition.guardrails` (root) directly, and the C# runtime reconstructs it from `agent.Guardrails` via `ToolFactory`'s fallback. It's also visible/editable in Studio Web.

Conversely, the config this test **requires** (guardrail in `resource.json` policies, with no requirement that it also be in the root) is the broken one: it does **not** run on the Python runtime and is **invisible** in the Studio Web UI.

Please invert the check: require the guardrail in `agent.json` root `guardrails[]` (primary), and accept the tool-resource copy as the mirror — not the reverse. At minimum, stop failing agent-root-only configs and drop the "display mirror / would not run" message.

apetraru-uipath (changes requested):
## The guardrail source-of-truth direction is inverted

Thanks for tightening the conversational-guardrail rules — the "built-in validators are autonomous-only" part looks right. But the PR repeats a claim about **where guardrails live** that is backwards, and it's now baked into the new test. I checked this against the actual product source (helix backend + Studio Web frontend + the Python / Unified Runtime), not the docs.

**Ground truth:** `agent.json` root `guardrails[]` is the source of truth for **both** the UI and the runtime. The per-tool `resources/<Tool>/resource.json` → `guardrail.policies[]` is a one-way **derived mirror**, not the runtime-effective source.

- **Studio Web UI** reads/writes the root only and *generates* the tool copies — `frontend-sw/src/utils/guardrailSync.ts` (`syncGuardrailsToTools()` maps agent-root → `tool.policies`; its own comment: *"Guardrails are stored at agent level (`agent.guardrails[]`)"*). `GuardrailsListingSection.tsx` never reads `resource.json`.
- **C# runtime** — autonomous *and* conversational both call the same `ToolFactory.MapResources` (`ConversationalAgentState.cs` vs `BasicAgentState.cs`). `ToolFactory.cs`: `var guardrails = resource.Guardrail ?? ToolGuardrailDefinitionForResource(agent, resource.Name);` with the comment *"This is needed for backwards compatibility."* Reading `resource.Guardrail` is the **backwards-compat** branch; the fallback reconstructs tool policies from `agent.Guardrails` (root).
- **Python / Unified Runtime** reads the root **only** and never reads `resource.json` policies at all: `build_guardrails_with_actions(agent_definition.guardrails, tools)` (uipath-agents-python `graph.py`, `voice/session.py`); no resource/tool model even defines a guardrails field.

**Consequence:** an **agent-root-only** guardrail runs on both runtimes and is visible/editable in Studio Web; a **resource-only** guardrail does *not* run on the Python runtime and is invisible in Studio Web. So the safe single source is the root.

I'm not disputing "write it in both places" — the CLI doesn't auto-sync and the product persists both on disk, so that's fine. The fix is just to correct which one is **authoritative** vs. the **mirror**. Requesting changes mainly because the new test now enforces the inverted rule (see the blocking comment on the test).

_Minor:_ this model may trace back to the deleted test's `uip agent migrate` note — that command doesn't exist (it's `uip agent refresh`), and refresh/pack don't inline tool policies up into the root.

apetraru-uipath:
### Don't hard-code the built-in validator list — it will silently rot as new validators ship

This PR spells out the full set of built-in validator IDs (`pii_detection`, `prompt_injection`, `harmful_content`, `intellectual_property`, `user_prompt_attacks`) in several places. New built-in validators will be added over time, and every one of these lists then drifts out of date with no signal. It also contradicts the skill's own rule in `guardrails-recommend.md` step 5: *"Derive the grouping … from the catalog's own fields — **do not hardcode validator names.**"*

The point these callouts are making ("built-in validators are autonomous-only; conversational supports only Custom `Tool` guardrails") is a statement about the **class** `$guardrailType: "builtInValidator"`, not about specific IDs — so it doesn't need the enumeration at all. Refer to the class and point at the authoritative dynamic source (`uip agent guardrails list` / the single Validators Quick Reference table), instead of repeating the names everywhere.

Please de-enumerate at each of these:

- `skills/uipath-agents/references/lowcode/capabilities/guardrails/guardrails-recommend.md:122` — conversational callout
- `skills/uipath-agents/references/lowcode/capabilities/guardrails/guardrails.md:15` — § Conversational Support
- `skills/uipath-agents/references/lowcode/critical-rules/conversational-critical-rules.md:7` — Critical Rule 1
- `skills/uipath-agents/references/lowcode/critical-rules/conversational-critical-rules.md:13` — "What NOT to Do" #2 (the friendly-name variant "PII, harmful content, prompt injection, IP, user prompt attacks")
- `tests/tasks/uipath-agents/lowcode/conversational/guardrail_custom_tool/check_conversational_custom_guardrail.py:34-39` — the `BUILTIN_VALIDATORS` name set. The script already detects built-ins via `g.get("$guardrailType") == "builtInValidator"`, so the hardcoded 5-name set is redundant **and** will rot the day a new validator ships. Drop the set and rely on the discriminator.

Suggested phrasing for the prose callouts:

> Built-in validators (any `$guardrailType: "builtInValidator"` — the UiPath Guardrails API validators returned by `uip agent guardrails list`) are autonomous-only; conversational agents run only `$guardrailType: "custom"` deterministic `Tool` guardrails.

If a concrete enumeration is genuinely useful, keep exactly **one** canonical copy — the Validators Quick Reference table in `guardrails.md` — and have every other spot link to it rather than duplicating the list.

apetraru-uipath (approved):

PR #1753, opened 2026-06-29
test(uipath-maestro-flow): scope inline-agent graders to named outputs

jiyangzh:
## Problem

The `billing_resolution_writer` and `billing_dispute_analyst` graders matched their needle against the **whole** `flow debug` payload (`assert_outputs_contain` / `collect_outputs`). The manual trigger echoes its inputs back into the outputs, so the match is satisfiable by the **input echo** — the grader passes even when:

- the inline agent **refuses** to draft (literal `input.X` passthrough), or
- the End node **never maps** the agent result into the output global (output null).

Both were observed this round: a run drafted a correct email but left `emailBody` null (End-mapping miss) and still scored 1.000. For `dispute_analyst`, the old token list (`290`, `300`, `contracted`, …) is present verbatim in the input `disputeDescription`, so the match proved nothing.

## Fix

Grade the **mapped output globals by name**, not the whole payload.

- `flow_check.py`: add `assert_output_nonempty(payload, name)` and `assert_named_output_contains(payload, name, needles, *, require_all=)` — scope to one declared output global (present, non-empty, needle match). Add `get_last_debug_raw()` to persist the trace.
- `check_billing_resolution_writer.py`: assert `emailSubject` non-empty and `emailBody` cites the invoice + approved credit.
- `check_billing_dispute_analyst.py`: assert the `determination` output (non-empty, named) and persist the debug trace for post-run grounding inspection. Drops the input-echo-satisfiable token match; deliberately does not assert grounding (unprovable at flow-debug level — dispute facts are in the prompt).
- Refresh the `billing_resolution_writer` reference fixtures to the working `binding:`-based input wiring.

## Validation

Verified the new helpers against real `flow debug` payloads:
- A run whose `emailBody` was null (End-mapping miss) now **FAILS** (old grader passed it).
- A correctly-wired run **PASSES** (`emailBody` cites `MCS-2026-04872` + `$1,610`).

Re-ran the affected tasks end-to-end with the tightened graders — `billing_resolution_writer`, `billing_invoice_lookup`, `billing_dispute_analyst` all pass at 1.000 with the scoped checks.

## Scope

Test-grading correctness only. Independent of the skill-docs fix for inline-agent input/output wiring (separate PR).

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
@@ -334,6 +334,73 @@ def assert_outputs_contain(
         )
 
 
+def get_last_debug_raw() -> str | None:
+    """Return the raw stdout of the most recent ``run_debug`` call (the full
+    ``uip maestro flow debug`` JSON envelope), or ``None`` if none ran yet.
+    Useful for persisting the execution trace for post-run inspection."""
+    return _LAST_DEBUG_RAW
+
+
+def assert_output_nonempty(payload: dict, name: str) -> Any:
+    """Assert a named output global (e.g. an End-node-mapped ``out`` variable)
+    is present and non-empty, and return its value.
+
+    Looks the variable up by name in the runtime payload's ``variables.globals``
+    dict and ``variables.globalVariables`` array (both casings). "Non-empty"
+    means: present, not ``None``, and — once stringified — not whitespace-only
+    (so ``""``, ``"   "``, ``{}``, ``[]`` all fail)."""
+    variables = _get_ci(payload, "variables", "Variables") or {}
+    globals_dict = _get_ci(variables, "globals", "Globals") or {}
+    value = _get_ci(globals_dict, name)
+    if value is None:
+        for v in _get_ci(variables, "globalVariables", "GlobalVariables") or []:
+            if str(_get_ci(v, "id", "Id", "name", "Name") or "").lower() == name.lower():
+                value = _get_ci(v, "value", "Value")
+                break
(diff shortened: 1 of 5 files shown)


rockymadden (approved):

bai-uipath (approved):
Fix what you agree with, otherwise lgtm. The grader scoping is the right fix — matching named output globals instead of the whole `flow debug` dump kills the input-echo false pass cleanly. The notes below are about the reference-fixture rewrite that rode along, none of them blocking the grader change itself.

**Key naming diverges from the wiring contract.** The refreshed writer fixture uses short keys — `agentInputVariables[].id: "customerName"`, `inputSchema` key `customerName`, `{{input.customerName}}` — but the documented contract is the flattened `<trigger>__output__<var>` form (`start__output__customerName`) in all three places. The short keys are internally consistent so they likely resolve at CLI debug, but the golden reference now teaches a different shape than the docs, and the canvas regenerates the flattened `id` from the binding path on round-trip. Fix: author the fixture with the flattened keys so reference and docs are identical.

**The fixture is never executed by the grader.** Nothing in the harness runs `*.reference.flow` — the check debugs the candidate flow. So the writer task passing validates the grader and the candidate's wiring, not this committed fixture; its runtime correctness is unverified. Fix: align it to the flattened form, or debug-run it once to confirm short keys actually populate `emailBody`.

**Analyst reference left on the deprecated wiring.** `billing_dispute_analyst` still has `agentInputVariables: []` + `{{ $vars.X }}` + empty `inputSchema` — the form that faults at debug under the new contract. It's not consumed by grading, so this is consistency only, but refresh it to match the writer.

PR #951, opened 2026-05-22
test(uipath-troubleshoot): 3 scenarios for Excel Read Range NRE/TIE playbook

Stefan-Virgil:
## Summary

Three coder-eval e2e scenarios exercising distinct branches of the `activity-packages/excel-activities/playbooks/read-range-null-reference.md` playbook (PR #950). **Stacked on #950 → #945 → #894 → #889.** CI here will fail until the parent PRs land; this PR merges cleanly after.

This completes the original Excel Read Range scope: 4 playbooks + 12 test scenarios across PRs #889, #893, #894, #944, #945, #949, #950, and this PR.

| Scenario | Branch | Surface signal | Resolution path |
|---|---|---|---|
| `excel-rr-nre-label` | Sensitivity label (branch 1) | Job-log Trace surfaces `sensitivity label: 'Confidential\Limited Access'; access principal: UIPATH\AUTOMATION1; decision: AccessDenied` + user "Confidential banner" clue | Grant Robot user label permission via Microsoft Purview, OR remove the label |
| `excel-rr-nre-namedrange` | Broken named range (branch 3) | Workflow source uses `Range: "MyDataRange"`; job logs Trace shows `resolving named range 'MyDataRange' (RefersTo: '#REF!')` | Excel Name Manager (`Ctrl+F3`) → fix or replace the broken named range |
| `excel-rr-nre-opaque` | Branches 2/4/5 collapsed | TIE wrapping NRE with no pointer; CLI rules out branches 1/3; user clue "workbook generated nightly by Python script" leans toward branch 2 | Workbook inspection checklist (open in Excel, Repaired dialog, Compatibility Checker, Name Manager, file size) — agent does NOT confidently pick a branch without inspection |

## Local results

Ran `coder-eval --repeats 3 -j 3` per scenario on Windows local (claude-sonnet-4-6 coder). **9/9 replicates passed, all at 1.000.**

| Scenario | Replicate scores | Mean | Status |
|---|---|---:|---|
| `excel-rr-nre-label` | 1.000 / 1.000 / 1.000 | 1.000 | 3/3 SUCCESS |
| `excel-rr-nre-namedrange` | 1.000 / 1.000 / 1.000 | 1.000 | 3/3 SUCCESS |
| `excel-rr-nre-opaque` | 1.000 / 1.000 / 1.000 | 1.000 | 3/3 SUCCESS |
| **Overall** | | **1.000** | **9/9** |

[...]

PR #1754, opened 2026-06-29
fix(test): align audit export tests with actual LTS-schema columns

t-hsia:
## Summary
- The `uip admin audit tenant export --file-format csv` CLI produces CSV columns using LTS-schema names (`Identifier`, `DateCreatedUtc`, `Action`, `Source`, `Category`, `ClientInformation`) — not the events-endpoint PascalCase names (`Id`, `CreatedOn`, `EventType`, `EventSource`, `EventTarget`, `ClientInfo`) that tests and docs assumed.
- Updated `REQUIRED` column lists in all 3 audit export verify tests (`audit_export_csv_verify_e2e`, `audit_export_verify_e2e`, `audit_org_export_verify_e2e`).
- Updated `audit-commands.md` reference to document actual LTS column names.

## Evidence
- **Main (unfixed) deterministic failure:** https://github.com/UiPath/skills/actions/runs/28403769097
- **Fix branch verification run:** https://github.com/UiPath/skills/actions/runs/28404795148

## Test plan
- [x] Confirmed deterministic failure on main — `AssertionError: header missing required columns: ['Id', 'CreatedOn', 'EventType', 'EventSource', 'EventTarget']`
- [ ] Verify fix branch run passes the CSV column validation criterion


yadvender-uipath:
### Downstream finding: the CLI's curated CSV column order is now stale (root cause worth fixing upstream)

Reviewed this against merged AuditService PR #951. The rename mapping in the tests and `audit-commands.md` is correct and 1:1 with the new `[JsonPropertyName]` values (`Id→Identifier`, `CreatedOn→DateCreatedUtc`, `EventType→Action`, `EventSource→Source`, `EventTarget→Category`, `ActorName→User`, `ClientInfo→ClientInformation`), and the documented CSV column order matches exactly what the CLI emits today. 👍

One thing to flag for context, though: the documented order is an **artifact of a now-stale list in the CLI**, not the long-term-store/DTO order. The exporter's `PREFERRED_CSV_COLUMNS` (in `packages/admin/audit-tool/src/utils/chunked-export.ts`) still lists the *old* PascalCase names. After #951 none of the renamed keys match that list, so `deriveCsvColumns` drops them into the "unknown extras, first-seen order" bucket — which is why the renamed columns (including the identity columns `Identifier`/`DateCreatedUtc`) now sort *after* the unchanged ones (`OrganizationId, TenantId, …, Status`) instead of leading the header.

So two notes for this PR:
- The doc line **"Columns follow the long-term-store field order"** is no longer accurate — the listed order is a side effect of the stale preferred-list partial match. Suggest rewording (e.g. "unchanged fields first, then the renamed fields in server order").
- The trailing CSV example still says **`nested objects (e.g. \`ClientInfo\`)`** — should be `ClientInformation` to match the renamed column.

I'm opening a CLI PR to update `PREFERRED_CSV_COLUMNS` to the new names, which restores the intended DTO-ordered header. Heads-up that once it merges, the **column order** documented here will change (identity columns move back to the front); the membership-based test assertions in this PR are unaffected since they don't check order. Will link the CLI PR here.

yadvender-uipath:
CLI fix for the stale preferred-column list: UiPath/cli#2771 — renames `PREFERRED_CSV_COLUMNS` to the new LTS-schema keys so the curated DTO header order is restored.

yadvender-uipath (approved):

PR #1755, opened 2026-06-30
feat(uipath-maestro-case): allow In-arg to bind a specific trigger via sourceTriggers

song-zhao-25:
## What

Lets an SDD §1.5 `Category=In` row bind the In-argument to a **specific trigger** via the existing `sourceTriggers` column, instead of always implicitly binding the primary trigger.

- `sourceTriggers` on an `In` row: **blank → primary trigger (T02)** — unchanged, backward-compatible default; a **single `T<N>` → that trigger**. Never a CSV (one trigger only).
- `sourceFields` stays **empty** on `In` rows — an In-arg *selects* a trigger, it does not *extract* a payload field (that remains the `Variable` operation).

## Why

Previously In-args had no way to express which trigger they attach to: `sourceTriggers` was forbidden on `In` rows, and the internal `triggerRef` field was dead (never populated from the SDD, no resolution rule — the emitted `elementId` was a silent placeholder that only worked by happenstance in single-trigger cases).

## How

| Layer | Change |
|---|---|
| **SDD authoring** (main change) | `sdd-generation-rules.md`, `sdd-template.md`, `sdd-template-examples.md` (new Use Case 3b) — allow a single `sourceTriggers` T-number on `In` rows; keep `sourceFields` empty; forbid CSV. |
| **Resolution** | In-arg `elementId` (formal slot + companion) and the bridge's host node resolve to `id-map[<sourceTriggers>].id`; blank → `id-map["T02"].id`. |
| **Dispatch** | `Category=In` is always handled by the variables plugin's Loop B (never Loop A payload-extraction); Loop A In cell is an explicit no-op (prevents a double-emit on a name collision). |
| **Validators** | New Phase-2 checks: `In` + CSV → ERROR; `In` + non-empty `sourceFields` → ERROR. Orphan-T check now covers `In`. Legacy `triggerRef` removed. |
| **Cross-refs** | `SKILL.md`, `implementation.md`, manual/timer/event trigger id-map notes reconciled. |

Out-arg rules unchanged (`Out` + `sourceTriggers` still rejected — direction mismatch).

## Test

`tests/tasks/uipath-maestro-case/multi_stages_tasks/in_arg_trigger_bind/` — a 2-trigger case (manual primary T02 + timer T03) with **three** In-args:

| In-arg | `sourceTriggers` | binds |
|---|---|---|
| `caseId` | *(blank)* | primary (manual) |
| `caseId2` | `T02` | primary (manual) — explicit-primary resolves identically to blank |
| `approverId` | `T03` | timer |

Structural-only check (no debug) asserting each In-arg's formal-slot + companion `elementId` + trigger-output bridge land on the correct node, with a leak-guard that no bridge spills onto the other trigger. `/lint-task`: OK.

## Backward compatibility

Blank `sourceTriggers` → primary trigger = the old implicit behavior. All existing single-trigger In-arg fixtures emit byte-identically; no test required adjustment.


song-zhao-25:
## ✅ coder-eval passing-run claim

Ran the new task locally against this branch's **modified skill** (`SKILLS_REPO_PATH=<worktree>`, default experiment, `claude-sonnet-4-6` agent):

```
coder-eval run tasks/uipath-maestro-case/multi_stages_tasks/in_arg_trigger_bind/in_arg_trigger_bind.yaml -e experiments/default.yaml
```

**Result: `status=SUCCESS`, 2/2 success criteria, weighted score `1.000`** (duration ~616s, 1 iteration).

| Criterion | Score |
|---|---|
| `uip maestro case validate` passes on the generated caseplan | 1.00 |
| Structural check — each In-arg's formal slot + companion `elementId` + trigger-output bridge land on the correct node (`approverId`→T03 timer; `caseId` blank & `caseId2` explicit-T02 → primary manual; no cross-trigger leak) | 1.00 |

Check-script output:
> OK: 2 triggers (manual primary + timer T03). In-arg 'approverId' (sourceTriggers=T03) bound to the timer node; 'caseId' (blank) and 'caseId2' (explicit T02) both bound to the primary manual node — each with formal slot + companion elementId + trigger-output bridge (source=vars.<slot id>) on its node, and no bridge leaking to the other trigger. Explicit-primary (T02) resolves identically to blank.

The agent, following the updated SDD/impl docs, authored the `sourceTriggers`-on-`In` wiring correctly end-to-end — confirming the change works as documented.

charlesliu9:
1. Eval can false-pass invalid companions
     tests/tasks/uipath-maestro-case/multi_stages_tasks/in_arg_trigger_bind/check_in_arg_trigger_bind.py:87 finds an inputOutputs companion when either name == arg or id == arg, but runtime resolution depends
     on id. A broken caseplan with {name: "caseId", id: "wrong"} would pass the check while =vars.caseId remains unresolved. Require id == arg and separately assert name == arg.

  2. New task uses non-taxonomy feature tags
     tests/tasks/uipath-maestro-case/multi_stages_tasks/in_arg_trigger_bind/in_arg_trigger_bind.yaml:13 adds feature:multi-trigger, feature:timer, feature:arguments, and feature:variables, but the repo’s tag
     taxonomy is closed and does not list these values. Keep feature:trigger and either use approved tags or omit the unsupported dimensions.

song-zhao-25:
Thanks @charlesliu9 — both addressed in 9228a4f6e.

**1. Companion false-pass.** You're right: the runtime resolves `=vars.<arg>` by the companion's `id` (`VariablesService.findVariableByVariableId`), so matching on `name OR id` could green-light a `{name:<arg>, id:<wrong>}` companion that never resolves. `_companion` now matches **strictly by `id`**, and `name == arg` is asserted separately — a broken companion now fails instead of passing.

**2. Non-taxonomy feature tags.** Trimmed to `feature:trigger` (the only value on the closed `feature:` list in `tests/README.md`); dropped `multi-trigger` / `timer` / `arguments` / `variables`.
> Note: the sibling tasks `multi_trigger.yaml` and `linear_three_stages.yaml` carry these same non-taxonomy tags — so there's a pre-existing inconsistency in the suite (and `lint-task` doesn't catch it). Happy to do a separate sweep to clean those up if you'd like.

Re-graded the preserved build with the tightened check — **2/2 pass**.

song-zhao-25:
## ✅ Re-verified after rebase + review fixes

Fresh full `coder-eval run` of a **new agent build** on the current branch head (`aea564eb9` — rebased onto latest `main`, with the tightened companion check and trimmed tags):

**Result: `status=SUCCESS`, 2/2 success criteria, weighted score `1.000`** (~774s, 1 iteration).

| Criterion | Score |
|---|---|
| `uip maestro case validate` passes on the generated caseplan | 1.00 |
| Structural check (now `id`-strict on the companion) — `approverId`→T03 timer; `caseId` (blank) & `caseId2` (explicit T02) → primary manual; no cross-trigger leak | 1.00 |

Check output:
> OK: 2 triggers (manual primary + timer T03). In-arg 'approverId' (sourceTriggers=T03) bound to the timer node; 'caseId' (blank) and 'caseId2' (explicit T02) both bound to the primary manual node — each with formal slot + companion elementId + trigger-output bridge (source=vars.<slot id>) on its node, and no bridge leaking to the other trigger. Explicit-primary (T02) resolves identically to blank.

This supersedes the earlier passing-run claim (which ran on the pre-rebase, pre-review-fix build). The feature still builds and binds correctly end-to-end on top of current `main`.

charlesliu9 (approved):

PR #1766, opened 2026-06-30
fix(uipath-platform): poll job logs in e2e check to beat ingestion lag

alexenica:
## What

`tests/tasks/uipath-platform/orchestrator/check_job_run_logs.py` read `uip or jobs logs` **once**, immediately after the job reached a terminal state.

## Why

Job log ingestion lags slightly behind the job's terminal transition. A single early read races that lag and returns empty, so the check fails with `FAIL: jobs logs returned empty` even though the job ran fine and its logs appear moments later. This made `skill-platform-job-run-logs-e2e` flaky (FAILURE 0.0 in the 2026-06-26 nightly).

Confirmed on `alpha`/`codereval`/`DefaultTenant`: every recent terminal job on the seeded stub **does** have logs (`execution started` / workflow log / `execution ended`) — the logs are real, the check was just reading too early.

## Fix

- Poll `jobs logs` for up to 60s (job is already confirmed terminal — we only wait for logs to flush) instead of reading once.
- Bump the check command `timeout` 30 → 90 so the poll fits inside it.

## Testing

On `alpha`/`codereval`/`DefaultTenant`:

| Test | Result |
|------|--------|
| Real terminal job with logs | `OK ... logs=3` in ~2s (no needless waiting) |
| Bogus job key | fails fast at the terminal-state check (~0.8s), never reaches the poll |

### DIFF
--- a/tests/tasks/uipath-platform/orchestrator/check_job_run_logs.py
+++ b/tests/tasks/uipath-platform/orchestrator/check_job_run_logs.py
@@ -6,6 +6,7 @@ Reads only the job key from `job_key.txt`; everything else queried live."""
 import json
 import subprocess
 import sys
+import time
 from pathlib import Path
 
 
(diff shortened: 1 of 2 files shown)


alexenica:
Folded into #1762 — all three uipath-platform e2e fixes (seed pagination, job-logs poll, role disambiguation) now ship in one PR.

PR #1782, opened 2026-07-01
feat(uipath-troubleshoot): add If-condition NRE + KeyNotFound coverage + tests

Stefan-Virgil:
> **Stacked on #1780** (`feat/troubleshoot-assign-runtime-exceptions`). Base will be retargeted to `main` and the branch rebased once #1780 merges — it depends on `key-not-found-exception.md`, which #1780 introduces.

## What

Extends `uipath-troubleshoot` runtime-exception coverage to faults thrown while resolving a **`System.Activities` If Condition** expression — for the two exceptions requested: `System.NullReferenceException` and `System.Collections.Generic.KeyNotFoundException`.

### Playbooks (extended, not duplicated — DRY)
Both exceptions already have per-exception playbooks. Rather than add redundant files, this adds an **`If` / `While` Condition** fault origin to each:
- `references/runtime-exceptions/playbooks/null-reference-exception.md`
- `references/runtime-exceptions/playbooks/key-not-found-exception.md`

Each notes the key nuance: the condition resolves **before either branch runs**, so the fault is in the `If` itself, not in a Then/Else activity.

### Tests (`tests/tasks/uipath-troubleshoot/runtime-exceptions/<scenario>/`)
Two faithful-replay e2e diagnose scenarios where an `If` Condition throws:
| Scenario | If Condition modeled |
|---|---|
| `if-null-reference-exception` | `If status.ToString() == "yes"` with `status` null |
| `if-key-not-found-exception` | `If config["FeatureEnabled"] == "true"` with the key absent |

Each ships README, RESOLUTION, task.yaml, mock OR fixtures (folders + Faulted `jobs get` with the `at If "…"` stack + error logs), and a `process/` snapshot. Graded on `skill_triggered` + `llm_judge` vs `RESOLUTION.md`.

## Validation — coder-eval

`experiments/default.yaml`, `--repeats 3 -j 3`. **6 / 6 SUCCESS @ weighted_score 1.0.**

| Scenario | 00 | 01 | 02 |
|---|---|---|---|
| if-null-reference-exception | ✅ 1.0 | ✅ 1.0 | ✅ 1.0 |
| if-key-not-found-exception | ✅ 1.0 | ✅ 1.0 | ✅ 1.0 |

Manifest verifier: 0 invalid / 0 parse errors. `/lint-task`: OK.


Stefan-Virgil:
Addressed the Theme 1 (Low) finding in 36da442eb: both `task.yaml` `description:` fields now read *"thrown while resolving an If Condition expression"* instead of *"…from an Assign expression"* (a copy-paste artifact from the Assign-based siblings). README/RESOLUTION/mock fixtures already referenced `If` correctly.

Description-only metadata change — no impact on mock dispatch, prompt, or success criteria, so the 6/6 @ 1.0 validation stands (no re-run needed).

MarinRzv (approved):

PR #1802, opened 2026-07-01
feat(maestro-case): build missing API workflows inline during planning

song-zhao-25:
## Summary

Third creatable kind at the Rule 17 empty-lookup gate (after agent #1622, agentic process #1723): a missing **API workflow** is built inline as an in-solution sibling by a sub-agent invoking `uipath-api-workflow`, then registered (`uip solution project add`), rediscovered offline (`registry search --type api --local`), I/O-verified from the sibling's `entry-points.json`, and bound (`resource:"process"` / `resourceSubType:"Api"` / `resourceKey="solution_folder.<name>"` / runtime `folderPath:""`).

Docs-only, single commit. All shared Create-on-Missing machinery ships with #1622 (merged) and the centralized adopt contract with #1833 (merged) — this PR adds the per-type surface + enum widenings, the same delta shape as the agentic leg.

## What changed

- **api-workflow plugin create surface** — `planning.md`: match-priority ladder, pre-gate in-solution sibling check, `## Creating an API workflow inline` (builder brief + runtime blockquote + failure contract); `impl-json.md`: Built-inline notes; ~35 enum/label/link widenings across SKILL.md and 9 references.
- **Builder brief overrides `uipath-api-workflow` rule 19a** ("edit Workflow.json only" / "I/O may be null") — without it a rule-abiding builder skips the entry-points back-fill and the warn-only verify binds an empty contract. Back-fill uses the **FLAT deploy shape** (`input.properties`, agent-identical, runtime-proven); the read path tolerates the `Workflow.json` wrapper shape with a warning.
- **Connectors delegated, not restricted** — the case skill is a pass-through: it hands the Purpose + pinned I/O to `uipath-api-workflow`, which owns the build (HTTP *and* Integration Service connectors included; `uip login` is ambient). The builder is non-interactive, so where that skill's rule 16 would stop and ask (a required Integration Service activity has no connection that pings), it returns `built:false` → the task becomes a placeholder — never a fake compute stub. When it *does* author an IntSvc connector it runs `uip api-workflow bindings sync` so the caller's existing `resources refresh` catalogues the connection into the solution. Never `uip api-workflow run` (vendor side effects); offline `validate` is the completion bar.
- **No build-kind choice** (JSON-DSL only) — §1b skip clause.
- **Gate label** is the generic `Create the missing resource(s) inline`, aligned with the RPA leg (#1803).

## Runtime evidence

- **Deploy e2e (alpha, 2026-07-01):** `solution pack → publish → deploy run` provisions the Api sibling as a runnable process (`<Pkg>.Api.<Name>`) in the case's folder; the case invoked it to **`Successful`** through the `folderPath:""` binding. **`case debug` does NOT provision Api siblings** — task faults incident `170007` by design; documented in the plugin + `phased-execution.md` Debug notes + `implementation.md` Step 13a (consent-gated deploy alternative, options enumerated).
- **L4 suite: 6/6 PASS (2026-07-02)** at `case_coding_agent_tests/create-on-missing-api-workflow-test/` (results.md + per-test run-log.md): T1 full loop incl. the rule-19a-override back-fill proof + SW upload; T2 idempotent re-run; T2b on-disk-unregistered adopt; T3 mixed agent+api select/skip routing (incl. flat-shape regression proof); T4 induced I/O mismatch → warn-not-block, nothing fabricated; T4b null-entry-points → Workflow.json-root fallback + warning.
- **T2b re-run post-#1833 (2026-07-06): PASS** — a live agent navigated the slim pointer → shared §3b → §4 chain, hit both collision errors, kind-checked via the `project.uiproj` fallback, applied the stale-decl remediation ("Added successfully"), adopted with hash-proven no-rebuild.
- **Connector coverage caveat:** the deploy e2e and the L4 suite exercise **compute** workflows (no vendor connection). Http-kind and pure-compute builds are fully covered. An **Integration Service** connector inline sibling — build + `bindings sync` + `resources refresh` connection-catalogue + connection provisioning at case deploy — is wired per `uipath-api-workflow` rule 16 but not yet e2e-proven end-to-end in a case solution; follow-up. It degrades safely today: no pinged connection → `built:false` → placeholder.

## Relationship to #1723 / #1803

Independent (no stacking); agentic processes and RPA stay non-creatable here. This branch uses the **generic `Create the missing resource(s) inline` gate label** (aligned with #1803), so the label no longer diverges between the legs. Remaining overlap is textual only: ~15 one-line enum widenings (creatable-set + `--type` unions) touch the same lines those PRs widen — whichever merges second resolves the mechanical conflicts.

In-repo coder-eval task deferred, matching both prior legs (the gate is an `AskUserQuestion` and the build spawns an `Agent`-tool sub-agent — not drivable by the default coder-eval experiment).


charlesliu9:
• Findings

  - High: skills/uipath-maestro-case/references/plugins/tasks/api-workflow/planning.md:86 makes inline-created API workflows “compute-only” and forbids both HTTP and Integration Service activities, but skills/uipath-maestro-case/references/phased-execution.md:202 and skills/uipath-maestro-case/references/placeholder-tasks.md:190 report these as resolved / needing no user action. For API workflow tasks whose purpose is external integration, this can silently generate a validating but non-functional sibling and mark it wired. Either degrade those cases to placeholder/build failure, or report them explicitly as compute-only scaffolds requiring user completion.

song-zhao-25:
Addressed in `551021c99` — valid catch, thanks.

**Root cause:** the case skill was imposing a **compute-only** restriction that `uipath-api-workflow` never asks for. Verified against that skill: connector activities (HTTP + Integration Service) are first-class (its rule 16), and its completion bar is offline `validate` (rules 20–21) — nothing there requires compute-only. So the restriction was case-skill-invented over-reach, and it was the sole reason an integration-purpose workflow could be reported "resolved" while being non-functional.

**Fix (removed the restriction, didn't add one):** the case skill is now a pass-through — it hands the Purpose + pinned I/O to `uipath-api-workflow` and lets it build whatever the purpose needs, connectors included (`uip login` is ambient). Because the build sub-agent is non-interactive, where that skill's rule 16 would stop and ask (a required Integration Service activity has no connection that pings), it returns `built:false` → the case skill's **existing** failure path turns the task into a placeholder + review item, never a validating-but-non-functional stub. So the contradiction you flagged (compute-only vs. "resolved / no user action") is gone: an inline API workflow is now either genuinely functional or an honest placeholder — same contract as the already-merged agent leg, so the shared completion-report lines are correct unchanged.

**One honest residual** (noted in the PR description): the deploy e2e + L4 suite cover the compute and Http paths; an Integration Service connector sibling's connection provisioning at *case deploy* (`bindings sync` + catalogue) is wired per rule 16 but not yet e2e-proven — tracked as follow-up. It degrades safely today (no pinged connection → `built:false` → placeholder).

charlesliu9 (approved):

PR #1803, opened 2026-07-01
feat(maestro-case): add RPA process leg to create-on-missing

song-zhao-25:
## What

Widens the Rule 17 gate's creatable set from {agent} to {agent, rpa}: a case task typed `rpa` whose registry lookup is empty can now be built inline as an in-solution sibling (delegated to `uipath-rpa`), registered, rediscovered via `--local`, and bound — instead of falling to a placeholder.

- ~25 gate/enumeration sites widened across SKILL.md, registry-discovery.md, planning.md, placeholder-tasks.md, phased-execution.md, phase-0-interview.md, implementation.md, case-commands.md. Gate label is now `Create the missing resource(s) inline`.
- New `rpa/planning.md` sections: pre-gate in-solution sibling check (`--type process --local` — the registry has no `rpa` type) and **Creating an RPA process inline** (pinned I/O + case↔.NET type map, no kind prompt — v1 always XAML/VB, deterministic `uip rpa init` scaffold brief, no-`resourceSubType` binding, `project.json` `entryPoints` read-back, failure/adopt contract). Matching Built-inline note in `rpa/impl-json.md`.
- Shared bindings layer (`bindings-v2-sync.md`, `variables/bindings/impl-json.md`) generalized: decoupled `folderPath:""` vs `resourceKey="solution_folder.<name>"` now covers agent / RPA; rpa omits `resourceSubType`/`metadata.subType`.
- Incident `170007` bridged into the debug path (Phase-5 Debug notes, Step-13a short-circuit, troubleshooting-guide): an inline RPA sibling faulting under `case debug` is by design — debug does not provision non-agent siblings; do not "fix" the runtime-correct `folderPath:""`.

## Runtime verification (alpha, 2026-07-01)

Full path proven live twice with a scratch case + `RpaClassify` sibling: `solution pack → publish → deploy run` provisions the sibling as a runnable process; the case's stage-entry rule starts the job (serverless, 4s, `Successful`); an assigned output round-trips into case vars (`TaskCompletedOutputs`). Engine composes PascalCase `JobArguments` from camelCase XAML args — matching is case-insensitive both directions. `case debug` on the same case reproduces `170007` ("process could not be found"), confirming the debug-path-only limitation the docs now describe.

## Review

Second commit addresses an 11-finding skill review (170007 debug bridge, bindings-layer generalization, match-priority ladder in rpa Registry Resolution, cross-kind name-collision guard, placeholder-upgrade contradiction, bounded retry, type map in the builder brief).

## Merge coordination

Independent of #1723 (agentic stays non-creatable here). Sibling branch `feat/case-create-missing-api-workflow` touches the same enumeration lines; shared-doc hunks (agent/planning.md, gate §1 rename rule) are byte-identical across both branches so the second-to-merge unions cleanly.


song-zhao-25:
Superseded by #2058 — the RPA create-on-missing leg was re-implemented against #1966's kind-agnostic `create-inline-common.md` structure (thinner delta, same behavior). Closing in favor of #2058.

PR #1834, opened 2026-07-02
test: drop Claude-biased --output-json and review-catalog grader criteria

bai-uipath:
## What

Rebalances the skills eval suite to grade on what the work produced rather than which exact command shape produced it. Two families of `command_executed` criteria scored correct work as a failure for non-Claude agents (Sonnet vs Codex vs Antigravity, late-June snapshot). This PR removes the form-only ones, rewrites the specific-command ones to drop the biasing flag while keeping the real assertion, removes a redundant review-catalog Read check, and updates the task-authoring guidance that generated these checks so they don't regrow. 309 task files across 18 skills, plus 2 authoring-guidance docs.

## `--output json` checks

The flag is a formatting preference — the skills teach it only when CLI output is parsed programmatically, and it is never itself a capability. As a **gating** check it penalizes an agent that reaches the same independently-verified outcome without typing the flag. Every one of these sits next to an outcome/artifact check that already confirms the work. Two dispositions:

- **Removed entirely — 195.** These asserted the flag habit, not any specific command:
  - 178 pure wildcard-verb checks (`(uip|$UIP)\s+.*--output json`) that matched *any* command carrying the flag.
  - 11 flag-habit repetition checks — `min_count > 1` on `<category> .*--output json` with descriptions like "used --output json on every eval command" / "on at least 3 aops-policy commands". Stripping the flag would leave a meaningless "ran N commands in this category" repetition check, so these are dropped.
  - 6 redundant duplicates — a sibling check in the same task already asserts the identical command without the flag.
- **Rewritten — flag stripped, command kept — 188.** Specific-command checks keep the command and args the task actually cares about; only the trailing formatting flag is removed. Examples: `uip admin tenants list ... --output json` → `uip admin tenants list`; `uip tm executions list --test-set-id ... --output json` → `... --test-set-id`. Two-ordering alternations collapse to the meaningful branch, and lookahead compounds retain every non-flag constraint (e.g. `--project-key`, `--object-type`, `--data-type`).

No `--output json` remains in any `command_executed` check, including the handful that were previously non-gating advisory (`pass_threshold: 0`) — those are normalized the same way for consistency. Skill documentation is unchanged: it still teaches `--output json` where output is parsed. Only the grading stops keying on the literal flag.

## uipath-review "consulted the catalog" checks — 16 removed

**The removed check** observes a `Read` tool call, not the review:

```yaml
- type: command_executed
  tool_name: Read
  command_pattern: 'agents-coded-rules\.md'   # did the agent Read this file?
  pass_threshold: 1.0                          # gating
```

**The outcome check sitting next to it (kept)** already verifies the review is correct:

```yaml
- type: run_command
  description: "Review report cites CODED_PII_IN_TRACES for the email_body param"
  command: "python3 check_coded_review_pii_in_traces.py"
  pass_threshold: 1.0
```

So the Read check grades *how the agent obtained the rules*, not *whether the finding is right* — and that splits along agent lines:

- **Claude** — its review workflow opens the catalog with `Read` → grader sees it → pass.
- **Antigravity** — rules arrive via native skill injection, no `Read` call → fail.
- **Codex** — reaches the file through its own access the matcher doesn't record → fail.

Late-June runs, review tasks where the finding was correct (outcome check passed) but the Read check failed:

| Agent | correct review, failed only the Read check |
|---|---|
| Codex | 7 |
| Antigravity | 2 |

Removing the Read check deletes a Claude-shaped observation, not any measure of review quality — the outcome check remains the gate.

## Measured impact on the non-Claude runs

Re-grading the late-June Antigravity and Codex runs against the post-change criteria, using each run's actual command traces (removed checks stop gating; rewritten checks are re-run with coder_eval's own match rules against the recorded commands). The change is monotonic — it only loosens gating, so no task can regress from pass to fail.

| Agent | Tasks graded | Pass — before | Pass — after | Recovered (fail→pass) |
|---|---|---|---|---|
| Antigravity | 762 | 525 (68.9%) | 539 (70.7%) | **+14** — 12 `--output json`, 2 review-catalog |
| Codex | 554 | 397 (71.7%) | 404 (72.9%) | **+7** — 7 review-catalog |

Every recovered task passed all of its outcome/artifact checks and was failing only on the removed/rewritten form checks. This is a deliberately narrow, safe slice of the total grader bias — it fixes only the `--output json` and review-catalog patterns. It does **not** touch the two larger levers found in the same analysis: the `skill_triggered` activation-detection gap (≈26 Antigravity tasks) and the prescribed-`uip`-verb / repetition form checks — both of which need harness-level changes and are out of scope here. That is why the deltas are a couple of points rather than the full gap. The rewritten `--output json` checks flipped no tasks in these two runs (where an agent skipped the flag it also skipped the command), so their value is structural: they stop docking future correct-without-flag runs rather than recovering past ones.

## Prevent regrowth — task-authoring guidance

These checks propagated because the authoring rule required them. `.claude/rules/test-writing.md` told authors (and `/generate-task`) to add a gating check matching `(uip|\$UIP)\s+.*--output\s+json` to every CLI-parsing task, and `tests/README.md` shipped it as the canonical `1.0`-weight example. Left unchanged, the next generated task regrows exactly what this PR removed. The guidance now grades the **outcome**, keeps `--output json` as prompt steering, and demotes any convention check to advisory (`pass_threshold: 0`):

```yaml
  - type: command_executed
    description: "Agent used --output json (advisory — convention only, non-gating)"
    command_pattern: '(uip|\$UIP)\s+.*--output\s+json'
    pass_threshold: 0   # advisory: the flag is outcome-invisible — never gate on it
```

## Also — schema fix for two unrelated tasks

`uipath-planner/asdd_crosswalk` and `uipath-planner/attended_reauth` used `expected: "yes"` on their `skill_triggered` criterion, which the current criteria model rejects (`expected_skill` is required; extra keys are forbidden) — both failed to load. Corrected to `expected_skill: "uipath-planner"`, matching the convention used across the rest of the suite. Included here so the whole suite loads clean.

## Validation

All touched task files load through coder_eval's task loader, every `command_pattern` compiles, no `--output json` check remains in any `command_executed`, and each task retains its outcome/artifact checks. Full-suite load is clean (869/869). A superset check over the three runs' real command traces confirms the rewrites only loosen matching: all 549 recorded commands that matched an old pattern (every one carrying `--output json`) still match the rewritten pattern — 0 regressions, so stripping the flag never docks a `--output json` run.

### DIFF
--- a/.claude/rules/test-writing.md
+++ b/.claude/rules/test-writing.md
@@ -53,7 +53,7 @@ Use side effects (`command_executed`, `file_exists`, `file_contains`, `json_chec
 
 Weight: `1.0` supporting · `1.5` core command/artifact · `2.0` artifact content · `3.0` primary validation · `5.0–6.0` e2e execution. `pass_threshold: 1.0` unless the criterion has multiple sub-assertions.
 
-When criteria parse CLI output, steer the prompt toward `--output json` and add a low-weight check matching `(uip|\$UIP)\s+.*--output\s+json`.
+When criteria parse CLI output, steer the prompt toward `--output json`, but grade the **outcome** (the parsed value, via `run_command` / `file_check`), NOT the literal flag. Never add a gating `command_executed` check on `(uip|\$UIP)\s+.*--output\s+json`: the flag is outcome-invisible (often the CLI default), so gating on it docks agents that reach the same result without typing it. To record convention adherence, use an advisory check (`pass_threshold: 0`).
 
 ## Prompts
 
--- a/tests/README.md
+++ b/tests/README.md
@@ -339,12 +339,12 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
-    description: "Agent used --output json on uip commands"
+    description: "Agent used --output json (advisory — convention only, non-gating)"
     tool_name: "Bash"
     command_pattern: '(uip|\$UIP)\s+.*--output\s+json'
     min_count: 1
     weight: 1.0
-    pass_threshold: 1.0
+    pass_threshold: 0   # advisory: the flag is outcome-invisible — never gate on it
 
   - type: command_executed
     description: "Agent linked flow project to solution"
@@ -478,7 +478,7 @@ Score is binary: 1.0 when matches ≤ `max_count` (default `0`), else 0.0. Empty
 
 | Weight | When to use | Example from existing tests |
 |--------|-------------|---------------------------|
-| `1.0` | Supporting checks | `--output json` flag used, presence of an auxiliary file |
+| `1.0` | Supporting checks | presence of an auxiliary file, an advisory (`pass_threshold: 0`) convention check such as `--output json` used |
 | `1.5` | Core behavior | `uip solution new` executed, `.flow` file created |
 | `2.0` | Important artifact content | `.flow` file contains the expected node type or handle wiring |
 | `3.0` | Primary artifact validity | `uip maestro flow validate` passes on the generated flow file |
--- a/tests/tasks/uipath-admin/apms_add_ip_range_smoke.yaml
+++ b/tests/tasks/uipath-admin/apms_add_ip_range_smoke.yaml
@@ -41,11 +41,3 @@ success_criteria:
     min_count: 1
     weight: 1.5
     pass_threshold: 1.0
-
-  - type: command_executed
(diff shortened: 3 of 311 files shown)


uipreliga (approved):
🚢

PR #1835, opened 2026-07-02
fix(ci): pre-bake test-manager-tool + solution-tool in the eval image

Mihaiii:
Pre-bake both tools next to the other nine, per this Dockerfile's existing policy: tools the tasks/checkers invoke are installed at build time so a feed or env problem fails the image build loudly instead of silently zeroing task scores.


uipreliga (approved):
🚢

PR #1846, opened 2026-07-06
test(uipath-maestro-case): add expense SDD finalize and build evals

abhiram-vad:
## Summary

- Adds an ExpenseReimbursement Phase 0 finalization eval that resumes from the generated `sdd.draft.md` artifact and verifies the finalized `sdd.md` mechanically and semantically.
- Adds an ExpenseReimbursement build-from-SDD eval that stages the generated `sdd.md`, proceeds past downstream hard stops, requires the agent to run `uip maestro case validate`, and validates the generated `caseplan.json`.
- Adds deterministic trigger fidelity coverage so the expense design must preserve the `expense_requests` record-created event, companion `expense_documents` / `expense_comments` objects, and trigger-sourced intake field mappings.
- Adds `check_expense_reimbursement_case.py` to assert the built case keeps the event trigger, 5-stage happy path, Rejected/Withdrawn terminal lanes, happy + terminal case exits, full task-type mix, Finance-to-Payment decision gating, finance-selected payment method usage, and Payment Tracking child case.

## Review Fixes

- Pins the checker to `ExpenseReimbursement/ExpenseReimbursement/caseplan.json` when present, so a valid Payment Tracking child case with its own `caseplan.json` does not make the grader fail on “multiple caseplan” discovery.
- Tightens completeness checks so disconnected terminal lanes, unconditional Payment transitions, Payment tasks that ignore `selectedPaymentMethod`, and Manual trigger downgrades fail deterministically.
- Removes the unavailable `AskUserQuestion` tool from the build eval’s allowed tools and adds a `command_executed` assertion for the agent’s validate run.
- Updates the uipath-maestro-case SDD guidance so tenant case-entity / business data-object record-created starts are authored as `Intsvc.EventTrigger` rows and preserved as unresolved placeholders if tenant registry data is missing, instead of being rewritten to Manual.

## Trigger Note

`expense_requests` is the customer/business data object named by the scenario, not a built-in object guaranteed to exist in the eval tenant. The SDD and generated caseplan should preserve it as the event source; a real customer tenant would need the corresponding entity/resource provisioned or mapped during registry/resource resolution. If it is unresolved during eval/build, the case should keep an event-trigger placeholder, not become a Manual case.

## Verification

- YAML parse for expense interview, finalization, and build task files.
- `sdd_check.py` passes on both staged expense SDD fixtures.
- `check_expense_trigger_sdd.py` passes on both staged expense SDD fixtures.
- `check-cli-verbs.py --json` passes on the expense interview, finalization, and build task files using the Python 3.9 compatibility shim for missing `sre_parse.POSSESSIVE_REPEAT`.
- `python3 -m py_compile` for `check_expense_reimbursement_case.py` and `check_expense_trigger_sdd.py`.
- `git diff --check`.
- Checker synthetic coverage: valid ExpenseReimbursement caseplan passes even with a sibling PaymentTracking `caseplan.json`; mutations fail for Manual trigger downgrade, orphaned Rejected lane, ungated Payment transition, and Payment tasks missing `selectedPaymentMethod`.

## Not Run

- Full `coder-eval` task execution; `tests/.venv/bin/coder-eval` is not installed in this worktree.


skills/uipath-maestro-case/assets/templates/sdd-template.md
```diff
@@ -207,6 +207,13 @@ DO NOT include in Configuration:
 - Meta notes like `No required event parameters` or `No user filter` (absence is the default; the skill discovers required params at `case spec` time).
 - Connector activity slug, HTTP method, or any spec-discovered detail.
 
+> **Tenant object starts are still event triggers.** If the user says a case starts
```
charlesliu9, line 210: nit: reformat may increase the readability

charlesliu9 (commented):
some lines seems copied from other windows, reformatting may help on readability.
Update: nvm, it seems Claude used to run in this way

charlesliu9 (approved):

PR #1900, opened 2026-07-06
feat(ci): support antigravity in run-coder-eval and bump coder_eval to 0.8.2

bai-uipath:
Adds antigravity (Gemini) as a third agent option in the `run-coder-eval` workflow, alongside `claude` and `codex`, so any task glob can be run against it on demand. Also adds a model-override input that applies to every agent (e.g. run `claude` on opus/haiku, or pin a specific codex/gemini model), removes the concurrency group so any number of runs can execute at once, and bumps the pinned `coder_eval` version to 0.8.2 (antigravity support first shipped in 0.8.1; 0.8.2 adds the `env_path_prepend` fix — see Review follow-ups below).

The `GEMINI_API_KEY` repo secret has been added for antigravity auth.

## Validation

**1. Nightly pipeline dry-run on 0.8.1** — [Azure build 12593759](https://dev.azure.com/uipath/coder_eval/_build/results?buildId=12593759&view=results). The actual `UiPath.Coder_Eval.Nightly` pipeline run in dry-run mode, pinned to coder_eval 0.8.1, over the `skills,activation` suites on both Linux (docker) and Windows (tempdir), on the default claude/Bedrock config. Dry-run selects one smoke task per skill plus an activation sample and gates on ≥90% pass with zero errors. **Succeeded** — confirms the 0.8.1 pin doesn't break the nightly's setup, either OS driver, or the scoring plumbing.

**2. Per-harness smoke runs** — the same offline task ([`uipath-solution/new_smoke.yaml`](https://github.com/UiPath/skills/blob/bai/run-coder-eval-antigravity/tests/tasks/uipath-solution/new_smoke.yaml): scaffold a solution via `uip solution init`, no tenant) run once against each harness on this branch. Each was verified from the container logs to route to its own agent backend and pass 1/1:

- claude — [run 28831393416](https://github.com/UiPath/skills/actions/runs/28831393416) → `agents.claude_code_agent`
- codex — [run 28831394470](https://github.com/UiPath/skills/actions/runs/28831394470) → `agents.codex_agent` (gpt-5.4)
- antigravity — [run 28831395466](https://github.com/UiPath/skills/actions/runs/28831395466) → `agents.antigravity_agent` (gemini-3.1-pro-preview)

## Review follow-ups

- **`env_path_prepend` blocker (mock `uip` shim bypassed)** — fixed upstream in coder_eval [#487](https://github.com/UiPath/coder_eval/pull/487), released in 0.8.2; this PR now pins 0.8.2, so mocked-replay tasks correctly shadow the real `uip` CLI. The 0.8.1 validation above still holds: 0.8.2 is 0.8.1 + #487, and `new_smoke.yaml` is a no-mock task the fix doesn't alter.
- **Local experiment-model leak** — running antigravity locally against an experiment YAML that pins a claude/codex model (e.g. `model: claude-sonnet-4-6`) overrides the gemini default and 404s. Pass `--model gemini-3.1-pro-preview` explicitly for local antigravity runs. CI is unaffected: the workflow's `agent_model` input defaults antigravity to `gemini-3.1-pro-preview`.


dmorosanu (approved):

dmorosanu:
Validation findings from running the antigravity agent on this branch (worktree, local runs, gemini-3.1-pro-preview):

**1. Experiment model leaks into the antigravity agent (local footgun).** With the standard experiment YAMLs (`model: claude-sonnet-4-6`), config-model precedence overrides the agent's gemini default and the run dies in seconds with a Gemini 404. Local runs require an explicit `--model gemini-3.1-pro-preview`. The CI model-override input added here covers CI; worth a README/PR-description line for the local path (or an agent-side guard that rejects/ignores non-gemini model ids).

**2. Blocker: `env_path_prepend` is not honored — mock `uip` shim bypassed.** `agents/antigravity_agent.py` documents the PATH-prepend hook as unused. Effect on any mocked replay task: the agent silently calls the real `uip` on the host with the developer's live tenant auth — observed concretely (final answer cited artifacts that exist only on the live tenant, absent from the scenario's mocks). This is the same defect class fixed for the codex agent in `8a53bbe` (codex applies the prepend; antigravity does not). Consequences: (a) replay results are contaminated and score 0 against fixtures, (b) the model-under-test executes commands against a real tenant while the harness believes it's sandboxed — mutation-capable, credential-bearing, and invisible in grading. Note the branch's validation task (`new_smoke.yaml`) is a no-mock task, so this path was never exercised.

Recommendation: port the `8a53bbe` PATH-prepend handling to `antigravity_agent` before this merges, or explicitly document the agent as live-only and unsafe for mocked replay suites.

bai-uipath:
1 - default model for the pipeline is set to gpt 5.4 for codex and gemini 3.1 pro for antigravity. No model id needed by the user.

2 - Fixed by bumping version to 0.8.2 which has https://github.com/UiPath/coder_eval/pull/487

uipreliga (approved):

PR #1917, opened 2026-07-07
chore(uipath-maestro-flow): tag DevCon tasks with path-to-ga

uisherif:
## Summary
- Adds `path-to-ga` to the `tags:` list on the 5 billing test tasks under `tests/tasks/uipath-maestro-flow/multi_node/`, so this GA-readiness bucket can be sliced with `--tags path-to-ga`.
- `devcon_expense_approval.yaml` is excluded — per [review feedback](https://github.com/UiPath/skills/pull/1917#issuecomment-4909352810), it's a structural-only (validate-only) test, and it's already mistagged `e2e`. Bundling it into `path-to-ga` here would conflate that pre-existing taxonomy issue with this change; fixing the `e2e`/`devcon` tagging on that file is left for a separate PR.

## ⚠️ Flag for reviewers
`path-to-ga` does **not** fit the closed tag taxonomy documented in `tests/README.md` (§ Tag Taxonomy): every existing tag maps to one of `skill` / `tier` / `mode:` / `lifecycle:` / `shape:` / `node:` / `resource` / `connector` / `windows` / `feature:`, and rule 5 says new tag values should be proposed in the PR rather than invented inline. This is a new cross-cutting dimension outside that taxonomy — flagging so reviewers can decide whether to (a) accept it as a new documented dimension, (b) fold it into an existing one, or (c) request a different approach. The tag-taxonomy lint mentioned in the README may also need updating to allow it.

## Tasks tagged
- `tests/tasks/uipath-maestro-flow/multi_node/billing_discrepancy_detector/billing_discrepancy_detector.yaml`
- `tests/tasks/uipath-maestro-flow/multi_node/billing_dispute_analyst/billing_dispute_analyst.yaml`
- `tests/tasks/uipath-maestro-flow/multi_node/billing_dispute_resolution/billing_dispute_resolution.yaml`
- `tests/tasks/uipath-maestro-flow/multi_node/billing_invoice_lookup/billing_invoice_lookup.yaml`
- `tests/tasks/uipath-maestro-flow/multi_node/billing_resolution_writer/billing_resolution_writer.yaml`

## Test plan
- [ ] Confirm CI's tag-taxonomy lint passes (or update it if `path-to-ga` needs to be added to the allowed vocabulary)
- [ ] `make tags TAGS="path-to-ga"` selects exactly these 5 tasks

### DIFF
--- a/tests/tasks/uipath-maestro-flow/e2e/devcon_expense_approval.yaml
+++ b/tests/tasks/uipath-maestro-flow/e2e/devcon_expense_approval.yaml
@@ -7,7 +7,7 @@ description: >
   human-in-the-loop skills work together across the full authoring lifecycle.
   Validate-only: inline HITL nodes block on human review and cannot be `flow debug`-ged
   headlessly, so correctness is enforced via the semantic checker rather than runtime.
-tags: [uipath-maestro-flow, uipath-human-in-the-loop, e2e, devcon]
+tags: [uipath-maestro-flow, uipath-human-in-the-loop, e2e, devcon, path-to-ga]
 
 run_limits:
   expected_turns: 29
--- a/tests/tasks/uipath-maestro-flow/e2e/devcon_expense_approval.yaml
+++ b/tests/tasks/uipath-maestro-flow/e2e/devcon_expense_approval.yaml
@@ -7,7 +7,7 @@ description: >
   human-in-the-loop skills work together across the full authoring lifecycle.
   Validate-only: inline HITL nodes block on human review and cannot be `flow debug`-ged
   headlessly, so correctness is enforced via the semantic checker rather than runtime.
-tags: [uipath-maestro-flow, uipath-human-in-the-loop, e2e, devcon, path-to-ga]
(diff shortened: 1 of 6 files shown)


rockymadden (approved):

rockymadden (commented):
I wonder if we should make the tag less specific to path to GA and more like a generally applicable tag, add to taxonomy, etc. Something like `importance:critical`

bai-uipath:
e2e/devcon_expense_approval should probably be excluded - it's a structural only test (ie having the e2e tag at all is wrong here).

sherifmak:
we discussed whether we give it "customer-use-case" or "e2e" but path to GA seemed appropriate since it focuses us on to the main goal of getting to GA. we can update the tags post this.

uisherif:
Fixed — reverted the `path-to-ga` tag on `devcon_expense_approval.yaml`, so this PR now only touches the 5 billing tasks. Left the pre-existing `e2e`/`devcon` mistagging on that file for a separate PR.

PR #1925, opened 2026-07-08
fix(uipath-troubleshoot): accept root flags and declared positionals in manifest command verifier

ShashankKarukonda-UiPath:
The `skill-troubleshoot-smoke-manifest-commands` verifier rejects three valid mocked-command shapes that landed with the DAP eval scenarios (#1587): bare `uip --version`, and kebab-case positional values — `is connections list <connector-key>` / `is resources describe <connector-key> <object-name>` — which the tokenizer can only parse as subcommand path tokens. Verified against the installed CLI schema: `@uipath/integrationservice-tool` declares `list [connector-key]` and `describe <connector-key> <object-name>` as positionals, and `-v, --version` is a root flag. These are verifier false positives, not fixture drift.

#1587 itself merged green because test-only PRs select only the changed tasks filtered by `--tags smoke` — its run logged `Running 0 task(s)`, so the manifest verification never executed. The breakage is latent and fails the smoke for **any** PR touching `skills/uipath-troubleshoot/**` (first hit: #1924).

Fix, in `verify_manifest_commands.py`:
- flag-only invocations validate against root `--help` flags instead of being rejected as unparseable
- during the path walk, a token that is not a subcommand is rerouted to the positional count when the parent leaf declares enough `Arguments`; flag and arity checks then run against that leaf

Still fails correctly on: typo'd subcommands (parent declares no/insufficient Arguments), excess positional arity, unknown flags, unknown root flags. Verified locally against the real CLI: all 15 integration-service-scoped commands pass, the three former BADs included.

Unblocks #1924 (ENGCE-59466). Follow-up worth considering separately: run manifest verification when `tests/tasks/uipath-troubleshoot/**` changes too, so fixture-only PRs can't skip it.

### DIFF
--- a/tests/tasks/uipath-troubleshoot/_shared/scripts/verify_manifest_commands.py
+++ b/tests/tasks/uipath-troubleshoot/_shared/scripts/verify_manifest_commands.py
@@ -209,9 +209,22 @@ def validate_shape(
 ) -> tuple[bool, str]:
     """Validate a (path, flag-set, positional-count) shape against the CLI."""
     if not path:
+        # A flag-only invocation (`uip --version`) is valid when every flag is
+        # a root-level flag — check against root --help instead of rejecting.
+        root_help = fetch_help(uip_bin, ())
+        allowed = (flag_names(root_help) if root_help else set()) | GLOBAL_FLAGS
+        if used_flags and all(f in allowed for f in used_flags):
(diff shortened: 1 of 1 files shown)


ShashankKarukonda-UiPath:
Closing — keeping the docs PR (#1924) free of test-infra changes; the verifier false-positive is left to the fixtures' owning team to address.

PR #1929, opened 2026-07-08
feat(hooks): twin .sh/.ps1 session hooks with polyglot dispatch for zero-install cross-platform support

RaduAna-Maria:
## Why

Session hooks required bash — absent on Windows machines without Git Bash, the plugin's primary enterprise environment. And no shell ships by default on **both** Windows and macOS, so a single-language port always leaves one platform needing an install. This PR ships every session hook as **twin implementations** — `.sh` (bash) + `.ps1` (Windows PowerShell 5.1 / pwsh 7+) — dispatched by a single bash/PowerShell **polyglot command** per `hooks.json` entry. Zero installs on every platform, and the same registration works under Codex.

## How the dispatch works

Each `hooks.json` entry registers one command, no `shell` field (relies on the runner's default shell selection):

```
echo `# <#` >/dev/null
bash "${CLAUDE_PLUGIN_ROOT}/hooks/<name>.sh"
exit $?
: <<'POLYEOF' #> > $null
& "${CLAUDE_PLUGIN_ROOT}/hooks/<name>.ps1"
if ($null -eq $LASTEXITCODE) { exit 1 } else { exit $LASTEXITCODE }
POLYEOF
```

sh-family shells: the backtick substitution is a comment → run the `.sh` → exit; the PowerShell branch is quoted-heredoc **data**, so even shells that parse the whole command up front (zsh) accept it. PowerShell: `` `# `` is a literal `#`, `<# … #>` block-comments the sh branch away, `> $null` silences the echo → runs the `.ps1` with explicit exit-code propagation (`exit N` in a `&`-invoked script otherwise collapses to 1 under `-Command`). Never put `#>` in the sh branch; keep the heredoc wrapper (documented in CONTRIBUTING.md § Hooks).

## Runner coverage (all verified, not inferred)

| Runner / platform | Hook shell | Branch | Verified |
|---|---|---|---|
| Claude Code — Linux | `sh -c` (dash on Debian/Ubuntu) | `.sh` | dash + bash in WSL Ubuntu |
| Claude Code — macOS | `sh -c` | `.sh` | dash is the stricter superset of macOS `sh` |
| Claude Code — Windows + Git Bash | Git Bash | `.sh` | live session: full event stream |
| Claude Code — Windows w/o Git Bash | PowerShell (pwsh → 5.1 fallback, `-ExecutionPolicy Bypass`) | `.ps1` | PS 5.1 `-Command`, incl. failure exit codes |
| Codex — macOS/Linux | `$SHELL -lc` (login shell — zsh on default macOS) | `.sh` | bash/zsh/dash `-lc`; zsh is why the heredoc exists |
| Codex — Windows | PowerShell | `.ps1` | live `codex exec` 0.142.4: SessionStart hook ran the `.ps1` twin with full stdin payload |

Codex reads plugin `hooks/hooks.json` in a Claude-compatible schema and substitutes `${CLAUDE_PLUGIN_ROOT}` itself; its `async` field is parsed-but-unsupported and it has no `SessionEnd`/`StopFailure` events (those entries simply never fire — matches the existing TELEMETRY.md cross-agent contract). Known limitation: a user whose *login shell* is fish (Codex POSIX only) gets a non-fatal parse failure — the hook is skipped.

## What changed

- **`hooks/*.ps1` twins** of the session hooks alongside the restored `.sh` originals (`suggest-permissions`, `set-session-env`, `send-telemetry`). Behavior contracts identical; `send-telemetry.ps1` uses `ConvertFrom-Json` for the same region-scoped extraction the `.sh` does with its string-aware awk pass.
- **BREAKING: the `ensure-uip` SessionStart hook is removed** (both twins + registration). The plugin no longer auto-installs `@uipath/cli` / `@uipath/rpa-tool` at session start — the CLI must already be on `PATH`; telemetry silently drops events when it is absent.
- **Twin sync rule**: any change to one twin requires the equivalent change to the other in the same PR — recorded in CLAUDE.md, CONTRIBUTING.md, the PR-review hook checklist, and `TWIN SCRIPT` headers in all eight scripts.
- **Contract guards parametrize over both twins** (`bash` + `pwsh`, both preinstalled on ubuntu runners) — CI enforces the sync rule.
- **`SessionEnd` is synchronous (`"timeout": 30`)**: measured on Claude Code 2.1.202 — async hooks still running at session teardown are killed after a <2.5s grace window, silently dropping the session-end event. All other events stay `"async": true`.

## Verification

- Polyglot matrix above, each cell tested with the exact registered command strings against the real hooks (dispatch, silence, exit-code propagation incl. `exit 2`/`exit 7`, stdin passthrough).
- `.ps1` twins: local harnesses mirroring the pytest guards — 36/36 telemetry assertions (event mapping, region scoping, Codex payload shapes, sanitization, drop paths), 6/6 set-session-env, 4/4 suggest-permissions.
- E2E: fresh `claude -p --plugin-dir` sessions delivered all five event shapes (`session-start`, `tool-use`, `completion`, `session-end`) through both branches; one unmocked run shipped to App Insights (session `4cabf753-7209-4789-b890-a3f74de5c3ee`).
- Session exit waits for the session-end handoff (~1–2s, 30s cap); all other hook events remain fully non-blocking.


gabrielavaduva (changes requested):
pwsh 7+ is not on macOS as default, we'll need something to work on both, or register 2 hooks? let's discuss offline

gabrielavaduva (approved):

PR #1562, opened 2026-06-18
feat(uipath-coded-apps): add default UI styling opt-in in Coded Apps

amrit-agarwal-1:
Web-app scaffold now offers a Q6 default-styling opt-in:

- yes (recommended): apollo-wind + apollo-core components, Apollo Vertex design tokens, and a next-themes Sun/Moon dark-mode toggle out of the box
- no: bare Tailwind so the user can bring their own component library

Both Q6 paths now share a single SDK initialization pattern: new UiPath() (no config) reads clientId/scope/orgName/tenantName/baseUrl/ redirectUri from <meta name="uipath:*"> tags injected by @uipath/coded-apps-dev locally (from uipath.json + .uipath/) or by the platform in production. The .env / VITE_UIPATH_* env-var path is gone.

Tailwind is now v4 (CSS-first) on both paths — no tailwind.config.js, configuration lives in src/index.css.

Skill-wide docs (SKILL.md env-var section, debug.md remediation steps, oauth-client-setup.md, oauth-scopes.md, validation-station.md, patterns.md buildTaskUrl helper) updated to reference uipath.json + uip login instead of VITE_UIPATH_* / .env, and to drop legacy-comparison framing.

PR #1930, opened 2026-07-08
fix(uipath-coded-apps): update test criteria for the uipath.json config convention

ninja-shreyash:
## Summary

Skill change #1562 replaced the `.env`/`VITE_UIPATH_*` SDK config convention with `uipath.json` (single config source, meta-tag injection) and rewrote `debug.md`'s canonical fixes to match. Two coded-apps tests still graded the old convention and failed every nightly (`2026-07-07_04-22-21`, `2026-07-08_04-20-05`) while the agent behaved correctly — the grader lagging the skill, not agent or skill defects. Post-#1562 a `VITE_UIPATH_*` `.env` is not a legacy variant but a non-functional artifact (the SDK reads only the `uipath.json`-injected meta tags), so the criteria grade `uipath.json` exclusively — no legacy fallback (review: [r3543030742](https://github.com/UiPath/skills/pull/1930#discussion_r3543030742)).

## Fixes

**`e2e_orchestrator_dashboard_web_app`**
- The three config criteria read `uipath.json` fields (`scope` covers `OR.Jobs`+`OR.Queues` — Rule #16; api-subdomain `baseUrl` — Rule #8; config-exists requires the `scope` key only, since `clientId` is legitimately an empty placeholder or omitted in test tenants).
- Turn-budget hardening after run 28935883495 exhausted `max_turns: 65` with every lifecycle criterion already passed: the prompt now writes `report.json` (also the `post_run` cleanup handle) immediately after deploy instead of last, bans task-tracking-tool ceremony (~14 turns in the failed run), and `max_turns` is 85 — even well-behaved runs live near 65.

**`integration_debug_catalog_classification`** — `debug.md`'s 401-scope fix is now "update the `scope` field in `uipath.json`", so the Fault B regex (`VITE_UIPATH_SCOPE|OR\.Tasks`) matched neither the dead env var nor the new canonical text: agents were failed for quoting the skill verbatim. Fault A/B regexes accept the `uipath.json`-era phrasing (bounded — vague answers like "check your scopes" still fail, and naming the dead `VITE_*` vars now correctly fails too).

## Verification

- **Full suite green on this branch: [51/51](https://github.com/UiPath/skills/actions/runs/28945485276)** (35 coded-apps + 16 dashboard), including both fixed tests and the hardened e2e.
- `e2e_orchestrator_dashboard_web_app`: nightly failure reproduced exactly pre-fix (0.868, same 3 criteria); post-fix criteria verified against both observed run artifacts (clientId omitted / empty) and negative configs (wrong scope, `cloud.uipath.com`); full local run **1.000 (22/22)** with zero task-tool calls and `report.json` written post-deploy.
- `integration_debug_catalog_classification`: reproduced pre-fix (0.773, Fault B — the agent's diagnosis quoted the new debug.md verbatim), post-fix **1.000 (6/6)**; regexes verified against the real `diagnosis.json`, vague answers, and dead-env-var answers.

## Follow-up (separate PR)

Nine tests pass today only because their prompts explicitly pin `.env.example`/`VITE_UIPATH_*` — they should migrate to the `uipath.json` convention so Rule-16 coverage grades what the skill actually teaches (`web_sdk_oauth_init` is the most affected: its prompt still teaches `import.meta.env` sourcing).

### DIFF
--- a/tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
+++ b/tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
@@ -192,8 +192,8 @@ success_criteria:
   # the three config criteria below accept EITHER convention so the test grades
   # the config contract, not which skill version taught it.
   - type: run_command
-    description: "SDK config exists: uipath.json with clientId+scope fields (current convention) OR .env/.env.example with VITE_UIPATH_* (legacy)"
-    command: "( F=$(find . -name 'uipath.json' -not -path '*/node_modules/*' -print -quit) && test -n \"$F\" && python3 -c \"import json,sys; d=json.load(open('$F')); sys.exit(0 if 'clientId' in d and 'scope' in d else 1)\" ) || find . \\( -name '.env.example' -o -name '.env' -o -name '.env.local' \\) -not -path '*/node_modules/*' -exec grep -lE '^VITE_UIPATH_' {} + | grep -q ."
+    description: "SDK config exists: uipath.json with a scope field (current convention; clientId may be an empty placeholder or omitted entirely in test tenants, so only scope is required) OR .env/.env.example with VITE_UIPATH_* (legacy)"
+    command: "( F=$(find . -name 'uipath.json' -not -path '*/node_modules/*' -print -quit) && test -n \"$F\" && python3 -c \"import json,sys; d=json.load(open('$F')); sys.exit(0 if 'scope' in d else 1)\" ) || find . \\( -name '.env.example' -o -name '.env' -o -name '.env.local' \\) -not -path '*/node_modules/*' -exec grep -lE '^VITE_UIPATH_' {} + | grep -q ."
     timeout: 10
     expected_exit_code: 0
     weight: 1.0
--- a/tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
+++ b/tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
@@ -4,7 +4,7 @@ description: >
   App dashboard that surfaces live Orchestrator state, from scratch. Given
   only a business goal, the agent must scaffold Vite + React, wire the
   `@uipath/uipath-typescript` SDK with subpath imports and constructor DI,
-  configure the SDK scope (uipath.json / legacy .env) to cover Jobs + Queues, configure
+  configure the uipath.json scope to cover Jobs + Queues, configure
   `vite.config.ts` correctly (Rule #9), use `getAppBase()` for any runtime
   asset URLs (Rule #10), build, pack, publish (default Web type), and
   deploy to a fresh per-run Orchestrator folder so `post_run` can clean it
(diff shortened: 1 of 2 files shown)


tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
```diff
@@ -4,7 +4,7 @@ description: >
   App dashboard that surfaces live Orchestrator state, from scratch. Given
   only a business goal, the agent must scaffold Vite + React, wire the
   `@uipath/uipath-typescript` SDK with subpath imports and constructor DI,
-  configure `VITE_UIPATH_SCOPE` to cover Jobs + Queues, configure
+  configure the SDK scope (uipath.json / legacy .env) to cover Jobs + Queues, configure
```
Raina451: lets remove "/ legacy .env" at all.

deepeshrai-tech (approved):

tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
```diff
@@ -187,25 +187,29 @@ success_criteria:
...
     pass_threshold: 1.0
 
+  # Skill change #1562 (Jul 6) replaced the .env/VITE_UIPATH_* SDK config with
+  # uipath.json ({clientId, scope, orgName, tenantName, baseUrl, redirectUri});
+  # the three config criteria below accept EITHER convention so the test grades
+  # the config contract, not which skill version taught it.
   - type: run_command
-    description: ".env or .env.example was created with VITE_UIPATH_* config (anywhere in the project tree)"
-    command: "find . \\( -name '.env.example' -o -name '.env' -o -name '.env.local' \\) -not -path '*/node_modules/*' -exec grep -lE '^VITE_UIPATH_' {} + | grep -q ."
+    description: "SDK config exists: uipath.json with clientId+scope fields (current convention) OR .env/.env.example with VITE_UIPATH_* (legacy)"
```
Raina451: isnt our skill always uses uipath.json instead of env for sdk config?

Raina451 (approved):
please check if we still want to mention legacy env, rest lgtm

tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
ninja-shreyash: no this got updated in this PR - https://github.com/UiPath/skills/pull/1562/changes#diff-d8f58212e2cbeabca4bcd8e594f1dda40bfebf4564108883452c6c6a34a4d608

ninja-shreyash: updated this

tests/tasks/uipath-coded-apps/e2e_orchestrator_dashboard_web_app.yaml
ninja-shreyash: updated

ninja-shreyash: Done — removed the legacy `.env` acceptance entirely (description + all three criteria), and also dropped the dead `VITE_*` alternatives from the debug-catalog regexes in the same pass: post-#1562 the SDK reads only the uipath.json-injected meta tags, so a `VITE_UIPATH_*` .env isn't a legacy variant but a non-functional artifact, and naming `VITE_UIPATH_SCOPE` as a fix is now a wrong answer (verified it's rejected). Criteria re-verified against both observed run artifacts (3/3) and the real diagnosis.json.

PR #1932, opened 2026-07-08
fix(uipath-ixp): ambiguous entity rename must ask, not guess (ACTV-88672)

alexandrujircan:
## Summary

Smoke test for [ACTV-88672](https://uipath.atlassian.net/browse/ACTV-88672) plus the skill fix it exposed.

- **New smoke** `tests/tasks/uipath-ixp/smoke/rename_ambiguous_entity.yaml` (`skill-ixp-rename-ambiguous-entity-smoke`), discovery variant: the prompt states ONLY the ambiguous goal — *"In my IXP workspace, rename subscriptions into 'Test Invoice'"* — with no workspace facts. A mock overlay (`_shared/mock_template_ambiguous/`) serves canned JSON for the read verbs: `projects list` reveals a project **titled** "subscriptions"; `projects get-taxonomy` reveals a field group "subscriptions" containing a field "subscriptions". PASS = the agent discovers the collision itself and asks which entity is meant (via `clarification.txt` — the harness is non-interactive) while executing **no** rename.
- **Skill fix** `skills/uipath-ixp/SKILL.md`: new Critical Rule 16 — ambiguous entity reference → ask which entity kind (project/group/field/data type) before any mutation; never guess, never mutate several candidates "to cover all cases".

The rule exists because the test caught the failure first: without it, the agent under test renamed **both** the group and the field without asking. With Rule 16 the agent lists projects, reads the taxonomy, and writes a clarification enumerating all three candidates with their discovered details.

## Grading design

- Negative guards read `mocks/calls.log` — the mock's record of **actually executed** `uip` invocations — via `file_contains` `excludes:`. `command_not_executed` regexes over Bash text false-match commands merely *quoted* in a correct clarification (this bit us in CI).
- A **discovery** criterion requires `projects list` AND `projects get-taxonomy` in the call log — it blocks a blind generic "which one do you mean?" (asked without looking) and doubles as the positive control: executed reads must land in the graded file, so a re-pointed log sink fails the test instead of making the excludes pass vacuously.
- `calls.log` is not dot-prefixed so it lands in the CI eval-report artifact (upload-artifact skips hidden files by default) — failed runs can be debugged from the artifact directly.
- `clarification.txt` is not a self-report: it is the ask-behavior under test, the only observable form of asking in a non-interactive run.

## Testing

- Ran `skill-ixp-rename-ambiguous-entity-smoke` locally with `experiments/smoke.yaml` after the final (discovery) design and it passed — 6/6 criteria, weighted score 1.000. The preserved sandbox shows the ideal trace: `projects list` → `get-taxonomy` → clarification citing discovered facts (project Name slug, sibling field, field kind), zero mutations.
- Regression: re-ran the three sibling rename smokes (`rename-project`, `rename-field`, `rename-field-group`) after adding Rule 16 — 3/3 passed, score 1.000; the rule does not cause over-asking on unambiguous renames.
- `scripts/check-cli-verbs.py`: no findings. Mock dispatch paths unit-tested directly (list/get/get-taxonomy/mutations/unknown-verb + logging).

[ACTV-88672]: https://uipath.atlassian.net/browse/ACTV-88672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/tasks/uipath-ixp/smoke/rename_ambiguous_entity.yaml
```diff
@@ -0,0 +1,70 @@
+task_id: skill-ixp-rename-ambiguous-entity-smoke
+description: >
+  Smoke: ambiguity guard (ACTV-88672). The workspace has a project TITLED
```
cezara98t: no need for ticket number

alexandrujircan: Removed in e9e564c.

alexandrujircan:
**CI smoke failure analysis (run 28997688778):** 35/37 (94.6% < 95% gate). Both failures were grading false-positives, not agent misbehavior:

- `skill-ixp-rename-ambiguous-entity-smoke` — the agent behaved correctly (asked, wrote `clarification.txt` naming all three candidates) but wrote the file via a bash heredoc whose text **quoted** the candidate rename commands. `command_not_executed` regexes the whole Bash call text, heredoc included, so a quoted `uip ixp projects update-title` counted as executed.
- `skill-ixp-negative-guards-smoke` (pre-existing) — same class: an explanatory `#` comment naming `--corrections f-100` (to say why NOT to use it) matched the forbidden-flag regex. Stochastic flake; candidate for a separate fix using the approach below.

**Fix (34c19b0):** the guards now assert on `mocks/.calls.log` — the mock's record of *actually executed* `uip` invocations — via `file_contains` `excludes:` (pure-Python checker; also works on Windows dev machines where `run_command` executes under cmd.exe). The mock template seeds the log so the path always exists; the seed line is the `includes:` anchor. Pattern documented in the mock template README.

Re-ran locally after the change: `skill-ixp-rename-ambiguous-entity-smoke` passed 4/4, weighted score 1.000. With this fix the suite is 36/37 (97.3%) even if `negative-guards` flakes again.

tests/tasks/uipath-ixp/_shared/mock_template_ambiguous/README.md
cezara98t, line 1: this is pretty specific to this new test case, while the name of the file and the description are general

alexandrujircan: Yes, this section is exactly for that test. It was added to serve as a kind of README file in which each mock will be explained in the future. If you think it's useless, I can remove it.

cezara98t (approved):

andrei-uipath (approved):

PR #1820, opened 2026-07-02
refactor(troubleshoot): tiered investigator (v2 architecture)

dmorosanu:
## Why

Eval data across 184 replay scenarios showed the previous 7-sub-agent orchestration (triage → scope-checker → generator → sequential testers → depth-verifier → presenter) cost 25–55 min per investigation while its marginal value sits in the playbook knowledge, not the choreography: 159/184 scenarios pass without the skill, 12 more flake-pass on retry, and the 13 stable-fails are all cases where a playbook fact or decision-tree discriminator is required. This PR keeps the knowledge and removes the choreography.

## What changed — three-tier architecture

~~- **Tier 0 — signature-index routing.** Every playbook declares `signatures:` frontmatter (currently 226 playbooks, 682 signatures, 30 justified `silent: true`); `scripts/build-signature-index.py` generates `references/signature-index.md` (grep-only routing table + no-signature symptom routing + signal-extraction cheatsheet) and lints: every playbook routable or silent, duplicate (kind,value) claims require discriminating notes, exclusion targets must exist, localizable message signatures get a warning (see #1842 below). Playbook bodies untouched.~~
- **Tier 1 — single-context investigation (the fast path, and the common case).** SKILL.md rewritten as a ~150-line protocol run entirely in one context: anchor → extract signals (mandatory AggregateException/inner-exception unwrap; route on language-invariant signals first, translating localized messages before index grep) → route via the signature index → walk the matched playbook's decision tree → mandatory format-forced verification checklist (cause named verbatim, evidence pinned vs sibling causes, runtime-evidence gate, resolution-branch alignment, causal precedence, fix scope incl. no unevidenced solution mechanisms) → present per `references/presenting.md` (discriminating datum carried into the final answer; interactive Healing-Agent apply-flow; approval gate: user source files are never modified without explicit approval, decline/non-answer = no edit). The previous Critical Rules section is restructured into `## 1. Invariants` — every load-bearing rule preserved or strengthened; the removed rules were orchestration mechanics that no longer have a boundary to protect.
- **Tier 2 — escalation** (`references/escalation.md`, loaded only on 6 defined triggers): 2–4 parallel read-only hypothesis probes + adjudication + a conditional fresh-eyes verifier; bounded spawn budget. On harnesses without a subagent-spawning tool, the same protocol runs as a **serial in-context fallback** (probes executed sequentially, per-probe blocks in notes.md, self-verify at reduced confidence) — validated end-to-end on a subagent-less harness (#1891).
[...]

PR #1936, opened 2026-07-08
fix(platform): front-load faulted-job root-cause redirect to uipath-troubleshoot in when_to_use

dmorosanu:
## Why

Faulted-job root-cause prompts ("investigate my failed job", raw error messages) intermittently activated `uipath-platform` instead of `uipath-troubleshoot`. The existing redirects sat deep in the `when_to_use` text, past the 1,536-char listing truncation, so the model never saw them.

## What

One line: prepend `NOT for diagnosing failures: 'why did my job/run fail', faulted/stuck job root-cause, error messages→uipath-troubleshoot.` at the head of `uipath-platform`'s `when_to_use`, where it survives truncation. Rebased onto the current `when_to_use` text (which has since gained the Data Fabric query phrases).

## Validation

3 CI rounds × the 4 routing-affected troubleshoot tasks = 12/12 activations routed correctly with this change (validated on the `feat/troubleshoot-v2` branch, see #1820 discussion). Extracted here from #1820 so the platform surface gets its own review; #1820 no longer touches `skills/uipath-platform/`.

## Merge coordination

Independent of #1820 — safe to merge in any order. Without it, troubleshoot activation on faulted-job prompts keeps the pre-fix intermittent miss rate.


MarinRzv (approved):

skills/uipath-platform/SKILL.md
```diff
@@ -1,7 +1,7 @@
...
... bucket / library / webhook / trigger / connector / connection / tenant / folder / robot / package / entity / record / choice set / trace / span / trace feedback / BYO LLM. Also 'upload to UiPath', 'create asset', 'start job', 'list queues', 'deploy a single package to Orchestrator', 'OAuth2 token', 'list trace spans / add trace feedback', 'create entity', 'add field', 'delete field', 'list entities', 'insert record', 'update record', 'delete record', 'query Data Fabric', 'search records', 'show every <thing> tagged/where', 'filter records by tag/choice/field', 'count by group', 'unique tag combinations', 'group by field', 'aggregate COUNT/SUM/AVG', 'create choice-set', 'create choiceset', 'add choice set values', 'delete choice set', 'upload file attachment', 'import CSV', 'load CSV into entity/list', 'load spreadsheet into entity', 'bulk load records from CSV', 'bulk import from CSV', 'register my own LLM key', 'configure a model substitution', 'my BYO LLM key stopped working / returns errors', 're-probe / audit a BYO configuration', 'uipath.com REST'. For `uip solution` ops or `.uipx` deploys→uipath-solution. For Data Fabric connector nodes inside a `.flow`→uipath-maestro-flow."
```
costin-uipath, line 4: Just a note: as per https://code.claude.com/docs/en/skills.md

"| `description`              | Recommended | What the skill does and when to use it. Claude uses this to decide when to apply the skill. If omitted, uses the first paragraph of markdown content. Put the key use case first: the combined `description` and `when_to_use` text is truncated at **1,536** characters in the skill listing to reduce context usage.                                                                                                                               |
| `when_to_use`              | No          | Additional context for when Claude should invoke the skill, such as trigger phrases or example requests. Appended to `description` in the skill listing and **counts toward the 1,536-character cap.** "

""NOT for diagnosing failures:.." is still in the first 1536 chars, but the second half of 'when_to_use' gets trimmed and is irrelevant. Maybe a compression would be useful.

costin-uipath (approved):

dmorosanu:
already fixed by other PR

PR #1938, opened 2026-07-08
uipath-rpa: make analyzer-rules list on-demand instead of an authoring prerequisite

gabrielavaduva:
## What

Demotes `uip rpa analyzer-rules list` from a **mandatory authoring-phase prerequisite** to an **on-demand** command in `uipath-rpa`:

- **`SKILL.md`** — Critical Rule 3 reframed from three-phase to two-phase validation (per-file `validate` + project-level `build`), with an explicit "do NOT pre-fetch the rule list" bullet; removed from the Capture-First flow, the post-`init` call-batching list, and the Key CLI Commands table (now marked on-demand).
- **`references/validation-guide.md`** — "Pre-generation: List Analyzer Rules" becomes "On demand: List Analyzer Rules": run it only when the user asks about the project's best-practice/analyzer rules, or when repeated violations of the same rule family suggest authoring against the full (scoped) rule set.
- **`references/environment-setup.md`**, **`references/ui-automation-guide.md`** — same cleanup in the batching note and capture-first step.

The command's reference documentation (`cli-reference.md § analyzer-rules list`) and the Task Navigation entry ("List project best-practice / analyzer rules") are untouched — it stays fully documented and discoverable when a user asks for it.

## Why

`validate` and `build` already enforce the enabled analyzer rules and report violations with rule IDs and recommendations, so pre-fetching the rule list is speculative cost on every authoring task: the unscoped call enumerates every rule across every package and can take a minute or more, and most of the output never influences the generated code. This was one of the wall-clock/token line items in the PILOT-6197 authoring-cost report ("feed analyzer constraints into the step that needs them, don't front-load them").

## Related issues

- **[UV-15204](https://uipath.atlassian.net/browse/UV-15204)** — "[RPA] [CLI] analyzer-rules step makes coding-agent workflow generation 3-4x slower — 7-8 min vs ~2 min when skipped" (CodingAgents, Queued). This PR is the skill-side mitigation: the step is no longer in the default authoring loop at all.
- **Known caveat — [UV-15226](https://uipath.atlassian.net/browse/UV-15226)** — "`uip rpa analyze` does not report Coded Workflow analyzer rules (e.g. ST-NMG-017) that Studio flags as validation errors" (Queued). This PR's premise ("validate/build enforce the rules, so skip the pre-fetch") holds for XAML authoring; for **coded workflows** there is a gap until UV-15226 is fixed, since some ST-NMG rules surface neither from the CLI validate/build path nor, now, from a pre-fetched rule list. The on-demand escape hatch (run the scoped list when the same rule family keeps biting, or when Studio flags errors the CLI missed) is the interim path for that case.

[UV-15204]: https://uipath.atlassian.net/browse/UV-15204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/skills/uipath-rpa/SKILL.md
+++ b/skills/uipath-rpa/SKILL.md
@@ -113,14 +113,14 @@ For the full decision flowchart, InvokeCode extraction rules, and detailed hybri
 
 ## Capture-First Fast Path
 
-When the request is "automate this dialog/form" or "build a UI test from these manual steps" — i.e. the bulk of the work is target capture, not coding — **defer authoring-phase prerequisites until target capture is complete**. The capture surface is interactive, app-state-sensitive, and time-bound; project-context discovery and analyzer rules add nothing during capture and steal time from it.
+When the request is "automate this dialog/form" or "build a UI test from these manual steps" — i.e. the bulk of the work is target capture, not coding — **defer authoring-phase prerequisites until target capture is complete**. The capture surface is interactive, app-state-sensitive, and time-bound; project-context discovery adds nothing during capture and steals time from it.
 
 **Fast-path order for capture-first tasks.** Read [ui-automation-guide.md](references/ui-automation-guide.md) and [uia-configure-target-workflows.md](references/uia-configure-target-workflows.md) in full first (Rule 7; the second is used in step 3). Then:
 
 1. **Pre-flight Window Baseline** — list top-level windows once; decide whether to launch the app ([§ Pre-flight: Window Baseline](references/ui-automation-guide.md)).
 2. **Inventory targets from manual steps** (Test Manager test case, PDD, or written script). Each "Click X" / "Enter Y" / "Select Z" / "Verify W" step maps to one OR element. Group by screen state ([§ Capturing from Manual Test Steps](references/ui-automation-guide.md)).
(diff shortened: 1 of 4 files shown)


skills/uipath-rpa/SKILL.md
```diff
@@ -149,12 +149,12 @@ uip rpa activities find --query log --output json > /dev/null 2>&1 &
...
-   - **Per-file** (after every create or edit): `uip rpa validate --file-path "<FILE>" --project-dir "<PROJECT_DIR>" --output json` until 0 errors. Catches structural XAML, missing references, analyzer rules, schema violations. Fix one thing per iteration.
+3. **Phase-gated validation.** Two-phase validation:
+   - **Per-file** (after every create or edit): `uip rpa validate --file-path "<FILE>" --project-dir "<PROJECT_DIR>" --output json` until 0 errors. Catches structural XAML, missing references, analyzer-rule violations, schema violations. Fix one thing per iteration.
```
RaduAna-Maria, line 153: Do we want to keep the validation after each edit? There might be a couple of small edits that each waits for validation and leads to a longer execution time. I know that there was an issue with `build` a couple of weeks ago. Let`s discuss

RaduAna-Maria (approved):

PR #1939, opened 2026-07-08
fix(uipath-maestro-flow): read node output shape from schema, don't guess it

rockymadden:
## Problem

The `skill-flow-slack-weather-pipeline` eval failed its end-to-end criterion. The reported fault was `[400300] Error evaluating expression in activity inputs (element getCurrentWeather1)` — but that was a **downstream symptom**.

Actual chain (from the debug runtime globals):
1. A Script read the Slack connector's output via the **raw Slack Web API envelope** — `$vars.slackNode.output.channel.purpose.value`.
2. The UiPath connector **normalizes/unwraps** that response, so the field is flat at `output.purpose.value`. The `.channel` wrapper doesn't exist → the script produced an empty city.
3. Empty city → geocode HTTP returned no `results`.
4. `getCurrentWeather1`'s `=js:$vars.geocodeCity1.output.body.results[0].latitude` threw `Cannot read property '0' of undefined` → flow Faulted.

The HTTP `.output.body` read was **correct**; it only broke because of the empty value upstream. `flow validate` passed because producer outputs are open schemas, so nothing caught it before debug.

Notably, the agent **did** fetch the connector's registry metadata (`registry get --connection-id`) — twice — but parsed only the **input** side, never `outputDefinition.fields`. It had the real output shape available and didn't read it.

## Fix (skill guidance only)

Test and CLI are correct and untouched (the lookup command already exists; the test rightly requires real end-to-end output). Three lean, generalized doc edits:

- **`node-output-wiring.md` (SSOT):** new *"Know the Field Path Before You Reference It"* — the durable, node-type-agnostic principle (read the producer's declared output shape; don't guess; a wrong path passes `validate` and faults at `debug`, often downstream). Per-node specifics deferred to the plugin docs to avoid a second maintenance site.
- **`connector/impl.md` Step 5b:** one line — read `outputDefinition.fields` (the output side of the `registry get --connection-id` call already run), not just inputs.
- **`diagnose/failure-modes.md`:** generalized the open-schema bullet and noted the fault can surface at a downstream node.

Links validated via `.maintenance/check-links.sh` (`checked=801 broken=0`).

## Open follow-up

Couldn't confirm from the eval artifacts whether `outputDefinition.fields` for `get-channel-info` is populated with the flat fields or comes back thin. If thin, the SSOT's "confirm against a real response" fallback is what carries the fix. Worth a live `registry get --connection-id` check on a tenant.


jiyangzh (approved):

skills/uipath-maestro-flow/references/shared/node-output-wiring.md
```diff
@@ -52,6 +52,16 @@ Three failure modes observed in agent-generated `.flow` files:
...
 
+## Know the Field Path Before You Reference It
+
+`=js:$vars.<node>.output.<field>` is only correct if `<field>` matches the producing node's **actual output shape**. Read that shape from the producer before referencing it — never guess it from an external API's documented response, from the inputs you passed that node, or from intuition.
+
+Guessing is silently dangerous: producer outputs are usually **open** schemas, so a wrong `<field>` passes `flow validate` and faults only at `flow debug` — often at a *downstream* node, as `Cannot read property '<x>' of undefined` (see [failure-modes.md](../diagnose/references/failure-modes.md#flow-validate-passes-flow-debug-faults)).
+
+To find the real shape: **CLI-owned nodes** (connector activities, `core.action.http.v2`) expose it via `uip maestro flow registry get <node-type> --connection-id <id>` — read `outputDefinition.fields`, not just the input fields. **User-owned nodes** document their output in the node's plugin `impl.md` (a script's shape is just the object it `return`s). If the declared schema is thin/open, confirm against a real response — a `flow debug` run's runtime `variables`, or a read-only `uip is resources run …`.
```
mrstark14, line 61: For dynamic connectors, the output sometimes depends on the vendor end which is returned during runtime. Won't this cause an issue for those connectors?
Also, is the fix working for the eval?

mrstark14 (commented):
Added a comment. Please check

skills/uipath-maestro-flow/references/shared/node-output-wiring.md
baishalighosh: @rockymadden ^

PR #1945, opened 2026-07-09
test(uipath-platform): fix DF task hygiene — drop stale criterion, ad…

AditiGoyalUipath:
…d smoke cleanup, prompt fixes

- e2e_product_catalogue: drop the "Id key in records update body" criterion — the CLI mandates Id in the update body (verified in `records.ts` and `--help`), so the negative-shape guard asserted absence of a thing the CLI hard-rejects without.
- smoke_choice_relationship: add pre_run/post_run cleanup for CE_SmokeExpense (entity) and CE_SmokePaymentStatus (choice set). Leftover artifacts from prior runs triggered Rule 8 in the agent, which correctly skipped `entities create` / `choice-sets create` and tanked four criteria. Cleanup runs entity-first, then CS, so the in-use reference clears before the CS delete.
- smoke_choice_relationship: remove the "CLI not connected to a live tenant, commands will fail" lines from the prompt — the eval CLI is connected, commands succeed, and the false disclaimer risks the agent silently skipping steps.
- integration_filter_contract / integration_filter_edge_cases: prompt terminology cleanup ("list" → "entity").

### DIFF
--- a/tests/tasks/uipath-platform/data-fabric/e2e_product_catalogue.yaml
+++ b/tests/tasks/uipath-platform/data-fabric/e2e_product_catalogue.yaml
@@ -95,14 +95,6 @@ success_criteria:
     weight: 2.5
     pass_threshold: 1.0
 
-  - type: command_executed
-    description: "Agent included Id key in records update body"
-    tool_name: "Bash"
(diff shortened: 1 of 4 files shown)


SuphalAthlur (approved):
LGTM

PR #1947, opened 2026-07-09
feat(uipath-ixp): make skill aware of default data types (ACTV-89215)

andrei-uipath:
## Ticket
[ACTV-89215](https://uipath.atlassian.net/browse/ACTV-89215) — *Make skill aware of default field types* (DU Cloud).

## Problem
From the DU Cloud Slack thread on the ticket: Claude created its own `Currency Amount` data type instead of reusing the built-in `Monetary Quantity`. As Cezara put it — *"the default field types are not covered in the skill, we should tell it to look at existing types before adding new ones."* The `uipath-ixp` skill documented `data-types add` but never that every IXP project ships with six built-in data types.

## Changes
- **CLI reference — new "Default data types" table** (`references/cli-reference.md`): all six defaults (`Exact Text`, `Inferred Text`, `Number`, `Date`, `Monetary Quantity`, `Boolean`) with their `--kind` / `--input-value` mapping and reuse guidance. Cross-referenced from the `fields add` / `groups add` rows. Names verified against the shared test fixture and the [official IXP docs](https://docs.uipath.com/ixp/automation-cloud/latest/user-guide/creating-and-configuring-fields).
- **Critical Rule 16** (`SKILL.md`): check existing `entity_defs` (`projects get-taxonomy`) and reuse a matching default before `data-types add`; add a new type only when none fits (e.g. a project-specific `Choice`). Uses the ticket's exact `Monetary Quantity` vs `Currency Amount` example.
- **New smoke `reuse_default_type.yaml`**: adding a monetary field must reuse `Monetary Quantity` **and must NOT** run `data-types add` — a direct regression guard for the reported bug.
- **Retargeted `taxonomy_edit_chain.yaml`**: this existing smoke explicitly instructed the agent to create a custom `Monetary Amount` money type — the exact anti-pattern this ticket removes. Since `money` always maps to the default, its chain now uses a `Choice` type (the only kind with no default, so creating it is correct), preserving the test's real purpose (the `data-types add` → `groups add` chain + the `fields add` anti-pattern) without contradicting Rule 16.

## Validation
- `scripts/check-cli-verbs.py`: clean on both tests (one advisory Info on a dynamic `--fields` pattern).
- `hooks/validate-skill-descriptions.sh`: passes (frontmatter `description` unchanged).
- `/lint-task` rubric: both tests verdict **OK** — no self-report, no prompt over-specification (the prompt says "a dollar amount," never names the type — the skill must supply it), meaningful coverage, not gameable without the skill, not a near-duplicate.
- Shape-only smoke tests run in `smoke-skills.yml` (CI injects the alpha token); they can't be executed in a local unauthenticated sandbox.

https://claude.ai/code/session_012ivra1ch9peFjwvqn3PJAc

[ACTV-89215]: https://uipath.atlassian.net/browse/ACTV-89215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tests/tasks/uipath-ixp/smoke/taxonomy_edit_chain.yaml
```diff
@@ -21,17 +26,22 @@ sandbox:
...
 initial_prompt: |
   On project `my_invoices-f1afa9ef-ixp`, set up the taxonomy for
-  invoice header extraction:
+  withholding-tax reconciliation:
 
-  1. Create a reusable `Monetary Amount` data type for representing
-     currency values throughout the project. Money kind, instructions:
-     "A currency value with implicit or explicit currency code, e.g.
-     $1,234.56 or 1234.56 EUR."
+  1. The built-in `Monetary Quantity` type is a generic amount
```
cezara98t: a bit handholding

skills/uipath-ixp/SKILL.md
```diff
@@ -50,6 +50,7 @@ Do not answer these from this skill. Re-activate `uipath-maestro-flow` and follo
...
+16. **Reuse the built-in data types before adding new ones.** Every IXP project ships with six default data types — `Exact Text`, `Inferred Text`, `Number`, `Date`, `Monetary Quantity`, `Boolean`. Before `data-types add`, or before picking a field's `--type` in `fields add`/`groups add`, check existing `entity_defs` (`projects get-taxonomy`) and reuse a matching default for straightforward values — e.g. `Monetary Quantity` for a plain currency amount, **never** a hand-rolled clone like `Currency Amount` that only re-implements the default. Add a new data type only when it carries something no default does — a project-specific `Choice`, or a reusable concept that needs its own tailored extraction instructions (e.g. a `Withholding Tax Amount` money type steering every field to a specific figure). The pre-trained defaults `Monetary Quantity`/`Number`/`Date` keep their fixed output format regardless of instructions, so never spin up a custom type just to reformat one. Full mapping: [CLI Reference § Default data types](references/cli-reference.md#default-data-types).
```
cezara98t: @alexandrujircan you were also adding a rule 16. be aware of conflicts

cezara98t: 1. are we sure that the list of default data types won't change?
2. "e.g. `Monetary Quantity` for a plain currency amount, **never** a hand-rolled clone like `Currency Amount` that only re-implements the default" it makes sense and the example is valid, but would it be flaky without it?(asking due to tokens usage)
3. same questions as 2 for the `Withholding Tax Amount` example; also, this influences the smoke test a lot

andrei-uipath: 1. a new data type at this point means a contract change in Acitivities and uptake in framework so we support that at runtime as well.. It's unlikely we add a new type at this point and if so I do expect we would update the skill. Do you propose letting the wording to be a bit more vague?
2. I'll rephrase for shorter token usage. Good point.
3. same

tests/tasks/uipath-ixp/smoke/taxonomy_edit_chain.yaml
andrei-uipath: Fair — dropped the rationale paragraph in f98806ff; the prompt now just states the task (create the type with these instructions), without explaining the built-in type or why creating one is OK.

skills/uipath-ixp/SKILL.md
andrei-uipath: Ack — #1932 is still open, so whichever PR merges second renumbers its rule (16 → 17) at rebase; the cross-references here say "Critical Rule 16" in two places (Quick Start data-types row + taxonomy_edit_chain description), will update those too if that happens.

andrei-uipath: Done in f98806ff:
1. Dropped the hard "six" count — the rule now says the project's `entity_defs` from `projects get-taxonomy` are the authoritative list, so a new default showing up doesn't make the rule wrong. The CLI-reference table stays as the mapping doc (that one we'd update alongside any contract change).
2. Kept `Currency Amount` but compressed to a parenthetical — it's the ticket's exact bug and what reuse_default_type.yaml guards, so I'd rather spend those ~10 tokens than risk the vague version.
3. Dropped the `Withholding Tax Amount` example from the rule entirely. That also decouples taxonomy_edit_chain.yaml from the rule text — the smoke is no longer answerable by pattern-matching an example in SKILL.md.

andrei-uipath: #1932 merged first — rebased onto main in fc0d0e57: our rule is now Critical Rule 17, cross-references (Quick Start data-types row + both smoke descriptions) renumbered.

tests/tasks/uipath-ixp/smoke/taxonomy_edit_chain.yaml
```diff
@@ -2,6 +2,11 @@ task_id: skill-ixp-taxonomy-edit-chain-smoke
 description: >
   Smoke: agent builds a small taxonomy with `data-types add` then
   `groups add --fields '[…]'` referencing the new type by name.
+  The new type is a money-kind type the agent SHOULD create — not a
+  generic clone of the built-in `Monetary Quantity`, but a reusable
+  concept carrying its own extraction instructions (the value-adding
+  branch of Critical Rule 17 / ACTV-89215; the plain-amount
```
alexandrujircan: I don't think ticket id brings any value here

tests/tasks/uipath-ixp/smoke/reuse_default_type.yaml
```diff
@@ -0,0 +1,51 @@
+task_id: skill-ixp-reuse-default-type-smoke
+description: >
+  Smoke: adding a monetary field, the agent reuses the built-in
+  "Monetary Quantity" data type instead of inventing a redundant custom
+  type (e.g. "Currency Amount"). Regression guard for ACTV-89215.
```
cezara98t: no need for ticket id

cezara98t (approved):

PR #1966, opened 2026-07-09
fix(maestro-case): dedup duplicate resource references at the create-on-missing gate

song-zhao-25:
## Stack

Depends on #2032, which establishes the portable Phase 0 to Phase 1 resource-name contract. This PR contains only the create-on-missing dedup behavior on top of that contract.

## Problem

Create-on-missing resolves task lookups independently. When several tasks reference the same missing resource, selecting them can request duplicate siblings with the same name. An agent and API workflow selected with the same name also collide in the flat `solution_folder.<name>` namespace.

## Changes

- Group the Rule 17 gate and Create selection by `(name, type)`, with all usages listed.
- Partition create-selected usages by canonical pinned I/O and confirmed intent.
- Build identical usages once and bind every usage to that sibling.
- Split and rename I/O-distinct or user-confirmed distinct-intent groups.
- Guard cross-type name collisions before parallel build.
- Apply granted rename write-back atomically to Section 2 and the Section 4 roll-up.
- Extract kind-agnostic inline-build rules into `plugins/tasks/create-inline-common.md`.

Inline creation remains limited to agents and API workflows. Regular RPA processes, Action Apps, child cases, connectors, and agentic processes remain placeholder-only. Connector creation and resolution behavior is unchanged; the two connector references only track the renamed shared gate option.

## Validation

- Skill validation
- `git diff --check`
- Coder-eval schema planning for the affected handoff task
- Stacked feature tree verified byte-for-byte against the pre-split combined PR

## Manual coverage

The existing T8 suite passed the primary dedup branches on alpha (`popoc / CaseManagement`):

| Unit | Branch exercised | Result |
|---|---|---|
| T8 | N-way identical-I/O merge | PASS |
| T8b | `A, B, A` I/O partition | PASS |
| T8b-r | granted write-back rerun | PASS |
| T8b-r-deny | denied write-back rerun | PASS |
| T8c-merge | ambiguous intent, merge | PASS |
| T8c-split | ambiguous intent, split | PASS |
| T8d | three I/O classes | PASS |
| T8e | cross-type same-name rename | PASS |
| T8f | deferred type cannot bridge conflicting pinned types | PASS |

D20 focused smoke coverage passed the distinct-intent and cross-type reason paths. The I/O grant-path atomic write-back remains partially exercised as documented in the prior PR history.


abhiram-vad (approved):
LGTM

charlesliu9 (approved):

PR #1968, opened 2026-07-09
docs(uipath-maestro-flow): connector connection binding name = <connector-key> connection

jiyangzh:
## Problem

A Maestro Flow using a Data Fabric (Data Service) connector node faults at `flow debug` with:

```
[102010] Integration Services invalid value in input
'Connection' has an invalid GUID value: '<bindings.uipath-uipath-dataservice connection>'
```

The connector connection binding never resolves — the raw `<bindings.…>` placeholder reaches Integration Services.

## Root cause — skill doc bug

The flow→BPMN converter resolves a connector's connection by matching the definition's `model.context[].connection` placeholder (`<bindings.<CONNECTOR_KEY> connection>`) against a top-level `bindings[]` entry **by `name`, name-only**. So the binding `name` must be the `<CONNECTOR_KEY> connection` convention.

Two docs instructed the **IS connection display name** (e.g. an email) instead:

- `connector/impl.md:557` — table row claimed binding `name` = the IS connection display name (`chandu.lella@uipath.com #3`). Self-contradictory (same sentence then says it must match the `<CONNECTOR_KEY> connection` placeholder).
- `connector/data-fabric/impl.md` — worse: an explicit rule (Step 3) *and* all 5 definition templates hardcoded `<bindings.<IS connection Name>>`. Following the doc literally produced a binding named after the connection's owner email + a rewritten definition placeholder → resolver finds no match → 102010.

## Fix

- `connector/impl.md`: correct the table row — binding `name` = `"<CONNECTOR_KEY> connection"`, matches the definition placeholder, name-only. Added the exact 102010 failure signature as a tripwire.
- `connector/data-fabric/impl.md`: correct the Step 3 rule and all 5 definition-template `connection` placeholders to `<bindings.uipath-uipath-dataservice connection>`. Clarified the display name belongs only in `bindings_v2.json`'s `displayName` and the connection resource filename.

## Validation

Reproduced on a failing devcon `BillingDisputeResolution` build (crmLookup faulted `102010`). Patched the built flow to match this fix (binding `name` → convention, reverted the definition placeholder) and re-ran `uip maestro flow debug`:

- `102010` gone, no raw `<bindings.…>` placeholder
- `crmLookup`, `erpLookup`, `merge`, `discrepancyDetection` all **Completed** (previously faulted at `crmLookup`)

No dependency on #1942 (disjoint files).

### DIFF
--- a/skills/uipath-maestro-flow/references/author/references/plugins/connector/data-fabric/impl.md
+++ b/skills/uipath-maestro-flow/references/author/references/plugins/connector/data-fabric/impl.md
@@ -27,7 +27,7 @@ Use the exact CamelCase `Name` (e.g. `BankDetails`). For Create/Update, also run
 
 Author the standard top-level `bindings[]` pair (`ConnectionId` + `FolderKey`) per [parent impl.md § Top-level `bindings[]` shape](../impl.md#top-level-bindings-shape-cli-emitted-reference-only). Data Fabric specifics:
 
-- Connection binding `name` = the IS connection display name (not the `<CONNECTOR_KEY> connection` placeholder), since `node configure` is run with this value pre-resolved.
+- Connection binding `name` = `uipath-uipath-dataservice connection` (the `<CONNECTOR_KEY> connection` convention) — must match the definition's `connection` placeholder in the templates below. NOT the IS connection display name / email: a display-name value silently faults at debug with `102010 'Connection' has an invalid GUID value: '<bindings.uipath-uipath-dataservice connection>'`. (The display name belongs only in `bindings_v2.json`'s `displayName` and the connection resource filename.)
 - Both bindings share the same `resourceKey` = `<connectionId>`.
 
 Add one `definitions[]` entry per activity type used — copy the complete entry from the Definitions Templates section below. The `form` block is required ([parent impl.md § Critical: Connector Definition Must Include `form`](../impl.md#critical-connector-definition-must-include-form)).
@@ -199,7 +199,7 @@ Add one entry per activity type used to `definitions[]` in the `.flow` file (sib
       { "name": "operation",    "type": "string" },
       { "name": "objectName",   "type": "string", "value": "QueryEntityRecordsCurated" },
       { "name": "method",       "type": "string", "value": "POST" },
-      { "name": "connection",   "type": "string", "value": "<bindings.<IS connection Name>>" },
+      { "name": "connection",   "type": "string", "value": "<bindings.uipath-uipath-dataservice connection>" },
       { "name": "folderKey",    "type": "string", "value": "<bindings.FolderKey>" },
       { "name": "activityConfigurationVersion", "type": "string", "value": "v1" },
       { "name": "metadata", "type": "json", "body": { "telemetryData": { "connectorKey": "uipath-uipath-dataservice", "connectorName": "Query Entity Records" }, "inputMetadata": {}, "errorState": { "hasError": false } } }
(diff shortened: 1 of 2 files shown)


jiyangzh:
## Converter proof (why the display-name binding can never resolve)

The fix is not just empirical — the CLI's flow→BPMN resolver enforces name-only matching against the `<connector-key> connection` placeholder in source.

`@uipath/flow-schema` — `src/packaging/resolve-context.ts` (`resolveContextBindingPlaceholders`):

```js
var CONTEXT_BINDING_RE = /^<bindings\.(.+)>$/;
const candidates = nodeBindingResource ? bindings.filter((b) => b.resource === nodeBindingResource) : bindings;
const name = match[1];                                  // inner text of <bindings.NAME>
const hasScope = typeof nodeResourceKey === "string" && nodeResourceKey.length > 0;
const binding = hasScope
  ? candidates.find((b) => b.name === name && b.resourceKey === nodeResourceKey)  // resource nodes: (name, resourceKey)
  : candidates.find((b) => b.name === name);                                       // connectors: name-only
if (!binding) {
  console.warn(`[flow-schema] Unresolved context binding placeholder "${entry.value}" ...`);
  return entry;                                          // leaves the raw <bindings.…> string
}
return { ...entry, value: `=bindings.${binding.id}` };
```

Mapping to this PR:

1. **Match key = placeholder inner text.** `name = match[1]` (between `<bindings.` and `>`). The connector definition ships `<bindings.uipath-uipath-dataservice connection>` (registry-owned), so the required binding `name` is `uipath-uipath-dataservice connection`.
2. **Lookup is `b.name === name`.** A binding named `aditi.goyal@uipath.com` never equals `uipath-uipath-dataservice connection` — the old "IS connection display name" guidance is unmatchable by construction.
3. **Name-only for connectors.** No `nodeResourceKey` on connector nodes → `hasScope=false` → `candidates.find(b => b.name === name)`, no resourceKey scoping. (`candidates` pre-filtered by `b.resource === nodeBindingResource` = `"Connection"` — why the `resource` casing also matters.)
4. **Miss → silent raw passthrough.** `console.warn("[flow-schema] Unresolved context binding placeholder …")` (the exact line seen in the failing debug run) + `return entry` unchanged → raw `<bindings.…>` reaches Integration Services → `102010 'Connection' has an invalid GUID value`.

So the display-name binding is guaranteed to leak the placeholder to runtime; only `name = "<connector-key> connection"` resolves. Corroborates the empirical repro (patch name → convention ⇒ crmLookup/erpLookup Completed).

skills/uipath-maestro-flow/references/author/references/plugins/connector/data-fabric/impl.md
```diff
@@ -27,7 +27,7 @@ Use the exact CamelCase `Name` (e.g. `BankDetails`). For Create/Update, also run
 
 Author the standard top-level `bindings[]` pair (`ConnectionId` + `FolderKey`) per [parent impl.md § Top-level `bindings[]` shape](../impl.md#top-level-bindings-shape-cli-emitted-reference-only). Data Fabric specifics:
 
-- Connection binding `name` = the IS connection display name (not the `<CONNECTOR_KEY> connection` placeholder), since `node configure` is run with this value pre-resolved.
+- Connection binding `name` = `uipath-uipath-dataservice connection` (the `<CONNECTOR_KEY> connection` convention) — must match the definition's `connection` placeholder in the templates below. NOT the IS connection display name / email: a display-name value silently faults at debug with `102010 'Connection' has an invalid GUID value: '<bindings.uipath-uipath-dataservice connection>'`. (The display name belongs only in `bindings_v2.json`'s `displayName` and the connection resource filename.)
```
rahul-katikineni, line 30: Any reason we need a separate skill for data fabric. Can we use the is skill here instead which probably already has this handling? 
Would also like to understand why this skill is under flow? Shouldn't this be on IS so that it is available cross surface.

@baishalighosh @chandusailella for thoughts

baishalighosh (changes requested):
Should be centralized via the is platform references, and is not needed to be done via
1. Connector 
2. Via surface (ie not flow specific) - needs to be refactored

skills/uipath-maestro-flow/references/author/references/plugins/connector/data-fabric/impl.md
AditiGoyalUipath: This was added earlier when we introduced a few workarounds to address issues with the Data Fabric connector.
We already have another PR that removes the Data Fabric Skills. We can merge that PR first, but we were waiting for a few remaining issues that @mrstark14 is addressing in this sprint.

baishalighosh:
We are investing in fixes for this in the upcoming sprint staring next week, we should not merge in this state, the skills need to be refactored to centralize binding emission and that will wholistically address these class of issues.

rockymadden (approved):

jiyangzh:
@baishalighosh assume this will be fixed by IS, closing this one.

PR #1977, opened 2026-07-10
fix(governance): clean up test-created policies on shared tenant

NishankSiddharth:
## Problem

The codereval tenant had accumulated ~155 governance policies (`iso-42001-*`, "Block ChatGPT") created by governance skill tests and never cleaned up. A leftover **Block ChatGPT** AI Trust Layer policy was blocking model calls and breaking maestro flow/case tests.

## Root cause

The compliance-pack and AOps tasks were authored assuming the sandbox has no live auth — their prompts say *"uip CLI is not connected to a live tenant — auth errors expected"* — and on that basis carried **no cleanup hooks**. That assumption is false in CI: `tests/experiments/smoke.yaml` and `nightly.yaml` pass `UIPATH_CLI_*` env auth into **every** sandbox and mount `~/.uipath`, and `smoke-skills.yml` mints a real service-account token. So the "dry" governance tasks actually mutate the live tenant on every PR and nightly run — `aops-policy create`, `compliance-packs state enable` — with nothing deleting the result. Epoch/date suffixes on the leftover names are agents dodging collisions with prior debris, so the pile compounded itself.

## Fix — cleanup scoped to exactly what each test creates

**`cleanup_compliance_pack.py`** (compliance-pack scope only): disables the `iso-42001-2023` pack state and deletes AOps policies in the `iso-42001` namespace — the deterministic `iso-42001-2023-<clause>-<product>` names the partial-apply flow creates. Scoped by that prefix so it never touches human-named production policies. Best-effort, always exits 0, degrades to warnings without auth (local runs unaffected). Logs every match/delete and **re-lists the tenant at the end** so the CI log confirms the cleanup.

**Per-task wiring:**
- Compliance-pack tasks call `cleanup_compliance_pack.py` — the creators/enablers (`apply_commands`, `org_scope`, `partial_apply`, `partial_apply_aitl`, `interactive_values`, `full_apply_e2e_live`) pre + post; the read-only guards (`already_configured`, `check_routing`, `disable`, `disambiguation`, `gap_scan`, `query_clause`, `state_list`) post-only, as a safety net against a live-auth agent failure.
- `nonregression_branch_a` creates a **named AOps policy**, not a compliance-pack policy, so it uses the surgical `cleanup_policy.py` keyed to the name pinned in its prompt (`test-nonreg-block-chatgpt`).
- `classify_ambiguous` gets **no hook** — correct behavior creates nothing, and a failure-created policy has an agent-chosen (non-deterministic) name; the failing "must not create" criterion is the signal.
- `full_apply_e2e_live`'s old `cleanup_compliance_pack.py` only disabled pack state and never deleted policies — replaced by the enhanced version.

## Self-review notes

- **Pagination bug found & fixed.** `uip gov aops-policy list --offset` is a **page index**, not a row offset (verified live: `--offset 1` returns rows 21-40). The pre-existing `cleanup_policy.py` did `offset += 20`, so it only ever saw the first 20 policies — part of why debris survived where cleanup did exist. Fixed in both `cleanup_compliance_pack.py` and `cleanup_policy.py`, with a dedupe-by-id guard. (`access-policy list --offset` is row-based — documented the per-kind difference.)
- **Branch-A name determinism.** The skill lets the agent derive the AOps policy name from intent, so a name-substring match wasn't guaranteed. Pinned the name in `nonregression_branch_a`'s prompt — consistent with the fixture-id convention `aops-policy_lifecycle_e2e` already uses — so its cleanup is exact.
- **Undeployed debris.** Accumulated policies show empty Deployments columns, so a plain delete clears them. Documented the delete-blocked-by-assignment edge case in the script for any future deployed-policy flow.
- **Accepted residual.** Failure-only guard tests (`disambiguation`, `classify_ambiguous`) can't be cleaned deterministically (unpredictable name; sweep is aops-policy only). Bounded to test-failure paths, which are themselves a loud signal — not worth over-broadening patterns and risking deletion of legitimate policies.

## Validation

- One-off tenant-sweep task ran in CI and purged the shared tenant (then removed from this branch — never lands in main).
- All 39 governance task YAMLs parse; both cleanup scripts compile; all hook timeouts within coder_eval's 300s cap.
- Read-only validation of `cleanup_compliance_pack.py` against the live tenant: correct pagination across all 77 policies, 0 unintended matches in the `iso-42001` namespace, read-only pack-state check works.


NishankSiddharth:
Latest test runs for updated tests. Each of them shows there are no stale policies anymore
1. https://github.com/UiPath/skills/actions/runs/29083406601
2. https://github.com/UiPath/skills/actions/runs/29085012830

bai-uipath (approved):
**Approve for its stated goal** — it stops the `iso-42001` debris accumulation and the leftover AITL policy that was breaking maestro tests, order-independently. The pagination fix is the real win: `aops-policy list --offset` is a page index, not a row offset, so the old `offset += 20` only ever saw the first 20 policies — that alone defeated the cleanup that already existed. Scoping is disciplined and the cross-validation in `check_full_apply_report.py` is a genuine integrity upgrade.

But it is not parallel-safe, and CI runs these tasks at `-j 10` (smoke) / ~20 (nightly) against one shared tenant. Two risks worth flagging (neither blocks the debris fix):

- **Tenant-global pack-state race.** `already_configured_smoke` (grades `command_not_executed: state enable`) and `apply_commands_smoke` (grades `command_executed: state enable`) both branch on whether the `iso-42001-2023` pack looks configured at coverage-read time. Every mutating task's pre+post cleanup now disables that pack, so a concurrent sibling toggling it steers the agent down the wrong branch → false fail. The coupling is pre-existing, but adding a pack-disable to every hook multiplies the toggle frequency into a likely hit.
- **In-flight "Block ChatGPT" window.** `nonregression_branch_a` creates a live model-blocking AITL policy; cleanup deletes it, but during the agent's run window it is active on the shared tenant and can block a concurrent OpenAI-provider model call — the exact original outage class, now bounded to a window rather than permanent.

jianjunwang2 (approved):

PR #1981, opened 2026-07-10
fix(test): raise e2e_01 task_timeout to 2400s

dushyant-uipath:
`e2e_01` mentions SharePoint and SAP in the prompt. The agent does real connector registry lookups for both, which burns most of the 1200s nightly default before the HITL node is ever placed. The flow produced has no HITL node and no edges — score 0.0, timeout.

Root cause: `turn_timeout: 2400` was already set (per-turn) but `task_timeout` was missing, so the 1200s nightly default applied to the whole task.

Fix: add `task_timeout: 2400` to match the per-turn limit.


dushyant-uipath:
Closing: superseded. e2e_01's timeout and scenario were rewritten multiple times since (#3196, #3180); current state is task_timeout/turn_timeout 1200/1200 with a scaled-down, mocked scenario.

PR #1990, opened 2026-07-10
fix(uipath-rpa): stale UIA doc corrections

danmunteanu1:
## What

Doc corrections surfaced by live end-to-end UIA benchmarks. The earlier fast-path overlay guide was removed from this branch — capture-flow optimization now lives package-side (UiPath/Activities#38449, merged); this PR keeps only the caller-side fixes. No frontmatter changes; Critical Rules preserved (additions only extend Rule 7a).

**Corrections (verified against package 26.10 docs + live runs):**
- `NGetText` output member is `TextString`, not `Text` — fixed in SKILL.md Rule 21 skip-tax, xaml/workflow-guide, xaml/xaml-basics-and-rules.
- Missing activity `Version` (e.g. `NGetText Version=\"V5\"`) passes validate+build, fails at runtime with `ThrowIfNotInTree` — new second cause + diff-against-starter diagnosis in xaml/csharp-expression-pitfalls, terse entry in xaml/common-pitfalls. (Cost a measured 15-minute debug loop when undocumented.)
- Attachment policy: the link path works first-try on cold agent-authored XAML (verified on package 26.10) — uia-configure-target-workflows now says link-first with per-reference embed fallback; the co-versioned package guide is source-of-truth on divergence.
- Stub-pattern acceptance: unconfigured targets are hard build errors, not warnings (ui-automation-guide).
- SKILL.md Rule 7a: runtime-failure trigger — locked/non-interactive Windows session makes every UIA scan fail (needs an unlock, not a fallback); persistent driver failure on an unlocked session routes to the Placeholder-Selector Stub Pattern.

## Checklist (existing-skill modification)

- Frontmatter valid, unchanged; no status markers added
- Critical Rules not removed
- No structural cross-skill dependencies; UIA boundary respected (category pointers only)
- Changes scoped to `skills/uipath-rpa`; all relative links resolve

## Related

- UiPath/Activities#38449 (merged) — package-side capture-flow batching + multi-screen guardrails; this PR carries the caller-side corrections from the same benchmark effort.


skills/uipath-rpa/references/xaml/workflow-guide.md
```diff
@@ -47,7 +47,7 @@ Analyze:
...
+**This is the most important discovery step. Read `<Activity>.md` BEFORE `activities get-default-xaml`, every time, even for activities that look simple.** Installed activity packages ship structured markdown at `{projectRoot}/.local/docs/packages/{PackageId}/activities/<Activity>.md`. The doc is the property surface; the CLI starter is not. `activities get-default-xaml` strips every property at its type default — for `NGetText` that means **all** output properties are absent from the starter, and authoring from the starter produces `NGetText.Value="..."` instead of the real output member `NGetText.TextString`. `validate` does not catch the wrong member name; only `build` does, after a wasted round-trip.
```
danmunteanu1, line 50: @gabrielavaduva is this a C# vs VBasic difference?

danmunteanu1: nvm, did some digging:

- Value → wrong: correct. It genuinely doesn't exist; build rejects it. The docs using Value as the unknown-member example is sound.
- Text vs TextString: both are valid, functional output properties. Text is the legacy backwards-compat one (hidden in the modern designer); TextString is the current typed one Studio surfaces in 26.x. So my four edits (Text→TextString) are a legitimate modernization — they match what current Studio shows and emits — but not a bug fix: the original Text examples would have compiled and run fine. I overstated them as broken.

danmunteanu1:
@alexandru-petrescu, take a look at the changes for the NGetText activity output argument.

alexandru-petrescu (approved):

AlvinStanescu (approved):

PR #1993, opened 2026-07-13
[DS-8893] feat(data-fabric): add MULTILINE_MAX field type, marker read contract, filter/sort exclusion, flag gating

aayushuipath:
## Why

`MULTILINE_MAX` shipped in `@uipath/uipath-typescript` 1.5.2 ([uipath-typescript#549](https://github.com/UiPath/uipath-typescript/pull/549)) and is bundled by `@uipath/data-fabric-tool` 1.198.0 — but the data-fabric skill's supported-types table explicitly forbids types not listed, so agents asked to store >10k-char text silently fall back to `MULTILINE_TEXT` (10k cap) and truncate. The type also has a read contract agents must know: `records list` / `records query` return a size marker (`"HasValue=true Length=N"`) instead of content, and echoing a listed record back through `records update` would overwrite the real content with the literal marker string.

## What

**`uipath-platform` → Data Fabric references**
- `entity-schema.md` — `MULTILINE_MAX` in the supported-types table; three-way intent mapping for `text / long text / paragraph / document body`; `lengthLimit` constraint (1–131072, UTF-16 byte budget); new *MULTILINE_MAX fields* section: tenant `MultilineMax` feature-flag gating (report enablement gap, never silently substitute), not filterable/sortable, marker read contract
- `records-query.md` — *Marker vs Full Content* section: full value only via `records get`; never persist the marker; never write it back (omit MULTILINE_MAX keys from update bodies)
- `filter-platform-contract.md` — MULTILINE_MAX sits outside the operator matrix entirely (the Text / Multiline column does not cover it); client-side scan only with explicit user approval
- `data-fabric.md` — Rule 21 (consolidated contract), tool floor `@uipath/data-fabric-tool 1.198.0+`, three troubleshooting rows (marker in list = expected; 400 on create = flag off; 400 on filter/sort = by design)

**`uipath-coded-apps` → `references/sdk/data-fabric.md`**
- enum comment + anti-shape 10: same marker / write-back / no-filter contract for direct SDK consumers (`getAllRecords` / `queryRecordsById` vs `getRecordById`)

**Tests**
- `tests/tasks/uipath-platform/data-fabric/integration_multiline_max.yaml` — lifecycle: create entity with MULTILINE_MAX → insert ~20k chars → observe marker on `records list` → full content via `records get` → filter attempt surfaced verbatim, no retry. Behavior-graded (server-side record-count check + llm_judge); tags follow sibling convention.

## Test run

The task requires the `MultilineMax` feature flag on the test tenant — first run is tracked in [DS-8895](https://uipath.atlassian.net/browse/DS-8895) (e2e validation of the agent flow against a flag-enabled tenant). Facts in the docs are sourced from the SDK constraint spec / JSDoc and the SDK's integration tests (marker regex `^HasValue=true`), not guessed.

Jira: [DS-8893](https://uipath.atlassian.net/browse/DS-8893)

[DS-8895]: https://uipath.atlassian.net/browse/DS-8895?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aayushuipath:
Pushed `8217a5a7` — doc facts now e2e-verified against alpha (tenant `dfe2etests`, flag on, tool `1.198.0-preview.80`): live marker form quoted (it carries a hint suffix), capacity settled (lengthLimit = UTF-16 **byte** budget; 65,536 chars pass, 65,537 rejected), filter/sort rejections quoted as the real server messages, write-back trap confirmed live, and the tool-floor row now notes 1.198.0 is only on the `preview` dist-tag (npm latest is still 1.197.0, whose bundle predates the type). Full run log on DS-8895.

skills/uipath-platform/references/data-fabric/data-fabric.md
```diff
@@ -339,6 +342,10 @@ Pass the query body via `--body` or `--file`; pagination uses `--limit` / `--cur
...
 | `--folder-key and --include-folders are mutually exclusive` | Both flags passed on `entities list` / `choice-sets list` | Pick one: `--folder-key <key>` for a single folder, OR `--include-folders` for tenant + every folder you can see |
 | Entity / choice set just created via `--folder-key <X>` doesn't appear in `entities list` / `choice-sets list` | Lists default to tenant-only | Re-run with `--folder-key <X>` (same key) or `--include-folders` |
+| `MULTILINE_MAX` field shows `HasValue=true Length=N — call Get Entity Record By Id …` in `records list` / `query` | Expected — list/query return a size marker, not content (Rule 21) | Full value via `records get <entity-id> <record-id>`. Never persist or write back the marker |
+| `entities create` / `addFields` with `MULTILINE_MAX` returns 400 on this tenant | `MultilineMax` feature flag not enabled on the tenant | Surface the enablement gap to the user; don't retry or substitute `MULTILINE_TEXT` without approval (Rule 18) |
```
Raina451, line 346: > feature flag not enabled on the tenant

we dont need to mention about ffs in skills.

Raina451: is this feature still behind ff and not enabled for all orgs in prod?

skills/uipath-platform/references/data-fabric/entity-schema.md
```diff
@@ -74,6 +75,24 @@ CLI needs UPPERCASE enum. Users write mixed-case + synonyms. Two paths:
 
 If the CLI rejects a `--body` with *"Cannot read properties of undefined (reading 'sqlTypeName')"*, the `type` value didn't match a known enum — almost always a casing issue. Re-emit with the exact UPPERCASE value from the table above.
 
+### MULTILINE_MAX fields
```
Raina451, line 78: do we really need this specific section? it is not a something common

Raina451 (approved):

skills/uipath-platform/references/data-fabric/data-fabric.md
aayushuipath: Legit point — no other skill documents an internal platform flag by name. Since #1993 merged before this round, fixed in follow-up PR #2086: flag terminology replaced with tenant-availability language ("not available on every tenant yet — preview rollout"), keeping the behavioral guidance so agents surface the 400 instead of retrying or silently substituting MULTILINE_TEXT.

aayushuipath: Yes — still gated: private preview, per-tenant enablement, default off in prod; enabled on the alpha test tenants (verified live on dfe2etests during the DS-8895 e2e). GA rollout is a later phase of the DS-8249 epic — which is why the skill needs the "may not be available on this tenant" failure-mode guidance at all. Wording de-FF'd in #2086.

skills/uipath-platform/references/data-fabric/entity-schema.md
aayushuipath: Kept it, for consistency with this file's own structure: Choice Set Fields, Relationship Fields, and FILE Fields each have a dedicated section here, because each has extras/contract a caller can't infer from the type table. MULTILINE_MAX is the same class — no filter/sort, size-marker reads, byte-budget lengthLimit. It also got slimmer in #2086 (the flag bullet became one availability line). Happy to fold it into the type-table row if you still prefer after that trim.

PR #1995, opened 2026-07-13
fix(api-workflow): grade tests against init's structured project path

rares-baesu-uipath:
## Problem

Skill PR #1744 made `uip api-workflow init <name>` mandatory — every API workflow now lands at `<Project>/Workflow.json`, never a bare `./Workflow.json` (a lone file is the `invalid_project_folder` shape Studio Web rejects — the Woolworths RCA).

But **18 tasks still graded the root `./Workflow.json`**. When the agent followed the skill and scaffolded a project, `file_exists` / `run` / `validate` all failed with FileNotFoundError — even when the workflow was correct. This surfaced as the `skill-api-workflow-loop-aggregation` failure (agent produced a valid `LoopAggregation/SumNumbers/Workflow.json`, scored 0.0).

It was a **coin-flip**, confirmed across two real runs:

| Run | Agent ran `init`? | File landed at | Root grading |
|-----|-------------------|----------------|--------------|
| loop-aggregation | yes (solution) | `LoopAggregation/SumNumbers/Workflow.json` | ❌ 0.0 |
| edit-add-activity | no (bare file) | `./Workflow.json` | ✅ 1.0 |

Same skill, same model — pass/fail decided purely by whether the agent obeyed the skill (project) or the literal prompt (bare file).

## Fix — structure-first: drive *and* grade the compliant shape

- **Tests (18):** each names a project and grades `<Name>/Workflow.json`, matching the already-correct `outlook_send`. Repointed run/validate/json/exists/contains criteria. 11 active + 7 `skip:true` testmanager tasks.
- **Skill:** rule 19a **"Which mode"** (default = `init` inside a solution; `--skip-solution-registration` only for CLI-only/local) + a new anti-pattern forbidding a lone `Workflow.json` with no project files. This closes the ambiguity that caused the coin-flip at the source.

**Discovery-based grading (`find … Workflow.json`) was considered and rejected:** it would *reward* the non-compliant bare-file output the skill exists to prevent.

## Verification

- `/lint-task`: **18 tasks OK** (path-pinning is the rubric's "ground-truth anchor" carve-out; no new issues)
- CLI-verb reachability: clean · YAML: valid · `check-skill-status.py`: OK · description gate: pass · SKILL.md links: resolve

## Passing-run claim

Ran the repointed tasks locally against a local build of the CLI (outside the coder-eval sandbox) and confirmed the structured paths work end to end: `uip api-workflow init <name>` scaffolds `<Name>/Workflow.json`, and `validate` / `run` against that path succeed — e.g. `skill-api-workflow-if-else-grade` passed (workflow authored at `GradeCheck/Workflow.json`, both `score=75 → PASS` and `score=30 → FAIL` runs green). The `<Name>/Workflow.json` grading paths match what `init` actually produces.

## Caveat

The confirming run above was **local (non coder-eval environment)** — we can't execute the coder-eval sandbox (needs auth) locally, so this isn't a sandbox smoke result. The named-project + explicit path is a strong anchor (matches `outlook_send`) but not an absolute guarantee against an agent adding a solution wrapper (`Solution/X/Workflow.json`). If a sandbox smoke run shows that drift, the deterministic fallback is pinning both names (`S/X/Workflow.json`, the maestro-flow pattern).

### DIFF
--- a/skills/uipath-api-workflow/SKILL.md
+++ b/skills/uipath-api-workflow/SKILL.md
@@ -82,6 +82,8 @@ Do NOT use for: `.flow` Maestro flows (→ `uipath-maestro-flow`), `.xaml` / cod
     ```
     It scaffolds `project.uiproj` + `Workflow.json` + `entry-points.json` + `bindings_v2.json` and, when run inside a solution, **auto-registers the project in the surrounding `.uipx`** (correct `ProjectRelativePath` + a fresh `Id`). Success → `Code: "ApiWorkflowInit"`. Then edit `Workflow.json` only.
 
+    **Which mode.** Default = `init` inside a solution (Studio Web-editable + deployable — what a shipped automation needs). Use `--skip-solution-registration` ONLY when the user explicitly wants a CLI-only/local workflow that never opens in Studio Web or ships in a solution; it still emits the full project folder (`<name>/Workflow.json` + siblings), just no `.uipx` wiring. Never emit a lone `Workflow.json` with no project files — even a throwaway local workflow gets a project.
+
     **Why it matters:** a legacy `project.json` + `workflows/WF_*.json` layout (no `.uiproj`) passes every runtime gate — `validate`, `run`, `pack`, `publish`, deploy — but Studio Web rejects it as `invalid_project_folder` and never shows it. `init` is the one step that can't produce the wrong shape. Full layout + field rules: [references/workflow-file-format.md](references/workflow-file-format.md#project-structure-studio-web-editable-contract).
 
     To **convert a legacy `project.json` project**, `init` a fresh sibling and move the existing workflow content into its `Workflow.json` (cleanest), or convert in place — see [references/troubleshooting.md](references/troubleshooting.md). Never wire it with `uip solution project add/remove` (errors on an already-registered name; `remove`+`add` destroys the project `Id`).
@@ -269,6 +271,7 @@ The mistakes an agent makes most often (each maps to a Critical Rule above — s
 - **Do NOT** read workflow inputs as `$input.<name>` from a non-first activity — use `$workflow.input.<name>`. See rule 13.
 - **Do NOT** invoke `uip api-workflow run` autonomously, and never with auth without an explicit "yes" — vendor calls have irreversible side effects (emails sent, tickets created). See rules 20–21.
 - **Do NOT** hand-assemble a project (`project.json` + `main.json`/`workflows/WF_*.json`). Scaffold with `uip api-workflow init <name>` — it writes the correct `project.uiproj` shape and registers it in the `.uipx`. The legacy `project.json`-only shape runs and packs but Studio Web rejects it (`invalid_project_folder`) and never shows it. See rules 19–19a.
+- **Do NOT** emit a lone `Workflow.json` with no project files, even for a quick local run. It runs under `uip api-workflow run` but is not a Studio Web project — can't be edited or shipped. Every workflow lives in an `init`-scaffolded project (`--skip-solution-registration` when no solution is needed). See rule 19a ("Which mode").
 - **Do NOT** wire a project into the solution with `uip solution project add/remove` — it errors on an already-registered name, and `remove`+`add` destroys the project `Id`. `init` registers it; for an already-built project, edit the `.uipx` `ProjectRelativePath` in place. See rule 19a.
 - **Do NOT** trust "it packed / published / ran" as proof a project opens in Studio Web — every runtime gate passes on the wrong shape. Scaffolding with `init` is what guarantees it (rule 19a).
 
--- a/tests/tasks/uipath-api-workflow/connector-intsvc/outlook_send.yaml
+++ b/tests/tasks/uipath-api-workflow/connector-intsvc/outlook_send.yaml
@@ -22,9 +22,11 @@ initial_prompt: |
   Do NOT ask for approval, confirmation, or feedback.
 
 success_criteria:
-  - type: file_exists
-    description: "Workflow.json was created in the project folder"
-    path: "OutlookSend/Workflow.json"
+  - type: run_command
+    description: "A Workflow.json exists inside a project folder (not a bare root file)"
+    command: 'WF=$(find . -name Workflow.json -not -path "./Workflow.json" -not -path "*/node_modules/*" | head -1); test -n "$WF"'
+    timeout: 10
+    expected_exit_code: 0
     weight: 1.0
     pass_threshold: 1.0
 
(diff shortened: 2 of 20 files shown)


rares-baesu-uipath:
Verified locally (against a local CLI build, **not** the coder-eval sandbox — that needs auth we can't run locally): `init` scaffolds `<Name>/Workflow.json` and `validate`/`run` against that path pass, e.g. `skill-api-workflow-if-else-grade` green on both score branches. Sandbox smoke run remains the post-merge confirmation.

liviubobocu (commented):
previous commands have been fixed

there is still an issue with the following fact:

esentially command_executed is not really testing the agent, but passes or fails based on the clI and it will incorrectly grade the agent itself 

Why run_execute is weak — with the real numbers

The seed (seed_valid_workflow.py) drops a already-valid Workflow.json. Now trace a hypothetical agent that does literally nothing:

- file_exists: Workflow.json (w=1.0) → passes — the seed made it
- command_executed: uip api-workflow run (w=2.0) → fails — agent ran nothing
- run_command: score 85 → PASS (w=1.5) → passes — the grader runs it
- run_command: score 40 → FAIL (w=1.5) → passes — the grader runs it

Score for a no-op agent = (1.0 + 1.5 + 1.5) / 6.0 = 0.67. The agent's actual behavior — the one thing an operate task exists to measure — only swings the score between 0.67 and 1.0. And the prompt tells the agent to "report the grade each run returns," but no criterion reads that report, so that instruction is ungraded.

(The two diagnose tasks don't have this problem: their seed is broken, so the run_command checks genuinely fail until the agent fixes the workflow. There, re-running in the grader is a valid signal. It's only broken for the valid-seed operate task.)

The concrete fix

Make the behavioral credit depend on an agent-produced artifact. Have the agent write what the runs returned to a file, then grade that file with json_check. Now the only path to the points is: the agent actually ran the workflow and captured the output.

Replace run_execute.yaml's prompt tail + success_criteria with:

initial_prompt: |
  The API workflow in ./Workflow.json is a score->grade evaluator: it takes a
  numeric `score` input and returns a `grade` of "PASS" (score >= 60) or "FAIL".
  It is already valid — do not rewrite it.

  Execute it locally for score 85 and again for score 40. Capture the grade each
  run returns and write them to ./results.json exactly as:
    {"grade_85": "<grade for score 85>", "grade_40": "<grade for score 40>"}

  Do NOT ask for approval or confirmation.

success_criteria:
  - type: command_executed
    description: "Agent executed the workflow with the local runner (the operate action)"
    tool_name: "Bash"
    command_pattern: '(uip|\$UIP)\s+api-workflow\s+run'
    min_count: 1
    weight: 2.0
    pass_threshold: 1.0

  - type: file_exists
    description: "Agent produced the results file (proves it captured run output)"
    path: "results.json"
    weight: 1.0
    pass_threshold: 1.0

  - type: json_check
    description: "Captured the real run outputs: score 85 -> PASS, score 40 -> FAIL"
    path: "results.json"
    assertions:
      - expression: "grade_85"
        operator: regex
        expected: "(?i)^pass$"
      - expression: "grade_40"
        operator: regex
        expected: "(?i)^fail$"
    weight: 3.0
    pass_threshold: 1.0

  ### Advisory sanity anchor: the seed itself must run correctly. Weight 0 so it
  #### never awards score for grader-side execution the agent didn't perform.
  - type: run_command
    description: "Sanity: seeded workflow routes correctly (not an agent signal)"
    command: 'uip api-workflow run ./Workflow.json --no-auth --input-arguments ''{"score":85}'' --output json | grep -q ''"PASS"'''
    timeout: 60
    expected_exit_code: 0
    weight: 0
    pass_threshold: 0

liviubobocu:
Run the commands together with skills for this PR with claude and the results highlight some failures in regards positioning of Workflow.json 

## Ran the repointed suite through coder-eval — the PR does not reliably pass (7/11)

I ran all 11 active repointed tasks through `coder-eval` locally (`experiments/default.yaml`, tempdir driver, `claude-code` / sonnet agent, `uip` CLI 1.197.1). **7 passed, 4 failed — and all 4 failures share one root cause: the workflow file lands one level deeper than the pinned `<Name>/Workflow.json`, because the agent creates a solution wrapper.**

### Results

| Task | Pinned path (graded) | Where the agent actually put it | Score |
|---|---|---|---|
| `if_else_grade` | `GradeCheck/Workflow.json` | `GradeCheck/Workflow.json` (root, no solution) | ✅ 1.000 |
| `validate_smoke` | `ReadyStatus/Workflow.json` | root project | ✅ 1.000 |
| `edit_add_activity` | `GreetCount/Workflow.json` | root project | ✅ 1.000 |
| `wait_activity` | `.../Workflow.json` | root project | ✅ 1.000 |
| `nested_loops` | `.../Workflow.json` | root project | ✅ 1.000 |
| `nested_control_flow` | `.../Workflow.json` | root project | ✅ 1.000 |
| `javascript_transform` | `.../Workflow.json` | root project | ✅ 1.000 |
| `loop_aggregation` | `SumNumbers/Workflow.json` | **`MySolution/SumNumbers/Workflow.json`** | ❌ 0.000 |
| `http_public_api` | `CatFact/Workflow.json` | **`CatFactSolution/CatFact/Workflow.json`** | ❌ 0.619 |
| `trycatch_continue` | `HandleError/Workflow.json` | **`HandleError/HandleError/Workflow.json`** | ❌ 0.000 |
| `dowhile_break` | `CountUp/Workflow.json` | **`CountUpSolution/CountUp/Workflow.json`** | ❌ 0.000 |

### Root cause: the two halves of this PR contradict each other

- The **test** half pins the un-nested path `<Name>/Workflow.json`.
- The **skill** half adds rule 19a: **"default = `init` inside a solution."**
- When an agent obeys that new default, it runs `uip solution init X` → `uip api-workflow init <Name>`, and the file lands at `<Solution>/<Name>/Workflow.json` → **every pinned criterion misses → 0.0.**

Confirmed with the CLI directly:

```
# root init → matches the pin
uip api-workflow init GradeCheck            → ./GradeCheck/Workflow.json      ✅

# skill's default (init inside a solution) → one level deeper, misses the pin
uip solution init MySol && cd MySol
uip api-workflow init GradeCheck            → MySol/GradeCheck/Workflow.json  ❌
```

This is the same coin-flip PR #1744 → this PR set out to eliminate, just relocated from "bare-file vs project" to "root-project vs solution-nested." The "`if_else_grade` passed locally" claim in the PR description is simply the lucky root-layout branch of that coin-flip — it masks a ~36% failure rate across the suite.

### Suggested fix: grade the shape, not the exact parent folder

Exact-path pinning is too brittle, because **both** layouts are legitimately compliant (the skill actually *recommends* the solution-nested one). Grade for "a `Workflow.json` inside some project folder" and reject only a bare root file — this passes for both compliant layouts and still forbids the non-compliant bare `./Workflow.json` the PR wants to prevent:

```yaml
# Replace the exact-path file_exists with a shape check that accepts nesting
# but rejects the bare root file:
- type: run_command
  description: "A Workflow.json exists inside a project folder (not a bare root file)"
  command: "test -n \"$(find . -name Workflow.json -not -path './Workflow.json' -not -path '*/node_modules/*' | head -1)\""
  expected_exit_code: 0
  weight: 1.0
  pass_threshold: 1.0

# Resolve the path once and reuse it in the run/validate criteria, e.g.:
- type: run_command
  description: "Sums [10,20,30] to 60 via the loop"
  command: "WF=$(find . -name Workflow.json -not -path './Workflow.json' -not -path '*/node_modules/*' | head -1); uip api-workflow run \"$WF\" --no-auth --input-arguments '{\"numbers\":[10,20,30]}' --output json | grep -qiE '\"total\"[ :]+60'"
  expected_exit_code: 0
  weight: 3.0
  pass_threshold: 1.0
```

> Note: the PR description rejected "discovery grading" on the grounds it would reward the bare-file output. This variant specifically **excludes** `./Workflow.json` (`-not -path './Workflow.json'`), so that objection doesn't apply — it rewards only the project shape, at any nesting depth.

_(Environment note: the `fast-json-patch` `MODULE_NOT_FOUND` lines in run stderr are harmless — the CLI falls back and the runs exit 0.)_

liviubobocu (approved):

PR #2018, opened 2026-07-14
feat(uipath-troubleshoot): playbook for UIAutomationNext dependency/version conflicts

Stefan-Virgil:
## What

Adds a UI Automation troubleshooting playbook for the **dependency / version-conflict** error class that `NClick` / modern Click activities hit when the `UiPath.UIAutomationNext.Activities` assembly fails to **load or bind** — a category no existing `ui-automation` playbook owned.

The package's existing playbooks all cover failures that happen *after* a selector is evaluated (targeting: `ambiguous-selector`, `timeout-issue`, `application-not-found`; execution/input-mode: `element-found-not-actionable`, `disabled-element`, `click-coordinate-off-screen`). This class fails *before* any selector is evaluated, so it was falling through.

## Why

Two distinct signatures, distinct root causes, distinct fixes — the playbook discriminates them on exception type:

| Signature | Symptom | Root cause | Fix |
|-----------|---------|-----------|-----|
| **A** | `MissingMethodException` — "Method not found: 'Void UiPath.UIAutomationNext.Activities…'" | Assembly loaded but a member is absent — version mismatch **inside `project.json`'s dependency set** (a UI Automation package bumped out of step with `UiPath.System.Activities` / the Studio line) | Align **all** foundational packages via Manage Packages |
| **B** | `FileNotFoundException` / `FileLoadException` — "Could not load file or assembly 'UiPath.UIAutomationNext.Activities'" (works in Studio, fails in Assistant/robot) | Exact pinned version can't be resolved — stale robot NuGet cache or a feed missing the version | Clean `%userprofile%\.nuget\packages`, verify feed, republish Lowest Applicable / Strict |

## Changes

- **New** `playbooks/dependency-version-conflict.md` — both signatures + a "Not this playbook" redirect block routing cat-1/cat-3 signals back to the existing selector/execution playbooks.
- `overview.md` — two new assembly-load/bind exception entries.
- `summary.md` — routing row.
- **New** `tests/.../uia-dependency-version-conflict/` — design-time **Signature A** scenario: `project.json` pins `UiPath.UIAutomation.Activities [24.10.9]` vs `UiPath.System.Activities [22.4.4]` on Studio 22.4; process never ran (no jobs), diagnosed from project source. Mirrors the mature `word-replace-text-version-mismatch` scaffold (project-source-driven, simulation persona, lean `skill_triggered + llm_judge` contract).

## Scope

Categories 1 (targeting) and 3 (execution/input-mode) from the source guide are already thoroughly covered by existing `ui-automation` playbooks, so per repo DRY rules they are not duplicated — this PR fills the one genuine gap (Category 2) and cross-links back.

## Validation

- `/lint-task`: **OK** (0 findings).
- coder-eval: **3/3 reps SUCCESS @ 1.0** (`skill_triggered` 1.00, `llm_judge` 1.00 on every replicate).


Stefan-Virgil:
Addressing the two **Issues for Manual Review** flagged by the automated reviewer.

### 1. Domain accuracy of the version pairing

The fixture pins `UiPath.UIAutomation.Activities [24.10.9]` against `UiPath.System.Activities [22.4.4]` on `studioVersion 22.4.3.0` — a UI Automation package bumped well ahead of the foundational System.Activities line and the installed Studio. Causal chain the RESOLUTION encodes:

- The 24.10 UI Automation package was compiled against a newer `UiPath.UIAutomationNext(.Activities)` runtime.
- With System.Activities / Studio held at the 22.4 line, the resolved UIAutomationNext runtime is **older** than the package's callers expect.
- At open/validate, the `N*` activities bind against the older assembly; the assembly **loads** but a member is **absent** → `MissingMethodException: "Method not found: 'Void UiPath.UIAutomationNext.Activities...'"`. (Contrast Signature B, where the assembly can't load *at all* — a different exception family.)

This mirrors the pattern already established by the mature, merged `word-replace-text-version-mismatch` sibling (`UiPath.Word.Activities [2.0.0]` vs `studioVersion 21.10.5.0` — a package ahead of the Studio line). The exact patch numbers are representative, not asserted as the sole real-world trigger.

Two points that make this robust to the exact-version nuance:

- **This is a faithful-replay *mock* — it never launches Studio.** `project.json` is the *evidence the agent must read and reason about*; the symptom (`MissingMethodException`) is supplied by the reporter in `initial_prompt`, exactly as a real user would paste it.
- **The diagnosis and fix are invariant to whether a live Studio surfaces this precise skew as `MissingMethodException` vs a NuGet dependency-resolution error** — both route to the same remediation (align all foundational packages to a mutually compatible line via Manage Packages), which is what the playbook and `RESOLUTION.md` already prescribe. The test grades that the agent (a) reasons from the project's dependency set rather than chasing Orchestrator jobs, and (b) does **not** misapply the Signature B cache-clean/republish fix.

Happy to adjust the specific versions if a domain expert prefers a particular known-incompatible pairing.

### 2. coder-eval validation

- **Local:** 3/3 reps `SUCCESS` @ score `1.000` (`skill_triggered` 1.00, `llm_judge` 1.00 on every replicate); run `runs/2026-07-14_11-37-40`.
- **CI:** e2e scenarios intentionally do **not** run on PR — the smoke gate runs only `smoke`-tagged tasks (per `tests/tasks/uipath-troubleshoot/CLAUDE.md` § Tier tag), and the sole troubleshoot smoke task is `smoke-manifest-commands`. That gate is **green** here, which confirms every `uip` verb this scenario mocks (`or folders list`, `or jobs list`, `docsai ask`) resolves against the current CLI catalog.

### Post-review change

Scrubbed the `manifest.json` `_doc` (commit `fa504580c`) so no agent-visible sandbox surface names the root cause — the `_doc` previously described the diagnosis in prose. Behaviorally inert (the dispatcher matches on rules, not the doc string), so no re-validation was needed.

MarinRzv (approved):

PR #2027, opened 2026-07-14
test(uipath-maestro-flow): add simulated interactive flow tasks

CarlesUIPath:
## Summary

Add 8 persona-driven simulated variants under interactive/ that test the agent's ability to clarify an ambiguous, non-technical request before building — covering weather, dice, customer escalation, expense approval, HITL schema design, IxP invoice extraction, and two Slack flows.

Each variant mirrors its non-simulated original's tags and adds the flat `simulation` marker (also applied to the existing solution_select task). customer_escalation ships a behavioral checker (check_escalation_behavior.py) that grades observable capabilities rather than node shape, since the persona can never name implementation details.

## Considerations

- `cli_dice_roller_simulated.yaml` was already implemented.
- All the simulated tasks now include the tag `simulation`, including the existing `solution_select.yaml`.
- **Important (please review):** Max turns and timeout should be lowered. @bai-uipath 

## EvalBoard visualisation

Simulated tasks will show a conversation history that will be visible in the Evalboard
image


bai-uipath:
**Good concept, needs two fixes before merge: land these as replacements (not additions), and restore `flow debug` on the four scenarios that can actually run.** The persona-driven "withhold requirements until asked" design is genuinely useful and cleanly built. What holds it back: it doubles suite cost by running alongside the originals it mirrors, and it swaps those originals' execution checks for static JSON greps.

## Must fix before merge

- **Replace the originals, don't run both.** All 8 variants are net-new task_ids under `interactive/`, and every scenario's non-simulated original still runs — `multi_node/bellevue_weather`, `multi_node/dice_roller`, `multi_node/slack_channel_description`, `multi_node/slack_weather_pipeline`, `multi_node/customer_escalation`, `e2e/devcon_expense_approval`, plus the HITL/IxP analogs. The suite now pays for both copies of each scenario. Fix: retire the non-simulated original for each scenario that gains a variant — but only after the debug gap below is closed, or you delete real signal. Once a variant runs debug it's a superset of its original (clarify + build + execute), so the swap is safe then.

- **Restore `flow debug` on the debuggable four (bellevue, dice, slack_channel_description, slack_weather_pipeline).** Every variant's success_criteria are 100% static — `uip maestro flow validate` plus Python grepping the `.flow` JSON — and these four regressed from originals that execute the flow and assert its output. `check_weather_flow.py` runs `run_debug()` and asserts the emitted verdict against the live weather; the variant only asserts both `'nice day'` and `'bring a jacket'` appear in the file, which a flow that hardcodes both strings and wires nothing still passes. Same story for dice (original asserts output is an integer in [1,6]; variant just greps for `random`), slack_channel_description, and slack_weather_pipeline. `flow debug` backs ~40 checkers across this same suite, so the environment supports it — this is lost coverage, not a platform limit. Fix: run debug and assert on runtime output for these four.

- **Static-only is correct for the other four — keep it.** customer_escalation (trigger flow, no live mailbox), expense_approval and hitl_schema_design (inline HITL blocks debug), and ixp_invoice (needs live IxP/SharePoint/SAP) genuinely can't execute, and their originals are validate-only too, so no regression. The per-task "validate-only because…" comments are the right call. One note: even here, assert wiring/structure over "both verdict strings present in JSON" — string presence is a weak behavioral proxy.

## max_turns / timeout (you flagged this)

- **Derive the budget from dialog length instead of guessing per task.** `run_limits.max_turns` is 40/60/80/100/140 across tasks whose sim `max_turns` are all 12–18 — near-identical dialog lengths. A clarify-then-build round is a handful of agent turns; ~12–18 rounds × ~4–6 build turns lands around 60–90, not 140. Pick one formula, apply it, and leave ~30% headroom over a clean run rather than 3×.

bai-uipath (approved):
actually I revise my last comment - looking at the results it is different enough from the one-shot ones so it's better to have them as separate tests (as you have in done in this PR)

PR #2028, opened 2026-07-14
feat(activation): adopt coder_eval early-stop in skill-activation suite

mohsen-uipath:
## What

| File | Change |
|------|--------|
| `tests/experiments/activation.yaml` | `run_limits.max_turns: 1 → 3`; add `run_limits.stop_early: true`; document semantics + harness constraint |
| `tests/tasks/activation/activation.yaml` | add `stop_when: auto` to all **23** stacked `skill_triggered` criteria |
| `tests/scripts/activation_gate.py` | gate's generated criterion also sets `stop_when: auto` — required: an armed experiment with no stop criterion is a hard `EarlyStopConfigError` at resolution |
| `tests/.coder-eval-version` | `0.8.8 → 0.8.10` — first release with `stop_when: auto` |
| `tests/README.md`, `tests/tasks/activation/README.md` | document early-stop behavior and the codex/antigravity escape hatch |
| `tests/tasks/uipath-maestro-flow/interactive/*_simulated.yaml` (×7) | drop `simulation.parallel_trials: true` — 0.8.10 removed the never-read field and its schema now rejects it (surfaced by the pin bump; behavior-neutral, trial concurrency was never driven by it) |

## Why

Activation is a recognition decision, not a work loop. The old hard `max_turns: 1`
cap truncated correct skill invocations the model emits *after* a brief reasoning
turn — recording them as false negatives. Raising the cap to 3 fixes that, but
naively tripling turns triples cost.

Arming early-stop on the stacked `skill_triggered` criteria ends a row the moment
it is decided: pass-stop when the expected skill engages, fail-stop when a wrong
skill engages; only no-engagement rows pay the full (small) turn budget. Verdicts
are unchanged — the any-engagement latch is monotonic and the authoritative
`check_all` re-runs on the frozen trajectory — while decided rows cost ~1 turn.

`stop_when: auto` (not `decided`): under the any-engagement latch each criterion
instance is live-decidable in exactly ONE polarity, which depends on the row —
the row's positive criterion (`skill_name == expected_skill`) can only live-pass,
the 22 distractors can only live-fail. `auto` arms each instance's own decidable
polarity; `decided` is rejected for `skill_triggered` because "both polarities"
is never true per-instance.

With a single positive criterion and no distractors, the gate's `auto` arms
pass-stop only — it cannot change `recall.yes`, only cost.

## Harness constraint

Cooperative early-stop is Claude Code-only. `--type codex` / `--type antigravity`
runs MUST add `-D run_limits.stop_early=false` (they then pay the full
`max_turns: 3` per row; verdicts unchanged). Documented in the experiment YAML
and both READMEs.

## Testing

Three independent armed-vs-control A/B runs on real dataset rows (claude-sonnet-5,
all 23 criteria stacked; control = `-D run_limits.stop_early=false`):

- 6-real-row armed run: **6/6 rows correct**; all 4 positives pass-stopped @ turn 1,
  first tool call (~5–7s); negatives ran to the cap as designed.
- Control run on the same rows: **verdict-identical on every row**; 31 assistant
  turns vs 17 armed (~45% reduction; ~45–55% across the three runs).
- Fail-stop path exercised live on an adversarial negative (misfire fail-stopped
  @ turn 1, first tool call); remaining fail-stop shapes covered by coder_eval
  unit tests.
- Gate smoke (`activation_gate.py`, 1 row): pass-stop @ turn 1, gate=PASS, exit 0.
- Released `coder-eval==0.8.10` wheel (what pin-installing CI workflows get):
  `coder-eval plan` validates task + experiment cleanly — "All tasks are valid!",
  exit 0.
- ADO nightly pipeline executed in **dry-run mode** against this branch with
  `coderEvalVersion=pinned`: all stages green, activation suite
  `pass_rate=100.0% (24/24)`, `gate=PASS`, upload clean to the isolated dry-run
  container. Full details in the [dry-run verification comment](https://github.com/UiPath/skills/pull/2028#issuecomment-5075437765).

## Note on metrics

A pass-stop can hide a distractor misfire on a later tool call, and a fail-stop
can hide a later correct invocation — authoritative precision/recall numbers come
from `-D run_limits.stop_early=false` runs (documented in
`tests/tasks/activation/README.md`).

## Verify cheaply (a few samples)

```bash
cd tests            # with SKILLS_REPO_PATH=<repo root> exported
coder-eval run tasks/activation/activation.yaml -e experiments/activation.yaml \
  --sample 5 --run-dir runs/verify -j 1
```

Expected: positive rows report `early_stop.reason = criterion_passed` at
`sdk_turn_index: 1`; negatives (rare in a random draw — sample more to hit one)
run to `max_turns: 3` with no `early_stop`.


bai-uipath (approved):
## Dry-run verification

Ran the ADO nightly pipeline in dry-run against this branch with `coderEvalVersion=pinned` (reads 0.8.4 from the branch's pin): https://dev.azure.com/uipath/coder_eval/_build/results?buildId=12707084

- ✅ **0.8.4 resolves end-to-end** — wheel installed from the branch pin, `coder-eval-agent:0.8.4` image pulled clean.
- ✅ **Positives early-stop on turn 1** — every positive fired `reason=criterion_passed deciding=skill_triggered sdk_turn=1`, scoring 23/23 from the frozen trajectory.
- ✅ **Negative runs to the cap** — `negative-033` reached `max_turns: 3` with no early-stop and scored correctly (the intended asymmetry).
- ✅ **Fail polarity works** — a wrong-skill misfire (`uipath-test-006`) early-stopped on `reason=criterion_failed`.
- ✅ **24 rows, 0 errors, upload clean** to the isolated `runs-dryrun` container.
- ℹ️ The build shows `GATE: FAIL`, but that's an activation-only harness artifact, not this PR: the gate's `activation_cases > 0` check reads a nested activation sub-run that exists only in the merged `skills,activation` layout, and the `review: FAILED` line is explicitly non-gating. The eval itself passed (92%, 0 errors).

mohsen-uipath:
## Dry-run verification

Ran the ADO nightly pipeline in dry-run against this branch with `coderEvalVersion=pinned` (reads **0.8.10** from the branch's pin): https://dev.azure.com/uipath/coder_eval/_build/results?buildId=12799116

- ✅ **0.8.10 resolves end-to-end** — log confirms `coder_eval pinned to 0.8.10 (skills/tests/.coder-eval-version)`, and `ghcr.io/uipath/coder-eval-agent:0.8.10` pulled clean (digest `97068ff2…`).
- ✅ **Positives early-stop on turn 1** — every early-stop event in the log fired `reason=criterion_passed deciding=skill_triggered sdk_turn=1 tool_call=1` (5.6–12.2s), each scoring **23/23** from the frozen trajectory.
- ✅ **Negatives run to the cap** — e.g. `negative-022` finished with no early-stop (19.3s) and scored 23/23 correctly (the intended asymmetry).
- ✅ **24 rows, 0 errors** — activation suite rollup: `pass_rate=100.0% (24/24)`.
- ✅ **GATE: PASS this time** — `suite rollup: variant=default suite=skill-activation … gate=PASS`. The activation-only gate artifact from the previous dry-run did not reproduce in the merged `skills,activation` layout.
- ✅ **Upload clean to the isolated dry-run container** — 338 blobs (0 errors) to `coderevaltests/runs-dryrun/2026-07-24_22-12-12`, `run.json`/`meta.json` confirmed; no Slack ping.
- ℹ️ No wrong-skill misfire occurred in this sample, so the fail polarity wasn't exercised live in this run (it was previously verified locally and by coder_eval's unit tests). All four pipeline stages green, total 53m58s.

uipreliga (approved):
Fix what you agree with and 🚢

uipreliga:
## Code review — multi-model (Gemini 3 + GPT-5.3-Codex + Opus)

Claims below were verified against the coder_eval 0.8.10 source, not taken from the PR description.

**The early-stop mechanics are correct.** `stop_when: auto` is the right choice, and the reasoning about why `decided` is unusable is exactly right (`skill_triggered.live_decidable_polarities` narrows each instance to one polarity). All 23 criteria are armed, per-row `${row.expected_skill}` substitution happens before `validate_early_stop` (`experiment.py:668-673`), and the `parallel_trials` removals are mandatory under the new pin. The gate-script fix in 0c1549d5d is a genuine catch — an armed run with no stop criterion hard-errors, and nothing links the two files at compile time.

The problems are that the PR is framed as a cost optimization while changing what the benchmark measures, and several of the new comments are wrong about the stop rule they document.

### High

**1. `BASELINES_PCT` is no longer comparable to what the gate measures** — `tests/scripts/activation_gate.py:20-25` *(flagged independently by all three reviewers)*

The comment claims baseline and gate "stay directly comparable." The baselines were measured at `max_turns: 1`; the gate resolves `-e tests/experiments/activation.yaml` so it now runs at `max_turns: 3`. Engagement is a monotone set union over the whole trajectory, so recall can only rise — every baseline becomes an easier floor and `DROP_PP = 10` absorbs more real regression than intended. A frontmatter edit that makes a skill fire only after the agent reads two neighbouring `SKILL.md`s scores as a miss at 1 turn and a hit at 3. Either re-baseline, or pin `run_limits: {max_turns: 1}` in the generated gate YAML.

**2. The deleted measurement-validity argument is not rebutted — and browsing now counts as activation** — `tests/experiments/activation.yaml:10-29`

The removed comment argued a >1 turn budget "would let the agent compensate for a weak skill description by exploring." That isn't theoretical: `_engaged_skill_names` (`criteria/skill_triggered.py:44-73`) matches `skills/<name>/` in **any string parameter of any tool call**, not just a `Skill` call. `allowed_tools` includes `Read`/`Grep`/`Glob`/`Bash` and the plugin is mounted at `$SKILLS_REPO_PATH`, so a plain `Read` of `skills/uipath-platform/SKILL.md` pass-stops a positive row. Two extra turns are two extra chances to convert an FN into a scored TP by browsing the tree. That is a substantive change to the metric, and the diff doesn't acknowledge it.

**3. Timeouts not scaled with the turn budget; timed-out rows leave the denominator** — `tests/experiments/activation.yaml:29-32`

`max_turns` triples but `task_timeout`/`turn_timeout` stay at 120s, so turns 2–3 have no wall-clock budget. A `task_timeout` isn't graded: the watchdog cancels `_evaluation_loop` (`orchestrator.py:466-507`) and `check_all` runs *inside* it at `orchestrator.py:1409`, so a TIMEOUT row has empty `success_criteria_results`. `reports.py:779-781` builds each criterion's slice as `[... for row in rows if i < len(...)]` — the row is **dropped from the denominator, not counted as a miss**. The rows that can actually consume 3 turns are the ones that didn't engage a skill, i.e. FN candidates. `activation_gate.py:143-148` reads only `recall.yes` and never checks `completion_rate`, which `_attach_row_accounting` does populate. Cheap fixes: raise `task_timeout`, and/or assert `completion_rate` in the gate.

### Medium

**4. The new prose describes a stop rule that does not exist** — `tests/README.md:181`, `tests/tasks/activation/README.md:39`

Both say the row ends "the moment a skill engages — pass-stop on the expected skill, fail-stop on a misfire." False for positive rows: `early_stop.py` gates fail-stop behind `if not any(verdicts[i] == "undecided" for i in pass_armed)`, so a misfire is *deferred* while the positive criterion is undecided. Consequences the docs get wrong: a positive row that engages only wrong skills runs to the cap (so "only no-engagement rows run to the cap" is false), and such rows report `criterion_failed`, not the promised pass-stop.

**5. "verdicts are unchanged" is false and self-contradictory** — `tests/experiments/activation.yaml:28`

The codex/antigravity note says an unarmed run yields unchanged verdicts. `tests/tasks/activation/README.md:39`, added in the same PR, says the opposite — a pass-stop hides later distractor misfires, so the armed run's precision is systematically optimistic relative to an unarmed one. A codex operator reading the YAML is told the two runs are comparable. (Also: "pay the full `max_turns: 3` on every row" — `max_turns` is a cap, not a floor. Same slip in three "runs to the cap" phrasings.)

**6. "Adding a new skill" instructions omit `stop_when: auto`, and omitting it fails silently** — `tests/tasks/activation/README.md:45`

Step 3 still says to append a criterion with just `skill_name:`. `validate_early_stop` only requires *at least one* armed criterion, so a 24th unarmed criterion raises nothing. For every new-skill positive row, all 23 existing criteria are distractors armed fail and `pass_armed` is empty — nothing to defer for — so the row fail-stops on the first distractor touch. An agent that reads a neighbouring `SKILL.md` before invoking the new skill is cut at the read and scores FN. The new skill's recall is deflated with no warning.

**7. Cost section stale** — `tests/tasks/activation/README.md:47-49`

Worth calibrating: for the **armed** path the numbers roughly survive. At ~95% recall only ~63 recall-miss positives plus 57 negatives can exceed one turn, and `max_turns` is a cap — worst case ≈ +18% turns. The real gap is the **unarmed** path this PR newly prescribes in two places (authoritative precision; every codex/antigravity run), which lifts all 1309 rows to up to 3 turns — plausibly 2–3× — against an unchanged 120s timeout (see #3). The section says nothing about it.

### Low

- **Duplicated, unenforced caveat**: `-D run_limits.stop_early=false` appears in three files and is enforced nowhere. No current CI path hits it — `activation-gate.yml` uses Claude Code, `run-coder-eval.yml` hardcodes `nightly.yaml` — so it's manual-run guidance only. Whether an external runner (dashboard / nightly activation suite) invokes this experiment with `--type codex` is not verifiable from this repo; if one does, it now hard-errors and the only shipped mitigation is prose.
- **Comment volume**: 19 comment lines for 4 config lines, 10 for one YAML key. The load-bearing parts are the codex override and the one-line "why `auto`, not `decided`". The rest re-derives coder_eval internals that will rot — and per #4/#5 is already inaccurate.
- **Stale neighbours**: `activation.yaml:8` "one-shot decision — short timeouts are fine" now sits directly above `max_turns: 3`; `activation-gate.yml:4-6` header says "15pp" and "2026-05-08" against the script's `DROP_PP = 10` and 2026-06-17.
- **No passing-run claim** (`.claude/rules/test-writing.md` §Workflow 5). The history is suggestive: b9eb0f457 / b1e4d9e81 shipped `stop_when: decided`, which is rejected for *every* row, and 0c1549d5d then fixed a gate YAML that would have hard-errored before a single row ran. A `--sample 5` run catches both.
- **Pre-existing count drift** in the paragraphs this PR edits: README says "22 criteria" / "~1190 rows"; actual is 23 and 1309 (1252 positives + 57 negatives). `uipath-insights` is absent from `BASELINES_PCT`, so it silently `SKIP`s the gate.

### Verified as correct (not defects)

The 0.8.8→0.8.10 bump doesn't change offline scoring — `_check_impl` was already any-engagement in 0.8.8; only the *live* verdict policy changed. `stop_when` is inert without `stop_early`, so the 23 new keys are harmless under other experiments. `-D run_limits.stop_early=false` parses and is applied before validation, so the escape hatch works. All 24 jsonl files carry `expected_skill` matching their filename, so the gate's "all positives" claim holds. Negative rows correctly never pass-stop (vacuous `pass_armed` guard). And the gate comment's "early stop cannot change `recall.yes` — only cost" is **true** as written.

### Overall

Ship-blocking issues are #1 and #3, both cheap: re-baseline (or pin the gate to `max_turns: 1`), and raise `task_timeout` alongside `max_turns`. #2 deserves an explicit decision recorded in the PR — the benchmark now measures "can the agent find the skill" rather than "does the description trigger it," which may be the better metric, but it shouldn't change by omission. #4–#6 are comment/prose corrections in text this PR authored.

mohsen-uipath:
Thanks for the thorough review — fixes pushed:

- **#1** (b2848af): gate task now pins `run_limits.max_turns: 1` (task layer wins the per-key merge — verified via config lineage), so it measures what the baselines measured.
- **#3** (b2848af): `task_timeout` 120 → 360; gate now fails on `completion_rate < 100%` so dropped rows can't inflate recall.
- **#4/#5** (0a6c287): READMEs + YAML comments now describe the real stop rule (fail-stop deferred while the positive is undecided) and drop the "verdicts are unchanged" claim.
- **#6** (0a6c287): "Adding a new skill" requires `stop_when: auto` and explains the silent recall deflation when omitted.
- **#7 + Low** (0a6c287, b04ddb3): cost section covers the unarmed path; counts fixed (23 criteria / 1309 rows); stale gate header + timeout table updated.

Not addressed here: #2 (metric-scope decision — for the full suite only now that the gate is pinned to 1 turn) and the `uipath-insights` baseline, which needs a measured number.

PR #2032, opened 2026-07-14
fix(maestro-case): preserve portable resource names across planning

song-zhao-25:
## Summary

This PR defines a portable resource-resolution contract between Case Management Phase 0 (SDD authoring) and Phase 1 (planning). After this change, an approved `sdd.md` contains enough resource intent for Phase 1 to repeat registry discovery on another machine, even when `tasks/registry-resolved.json` was not transferred.

The contract now covers:

- `process`, `agent`, `rpa`, and `api-workflow` tasks
- Action App tasks
- child-case (`case-management`) tasks

It also prevents an old `registry-resolved.json` from silently overriding an edited SDD and preserves the fields Phase 2 needs to create bindings for resolved Action tasks.

## Why

`tasks/registry-resolved.json` is a local planning audit/cache. It may not exist when Phase 0 and Phase 1 run on different machines, and it may be stale when an SDD is edited between runs.

Previously, an unresolved task could lose its intended resource name by writing `<UNRESOLVED>` into the same SDD field Phase 1 needed as its registry query. Without the Phase 0 cache, Phase 1 then had no portable lookup key and could fall back to the task display name, which is not necessarily the deployed resource name.

There were two related correctness gaps:

1. A resolved entry in an old cache could be reused without proving that its task, type, resource name, folder, and identity still matched the current SDD.
2. Resolved Action task entries included the Action App ID but omitted the selected app title and deployment folder. Phase 2 consumes those fields to create `data.name` and `data.folderPath` bindings, so the resulting task could lose its runtime resource binding.

### 1. Portable names are always concrete

The SDD separates portable resource intent from tenant-specific resolution state:

| Task type | Portable lookup name | Resolution fields |
|---|---|---|
| `process`, `agent`, `rpa`, `api-workflow` | `Resolved Resource` | `Folder Path`, `Resource Identity` |
| `action` | Action App title in `HITL Implementation` | `Deployment Folder`, `Action App ID` |
| `case-management` | `Child Case` | `Folder Path`, `Resource Identity` |

The portable lookup-name field must always contain the intended concrete name. Only folder and identity fields use `<UNRESOLVED>` when no live resource was selected.

Phase 0 writes the selected registry entry's canonical name when resolved. When unresolved, it preserves the user-requested name instead of replacing it with `<UNRESOLVED>`.

### 2. `sdd.md` is authoritative; the registry audit is optional

Every portable-resource resolution written to `tasks/registry-resolved.json` now has required association and lookup fields:

```json
{
  "stage": "Review and Follow-up",
  "task": "Review Request",
  "taskType": "action",
  "cacheFile": "action-apps-index.json",
  "searchQuery": "PortableReviewActionProbeQ91",
  "matches": [],
  "selected": null,
  "rationale": "..."
}
```

Phase 1 associates an entry by exact `stage` + `task` and reuses it only when all of these match the current SDD:

- task type
- compatible cache type
- portable search name
- selected entry's canonical name
- exact deployment folder
- resource identity
- version metadata when the SDD carries a version

Missing, unresolved, or mismatched data makes the entry stale. Phase 1 ignores that entry, discovers from the SDD's portable name, and replaces the task's audit entry. The cache can never upgrade or override unresolved or edited SDD fields.

Older audit files that lack the new association fields are treated as stale and rebuilt rather than trusted.

### 3. Resolved Action tasks keep their runtime binding inputs

Resolved Action `tasks.md` entries now include all three selected-resource fields:

```markdown
## T<n>: Add action task "Resolved App Review" to "Review and Follow-up"
- taskTypeId: b20c471c-adef-4b37-a884-897f56ca53bc
- name: "purchaseorderapp-1782974854"
- folder-path: "Shared"
```

Phase 2 uses `name` and `folder-path` to create the Action App binding pair and uses `taskTypeId` for schema discovery/audit.

### 4. Action Apps and child cases remain non-creatable

This change gives Action and child-case tasks the same portable-name handoff behavior as the runnable task types, but it does not add inline creation for either type. A zero-match Action App or child case proceeds through the existing unresolved-resource approval gate and placeholder path.

## Phase 1 Decision Flow

```text
Read the type-specific portable name from the approved SDD
  |
  +-- missing or <UNRESOLVED> name
  |     -> surface an SDD contract error; never substitute Task Name
  |
  +-- concrete name
        |
        +-- matching task entry exists in registry-resolved.json
        |     |
        |     +-- type/cache/name/folder/identity/version all match SDD
        |     |     -> reuse the cached resolution
        |     |
        |     +-- any field is missing, unresolved, or different
        |           -> treat as stale and re-run discovery from SDD
        |
        +-- cache entry or file is absent
              -> run discovery from SDD

Discovery result
  |
  +-- exact compatible match
  |     -> select canonical name, exact folder, and identity
  |     -> write corrected audit entry and resolved tasks.md entry
  |
  +-- zero matches
        -> run the existing unresolved-resource approval gate
        -> Action App / child case: placeholder only
        -> agent / API workflow: existing create-or-placeholder behavior
```

### Cross-machine runnable handoff

Before, an unresolved API workflow could lose the only usable query:

```markdown
**Resolved Resource:** <UNRESOLVED>
**Folder Path:** <UNRESOLVED>
**Resource Identity:** <UNRESOLVED>
```

After, intent remains portable while resolution state stays explicit:

```markdown
**Resolved Resource:** FinancialPostingFunction
**Folder Path:** <UNRESOLVED>
**Resource Identity:** <UNRESOLVED>
```

Phase 1 can now search `api-index.json` for `FinancialPostingFunction` without the Phase 0 cache.

### Stale Action selection

If the cache says task `Resolved App Review` selected `SimpleApprovalApp`, but the current SDD says `purchaseorderapp-1782974854` in `Shared`, Phase 1 rejects the cache entry and searches `action-apps-index.json` using the current SDD title. It then replaces the stale audit entry and emits the selected app title, folder, and ID into `tasks.md`.

### Display name is not resource intent

For a task displayed as `Review Request` whose intended Action App is `PortableReviewActionProbeQ91`, discovery uses `PortableReviewActionProbeQ91`. It never searches for `Review Request`.

## Scope and Compatibility

- This is intentionally a stricter pre-release contract; full backward compatibility is not required because this behavior has not shipped to customers.
- An older SDD that has `<UNRESOLVED>` in a portable name field is surfaced as invalid instead of receiving a guessed lookup name.
- An older `registry-resolved.json` without the required task association or matching resolution fields is safely treated as stale and regenerated.
- Connector discovery and connector binding behavior are unchanged and out of scope.
- Create-on-missing deduplication is deliberately separate in stacked PR #1966.

## Test Coverage

| Coder-eval task | Coverage |
|---|---|
| `skill-case-phase-0-finalize-draft` | Phase 0 preserves concrete unresolved agent names in task blocks and the Section 4 rollup while folder and identity remain unresolved. |
| `skill-case-registry-handoff-sdd-only` | Phase 1 regenerates an absent cache for API-workflow and agent tasks using SDD resource names, then verifies selected entries against the live type-specific caches. |
| `skill-case-registry-handoff-action-case-sdd-only` | Phase 1 rejects deliberately stale Action/child-case entries, checks full exact match sets against `action-apps-index.json` and `caseManagement-index.json`, verifies placeholders for misses, and verifies `name`/`folder-path`/ID for the resolved Action task. |

Validation completed:

- Skill frontmatter and structure validation
- `git diff --check`
- Coder-eval schema planning for all three tasks
- CLI verb validation for the handoff tasks
- Positive and negative controls for the deterministic registry checkers
- `skill-case-registry-handoff-action-case-sdd-only` executed with coder-eval CLI overrides `--type codex --model gpt-5.4`: **6/6 criteria passed, score 1.000**
- PR CI and skill smoke checks passed

Behavioral coder-eval execution still pending for:

- `skill-case-phase-0-finalize-draft`
- `skill-case-registry-handoff-sdd-only`

The no-cache task's GPT override was attempted without modifying task YAML, but coder-eval currently cannot unset its task-local Claude-only `sdk_options` through `-D`; it rejects the configuration before execution.


abhiram-vad (approved):
LGTM

charlesliu9 (approved):

PR #541, opened 2026-05-04
feat(hitl-skill): add Case Management surface — hitl-casetask-action.md + SKILL.md routing

dushyant-uipath:
## What

Adds Human-in-the-Loop support for the **Case Management** surface to the `uipath-human-in-the-loop` skill.

## Changes

**`SKILL.md`**
- Adds Case surface detection: if `caseplan.json` is present (root type `case-management:root`), route to `hitl-casetask-action.md`
- Step 3 now presents two options for Case: QuickForm and App-based
- Step 5 has a Case section linking to the reference doc

**`references/hitl-casetask-action.md`** (new file)
Three paths, confirm before writing:

| Path | When to use |
|---|---|
| QuickForm | Structured form fields, no deployed app. Schema lives in a separate `.hitl.json` file alongside `caseplan.json`. Action task gets `context[]` entries: `hitlType: "quick"`, `_schemaFileId`, `hitlSchemaId`. |
| Generic | Simple approve/reject, no form fields. `data.taskTitle` only. |
| App-based | Existing deployed Action Center app. `hitlType: "custom"` + `=bindings.*` references. |

Key implementation detail for QuickForm: the `.hitl.json` file uses a **unified `fields[]` array** where each field has a `direction` property (`"input"` / `"output"` / `"inOut"`). This is what the runtime sync hook (`useHitlQuickFormBpmnSync`) reads from `parsed?.fields`. The `getDefaultHitlSchema()` function creates an initial file with separate `inputs/outputs/inOuts` keys (empty), but that's just the blank starting state.

**Eval tests**
- `smoke_09_case_quickform.yaml` — QuickForm lifecycle: creates caseplan.json, adds QuickForm task, checks `.hitl.json` file + context entries
- `smoke_10_case_generic.yaml` — Generic task lifecycle
- `quality_12_case_path_selection.yaml` — Path selection for data enrichment: should pick QuickForm (not Generic, not App-based)

## Testing

1. Check out the branch: `git checkout feat/hitl-case-surface`
2. Point the skill runner at the branch and run the three new eval tests:
   - `skill-hitl-smoke-case-quickform`
   - `skill-hitl-smoke-case-generic`
   - `skill-hitl-quality-case-path-selection`
3. Manually: ask Claude Code with this skill to "add a QuickForm HITL to my case project" on a real `caseplan.json` and verify:
   - A `.hitl.json` file is created with `fields[]` + `direction` (not separate arrays)
[...]

PR #2056, opened 2026-07-15
feat(hitl-skill): add Case Management surface — hitl-casetask-action.md + SKILL.md routing

dushyant-uipath:
## What

Adds Human-in-the-Loop support for the **Case Management** surface to the `uipath-human-in-the-loop` skill. Supersedes #541 (closed — same branch, force-pushed after a rebase onto latest `main` plus fixes described below; GitHub won't allow reopening a closed PR whose branch was force-pushed).

## Changes

**`SKILL.md`**
- Adds Case surface detection: if `caseplan.json` is present (root type `case-management:root`), route to `hitl-casetask-action.md`
- Step 3 now presents two options for Case: QuickForm and App-based
- Step 5 has a Case section linking to the reference doc

**`references/hitl-casetask-action.md`** (new file)
Two paths, confirm before writing:

| Path | When to use |
|---|---|
| QuickForm | Structured form fields, no deployed app. Schema lives in a separate `.hitl.json` file alongside `caseplan.json`. Action task gets `context[]` entries: `hitlType: "quick"`, `_schemaFileId`, `hitlSchemaId`. |
| App-based | Existing deployed Action Center app. `hitlType: "custom"` + `=bindings.*` references. |

Key implementation detail for QuickForm: the `.hitl.json` file uses a **unified `fields[]` array** where each field has a `direction` property (`"input"` / `"output"` / `"inOut"`).

Also documents that every action task written via direct JSON — QuickForm or App-based — needs `entryConditions[]` (e.g. `current-stage-entered`) on the task node itself. This is a general Case Management structural requirement, not HITL-specific, but its absence from the doc caused the agent to inconsistently add it, which surfaced as a `uip maestro case validate` failure (`Task has no entry rules`) during local verification below.

**Eval tests**
- `smoke_09_case_quickform.yaml` — QuickForm lifecycle: creates `caseplan.json`, adds a QuickForm task, checks `.hitl.json` file + context entries
- `quality_12_case_path_selection.yaml` — Path selection for a data-enrichment scenario: should pick QuickForm (not App-based)

## Verification (actually run locally, not just described)

Ran both new tests plus a regression sample of 8 existing `uipath-human-in-the-loop` smoke/quality tests via `coder-eval` against this branch:

| Task | Result |
|---|---|
| `skill-hitl-smoke-case-quickform` (new) | SUCCESS, 1.000 |
| `skill-hitl-quality-case-path-selection` (new) | SUCCESS, 1.000 |
| 7 existing smoke/quality regression tests | SUCCESS, 1.000 each |
| `skill-hitl-smoke-neg-admin` | Pre-existing known-flaky llm_judge test (0.4), already slated for deletion via #1920/#1921 — unrelated to this change |

Along the way, fixed two test-authoring bugs found in the original tests (not skill bugs): `path: "*.hitl.json"` isn't glob-matched by this repo's `coder-eval` version (treated as a literal filename), and a `grep -L` bindings-check had non-portable exit-code semantics between BSD and GNU grep. Both criteria now use `find`-based `run_command` checks matching this repo's existing convention (see e.g. `uipath-coded-apps` tests). Also added `metadata.caseExitRules` to both tests' seed fixtures — a pre-existing case-level completion-rule requirement unrelated to HITL that a bare-bones seed `caseplan.json` was missing.

Manually confirmed the actual produced artifacts match the documented shapes:
- `.hitl.json` has `fields[]` + `direction` (not separate `inputs`/`outputs`/`inOuts` arrays)
- `caseplan.json` action task has `data.context[]` with `hitlType: "quick"`, `_schemaFileId`, `hitlSchemaId`
- No `data.formType` or inline `data.schema` in `caseplan.json`
- `root.data.uipath.bindings[]` unchanged for the QuickForm path

### DIFF
--- a/skills/uipath-human-in-the-loop/SKILL.md
+++ b/skills/uipath-human-in-the-loop/SKILL.md
@@ -1,12 +1,12 @@
 ---
 name: uipath-human-in-the-loop
-description: "UiPath Human-in-the-Loop / HITL node authoring — building approval gates, escalations, write-back validation, and data enrichment checkpoints in Flow, Maestro, or Coded Agents. NOT for managing, reassigning, or monitoring tasks at runtime (use uipath-tasks for that)."
+description: "UiPath Human-in-the-Loop / HITL node authoring — building approval gates, escalations, write-back validation, and data enrichment checkpoints in Flow, Maestro, Case Management, or Coded Agents. NOT for managing, reassigning, or monitoring tasks at runtime (use uipath-tasks for that)."
 allowed-tools: Bash, Read, Write, Edit, Glob, Grep
 ---
 
 # UiPath Human-in-the-Loop Assistant
 
-Recognizes when a business process needs a human decision point, designs the task schema through conversation, and wires the HITL node into the automation — Flow, Maestro, or Agent.
+Recognizes when a business process needs a human decision point, designs the task schema through conversation, and wires the HITL node into the automation — Flow, Maestro, Case Management, or Agent.
 
 > **Coded agents:** for wiring HITL inside a coded agent, use the `uipath-agents` skill — see `skills/uipath-agents/references/coded/capabilities/human-in-the-loop.md`.
 
@@ -72,6 +72,11 @@ Run these checks in order:
 # Check for a .flow file (Flow project)
 find . -name "*.flow" -maxdepth 4 | head -5
 
+# Check for a Case Management project. The on-disk filename varies
+# (caseplan.json, or content/<name>.json.bpmn under a `uip maestro case init`
+# project) — detect by content marker, not by filename.
+grep -rl '"case-management:root"' --include="*.json*" . 2>/dev/null | head -3
+
 # Check for agent.json (Low-Code Agent project)
 find . -name "agent.json" -maxdepth 4 | head -3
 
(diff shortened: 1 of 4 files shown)


dushyant-uipath:
@claude please re-review — the branch has been substantially updated since your last review (which was against the initial push). Since then:

- Field types fixed: `text`→`string`, `dateTime`→`datetime` in `hitl-casetask-action.md` (your High-severity finding) — confirmed canonical against `hitl-schema-types`' own `field-type.ts`.
- "Present three options" → "Present two options" in the Case surface section (your Medium finding).
- Added a `caseplan.json` detection command to Step 1 (content-marker based — `grep -rl '"case-management:root"'` — since the on-disk filename varies; a literal `find -name caseplan.json` would miss real projects, confirmed live).
- Frontmatter `description` now mentions Case Management.
- `id`-generation convention clarified (lowercase, no separator — verified against real agent output, not guessed).
- `entryConditions[]` requirement documented (root-caused a real timeout bug during verification).
- Removed the `choice.json` self-report pattern from `quality_12` per the earlier lint-task finding, replaced with a case-insensitive artifact check.
- Rebased onto latest `main` and fixed a newly-added, unrelated `sandbox.driver: tempdir` lint rule.

All fixed since your last pass. Both new eval tests verified locally end-to-end (`skill-hitl-smoke-case-quickform`, `skill-hitl-quality-case-path-selection`, both SUCCESS/1.000), and CI is green.

Still open from your last review, for reference: no dedicated `e2e` test for the Case surface (flagged as Medium, but existing local verification covers the same ground), and `mode:*` tags missing (repo-wide gap, not introduced by this PR).

Sahil-Wakode (approved):

PR #2062, opened 2026-07-16
feat(uipath-troubleshoot): add Click silent no-op playbook (UI Automation)

Stefan-Virgil:
## What

Adds `click-silent-no-op.md` to the UI Automation package — the **no-signature counterpart** to `verify-execution-failure.md` — for the "NClick reported `Successful` but did nothing" family (no exception, job `Successful`, zero Error logs).

Covers:
- **Causes:** input-method/target-tech mismatch (`Simulate`/`SendWindowMessages` on Java/SAP/legacy Win32), overlay intercept, wrong/duplicate element, lost focus, DOM race.
- **Routing:** via the no-signature table ("Job Successful but the action had no effect").
- **Source-required investigation:** read `InteractionMode` + enclosing scope target technology; establish runtime proof of no-effect before concluding (§6 gate).
- **Resolution:** cause decision tree (A–E) + closing the detection gap (F) with a real Verify Execution target.

## Wiring

- `summary.md` — new row.
- `investigation_guide.md` item 8 — redirect to the dedicated playbook (keeps the wrong-page pointer as "see also").
- `scope-container-wrong-page.md` — reciprocal cross-link from its silent variant.

## Notes

- Self-contained; degrades gracefully. No cross-skill file imports.
- Cross-references `verify-execution-failure.md` (fires when `VerifyOptions` was set and threw) so the two are unambiguous.
- A companion coder-eval test (Simulate-on-Java scenario) ships in a separate PR (`test/troubleshoot-click-noop-simulate-tech`) — **merge this playbook PR first** so that test grades against the real playbook in CI.


Stefan-Virgil:
Addressed the review (`599452aa`):

- **Activity-list mismatch (Low):** added `NKeyboardShortcuts` to the `summary.md` row so it matches the playbook's "applies to" list — both now enumerate the same 7 input activities. Send Keyboard Shortcuts posts input via the input method and can silently no-op on unsupported target tech, so it belongs in the family.

On the manual-review flags:
- **Domain accuracy** (`Simulate`/`SendWindowMessages` unsupported on Java/SAP/legacy Win32): this matches UiPath's documented input-method support — Simulate/Window-messages work for web + WPF/WinForms but not Java (JAB) or SAP GUI, which require Hardware Events. The playbook also directs the agent to confirm via `docsai` at investigation step 6 rather than asserting it blind.
- **`NKeyboardShortcuts` in the family:** kept it in — its driver posts input via the same input-method setting, so it exhibits the same silent-no-op mode on unsupported tech.

Stefan-Virgil:
@claude please re-review — the Low activity-list finding was addressed in `599452aa` (added `NKeyboardShortcuts` to the summary.md row to match the playbook).

MarinRzv (approved):

PR #2063, opened 2026-07-16
test(uipath-troubleshoot): add NClick Simulate-on-Java silent no-op scenario

Stefan-Virgil:
## What

e2e diagnose scenario exercising the **Click silent no-op** playbook (added in #2062).

A modern `NClick` with `InteractionMode=Simulate` targets a **Java** desktop app, where Simulate is unsupported — so the click reports `Successful` while the Submit button is never actuated: job `Successful`, zero Error logs, claim status still `Draft`. The agent must route via the no-signature table to `click-silent-no-op.md`, confirm the no-op from runtime logs, and name the input-method/target-tech mismatch as the root cause with the Hardware Events fix.

## Fixtures

- Manifest-driven `uip` mock (`data/m/r/`): folders, jobs list/get (Successful, HA disabled), Error logs (empty), Info logs (status still `Draft`); `docsai ask` is **passthrough** (no canned answer on disk).
- Frozen project source (`process/`): `Simulate` Click inside a Java `Use Application` scope, no Verify Execution.
- `RESOLUTION.md` reference for the judge.

## Grading — conclusion, not trajectory

- `skill_triggered` (gate)
- `command_executed` **advisory** (`pass_threshold: 0`) — records whether the agent took the Orchestrator path; non-gating because this is a **source-required** scenario (the agent can equally diagnose from `Main.xaml`).
- `llm_judge` against `RESOLUTION.md` (threshold 0.7).

## Hint hygiene

The agent-visible surface (`process/` + `data/`) carries **no diagnosis hints** — only legitimate evidence (`InteractionMode=Simulate`, `app='java.exe'` selectors, `Draft` status log). The input-method/target-tech knowledge lives in the skill's playbook, not a planted fixture. Verified by scan.

## Validation

**3/3 SUCCESS** over 3 parallel reps (`experiments/default.yaml`, `--repeats 3 -j 3`): `skill_triggered`=1.0 and `llm_judge`=1.0 in every rep; the advisory `command_executed` confirmed the expected path split (1/3 Orchestrator, 2/3 source-only).

## Merge order

Depends on #2062 for correct CI grading — **merge #2062 first**.


Stefan-Virgil:
Addressed both Medium findings (`4ed2b92d`) — thanks for catching the suite convention (`tests/tasks/uipath-troubleshoot/CLAUDE.md`), which I'd missed:

- **(1) `command_executed`:** removed. The task now has exactly the two-criteria contract — `skill_triggered` + `llm_judge`. Agreed it encodes one investigation path (and this is a source-required scenario where the agent can legitimately diagnose from `Main.xaml` alone), so even the non-gating advisory form was against the convention.
- **(2) `llm_judge` prompt:** replaced with the canonical generic rubric (verbatim match with `uia-verify-execution-failure/task.yaml`). The scenario-specific scoring detail stays in `RESOLUTION.md`, which the judge reads via `include_reference`.

On the manual-review notes:
- **`README.md` link to `click-silent-no-op.md`:** expected to be broken until #2062 merges — merge order (#2062 first) is called out in both PR descriptions. README is docs-only, not read in the sandbox.
- **Lost path-split telemetry:** acceptable — it was only incidental (1/3 Orchestrator, 2/3 source) and not part of the grading contract.

The prior 3/3 validation stands: this change only removes a criterion and reverts the judge wording to the generic rubric, so `skill_triggered` + `llm_judge` (which were 1.0 across all 3 reps) are unaffected.

Stefan-Virgil:
@claude please re-review — both Medium findings addressed in `4ed2b92d`: removed the `command_executed` criterion and reverted the `llm_judge` prompt to the canonical generic rubric, per the suite two-criteria convention.

MarinRzv (approved):

PR #2098, opened 2026-07-16
perf(maestro-case): use bounded edits for faster case builds

song-zhao-25:
## Summary

Reduce Maestro Case build latency and token cost by preventing the agent from repeatedly regenerating a populated `caseplan.json`.

The LLM remains the sole author of the case JSON. This change only narrows how its mutations are applied:

- Allow `Write(caseplan.json)` only for the initial T01 root scaffold.
- Require bounded `Edit` operations for every later mutation.
- Group edits by stable target, such as one stage task array, one task I/O array, one condition target, or one SLA target.
- Cap each replacement payload at 30 KB and split larger targets.
- Track progress at section level instead of creating a todo update for every T-entry.
- Remove conflicting whole-file-write wording from cross-cutting and trigger references.

## Motivation

The Golden Expense E2E baseline took 52m55s and cost $9.73. Model generation dominated the run:

- 217,198 output tokens
- 13,932,735 cache-read tokens
- three `caseplan.json` writes, including a 73 KB populated rewrite
- one 32,000-token response without a tool call before context reconstruction

The existing guidance allowed large sections to be composed in reasoning and emitted through whole-section or whole-file writes. That concentrates output into long model turns and risks silently dropping unrelated fields. Bounded edits should reduce those large generations while preserving LLM-authored JSON and the existing phase/validation model.

## Scope and compatibility

- No task YAML, grader, workflow, CLI, or JSON schema changes.
- No deterministic renderer or helper script is introduced.
- T01 scaffold behavior is unchanged.
- Existing validation boundaries and hard stops remain unchanged.
- The primary behavior change is mutation granularity and progress bookkeeping.

## Validation

- `git diff --check`
- `quick_validate.py skills/uipath-maestro-case` — passes
- Confirmed the PR diff contains only the nine Maestro Case skill/reference files.

The unchanged Golden Expense E2E was dispatched in [run 29539778252](https://github.com/UiPath/skills/actions/runs/29539778252). Four attempts failed before task execution because `astral-sh/setup-uv@v4` received GitHub's “Unicorn” error page. Claude, Docker build, and the grader never started, so performance comparison is still pending. This PR is draft until a complete E2E run confirms quality, duration, and cost.

## Evaluation target

Compare the next successful run against the baseline for:

- deterministic grader result
- total duration and cost
- output and cache-read tokens
- largest `caseplan.json` write/edit
- number of duplicate schema lookup calls


song-zhao-25:
Closing this draft as a documented negative performance experiment. Accuracy remained 6/6 with score 1.000, but duration improved only 6% while cost increased 25%, cache-read tokens increased 51%, and agent/tool turns increased about 26%. The retained findings and recommended coarser-batching follow-ups are recorded in [the performance investigation](https://github.com/UiPath/skills/blob/experiment/maestro-case-llm-build-speed/docs/performance/maestro-case-build-performance.md).

PR #2100, opened 2026-07-17
docs(uipath-ixp): note that folder deployment is not supported

andrei-uipath:
Sets the expectation in the uipath-ixp skill that publishing an IXP model does not deploy it to an Orchestrator folder — that folder/environment binding is a product-side step with no `uip ixp` (or other CLI) equivalent. There are live cases where a project is created for use in a Flow and an agent then attempts folder deployment as the final step, landing the model locally instead of in the target folder (ACTV-89471).

The note lands at the publish guidance in SKILL.md and in the cli-reference Deployments section. Rather than only telling the agent what it can't do, it directs the correct behavior: publishing is the last step this skill performs, and when the user explicitly asks to deploy, the agent hands the folder/environment binding back to them (done in-product) and never falls back to a local or improvised deployment. (Publishing/tagging via `uip ixp` stays within the project; deploying to a folder — the product-side step — is what makes a model available to Maestro Flow.)

ACTV-89471: https://uipath.atlassian.net/browse/ACTV-89471

https://claude.ai/code/session_0164ykfgunBCrvhrUMHbaGEQ


richsilv (commented):
I think the agent could do with some instruction that the correct thing to do is tell the user that _they_ need to deploy the model, rather than just that the agent itself can't do so. Otherwise I'm concerned that it will try to find a workaround by using the CLI docs or something.

skills/uipath-ixp/references/cli-reference.md
```diff
@@ -129,7 +129,7 @@ Structural edits to a field within an existing field group. For instruction-only
 
 ## Deployments
 
-For working with runtime (deployed) IXP models — separate from the training workflow above.
+For working with runtime (deployed) IXP models — separate from the training workflow above. Publishing/tagging **does not** deploy a model to an Orchestrator folder — folder/environment binding is completed **by the user in-product**. A published model is referenced directly by downstream consumers (e.g. Maestro Flow).
```
richsilv: I think this is a bit ambiguous - it uses "publishing/tagging" at the start to mean publishing to the project (which is not available to Flow), and then "published model" later to refer to the model deployed to a folder which *is* available to Flow.

richsilv (commented):
I think the wording is a bit ambiguous, partly as a result of our ridiculous historic naming of stuff in IXP.

But today:
_Publish_ - is really only something you do with packages (but we have been using it interchangeably with deployment)
_Deploy to project_ - is the thing that makes the model available to tag, doesn't require publishing a package first, and doesn't make it available to Flow
_Deploy to folder_ - requires a published package (although the package publishing can happen in the same step), and does make it available to Flow

I don't think much needs to change here, apart from the references to publish(ed) in the Deployments section.

richsilv (approved):

cezara98t (approved):

PR #2101, opened 2026-07-17
fix(uipath-ixp): hard DO-NOT-TRIGGER for flow work in frontmatter description

andrei-uipath:
Skill selection happens on the frontmatter description alone — the body-level "When NOT to Use This Skill" deferral shipped in #1348 (RE-12440) is only read *after* uipath-ixp activates, so it never prevents the misfire [ACTV-89470] describes: agents picking uipath-ixp during flow discovery. In the description, the deferral was one soft trailing sentence ("For IxP nodes or model listings inside a .flow / Maestro flow→uipath-maestro-flow") that loses to the strong positive triggers ("pull metrics and model versions, publish/tag/roll back model versions").

This replaces it with a hard negative trigger in the same idiom uipath-maestro-flow already uses: "DO NOT TRIGGER during .flow / Maestro Flow work — discovering or listing IxP / document-extraction models, extractors, or nodes available to Maestro Flow, and adding or wiring an IxP node, belong to uipath-maestro-flow even when they sound like IXP model management." (Per review, the earlier "Standalone IXP projects ONLY" qualifier was dropped as ambiguous — the capability list already carries the positive scope and the negative trigger states the exclusion precisely.) Description is 727 chars (cap 1024); body text unchanged.

https://claude.ai/code/session_0164ykfgunBCrvhrUMHbaGEQ

[ACTV-89470]: https://uipath.atlassian.net/browse/ACTV-89470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-ixp/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
+description: "UiPath IXP (Document Understanding) via `uip ixp` — create projects (with autopilot taxonomy suggestion, an imported taxonomy file, or empty), upload/download/delete documents, author the taxonomy (field groups, fields, data types, per-field and overall extraction instructions), configure the extraction model and pre-processing, review/confirm/unconfirm predictions, mark fields missing, pull metrics and model versions, publish/tag/roll back model versions. Standalone IXP projects ONLY. DO NOT TRIGGER during .flow / Maestro Flow work — discovering or listing IxP / document-extraction models, extractors, or nodes available to Maestro Flow, and adding or wiring an IxP node, belong to uipath-maestro-flow even when they sound like IXP model management."
```
richsilv: Is "standalone" a bit ambiguous? It would be nice to use "design-time", but the Flow is in design-time so maybe that's even more ambiguous. But perhaps it only needs the "DO NOT TRIGGER" bit - that would be my suggestion.

richsilv (commented):
I think @rockymadden will want to see this given that it's a change to front-matter.

richsilv (approved):
Looks good to me. Not sure if you need a coding agents team reviewer to change the front matter.

cezara98t (approved):

PR #2103, opened 2026-07-17
feat(skills): add performance scenario and report workflows

neha-pahuuja:
Adds the perf-scenario command surface and operational rules to the uipath-test SKILL.md, the scenario lifecycle guide (adapted from #789) and a new report-authoring guide covering single-run persona reports and multi-run comparison reports (structure mirrors the AgentHub tool contracts), plus eval tasks for the workflows.

Builds on the performance scenario skill draft in #789.


tests/tasks/uipath-test/e2e_perf_full_lifecycle.yaml
```diff
@@ -0,0 +1,167 @@
+task_id: skill-test-perf-full-lifecycle
```
ganeshborle, line 1: You can add the skip marker here. skip:true

tests/tasks/uipath-test/quality_perf_add_loadgroup.yaml
```diff
@@ -0,0 +1,70 @@
...
+tags: [uipath-test, integration, mode:operate, lifecycle:setup, feature:perf-loadgroup]
+
+sandbox:
+  python: {}
+  node:
+    env_packages:
+      - "@uipath/test-manager-tool"
+
+initial_prompt: |
+  Before starting, load the uipath-test skill and follow its workflow.
```
ganeshborle: `  Before starting, load the uipath-test skill and follow its workflow.`
This line should not be present. Its the real user prompt so that skill should load automatically. This will get evalulated as skill invocation.

tests/tasks/uipath-test/quality_perf_execute_wait.yaml
```diff
@@ -0,0 +1,63 @@
...
+  max_turns: 50
+
+sandbox:
+  python: {}
+  node:
+    env_packages:
+      - "@uipath/test-manager-tool"
+
+initial_prompt: |
+  Before starting, load the uipath-test skill and follow its workflow.
```
ganeshborle: Same, maybe remove this statement from all the places.

tests/tasks/uipath-test/quality_perf_report_compact.yaml
```diff
@@ -0,0 +1,83 @@
...
+  the Release Manager persona's structural sections (a clear go/no-go
+  statement, SLO citation, success-rate reference) without raw
+  time-series dumps.
+tags: [uipath-test, integration, mode:operate, lifecycle:generate, feature:perf-scenario]
+
+agent:
+  type: claude-code
+
+run_limits:
+  max_turns: 60
```
ganeshborle: See, if you can reduce the max_turns from all the tests.

tests/tasks/uipath-test/e2e_perf_full_lifecycle.yaml
navneet-ghub: Done in 435320aaa — added `skip: true` with a comment explaining the blocker: the e2e needs a Performance Testing runtime on the eval tenant plus the CI bot's `PerformanceScenarios.*` Test Manager permissions (perf-scenario reads currently return HTTP 403). Will re-enable once both exist.

tests/tasks/uipath-test/quality_perf_add_loadgroup.yaml
navneet-ghub: Done in 435320aaa — removed the line here. Skill activation for these prompts is covered separately by the activation-gate entries in `tests/tasks/activation/uipath-test.jsonl`, and none of the perf tasks grade `skill_triggered`, so nothing depended on it.

tests/tasks/uipath-test/quality_perf_execute_wait.yaml
navneet-ghub: Done in 435320aaa — swept the line from all eight perf task prompts (both smoke tasks, the five quality tasks, and the e2e).

tests/tasks/uipath-test/quality_perf_report_compact.yaml
navneet-ghub: Done in 435320aaa — reduced across the board: quality tasks 50/60 → 35 (each runs a handful of CLI calls plus report authoring; 35 leaves retry headroom over the ~15–20 turns they actually use), e2e 120 → 60 (the full pipeline is ~25–30 turns since each `--wait` is a single call). Kept above the smoke-experiment default region to avoid flaky `max_turns_exhausted` failures.

ganeshborle (approved):
LGTM

PR #2104, opened 2026-07-17
Add InvokeWorkflowFile troubleshooting playbook coverage (design-time + runtime)

Stefan-Virgil:
## What

Comprehensive troubleshooting coverage for `UiPath.Core.Activities.InvokeWorkflowFile`, which previously had a single runtime "invoke faulted" playbook and **zero test coverage**.

### Playbook (classic-activities)

- **New** `invoke-workflow-file-design-time-errors.md` — Studio design/build-time errors: unknown-member `ArgumentsVariable` (package-version mismatch), Cache Mechanism Error code 7, invoked `.xaml` outside the project root, unmapped required argument, plus the maintenance checklist (re-add activity, fix double-backslashes, rebuild `project.json`).
- **Completed** `invoke-workflow-failed.md` — added the **persistence-not-supported** branch (it was listed only as a symptom, with no cause/investigation/resolution), plus `ignoredFiles` / `UnSafe` / `TargetSession` specifics. Property names verified against the repo's own activity docs.
- Cross-redirected the two playbooks so design-time vs runtime route cleanly; each error signature resolves to one dominant playbook via the skill's grep routing.

### Tests — 6 e2e scenarios (`skill_triggered` + `llm_judge`, canonical troubleshoot shape)

| Class | Scenario | Cause |
|---|---|---|
| Design-time | `classic-invoke-cache-error-7` | project cache corruption |
| Runtime | `classic-invoke-child-throws` | child workflow throws (one-hop tracing) |
| Runtime | `classic-invoke-arg-mismatch` | stale argument key → `ArgumentException` |
| Runtime | `classic-invoke-session-isolated` | non-`Current` `TargetSession` + `UnSafe=False` |
| Runtime | `classic-invoke-ignored-file` | invoked `.xaml` in `ignoredFiles` → runtime FNF |
| Runtime | `classic-invoke-persistence` | `RunJob` Suspend + `supportsPersistence: false` |

Prompts use investigate framing with an Orchestrator-folder anchor so the fault routes through the skill's CLI investigation rather than a direct source read.

### Split out for rework

Three design-time scenarios — `classic-invoke-arg-required`, `classic-invoke-file-not-packed`, `classic-invoke-unknown-member` — were moved to branch `test/troubleshoot-invoke-workflow-design-time`. Their faults are fully source-visible with an obvious fix, so the agent resolves them directly (via `uip rpa validate`/`build`/`pack` or a source read) without invoking the troubleshoot skill — `skill_triggered` lands 0/3–1/3 regardless of prompt wording. They will be reworked with runtime-only evidence (mocked job logs, cause hidden from the source) before shipping. Both playbooks still document all of these cases.

## Validation

Re-validated with 3 coder-eval replicates per scenario (`experiments/default.yaml`): **skill_triggered 17/18** (`arg-mismatch` 2/3, the other five 3/3), with `llm_judge` ≥ 0.9 on every run.

## Notes

- Each scenario carries the fault **only** in retrieved evidence (logs/source), never in project/workflow names, prompts, or manifests — no-hints rule enforced.
- Fixtures follow the faithful-replay template; runtime scenarios use canned faulted-job responses (`jobs get`/`logs`/`history`/`traces` + `folders`/`jobs list`), the design-time `cache-error-7` scenario is source-only with `docsai` passthrough. Mock fixture filenames are hashed to `sha1[:10]`.


Stefan-Virgil:
Addressed the review finding in fd214b3e: added the required `lifecycle:discover` tag to all 9 task YAMLs.

`discover` is the correct phase per the [tag taxonomy](https://github.com/UiPath/skills/blob/main/tests/README.md#tag-taxonomy) — these are read-only diagnostic investigations (list/get/inspect existing state), matching the current `db-*` troubleshoot scenarios which use `lifecycle:discover`.

Minor note for the record: the older troubleshoot tasks in this directory (e.g. `classic-movefile-source-not-found`, `classic-openapp-wrong-scope-type`, `sys-getasset-*`) do **not** carry a lifecycle tag — only ~43/204 do — so the new tasks are now *more* compliant than most existing siblings, not less. Bringing the rest up to standard is out of scope for this PR.

MarinRzv (approved):

PR #2119, opened 2026-07-18
fix(maestro-case): preserve same-name output bindings

song-zhao-25:
## Summary

- preserve every explicit output-binding operator and both operands from SDD through `tasks.md`
- reject bare SDD Outputs rows; reserve bare `tasks.md` outputs for schema discovery
- dispatch output generation by operator so equal-name extracts such as `greeting -> greeting` retain reassignment semantics and `originalVar`
- resolve both whole-value `<-` and in-expression `$xref` references through one output-reference-ID algorithm, using source `.id` rather than a reassigned output `.var`
- resolve custom `=` outputs, which intentionally have no `.id`, through their verified root Case-variable companion
- define equal-name extraction as a controlled alias while still deduplicating independently minted output IDs
- centralize the canonical output-row grammar in the shared I/O-binding planner and trim duplicate allocation guidance
- add a general `io_binding` e2e matrix using existing deterministic tenant resources; keep the original `single_node/api-workflow` test unchanged
- make Case e2e solution cleanup non-interactive with the CLI-required `--yes`, add regression coverage, and wire cleanup into the new task

## Regression coverage

The new `tests/tasks/uipath-maestro-case/io_binding/` eval covers:

- literal, Case-variable, prior-task, and JavaScript-expression inputs
- schema-discovered bare output auto-minting without a bare SDD row
- renamed and nested extraction
- literal, variable-copy, computed, and metadata assignments
- the exact `estimatedAge -> estimatedAge` controlled-alias regression
- whole-value and `$xref` consumers of a collision-safe reassigned output whose `id` differs from its Case-variable `var`
- whole-value and `$xref` consumers of a custom output whose missing `.id` requires root-companion fallback
- global output-ID uniqueness, lossless `tasks.md`, tenant-resource reuse, solution shape, Case debug execution telemetry, and post-run tenant cleanup

## Verification

- `uip solution delete --help` confirms `--yes` is required for the non-interactive delete path
- cleanup regression test passed after first reproducing the missing-`--yes` failure
- 17 focused Maestro Case checker tests pass together, including Athena and CM Golden Expense collection
- skill structure validation, task YAML dry-run planning, Python compilation, and `git diff --check` pass
- the CLI verb scan reports only five pre-existing stale strings unrelated to this change
- all review-fix and coverage commits are pushed; the expanded eight-task `io_binding` coder eval has not yet been rerun

## Base

This PR targets `main` directly.


charlesliu9:
is the test failure due to the issue fixed in another PR?

song-zhao-25:
@charlesliu9 yes

charlesliu9:
equal-name collision rules are contradictory. The global rule says unrelated collisions must be suffixed, but the implementation recipe says equal-name extracts bypass the uniqueness pool and always keep id == var. If another task already owns greeting, greeting -> greeting can therefore produce duplicate task-output IDs. Clarify that only the root companion collision is exempt; collisions with task, trigger, or rule outputs must still allocate greeting2, with originalVar and target updated accordingly. See skills/uipath-maestro-case/references/plugins/variables/global-vars/impl-json.md:74 and skills/uipath-maestro-case/references/plugins/variables/io-binding/impl-json.md:39.

charlesliu9 (approved):

PR #2121, opened 2026-07-18
feat(uipath-agents): discover llm-as-judge models via CLI [AL-515]

apetraru-uipath:
## What changed?

The coded guardrail skills (`guardrails.md`, `guardrails-recommend.md`) told the agent to *"ask the user which model ID"* for an `llm_as_judge` guardrail's `model` parameter. Now that the CLI ships `uip agent guardrails llm-as-judge-models` (PR UiPath/cli#3113), the skills point at it instead: run the command, pick a `ModelId` from the result, and fall back to asking only if it returns nothing or fails.

- `guardrails.md` — the LLM-as-judge tenant-availability note now runs `uip agent guardrails llm-as-judge-models --output json` and recommends a non-preview model (Haiku/mini class as a sound judge default).
- `guardrails-recommend.md` — the `enum`/`model` parameter-mapping row now discovers models via the command instead of asking blindly.

## How has this been tested?

Doc-only change. The referenced command already exists and is tested in the CLI repo. The existing `llm_as_judge` test task (which passes an explicit model) still exercises the valid explicit-model path.

None — guidance-only. The CLI-side deterministic reviewer rule that validates the chosen model is a separate PR in UiPath/cli.

### DIFF
--- a/tests/tasks/uipath-agents/lowcode/guardrails/llm_as_judge/llm_as_judge.yaml
+++ b/tests/tasks/uipath-agents/lowcode/guardrails/llm_as_judge/llm_as_judge.yaml
@@ -1,0 +1,93 @@
+task_id: skill-agent-guardrail-lowcode-llm-as-judge
+description: >
+  Built-in LLM as Judge guardrail (low-code). End-to-end test: scaffold a
+  standalone low-code agent and add a builtInValidator guardrail for
+  llm_as_judge at Agent scope with a block action. Verifies the skill
+  discovers a model via `uip agent guardrails llm-as-judge-models` (the
+  model Options come back empty from `guardrails list`), then writes the
+  correct shape: $guardrailType builtInValidator, validatorType
+  llm_as_judge, an `enum` model parameter with a non-empty value, a `text`
+  guardrailText parameter, and PascalCase scope values.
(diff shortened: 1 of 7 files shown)


tests/tasks/uipath-agents/lowcode/guardrails/llm_as_judge/llm_as_judge.yaml
```diff
@@ -0,0 +1,95 @@
...
+  llm_as_judge, an `enum` model parameter with a non-empty value, a `text`
+  guardrailText parameter, and PascalCase scope values.
+tags: [uipath-agents, e2e, mode:build, lifecycle:generate, low-code, guardrail]
+initial_prompt: |
+  Create a UiPath low-code agent "JudgeAgent" that answers customer support questions
+  about their accounts. Add an LLM as Judge guardrail at Agent scope with a block action
+  and the message "Response violated the judge rule." The judge rule is: "The agent must
+  only provide factual account information and must never offer refunds or price
+  negotiation." For the model parameter, discover an available model by running
+  `uip agent guardrails llm-as-judge-models --output json` and use one of the returned
```
andreibalas-uipath: The test's prompt should not discole HOW the LLM as judge model should be discovered via the CLI. It should be inferred based on the skill's instructions.

valentinabojan (approved):

PR #2132, opened 2026-07-20
chore(tests): retarget experiments + bump coder-eval pin for the eval refresh

bai-uipath:
## What

Updates the coder_eval experiment configs and version pin for the multi-harness eval refresh.

The experiment `model` moves to `claude-sonnet-5`, and the hardcoded `agent.type: claude-code` is dropped so the experiments are harness-agnostic — the runner selects the harness per run and coder-eval's baseline default supplies claude-code otherwise. The activation experiment also cleans the per-task `.agents` catalog copy between rows so a full non-Claude activation sweep doesn't balloon the run dir. The coder-eval pin moves to 0.8.6, which ships the refreshed-model pricing.

## Why

Keep local and CI runs on the same model as the nightly, let the same experiments drive every harness now that the nightly runs codex and antigravity too, and pull in the pricing that lands in coder-eval 0.8.6.


uipreliga (approved):
🚢

PR #2143, opened 2026-07-21
MST-12614: Prevent edges added to the case

charlesliu9:
Edges are expected to be empty, however it may still be added to the case in some runs.

The guardrails that exist (and where)
Location	                                          Rule
phased-execution.md:46	        "Edges — none authored; schema.edges stays []."
case/impl-json.md:254-256	         Root written with empty edges: [] + post-write check edges length 0
stages/planning.md:35	                 "Do NOT create edges for any stage."
triggers/event/impl-json.md:172	"No edge from this trigger (edges retired)"
case-editing-operations.md:333	"The skill never creates edges"

Three gaps are now closed:
Write-time no-edges rule in stages/impl-json.md + post-write edges === [] check
Same post-write check added to the manual and timer trigger recipes
"Do not copy into a build" banner on the schema's edge examples


song-zhao-25:
could we clean up edge related info from skill as much as possible and keep only minimal existence to support edit case

charlesliu9:
> could we clean up edge related info from skill as much as possible and keep only minimal existence to support edit case

We currently only have the edge skeleton description in case-schema, all others fall into the following buckets:
Bucket 1 — Structural: the empty edges: [] array (must stay)
Bucket 2 — Brownfield read/edit support (the "keep minimal" core)
Bucket 3 — Scattered "don't author edges" prose

Are you suggesting remove any above or something else?

song-zhao-25 (approved):

PR #2166, opened 2026-07-21
test(maestro-bpmn): clarify live debug script output mapping

nikhil-maryala:
Clarifies the live debug eval and script guidance around supported Jint return-to-output mapping instead of direct Globals/vars mutation. Focused coder-eval to run: live_debug_e2e plus script task guidance tasks.

### DIFF
--- a/skills/uipath-maestro-bpmn/references/structural-bpmn.md
+++ b/skills/uipath-maestro-bpmn/references/structural-bpmn.md
@@ -156,6 +156,9 @@ template, but the runtime contract is fixed:
 - Map the return back through `source="=result.response"` (scalar) or
   `source="=result.response.<field>"` (object field); `var` points at a declared
   variable id (do not put the target id in `name`).
+- Do not mutate `Globals.*`, `vars.*`, or process variables inside the script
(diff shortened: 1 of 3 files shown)


nikhil-maryala:
Superseded by #2136 and closing this one, per consolidation (both PRs edited the same `live_debug_e2e` task + `check_live_debug.py` and would conflict).

#2136 already covers the shared core (name-based `product` read from `variables-all`) and adds more: live-recovery grading fallback, turn-budget/anti-wander fixes, and the jint/gateway smoke-check init-nesting fix. This PR's one net-new insight — **Jint scripts must `return` + `uipath:output`-map rather than mutate `Globals.*`/`vars.*`** — has been folded into #2136 in f94670658:
- `structural-bpmn.md`: documents the no-Globals/vars-mutation contract.
- `check_live_debug.py`: fails if an authored script body reads/mutates `Globals.*`/`vars.*` (comments stripped, scoped to the agent's project files).

I did not port the stricter exact-id/exact-source `.bpmn` assertions — #2136's runtime `product==42` read already proves the mapping worked, and rigid structural matching risks false-rejecting valid variants. Thanks!

PR #2172, opened 2026-07-21
feat(uipath-maestro-bpmn): validate via CLI, remove bundled validator

DevMomo:
> **Rebase note (main moved):** main since added a "registry-evidence" fallback and `references/registry-workflow.md` that both read `validator/bpmn-spec.json`. To avoid breaking that, this PR now **keeps `validator/bpmn-spec.json`** as shared registry data and deletes only the rule-engine files (`rules.mjs`, `model.mjs`, `validate-bpmn.mjs`, `uipath-moddle.v1.json`, samples, tests). The bundled *validator* is still removed; validation is via `uip maestro bpmn validate`.

## What

Switches the `uipath-maestro-bpmn` skill to validate BPMN with `uip maestro bpmn validate` and deletes the validator that was bundled with the skill.

- **SKILL.md** step 4 and **references/structural-bpmn.md** now run `uip maestro bpmn validate <file> --output json` as the validation step (exit 0 = valid; exit 1 lists each issue with its rule code; warnings don't block).
- **references/cli-conventions.md** drops the old "there is **no** `uip maestro bpmn validate` command" claim and adds a CLI-freshness note: if `validate` is unknown or runs only the deploy-readiness checks, update the CLI.
- Deletes `skills/uipath-maestro-bpmn/validator/` (rules, model, moddle descriptor, the checked-in `bpmn-spec.json` copy, fixtures, tests) and `.github/workflows/test-maestro-bpmn-validator.yml`.
- Updates the coder-eval tasks under `tests/tasks/uipath-maestro-bpmn/` to drive and grade with the CLI command instead of `node validate-bpmn.mjs`.

## Why

The CLI's `uip maestro bpmn validate` now runs the full PO.Frontend canvas rule set offline — the structural rules were ported into `@uipath/maestro-sdk` and wired into the command in **UiPath/cli#3135** (merged). That removes the reason the skill shipped its own validator, and it kills the drift risk of a second checked-in copy of the PO.Frontend rule engine + extension spec. The CLI is now the single source of truth.

## Dependency / sequencing

Requires a released `uip` that contains cli#3135. The skill's freshness note tells the agent to update the CLI if `validate` is missing or runs only deploy-readiness checks; the `integration`-tier eval (`diagnose/validate_fix_loop`) needs a runner CLI new enough to have the structural rules. **Hold merge until a CLI release includes cli#3135.**

This also supersedes the validate-reconciliation portion of the still-open #1691 (which framed the CLI validate as a deploy-readiness *complement* to the bundled validator — no longer accurate now that the CLI carries the structural rules).

## Test plan

**Static checks**
- `hooks/validate-skill-descriptions.sh` passes; `scripts/check-skill-status.py` → OK (24 skills).
- All 70 `tests/tasks/uipath-maestro-bpmn/**` YAMLs parse; no remaining references to the bundled validator anywhere in the repo.

**Eval suite (run locally against a CLI built from `main` with cli#3135)**

Built `uip` from the merged CLI and ran the coder-eval harness (`make install`) on the edited tasks via `experiments/default.yaml`:

| Task | Result | Verifies |
|---|---|---|
| `diagnose/validate_fix_loop` (integration) | 4/4, score 1.000 | Agent runs `uip maestro bpmn validate`; after repair the grading command returns exit 0 / `Valid` |
| `debug/debug_not_validation` | 4/4, score 1.000 | Edited `command_pattern` matches — agent validates locally via the CLI verb |
| `structural/error_boundary_handler` (authoring) | 4/4, score 1.000 | Agent authors a BPMN and validates it with `uip maestro bpmn validate` |

Raw command spot-checks against the #3135 CLI: the broken fixture → `Failure` / exit 1 with `MISSING_CONDITION_EXPRESSION` + `VARIABLE_DOES_NOT_EXIST`; a valid file → `Success` / exit 0.

**⚠️ Release gating (must-read before merge)**

**Self-guarding eval:** `validate_fix_loop` now has a `pre_run` assertion that the pristine broken fixture *fails* validation. On a CLI too old to have cli#3135 the fixture would validate clean, so the guard aborts the run with a clear "update the CLI" message instead of passing vacuously — the eval can no longer silently false-pass on an old CLI.

These evals pass only against a `uip` that contains cli#3135. The **currently-released** CLI predates it: I confirmed the eval's broken fixture validates as `Valid` / exit 0 on the released CLI, which breaks `validate_fix_loop`'s premise. The skills-repo CI runner installs `@uipath/cli@latest`, so **do not merge this PR until a `uip` release includes cli#3135** — otherwise the eval CI here will be unreliable.


nikhil-maryala (approved):

rockymadden (approved):
Looks good.

bai-uipath (approved):

PR #2192, opened 2026-07-22
fix(uipath-mcp-servers): improve availability gate eval

edis-uipath:
## Summary

- keep the availability-gate prompt focused on the user goal instead of assuming a production endpoint
- mock an authenticated `uip login status` response so environment discovery stays deterministic
- preserve the mocked `IsActivityNotAvailable` response and safety checks

## Validation

- `coder-eval plan tests/tasks/uipath-mcp-servers/is-activity-availability-gate/task.yaml -e tests/experiments/default.yaml`
- verified the shared mock dispatcher matches `login status` with exit code 0
- verified the candidates fixture returns `Reason: IsActivityNotAvailable` with exit code 1
- passing local coder-eval run `2026-07-22_15-26-02`: Codex `gpt-5.6-terra`, score 1.0, 5/5 criteria passed

### DIFF
--- a/tests/tasks/uipath-mcp-servers/is-activity-availability-gate/task.yaml
+++ b/tests/tasks/uipath-mcp-servers/is-activity-availability-gate/task.yaml
@@ -13,18 +13,15 @@ sandbox:
   - {type: template_dir, path: fixtures}
   mock_path_dirs: [mocks]
 initial_prompt: |
-  For an IS-activity tool, read references/is-activity-workflow.md BEFORE any uip call and use the exact `uip agenthub mcp-tools create-is-activity` flags it specifies (`--mcp`, `--target-identifier`, `--description`, `--metadata`, `--input-schema`, `--output-schema`); for a resource tool, read the resource guidance and run `mcp-tools candidates`/`template` as it directs. Use the exact CLI commands and flags the skill documents — do not invent flag names.
-
-  Use the `uip` CLI as if it were the real UiPath CLI (it is mocked in this
-  sandbox; manifest-driven canned responses).
-
   TASK: Add a Jira Create Issue tool to my `test-mcp-skill` MCP server in
   the Shared folder. The target is the Jira Integration Service activity
-  (connector uipath-atlassian-jira). I am running against my production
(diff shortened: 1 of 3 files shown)


tests/tasks/uipath-mcp-servers/is-activity-availability-gate/task.yaml
```diff
@@ -13,18 +13,15 @@ sandbox:
...
   TASK: Add a Jira Create Issue tool to my `test-mcp-skill` MCP server in
   the Shared folder. The target is the Jira Integration Service activity
-  (connector uipath-atlassian-jira). I am running against my production
-  tenant on cloud.uipath.com. There is no user available for follow-up
-  questions in this session — if you hit a blocker you cannot resolve
-  within the skill's rules, stop and write up the outcome instead of
-  working around it.
+  (connector uipath-atlassian-jira).
+
+  DO NOT assume I am running against my production tenant on cloud.uipath.com.
```
radu-mocanu, line 20: nit: this seems like a skill gap, we should not expect the user to specify this

edis-uipath: the test is testing explicitly what happens when the agent hits something that is not available in CLI 

in this case the skill text already specifies that the feature is only available in alpha and staging and it will not hit the cli if you tell them that the tenant is in cloud (becuase we do not have feature flags on skills)

there is no point to test what happens if the agent knows that this is not available, the point is to test what happens when it ignores that instruction and hits it anyways

radu-mocanu (approved):

tests/tasks/uipath-mcp-servers/is-activity-availability-gate/task.yaml
edis-uipath: this is the gate it goes around, so it will hit the CLI gate 
 
https://github.com/UiPath/skills/blob/b357b6a16acfdaa4631781ff38af0724b2cb2d18/skills/uipath-mcp-servers/references/is-activity-workflow.md?plain=1#L7

PR #2210, opened 2026-07-22
test(tasks): drop sandbox.driver tempdir overrides + gate against reintroduction

bai-uipath:
## What

Remove the `sandbox.driver: tempdir` override from all 190 task YAMLs that had it, and add a CI gate that fails the build if any task reintroduces it.

## Why

The sandbox driver is decided by the run environment, not by the task. The Linux nightly slice runs every non-windows task under `driver: docker` — the `skills-image` bakes the `uip` CLI and every tool plugin. The Windows slice selects tasks by the `windows` tag and forces `--driver tempdir` on the CLI (which wins over any YAML), so no task needs a YAML driver override.

A task that pinned `driver: tempdir` opted out of the docker image and ran on the bare host, where the tool plugins are not installed. Any tool-backed command then failed — e.g. `skill-flow-interactive-customer-escalation-triage`'s `uip maestro flow validate` needs `@uipath/maestro-tool`, which isn't on the host, so it failed with "No compatible version found for the CLI's 1.199.x line" and scored zero on otherwise-correct runs. The gate prevents that footgun from returning.

Existing `driver: docker` pins are left in place (they match the Linux default and are harmless); the gate reports them as an informational note rather than blocking.

## Validation

Ran the affected tasks on the docker driver (skills-image with the full baked toolchain: `uip` 1.199.0-dev.7959 + 31 tool plugins, coder-eval 0.8.8, Bedrock backend) to confirm they run cleanly in the container they were opting out of. Two cuts, **42 tasks across 9 of the affected categories**.

**Headline: every task executed in-container (`sandbox_path` under `/work/output`) — 0 ran on the host, and there were 0 tool-resolution / host-missing failures.** That is the failure class this PR removes, and it is gone.

### Cut 1 — 12 tasks, 12/12 passed (1.0)

| Task | Category | Score |
|---|---|---|
| `skill-flow-interactive-customer-escalation-triage` | maestro-flow *(the original bug)* | 1.0 ✅ |
| `skill-bpmn-smoke-registry-discovery` | maestro-bpmn *(tool-backed)* | 1.0 ✅ |
| `skill-agent-conversational-guardrail-custom-tool` | agents | 1.0 ✅ |
| `skill-ixp-change-field-type-smoke` / `list-documents` / `negative-guards` | ixp | 1.0 ✅ |
| `skill-mcp-servers-cross-folder-create` / `remote-create` | mcp | 1.0 ✅ |
| `skill-platform-solution-pack-determinism-integration` | solution | 1.0 ✅ |
| `uipath-coded-apps-dashboard-diagnose-broken-metric` / `edit-remove-widget` / `refuse-invocation-timeline` | coded-apps | 1.0 ✅ |

The two tool-backed tasks (`customer-escalation-triage`, `bpmn-smoke-registry-discovery`) are the direct proof: these are exactly the tasks that failed on the toolless host with "No compatible version of `@uipath/maestro-tool`", and both pass on the docker driver because the tool is baked into the image.

### Cut 2 — 30 tasks (troubleshoot + review), the two largest categories by count

- **troubleshoot: 15/15 passed** (10 at 1.0, 5 at 0.85–0.925 partial credit) — clean sweep.
- **review: 8/15 passed at 1.0**; the other 7 hit `MAX_TURNS_EXHAUSTED` (3) or a 1200s agent-turn timeout (4).

Those 7 review misses are **not** driver-related — they are pre-existing agent-loop behavior on long static-analysis tasks:
- All 7 ran in-container with commands succeeding (e.g. `skill-review-rpa-selector-brittle`: 60 assistant turns, 38 successful commands, ran the full 1202s before the wall-clock turn timeout). Nothing failed on a missing tool.
- Review tasks are static analysis (read an agent/RPA project, flag issues); they don't invoke the baked tool plugins that the tempdir removal restores, so the driver is orthogonal to their outcome.
- Turn-timeouts and max-turns are a suite-wide, pre-existing phenomenon — a single normal nightly (2026-06-23) logged 19 `MAX_TURNS_EXHAUSTED` and 20 turn-timeouts across the suite.

Net: the driver switch introduces no environment or tool-resolution regressions; the residual review flakiness is independent of this change and would occur on either driver.

### DIFF
--- a/.github/workflows/task-driver-gate.yml
+++ b/.github/workflows/task-driver-gate.yml
@@ -1,0 +1,40 @@
+name: Task Driver Gate
+
+# Fails the build if any coder-eval task YAML pins `sandbox.driver: tempdir`.
+#
+# The sandbox driver is chosen by the run environment, never by the task:
+#   - Linux nightly slice  -> `driver: docker` (skills-image bakes the uip CLI
+#     + every tool plugin), set in tests/experiments/nightly.yaml.
+#   - Windows slice        -> selected by the `windows` tag and forced onto
+#     `--driver tempdir` on the CLI (wins over YAML).
+# A task that pins `driver: tempdir` runs on the toolless host and breaks any
+# tool-backed command (e.g. `uip maestro flow validate`). See
+# scripts/check-task-driver.py for the full rationale.
+#
+# No `paths:` filter so this can be wired as a required check — it is a
+# sub-second full scan that always reports a definitive pass/fail.
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: task-driver-gate-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  gate:
+    runs-on: ubuntu-latest
+    name: No task pins sandbox.driver tempdir
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with: { python-version: '3.13' }
+      - name: Install PyYAML
+        run: pip install --quiet pyyaml
+      - name: Check task drivers
+        run: python3 scripts/check-task-driver.py tests/tasks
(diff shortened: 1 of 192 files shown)


rockymadden (approved):
Looks good to me and makes sense.

PR #2219, opened 2026-07-23
fix(guardrails): make Terra trajectory gates explicit

apetraru-uipath:
## What changed?

- Route a named Tool plus an exact mechanical predicate to a custom
  deterministic guardrail before built-in catalog ranking. Quoted text and a
  distinct all-caps token such as `CONFIDENTIAL` are explicitly treated as
  literals, not broadened into an LLM-based semantic category.
- Require the deterministic result to use the exact `resource.json` Tool name,
  Tool scope, the matching custom rule (`word/contains` for a literal), and a
  blocking action. Before validation, re-read `agent.json`; if a built-in was
  written, replace it with the required custom rule.
- Make the required coded SDK documentation fetch an explicit first-step
  completion gate.
- Make escalation lookup unconditional: after `resources list`, both the
  exact-match and no-match branches must attempt
  `uip solution resources get` before editing, rejecting, or responding.
- Require low-code scaffolding before escalation lookup, and require
  `uip agent validate` after refresh before reporting conversational work
  complete.
- Make the three coded guardrail documentation criteria agent-portable: they
  still require the exact UiPath SDK guardrail URL, but accept either Claude's
  `WebFetch` or Codex's Bash-based fetch.
- Limit changes to `uipath-agents` guardrail references and three coded
  guardrail task definitions. No non-guardrail skill or workflow is changed.

## How has this been tested?

- `94 passed` — `pytest tests/scripts -q`
- All six targeted task definitions resolve with `coder-eval plan` under
  `tests/experiments/smoke.yaml`.
- Codex `gpt-5.6-terra` on the exact PR head (`2f9ec36e5`):
  [6/6 passed with every weighted and individual criterion score at
  1.0](https://github.com/UiPath/skills/actions/runs/30071448969).
  These are the six non-Maestro guardrail failures from the official
  `2026-07-23_04-33-01` daily Codex baseline.
- The literal custom-rule branch also
  [passed in isolation](https://github.com/UiPath/skills/actions/runs/30071217405)
  on the same head, producing `$guardrailType: "custom"`, Tool scope, the
  exact `Send message to channel` name, `word/contains`, and
  `value: "CONFIDENTIAL"`.
- PR smoke uses Claude Sonnet 5. On the exact head, all three guardrail smoke
  tasks passed at `1.0` (including escalation), and the reviewer threshold
  passed. The broad
  [26-task aggregate](https://github.com/UiPath/skills/actions/runs/30071209972)
  is 23/26 because three unrelated tasks failed by timeout/max-turn exhaustion:
  coded push/pull, bindings sync, and coded Integration Service.
- The Task Driver Gate is green after rebasing onto current `main`.
- `git diff origin/main --check` passes.

- [x] None
- [ ] Yes

### DIFF
--- a/tests/tasks/uipath-agents/coded/guardrails/recommend_all/recommend_all.yaml
+++ b/tests/tasks/uipath-agents/coded/guardrails/recommend_all/recommend_all.yaml
@@ -47,7 +47,6 @@ success_criteria:
 
   - type: command_executed
     description: "Agent fetched the UiPath Python SDK guardrail docs"
-    tool_name: "WebFetch"
     command_pattern: 'uipath\.github\.io/uipath-python/.*guardrails'
     min_count: 1
     weight: 1.5
--- a/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails-recommend.md
+++ b/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails-recommend.md
@@ -14,6 +14,11 @@ Both workflows are driven by live data — the catalog (`uip agent guardrails ca
 
 > Full three-fetch mandate applies to **Recommend mode**. In **Validate mode** of an existing guardrail the SDK docs are the authoritative, sufficient source for a validator's scope/stage — `catalog` (relevance metadata) and `list` (tenant entitlement) are recommended cross-checks, not a hard prerequisite for a scope/placement fix. See [Validate Mode](#validate-mode).
 
+**Required first operation in both modes:** use `WebFetch` on
+`https://uipath.github.io/uipath-python/core/guardrails/` before catalog calls,
+project inspection, analysis, or edits. A coded guardrail recommendation or
+validation is not grounded until this WebFetch has completed.
+
 ### Catalog (cacheable — 30-minute TTL)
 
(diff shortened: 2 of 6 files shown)


tests/tasks/uipath-agents/coded/guardrails/recommend_all/recommend_all.yaml
```diff
@@ -47,7 +47,6 @@ success_criteria:
 
   - type: command_executed
     description: "Agent fetched the UiPath Python SDK guardrail docs"
-    tool_name: "WebFetch"
     command_pattern: 'uipath\.github\.io/uipath-python/.*guardrails'
     min_count: 1
```
apetraru-uipath, line 51: I removed tool_name: "WebFetch" because it made the documentation-fetch criterion Claude-specific.
Claude Sonnet fetches the SDK page with WebFetch, while Codex/Terra typically uses Bash with curl. The exact UiPath SDK URL remains mandatory through command_pattern, so the test still verifies the required documentation step—it just accepts either transport.
Without this change, Codex could fetch the correct page and perform the task correctly but still receive a false failure solely because it didn’t use Claude’s tool name. The tradeoff is that any tool matching the exact URL is accepted, which is intentional for cross-agent tests.

valentinabojan (approved):

PR #2227, opened 2026-07-23
ci(smoke): authenticate guardrail CLI against Alpha

apetraru-uipath:
## What changed?

- Replace the PR smoke workflow's client-credentials CLI auth with the existing Alpha bot-user ROPC helper already used by the nightly workflow.
- Mount the standard UiPath auth file into the smoke image.
- Fail setup early unless both `uip login status` and `uip agent guardrails list` succeed.

This PR changes only `.github/workflows/smoke-skills.yml`. It does not modify tests, task definitions, guardrail skills, review skills, docs, or Maestro files.

## Why?

The previous service-account token could authenticate successfully but lacked the user-only access required by the `agentsruntime_` guardrail definitions endpoint, causing guardrail catalog/list calls to return 401. The Alpha bot-user flow includes the required user scopes and was validated by the live pipeline on #2219.

## How has this been tested?

- Workflow YAML parses successfully.
- `git diff --check` passes.
- Existing `tests/scripts` baseline: 94 passed.
- The workflow hunk exactly matches the Alpha auth/preflight change validated on #2219.

- [x] None
- [ ] Yes

### DIFF
--- a/.github/workflows/smoke-skills.yml
+++ b/.github/workflows/smoke-skills.yml
@@ -359,77 +359,39 @@ jobs:
             -f tests/docker/Dockerfile \
             .
 
-      # Pre-auth uip via the CLI's documented env-var bypass so non-RPA
-      # smoke tasks (orchestrator / integration-service / data-fabric / ...)
-      # have a logged-in CLI without the agent ever calling `uip login`.
-      # See smoke-rpa-skills.yml for the chain rationale; this Linux job
-      # doesn't go through Helm, but the same UIPATH_CLI_* env vars
-      # satisfy the JS CLI's general auth state too.
-      - name: Mint UiPath service-account token + enable env-auth
+      # Authenticate as the existing Alpha bot user. Agent guardrail discovery
+      # is served by agentsruntime_ and requires user-only scopes such as
+      # StudioWebBackend; an External App client_credentials token can report a
+      # valid login while still receiving 401 from the guardrail endpoints.
+      #
+      # Reuse the same ROPC helper as the nightly workflow. It writes the
+      # standard ~/.uipath/.auth file that the smoke experiment mounts into
+      # every Docker sandbox.
+      - name: Mint UiPath Alpha bot token (ROPC)
         env:
-          UIPATH_CLIENT_ID:     ${{ secrets.UIPATH_CLIENT_ID }}
-          UIPATH_CLIENT_SECRET: ${{ secrets.UIPATH_CLIENT_SECRET }}
+          UIPATH_URL:               https://alpha.uipath.com
+          UIPATH_ORGANIZATION_NAME: ${{ secrets.UIPATH_ORG_NAME }}
+          UIPATH_ORGANIZATION_ID:   ${{ secrets.UIPATH_ORG_ID }}
+          UIPATH_TENANT_NAME:       ${{ secrets.UIPATH_TENANT_NAME }}
+          UIPATH_TENANT_ID:         ${{ secrets.UIPATH_TENANT_ID }}
+          CLIENT_ID:                ${{ secrets.UIPATH_ROPC_CLIENT_ID }}
+          CLIENT_SECRET:            ${{ secrets.UIPATH_ROPC_CLIENT_SECRET }}
+          CE_USERNAME:              ${{ secrets.UIPATH_BOT_USERNAME }}
+          CE_PASSWORD:              ${{ secrets.UIPATH_BOT_PASSWORD }}
         run: |
           set -euo pipefail
-          TOKEN=$(curl -fsS -X POST \
-            "https://alpha.uipath.com/identity_/connect/token" \
-            -H "Content-Type: application/x-www-form-urlencoded" \
-            -d "grant_type=client_credentials" \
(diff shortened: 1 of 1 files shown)


akshaylive (approved):

uipreliga:
## Code review — multi-model (Gemini 3 · GPT-5.3-Codex · Opus)

Reviewed the auth-strategy swap in `smoke-skills.yml` (client_credentials `UIPATH_CLI_*` env auth → ROPC bot login via `refresh-auth.sh` + `.auth` mount + fail-fast `docker run`). Findings below are synthesized across three reviewers and independently verified. **Comment-only — nothing changed.**

### 🔴 Blockers (secret hygiene — both cheap to fix)

**1. Minted token is no longer masked, and the fail-fast prints CLI auth state to the run log**
`smoke-skills.yml:381-394` — The old step did `echo "::add-mask::$TOKEN"`. The new path mints the token inside `refresh-auth.sh` with **no** `::add-mask::` anywhere, so the token is unmasked for the whole job. The fail-fast then runs `uip login status --output json` and `uip agent guardrails list --output json` straight to the console log. If `login status` emits the token it leaks in clear text; even if it doesn't today, losing the mask is a hygiene regression.
→ `::add-mask::` the `UIPATH_ACCESS_TOKEN`/`UIPATH_REFRESH_TOKEN` after `refresh-auth.sh`, and/or redirect the fail-fast commands to `/dev/null` (they're run for exit code, not output).

**2. Report-redaction step scrubs a var that no longer exists, and misses the one that does**
`smoke-skills.yml:481-487` — The `names` tuple keys on `UIPATH_CLI_AUTH_TOKEN`, which this PR stops setting (filtered out by the `len(...) >= 8` guard → dead no-op). The real credential now lives in the mounted `~/.uipath/.auth` as `UIPATH_ACCESS_TOKEN`/`UIPATH_REFRESH_TOKEN` (`:rw` into every sandbox). Those names are **not** redacted, so if a task `cat`s `/.uipath/.auth` or a `uip` error echoes the token, it ships unredacted in the uploaded `task.log`/`task.json`/`*.html` artifacts (14-day retention).
→ Add `UIPATH_ACCESS_TOKEN`/`UIPATH_REFRESH_TOKEN` (sourced from `~/.uipath/.auth`) to the `names` tuple; drop the dead `UIPATH_CLI_AUTH_TOKEN` entry.

### 🟡 Low — cleanup / follow-ups

- **Dead passthrough config** — `tests/experiments/smoke.yaml:20-27` still lists `UIPATH_CLI_ENABLE_ENV_AUTH`, `UIPATH_CLI_AUTH_TOKEN`, `UIPATH_CLI_ORGANIZATION_*`, `UIPATH_CLI_TENANT_*` under `env_passthrough_extra`. The workflow no longer sets any of them → forwards nothing. Harmless no-op; prune for clarity (or note deferred).
- **`TRACES_SMOKE_PROCESS_KEY` no longer exported to `$GITHUB_ENV`** — **no active regression** (both consumers `traces_e2e.yaml:7` / `traces_feedback_e2e.yaml:8` are `e2e`-tagged; smoke runs `--tags smoke`). Latent only if those tasks are ever run here. Consider a one-line comment or re-adding the export.
- **Fail-fast relies on implicit image `HOME=/`** — `smoke-skills.yml:387-394`, mount target is absolute `/.uipath`, so `uip` finds `/.uipath/.auth` only if `HOME=/`. Matches the proven nightly mount, so likely fine; worth confirming `HOME` is baked into the `coder-eval-agent` image rather than set at sandbox launch.
- **Stale task prose** — `lowcode_review_guardrail_action_ineffective.yaml:13`, `lowcode_review_guardrail_unknown_validator.yaml:10` still say the runner passes `UIPATH_CLI_*` auth; now delivered via the `.auth` mount. Docs only.

### ✅ False positive discarded
- "Redaction step raises `KeyError` on the unset `UIPATH_CLI_AUTH_TOKEN`" — incorrect. `values = [os.environ[name] for name in names if len(os.environ.get(name, "")) >= 8]`: the `.get()`-length guard evaluates first, so the unset var is filtered before `os.environ[name]` is ever indexed. No crash.

### 👍 Positives (all three reviewers)
- Genuine simplification (77→39 lines): removed the `run_uip_with_tenant` wrapper, the tenant-list round trip, and the embedded Python.
- Fail-fast exercises the **real** auth path (same `~/.uipath:/.uipath:rw` mount tasks use), so a green check reflects what tasks see.
- Correct scope rationale — guardrail discovery needs user scopes (`StudioWebBackend`) a client_credentials token lacks.
- DRY with nightly (`run-coder-eval.yml`) via the shared `refresh-auth.sh`; atomic auth-file write (`umask 077` + temp + `mv -f`); version-sync safe (`UIPATH_CLI_DISABLE_VERSION_SYNC=1` baked in the image).

### Verdict
Solid, well-motivated refactor that fixes the guardrail 401s and simplifies the step. Recommend addressing the two secret-hygiene blockers before merge; the rest are low-priority cleanups.

<sub>🤖 Multi-model review via `/code-review` (Gemini 3, GPT-5.3-Codex, Opus 4.8).</sub>

PR #1694, opened 2026-06-26
test(uipath-agents): cover low-code eval CLI + attachments/datafabric context

mjnovice:
## Why

Two test-coverage gaps in the **uipath-agents** skill (low-code side), surfaced by `/test-coverage`:

1. **Low-code eval subsystem (High).** The `uip agent eval` engine is separate from `uip codedagent eval`. Coded had 6 `eval_*` tasks; low-code had **zero** in `main`.
2. **Context variants (Medium).** Only the `index` context variant was tested; `attachments` and `datafabricentityset` were untested.

> The guardrail-variant gap from the same report (`user_prompt_attacks`, `intellectual_property`, custom number/boolean rules, log/filter actions) was **already closed** by #1684 (AL-487), so no work there.

## What

| Task | Tier | Runs offline? | Covers |
|---|---|---|---|
| `lowcode/eval_authoring` | smoke | ✅ | `uip agent eval evaluator add` + `eval add` + `validate`; deterministic evaluator wired with ≥2 test cases |
| `lowcode/eval_run` | e2e | ❌ (live tenant) | `solution upload` → `eval run start --wait` → results export |
| `lowcode/context_attachments` | smoke | ✅ | `contextType: "attachments"` resource + validate |
| `lowcode/context_datafabric` | smoke | ✅ | `contextType: "datafabricentityset"` resource (entitySet shape, no `indexName`/`settings`) + validate |

Each task has a YAML + a Python check script.

## Notes

- A prior unmerged attempt (`eval_run_lifecycle`, PR #1183) over-reached on the cloud lifecycle (push + 2 runs + status + list + compare + 7 self-report files) and **failed at 0.39**, and used the now-stale `uip agent push`. These tasks use the current `uip solution upload` path the skill teaches, and keep the single e2e narrow.
- All four lint clean (`0 Critical/High/Medium`); every `uip` verb resolves against the catalog; check scripts verified against positive **and** negative fixtures.

## Validation

- ✅ Lint, verb-reachability, YAML parse, check-script compile + positive/negative fixtures (local).
- ⏳ `coder-eval` run-claims: the three offline tasks will run via `smoke-skills.yml` on this PR. The `eval_run` e2e needs a live tenant — recommend one manual run to confirm `eval run start --wait` completes before relying on it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2233, opened 2026-07-23
fix(tests): drop tempdir driver pins from classic-invoke troubleshoot tasks

mjnovice:
## Why

Every open PR currently fails the **Task Driver Gate** (`No task pins sandbox.driver tempdir`). #2104 added the six `classic-invoke-*` troubleshoot tasks with `sandbox.driver: tempdir` while #2210 — which removed all such pins and introduced the gate — was in flight. The two crossed without conflicting, and since the gate only runs on `pull_request`, main itself never surfaced the breakage.

## What

Deletes the `driver: tempdir` line from the six task YAMLs, exactly as #2210 did for every other task. All six are Linux tasks (no `windows` tag, mock `template_dir` + python sandbox), so per `scripts/check-task-driver.py` the driver must come from the run environment, not the YAML.

Verified locally:

```
$ python3 scripts/check-task-driver.py tests/tasks
OK — no task pins `sandbox.driver: tempdir`.
```

First observed on PR #1694 after updating its branch with main.

### DIFF
--- a/tests/tasks/uipath-troubleshoot/activity-packages/classic-invoke-arg-mismatch/task.yaml
+++ b/tests/tasks/uipath-troubleshoot/activity-packages/classic-invoke-arg-mismatch/task.yaml
@@ -16,7 +16,6 @@ run_limits:
   turn_timeout: 3600
 
 sandbox:
-  driver: tempdir
   python: {}
   template_sources:
     - type: template_dir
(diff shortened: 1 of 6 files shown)


bai-uipath (approved):
lgtm 🙏

MarinRzv (approved):

PR #2247, opened 2026-07-24
test(troubleshoot): drop project-directory pointer from task prompts

MarinRzv:
## What

Removes the project-directory pointer — "the project source is in the current working directory" (and the bundled "…from Orchestrator and the project source" variants) — from **174** `uipath-troubleshoot` task prompts. Where a prompt named the location, that fact now lives in the task `simulation` and is provided only when the agent asks (a new constraint; any conflicting "never invent data" constraint was amended to exempt it).

## Why

In a full suite run, `uipath-troubleshoot` failed to activate on ~34% of tasks (WARN: correct diagnosis, but the skill never fired). Root cause: the prompt handed the agent the project location up front, so its first step was to read local files directly instead of invoking the skill. The agent can't see its turn budget, so that first-step decision is decisive.

Bare/generic diagnostic prompts (already in the activation set) activate ~100%; the specific-symptom prompts that pointed at the cwd are the ones that diverged.

## Validation

Re-ran a previously non-activating task under its real config (populated sandbox, mock `uip`, `max_turns: 60`, simulation):

| `csv-append-column-mismatch` | Skill activated | Judge | Result |
|---|:---:|---:|:---:|
| Before (prompt had the pointer) | ❌ no | 1.00 | WARN |
| After (this branch) | ✅ yes | 1.00 | PASS |

## Scope

- Only `tests/tasks/uipath-troubleshoot/**/task.yaml` (prompt + simulation). No skill sources, references, mocks, tags, success criteria, or run limits changed.
- Activation-dataset experiments are intentionally **not** in this branch.


costin-uipath (approved):
Will this cover future generated scenarios? Can we do that by ammending generate_scenario.py?

PR #2268, opened 2026-07-24
fix(governance): correct disambiguation judge target + surface catalog-get verb

sauravrathore01:
Claude Run: https://github.com/UiPath/skills/actions/runs/30198981849
Codex Run: https://github.com/UiPath/skills/actions/runs/30206116172
Antigravity Run: https://github.com/UiPath/skills/actions/runs/30211284466

## What

Two fixes from investigating `uipath-governance` compliance smoke-test failures (`skill-governance-compliance-disambiguation`, `skill-governance-compliance-query-clause`).

### 1. `disambiguation_smoke.yaml` — judge graded the wrong source (false failure)

The prompt directs the agent's response into `output.txt`, but the `llm_judge` criterion read only the agent's chat output (`include_agent_output: true`, `files: []`). The agent actually behaved **correctly** — ran 0 governance mutations (all three `command_not_executed` checks passed) and wrote a proper numbered disambiguation question to `output.txt` — but the judge only saw Codex's plan-narration preamble (`"I'll apply a UiPath governance policy..."`) and scored 0.0, dragging the weighted score to 0.727 → FAILURE.

**Fix:** point the judge at the actual deliverable (`files: ["output.txt"]`) and drop the misleading chat narration (`include_agent_output: false`). The `command_not_executed` checks remain the deterministic no-mutation guardrail.

### 2. `SKILL.md` — surface the canonical `catalog get` verb inline (real failure)

For `query-clause`, the agent read `SKILL.md`, never opened the linked `query/impl.md`, guessed non-existent verbs (`uip gov compliance`, `compliance-standard`, `standards`), and answered from memory — so the required `uip gov compliance-packs catalog get` never ran (criterion 0.0 → FAILURE). The skill and test are both correct; the agent stopped at `SKILL.md`.

**Fix:** inline the canonical command (with the `iso-42001-2023` pack-ID example) into the compliance query routing-table row, so a weak-navigation agent sees the verb in the file it demonstrably reads without a second file-open.

## Verification

- Grader regex `uip\s+gov\s+compliance-packs\s+catalog\s+get\b` now matches text physically present in `SKILL.md`. ✅
- `hooks/validate-skill-descriptions.sh` passes (exit 0). ✅
- `/lint-task` on `disambiguation_smoke.yaml`: verdict OK. ✅
- **Not yet run:** the live eval re-run (needs `CODEX_API_KEY`) — the model-behavior confirmation still has to happen in CI/smoke. Fix 1 is deterministically correct; fix 2 is a mitigation whose pass depends on the model now running the surfaced command.

Generated with Claude Code

### DIFF
--- a/tests/tasks/uipath-governance/compliance-pack/full_apply_e2e_live.yaml
+++ b/tests/tasks/uipath-governance/compliance-pack/full_apply_e2e_live.yaml
@@ -46,26 +46,31 @@ success_criteria:
     weight: 1.0
     pass_threshold: 1.0
 
-  # Advisory, not gating. command_executed matches the literal top-level command
-  # string; an agent that batches its CLI calls into a script file and runs
-  # `bash run.sh` executes state coverage/enable correctly but exposes only
-  # `bash run.sh` to the matcher, so the string-match false-fails a correct run.
-  # The real invariant — the pack actually ends up enabled — is gated below by
-  # the weight-5.0 post-hoc `state get` outcome check (and report.json content),
-  # which passes only if these commands genuinely ran against the tenant.
-  - type: command_executed
-    description: "Agent called state coverage to analyze posture (advisory — outcome gated by post-hoc check)"
-    command_pattern: 'uip\s+gov\s+compliance-packs\s+state\s+coverage\b'
-    min_count: 1
+  # Gate on the CLI RESPONSE captured in output.txt, not the command string.
+  # command_executed matches only the literal top-level command, so an agent that
+  # batches its calls into a script (`bash run.sh`) false-fails even when the
+  # commands ran. But grading solely the end-state (pack enabled) would false-PASS
+  # if the pack were already enabled by something else on the shared tenant (a
+  # concurrent disable_smoke pre_run, a racing cleanup). The CLI echoes a per-call
+  # response envelope (`"Code": "CompliancePacksState<Verb>"`) into output.txt via
+  # the agent's `> output.txt` redirect — present regardless of script-nesting, and
+  # written ONLY when THIS agent actually ran the call. So it proves the command
+  # ran this run without depending on how it was invoked or on prior tenant state.
+  - type: file_contains
+    description: "state coverage ran this run (CLI response captured in output.txt — survives script-nesting, not satisfied by a pre-enabled pack)"
+    path: "output.txt"
+    includes:
+      - "CompliancePacksStateCoverage"
     weight: 3.0
-    pass_threshold: 0
+    pass_threshold: 1.0
 
-  - type: command_executed
-    description: "Agent called state enable to apply missing settings (advisory — outcome gated by post-hoc check)"
-    command_pattern: 'uip\s+gov\s+compliance-packs\s+state\s+enable\b'
-    min_count: 1
+  - type: file_contains
+    description: "state enable ran this run (CLI response captured in output.txt — survives script-nesting, not satisfied by a pre-enabled pack)"
+    path: "output.txt"
+    includes:
+      - "CompliancePacksStateEnable"
     weight: 3.0
-    pass_threshold: 0
+    pass_threshold: 1.0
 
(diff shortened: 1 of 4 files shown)


tests/tasks/uipath-governance/compliance-pack/full_apply_e2e_live.yaml
```diff
@@ -46,19 +46,26 @@ success_criteria:
...
 
+  # Advisory, not gating. command_executed matches the literal top-level command
+  # string; an agent that batches its CLI calls into a script file and runs
+  # `bash run.sh` executes state coverage/enable correctly but exposes only
+  # `bash run.sh` to the matcher, so the string-match false-fails a correct run.
+  # The real invariant — the pack actually ends up enabled — is gated below by
+  # the weight-5.0 post-hoc `state get` outcome check (and report.json content),
+  # which passes only if these commands genuinely ran against the tenant.
   - type: command_executed
-    description: "Agent called state coverage to analyze posture"
```
abhinavmahanti, line 50: What if the pack is in enabled state itself (due to disable pack pre run failing or disable_smoke enabling it concurrently on the same tenant)? If we remove the coverage and enable call checks and just rely on the outcome, it can lead to a false positive (we are checking if the pack is enabled at the end or not with this change, not that the agent enabled this pack or not)

sauravrathore01: Good catch — you're right that grading pure end-state ("pack enabled") false-positives when the pack is already enabled by something else on the shared tenant (a concurrent `disable_smoke` pre_run, a racing cleanup).

Fixed in 8688f1fc7: instead of demoting to advisory + relying on the outcome check, the coverage/enable criteria now gate on the **CLI response captured in `output.txt`** — `"Code": "CompliancePacksStateCoverage"` / `...StateEnable`. That marker:
- is written **only when this agent actually ran the call** (it's this task's own captured stdout), so a pre-enabled pack can't satisfy it — `setup_enable_compliance_pack.py` never writes this task's `output.txt`;
- appears **regardless of script-nesting** (survives the `bash run.sh` batching that caused the original false-negative), since it's the response body, not the command string.

So it grades "the agent enabled the pack **this run**", not "the pack happens to be enabled at the end". The weight-5.0 post-hoc `state get` outcome check stays as the complementary end-state gate. PTAL.

abhinavmahanti (approved):

tests/tasks/uipath-governance/compliance-pack/full_apply_e2e_live.yaml
sauravrathore01: Follow-up — the latest CI run proved your concern applies in *both* directions, so I've hardened this further (54d113ecd).

Run 30260025266 on the shared tenant: the agent correctly ran coverage + enable (its captured response shows `Result: Success`, `Active: true` at 11:04:49), but a **sibling compliance task's `post_run` cleanup** — which disables iso-42001 — toggled the pack off ~17s later (`LastToggledAt: 11:05:06`). So the live post-hoc `state get` re-read `Active: false` and false-**failed** a correct run. Same shared-tenant race you flagged, just the disable direction.

So the live re-read is racy both ways (pre-enabled → false pass; concurrently-disabled → false fail). Fix:
- **Gate** on the agent's own captured `CompliancePacksStateEnable` + `CompliancePacksStateCoverage` responses in `output.txt` — written only on a real successful call *this run*, immune to later concurrent toggles, and not satisfiable by prior tenant state.
- **Demote** the live post-hoc `state get` to advisory (`pass_threshold: 0`) — kept as a best-effort signal that lands green when no cleanup races.

Verified the exact failing run re-scores to SUCCESS under the new config. Thanks for pushing on this — the outcome check was the wrong instrument on a shared, parallel-mutated tenant.

IsabellaCapriottiUIPath (approved):

PR #2269, opened 2026-07-24
fix(uipath-agents): rebuild deploy_folder_and_rebump fixture as a real LangGraph agent; target tenant feed

andreibalas-uipath:
## Why

The 2026-07-24 nightly failed `skill-agent-coded-deploy-folder-and-rebump` at 0.625 for two independent reasons:

1. **Fixture misclassification.** The `acme-echo` fixture was a plain function-map project (`uipath.json` `functions` map, no framework config, no LLM) — exactly the shape both skill descriptions route to `uipath-functions` ("deterministic Python Coded Functions … → uipath-functions"). The agent followed that redirect and deployed via `uip functions publish`, never running the graded `uip codedagent deploy`. The CLI itself classifies this shape as `ProjectType: Function`, so the ambiguity is real down to the runtime.
2. **Folder-feed mismatch.** `codedagent deploy --folder <name>` matches **package feed names**, not Orchestrator folder paths (`uipath/_cli/cli_publish.py`), and the run tenant's `Shared` folder has no dedicated feed — the graded folder deploy could never succeed cloud-side and always died at feed lookup with `Folder 'Shared' not found`.

## What

**Fixture** — replaced with a scratch-built LangGraph coded agent: `langgraph.json`, `StateGraph` echo node calling `UiPathChatAnthropicBedrock`, `GraphInput`/`GraphOutput` Pydantic models, `project.uiproj` with `ProjectType: "Agent"`, plus the full `uipath init` output (`entry-points.json`, `agent.mermaid`) so the project satisfies the deploy prerequisites and matches the prompt's "already set up locally" premise. Local state excluded (`.env`, `.agent/`, `uv.lock`, `evaluations/`, `.uipath/`).

**Task** — prompt and criterion now target the tenant package feed (`--tenant` / `-t`) instead of the feed-less `Shared` folder; the task description documents the routing rationale.

**Known follow-up (deliberately out of scope):** after a successful run the tenant feed holds `acme-echo 0.0.2`, so subsequent uploads 409 (`Package already exists`). The prompt tolerates cloud errors and the criterion grades the deploy attempt, so the task stays green. True cleanup needs a raw Orchestrator OData call — there is no `uip` verb to delete a tenant-feed package version (`or packages` has no `delete`; `solution packages delete` is solution-feed only) — so per review decision we leave it as-is rather than introduce the repo's first REST-based cleanup script.

## Passing runs

Local `coder-eval run … -T codex -m gpt-5.6-terra` (2026-07-24, tempdir driver):
- LangGraph fixture + `--folder` (pre-retarget): **1.000 (2/2)**, 66.6s, no drift to `uip functions`, no init recovery
- LangGraph fixture + `--tenant` (final): **1.000 (2/2)**, 45.7s, single deploy attempt, package genuinely published to the tenant feed
- 409 tolerance re-run: in flight, will be posted as a PR comment


andreibalas-uipath:
**409-tolerance re-run confirmed** (local, codex/`gpt-5.6-terra`, `runs/deploy-rebump-409`): with `acme-echo 0.0.2` already in the tenant feed from the previous run, the task still scores **1.000 (2/2)** in 53.7s. The deploy packaged successfully, Orchestrator returned `409 Package already exists` (errorCode 1004), and the agent moved on per the prompt's cloud-error tolerance. So repeated runs against the same tenant stay green — the 409 only means the upload itself is a no-op until the version changes or the package is removed (no CLI verb exists for that; deliberately out of scope per PR description).

radu-mocanu (approved):

PR #2275, opened 2026-07-24
fix(maestro-flow): doc fixes proven by the DevCon billing task sweep

jiyangzh:
## TL;DR

Four agent-authored `.flow` defects pass `flow validate` and only fault at `flow debug`. Docs now pin all four contracts; each was proven by a one-change experiment (fail→pass) and re-verified by fresh unaided builds scoring **1.000** on the DevCon `billing_dispute_resolution` e2e (2026-07-24 and 2026-07-27).

1. Grounded inline agent needs a retrieval cap + decide-anyway fallback in its prompt (`TERMINATION_MAX_ITERATIONS`)
2. Never author `outputs.output` on Orchestrator-job nodes — declare `error` only, the converter injects `output` (`=this`); an invented `source` is null downstream
3. Agent input `type` must match the bound node's real shape — Data Service queries return `array` (`AGENT_STARTUP.INPUT_VALIDATION_ERROR`)
4. Hand-written CEQL: field names bare, only values quoted (IS 400 → incident `102003`)

Every claim verified against flow-converter source, live registry, Studio Web serialization, or a fault record — details in the comments.

## Summary

Three authoring gaps let a coding agent build a `.flow` that passes `uip maestro flow validate` and then faults under `uip maestro flow debug`. All three surface as a bare flow incident `170002`, so each doc note now names the underlying fault code.

Found by running the DevCon `billing_dispute_resolution` e2e task: **three runs stuck at 0.500**, each dying at a *different* node, then **1.000** once the docs were corrected.

## What changed (docs only — no frontmatter, no code)

1. **Grounded inline agents looped on retrieval** until `AGENT_RUNTIME.TERMINATION_MAX_ITERATIONS`. The stop-condition rule lived two link hops away behind a 17-line hub file, and the authoring agent never reached it (its transcript had `agent-prompting-guide` ×10 but `Tool-call criteria` ×0).
   - `inline-agent/impl.md` step 3 now links `autonomous-agent-prompting-guide.md#1-system-prompt-skeleton` directly and inlines the two mandatory sentences (call cap + decide-anyway).
   - `maxIterations` guidance corrected: keep the scaffold default 25 whenever the agent holds a tool or context handle; `≤5` only for a tool-less single-shot agent. It is a kill switch, not a loop fix — the same prompt died at 5 and again at 25.
   - `context/index.md` gains the matching gotcha; the prompting guide's skeleton gains `Call at most <N> times` plus a never-end-without-an-answer line.

2. **The api-workflow node template shipped a dead output binding.** It prescribed `outputs.output` with `source: "=result.response"`. That node type's registry `outputDefinition` declares **only `error`**, and flow-converter's `applyOrchestratorActivityOutputs` injects the working output (`source: "=this"`) *only when the instance omits it*. Hand-authoring it therefore made `$vars.<node>.output` null at runtime while validate stayed green; a downstream agent binding it as a required object died with `AGENT_STARTUP.INPUT_VALIDATION_ERROR`.
   - Template now declares `error` only, with the mechanism explained and a Debug-table row for the null symptom.
   - `planning-impl.md`: author only the output keys the registry declares.

3. **Nothing pinned agent input types to the bound node's real shape.** Data Service query-entity-records nodes return an **array**; typing that key `object` faults `AGENT_STARTUP.INPUT_VALIDATION_ERROR` before the model runs, and neither `flow validate` nor `agent validate` can catch it. Added to `inline-agent/impl.md` anti-patterns, plus: mark a key `required` only when the binding can never be empty.

## Verification

Each fix was first proven by a one-change experiment on the failing artifact (patch → the same flow passes), then end-to-end: with these docs the agent authored **all three correctly, unaided, first try**.

| Run | Score | Died at | Fault |
|---|---|---|---|
| 1 | 0.500 | analyst, mid-run | `TERMINATION_MAX_ITERATIONS` |
| 2 | 0.500 | writer, startup | api-workflow output null |
| 3 | 0.500 | analyst, startup | array typed as `object` |
| 4 | **1.000** | — | 3/3 criteria passed |

Run 4 (`runs/2026-07-24_14-06-25`) audited span by span: 17 nodes, real tenant data (1 CRM row, 8 ERP rows), 4 real SOP-index retrievals, real API-workflow round-trip, all five returned out variables `$vars`-bound to agent outputs — no literals, no hardcoded verdicts in any script.

Known unchanged gap: IxP extraction returns 0 fields on the headless debug lane (documented in `check_e2e_billing_dispute_resolution.py:44`), identical in all four runs.


jiyangzh:
## Follow-up commit `0e3d1c0e2` — reviewer findings addressed

An independent review pass checked every claim above against primary sources (registry, flow-converter, run artifacts). Three of my statements were wrong or overreaching, and the defect turned out to be broader than api-workflow. All fixed in `0e3d1c0e2`, each verified per node family with `uip maestro flow registry get`.

### Registry ground truth

| Family | serviceType | `outputDefinition` | `output.source` | `error.source` |
|---|---|---|---|---|
| api-workflow | `ExecuteApiWorkflowAsync` | **varies** per published workflow | `=this` when declared | `=Error` |
| rpa-workflow | `StartJob` | `error` only | injected `=this` | `=Error` |
| agent | `StartAgentJob` | `output` + `error` | `=this` | `=Error` |
| agentic-process | `StartAgenticProcess` | `output` + `error` | `=this` | `=Error` |
| `core.subflow` | — | `error` only | injected `=this` | `=Error` |
| connector / `core.action.script` / `core.action.transform` | `Intsvc.ActivityExecution` etc. | `output` + `error` | **`=result.response`** — correct here | `=Error` |

### Corrections to the original commit

1. **"This node type declares no `output`" was wrong.** It varies: `07d45a91` and `c09cf5f6` declare `error` only, while `84e2a2f6`, `8d578fb1` and `be88b8d6` declare `output` with `source: "=this"`. The rule is now "mirror the registry for the node type you are wiring", not "never author `output`".
2. **A fallback I added recommended the failing path.** "Read `outputDefinition` when unsure of a node's shape" — for `query-entity-records` the registry says `output.type: "object"` with no schema even though the activity returns an array, i.e. it produces exactly the `object` typing that faults. Replaced with: bind the leaf, or confirm the shape from one debug run.
3. **`"=result.Error"` was dead too** — the converter's `ERROR_OUTPUT_DEFINITION` uses `"=Error"` and `result.Error` appears 0 times in it. Swept across every flow reference that carried it.

### Same defect in sibling docs (now fixed)

`plugins/rpa/impl.md`, `plugins/agent/impl.md`, `plugins/agentic-process/impl.md`, `plugins/subflow/impl.md` and the generic `<NODE_TYPE>` template in `editing-operations-json.md` all taught `output` + `=result.response` on families whose registry says otherwise. Also corrected the note in `editing-operations-json.md` claiming the emitter ignores the instance `outputs` block — true for most nodes, **not** for Orchestrator-job nodes, whose `source` the converter copies verbatim. That belief is why this survived.

### Evidence added for the one unproven sentence

The "5 dies at 5, 25 dies at 25" claim now has a fault record: debug instance `f88fa4bf-08af-4a83-9aec-de7f20d726f4` → `AGENT_RUNTIME.TERMINATION_MAX_ITERATIONS`, *"Maximum iterations of '25' reached."*, with 25 `contextGroundingTool` spans on the analyst element. (Reviewer couldn't find it because that run was executed by hand, outside `tests/runs/`.)

### Left untouched deliberately

- `plugins/queue`, `plugins/flow`, `plugins/scheduled-trigger`: their registry entries do not resolve on this tenant, so the correct `source` is unverified. `=result.Error` → `=Error` was applied (converter-level fact), but their `output` sources were not changed.
- `core.action.script` / `core.action.transform` / connector examples keep `=result.response` — verified correct.
- Claim 1's diagnosis is softened in fact if not in wording: the failing agent never opened the hub file *or* the guide, so the effective fix is the inlined sentences rather than the retargeted link. The link fix stands on its own merit.

jiyangzh:
## Third commit `15d03207d` — validated against a Studio Web solution package

Checked the docs against `Solution 40.uis` — a solution authored in Studio Web, i.e. the product's own serialization rather than the registry's declaration. Two families were still wrong, including the one the previous commit flagged as unverified.

What Studio Web actually writes:

| Node | `outputs` block |
|---|---|
| `core.trigger.manual` | `output` with `"source": "null"` (literal string) |
| `uipath.core.flow.*` (published flow) | `error` only, `"source": "=Error"` — **no `output`** |
| `uipath.agent.autonomous` | `error` only, `"source": "=Error"` |
| `uipath.agent.resource.context.index.*` | `outputs: null` |
| `core.control.end` | `outputs: null` |

And `variables.nodes[]` declares `autonomousAgent1.output` even though the node's own `outputs` has no `output` entry — direct confirmation that the injected/manifest output is what surfaces downstream, which is the mechanism this PR documents.

Fixed accordingly:
- `plugins/flow/impl.md` — was `output` + `"=result.response"`; now `error` only, with the reason.
- `core.trigger.manual` examples in `shared/file-format.md` (×2) and `plugins/subflow/impl.md` — `"=result.response"` → `"null"`.
- `shared/file-format.md` — the `source` field description no longer implies `"=result.response"` is universal; it now states the rule per family (`"=Error"` always; `"=result.response"` for connector/script/transform/HTTP; `"=this"` or omitted for Orchestrator-job nodes).

`core.action.script` inside the subflow example keeps `"=result.response"` — registry-confirmed correct.

jiyangzh:
## Fourth runtime-only contract, found and closed the same way — `5b15ec06c`

Re-running the DevCon e2e after the tenant recovered surfaced one more validate-blind defect: the agent hand-wrote the Data Service `queryExpression` as `'accountNumber' = '...'` — CEQL reads the quoted token as a string literal, IS rejects with 400 `"Expected a field name expression but got 'StringValue'"`, surfaced as flow incident `102003` on both lookup nodes. `flow validate` passes; only `node configure --detail.filter` validates the expression, and a string hand-written into the `.flow` JSON bypasses it.

Proof (same one-change standard as the other three): unquoting the two field names in the faulted artifact — nothing else — took the flow from Faulted to a completed SOP-grounded resolution.

Doc fix: one row in the connector plugin's Debug table (symptom → cause → fix: field names bare, values quoted; or author via `--detail.filter` so the CLI compiles the CEQL).

**Fresh-build verification with all four fixes live** (`runs/2026-07-27_14-17-13`): **3/3, score 1.000**, one iteration. The new agent wrote bare CEQL field names unaided, plus the three earlier contracts (retrieval cap + maxIterations 25, `error`-only api-workflow outputs, `array` input types). Tenant-side spans confirm real execution: 13/13 elements Completed, real SOP retrieval (`billing-dispute-sop.pdf`), real credit-memo round-trip, resolution citing the invoice.

That makes two independent unaided passes (2026-07-24 run with fixes 1–3, 2026-07-27 run with fixes 1–4). The intervening failures were tenant capacity (stuck-job backlog, since cleared) plus this bug.

rockymadden (approved):

rockymadden:
Can you address the medium finding?

jiyangzh:
Both review findings addressed in `a042cc05c`: the **Medium** (adjudicated valid — `action-nodes.md`'s blockquote made an *exclusive*-exception claim that foreclosed the Orchestrator-job exception; now mirrors `editing-operations-json.md`'s wording) and the **Low** (stale hub link at `inline-agent/impl.md:118` → `autonomous-agent-prompting-guide.md`).

PR #2280, opened 2026-07-26
test(admin): migrate identity coder-eval tests off command_executed → read-back verify (PLT-107839)

t-hsia:
## PLT-107839 — switch admin **identity** coder-eval tests off `command_executed` → deterministic, agent-agnostic criteria

Migrates all **19 `identity_*`** admin tests from brittle `command_executed` command-string matching (and removes every `skill_triggered` check) to **deterministic outcome grading**: the agent retrieves/mutates real tenant state and the harness verifies the *result*, per the ticket.

Command-string matching was the #1 cross-agent flakiness source — Claude vs codex differ on flag order, `$UIP` vars, verb choice, batching, and how they save output. It grades *typing*, not *achieving*.

### Approach by archetype
| Archetype | Tests | How it's graded now |
|---|---|---|
| Read-only list/get | 6 | retrieve → `run_command` asserts real records from the tenant (envelope/save-shape tolerant) |
| Diagnose / troubleshoot | 5 | read-back on the diagnostic retrieval (4) + `llm_judge` on the root-cause conclusion (1); `command_not_executed` guards kept |
| Mutations | 7 | **real** create/revoke/rotate/add against the live tenant + `pre_run` seed / `post_run` cleanup, then a verify script **reads the resource back** from the tenant |
| E2E (user lifecycle) | 1 | `pre_run` seeds the user → agent **updates** surname → read-back proves the update transition landed |

All grading is **agent-agnostic and save-shape robust**: list reads use a recursive dict-list finder; object reads (SMTP, effective-access) use recursive signature-field search; mutation verifies query the tenant directly (immune to how the agent saves output); the one `llm_judge` grades the conclusion. `command_not_executed` safety guards are retained (agnostic negative guards, not the brittle positive match the ticket targets).

### Enabler
The coder-eval runner authenticates `uip` as a real licensed user against a live tenant, so read-back is possible. A Phase-0 probe confirmed **read** access; a `create_pat` probe confirmed **write** access before the mutation conversions.

### Validation
Sequential runs only (`-j1`, shared live org). **codex 18/18** (full set, sequential) + fed-cred; **Claude 19/19 on three consecutive runs**. Each phase (read-only, diagnose, mutations, e2e) passed an independent critical-review gate; all findings addressed.

### Notable fixes made along the way
- **`cleanup_pat.py` casing bug**: matched lowercase `description` but the tenant returns PascalCase `Description`, so historical create-PAT tests never revoked their tokens → leaked into the 5-PAT cap on the shared alpha org. Fixed; cleanup now distinctive-marker-scoped (never touches real PATs).
- Three agent-save-variance flakes (`login_failure`, `access_denied`, `smtp`) — Claude wrapped/reshaped saved output where codex dumped raw. Fixed by making checks save-shape tolerant.

### Follow-up for the Identity team (platform bug, NOT a test defect)
`uip admin external-apps federated-credentials create` returns `HTTP 400: Client <id> not found in partition <N>` for a **freshly-created** external app. Alpha App Insights (`plt-alp-we-01-appins`) investigation confirmed an **identity-service create→read cross-partition inconsistency**: `create` returns 201 and the app lists fine, but its client record isn't materialized in the org partition the by-id/fed-cred read paths query (partition is otherwise healthy; the same partition works for established clients and later fed-cred creates). Not replication lag (persists past 75s + a later 404). The `federated_credentials` test works around it with a **persistent host app** (created once, never deleted; the test churns only the credential). Worth filing/fixing platform-side so create→immediately-use is consistent.


bai-uipath (approved):
Fix if you agree, otherwise lgtm — grading against real tenant state is the right call, the mutation verifies are the strong part of this, and clean codex + Claude runs are good evidence the cross-agent flakiness is actually gone rather than just moved.

- **PAT cleanup markers cross-delete between concurrent tasks.** `cleanup_pat_marker.py` and `setup_revoke_pat.py` both sweep by the substring `ce-identity-smoke`, which also matches the other task's token, and the nightly runs these adjacent tasks in parallel. The bad direction is the revoke task: its seed gets swept away, and `verify_pat_revoked.py` reads "seed absent" as success. Sequential validation runs wouldn't have surfaced this. **Fix:** exact-match each task's own description, as the ext-app cleanups already do.

- **The smtp diagnose task now grades identically to the smtp get task.** Same verification block in both, so the only difference is prompt framing and nothing grades the diagnosis. **Fix:** add a judge on the agent's conclusion, or fold the two into one task.

- **Both PAT-list checks pass on an empty collection.** The `if coll:` guard makes every assertion conditional, and empty is the common case given the cleanups revoke test tokens. **Fix:** seed a marker PAT in pre_run and assert it shows up in the saved file.

Not a blocker: on the lifecycle e2e the harness does the invite and the delete while the agent does one field update — `verify_user_updated.py` already checks existence *and* surname, so dropping the invite from the seed script gets you invite + update graded for free.

yadvender-uipath (approved):

PR #2287, opened 2026-07-27
Improve Data Fabric skills and coder evals

AditiGoyalUipath:


uipreliga (commented):
## Review: Data Fabric eval cleanup

Reviewed with three independent passes (Claude Opus, Gemini 3 Pro, GPT-5.3-Codex) plus manual verification of every load-bearing claim. Findings marked **[3/3]** / **[2/3]** were flagged independently by that many reviewers.

**Verdict: changes requested.** Three of the four goals are right — the `_shared` retry/folder-key hardening, `fail_on_error: true` on seeds, and replacing command-shape regex with live-schema `run_command` oracles (`integration_multiline_max.yaml:44` is the model to copy) are real improvements, and the doc trim closed four genuine doc/test contradictions. But "reduce flakiness" was largely applied by **weakening graders** rather than making them deterministic, and in several places it inverted: tasks now pass while proving nothing.

### Blocking (3)

1. **`integration_entity_scope.yaml` grades nothing** — one `uip df entities list` call scores 12.5/12.5. See inline at `:54`.
2. **`integration_files.yaml` passes for `cp`** — no `files download` criterion survives, and `cmp` compares two local files. See inline at `:62`.
3. **`smoke_choice_relationship.yaml` self-poisons permanently** — both cleanup hooks deleted while `--expected 1` kept. Degrades every *future* run, not just this PR. See inline at `:226`.

### Not visible in this diff

**`_shared/verify_record_field.py:65` was left out of the tenant-scope fix — the 5th copy.** This PR adds the identical `TENANT_SCOPE` / `entity_folder_key()` normalisation to four scripts (`assert_record_count.py:49`, `cleanup_entities_by_name.py:48`, `verify_field_binding.py:50`, `verify_field_constraints.py:51`) because the CLI reports the all-zeros GUID as the folder key for tenant-scoped entities. `verify_record_field.py` still does:

```python
return (
    e.get("ID") or e.get("Id") or e.get("id"),
    e.get("FolderKey") or e.get("folderKey") or "",   # :65 — not normalised
)
```

`sample_record()` then passes that straight to `--folder-key` (`:72-73`), i.e. `--folder-key 00000000-0000-0000-0000-000000000000` for a tenant-scoped entity. Its one remaining consumer is `e2e_expense_tracker.yaml:219` at **weight 3.0**, on a tenant-scope-capable path. If the premise behind the other four fixes holds, that criterion fails deterministically.

**Fix:** hoist the helper into `_shared/df_common.py` and import it in all five. These are test helpers under a shared `_shared/` dir, so the skill self-containment rule in `CLAUDE.md` does not apply — and five copies of a 9-line function is exactly how this bug happened. Note also that `verify_field_binding.py:70` and `verify_field_constraints.py:77` parse `Data` **only as a list**, while `assert_record_count.py:74` and `cleanup_entities_by_name.py:76-81` also handle `{"Records": [...]}`. Both shapes cannot be right; if the dict shape is what the CLI emits, the two narrow readers burn all their new retries and then report a misleading "entity not found".

### Coverage lost with no replacement

14 tasks deleted. Defensible: `integration_filter_contract` (a CLI/platform-contract test masquerading as a skill eval — and its anti-pattern guards survive in `integration_query_filters.yaml:81-123`), `integration_naming_edge_cases` + `integration_constraint_violations` (exhaustive boundary probes with `--allow-mismatches` escape hatches), `integration_choice_relationship` / `integration_preseeded_entity` (near-duplicates of `e2e_expense_tracker` / `integration_query_filters`), `smoke_entities` / `smoke_records` / `smoke_bulk_import` (folded into survivors).

Silent regressions worth acknowledging in the PR body even if the trim stands:

| Capability | Deleted with | Still documented? |
|---|---|---|
| Cursor-walk correctness | `verify_pagination.py`, `extract_page_ids.py` | n/a — only remaining pagination task now self-reports |
| Filter edge cases (null vs empty vs missing, `'`/`%`/`_`/`\` escaping, case sensitivity, page-size 0/oversized/past-end) | `integration_filter_edge_cases` | Yes — `filter-platform-contract.md`, now untested |
| `CHOICE_SET_MULTIPLE` set-equality / not-contains query shapes | `smoke_choice_multiple_filter` | Yes — `records-query.md` § Filtering on Choice-Set Fields |

### Low / informational

- **`tests/tasks/uipath-test/integration_release_readiness_qa_lead.yaml:137`** — comment points at `integration_preseeded_entity.yaml:189-190`, deleted here. It is the sole surviving pointer to the persistent-shared-fixture pattern it documents; repoint at `integration_query_filters.yaml` (same `seed_entity.py` / `revert_entity_to_seed.py` shape).
- **MULTILINE_MAX version floor removed** — the trim dropped `entity-schema.md`'s "`@uipath/data-fabric-tool` 1.198.0+" note and the Tool Version Requirements table, while `integration_multiline_max.yaml:42-48` newly asserts `Body.Name=MULTILINE_MAX` at weight 3.0 with no preview-install step. Fine if `@latest` bundles the type; unpassable if not.
- **`RecordOwner` added inconsistently** — `entity-schema.md:106` now lists six system fields; `entity-schema.md:331` and `bulk-import.md:36` still list five.
- **`SKILL.md:43`** (untouched here) still promises `data-fabric.md` carries the "request-body schema, per-type operator matrix" — the trim removed the body-schema table, and the operator matrix lives in `filter-platform-contract.md`.
- **Weight convention drift** — `0.67/0.67/0.66` in `integration_files.yaml:74,82,90` and `integration_error_paths.yaml:130,138,146`; the ladder in `.claude/rules/test-writing.md` is `1.0 · 1.5 · 2.0 · 3.0 · 5.0–6.0`. Arithmetically fine, off-convention. `1.0` each, or one criterion with `min_count: 3`, reads better.
- **`cleanup_entities_by_name.py:102-108`** — `first_folder_key()` falls back `Key → key → Id → id`. Orchestrator folders carry both a GUID `Key` and a numeric `Id`; `--folder-key` wants the GUID, so a CLI version emitting only `Id` would silently pass a number. Prefer Key-only plus a `SKIP:`. It also prints `SKIP:` diagnostics to **stdout** from a pure-lookup-named function, and its `folders[0]` choice is invisibly coupled to the two consuming prompts' "pick the first folder returned".
- **`integration_files.yaml:80`** — regex embeds a literal `café-📎\.txt` matched against raw command text; any shell-escaping form or Unicode normalisation difference silently misses.
- **`integration_entity_scope.yaml:19-23`** promises an "invalid folder ID" the prompt no longer supplies (the literal all-zeros GUID was removed). Plausibly deliberate, since `TENANT_SCOPE` now treats that GUID as *tenant scope* — but either restore a clearly-invalid non-zero GUID or reword to "a folder ID you make up".
- **`smoke_negative_guards.yaml:2`** — says "four requests for operations that the skill explicitly prohibits", then lists three prohibitions plus one advisory. Separately: the new `min_count:0/max_count:0` guard on `entities create Order` docks an agent that verifies the rejection empirically. `data-fabric.md:18`'s immediate-refusal rule only applies when the prompt establishes the disqualifying fact, and here it doesn't — is refusing-from-doc-knowledge the intended contract?
- **Stale `description:` text** — `integration_csv_import.yaml:5` says "confirm exact field names (case-sensitive)" while the docs now require `Fields[].DisplayName` headers; `integration_folder_binding.yaml:9` says FILE fields need same scope class while `entity-schema.md:198` now excludes FILE.
- **Inconsistent cleanup flags** — `e2e_employee_directory.yaml:205,210` gained `--include-folders`; `e2e_product_catalogue.yaml:178,183` did not, though both prompts say tenant scope.
- **`integration_entity_scope.yaml:38-41`** — last two prompt bullets sit two columns left of the first three. YAML-valid inside the `|` block, reads as a truncated edit.
- `.claude/rules/test-writing.md` step 5 wants a passing-run claim in the PR body; almost every grader here changed, so it's worth adding.

### Verified clean

- All 18 remaining task YAMLs parse, carry `skill` + `tier` + `mode:` + `lifecycle:` tags, and none pin `@uipath/cli` in `sandbox.node.env_packages`.
- `python3 scripts/check-cli-verbs.py tests/tasks/uipath-platform/data-fabric/*.yaml` → clean (catalog uip 1.197.0, 1230 verbs).
- No dangling filesystem references to `extract_page_ids.py`, `verify_constraint_outcomes.py`, `verify_filter_contract.py`, `verify_pagination.py`, `_shared/inputs/*`, `codeeval_test_entity.*`, or any deleted task id. All seed JSONs and surviving helpers still have consumers.
- Every anchor link and `Rule N` cross-reference in the seven trimmed docs still resolves. The one broken anchor (`entity-schema.md:29` → `#ui-broken-types-do-not-use`, single hyphen vs the actual double) pre-dates this PR.
- `bulk-import.md` still fully supports the new `smoke_negative_guards` task 4 (complex-field CSV limitation, `records insert --file` alternative, `files upload`-after-row-exists).
- Removing the "OFFLINE SMOKE TEST" preambles is safe — `smoke-skills.yml` authenticates against a live alpha tenant.

### Worth keeping

- `assert_record_count.find_entity_id` / `total_count` are correctly written retry loops — bounded attempts, `attempt + 1 < ATTEMPTS` so no trailing sleep, last-error accumulation, diagnostic on exhaustion. `verify_field_constraints` should be rewritten to match.
- The `cleanup_entities_by_name.delete_entity` propagation-poll fix is right: `--include-folders` doesn't scope the query, `--folder-key` does, so the ghost-entity poll now observes the correct index.
- `fail_on_error: false → true` on six seed steps — silently running a brownfield task against a missing fixture produced exactly the false failures this PR targets.
- `INTEGER → DECIMAL` + `decimalPrecision: 0` in three seed schemas makes the fixtures consistent with the skill's own Rule 12b.
- The doc trim fixed four latent doc/test conflicts: `records-query.md:69` `Rule 16 → 17`; `data-fabric.md:59` Rule 4 previously said `Order` is *not* rejected — the opposite of what `smoke_negative_guards` task 3 grades; `choice-sets.md` no longer prescribes `snake_case` value names that would fail the `paid,unpaid,refunded` fixture; and `bulk-import.md`'s new `Fields[].DisplayName` header rule is what makes `integration_csv_import` passable given its deliberately-diverging seed display names.
- Per-task isolated fixtures (`smoke_employees` / `smoke_projects` / `smoke_documents`) remove real cross-task coupling.

tests/tasks/uipath-platform/data-fabric/integration_entity_scope.yaml
```diff
@@ -47,13 +51,9 @@ success_criteria:
...
     pass_threshold: 1.0
 
-  - type: command_executed
-    description: "Agent created the folder-scoped entity (CE_FolderScopeTest) — only fires when the tenant has ≥2 folders; soft assertion since the prompt allows skipping this branch on single-folder tenants"
-    tool_name: "Bash"
-    command_pattern: '(?s)uip\s+df\s+entities\s+create\s+CE_FolderScopeTest'
-    min_count: 1
-    weight: 2.0
-    pass_threshold: 0.0
+  # Advisory folder-scoped-create check removed — the prompt permits
```
uipreliga: **Critical — this task now grades nothing.** After this change all 7 criteria award full marks to an agent that runs one `uip df entities list` and stops (12.5 / 12.5):

| criterion | why it passes for free |
|---|---|
| `uip login status` (`:49`) | `min_count: 0` |
| must NOT recreate `CE_TenantScopeTest` | inaction passes |
| `entities list` `min_count: 1` | one call |
| `fieldName` in create body | `pass_threshold: 0.0` |
| `entities list … --folder` | `pass_threshold: 0.0` |
| must NOT `entities delete` | inaction passes |
| must NOT `removeFields` | inaction passes |

Nothing verifies that the tenant entity is visible from a folder context, that `CE_FolderScopeTest` landed in folder A, that it is invisible from folder B, or that the invalid-folder create errored — i.e. none of "S-SC-01 through S-SC-05" that `description:` still claims.

The justification in this comment also rests on a factual error: the removed check had `pass_threshold: 0.0`, and per `tests/README.md` + `.claude/rules/test-writing.md` ("To record convention adherence, use an advisory check (`pass_threshold: 0`)") that means *zero fraction of `min_count` required* — it can never "drag the score". The same reasoning removed the `--native-only` advisory.

**Fix:** restore the deleted `llm_judge` (weight 2.5, `pass_threshold: 0.5`) and the folder-scoped-create advisory. Better: add a deterministic `run_command` that lists with `--folder-key <B>` and asserts `CE_FolderScopeTest` is absent.

tests/tasks/uipath-platform/data-fabric/integration_entity_scope.yaml
```diff
@@ -65,10 +65,10 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
-    description: "Agent listed entities at least twice (different scope contexts)"
+    description: "Agent listed entities in the requested scope contexts (calls may be grouped in one Bash action)"
     tool_name: "Bash"
     command_pattern: 'uip\s+df\s+entities\s+list'
```
uipreliga, line 66: `min_count` 2 → 1, but the description still says "the requested scope **contexts**" (plural). This collapses the *count*, not the *content* — one `--folder-key <A>` listing now satisfies a criterion that requires A **and** B.

If the rationale is that the agent may group calls into one Bash action, keep `min_count: 1` but assert both keys appear:

```yaml
command_pattern: '(?s)entities\s+list.*--folder-key.*entities\s+list.*--folder-key'
```

Same issue at `integration_folder_records.yaml:73-79`, whose description says "listed entities in **both** folder contexts".

tests/tasks/uipath-platform/data-fabric/integration_query_filters.yaml
```diff
@@ -43,34 +41,34 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
-    description: "Agent ran filter queries with --body (AND, OR, nested, sort, projection)"
+    description: "Agent ran filter queries with --body (queries may be grouped in one shell action)"
     tool_name: "Bash"
-    command_pattern: '(?s)uip\s+df\s+records\s+query\s+.*--body'
-    min_count: 4
+    command_pattern: '(?s)uip\s+df\s+records\s+query\s+.*(?:--body|--file)'
```
uipreliga: **High [3/3] — the `|--file\s+\S+` alternation launders every body-shape criterion in this file.**

A single `uip df records query <id> --file q.json`, with `q.json` containing `{}`, satisfies the positive criteria at `:46` (3.0), `:70` (2.5), `:84` (2.5), `:128` (2.0), `:136` (2.0), `:144` (2.0) = **14.0 weight** — with zero evidence the body contained `filterGroup`, `queryFilters`, `logicalOperator`, `sortOptions`, or `selectedFields`.

It compounds: the four anti-pattern guards at `:92`, `:101`, `:110`, `:119` (`"filters":`, `"field":`, `"=="`, `Equals`) did **not** get the `--file` branch, so the same opaque call auto-passes those too — another 8.0. That's ~22 of ~33 total weight from one command that proves nothing.

`min_count` also dropped 3/4/2 → 1 across these, so one query now satisfies criteria written to prove *distinct* filter shapes.

**Fix:** don't alternate on the flag. Either drop the shape criteria and lean on the `llm_judge` (`:149`) + `assert_record_count`, or keep them `--body`-only and add a companion `run_command` that greps the payload. `.claude/commands/lint-task.md` names this exactly: "`command_executed` patterns so loose that any usage of the CLI passes".

Same pattern in `e2e_employee_directory.yaml:101,117,125`, `e2e_product_catalogue.yaml:90`, `e2e_expense_tracker.yaml:146`. Of those only `e2e_employee_directory.yaml:109` has a compensating deterministic check.

tests/tasks/uipath-platform/data-fabric/integration_query_filters.yaml
```diff
@@ -127,47 +125,23 @@ success_criteria:
...
-    command_pattern: '(?s)uip\s+df\s+records\s+query.*(?:aggregates.*groupBy|groupBy.*aggregates)'
-    min_count: 1
-    weight: 2.5
-    pass_threshold: 1.0
-
-  - type: command_executed
-    description: "Agent used COUNT aggregate function with an alias"
-    tool_name: "Bash"
-    command_pattern: '(?s)uip\s+df\s+records\s+query.*"function"\s*:\s*"COUNT".*"alias"'
+    command_pattern: '(?s)uip\s+df\s+records\s+query.*(?:selectedFields|--file\s+\S+)'
```
uipreliga: **High [2/3] — prompt asks, nothing grades.** This change deleted the criteria for the `in` / `valueList` operator, `groupBy` + `aggregates`, and `COUNT` + `alias`, but `initial_prompt` still asks for both behaviours:

> - Orders whose status is one of {Pending, Active} — how many?
> - Count of orders grouped by status — one count per status.

Aggregate and IN-operator coverage is now unenforced. Either restore the three criteria or drop those two asks from the prompt so the task's scope matches what it grades.

tests/tasks/uipath-platform/data-fabric/integration_files.yaml
```diff
@@ -52,35 +59,35 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
-    description: "Agent uploaded file at least twice (v1 and v2 overwrite)"
+    description: "Agent deleted the file attachment"
```
uipreliga, line 62: **High — this task now passes for `cp`.** The `files upload --file` (`min_count: 2`, weight 3.0) and `files download --destination` (`min_count: 2`, weight 2.5) criteria were both replaced here, and **no `files download` criterion survives anywhere in the file**. The only upload criteria left are the three edge-case filename probes (`:69-91`, 2.0 combined).

The byte-integrity checks that remain compare two *local* files:

```yaml
- command: "cmp -s original.txt downloaded.txt"      # :149
- command: "cmp -s updated.txt verify_update.txt"    # :157
```

So an agent that writes the four files locally and runs `cp original.txt downloaded.txt; cp updated.txt verify_update.txt` — never touching the platform — collects 1.0 (login, `min_count: 0`) + 1.5 (records lookup) + 4.5 (four `file_exists`) + 1.0 (`empty.txt`) + 6.0 (both `cmp`) + 4.0 (both must-NOTs) = **18.0 / 24.0 = 0.75**. The round-trip through Data Fabric — the entire point per `description:` lines 3-8 — is unenforced.

**Fix:** restore `command_executed` for `files\s+upload\s+.*--file` (`min_count: 2`) and `files\s+download\s+.*--destination` (`min_count: 2`). The `file_exists` + `cmp` checks are a good addition, but they only mean something once the download is required to be what produced `downloaded.txt`.

tests/tasks/uipath-platform/data-fabric/smoke_choice_relationship.yaml
```diff
@@ -191,10 +224,11 @@ success_criteria:
     pass_threshold: 1.0
 
 pre_run:
```
uipreliga, line 227: **Blocking — this task self-poisons after one bad run, permanently.**

`main` had both:

```yaml
pre_run:
  - command: "… cleanup_entities_by_name.py --name-prefix CE_SmokeExpense"
post_run:
  - command: "… cleanup_entities_by_name.py --name-prefix CE_SmokeExpense"
```

This rewrite deleted **both** and added **no `post_run` at all** (the file ends at this `pre_run` block), while converting the task from offline shape-matching to live brownfield and keeping `--expected 1` at `:134` (weight 3.0).

The first run where the agent inserts a second expense — or dies between insert and completion — leaves `CE_SmokeExpense` at 2 records. With no pre-run reset, that criterion then fails on **every subsequent run forever**. It also leaks `CE_SmokeExpense`, `CE_SmokePaymentState0726`, and the agent-added `CE_SmokeEmployees.status` column into the shared tenant.

**Fix:** add a `pre_run` reset for `CE_SmokeExpense` — either `cleanup_entities_by_name.py`, or `seed_entity.py --reset` with an empty records file so the "reuse existing entity" path is still exercised — plus a matching `post_run`.

Two asides: `CE_SmokePaymentState0726` hardcodes a date stamp into both the prompt (`:56`) and the grader (`:124`) — a throwaway name in a permanent test; and this is tagged `smoke` yet carries `max_turns: 120` / `task_timeout: 2400` (`:8-12`), 2.7× the `smoke.yaml` default of 900s, in the PR gate.

tests/tasks/uipath-platform/data-fabric/smoke_choice_relationship.yaml
```diff
@@ -84,59 +82,58 @@ success_criteria:
...
-    tool_name: "Bash"
-    command_pattern: '(?s)uip\s+df\s+entities\s+(create|update)(?=.*RELATIONSHIP)(?=.*referenceEntityId)(?=.*referenceFieldId)'
-    min_count: 1
-    weight: 3.0
+  - type: run_command
+    description: "Live Expense schema contains all requested complex field types and amount constraints"
+    command: >-
+      python3 $SKILLS_REPO_PATH/tests/tasks/uipath-platform/data-fabric/_shared/verify_field_constraints.py
+      --entity-name CE_SmokeExpense --ignore-field-case
+      --assertions 'amount.Name=DECIMAL,amount.decimalPrecision=2,amount.minValue=0,amount.maxValue=1000000,category.Name=CHOICE_SET_SINGLE,tags.Name=CHOICE_SET_MULTIPLE,submitter.Name=RELATIONSHIP,project.Name=RELATIONSHIP,Receipt.Name=FILE,paymentStatus.Name=CHOICE_SET_SINGLE'
```
uipreliga: `paymentStatus.Name=CHOICE_SET_SINGLE` is asserted inside a 10-assertion, weight-4.0 criterion, but the prompt (`:56-58`) only says *"`CE_SmokeExpense` has it as a payment-status column"* — the column name is never pinned. `PaymentState`, `payment_status`, or `state` all fail, and since `verify_field_constraints.py` exits 1 on *any* failure (`:188-191, 202-204`), one miss drags the other nine assertions down with it.

**Fix:** name the column explicitly in the prompt, and split multi-field assertion blobs into per-field criteria so a single miss doesn't zero the rest.

tests/tasks/uipath-platform/data-fabric/integration_large_pagination.yaml
```diff
@@ -71,25 +59,39 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
```
uipreliga: **High [2/3] — pagination correctness is now pure self-report.** `min_count` dropped 9 → 1 at the same time the deterministic oracles were deleted (`extract_page_ids.py`, `verify_pagination.py --mode walk --count 500` / `--mode offset --against walk_ids.txt --start 180 --stop 240`).

The 9-page / 500-unique-ID assertion now survives only as the `llm_judge` below, whose prompt says:

> Grade the reported outcome only; command criteria separately verify execution, so do not require files, screenshots, logs, or other artifacts here.

But the command criterion it defers to is this one — and a single `--limit 1` satisfies it. So the agent can assert "9 pages, 500 unique IDs, no duplicates" without walking anything.

This is the pattern `.claude/rules/test-writing.md` § *Grade Behavior, Not Self-Reports* prohibits, and `.claude/commands/lint-task.md` rates "`llm_judge` as the only or dominant criterion (graded by an LLM with no ground truth)" as **High**.

**Fix:** restore `verify_pagination.py`. A helper that reads an ID file the agent produced is deterministic and *less* flaky than an LLM judging prose — this de-flaking change made the grader noisier, not quieter. At minimum require `--limit 60` with `min_count: 5`.

tests/tasks/uipath-platform/data-fabric/integration_large_pagination.yaml
```diff
@@ -122,10 +124,50 @@ success_criteria:
...
```
uipreliga, line 144: **Medium [2/3] — these CRUD criteria don't verify what their descriptions claim.**

- **"as a batch"**: `(?s)…insert.*(?:\[|--file\s+\S+|--body\s+["']?\$\w+)` — with `(?s).*`, any insert containing a `[` matches, so a single-row `[{…}]` passes. Try `insert.*\[.*Smoke-WriteTests.*Smoke-ReviewPR.*Smoke-Deploy.*\]`.
- **"with their Id values"**: accepts `--file` / `$VAR`, so the `"Id":` requirement is unverified.
- **"the three temporary rows, not the entity"**: satisfied by any single `records delete`.

Bigger issue: the previous "must NOT mutate the seed" guard was removed and **no end-state record-count assertion replaced it**, so a wrong-row deletion against the shared 500-record fixture goes undetected in-task. It self-heals only because the next `pre_run` re-seeds on count mismatch — and the `post_run` revert exits 0 even when `wipe_records` bails on a transient error (`seed_entity.py:167-172`). Add:

```yaml
- type: run_command
  description: "End state: CE_LargePaginationTest back to the canonical 500 records"
  command: "python3 $SKILLS_REPO_PATH/…/_shared/assert_record_count.py --entity-name CE_LargePaginationTest --expected 500"
  timeout: 180
  expected_exit_code: 0
  weight: 2.5
  pass_threshold: 1.0
```

Also scope creep: `description:` still says the task focuses on "pagination correctness at scale", but a full record-CRUD sequence was bolted on (max_turns 50 → 60), converting a read-only brownfield task into a mutating one. CRUD coverage lost with `smoke_records.yaml` was folded in here rather than given its own task, muddying both objectives.

tests/tasks/uipath-platform/data-fabric/smoke_csv_type_confirm.yaml
```diff
@@ -12,95 +12,40 @@ run_limits:
...
     weight: 3.0
     pass_threshold: 1.0
 
-  - type: command_executed
-    description: "Agent engaged the CLI before pausing — proves it didn't just write the self-report file"
-    tool_name: "Bash"
-    command_pattern: 'uip\s+df\s+(entities|choice-sets|--help)'
-    min_count: 1
-    weight: 1.5
+  - type: llm_judge
```
uipreliga, line 39: **High [2/3] — deterministic grading replaced by a 9-clause binary judge.** Four checks (weight 9.0, including `file_contains ["DATETIME_WITH_TZ","BOOLEAN","DECIMAL","RELATIONSHIP"]`) were replaced by this single `llm_judge`.

The task is now exactly two criteria: this judge, and a `min_count:0/max_count:0` negative guard that an agent doing **nothing at all** passes. So the only discriminating signal is monologue — `.claude/rules/test-writing.md`: "Never grade on agent monologue."

The rubric is also the highest-variance shape available: *"Score 1 only if all five columns are covered, the four ambiguities are surfaced, and the response clearly pauses; otherwise 0"* — all-or-nothing on a 9-part conjunction, at `pass_threshold: 1.0`. And it asks the judge to check "use only UI-compatible field types" without telling it which those are; the deleted `file_contains` pinned them.

**Fix:** split the conjunction into separate criteria with `pass_threshold: 0.5`, and keep at least one deterministic artifact check. Same shape at `smoke_choiceset_scope_ask.yaml:30`.

tests/tasks/uipath-platform/data-fabric/smoke_choiceset_scope_ask.yaml
```diff
@@ -49,3 +26,14 @@ success_criteria:
     max_count: 0
     weight: 3.0
     pass_threshold: 1.0
+
+  - type: llm_judge
```
uipreliga, line 29: Same shape as `smoke_csv_type_confirm.yaml:39`: `file_exists scope_question.md` + `file_contains ["folder","tenant"]` (weight 5.0) were replaced by one `llm_judge`, leaving the task with this judge plus a negative guard an idle agent passes. The entire positive signal is one LLM judgement on prose.

Separately — the prompt now says:

> I have not decided whether it belongs at tenant level or in a folder. **Ask me before creating it.**

That telegraphs the expected behaviour, converting a Rule 19 *knowledge* test ("does the agent know scope must be resolved first?") into an instruction-following test. Consider dropping the last sentence and grading the `AskUserQuestion` call instead.

tests/tasks/uipath-platform/data-fabric/_shared/assert_record_count.py
```diff
@@ -41,45 +46,67 @@ def run_uip(*args: str) -> tuple[int, str, str]:
...
-            continue
-        try:
-            data = json.loads(out)
-        except json.JSONDecodeError:
-            continue
-        inner = data.get("Data") if isinstance(data, dict) else None
-        recs = inner if isinstance(inner, list) else (inner or {}).get("Records") or (inner or {}).get("records") or []
-        for ent in recs:
-            if not isinstance(ent, dict):
+    for attempt in range(ENTITY_LOOKUP_ATTEMPTS):
```
uipreliga, line 65: **Medium [3/3] — the retry budget can't fit the harness timeout.** `find_entity_id` does 3 attempts × 2 uip calls (each up to `UIP_TIMEOUT_SECONDS = 60`) + 2×2s sleeps; `total_count` adds 3 × 60s + 2×3s. Worst case ≈ **9 minutes**; even a moderately slow tenant at ~5s/call burns ~55s.

But 8 call sites pass `timeout: 60`: `integration_csv_import.yaml:96`, `integration_folder_binding.yaml:134,142`, `integration_folder_records.yaml:105`, `integration_multiline_max.yaml:69`, `e2e_expense_tracker.yaml:196,204`, `e2e_product_catalogue.yaml:131`, `e2e_employee_directory.yaml:159`. The harness kills the script mid-retry, so attempts 2-3 are unreachable and the criterion reports an opaque timeout instead of the intended `FAIL:` message — the opposite of the de-flaking goal.

**Fix:** either `UIP_TIMEOUT_SECONDS = 15` with `ATTEMPTS = 2` (both external reviewers' suggestion), or raise those `timeout:` values to ≥180.

Secondary: the inner `for extra in (["--include-folders"], [])` was a *capability probe* for older CLIs, not something transient. Nesting it in the retry loop re-runs the unsupported flag on every attempt, doubling call count for no benefit. Latch it:

```python
if extra and "unknown option" in (err or "").lower():
    unsupported_include = True
```

tests/tasks/uipath-platform/data-fabric/_shared/verify_field_constraints.py
```diff
@@ -45,11 +48,27 @@ def run_uip(*args: str) -> tuple[int, str, str]:
...
+    return "" if str(folder_key).lower() == TENANT_SCOPE else str(folder_key)
+
+
 def find_entity_id(name: str) -> tuple[str | None, str | None]:
     """Return (id, folder_key) — folder_key is empty for tenant-scoped."""
-    for extra in (["--include-folders"], []):
+    # The tenant list endpoint is occasionally eventually consistent directly
+    # after entity creation. Try folder-aware discovery once, then retry the
+    # broadly supported tenant listing before declaring the entity absent.
+    for attempt, extra in enumerate((["--include-folders"], [], [], [])):
```
uipreliga: **High [2/3] — this retry can't help the case it was added for.** `--include-folders` is tried only on attempt 0; attempts 1-3 are all the plain tenant listing. So for a **folder-scoped** entity that hasn't propagated yet, the retry adds nothing — and folder-scoped entities are exactly the case in `e2e_expense_tracker.yaml` (prompt `:19-20`), which is the task these criteria guard (`:95, :103, :111, :119`).

Two more defects in the same loop:
- The `attempt >= 1` sleep fires after the **final** attempt too — 2s burned before returning `(None, None)`.
- The `json.JSONDecodeError` branch (`:75-76`) `continue`s without sleeping, unlike the other two paths.

And `:77` handles `Data` only as a **list**, while `assert_record_count.py:73-74` and `cleanup_entities_by_name.py:76-81` also handle a dict with `Records`/`records`. Both shapes can't be right; if the dict shape is what the CLI emits, this function never finds anything and the new 6s of sleeping is pure waste — then reports a misleading "entity not found", failing the criterion.

**Fix:** mirror `assert_record_count.find_entity_id` — outer `for attempt in range(3)`, inner `for extra in (["--include-folders"], [])`, sleep guarded by `attempt + 1 < ATTEMPTS`, same dict/list `Data` handling. Or better, share one implementation (see the review body on `verify_record_field.py`).

tests/tasks/uipath-platform/data-fabric/_shared/seed_choice_set.py
```diff
@@ -117,15 +117,17 @@ def create_choice_set(name: str, display_name: str, description: str) -> str | N
     return d.get("Id") or d.get("ID") or d.get("id")
 
 
-def create_value(cs_id: str, value_name: str, display_name: str | None = None) -> None:
+def create_value(cs_id: str, value_name: str, display_name: str | None = None) -> bool:
```
uipreliga, line 120: Nit: the signature changed to `-> bool`, but the sole caller at `:179` discards the result:

```python
create_value(cs_id, v, display_name=v.capitalize())
```

Either use it (e.g. abort value seeding after N failures, or make the "OK: seeded" line conditional) or revert to `-> None`.

tests/tasks/uipath-platform/data-fabric/integration_error_paths.yaml
```diff
@@ -102,19 +107,43 @@ success_criteria:
...
+  - type: command_executed
+    description: "Agent attempted the non-file-column upload (temp.txt targeted at the Score column)"
+    tool_name: "Bash"
+    command_pattern: '(?s)uip\s+df\s+files\s+upload\s+\S+\s+\S+\s+"?Score"?(?:\s|$)'
+    min_count: 1
+    weight: 0.67
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent attempted the nonexistent-record file upload"
```
uipreliga: **Medium [2/3] — this pattern matches the two *other* upload probes, so it can't identify its own scenario.** The alternation includes a bare GUID (`[0-9a-f]{8}-…`) and `no-such`. The second positional after `upload` is the record-id, which is a **real UUID** in the immediately preceding probe ("attach it to the Score column", prompt `:63-64`) — and `no-such` matches the third probe's `no-such-file.bin`. So the criterion passes if the agent does either other probe and skips this one.

`does-not-exist` also lingers in the alternation although the prompt no longer supplies literal `does-not-exist-ccc` IDs — that coupling is now stale.

**Fix:** drop the GUID and `no-such` alternatives, keep only sentinel names the prompt actually induces — or grade the outcome (the surfaced error) via the `llm_judge` at `:184` instead.

tests/tasks/uipath-platform/data-fabric/integration_csv_import.yaml
```diff
@@ -40,25 +45,25 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
-    description: "Agent retrieved entity schema before creating CSV"
+    description: "Agent inspected the entity schema before writing (entities get or the schema-bearing entities list)"
```
uipreliga: **Medium [2/3] — this relaxation makes the criterion unconditional.** The entity is pre-seeded, so the agent **must** call `entities list` to resolve its ID before anything else. Adding `list` to the alternation means the weight-2.0 criterion can no longer fail, and it duplicates the discovery criterion above it.

That matters here specifically: the fixture (`_shared/seeds/integration_inventory.schema.json`) deliberately diverges display names from field names (`SKU` → `Stock-Keeping Unit`), so schema inspection is the actual skill under test — and `description:` line 5 still claims the task "retrieves the schema to confirm exact field names (case-sensitive)".

**Fix:** revert to `entities\s+get\s+`, or keep `list` but require a second distinguishing signal.

tests/tasks/uipath-platform/data-fabric/e2e_employee_directory.yaml
```diff
@@ -84,66 +90,55 @@ success_criteria:
...
-    command_pattern: '(?s)uip\s+df\s+records\s+update.*\"Id\"'
+    command_pattern: '(?s)uip\s+df\s+records\s+update.*(?:\\?"Id\\?"\s*:|--file\s+\S+)'
     min_count: 1
     weight: 2.5
     pass_threshold: 1.0
 
+  - type: run_command
+    description: "End state: employee field constraints match the requested schema"
+    command: >-
+      python3 $SKILLS_REPO_PATH/tests/tasks/uipath-platform/data-fabric/_shared/verify_field_constraints.py
```
uipreliga, line 133: **High [2/3] — new flakiness in a de-flaking PR.** This asserts `FullName`, `Department`, `JobTitle`, `Salary`, `PerformanceScore` at weight 3.0, but:

1. It **omits `--ignore-field-case`**, unlike all four `e2e_expense_tracker.yaml` calls (`:95, :103, :111, :119`).
2. The prompt (`:32-49`) never pins those names — it says "Full name", "Job title", "a performance score (0 to 100)".

So `fullName`, `full_name`, `Name`, `EmployeeName`, `Score` all fail — and `--ignore-field-case` would only fix case, not tokenisation. Since `verify_field_constraints.py` exits 1 on any single failure, one naming miss zeros all five assertions.

**Fix:** pin the column names in the prompt (*"name the columns `FullName`, `Department`, `JobTitle`, `HireDate`, `Salary`"*), add `--ignore-field-case` here for consistency, and split into per-field criteria so one miss doesn't drag the rest.

Moving these constraints from command-shape regex to a live-schema check is the right direction — the problem is only the unpinned names.

tests/tasks/uipath-platform/data-fabric/e2e_product_catalogue.yaml
```diff
@@ -115,6 +112,19 @@ success_criteria:
     weight: 1.0
     pass_threshold: 1.0
 
+  - type: run_command
+    description: "End state: the catalogue contains the requested field types"
+    command: >-
+      python3 $SKILLS_REPO_PATH/tests/tasks/uipath-platform/data-fabric/_shared/verify_field_constraints.py
```
uipreliga, line 118: Same as `e2e_employee_directory.yaml:133`: `--ignore-field-case` is missing, and `ProductName` / `InStock` are asserted from prompt text "product name" / "in stock — yes/no" (`:26-29`) that never pins them. `Product_Name`, `Name`, `IsInStock`, `Stock` all fail and take the whole criterion with them. (`SKU` and `Poster` *are* pinned at `:30-31` — do the same for the other two.)

skills/uipath-platform/references/data-fabric/entity-schema.md
```diff
@@ -184,72 +186,32 @@ uip df entities update <entity-id> \
...
-
-System entities live at tenant level but are exempt from the folder ↔ tenant block — that's how FILE fields work on folder-scoped entities (they point at the tenant-level `EntityAttachment` system entity). The exemption is specific to system entities; ordinary tenant entities and choice sets stay blocked.
-
 Surface this constraint to the user **before** invoking `entities create` / `addFields` whenever the proposed parent and target sit on opposite sides of the tenant ↔ folder boundary AND the target is not a system entity. Do not silently fall back to a different field type — see Rule 18 (no silent substitution).
```
uipreliga: **The "system entity" exemption is still referenced here but its definition was deleted.** This line says:

> …whenever the proposed parent and target sit on opposite sides of the tenant ↔ folder boundary **AND the target is not a system entity**.

The trim removed every definition of that exemption:
- the matrix row `| Folder | Tenant **system** entity (\`EntityAttachment\`, \`User\`) | ✅ | Omit — platform-managed |`
- the paragraph *"System entities live at tenant level but are exempt from the folder ↔ tenant block — that's how FILE fields work on folder-scoped entities…"*
- the `referenceFolderKey` bullet's "(b) the target is a tenant-level system entity" clause
- the matching "**Exception: tenant-level system entities**" sentence in `data-fabric.md` § Folder Scope

`grep -rn "system entit" references/data-fabric/` now returns only this dangling clause and an unrelated FILE note at `:223`. The current matrix (`:205-211`) has no system-entity row, so an agent reading this rule has no way to evaluate the condition it hinges on — and that paragraph was the explanation for why FILE fields work on folder-scoped entities at all, which `integration_folder_binding.yaml` grades.

**Fix:** either end the sentence at "…tenant ↔ folder boundary." (`:198` already covers the FILE case) or restore one defining sentence plus the matrix row.

uipreliga (approved):
Fix what you agree with and 🚢

niket0503 (approved):

PR #2291, opened 2026-07-28
feat(uipath-api-workflow): evals contract + test-until-green loop, with tests

rares-baesu-uipath:
## Summary

Teaches `uipath-api-workflow` the **evals / test-until-green** capability of Studio Web's API-workflow **Evaluations** panel (`@uipath/unified-evals`), as a test-driven authoring flow, and adds the test tasks that grade it.

### How the skill behaves

**TDD gate (Critical Rule 22 — also the opening block of the skill and Core Principle 0).** On any create or edit request the agent first checks whether the project has an `evals/` folder (inside the project directory, next to `Workflow.json`):

- **No `evals/` folder** → the Evaluations feature is not enabled for that project. No test or loop-mode question, no folder creation — the workflow is authored normally.
- **`evals/` present** → the agent **ends its turn** with two questions before writing `Workflow.json` or any eval file:
  1. **Tests** — rows exist: *"should anything change, or use them as-is?"*; empty set: *"want me to add some?"* (with proposed cases).
  2. **Loop mode** — *"run the tests and re-try until every case passes, or author once and you verify?"* A "yes" is also the rule-21 consent for `--no-auth` runs.

  Once answered: eval set first, then the workflow, then the rows are run. **Later change requests that alter behavior** (thresholds, rounding, branch order …) re-open the question: the agent lists the rows the change would invalidate and asks before editing, then re-runs. Prompts that explicitly say not to ask skip the gate (existing tests kept as-is, author once, rows still run).

**Eval file contract** (`references/testing-and-evals.md`): `evals/<scope>/eval-sets/*.json` + `evals/<scope>/evaluators/*.json` (scope `default`; Studio Web seeds `evaluation-set.json` and `exact-match-evaluator.json` when the panel is first opened); `evaluatorRefs` / `evaluationCriterias` keys are evaluator **file base names**; `uipath-exact-match` is a strict deep-equal of the **raw** workflow output under `targetOutputKey` (`*` / dotted path / list); run results are in-memory only.

**Scoring recipe**: `uip api-workflow run` PascalCases every key in its printed `Data` (`grade` → `Grade`), so expectations are derived from the raw Response in the debug log (`--log-level debug --log-file`) or from `output.schema`, never from `Data`. Inside Studio Web, rows run through the Evaluations panel or consent-gated `RunProject`.

### Changes

- `skills/uipath-api-workflow/SKILL.md` — description (test/eval, `evals/` trigger, one-line edits load the skill); TDD gate block, Core Principle 0, rule 22; Phase 0 discovery of `evals/`; Phase 2 precondition; Quick Start step; When-to-Use bullets; anti-patterns.
- `skills/uipath-api-workflow/references/testing-and-evals.md` — new: file contract (§1), running/scoring a row (§2), the interactive protocol (§3), stale-expectation guard and gotchas (§4).
- `tests/tasks/uipath-api-workflow/evals/` — six local-only integration tasks (`feature:eval`), scored the way the panel scores:
  - `author_eval_set` (mode:build) — fill the panel-seeded empty dataset of a correct project; grades the file contract and that rows use the raw `grade` key.
  - `loop_until_green` (mode:diagnose) — correct dataset, workflow with two bugs (`Grade` casing, `>` vs `>=`); fix the workflow, leave `evals/` byte-for-byte untouched.
  - `stale_expectations` (mode:build) — lower a threshold 60 → 50 and re-derive only the row it invalidated.
  - `gate_create` / `gate_edit` / `gate_off_no_evals` (mode:build) — create on a panel-seeded project and a behaviour-changing edit on a project with tests: the agent must stop and ask; the same create on a project **without** `evals/`: the agent must simply author. Deterministic nothing-written checks plus an `llm_judge` over the final response (needs the Anthropic judge transport — runs in CI).
  - `_shared/` — parameterised fixture, panel-equivalent scorer, generic contract / rows-pass checks.

Scope: `skills/uipath-api-workflow/**` and `tests/tasks/uipath-api-workflow/**` only.

## Verification

- `hooks/validate-skill-descriptions.sh` (1018 ≤ 1024), `npm run skills:validate` + `skills:build`, `scripts/check-skill-status.py`, `scripts/check-cli-verbs.py` on the new tasks — all OK; no duplicate `task_id`s.
- `npm run skills:test`: 51/52 locally; the one failure (`pack builds complete… npm packages`, an `npm publish` dry-run assertion) fails identically on `origin/main` in this environment and is not among CI's failures.
- Each check script was dry-run against its seed in both directions (fails on the seeded/tampered state, passes on a correct fix).
- Ran the tasks locally with `coder-eval` (`experiments/default.yaml`, claude-code agent): `author_eval_set`, `loop_until_green`, `stale_expectations`, `gate_off_no_evals` — SUCCESS, score 1.000; the PR-gate smoke task `if_else_grade` passes unchanged. `gate_create` and `gate_edit` pass their deterministic criteria (nothing written; the final message asks the two questions, and on the edit path flags the stale `score 55` row); their `llm_judge` is left for CI. `gate_edit` is known-flaky for a reason outside the skill: in 1 of 4 local runs Claude Code did not load the skill for the one-line edit at all.


tests/tasks/uipath-api-workflow/evals/_shared/eval_scoring.py
```diff
@@ -0,0 +1,153 @@
...
+def score_project(project_dir, workflow_path=None):
+    """Run every row of every eval set through the workflow and score it with its
+    referenced evaluators. Returns a list of dicts; `verdict` is pass|fail|error."""
+    project_dir = Path(project_dir)
+    workflow_path = Path(workflow_path or project_dir / "Workflow.json")
+    evaluators = load_evaluators(project_dir)
+    results = []
+    for set_path in load_eval_sets(project_dir):
+        eval_set = json.loads(set_path.read_text())
+        refs = [str(r).removesuffix(".json") for r in eval_set.get("evaluatorRefs", [])]
```
liviubobocu: The shared scorer accepts evaluator refs the panel orphans. 
A dataset whose evaluationCriterias keys are "exact-match-evaluator.json" scores 0/3 in the panel and gets EXIT=0 from both check_eval_contract and check_rows_pass. That's 6.0 of the weight in loop_until_green + stale_expectations.

tests/tasks/uipath-api-workflow/evals/gate_create/check_gate.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python3
```
liviubobocu, line 1: Both gate tasks pass on an agent that does nothing. Reproduced directly:

$ python3 gate_edit/seed.py && python3 gate_edit/check_gate.py
OK: workflow and dataset untouched — the agent stopped at the tests gate
EXIT=0

skills/uipath-api-workflow/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
```
liviubobocu, line 1: Core Principle 0 orders a run the user never consented to. SKILL.md:41 says "once answered: eval set first, workflow second, then run the rows (TDD)" — dropping the loop-mode conditional that rule 22 has at :137 ("then run the rows (loop mode) or hand over (author once)"). Core Principles are at the top and always read; rule 22 is 90 lines down. Directly violates the skill's own rule 21.

skills/uipath-api-workflow/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
 name: uipath-api-workflow
```
liviubobocu, line 2: The "don't prompt me" escape hatch runs rows with no consent at all. SKILL.md:137 and testing-and-evals.md:100 both end with "author once, and run the existing rows before reporting." "Don't ask for confirmation" waives the gate questions; rule 21's run consent is a separate safety gate that rule 22 itself says only a loop-mode "yes" satisfies. Here neither was given.

skills/uipath-api-workflow/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
-description: "UiPath API Workflow assistant — author, run, validate, package, publish, deploy, and troubleshoot JSON workflows for `uip api-workflow`. Covers logical/hierarchical structure (Sequence, Assign, JavaScript, If with #Wrapper/#Then/#Else, ForEach, DoWhile, Break, TryCatch, Wait, Response — nested patterns) AND HTTP / Integration Service connector activities (Gmail, Outlook, GitHub, Slack) authored via `uip api-workflow registry resolve`/`stub`. Operate: run locally, manage IS connections (`uip is connections`), pack/publish/deploy via `uip solution`, invoke published workflows via HTTP/schedule/event triggers. Diagnose: validate → run --no-auth loop, root-cause run/expression/connection faults, inspect job logs & traces. Triggers on UiPath API workflows, project type \"Api\", JSON files with `document.dsl`/`do[]`, those activity types, or fetching from a public/vendor API. For .flow Maestro→uipath-maestro-flow. For .xaml/coded RPA→uipath-rpa. For coded agents→uipath-agents. For Coded Apps→uipath-coded-apps."
```
liviubobocu, line 3: an evals/ folder is claimed as a trigger, but it isn't an API-workflow signal. Low-code agents use evals/ too (skills/uipath-agents/references/lowcode/agent-definition.md:15), as do inline agents inside .flow projects. This PR makes uipath-api-workflow the only skill in the repo naming the "Studio Web Evaluations panel", while uipath-maestro-flow separately claims "run Studio Web evals". There's already a live cross-fire row in the repo's own recall set — uipath-agents-014: "I edited my evaluation set in Studio Web — pull the changes back to my local repo". Neither side has a redirect. The layouts actually differ and would disambiguate cheaply: api-workflow is evals/<scope>/eval-sets/, agents is evals/eval-sets/ with no scope segment.

skills/uipath-api-workflow/references/testing-and-evals.md
```diff
@@ -0,0 +1,133 @@
...
+   - **Report progress each iteration** — e.g. "iteration 2: 3/4 rows pass; fixing row 'unpaid invoice' (expected status `unpaid`, got `paid`)".
+   - **Stop conditions:** all green; OR no progress after a few iterations (cap it — see Infinite Loop Prevention in SKILL.md); OR the user interjects. Always end with a per-row summary of the final state.
+   - **Concurrency / interruptibility (optional but preferred):** run the run-and-score portion in a **background subagent** so the main thread stays responsive and can report progress. If a new user message changes the workflow or the intent, stop the in-flight run, incorporate the change, then restart the loop. Never keep scoring against a workflow the user has just asked you to change.
```
liviubobocu: The CLI is not slow. Two independent measurements agree: ~0.72–0.80s warm per row, ~1.2–1.6s cold; --log-level debug --log-file costs nothing measurable. A 5-row × 3-iteration loop is ~18s of compute. §3's "run the loop in a background subagent" is not justified by run cost.

What is expensive is turns, and the doc gives no guidance. Same 5×3 loop: 30 tool calls if you follow §2 literally (two fenced snippets = run call + scrape call, per row), 3 if you batch the whole iteration into one bash invocation. A 10× spread with nothing in the doc pointing either way.

skills/uipath-api-workflow/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
+description: "UiPath API Workflow assistant — author, run, validate, package, publish, deploy, and troubleshoot JSON workflows for `uip api-workflow`. Load for ANY create or edit of a `Workflow.json` / API workflow project, even one-line changes — with `evals/` present, tests come first (TDD). Covers structure (Sequence, Assign, JavaScript, If with #Wrapper/#Then/#Else, ForEach, DoWhile, Break, TryCatch, Wait, Response — nested) AND HTTP / IS connector activities via `uip api-workflow registry`. Operate: run locally, manage IS connections, pack/publish/deploy via `uip solution`, invoke via trigger. Test/eval: `evals/` datasets (exact-match, Studio Web Evaluations panel); loop until green. Diagnose: validate → run --no-auth loop, root-cause. Triggers on API workflows, project type \"Api\", JSON with `document.dsl`/`do[]`, those activity types, an `evals/` folder, or public API fetch. For .flow Maestro→uipath-maestro-flow. For .xaml/coded RPA→uipath-rpa. For coded agents→uipath-agents. For Coded Apps→uipath-coded-apps."
```
liviubobocu: Rule 22 mandates writing expectedOutput rows before the workflow. §2 says row keys must match the workflow's raw output casing, derivable only from a run or from output.schema. But uip api-workflow init scaffolds neither — it emits metadata.variables: [] with no input/output keys at all. So the agent must commit to key casing it has no way to derive, three steps before the artifact that defines it exists.

The dry-run agent hit exactly this on its first attempt while actively looking for it: proposed grade, authored Grade. And the trap is invisible from normal CLI output — on the broken workflow and the fixed one, --output json printed byte-identical Data: {"Grade":"PASS"}.

Worse, the doc diagnoses this and withholds the cure: it never supplies a comparator. §2 stops at printing the raw JSON, then §4 warns that "an LLM eyeballing a run … forgives key casing". Every agent will hand-roll a different comparison, and the dry-run agent eyeballed it too.

Fix: add an ordering step — declare output.schema in the skeleton first, derive row keys from it — and ship a copy-paste batch comparator. The comparator also fixes the turn-cost problem below.

liviubobocu (approved):

rares-baesu-uipath:
Addressed all seven review comments in the latest commit — each turned out to be real:

1. **Scorer accepted `<ref>.json` criteria keys** → `eval_scoring.py` / `check_eval_contract.py` are now strict: keys must equal the evaluator file base name, a `.json` suffix is a contract error, and a row with no criteria for a declared evaluator is an error (mirrors the panel's "No expected output could be matched"). Verified: a `.json`-keyed dataset now fails both checks and the shipped comparator; the correct key passes.
2. **Gate tasks passed on a do-nothing agent** → both gate tasks add a deterministic `command_executed` "agent inspected the project/evals" criterion, and the descriptions now say the `llm_judge` is the deciding criterion (the untouched-files check is only a necessary condition).
3. **Core Principle 0 ordered a run without consent** → rows run only in loop mode; "author once" means hand over without running.
4. **"Don't ask" ran rows with no consent** → it waives the two questions, not rule 21: rows run only when the request itself asks for a run/loop; otherwise hand over the command.
5. **`evals/` as a trigger** → description + trigger bullet key on the API-workflow layout (`evals/<scope>/eval-sets/` next to `Workflow.json`) and redirect agent evals (`evals/eval-sets/`, no scope) → `uipath-agents`, Flow evals → `uipath-maestro-flow`.
6. **"CLI is slow" / background subagent** → removed; replaced with turn-cost guidance: one iteration = one command.
7. **Row keys had no source + no comparator** → TDD order now starts with declaring `input.schema` / `output.schema`; §2 ships a copy-paste `score-evals.py` batch comparator (raw output from the debug log, strict deep-equal under `targetOutputKey`, file-base-name refs) that scores every row in one tool call.

PR #2293, opened 2026-07-28
fix(tests): grade simulated troubleshoot scenarios on the full dialog + CI gate

dmorosanu:
## Problem

Six `uipath-troubleshoot` scenarios failed the 2026-07-28 nightly (`2026-07-28_04-39-07`) with `skill_triggered` = 1.00 and `llm_judge` = **0.00**. In every case the agent had already produced the correct root cause and fix.

Their `llm_judge` sets `include_agent_output: true` **without** `include_dialog: true`, so the judge receives only the agent's *final* turn. Under `simulation.enabled: true` the simulated user keeps talking after the diagnosis lands, so the final turn is an acknowledgement. What the judge actually graded:

| Task | Graded turn | Diagnosis it never saw |
|---|---|---|
| `rpa-executor-service-disconnect` | `"Will do."` | Robot service shut down / lost Orchestrator connection mid-run -> rerun |
| `web-http-401-auth` | `"You're welcome. The corrective change is limited to that existing header expression..."` | Bare token in `Authorization`, no `Bearer ` prefix, with the exact expression fix |
| `rpa-is-folder-permission-cns` | `"Sounds good. Once the grant is in place..."` | `CNS1045` / 403, robot lacks `Connections.View` in Finance, 3-step grant |
| `rpa-preflight-failure` | `"Kept the investigation evidence in .local/investigations."` | Missing `PO_PricingTable` asset + inaccessible Outlook connection; `Throw` ruled out as propagation |
| `rpa-is-personal-workspace-connection-cns` | `"That should resolve the unattended failure..."` | Process bound to a personal-workspace OneDrive connection; create in shared folder, rebind, republish |
| `assign-linq-no-data-rows` | a blocked `uip rpa pack` report (turn 4 was a *publish* request) | Unguarded `CopyToDataTable()` on an empty LINQ result + `If(rows.Any(), ..., dtInvoices.Clone())` |

Judge findings confirm it, e.g. for `rpa-executor-service-disconnect`: *"Agent output is 'Will do.' - no root cause identified ... Zero diagnostic content provided."*

## Changes

1. **`include_dialog: true` on the six failing tasks.**
2. **Canonical judge block in `tests/tasks/uipath-troubleshoot/CLAUDE.md`** - the block authors are told to "copy verbatim into every new scenario" omitted `include_dialog`, while the prose below it promised the judge grades the diagnosis "wherever it appears in the dialog". That is the origin of the defect; without this change the new gate would block every scenario authored per the repo's own instructions.
3. **Full sweep: 188 more criteria across the whole troubleshoot tree**, including all 143 `skip: true` tasks so they are correct the moment anyone unskips them. 297 of 298 task YAMLs now carry the flag; the one holdout has no `simulation` block, where the final turn *is* the whole answer. Insertions are structural (walk `success_criteria`, only touch an `llm_judge` in a simulation-enabled task that sets `include_agent_output: true` and lacks `include_dialog`), not a text replace: diff is `+188/-0`, and every file parses and reads back the flag. **No `skip:` flag was changed.**
4. **`scripts/check-judge-dialog.py` + `.github/workflows/judge-dialog-gate.yml`** - deterministic gate modeled on `task-driver-gate.yml`, emitting `::error` annotations at the offending criterion's line.

## Gate scoping

Restricted to task YAMLs a PR **adds or modifies** (`--diff-filter=AM`). No `paths:` filter, so it can be wired as a required check - it exits 0 when a PR touches no troubleshoot task YAMLs. `workflow_dispatch` scans the whole skill.

Now that the tree is clean it could be switched to a full-tree scan (strictly stronger: catches an unflagged task arriving via a merge, not just one the author edited). Left PR-scoped for now; it is a three-line change.

## Validation

**Gate:** green on this PR (10s). Log confirms it is non-vacuous - the diff step resolved exactly the six task YAMLs, then `OK - every simulated task's llm_judge grades the full dialog.` Also verified locally: 6 correct annotations on the pre-fix state, clean on the fixed files, clean on empty input, and clean across all 298 files after the sweep.

**coder-eval, all six scenarios: 6/6 SUCCESS, every criterion 1.00**, no simulator failures, all clean `stop_token` (local, claude-sonnet-4-6 via Bedrock, experiment `default.yaml`):

| Task | turns | `llm_judge` nightly -> now | score |
|---|---|---|---|
| `rpa-preflight-failure` | 3 | 0.00 -> **1.00** | 1.000 |
| `web-http-401-auth` | 3 | 0.00 -> **1.00** | 1.000 |
| `assign-linq-no-data-rows` | 2 | 0.00 -> **1.00** | 1.000 |
| `rpa-executor-service-disconnect` | 1 | 0.00 -> **1.00** | 1.000 |
| `rpa-is-folder-permission-cns` | 1 | 0.00 -> **1.00** | 1.000 |
| `rpa-is-personal-workspace-connection-cns` | 1 | 0.00 -> **1.00** | 1.000 |

Three ran as genuine multi-turn dialogs (3, 3, 2 turns) - the exact condition the fix targets. `rpa-preflight-failure` is the sharpest case: its final turn's `agent_output` was the **empty string**, so the old config would have handed the judge nothing; with `include_dialog` it scored 1.00. Judge prompts now contain `REFERENCE SOLUTION` + `DIALOG` + `AGENT OUTPUT` where the nightly's had `AGENT OUTPUT` alone.

### Known-unrelated blocker in `run-coder-eval.yml`

A `Run Coder Eval` dispatch on this branch (codex arm) returned 4/6, but **all six tasks ended `turns=1 / stop=error / simulator_failures=1`**:

```
API Error (eu.anthropic.claude-sonnet-5): 400 The provided model identifier is invalid.
```

The user simulator sets `model=None` by design and inherits the Bedrock **route** model (`BEDROCK_MODEL` + `AWS_REGION` secrets), which currently resolve to an EU inference profile that does not carry sonnet-5. No workflow input can override it (`agent_model` only reaches the agent, which is why the codex agent ran while the simulator died; with `agent=claude` the agent would fail identically). That run therefore never produced a multi-turn dialog and is not evidence either way for this PR.

This blocks `run-coder-eval.yml` for **any** simulated task, well beyond this PR - it needs a secrets change (point `BEDROCK_MODEL` at a model valid in the configured `AWS_REGION`, e.g. the `us.anthropic.claude-sonnet-4-6` the nightly VM uses). The post-merge nightly runs in the VM environment where the simulator works, so it is the clean CI before/after.

## Not in scope

Two genuine nightly failures remain open: `connector-aggregate` (the `connection-not-resolved.md` playbook has no transient-5xx branch, so a 502 Bad Gateway is diagnosed as a broken connection - the real client-facing defect) and `o365-interactive-token-timeout` (agent fused two GUIDs into a nonexistent folder key; the permissive `unmocked_default: []` masked it as "no jobs").

### DIFF
--- a/tests/tasks/uipath-troubleshoot/CLAUDE.md
+++ b/tests/tasks/uipath-troubleshoot/CLAUDE.md
@@ -273,6 +273,7 @@ Every `llm_judge` criterion across all troubleshoot tasks uses the **same** prom
   weight: 3.0
   pass_threshold: 0.7
   include_reference: true
+  include_dialog: true
   include_agent_output: true
   prompt: |
     Grade the agent's final answer against the attached RESOLUTION.md.
@@ -289,12 +290,15 @@ Every `llm_judge` criterion across all troubleshoot tasks uses the **same** prom
     Return JSON: {"score": <float>, "rationale": "<one sentence>"}
 ```
 
-**What the judge sees** (both flags MUST be `true`):
+**What the judge sees** (all three flags MUST be `true`):
 
 - `include_reference: true` — passes `RESOLUTION.md` (the file named under `reference:` at the task root)
+- `include_dialog: true` — passes the full user<->agent dialog, every turn
 - `include_agent_output: true` — passes the agent's final user-facing response
 
 That is **all** the context the judge gets. The contract: agent's diagnosis (wherever it appears in the dialog) vs. RESOLUTION.md → score. Tool calls are deliberately excluded — the judge grades the presented diagnosis, not how it was reached.
+
+`include_dialog` is what makes "wherever it appears in the dialog" true. Without it the judge receives only the final turn, and the simulator keeps talking past the diagnosis — so the graded turn is an acknowledgement ("Will do.", "You're welcome.") and a correct investigation scores `0.00`. Six scenarios failed this way in the 2026-07-28 nightly. `scripts/check-judge-dialog.py` (CI: `judge-dialog-gate.yml`) blocks a PR that reintroduces it.
 
 **Forbidden on `llm_judge`:**
 
--- a/tests/tasks/uipath-troubleshoot/CLAUDE.md
+++ b/tests/tasks/uipath-troubleshoot/CLAUDE.md
@@ -325,6 +325,7 @@ success_criteria:
     weight: 3.0
     pass_threshold: 0.7
     include_reference: true
+    include_dialog: true
     include_agent_output: true
     prompt: |
       Grade the agent's final answer against the attached RESOLUTION.md.
--- a/tests/tasks/uipath-troubleshoot/activity-packages/classic-browserscope-efail/task.yaml
+++ b/tests/tasks/uipath-troubleshoot/activity-packages/classic-browserscope-efail/task.yaml
@@ -45,6 +45,7 @@ success_criteria:
     weight: 3.0
     pass_threshold: 0.7
     include_reference: true
+    include_dialog: true
     include_agent_output: true
     prompt: |
       Grade the agent's final answer against the attached RESOLUTION.md.
--- a/tests/tasks/uipath-troubleshoot/activity-packages/classic-movefile-source-not-found/task.yaml
+++ b/tests/tasks/uipath-troubleshoot/activity-packages/classic-movefile-source-not-found/task.yaml
@@ -47,6 +47,7 @@ success_criteria:
     weight: 3.0
     pass_threshold: 0.7
     include_reference: true
+    include_dialog: true
     include_agent_output: true
     prompt: |
       Grade the agent's final answer against the attached RESOLUTION.md.
--- a/tests/tasks/uipath-troubleshoot/activity-packages/classic-openapp-wrong-scope-type/task.yaml
+++ b/tests/tasks/uipath-troubleshoot/activity-packages/classic-openapp-wrong-scope-type/task.yaml
@@ -47,6 +47,7 @@ success_criteria:
     weight: 3.0
     pass_threshold: 0.7
     include_reference: true
+    include_dialog: true
     include_agent_output: true
     prompt: |
       Grade the agent's final answer against the attached RESOLUTION.md.
--- a/tests/tasks/uipath-troubleshoot/activity-packages/classic-runlocaltriggers-generated-missing/task.yaml
+++ b/tests/tasks/uipath-troubleshoot/activity-packages/classic-runlocaltriggers-generated-missing/task.yaml
@@ -58,6 +58,7 @@ success_criteria:
     weight: 3.0
     pass_threshold: 0.7
     include_reference: true
+    include_dialog: true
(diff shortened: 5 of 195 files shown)


dmorosanu:
Addressed both bot comments.

**1. Lean-default template missing `include_dialog` (lint bot, out-of-rubric) - fixed in b867e4a50.** This was a genuine miss and the same defect class as the root cause: I updated the canonical block at line 271 but not the "Lean default for every new scenario" snippet at line 323, which is the fuller `success_criteria` copy-paste and arguably the more likely one to be used. Swept the rest of the repo for the same pattern - `grep` over every `*.md` and `tests/templates/` shows these were the only two blocks, and both now carry the flag (lines 276 and 328).

**2. `criterion(a)` pluralization (Claude review nit) - fixed in 9b09b68f9.** Now switches noun and verb together, verified in all three cases:
```
FAIL - 1 llm_judge criterion grades the last turn only:
FAIL - 2 llm_judge criteria grade the last turn only:
OK - every simulated task's llm_judge grades the full dialog.
```

**3. Passing-run evidence - now in the PR description.** Both bot comments ran against the pre-sweep commit, before the results landed. All six scenarios: **6/6 SUCCESS, every criterion 1.00**, no simulator failures, clean `stop_token` (local, claude-sonnet-4-6 via Bedrock, experiment `default.yaml`). `llm_judge` went 0.00 -> 1.00 on all six. Three ran as genuine multi-turn dialogs (3, 3, 2 turns), which is the condition the fix targets; `rpa-preflight-failure`'s final turn was the empty string, so the old config would have graded nothing.

**Scope changed since those reviews:** the PR now also sweeps the remaining 188 criteria across the whole troubleshoot tree (including all 143 `skip: true` tasks), so the "48 remaining tasks" the review listed under *What's Missing* are handled here rather than as follow-up. No `skip:` flag was changed. Worth a reviewer's attention: 48 of the swept tasks are active and currently pass at 1.00, so their grading input changes - verified structurally (parse + flag readback + gate clean) but not re-run. Tonight's nightly is the real check on those.

Two open questions from the review that need a team decision, not a code change:
- **Gate as a required check** - it is deliberately `paths:`-filter-free so it can be wired into branch protection. Now that the tree is clean it could also be switched from PR-diff-scoped to a full-tree scan (strictly stronger, ~3 lines).
- **`run-coder-eval.yml` cannot currently run any simulated task.** A dispatch on this branch had all 6 tasks end `turns=1 / stop=error / simulator_failures=1` on `API Error (eu.anthropic.claude-sonnet-5): 400 The provided model identifier is invalid`. The simulator sets `model=None` by design and inherits the Bedrock route model from the `BEDROCK_MODEL` / `AWS_REGION` secrets, so no workflow input overrides it. Needs a secrets change; unrelated to this PR but blocks that workflow for every simulated task.

MarinRzv (approved):

costin-uipath (approved):

rockymadden (approved):
lgtm

PR #2296, opened 2026-07-28
feat(uipath-ixp): read the configured model + pre-processing [ACTV-88693]

andrei-uipath:
`[P1] [Smoke] Query model and preprocessing settings` — [ACTV-88693](https://uipath.atlassian.net/browse/ACTV-88693).

The skill documented `configure-model` (the write) but nothing for reading what a project is currently set to. Both improvisations an agent has are wrong:

- **`list-models`** — which the "Describe this project" row already pointed at for the "current model". Its `ModelName` is the trained labeller's *family* (`gemini_ixp`), never a `--model` value like `gemini_2_5_flash`, and it says nothing about pre-processing.
- **`configure-model`** — a read-modify-write that returns the merged config, so probing it to *find out* the current values rewrites them.

The settings are in `projects get-taxonomy` under `Data.dataset._model_config`. `model_version` holds the `--model` value verbatim; `input_config` stores the underlying mode and has to be inverted back to the `none|table_mini|table` token. `input_config: null` means never configured — **not** `none`, which is the explicit image-only setting.

## Changes

- `references/cli-reference.md` — new **Reading the current model and pre-processing** section with the inversion table; `get-taxonomy` row notes `_model_config`; `configure-model` row flags the read-modify-write; `list-models` row states what `ModelName` actually is.
- `SKILL.md` — Task Navigation row for the query, and the "Describe this project" row now says "current model **version**" so it stops reading as the extraction model.
- `tests/tasks/uipath-ixp/smoke/query_model_and_preprocessing.yaml` — grades the `get-taxonomy` read plus a negative guard that no `configure-model` write is produced.

## Verification

Mapping is not inferred from the CLI help text — `--preprocessing` → `input_config` is verified against `ProjectService.MapPreprocessing` in designtime-api, and `_model_config` against a live alpha `get-taxonomy` capture (`model_version: gemini_2_5_flash`, `input_config: null`). `ModelName: gemini_ixp` is confirmed from a live `list-models` capture and the generated `ModelName` enum.

**Ran `skill-ixp-query-model-preprocessing-smoke` in CI (run [30352701555](https://github.com/UiPath/skills/actions/runs/30352701555)) and it passed** — `status=SUCCESS score=1.000`, 2/2 criteria. That was the first real agent run: this workspace's API quota is exhausted until 2026-08-01, so every local `coder-eval run` dies `agent_api_error` before the agent starts (`status=ERROR`, not a content failure). Before CI, validation was offline:

- `coder-eval plan` — schema valid, 2 criteria.
- `check-task-driver.py` + `validate-skill-descriptions.sh` — pass.
- Pattern discrimination against hand-built `calls.log` fixtures: three correct shapes (get-taxonomy alone, with a preceding `projects get`, with extra `list-models` context) all score 6.0/6.0; `list-models`-only 3.0, destructive `configure-model` probe 3.0, probe-with-no-read 0.0.

Suite-level note: that run came in at 95.2% (40/42) against a 95% gate, with `per-occurrence-confirm` and `update-prompts-heredoc` failing. Both are pre-existing flakes, not this PR — the earlier ACTV-88681 runs on an untouched tree failed `update-prompts-heredoc` (0.67) and `confirm-predictions` (0.78) at 95.1%, and the failing set rotates between runs. None of the flaky tasks touch the model-config rows changed here.

Note the last bullet grades patterns directly — `coder-eval evaluate` reseeds `mocks/` from `template_sources`, so it never reads a hand-placed `calls.log`.

[ACTV-88693]: https://uipath.atlassian.net/browse/ACTV-88693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-ixp/references/cli-reference.md
```diff
@@ -14,15 +14,38 @@ All commands use `uip ixp` prefix. Always append `--output json` when parsing ou
...
-| `uip ixp projects configure-model <project-name> [options] --output json` | Configure extraction model. Options: `--model` (gemini_2_5_flash/gemini_2_5_pro/gpt_4o_2024_05_13) and `--preprocessing` (none/table_mini/table). |
-| `uip ixp projects list-models <project-name> --output json` | List all model versions and tags. Returns `Models[]` (`Version`, `ModelName`, `Pinned`, `TrainedTime`, `Description`), `Tags[]` (`Name`, `Version`, `UpdatedAt`), and `MaxPublished`. |
+| `uip ixp projects configure-model <project-name> [options] --output json` | Configure extraction model. Options: `--model` (gemini_2_5_flash/gemini_2_5_pro/gpt_4o_2024_05_13) and `--preprocessing` (none/table_mini/table). At least one is required. **Write-only, and a read-modify-write** — it merges what you pass into the current config and returns the merged result, so calling it to *find out* the current settings rewrites them. To read them, use [`get-taxonomy`](#reading-the-current-model-and-pre-processing). |
```
andrei-uipath: why is this needed: At least one is required. **Write-only, and a read-modify-write** — it merges what you pass into the current config and returns the merged result, so calling it to *find out* the current settings rewrites them.?

andrei-uipath: Fair — removed in 2e1b4d7, leaving just the cross-reference.

It was the third copy of the same warning: the SKILL.md Task Navigation row and the new reference section both already say `configure-model` is a read-modify-write and must not be used to read. And an agent that's *on* this row has already decided to write — the one who needs the warning arrives via Task Navigation or the section. Same argument that took Critical Rule 18 out of #2288.

I also dropped "At least one is required" — it's true (the CLI throws `Nothing to configure. Pass --model and/or --preprocessing.`) but it's unrelated to this ticket. Happy to add it on its own if you want the row to document it.

alexandrujircan (approved):

PR #2299, opened 2026-07-28
test(uipath-troubleshoot): simulated user declines fix application in scenario sims

dmorosanu:
## Why

Nightly duration analysis (2026-07-28 runs) showed the claude arm spends 2x the wall-clock of the codex arm on the uipath-troubleshoot suite (817m vs 403m over 155 tasks). A third of the gap comes from 39 scenarios where the agent, after delivering the diagnosis, offers to apply the fix and the simulated user approves - pulling the run into a full edit + validate/build/pack workflow (8-31m per task) that the judge never grades: the scenario contract is skill_triggered + llm_judge on the diagnosis vs RESOLUTION.md only.

## What

- Append one constraint to `simulation.constraints` in all 296 scenario task.yamls: the simulated user DECLINES fix application (source edits, validate/build/pack/publish) and asks only for the root cause and exact fix steps. Wording follows the existing precedent in `activity-packages/uia-application-not-found/task.yaml` (already had it; untouched).
- Same line added to the generator template in `_shared/scripts/generate_scenario.py` so new scenarios inherit it.

## Validation

Local re-runs of the four worst nightly offenders (sequential, sonnet-4-6 via bedrock), all SUCCESS, zero source edits and zero build commands in transcripts:

| Task | Nightly claude (old sim) | Re-run (new sim) |
|---|---|---|
| replace-text-silent | 31.1m, fix workflow entered | 11.9m, score 0.85, clean |
| excel-rr-sheet-dynamic-empty | 21.4m, fix workflow | 6.6m, score 1.00, clean |
| excel-rr-sheet-bytes | 14.2m, fix workflow | 3.5m, score 1.00, clean |
| uia-type-into-method-not-found | 16.9m, fix workflow | 6.0m, score 0.85, clean |

Runs of the remaining 35 affected scenarios are in progress; results will be posted on this PR.

PR #2301, opened 2026-07-28
ci: block troubleshoot tasks whose llm_judge grades only the last turn

MarinRzv:
## Problem

With `simulation.enabled: true` the simulated user keeps talking after the agent has delivered its diagnosis ("thanks, I'll get an admin to grant that"). The agent's **final** turn is then a one-line acknowledgement — "Will do.", "You're welcome." — while the real diagnosis sits two turns back.

An `llm_judge` criterion that sets `include_agent_output: true` **without** `include_dialog: true` receives only that final turn. The judge sees no diagnosis and scores `0.00`, so a fully correct investigation fails the task.

Six troubleshoot tasks failed exactly this way in the 2026-07-28 nightly, each with a correct root cause in turn 1.

## Change

| File | Purpose |
|------|---------|
| `scripts/check-judge-dialog.py` | Flags `llm_judge` criteria that grade the last turn only. Accepts task YAML paths or a directory. |
| `.github/workflows/judge-dialog-gate.yml` | Runs the check on `pull_request` (and `workflow_dispatch`), scoped to the troubleshoot task YAMLs the PR adds or modifies. |

A task is reported when **all** of these hold:

1. `simulation.enabled` is `true` — multi-turn, so a last-turn-only grade can miss the diagnosis. Single-turn tasks are fine; their last turn *is* the answer.
2. It has an `llm_judge` criterion with `include_agent_output: true`.
3. That criterion does not set `include_dialog: true`.

Failures are emitted as `::error file=…,line=…` annotations, so they render inline on the PR diff at the offending `- type: llm_judge` line.

## Scoping

The gate checks only the task YAMLs a PR **adds or modifies** (`--diff-filter=AM`), so the ~194 existing tasks that predate this rule are not retroactively blocked. `workflow_dispatch` scans the whole skill directory instead, for a full sweep on demand.

There is no `paths:` filter on the workflow, so it can be wired as a required check — it exits 0 when a PR touches no troubleshoot task YAMLs. To cover other skills later, widen the pathspec in the "Collect changed task YAMLs" step.

## Verification

Run locally (`pip install pyyaml`):

```bash
# current main — every task still grades the last turn only
python3 scripts/check-judge-dialog.py tests/tasks/uipath-troubleshoot
# FAIL - 194 llm_judge criteria grade the last turn only:   (exit 1)

# a task fixed by #2293
python3 scripts/check-judge-dialog.py <fixed-task.yaml>
# OK - every simulated task's llm_judge grades the full dialog.   (exit 0)

# no paths (PR touches no tasks)
python3 scripts/check-judge-dialog.py
# OK - ...   (exit 0)
```

All three paths confirmed, plus the singular-grammar branch (`1 llm_judge criterion grades`).

## Relationship to #2293

Split out of #2293, which now carries only the task-YAML fixes (`include_dialog: true` on 194 tasks + the `CLAUDE.md` judge template). The two can merge in either order — because the gate is scoped to changed files, merging this first does not retroactively fail anything.


dmorosanu:
## Added: troubleshoot scenario contract gate (moved here from #2311)

Pushed `33af29dc4` onto this branch, plus a merge of latest `main` (`1bfaebcb0`) which this branch needed — it was based on a commit predating #2299, so the new gate could not pass without it. Both gates are green on the merged tree:

```
$ python3 scripts/check-troubleshoot-tasks.py
OK - 297 scenario(s) + the generator template satisfy the troubleshoot scenario contract.

$ python3 scripts/check-judge-dialog.py
OK - every simulated task's llm_judge grades the full dialog.
```

### Why

The troubleshoot suite is ~300 near-identical faithful-replay scenarios grown from one generator template, and nothing verified they stayed in that shape. The drift that motivated this is expensive rather than loud: in the 2026-07-28 nightly, 39 scenarios let the simulated user approve fix application, sending runs into an ungraded edit + validate/build/pack loop worth ~126 minutes of wall clock. Every one of those tasks still passed, so no signal existed. #2299 fixed the 296 files; this adds the gate that keeps them fixed.

### What

- **`tests/tasks/uipath-troubleshoot/_shared/scenario-contract.yaml`** — the machine-readable master contract: required simulation constraints (currently the decline-fix constraint from #2299), required + forbidden success-criteria types, the canonical `llm_judge` shape (`weight`/`pass_threshold`/`include_reference`/`include_agent_output`, no `files:`), structural keys (`reference.file`, `sandbox.mock_path_dirs`, required tags, `run_limits`), and the exemption list (`smoke-manifest-commands`, the one non-scenario task).
- **`scripts/check-troubleshoot-tasks.py`** — enforces the contract against every scenario **and against the generator template**, which it renders as a synthetic scenario. A rule therefore cannot apply to the existing scenarios but not to the generator that produces new ones.
- **`.github/workflows/troubleshoot-task-contract.yml`** — PR gate, no `paths:` filter so it can be wired as a required check (sub-second full scan, definitive pass/fail). Same shape as `task-driver-gate.yml`.
- Suite `CLAUDE.md` points at the contract and the local command.

Rules were derived from what is already documented as MUST in the suite CLAUDE.md **and** currently holds for 100% of scenarios, so the gate is green from the first commit. Constraints that are only mostly-followed (`Never invent...`, `Never instruct the agent to stop...`, 3-5 scenarios each) were deliberately left out rather than enforced retroactively.

### Verification

Negative tests — injected drift, all correctly failed with an annotated line + fix hint:

| Injected drift | Caught |
|---|---|
| decline-fix constraint removed from a scenario | yes |
| decline-fix constraint removed from the generator template only | yes |
| `file_exists` criterion added | yes |
| `llm_judge.include_reference: false` | yes |
| `llm_judge.weight: 2.0` | yes |
| `mock_path_dirs: [mocks]` | yes |
| `uipath-troubleshoot` tag removed | yes |

### Possible consolidation

`check-judge-dialog.py`'s rule (llm_judge must grade the full dialog, not only the last turn) is the same kind of suite-wide invariant the contract file now expresses declaratively. Folding it in as a contract rule would leave one gate instead of two — happy to do that here or as a follow-up, your call since this is your PR.

### Pre-existing issues found, deliberately NOT changed

1. The generator template pins `sandbox.driver: tempdir`, which `scripts/check-task-driver.py` blocks for task YAMLs. That gate only scans `tests/tasks/**.yaml`, so the template inside a `.py` file escapes it — newly generated scenarios would carry a forbidden driver until someone notices.
2. `scripts/check-task-driver.py` calls `read_text()` with no encoding, so it crashes with `UnicodeDecodeError` on a Windows shell (CI is unaffected; Linux defaults to UTF-8).

dmorosanu:
## Consolidated the judge-dialog rule + fixed three pre-existing bugs

Three commits pushed (`fd778cc9a`, `abc9e1f9c`, `50c5232a6`).

### 1. `ci(troubleshoot): fold the judge-dialog rule into the scenario contract`

`include_dialog: true` is now a required `llm_judge` flag in `_shared/scenario-contract.yaml`, enforced by `check-troubleshoot-tasks.py`. `scripts/check-judge-dialog.py` and `.github/workflows/judge-dialog-gate.yml` are removed — **227 net lines deleted**, one gate instead of two.

Why this is a clean fold rather than a lossy one:

- All **297/297** simulated scenarios already set `include_dialog: true`, so the rule lands green with no migration.
- The standalone check's guard clauses (`simulation.enabled` true, `include_agent_output` true) are unconditionally true inside this contract — it already requires both on every scenario — so the rule collapses to one entry in `require_true` with no conditional machinery.
- Enforcement strengthens from changed-files-only to a full 297-scenario scan, which catches drift a path filter misses (a bad merge, a revert, a bulk edit).
- The rationale from the original docstring (six tasks scoring 0.00 in the 2026-07-28 nightly with correct root causes in turn 1) is preserved verbatim as the rule's `why` in the contract file, and the suite `CLAUDE.md` pointer now names the consolidated gate.

**This fold also caught a real bug:** the generator template never emitted `include_dialog`, so every newly generated scenario would have shipped with exactly the last-turn-only grading defect this rule exists to prevent. Added to the template.

Also improved the checker while validating the fold: a *missing* required flag has no line of its own, so annotations were landing on line 0 — they now anchor on the `- type: llm_judge` line, and each flag carries its own fix hint instead of a shared generic one.

### 2. `fix(troubleshoot): drop the forbidden sandbox.driver pin from the scenario template`

The generator template pinned `sandbox.driver: tempdir`, which `scripts/check-task-driver.py` blocks for task YAMLs — but that gate only scans `tests/tasks/**.yaml`, so a template inside a `.py` file escaped it. Every newly generated scenario would have carried a forbidden driver that opts out of the docker image and runs on the toolless host. Rendered template now yields `sandbox: [python, template_sources, mock_path_dirs]`, no `driver`.

### 3. `fix(scripts): read task YAMLs as UTF-8 in the driver check`

`check-task-driver.py` called `read_text()` with no encoding, so it died with `UnicodeDecodeError` on any non-ASCII task YAML under a Windows shell (CI was unaffected — Linux defaults to UTF-8), making the check unrunnable locally on Windows. Now reads UTF-8 explicitly and tolerates a decode error the same way it already tolerated malformed YAML.

### Verification

```
$ python3 scripts/check-troubleshoot-tasks.py
OK - 297 scenario(s) + the generator template satisfy the troubleshoot scenario contract.

$ python3 scripts/check-task-driver.py tests/tasks     # previously crashed on Windows
OK - no task pins `sandbox.driver: tempdir`.
```

Negative tests for the folded rule, both correctly failing with an annotated line and the specific rationale:

| Injected drift | Caught |
|---|---|
| `include_dialog` removed from a scenario | yes — annotated on the judge criterion line |
| `include_dialog` removed from the generator template only | yes |

PR #2307, opened 2026-07-28
test(admin): add outcome-graded user-delete smoke (PLT-107839 follow-up)

t-hsia:
## What

Adds `identity_delete_user_smoke` — the deferred **user-delete** coverage from the identity migration (#2280, PLT-107839).

The `identity_user_lifecycle_e2e` test grades the *update* tail but exercises delete only as teardown (a single read-back can't grade both the updated state and a subsequent delete — the delete erases the evidence). This dedicated smoke grades the **delete** deterministically.

## How it's graded (agent-agnostic, outcome-based)

Mirrors the blessed `identity_pat_revoke_smoke` pattern:

1. **`pre_run`** (`setup_delete_user.py`) — invites a distinct marker user (`ce-identity-delete@example.com`) and records its id to a temp state file.
2. **agent** — deletes the user (`list → identify → delete`).
3. **verify** (`verify_user_deleted.py`) — asserts that exact seeded id is **gone** from the live tenant. The state file proves the seed existed, so a do-nothing agent **fails** rather than trivially passing on absence.
4. **`post_run`** (`cleanup_delete_user.py`) — deletes any leftover user + clears state (idempotent; distinct marker keeps it `-jN`-safe alongside the lifecycle test's `ce-identity-lifecycle` marker).

No `command_executed` / `skill_triggered` — consistent with the migration.

## Validation

- `/lint-task`: **OK** (0 findings).
- Live coder-eval run: _pending — will post the passing verdict below._

### DIFF
--- a/tests/tasks/uipath-admin/cleanup_delete_user.py
+++ b/tests/tasks/uipath-admin/cleanup_delete_user.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python3
+"""Best-effort cleanup: delete the delete-smoke marker user if the agent left it
+behind (e.g. it never ran the delete), so the test does not leak users onto the
+org. Also clears the seed state file. Idempotent; always exits 0."""
+
+import logging
+import os
+import sys
+import tempfile
+
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), '_shared'))
+from admin_helpers import run_cli
+
+logging.basicConfig(level=logging.INFO, format="cleanup_delete_user: %(message)s")
+logger = logging.getLogger(__name__)
+
+MARKER_EMAIL = "ce-identity-delete@example.com"
+SEARCH = "ce-identity-delete"
+STATE_FILE = os.path.join(tempfile.gettempdir(), "ce_delete_user_seed.txt")
+
+
+def main():
+    data = run_cli(["admin", "users", "list", "--search", SEARCH])
+    if data and data.get("Result") == "Success":
+        for u in data.get("Data", []):
+            if (u.get("Email") or u.get("email") or "").lower() == MARKER_EMAIL:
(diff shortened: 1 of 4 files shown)


t-hsia:
## Validation + review complete — ready for review

### Live coder-eval (alpha, sequential runs)
On the final hardened commit `f1011502f`, both agents pass:

| Agent | Result |
|---|---|
| codex | 1/1 PASS · score 1.000 |
| claude | 1/1 PASS · score 1.000 |

Agent-agnostic, as PLT-107839 requires. (Earlier commits also passed codex+claude at each stage.)

### Critical review — 3-reviewer council
Convened a council with three independent adversarial lenses. Consolidated: one FIX-BEFORE-SHIP (correctness) + two SHIP-WITH-NITS, all addressed in `f1011502f`:

- **[HIGH] get-by-id guard soundness** — a transient CLI failure surfaces as `None` just like a genuine 404, so a single blip during a rename-attack silently skipped the anti-cheat guard. Now retried 3×: a live renamed user surfaces on retry; a real not-found stays `None` across all attempts (and the search check already proved the CLI is up). Lone false-pass closed.
- **[MED] flake hardening** — delete-propagation poll 4→6 attempts (bounded 15s/call); setup waits for leftover-delete to propagate before re-inviting the same email (no "already exists" seed collision); `pre_run`/criterion timeouts 60/40→90/90s.
- **[LOW] parity/docs** — added `delete` verb tag (parity with `pat_revoke`'s `revoke`); documented the single-marker/serial-run constraint in the setup docstring.

Two facts that shaped the fixes: `uip admin users list` **paginates**, so the `--search`-scoped check (not an unfiltered list) is the pagination-safe primary signal; and the passing runs empirically prove `users get` on a deleted id returns no live record, so soft-delete tombstones are a non-issue on this tenant.

yadvender-uipath (approved):

PR #2309, opened 2026-07-28
chore (ipe-evals): fix ipe eval failures for codex (#ENGCE-60054)

mrstark14:
## Summary

Fixes IPE eval failures surfaced by the codex evalboard run [`2026-07-28_04-39-07`](https://coder-evalboard.uipath-dev.com/runs/2026-07-28_04-39-07) (#ENGCE-60054). Two classes of fixes: tasks that failed because the tenant lacks the connector connection the prompt implied, and grading regexes/checks that produced false negatives on legitimate agent behavior.

### `connector_features/multiselect.yaml` — rewritten (+ new `check_multiselect_flow.py`)
The Act! 365 create-contact scenario failed on codex because the tenant has **no enabled `uipath-act-act365` connection** — the agent halted asking for connector prerequisites instead of building ([failed run](https://coder-evalboard.uipath-dev.com/runs/2026-07-28_04-39-07/skill-flow-ipe-multiselect)). The old criteria also never graded the multiselect field itself.

Rewritten to a Slack group direct message (`uipath-salesforce-slack` / `create-group-direct-message`) using the connection that **does** exist in `Shared/uipath-maestro-flow`. The `users` field is the multiselect under test. New sidecar `check_multiselect_flow.py` grades the actual feature: Slack connector node present **and** the `users` multiselect carries exactly 3 entries (name-agnostic across `**/*.flow`).

### Connection-folder hint added (1 line each)
`dtl_load_by_default_true.yaml`, `enum.yaml`, `searchable_joins.yaml` — prompts now say "Use the connection present in `Shared/uipath-maestro-flow`" so agents resolve the intended tenant connection instead of searching blind.

### `single_node/openmeteo_weather` — accept any connector activity flavor
Run `2026-07-28_04-39-07` false-failed agents that legitimately reached the same API via the generic `get-record` activity over `V1Forecast`. Prompt + `check_openmeteo_weather.py` now accept any `uipath.connector.custom-codereval-openmeteoapis.*` node (curated `getcurrentweather` or generic), still rejecting `core.action.http.v2` proxies and mocks.

### `single_node/outlook_trigger_inbox.yaml` — regex false-negative fix
The `command_executed` patterns required the literal connector key; agents that passed it via a shell variable (`KEY=...; uip is triggers objects "$KEY" EMAIL_RECEIVED`) were docked. Patterns now accept `$VAR`/`${VAR}` or the literal key.

### `connector_features/jdbc_databricks_query.yaml` — drop skill-load criterion
Removed the `command_executed` check on the Skill tool — grading skill activation is not an outcome check.

## Test runs — multiselect (rewritten task)

Local `coder-eval` runs against `experiments/default.yaml`, **6/6 passing at score 1.00** (1 confirmation + 5 parallel):

| Run directory | Status | Score | Duration |
|---|---|---|---|
| `runs/2026-07-29_02-37-21` | SUCCESS | 1.0 | 369s |
| `runs/2026-07-29_02-41-01` | SUCCESS | 1.0 | 562s |
| `runs/2026-07-29_02-41-06` | SUCCESS | 1.0 | 428s |
| `runs/2026-07-29_02-41-10` | SUCCESS | 1.0 | 349s |
| `runs/2026-07-29_02-41-18` | SUCCESS | 1.0 | 380s |
| `runs/2026-07-29_02-41-22` | SUCCESS | 1.0 | 375s |

All three criteria passed in every run: valid `.flow` produced, Slack group-DM node with 3-entry `users` multiselect, and `uip maestro flow validate` executed. An exploratory pre-rewrite run (`runs/2026-07-29_02-19-59`) verified the agent resolves the Slack connection and recipients before the criteria were authored.

Also re-verified locally at 3/3 score 1.00 each: `dtl_load_by_default_true` (`runs/2026-07-28_18-16-53/-58/-17-03`) and `openmeteo_weather` (`runs/2026-07-29_01-23-09/-12/-17`).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -79,7 +79,7 @@
 /tests/tasks/uipath-solution/ @UiPath/team-merlot @UiPath/team-orange
 
 # Flow skill
-/skills/uipath-maestro-flow/ @bai-uipath @rockymadden @gozhang2 @tmatup @akshaylive @jiyangzh @nikhil-maryala @baishalighosh @DevMomo
+/skills/uipath-maestro-flow/ @bai-uipath @rockymadden @gozhang2 @tmatup @akshaylive @jiyangzh @nikhil-maryala @baishalighosh @DevMomo @chandusailella @mukundbayyaram
 # Flow test tasks (general) — more specific subfolder rules below override this
 /tests/tasks/uipath-maestro-flow/ @bai-uipath @rockymadden @gozhang2 @tmatup @akshaylive @jiyangzh @nikhil-maryala @baishalighosh @DevMomo
 
@@ -117,7 +117,7 @@
 /skills/uipath-maestro-flow/references/plugins/connector/ @baishalighosh @chandusailella @bai-uipath @rockymadden @DevMomo
 /skills/uipath-maestro-flow/references/plugins/connector-trigger/ @baishalighosh @chandusailella @bai-uipath @rockymadden @DevMomo
 /skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/ @AlexandruCGhimisi @andreibalas-uipath @al3xanndru
-/tests/tasks/uipath-maestro-flow/connector_features/ @baishalighosh
+/tests/tasks/uipath-maestro-flow/connector_features/ @baishalighosh @chandusailella @mukundbayyaram
 
 # IXP
 /skills/uipath-maestro-flow/references/author/references/plugins/ixp/ @jcalero @constantinmuraru @scallaway-uipath @tommilligan @joe-prosser @vlad-crisan @richsilv @AWilcke
--- a/tests/tasks/uipath-maestro-flow/connector_features/check_multiselect_flow.py
+++ b/tests/tasks/uipath-maestro-flow/connector_features/check_multiselect_flow.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+"""Check the Slack group-DM multiselect flow.
+
+Passes (exit 0) when any .flow file in the workspace has:
+  1. a Slack connector node (type contains 'slack'), and
+  2. a 'users' multiselect field on that node with exactly 3 entries.
+
+Name-agnostic: the prompt does not fix a project name, so every .flow file
+is checked ('**' glob skips dot-dirs, so .uipath/.skills flows are ignored).
+"""
+import glob
+import json
+import sys
+
+
+def find_users_list(obj):
+    """Recursively find a 'users' key holding a list."""
+    if isinstance(obj, dict):
+        for key, value in obj.items():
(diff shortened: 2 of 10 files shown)


bai-uipath (approved):
**Right diagnosis, two matchers need tightening before merge.** The jdbc removal (only `tool_name: "Skill"` criterion in the suite) and the openmeteo loosening (`assert_flow_uses_connector_target` and live debug still gate it) are both clean.

- **`check_multiselect_flow.py` is too loose to be the task's only grader.** Verified: a `core.logic.mock.slack` node passes, `send-message` passes, `users: ["a","b","c"]` passes, and a decoy `queryParameters.filter.users` beats the real `bodyParameters.users` because `find_users_list` walks unordered. **Fix:** exclude mock/HTTP types, pin `create-group-direct-message`, read `inputs.detail.bodyParameters.users` directly, require three distinct entries. There's no `flow debug` behind this one, so precision matters more than usual.

- **Drop `triggers objects` and `triggers describe` from the gate (`weight: 0`) rather than hardening their regexes.** Skipping describe means no `parentFolderId`, which `check_folder_id_fresh` already fails on precisely, so they double-charge one defect and add a second way to fail a correct flow built via another discovery route. That also dissolves the false positive the loosening introduced: `KEY=uipath-google-gmail; uip is triggers objects "$KEY" EMAIL_RECEIVED` now passes an Outlook criterion (verified). If you'd rather keep them gating, use `(?s)\A(?=.*uipath-microsoft-outlook365)(?=.*uip\s+is\s+triggers\s+objects).*EMAIL_RECEIVED`.

- **Keep `resources run list ... MailFolder` as a command check.** Exchange folder IDs are stable per mailbox, so a remembered same-connection ID passes the freshness check and the artifact genuinely can't distinguish it from a fresh resolve. That's the one criterion here where the transcript is the only instrument.

- **`complex_array.yaml` has the identical Act!365 root cause and is untouched**, and it's the only other Act!365 task, so it keeps failing the same way. Consider `check_connection_available.py` as a precondition on both (`enum.yaml:37`, `path_params.yaml:37` already do this) instead of rewriting the scenario, which keeps the coverage.

**Minor:** `check_multiselect_flow.py:52` returns instead of continuing, so a wrong first Slack node hides a correct second. Event-as-variable is still a false negative. `multiselect.yaml` keeps `expected_turns: 37` from the multi-node version and hardcodes a personal address where siblings use `is-sandboxes`. Connection hint in `dtl_load_by_default_true.yaml:19` lands after "Validate the final flow file"; siblings put env facts in the opener. Worth confirming `users` is genuinely the IPE multiselect on `create-group-direct-message`, since nothing in the repo documents it.

chandusailella (approved):

aoltean16 (approved):

rockymadden (approved):

PR #2311, opened 2026-07-29
ci(troubleshoot): gate scenario task YAMLs against the suite contract

dmorosanu:
## Why

The troubleshoot suite is ~300 near-identical faithful-replay scenarios grown from one generator template, and nothing verified they stayed in that shape. The drift that motivated this is expensive rather than loud: in the 2026-07-28 nightly, 39 scenarios let the simulated user approve fix application, sending runs into an ungraded edit + validate/build/pack loop worth ~126 minutes of wall clock. Every one of those tasks still passed, so no signal existed. #2299 fixed the 296 files; this adds the gate that keeps them fixed.

## What

- **`tests/tasks/uipath-troubleshoot/_shared/scenario-contract.yaml`** — the machine-readable master contract: required simulation constraints (currently the decline-fix constraint from #2299), required + forbidden success-criteria types, the canonical `llm_judge` shape (`weight`/`pass_threshold`/`include_reference`/`include_agent_output`, no `files:`), structural keys (`reference.file`, `sandbox.mock_path_dirs`, required tags, `run_limits`), and the exemption list (`smoke-manifest-commands`, the one non-scenario task).
- **`scripts/check-troubleshoot-tasks.py`** — enforces the contract against every scenario **and against the generator template**, which it renders as a synthetic scenario. A rule therefore cannot apply to the existing scenarios but not to the generator that produces new ones.
- **`.github/workflows/troubleshoot-task-contract.yml`** — PR gate, no `paths:` filter so it can be wired as a required check (sub-second full scan, definitive pass/fail). Same shape as `task-driver-gate.yml`.
- Suite `CLAUDE.md` points at the contract and the local command.

Rules were derived from what is already documented as MUST in the suite CLAUDE.md **and** currently holds for 100% of scenarios, so the gate is green on main from the first commit. Constraints that are only mostly-followed (`Never invent...`, `Never instruct the agent to stop...`, 3-5 scenarios each) were deliberately left out rather than enforced retroactively.

## Verification

```
$ python3 scripts/check-troubleshoot-tasks.py
OK - 297 scenario(s) + the generator template satisfy the troubleshoot scenario contract.
```

Negative tests (all correctly fail with an annotated line + fix hint):

| Injected drift | Caught |
|---|---|
| decline-fix constraint removed from a scenario | yes |
| decline-fix constraint removed from the generator template only | yes |
| `file_exists` criterion added | yes |
| `llm_judge.include_reference: false` | yes |
| `llm_judge.weight: 2.0` | yes |
| `mock_path_dirs: [mocks]` | yes |
| `uipath-troubleshoot` tag removed | yes |

## Notes for the reviewer

Two pre-existing issues found while building this, deliberately NOT changed here:

1. The generator template pins `sandbox.driver: tempdir`, which `scripts/check-task-driver.py` blocks for task YAMLs. The existing gate only scans `tests/tasks/**.yaml`, so the template escapes it — newly generated scenarios would carry a forbidden driver until someone notices.
2. `scripts/check-task-driver.py` calls `read_text()` with no encoding, so it crashes with `UnicodeDecodeError` on a Windows shell (CI is unaffected because Linux defaults to UTF-8).

### DIFF
--- a/.github/workflows/troubleshoot-task-contract.yml
+++ b/.github/workflows/troubleshoot-task-contract.yml
@@ -1,0 +1,42 @@
+name: Troubleshoot Task Contract
+
+# Fails the build if any uipath-troubleshoot scenario task YAML drifts from the
+# suite contract in tests/tasks/uipath-troubleshoot/_shared/scenario-contract.yaml
+# (required simulation constraints, required/forbidden success-criteria types,
+# canonical llm_judge shape, structural keys). The generator template in
+# _shared/scripts/generate_scenario.py is validated as a synthetic scenario in
+# the same pass, so a rule can never apply to the scenarios but not the
+# generator that produces new ones.
+#
+# Motivating regression: 39 scenarios let the simulated user approve fix
+# application, sending nightly runs into an ungraded edit + validate/build/pack
+# loop worth ~126 minutes of wall clock. Nothing flagged it — the tasks still
+# passed. See scripts/check-troubleshoot-tasks.py for the full rationale.
+#
+# No `paths:` filter so this can be wired as a required check — it is a
+# sub-second full scan that always reports a definitive pass/fail.
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: troubleshoot-task-contract-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  gate:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


dmorosanu:
dup https://github.com/UiPath/skills/pull/2301

dmorosanu:
Superseded by #2301 — the commit was moved onto `ci/judge-dialog-gate` as `33af29dc4` (along with a merge of latest main that branch needed). Full description posted there: https://github.com/UiPath/skills/pull/2301#issuecomment-5114574748

PR #2312, opened 2026-07-29
fix(ixp): occurrence indices are read-scoped, confirmed rows sort first

andrei-uipath:
`get-predictions` does not return a repeatable group's rows in a fixed order. The server pairs annotations with predictions and lists the **matched pairs first**, so confirming one row of `Line Items` hoists it to `Occurrence` 0 on the next read and renumbers the rest — the IXP UI shows it first too. Values and page locations are unaffected; only the position in the read moves.

The skill said the opposite: *"read `Occurrence` directly; it equals document order"*, and described `--occurrence N` as an index reusable across calls. An agent that confirms row 2, then unconfirms row 2 off the same read, hits a different row.

Changes (docs only, `skills/uipath-ixp/`):

- **SKILL.md** — new Critical Rule 18 (`Occurrence` is read-scoped); Rule 13 no longer calls it document order; confirm/unconfirm occurrence rows in Task Navigation; two Common Pitfalls rows for "the confirmed row is now first" and "the second `--occurrence` call hit the wrong row".
- **label-documents-guide.md** — new *Occurrence numbers are read-scoped* section; per-occurrence confirm now steers to one batched `--updates` call; unconfirm takes its index from a fresh read.
- **cli-reference.md** — new *Occurrence numbering and read order* section; `get-predictions`, `confirm --occurrence`, `unconfirm --occurrence` rows updated.

Guidance in one line: batch all target rows into one `--updates` call, re-read `get-predictions` between sequential per-occurrence writes, and identify rows by value rather than index.

Server-side behaviour verified live (incl. UI) on alpha; tracked as ACTV-89753, where the API fix will make occurrence numbering stable in document order. This PR aligns the docs with what the server does today and stays correct after that lands.

https://claude.ai/code/session_01UUHFJKqqCSFbpigcZjKmVH


skills/uipath-ixp/references/cli-reference.md
```diff
@@ -155,15 +155,27 @@ Add before deleting: if the add fails, the field is still in its original group.
...
+| `uip ixp labellings get-predictions <project-name> [document-id] --output json` | Get IXP model predictions for all documents (or a single document). Returns `Data: { ProjectName, TotalDocuments, DocumentsWithPredictions, Predictions[] }`. Each `Predictions[]` entry is one document `{ DocumentId, Labels[] }`; each label is `{ Name, Occurrence, Fields[] }`; each field is `{ FieldId, FieldName, FormattedValue }`. This is the model's **prediction** layer, not the confirmed/annotation layer. Each label carries an explicit `Occurrence` (the value for `--occurrence`/`--updates`); it is 0-based and usually runs 0..N-1 in document order, but do NOT assume it is contiguous or starts at 0 — a single-occurrence group can come back as `Occurrence` 1 with no 0. Always target the actual `Occurrence` value reported here, never a positional guess. **The order is not stable across writes**: the server lists annotation↔prediction matched pairs first, so confirmed rows of a repeatable group sort to the front and the rest renumber — see [Occurrence numbering and read order](#occurrence-numbering-and-read-order). |
```
alexandrujircan, line 158: where is #occurrence-numbering-and-read-order reference?

andrei-uipath:
Added the smoke task: `tests/tasks/uipath-ixp/smoke/batch_occurrence_confirm.yaml` (`skill-ixp-batch-occurrence-confirm-smoke`).

Five `Line Items` extractions, four correct and one invented. Graded shapes:

| Weight | Criterion |
|---|---|
| 3.0 | occurrences 0, 1, 3, 4 confirmed in **one** `--group` + `--updates` call |
| 2.0 | did NOT chain single-occurrence `--occurrence` confirms |
| 2.0 | did NOT confirm occurrence 2 (the invented row) |
| 1.5 | did NOT use plain `--fields` without `--group` |
| 1.5 | did NOT confirm the whole document blind |

The chaining check is the point of the task: the first confirm renumbers the group, so every later index in a chain addresses a row the agent never validated. Indices are not stated in the prompt — the agent derives them.

**Local run is blocked**: the workspace API key is over its usage limit until 2026-08-01, so the task errored with `agent_api_error / 400 usage limits` before the agent ran. Not claiming a passing local run. What I did verify is the grading itself — I replayed six synthetic `calls.log` shapes through the five patterns:

| Shape | Score |
|---|---|
| ideal, one batched call (quoted + unquoted, as the mock logs it) | 10.0/10.0 |
| batched but including the invented row | 8.0/10.0 |
| split into two `--updates` calls | 7.0/10.0 |
| chained `--occurrence` calls | 5.0/10.0 |
| plain `--fields`, no `--group` | 5.5/10.0 |
| blind whole-document confirm | 5.5/10.0 |

Happy to re-run for real once the key resets, or if someone with quota wants to kick it off sooner.

No integration/e2e counterpart on purpose: the only thing a live test could pin is today's reordering, which ACTV-89753 deliberately changes. Order guarantees belong in the designtime-api service tests.

misupantea-uipath (approved):

andrei-uipath:
The batching task above grades the *mitigation*, not the ordering fact — an agent that batches for unrelated reasons scores full marks. Added the task that grades the fact: `tests/tasks/uipath-ixp/smoke/reordered_occurrence_read.yaml`.

Fixture is a partly-confirmed `Line Items` group, so the read comes back with matched pairs first and `Occurrence` no longer tracks the page:

```
invoice, top to bottom          get-predictions
1. Steel bolts   $400           Occurrence 0 — Copper wire  (confirmed earlier)
2. Copper wire   $150           Occurrence 1 — Pallet wrap   (confirmed earlier)
3. Shrink film   $24            Occurrence 2 — Steel bolts   (not confirmed)  <- confirm this
4. Pallet wrap   $36            Occurrence 3 — Shrink film $2,400 (wrong total)
```

The correct row is the **first line on the page** but **`Occurrence` 2**. Numbering the invoice top-to-bottom — what the old docs invited — yields `--occurrence 0` and re-confirms Copper wire. So the two answers are distinct and the check has teeth.

| Weight | Criterion |
|---|---|
| 3.0 | confirms occurrence 2 (index from the read) |
| 2.5 | did NOT confirm occurrence 0 (page-order guess) or 1 (already confirmed) |
| 2.0 | did NOT confirm or `--corrections` occurrence 3 (wrong total, not OCR garble) |
| 1.5 | did NOT confirm without scoping to `--group` |

Same deterministic replay (local agent run still blocked by the API quota): correct shapes — `--occurrence 2`, an `--updates` entry, quoted/spaced variants — score 9.0/9.0; the page-order trap scores 3.5/9.0; re-confirming 0/1 alongside 2 scores 6.5; sweeping in the wrong row 7.0; `--corrections` on it 4.0; unscoped `--fields` or a blind whole-doc confirm 4.5.

What no offline task can check is that the *server* reorders — the mock returns nothing. These two pin agent behavior against a reordered read; the server-side order guarantee belongs in the designtime-api service tests (ACTV-89753 / #35458).

andrei-uipath:
Dropped `batch_occurrence_confirm.yaml` (849e2a6b3). It near-duplicated `per_occurrence_confirm.yaml`, and the only thing making the batched shape mandatory was a prompt clause banning re-reads — chain-plus-re-read is a shape the skill permits, so the task was gating on an artificial constraint. `reordered_occurrence_read.yaml` carries the behavior worth pinning.

Net diff on this PR: the three doc files plus one smoke task.

skills/uipath-ixp/references/cli-reference.md
andrei-uipath: It's in this same file, nine lines below the table — `cli-reference.md:167`, added in this PR (just outside the diff hunk shown here): https://github.com/UiPath/skills/blob/fix/ixp-occurrence-read-order/skills/uipath-ixp/references/cli-reference.md#L167

`### Occurrence numbering and read order` → anchor `#occurrence-numbering-and-read-order`. It covers why there is no stable row key (`field_group.id` is the taxonomy group id, identical for every row), the matched-pairs-first order, and the three call-shape rules. The `confirm --occurrence` and `unconfirm --occurrence` rows link to it too.

The sibling section in `label-documents-guide.md` is `#occurrence-numbers-are-read-scoped`, linked from step 2a and the per-occurrence blocks.

tests/tasks/uipath-ixp/smoke/reordered_occurrence_read.yaml
```diff
@@ -0,0 +1,98 @@
...
+  mock_path_dirs: [mocks]
+  template_sources:
+    - {type: template_dir, path: ../_shared/mock_template}
+
+initial_prompt: |
+  On project `freight_docs-df33115f-ixp`, document `doc-freight-13`, the
+  `Line Items` field group has four extractions. Two of them were confirmed
+  in an earlier session; the other two are still unconfirmed.
+
+  The invoice lists these four line items, top to bottom:
```
alexandrujircan: Why present this info? A coding agent wouldn't have access to the real order from the invoice, right?

tests/tasks/uipath-ixp/smoke/reordered_occurrence_read.yaml
```diff
@@ -0,0 +1,98 @@
+task_id: skill-ixp-reordered-occurrence-read-smoke
+description: >
+  Smoke: on a partly-confirmed repeatable group the read returns matched
+  annotation/prediction pairs FIRST, so `Occurrence` no longer follows the
+  order the rows appear on the page. The agent must take the index from the
+  `get-predictions` output, not from the row's position on the invoice —
+  here the target line is the first line on the page but `Occurrence` 2.
```
alexandrujircan: So the agent was looking at the row's position on the invoice before? And this test doesn't pass without the skill change?

tests/tasks/uipath-ixp/smoke/reordered_occurrence_read.yaml
andrei-uipath: Because the agent doesn't really call the IXP service, so it has to know somehow this info. It could get it from the first get-predctions response (before anything is annotated)

tests/tasks/uipath-ixp/smoke/reordered_occurrence_read.yaml
andrei-uipath: that is a good question

andrei-uipath:
## A test that fails without this change

Replaced the two shape-graded tasks with one that grades **wire effect**, and measured it against both doc versions before believing it.

`tests/tasks/uipath-ixp/_shared/reordering_mock/` is a stateful `uip ixp labellings` double. Fixtures come from a live `get-predictions` capture of a 16-row `Invoice > LineItems` document — captured envelope, label names, field ids (`dc2c87df35fe873c`, `f0e390b50f3a42f5`), row count and response shapes; only the line text is synthetic, since the source document belonged to a live tenant. It reproduces five verified behaviours: matched-pairs-first read order, `--occurrence N` addressing that order, one-invocation-one-snapshot batch semantics, the silent no-op unconfirm, and unscoped/group-wide spread.

Two rows (3 and 15) start annotated, so the very first read is already reordered and a stale index destroys real work. `occurrence_rollback_after_confirm` asks for one row to be confirmed and then rolled back, and grades which **row** the double actually mutated.

### Differential, 2 runs per arm, agent given only one doc version

| arm | mutations the double applied | end state | score |
|---|---|---|---|
| pre-change docs, run 1 | `confirm occ 2 → row 0`; `unconfirm occ 2 → row 15` | `0 3` | 6.0/14.0 |
| pre-change docs, run 2 | `confirm occ 2 → row 0`; `unconfirm occ 2 → row 15` | `0 3` | 6.0/14.0 |
| this PR's docs, run 1 | `confirm occ 2 → row 0`; `unconfirm occ 0 → row 0` | `3 15` | **14.0/14.0** |
| this PR's docs, run 2 | `confirm occ 2 → row 0`; `unconfirm occ 0 → row 0` | `3 15` | **14.0/14.0** |

Both pre-change runs reused the index they had just written with, wiped a row reviewed in an earlier session, and left the target row still confirmed. Neither re-read. One of them reported: *"the confirm response (PredictionsConfirmed: 2) and the later unconfirm response (FieldsUnconfirmed: 2, Unmatched: []) both matched expectations, so no re-check was needed"* — the failure is invisible from the responses, which is the whole point. Both post-change runs cited Rule 18, re-read, and found the row at occurrence 0.

Scenario matrix through the same double (deterministic, no agent): correct 14.0 · `mark-missing` instead of unconfirm 11.0 · stale-then-self-corrects 9.0 · stale 6.0 · blanket document rollback 4.5 · one throwaway call 1.0 · nothing 0. Only the correct path scores full marks, and the destructive answers rank below inaction.

### Dropped

- `occurrence_second_write_after_confirm` — cannot discriminate. With annotated rows sorting first, a stale index still resolves correctly whenever every newly-annotated row has a lower document index, so with two targets ascending write order is always safe; all four arms scored 11.0/11.0. No choice of target pair fixes it.
- `reordered_occurrence_read` — 9.0/9.0 on both doc versions. It graded guidance that already existed ("always target the actual `Occurrence` value reported here").

### Review findings folded in

Two adversarial reviewers went at the double and the grading. Fixed: `--occurrence=8` silently escalating to a group-wide write; an infinite loop when a flag's value was missing (bash `shift 2` with one arg left); `--occurrence abc` fabricating row 0 via arithmetic expansion; unvalidated `--group` accepting any string; constant `PredictionsConfirmed`/`FieldsUnconfirmed`; `unconfirm` and `mark-missing` returning the confirm envelope; blanket writes logging no `row=` (so the most destructive answer outscored the targeted-wrong one); a vacuous ~48% floor from an empty log file; and the answer key plus grading substrate sitting inside the PATH directory agents are told to leave alone — both now under `_svc/`.

Still not claiming a `coder-eval` run: the workspace API key is over quota until 2026-08-01, so the arms above ran through the same fixtures and the same criteria, driven by agents given only the skill docs, rather than through the coder_eval harness. CI smoke will be the first harness run.

andrei-uipath:
Trimmed the fixture from sixteen rows to three (`de43c7c01`) — the extra rows showed nothing the three do, and the whole ordering now fits in one glance:

```
rows            0 Service call S50FT unit #2340   <- target
                1 Recovery/Eco fee                <- control, never touched
                2 PO 40021 shoe kit, left         <- confirmed in an earlier session

read            [2, 0, 1]   target reads at occurrence 1
confirm occ 1   [0, 2, 1]   occurrence 1 is now row 2
unconfirm occ 1             destroys row 2, answers FieldsUnconfirmed: 2, Unmatched: []
```

Re-measured rather than assumed, since the prompt changed. 2 runs per arm, each agent given one doc version only:

| arm | mutations the double applied | end state | score |
|---|---|---|---|
| pre-change docs, run 1 | `confirm occ 1 → row 0`; `unconfirm occ 1 → row 2` | `0` | 6.0/14.0 |
| pre-change docs, run 2 | `confirm occ 1 → row 0`; `unconfirm occ 1 → row 2` | `0` | 6.0/14.0 |
| this PR's docs, run 1 | `confirm occ 1 → row 0`; `unconfirm occ 0 → row 0` | `2` | **14.0/14.0** |
| this PR's docs, run 2 | `confirm occ 1 → row 0`; `unconfirm occ 0 → row 0` | `2` | **14.0/14.0** |

Same 6.0-vs-14.0 split as the sixteen-row version. Both pre-change runs reused the index and wiped the earlier session's annotation while leaving the target confirmed — the inverse of what was asked. Both reasoned that they had followed the rule against positional guessing, which they had: the pre-change docs say to read the reported `Occurrence`, and say nothing about that value expiring.

Deterministic scenario matrix unchanged: correct 14.0 · `mark-missing` substitute 11.0 · stale-then-self-corrects 9.0 · stale 6.0 · blanket document rollback 4.5 · one throwaway call 1.0 · nothing 0.

Fixture is 21 files instead of 34, and `_svc/fixtures/` now holds three row blocks.

alexandrujircan (approved):

PR #2315, opened 2026-07-29
fix(tests): unblock two uipath-ixp smoke tasks failing on the grader, not the skill

andrei-uipath:
Both [2026-07-29 nightly](https://coder-evalboard.uipath-dev.com/runs/2026-07-29_04-37-44) uipath-ixp failures are grading bugs. Neither agent did anything wrong; the `uipath-ixp` skill needs no change.

## `skill-ixp-move-field-between-groups-smoke` — unwinnable by any agent (scored 0.54)

The base mock exits 1 on every invocation, which breaks this task two ways:

1. **The carried-over values are unobtainable.** `projects get-taxonomy` returns nothing, yet the criterion requires `fields add … --type … --instructions …` with the field's existing type and instructions.
2. **The graded `fields delete` is wrong to issue.** The skill teaches add-before-delete so a failed add leaves the field in place (`cli-reference.md`: "if the add fails, the field is still in its original group"). With `fields add` guaranteed to fail, withholding the delete is the documented correct behavior.

The agent said as much:

> Couldn't complete the move: the required taxonomy lookup failed because this sandbox is not connected to an IXP tenant. I did not guess the field's existing type or extraction instructions, so no add/delete operation was issued.

Added 2026-07-28 in #2288, so this was its first nightly — it has never passed.

**Fix.** New `_shared/mock_template_taxonomy` overlay (same layering pattern as the existing `mock_template_ambiguous`, listed second) serving canned JSON for the reads and success for the two `fields` mutations. `get-taxonomy` is stateful, so an agent that re-reads to verify sees its own writes instead of stale state it would try to re-apply:

| After | `Invoice Header` | `Line Items` |
|-------|------------------|--------------|
| — | Total Amount | — |
| `fields add` | Total Amount | Total Amount (**new** `field_id`) |
| `fields delete` | — | Total Amount |

The intermediate both-groups state is what the real backend does between add and delete, and the new `field_id` is why a move drops the field's confirmed labels. `--type` now asserts the value `Monetary Quantity` rather than flag presence — only knowable from the read, via the `moon_form` entry's `field_type_id`.

## `skill-ixp-update-prompts-heredoc-smoke` — quoting, not behavior

The `calls.log` criterion proving the fully expanded, correct payload reached the CLI **passed**. Only the criterion regexing raw Bash text failed:

- pattern wanted: `--updates ["']?$(cat …json)`
- codex emitted: `--updates \""'$(cat …json)"`

Codex wraps commands in `/bin/bash -lc "…"`, backslash-escaping the inner quotes. `["']?` allows one bare quote and no backslash, so the **same correct command** passed at score 1.0 under `claude-code` on [2026-07-20](https://coder-evalboard.uipath-dev.com/runs/2026-07-20_04-26-57) and failed under codex on 2026-07-29. The trigger was the agent switch, not a task edit — the pattern was always this brittle, just never exercised by codex.

**Fix.** Quote runs are now `(?:\?["'])*`.

This is the anti-pattern `.claude/rules/test-writing.md` ("grade the outcome, not the literal flag") and `mock_template/README.md` ("not regexes over agent-authored Bash text") already warn about. The stricter alternative is to **delete** the criterion — criteria 1 and 3 already prove the file was written and the right payload arrived; only the file→CLI linkage is lost. Happy to drop it instead if reviewers prefer. `grep` confirms this was the only `$(cat` pattern of its kind in `tests/`.

## Verification

⚠️ **`coder-eval` passing-run claim is pending** — not yet attached. What was verified instead:

- Executed the new mock under `sh` and graded the real `calls.log` it produces against all 6 `move_field_between_groups` criteria — all pass, including the 4 negative guards (confirmed they still bite by probing `groups add`).
- `get-taxonomy` emits valid JSON at all three states; type resolves through `field_type_id`; `field_id` matches no `entity_def` at every state.
- Regex-tested the new `command_pattern` against the **actual recorded commands** from both nightly runs: matches codex 2026-07-29 and claude 2026-07-20, still rejects an inline payload, a different project, and `groups` instead of `fields`.
- `mocks/uip` committed mode `755`, LF-clean, `sh -n` clean.
- `scripts/check-task-driver.py` and `scripts/check-cli-verbs.py` clean — 0 High, 0 Medium (the 1 Info is pre-existing on criterion 1).
- Confirmed the two-`template_dir` layering mechanism works under codex in this same nightly: `skill-ixp-rename-ambiguous-entity-smoke` uses it and scored 1.0.

https://claude.ai/code/session_01HfNsX5UrVVqZfHUFhsCn6E

### DIFF
--- a/tests/tasks/uipath-ixp/_shared/mock_template/README.md
+++ b/tests/tasks/uipath-ixp/_shared/mock_template/README.md
@@ -34,3 +34,8 @@ disobedient agent cannot reach the cloud with the harness-injected token.
 
 Integration/e2e tasks use `live_calls_template`; its wrapper records the same log
 format and delegates unchanged to the real CLI.
+
+A task whose correct path reads before it writes cannot be graded on this mock
+alone — the read fails, so there is nothing to carry into the graded write, and
+an agent that declines to invent the values correctly stops. Overlay
+`mock_template_taxonomy` (listed second) to serve that read.
--- a/tests/tasks/uipath-ixp/_shared/mock_template_taxonomy/README.md
+++ b/tests/tasks/uipath-ixp/_shared/mock_template_taxonomy/README.md
@@ -1,0 +1,68 @@
+# uipath-ixp taxonomy-serving mock
+
+Overlay for smoke tasks whose correct path **starts with a read**. Overlays the
+base [`mock_template`](../mock_template/README.md) — list it SECOND so its
+`mocks/uip` wins:
+
+```yaml
+sandbox:
+  mock_path_dirs: [mocks]
+  template_sources:
+    - {type: template_dir, path: ../_shared/mock_template}
+    - {type: template_dir, path: ../_shared/mock_template_taxonomy}
+```
+
+## Why it exists
+
+The base mock fails every invocation. That is right for a task whose graded
+commands are independent, and wrong for one that reads before it writes:
+
+- **A value carried from the taxonomy is unobtainable.** If grading requires
+  `fields add --type … --instructions …` with the field's existing type and
+  instructions, a failing `get-taxonomy` leaves the agent nothing to carry. An
+  agent that declines to invent them stops — the correct call, graded as a fail.
+- **A mutation gated on an earlier mutation is wrong to issue.** The skill
+  teaches add-before-delete so a failed `fields add` leaves the field where it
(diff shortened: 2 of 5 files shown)


andrei-uipath:
Superseded by #2316 (move-field mock) and #2317 (heredoc criterion) — split per the one-logical-change-per-PR rule. Same content, no changes.

PR #2322, opened 2026-07-29
Feat/tm playwright first mile

addyGarg:
## What

Teaches the `uipath-test` skill the **Playwright first mile**: take a Playwright suite from a repo to executed results in Test Manager, using only `uip` commands.

- **New** `references/playwright-first-mile-guide.md` — the seven-step pipeline (pack → upload → automatic ingestion → label-filled test set → context probe → project-scoped run → results), plus how to get a *fix* back in and what to do when a run produces no results.
- **`SKILL.md`** — the external-package concept, a Pack section, the `playwright-context` row, the `--playwright-projects` note on the run row, the `--labels` selector on `testcases add`, a navigation entry, and one anti-pattern (never `link-automation` a Playwright test case — ingestion links them).
- **One smoke task** (`playwright_package_prep_smoke.yaml`) that passes on the currently published CLI.

The pipeline differs from the Studio/RPA one in a way worth calling out: there is **no link step**. Uploading the package is enough — ingestion creates one test case per Playwright test, already bound, and labels each with `PW_Tag_*`, `PW_Project_*`, `PW_Suite_*`, `PW_File_*`.

## Why

The CLI commands landed in UiPath/cli#3226 and the Test Manager side is deployed, but nothing taught an agent the order, the traps, or how to tell a Playwright test set from an RPA one. Without this, an agent reaches for `link-automation` (wrong pipeline), polls ingestion with a filter that never matches, or applies project scoping to a test set that cannot take it.

## How it was validated

Not by reading — by running. **Six cold agents**, each given only this skill and a logged-in `uip`, on a local serverless stack and on the alpha `testmanagerdev/DevTest` tenant. Every gap they hit was folded back in:

- ingestion polling with `--filter <PackageName>` never matches (names are `"<suite> > <title>"`, package only in the description) → a false "ingestion never happened"
- `IsAutomated: false` is normal on ingested test cases
- folder choice must come from the unflagged `folders list`; the default folder needs a Cloud Robots–Serverless machine
- `report get` requires `--project-key`/`--test-set-key` (verified against the CLI)
- no-results triage: `Cancelled`, `None`-results and stuck-`Pending` shapes

Latest full run on alpha: pack → upload → ingestion (~10 s) → label-filled test set → chromium-only run → **1 passed / 1 failed with the real Playwright error surfaced**, then a fix-and-republish loop (bump version, re-upload, ingestion updates in place, re-run the same test set) → **2/2 passed**, confirmed by `report get`, `testcaselogs list` and the JUnit export.

## Related

- **UiPath/skills#2401** — the integration and e2e eval tasks, split out: they drive the full pipeline and cannot pass until the commands reach npm `latest` (the eval runner installs `@uipath/cli@latest` and tasks may not pin a version). Keeping them here would mean a scheduled task that fails every run.
- **UiPath/cli#3336** — registers `playwright-context` on stable builds (it was `previewCommand`, so absent from every stable release while `--playwright-projects` worked). Until that lands, a stable CLI answers `unknown command` for the probe; the guide's availability check handles it — skip the probe, and only drop project scoping if the run flag itself is rejected.
- **UiPath/cli#3226** — the commands this skill documents, merged into the 1.200 line.

This PR can merge independently: nothing in it needs an unreleased command to *function*, because the guide gates on availability.


addyGarg:
## Test coverage + passing-run claims

Two coder_eval tasks added per `.claude/rules/test-writing.md`:

**`playwright_pack_generate`** (integration · mode:build · lifecycle:generate) — offline: scaffold a minimal Playwright suite → `tm pack --type playwright --no-create-test-cases` → assert the `.nupkg` exists and embeds `testCases.json`, plus the never-link-automation anti-pattern guard. *Passing-run claim:* all criteria simulated green against a CLI built from UiPath/cli#3226 (pack Success, TestCount 2, nupkg + embedded metadata verified via `unzip`). Deliberately tiered **integration**, not smoke: the PR-gate smoke runner installs `@uipath/cli@latest`, which lacks these commands until cli#3226 publishes — retag to smoke then.

**`playwright_first_mile_e2e`** (e2e · mode:operate · lifecycle:setup · **path-to-ga**) — the full pipeline to executed results on a capable tenant (pack → upload → auto-ingestion → label fill → `playwright-context` probe → chromium-only run → results). *Passing-run claim:* the equivalent flow was executed live by five context-free agent runs during skill authoring — on a full local serverless stack it reached **2/2 passed, chromium-only, selection persisted**; on alpha DevTest the first mile is green end-to-end with execution currently blocked by a tenant-side serverless identity fault (`Serverless.Runtime.CannotIssueUserTokenDueToUserNotPartOfOrg`), hence the path-to-ga tag.

Tenant prerequisites for the e2e task are documented in the task header and in `references/playwright-first-mile-guide.md`.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,148 @@
...
+## Step 3 — Wait for ingestion
+
+Ingestion is asynchronous and automatic. Poll until the auto-created test cases appear:
+
+```bash
+uip tm testcases list --project-key <PROJECT_KEY> --output json
+```
+
+- Poll **unfiltered** and count. Do NOT pass `--filter <PackageName>` — the auto-created test case *names* are `"<suite> > <test title>"`; the package name appears only in the description, which `--filter` does not search, so a package-name filter stays empty forever and reads as a false "ingestion never happened".
+- Expect exactly `TestCount` new test cases (from Step 1), typically within 1–2 minutes. `TestCount` is one per Playwright **test**, NOT multiplied by the number of Playwright projects (2 tests × 2 projects → 2 test cases).
```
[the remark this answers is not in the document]
addyGarg: Good catch — the poll now takes a pre-upload baseline and waits for the count to grow by TestCount, so pre-existing test cases can't fake completion.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,148 @@
...
+Set the project's default Orchestrator folder FIRST — both the Step 5 probe and the Step 6 run resolve packages through it (Critical Rule #10):
+
+```bash
+uip or folders list -n <folder-name> --all --output json   # Key field = folder key
+uip tm project set-default-folder --project-key <PROJECT_KEY> --folder-key <FOLDER_KEY> --output json
+```
+
+```bash
+uip tm testsets create --project-key <PROJECT_KEY> --name "PW Smoke" --output json
+uip tm testcases add --test-set-key <TEST_SET_KEY> --labels "PW_Suite_<name>" --output json
```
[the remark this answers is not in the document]
addyGarg: Agreed. Added: labels aren't package-qualified — in multi-package projects fill by a package-unique label (PW_File_*, unique suite name) or explicit keys.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,168 @@
...
+uip or folders list --output json    # WITHOUT --all: only folders you are a member of
+uip tm project set-default-folder --project-key <PROJECT_KEY> --folder-key <FOLDER_KEY> --output json
+```
+
+- Pick the folder from the **unflagged** `folders list` — `--all` returns every folder *visible* to you, including ones where you have no rights; choosing one of those fails later with `folderNotFoundOrNoAccess`.
+- **The folder must have a Cloud Robots – Serverless machine assigned** or the run's job creation 500s and the execution is instantly `Cancelled`. Check with `uip or machines list --folder-key <FOLDER_KEY> --output json`; if none, create and assign one (one serverless machine per folder):
+
+```bash
+uip or machines create -n <name> --serverless --testing-slots 2 --output json
+uip or machines assign <name> --folder-key <FOLDER_KEY> --output json
```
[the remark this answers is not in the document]
addyGarg: You're right — verified against the CLI (`machines assign <machine-keys...>`, GUIDs). The snippet now captures Data.Key from create and assigns by key.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
```diff
@@ -0,0 +1,109 @@
...
+    tool_name: "Bash"
+    command_pattern: 'uip\s+tm\s+testsets?\s+playwright-context'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent ran the test set scoped to the chromium project (hidden flag)"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+tm\s+testsets?\s+run\s+.*--playwright-projects\s+chromium'
```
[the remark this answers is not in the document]
addyGarg: Tightened — the criterion now rejects any extra project after chromium.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
```diff
@@ -0,0 +1,109 @@
...
+    tool_name: "Bash"
+    command_pattern: 'uip\s+(or|orchestrator)\s+packages\s+upload'
+    min_count: 1
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent polled for automatic ingestion with an unfiltered test case list (guide Step 3)"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+tm\s+testcases?\s+list\s+--project-key\s+PWE2E(?!.*--filter)'
```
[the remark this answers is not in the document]
addyGarg: Fixed — the lookahead now covers the whole command, so a --filter in any position disqualifies the poll.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
```diff
@@ -0,0 +1,109 @@
...
+    description: "Agent fetched results (report/testcaselogs/stats) rather than stopping at run"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+tm\s+(report\s+get|executions\s+(testcaselogs\s+list|get-stats))'
+    min_count: 1
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: run_command
+    description: "E2E outcome: latest PWE2E execution reached a terminal state with results (Passed on a capable tenant)"
+    command: "uip tm executions list --project-key PWE2E --output json 2>/dev/null | grep -qE '\"Status\": \"(Finished|Passed)\"'"
```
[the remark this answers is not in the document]
addyGarg: Fair point on the durable project — the check now inspects only the newest execution (--limit 1).

skills/uipath-test/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: uipath-test
-description: "UiPath Test Manager — manage test projects, cases, sets, executions; generate reports. For Orchestrator→uipath-platform. For test automation→uipath-rpa."
+description: "UiPath Test Manager — manage test projects, cases, sets, executions; generate reports; pack and run Playwright suites on serverless (tm pack, PW_* labels, --playwright-projects). For Orchestrator→uipath-platform. For Studio/RPA test automation authoring→uipath-rpa."
```
ganeshborle: do not include commands here

skills/uipath-test/SKILL.md
```diff
@@ -29,6 +29,7 @@ UiPath Test Manager is a web application that manages the testing lifecycle of p
 - **Test case logs** - Logs of a **test case** in an execution. A **testcase** can be navigated from **testcaselogs**.
 - **Test step logs** — Step-level logs within a **test case log**.
 - **Test case log assertions** - Assertion steps of a test case log in an execution.
+- **External (Playwright) test packages** - A Playwright suite packed with `uip tm pack --type playwright` and uploaded to Orchestrator. Ingestion then auto-creates one test case per Playwright test (no link step) and labels each with `PW_Tag_*`, `PW_Project_*`, `PW_Suite_*`, `PW_File_*`. These run on serverless cloud runtimes — see [references/playwright-first-mile-guide.md](references/playwright-first-mile-guide.md).
```
ganeshborle: This section just describes what every term in the test manager means, so avoid adding commands here.

skills/uipath-test/SKILL.md
```diff
@@ -85,7 +86,7 @@ Common `uip tm` commands organized by resource type.
 | `uip tm testcases steps delete --project-key <PROJECT_KEY> --step-id <UUID> --yes` | Delete a step. |
 | `uip tm testcases list-result-history --project-key <PROJECT_KEY> --test-case-id <TEST_CASE_ID>` | List test case log result history for a specific test case. Optional `--only-failed`, `--filter`, `--limit`, `--offset`. |
 | `uip tm testcases run --project-key <PROJECT_KEY> --test-case-id <TEST_CASE_ID>` | Start a new execution for one or more test cases. **Uses `--test-case-id <UUID>` (space-separated for multiple).** Optional `--async`, `--name`, `--folder-key`, `--robot-user-key`, `--machine-key`. |
-| `uip tm testcases add --test-set-key <TEST_SET_KEY> --test-case-keys <KEY1,KEY2,...>` | Add test cases to a test set (comma-separated keys). |
+| `uip tm testcases add --test-set-key <TEST_SET_KEY> (--test-case-keys <KEY1,KEY2,...> \| --labels <name...>)` | Add test cases to a test set — by explicit keys, OR every test case carrying at least one of the given labels (variadic, space-separated; quote names containing spaces; OR-match, exact, case-sensitive). Pass exactly one of the two selectors. |
```
ganeshborle: Could say; both options are mutually execlusive.

skills/uipath-test/SKILL.md
```diff
@@ -103,10 +104,13 @@ Common `uip tm` commands organized by resource type.
 | `uip tm testsets update --test-set-key <TEST_SET_KEY> --name <TEST_SET_NAME>` | Update a test set name or description. |
 | `uip tm testsets delete --test-set-key <TEST_SET_KEY>` | Delete a test set by its key. |
 | `uip tm testsets list-testcases --project-key <PROJECT_KEY> --test-set-key <TEST_SET_KEY>` | List test cases assigned to a test set. |
-| `uip tm testsets run --test-set-key <TEST_SET_KEY>` | Run a test set and return the execution ID. Optional `--execution-type <automated\|manual\|mixed\|none>` (default `automated`), `--input-path <FILE>` for parameter overrides. |
+| `uip tm testsets run --test-set-key <TEST_SET_KEY>` | Run a test set and return the execution ID. Optional `--execution-type <automated\|manual\|mixed\|none>` (default `automated`), `--input-path <FILE>` for parameter overrides, `--wait` to block until terminal. For Playwright test sets, optional `--playwright-projects <names...>` (hidden from `--help`; see below). |
```
ganeshborle: `--wait` to block until terminal.  -> this is generic option; should not go here!

ganeshborle: (hidden from `--help`; see below). -> this is questionable. We are telling coding agent that it is there and use it but hiding it from the cli. This would again bring the scenario that if it is not available in the prod; coding agent would still try it because of this skill.

skills/uipath-test/SKILL.md
```diff
@@ -103,10 +104,13 @@ Common `uip tm` commands organized by resource type.
...
+| `uip tm testsets playwright-context --test-set-key <TEST_SET_KEY>` | Probe whether a test set is a Playwright test set: returns `IsPlaywright` plus the available/selected Playwright project names. Hidden from `--help` but functional. See [references/playwright-first-mile-guide.md](references/playwright-first-mile-guide.md). |
```
ganeshborle: same comment for the --help!

skills/uipath-test/SKILL.md
```diff
@@ -103,10 +104,13 @@ Common `uip tm` commands organized by resource type.
...
 
 > Keys use the format `PROJECT_KEY:NUMBER` (e.g., `INV:42`). To add or remove test cases in a test set, use `uip tm testcases add` / `uip tm testcases remove` — those verbs live under the `testcases` group, not under `testsets`.
 
+> **Playwright test sets:** `--playwright-projects <names...>` (space-separated, case-sensitive `playwright.config` project names) runs only the selected projects and persists the selection on the test set. It requires every test case in the set to come from one single Playwright package; unknown names fail fast listing the valid ones. Probe first with `uip tm testsets playwright-context` and branch on `IsPlaywright`. Full pipeline: [references/playwright-first-mile-guide.md](references/playwright-first-mile-guide.md).
```
ganeshborle: Should it go here?

skills/uipath-test/SKILL.md
```diff
@@ -141,7 +145,7 @@ Common `uip tm` commands organized by resource type.
 
 | Command | Purpose |
 |---|---|
-| `uip tm report get --execution-id <EXECUTION_ID>` | Get a summary report for a completed test execution. Optional `--project-key`, `--test-set-key`, `--query`. |
+| `uip tm report get --execution-id <EXECUTION_ID> (--project-key <KEY> \| --test-set-key <KEY>)` | Get a summary report for a completed test execution. One of `--project-key`/`--test-set-key` is required to identify the project. Optional `--query`. |
```
ganeshborle: Can you validate this locally?
Optional `--query` should also get removed (should have been removed from the parent as well :) as it is common option across commands )

skills/uipath-test/SKILL.md
```diff
@@ -156,6 +160,14 @@ Common `uip tm` commands organized by resource type.
 |---|---|
 | `uip tm result download --execution-id <EXECUTION_ID>` | Download test execution results as JUnit XML. Optional `--project-key`, `--test-set-key`, `--result-path <DIR>`. |
 
+### Pack Commands (Playwright)
```
ganeshborle, line 163: I think you also need to update readme.md of cli with these new commands. But make sure you update that only when the code is available in the prod.

skills/uipath-test/SKILL.md
```diff
@@ -305,11 +317,16 @@ If the probe in Rule #2 shows singular subjects, the CLI predates the closed-ver
 | I want to... | Start here |
 |---|---|
 | **Generate a shareable test report** (tester or release manager view) | [references/test-result-report-guide.md](references/test-result-report-guide.md) |
-| **Publish a project and link it to a Test Manager test case** | [references/publish-and-link-guide.md](references/publish-and-link-guide.md) |
+| **Publish a project and link it to a Test Manager test case** (Studio/RPA) | [references/publish-and-link-guide.md](references/publish-and-link-guide.md) |
+| **Pack, ingest, and run a Playwright suite on serverless** (pack → upload → labels → run) | [references/playwright-first-mile-guide.md](references/playwright-first-mile-guide.md) |
+
 
+> **Known CLI quirks:** a command can emit its JSON envelope and then sit before exiting (`[Telemetry] flush timed out`) — if you have the JSON, it succeeded. Auto-updater lines ("Checking for updates…") can precede the envelope, same hazard as progress lines (Rule #3). And `login status` can report "Logged in" (even with a stale `ExpirationDate`) while calls 401 — the non-interactive fix is `uip login refresh`; only fall back to interactive `uip login` if refresh fails.
```
ganeshborle: This is not required!

skills/uipath-test/SKILL.md
```diff
@@ -219,11 +231,11 @@ Object labels are tag-style metadata applied to Requirement, TestCase, TestSet,
 
 1. **Always check login first** — run `uip login status --output json` before any Test Manager operation. If not authenticated, run `uip login` to sign in.
 2. **Probe the CLI surface once per session, before the first `uip tm` command.** Run `uip tm testcases --help --output json` (any flags accepted). Result `Success` → post-rename CLI; use the command tables above as-is. `unknown command` / non-zero exit → pre-rename CLI; translate via the [Pre-rename fallbacks](#pre-rename-fallbacks) table before each call. Re-probe on any later `unknown command` error.
-3. **Always pass `--output json`** to every `uip` command — no exceptions. Structured JSON output is what you need to reason about results reliably, even when you only plan to summarize them back to the user.
+3. **Always pass `--output json`** to every `uip` command — no exceptions. Structured JSON output is what you need to reason about results reliably, even when you only plan to summarize them back to the user. Note: stdout may carry human progress lines *before* the JSON envelope (e.g. `Resolved project …`, run/wait progress) — when parsing, extract the JSON object rather than parsing stdout wholesale.
```
ganeshborle: I think this change is not very necessary!

skills/uipath-test/SKILL.md
```diff
@@ -219,11 +231,11 @@ Object labels are tag-style metadata applied to Requirement, TestCase, TestSet,
...
 4. **Cap retries at 3** for any failing `uip` CLI command. After 3 failures, stop and report the error to the user (see Rule — never fall back to direct REST APIs).
 5. **Handle empty results** — if a list command returns an empty array, stop and inform the user rather than proceeding with a null key.
 6. **Confirm before delete** — always confirm the target resource key with the user before running any `delete` command. All delete commands require `--yes` (or `-y`) to proceed; omitting it exits without deleting.
-7. **For operations requiring folder key** — use `uip or folders list -n <folder-name> --all --output json` (run `/uipath-platform` for folder management details).
+7. **For operations requiring folder key** — when the user named a folder, look it up with `uip or folders list -n <folder-name> --all --output json`; when *choosing* a folder yourself, use the unflagged `uip or folders list --output json` (memberships only — `--all` includes folders you cannot actually use). Run `/uipath-platform` for folder management details.
```
ganeshborle: This can rephrased I think!

skills/uipath-test/SKILL.md
```diff
@@ -261,7 +273,7 @@ If the probe in Rule #2 shows singular subjects, the CLI predates the closed-ver
   For more authentication details, run `/uipath-platform`.
 
 ### Confirm project scope
-  Ask the user for the project name or key before any Test Manager call. For multi-project scenarios, collect ALL names or keys in one prompt. Resolve each to a `PROJECT_KEY`:
+  Ask the user for the project name or key before any Test Manager call (unless the task already names one or authorizes creating a new project). For multi-project scenarios, collect ALL names or keys in one prompt. Resolve each to a `PROJECT_KEY`:
```
ganeshborle: This is not required I think. Coding agents are smart to deduce it by themselves.

skills/uipath-test/SKILL.md
```diff
@@ -305,11 +317,16 @@ If the probe in Rule #2 shows singular subjects, the CLI predates the closed-ver
...
 - **Do NOT skip the surface probe** (Critical Rule #2). On a pre-rename CLI, post-rename commands fail with `unknown command`; on a post-rename CLI, pre-rename commands fail the same way. The skill targets the post-rename surface and falls back per the [Pre-rename fallbacks](#pre-rename-fallbacks) table. Picking the wrong shape without probing burns a retry on every call.
 - **Do NOT guess command names — verb-noun composites are required.** The CLI uses explicit verb-noun forms; bare verbs do not exist. Confirm with `uip tm <resource> --help --output json`.
+- **Do NOT `link-automation` Playwright test cases.** Playwright ingestion links them to the package automatically; the manual link step belongs to the Studio/RPA pipeline only.
+- **Do NOT conclude the Playwright commands are missing because `--help` omits them.** `uip tm testsets playwright-context` and `run --playwright-projects` are intentionally hidden until the capability is broadly available — they are functional. Probe by running them; on a Test Manager or CLI without the capability they fail with explicit instructions.
```
ganeshborle: Same comment as before. It shouldn't be here. For the demo, you can load the local build of skills and showscase if needed. If we keep this in the skills, then by next Friday the next release branch will get cut and this will get released.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
```diff
@@ -0,0 +1,109 @@
+# Tenant prerequisites (documented in references/playwright-first-mile-guide.md):
```
ganeshborle: Did we run and tested it?
Adding this tests will run daily!

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
```diff
@@ -0,0 +1,109 @@
...
+#   - Test Manager with the Playwright feature flag enabled
+#   - a folder the runner user is a member of, with a Cloud Robots - Serverless
+#     machine assigned (the task lets the agent discover/provision this)
+#   - @uipath/cli new enough to carry tm pack --type playwright (UiPath/cli#3226)
+# Must-pass once the CLI is published and the e2e tenant has the serverless
+# Playwright runtime; until then this documents the target bar. (The GA-track
+# tag is deliberately omitted pending an allowlisted approver — see PR notes.)
+task_id: skill-test-playwright-first-mile-e2e
+description: >
+  Full Playwright first-mile e2e: agent uses the uipath-test skill to take a
```
ganeshborle: As it is e2e, do we need to have pre_data and post_data defined in this test ?

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
```diff
@@ -0,0 +1,109 @@
...
+tags: [uipath-test, e2e, mode:operate, lifecycle:setup, feature:test-case]
+
+max_iterations: 1
+run_limits:
+  expected_turns: 50
+  max_turns: 49
+  turn_timeout: 1200
+
+initial_prompt: |
+  Create a minimal Playwright test suite in ./pw-suite with two passing tests
```
ganeshborle: Can you rephrase this initial_prompt like a human. We shouldn't give much intstructions in the prompt and rather coding agent should deduce it by themselves with the help of skills and tool

tests/tasks/uipath-test/playwright_pack_generate.yaml
```diff
@@ -0,0 +1,62 @@
...
+  `uip tm pack --type playwright` — no tenant calls needed (pack is offline and
+  `--no-create-test-cases` removes the project-key requirement). Asserts the
+  agent (a) produces a project that satisfies pack's stated prerequisites
+  (@playwright/test dependency, a lockfile, a playwright.config), (b) packs
+  with --type playwright, (c) the resulting .nupkg embeds the testcases.json
+  test metadata, and (d) it never reaches for the Studio/RPA pipeline
+  (link-automation) — the skill, not the prompt, must teach all of this.
+tags: [uipath-test, integration, mode:build, lifecycle:generate, feature:test-case]
+
+initial_prompt: |
```
ganeshborle: Also, please add one smoke test.

tests/tasks/uipath-test/playwright_pack_generate.yaml
```diff
@@ -0,0 +1,62 @@
...
+    description: "Agent packed with tm pack --type playwright"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+tm\s+pack\s+.*--type\s+playwright'
+    min_count: 1
+    weight: 3.0
+    pass_threshold: 1.0
+
+  - type: file_exists
+    description: "External test package was produced"
+    path: "out/PwPackCheck.1.0.0.nupkg"
```
ganeshborle: Make sure to clean these files up once the run is finished.

ganeshborle (changes requested):

skills/uipath-test/SKILL.md
addyGarg: Done — the description no longer names commands; it just says the skill covers packaging and running external Playwright test suites (still enough for skill activation).

skills/uipath-test/SKILL.md
addyGarg: Agreed — the concept now describes the term only, no commands.

skills/uipath-test/SKILL.md
addyGarg: Reworded to 'The two selectors are mutually exclusive.'

skills/uipath-test/SKILL.md
addyGarg: Removed — the run row is back to its original text; --wait is generic and stays out of it.

addyGarg: You're right, and I checked: the published test-manager-tool (1.198.0) has no Playwright support at all, so anything documented here would fail for every user today. Removed the row/flag mentions and all 'hidden from --help' guidance from SKILL.md. The reference guide now opens with an availability gate: probe the CLI first, stop if the capability is absent, and if the optional project-scoping commands are missing, run the test set without scoping instead of retrying — so no instruction to call something help does not list.

skills/uipath-test/SKILL.md
addyGarg: Same fix — the playwright-context row is gone from SKILL.md.

skills/uipath-test/SKILL.md
addyGarg: No — removed. The project-selection detail lives only in the reference guide, behind the availability gate.

skills/uipath-test/SKILL.md
addyGarg: Validated locally: 'uip tm report get --execution-id <id>' alone exits with 'Provide --project-key or --test-set-key to identify the project.', so the requirement is real — the row now records it (and cites the verification). --query removed, including from the original text.

skills/uipath-test/SKILL.md
addyGarg: Done — the description no longer names commands; it now just says the skill covers packaging and running external Playwright test suites, which is still enough for skill activation.

skills/uipath-test/SKILL.md
addyGarg: Agreed on both counts — noted as an action item, to be done when the commands are actually in prod (not in this PR). The Pack Commands section is also removed from SKILL.md for the same reason.

skills/uipath-test/SKILL.md
addyGarg: Removed.

skills/uipath-test/SKILL.md
addyGarg: Reverted — Rule 3 is back to its original text.

skills/uipath-test/SKILL.md
addyGarg: Rephrased: look up by name with 'folders list -n <name> --all' when the user named a folder; list without --all when picking one yourself, so you only see folders you are a member of.

skills/uipath-test/SKILL.md
addyGarg: Reverted — the carve-out is gone.

skills/uipath-test/SKILL.md
addyGarg: Understood, and that release-cut timing is the deciding argument. All Playwright command surface is out of SKILL.md; the guide is availability-gated (probe, then stop or degrade). Happy to go further and hold the whole guide back if you would rather it not ship before the CLI does — say the word and I will trim the PR to the generic fixes only.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
addyGarg: Honest answer: no — it cannot pass yet (needs the unpublished CLI and an e2e tenant with serverless Playwright), and a scheduled task that always fails is just noise. I have withdrawn the e2e task and left an action item to add it once both are in place.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
addyGarg: Moot now that the e2e task is withdrawn — and yes, when it comes back it will need explicit setup/teardown for the tenant resources it creates (project + package), which is part of why it is better added when it can actually run.

tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
addyGarg: Also moot with the e2e withdrawn; the new smoke task's prompt is written the way you describe — 'I have a Playwright suite I want to run through UiPath Test Manager… tell me whether this CLI can package it' — with the workflow left to the skill.

tests/tasks/uipath-test/playwright_pack_generate.yaml
addyGarg: Added — playwright_package_prep_smoke.yaml (tier: smoke). It grades the preparation work and the capability check, so it passes on today's published CLI and stays valid after the new commands ship.

tests/tasks/uipath-test/playwright_pack_generate.yaml
addyGarg: Handled by scoping: the smoke task only writes into the sandbox working directory and creates no tenant resources, so there is nothing left behind. The tenant-touching e2e task (withdrawn) is where teardown matters, and it will carry it.

addyGarg:
## Review round applied — PR is now release-safe

Thanks @ganeshborle — all 19 comments addressed, replied inline, resolved. Two decisions worth surfacing:

**1. Nothing Playwright is advertised in SKILL.md any more.** I verified the published `@uipath/test-manager-tool` (1.198.0) carries **no** Playwright surface — no `--type playwright`, no `playwright-context`. So your release-cut concern is exactly right: anything documented would fail for every user today. Removed from SKILL.md: the `playwright-context` row, the `--playwright-projects` mention, the Playwright callout, the Pack Commands section, and all "hidden from `--help`" guidance. The generic notes I had added (JSON-envelope, CLI quirks, autonomy carve-out) are reverted, and Rules 3/7 are restored/rephrased.

The reference guide now opens with an **availability gate**: probe the CLI, stop and tell the user if the capability is absent, and if the optional project-scoping commands are missing, run the test set without scoping rather than retrying. So no path instructs an agent to call something `--help` does not list. If you would rather the guide itself not ship before the CLI does, I can trim this PR to the generic fixes only — happy either way.

**2. The e2e task is withdrawn; one smoke task added.** Honest answer to "did we run and test it?" — it cannot pass yet (needs the unpublished CLI plus an e2e tenant with serverless Playwright), and a scheduled always-failing task is just noise. `playwright_package_prep_smoke.yaml` replaces it: it grades the preparation work and the capability check, so it passes on todays published CLI and stays valid after the new commands ship, writes only into the sandbox working directory, and creates no tenant state to clean up.

### Action items (tracked, not in this PR)
1. Re-add the e2e task (with proper setup/teardown for the project + package it creates) once the CLI is published and an e2e tenant has serverless Playwright.
2. Update the cli README with the new commands — only when they are in prod.
3. Re-add the Playwright command surface to SKILL.md when the capability is live, and unhide it in the CLI at the same time.
4. TM API: add an explicit boolean to the playwright-project-context response instead of clients inferring "not Playwright" from an empty payload (raised on UiPath/cli#3226 too).

addyGarg:
## Split: held tests moved out

Per the review direction, this PR now contains only what is safe to ship today, and the tests that need the unreleased CLI live in their own branch to merge later.

**Stays here** — the skill content (lean: no command rows, no hidden-command guidance, availability-gated guide) plus one **smoke** task that passes on the currently published CLI (suite prep + capability check + the never-`link-automation` guard).

**Moved out** — the integration and e2e tasks, on branch `test/tm-playwright-integration-e2e`. They cannot pass until the external-package commands reach npm `latest` (the eval runner installs `@uipath/cli@latest` and tasks may not pin a version), so keeping them here would only add scheduled noise. Their PR carries the merge preconditions and a one-line manual run command.

Also folded the three failure-triage sections in the guide into one — same guidance, a third of the words.

addyGarg:
## Decision on the two open items

**1. The command surface stays in `SKILL.md`.** I had stripped the `playwright-context` row, the `--playwright-projects` note and the pack section on the grounds that no published CLI has them. Restoring them, with the reasoning that changed my mind: #3226 is merged into the **same 1.200 line** this skill targets (`version-manifest.json` → `targetCli: ^1.200.0`), and `uip skills install` resolves skills by the CLI's own minor — so a CLI that can pull skills 1.200 *is* a 1.200 CLI and carries all four commands. The pairing you were protecting is satisfied by construction.

The distinction that settles it: the only person who can be inconvenienced is someone who *manually* tries the Playwright flow on an older CLI, and for them the guide's availability check says stop and report — a one-off, graceful outcome. A test in CI, by contrast, breaks continuously for everyone. So the commands are documented; the tests are held (see below).

What I kept from your review: no command names in the description or in Concepts, no generic flags (`--wait`, `--query`) on command rows, and **no "hidden from `--help`" guidance anywhere** — that was your real objection and it stays fixed. The callout now states the Test Manager + CLI requirement plainly and points at the guide's availability check.

**2. Integration and e2e tasks are split into their own PR** — branch `test/tm-playwright-integration-e2e`, to merge once the commands reach npm `latest` (the eval runner installs `@uipath/cli@latest` and tasks may not pin a version). Its description carries the three preconditions and a one-line manual run command. That answers your "did we run and test it? / this runs daily" concern directly: nothing that can fail on a schedule lands here.

## Two things worth your eyes before merge

- The **reference guide** (165 lines) is the largest part of this PR and none of your 19 comments touched it — every one was on `SKILL.md` or the test YAMLs. If you want the same leanness pass applied there, say so and I will trim it (my own read: roughly 25–30 lines could go, the earned parts being the ingestion-poll trap and the no-results triage).
- The remaining smoke test is deliberately modest: most of what it grades a capable agent might do unaided, the genuinely skill-informed signal being the lockfile requirement. Real coverage lives in the held PR.

addyGarg:
## Related PRs

**UiPath/skills#2401 — integration + e2e eval tasks.** Deliberately split out of this PR: those tasks drive the full pipeline, so they cannot pass until the external-package commands reach npm `latest` (the eval runner installs `@uipath/cli@latest` and tasks may not pin a version). Keeping them here would mean a scheduled task that fails every run. This PR therefore ships the skill plus one smoke task that passes on the currently published CLI; #2401 carries its merge preconditions in the description and each task header.

**UiPath/cli#3336 — register `playwright-context` on stable, hidden from help.** Found while validating this skill against a freshly built CLI: in cli#3226 the probe was registered with `previewCommand`, which gates on the binary carrying a prerelease tag — so it was absent from every *stable* release, permanently, while `--playwright-projects` (only `hideHelp`) worked. The skill documents both, so the two must have the same availability. #3336 is the one-line fix. Until it is in a released line, a stable CLI answers `unknown command` for the probe — which the guide handles: its availability check says skip the probe and carry on, and only drop project scoping if the run flag itself is rejected.

Ordering, for anyone picking this up: this PR can merge independently (nothing in it requires an unreleased command to *function* — the guide gates on availability). #3336 should land in the same 1.200 line as cli#3226. #2401 merges last, once `npm view @uipath/cli version` reports 1.200.x and the eval tenant has Playwright support (its e2e task additionally needs working serverless execution).

## Validation status

The flow this skill describes is verified end to end on the alpha `testmanagerdev/DevTest` tenant with a CLI built from cli `main` and this skill installed: pack → upload → ingestion (~10 s) → label-filled test set → chromium-only run → **1 passed / 1 failed with the real Playwright error surfaced**, then a fix-and-republish loop (bump version, re-upload, ingestion updates in place, same test set re-run) → **2/2 passed**, confirmed by `report get`, `testcaselogs list` and the JUnit export. The gaps that run exposed are folded into the guide.

ganeshborle:
Maybe add a PR description.

skills/uipath-test/SKILL.md
```diff
@@ -85,7 +86,7 @@ Common `uip tm` commands organized by resource type.
 | `uip tm testcases steps delete --project-key <PROJECT_KEY> --step-id <UUID> --yes` | Delete a step. |
 | `uip tm testcases list-result-history --project-key <PROJECT_KEY> --test-case-id <TEST_CASE_ID>` | List test case log result history for a specific test case. Optional `--only-failed`, `--filter`, `--limit`, `--offset`. |
 | `uip tm testcases run --project-key <PROJECT_KEY> --test-case-id <TEST_CASE_ID>` | Start a new execution for one or more test cases. **Uses `--test-case-id <UUID>` (space-separated for multiple).** Optional `--async`, `--name`, `--folder-key`, `--robot-user-key`, `--machine-key`. |
-| `uip tm testcases add --test-set-key <TEST_SET_KEY> --test-case-keys <KEY1,KEY2,...>` | Add test cases to a test set (comma-separated keys). |
+| `uip tm testcases add --test-set-key <TEST_SET_KEY> (--test-case-keys <KEY...> \| --labels <name...>)` | Add test cases to a test set — by explicit keys, OR every test case carrying at least one of the given labels. Both selectors are variadic and space-separated (comma-separated keys still accepted; quote label names containing spaces). Label matching is OR, exact and case-sensitive. The two selectors are mutually exclusive. |
```
ganeshborle: Add little bit elaborate syntax like Key1,key2; label1,labe2

skills/uipath-test/SKILL.md
```diff
@@ -156,6 +160,14 @@ Common `uip tm` commands organized by resource type.
 |---|---|
 | `uip tm result download --execution-id <EXECUTION_ID>` | Download test execution results as JUnit XML. Optional `--project-key`, `--test-set-key`, `--result-path <DIR>`. |
 
+### Pack Commands (Playwright)
+
+| Command | Purpose |
+|---|---|
+| `uip tm pack --project-path <dir> --type playwright --project-key <PROJECT_KEY> --name <PackageName> --package-version <ver> -o <out-dir>` | Pack a Playwright suite into a `.nupkg` external test package. Requires a lockfile and `@playwright/test` in the project. `--project-key` targets the Test Manager project where ingestion auto-creates the test cases; `--no-create-test-cases` skips that; `--dry-run` previews. Upload with `uip or packages upload <nupkg>`. |
```
ganeshborle: Does version have a specific syntax, if yes you could add it here and/or in the cli help for this command.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,178 @@
...
+- `PW_File_<path>` — the spec file
+
+> **Do NOT run `uip tm testcases link-automation` on Playwright test cases.** They are linked by ingestion; manual linking is the RPA pipeline and will corrupt the association.
+
+> **Availability gate — check this first.** The external-package commands are newer than most installed CLIs. Run `uip tm pack --help --output json`: if it exposes no `--type` option, this pipeline is not available on this CLI — tell the user and stop rather than improvising. Step 5's probe is optional: if `testsets playwright-context` answers `unknown command`, skip it and carry on — its absence says nothing about project scoping. Only if `run --playwright-projects` itself is rejected (`unknown option`) is scoping unavailable in this build; then run the test set without it rather than retrying.
+
+## Prerequisites
+
+- A recent `@uipath/cli` — this flow's commands (`tm pack --type playwright`, `testcases add --labels`, `testsets playwright-context`, `run --playwright-projects`) do not exist on older CLIs and have no pre-rename fallback. If `uip tm pack --help` does not show `--type`, upgrade the CLI before anything else.
```
ganeshborle: This will get obsolete, release version of cli and skills are same. Do not add this line.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,178 @@
...
+- Logged in: `uip login status --output json`. If not, `uip login`.
```
ganeshborle: Lets not duplicate the rules from mail skill.md

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,178 @@
...
+
+- `--project-key` is **required** by default because test-case auto-creation is on; pass `--no-create-test-cases` to pack without it (label metadata stays embedded for later use).
+- Preview with `--dry-run` (writes nothing).
+- Capture `Data.Output` (the `.nupkg` path) and `Data.TestCount` from the JSON output. `TestCount` is the number of test cases ingestion will create — remember it for Step 3.
+- Pack fails loudly when the lockfile or `@playwright/test` is missing — fix the project, do not improvise around it.
+
+## Step 2 — Upload to Orchestrator
+
+```bash
+uip or packages upload "<out-dir>/<PackageName>.1.0.0.nupkg" --output json
```
ganeshborle, line 59: You can add, check the latest command from that package to be always safe.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,178 @@
...
+```
+
+Each re-upload needs a **new `--package-version`** at pack time — Orchestrator feeds reject an existing version.
```
ganeshborle: Mention version format

tests/tasks/uipath-test/playwright_package_prep_smoke.yaml
```diff
@@ -0,0 +1,66 @@
+# Passes on the currently published CLI and stays valid after the external-package
+# commands ship: it grades the preparation work and the capability check the skill
+# prescribes, not the packaging call itself. Everything happens in the sandbox
```
ganeshborle, line 3: Did we run this test?

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Step 3 — Wait for ingestion
+
+Ingestion is asynchronous and automatic. Poll until the auto-created test cases appear:
+
+```bash
+uip tm testcases list --project-key <PROJECT_KEY> --output json
+```
+
+- Poll **unfiltered** and count. Do NOT pass `--filter <PackageName>` — the auto-created test case *names* are `"<suite> > <test title>"`; the package name appears only in the description, which `--filter` does not search, so a package-name filter stays empty forever and reads as a false "ingestion never happened".
```
ganeshborle: What if there are multiple test cases?

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,178 @@
...
+- In a project that already holds test cases, take a **baseline count before Step 2** and wait for it to grow by `TestCount` — an absolute count can look satisfied by pre-existing rows (or time out on a large paginated list). New rows are also recognizable by their `"<suite> > <title>"` names and package description.
```
ganeshborle: Lets not have vague logic, model can hallucinate!

skills/uipath-test/SKILL.md
addyGarg: Added examples for both: --test-case-keys DEMO:1 DEMO:2 (comma form still accepted) and --labels PW_Tag_smoke "PW_Suite_Checkout flow" for the quoting case.

skills/uipath-test/SKILL.md
addyGarg: Yes — checked it against the CLI: NuGet/SemVer style, three numeric parts with an optional prerelease suffix (1.0.0, 1.0.1-beta.1); 1.0 and non-numeric strings are rejected with "Invalid package version". Documented on the pack row and in the guide. Agreed the CLI help should say it too — that is a one-line change in the cli repo, happy to raise it.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Removed. You are right that it goes stale given the lockstep release.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Removed — the login/auth bullets duplicated SKILL.md's rules.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Added: 'uip or packages list --search <PackageName>' returns Version and IsLatestVersion, so you can see the highest published version before choosing the next one.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Added there too — same NuGet/SemVer rule, with the 1.0 rejection called out.

tests/tasks/uipath-test/playwright_package_prep_smoke.yaml
addyGarg: Yes, and it caught a real bug: it ran on CI and failed at 0.88 because I had asserted playwright.config.ts by exact name while the agent wrote a valid config with a different extension. Fixed to match playwright.config.*, and I also fixed a second criterion that counted tests per file (a valid two-file suite would have failed). All four criteria re-verified locally against a real suite; the gate is green now.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Good catch — the old wording only worked for an empty project. It now matches the test case names against the tests the pack step reported (names are exactly "<suite> > <test title>"), which works whether the project already had test cases and whether this is a first upload or a re-upload.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Agreed — the baseline-count heuristic is gone. The check is now deterministic: the expected names either exist or they do not, with an explicit stop-and-report if they never appear.

ganeshborle:
Please rebase the changes.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,199 @@
...
+The key difference from the RPA pipeline: there is **no link step**. Uploading the package is enough — ingestion creates one Test Manager test case per discovered Playwright test, already bound to the package, and labels each with:
+
+- `PW_Tag_<tag>` — one per Playwright tag (`@smoke` → `PW_Tag_smoke`)
+- `PW_Project_<name>` — one per Playwright project the test runs in
+- `PW_Suite_<name>` / `PW_Path_<chain>` — describe-block grouping (name and full chain)
+- `PW_File_<path>` — the spec file
+
+> **Do NOT run `uip tm testcases link-automation` on Playwright test cases.** They are linked by ingestion; manual linking is the RPA pipeline and will corrupt the association.
+
+> **If a command is missing.** The probe in Step 5 is optional — if `testsets playwright-context` answers `unknown command`, skip it and carry on; its absence says nothing about project scoping. Only if `run --playwright-projects` itself is rejected (`unknown option`) is scoping unavailable in that build; run the test set without it rather than retrying.
```
ganeshborle: Can you please validate this? Maybe add a little bit elaboration. I didn't get this part; 
```
if `testsets playwright-context` answers `unknown command`, skip it and carry on;
```

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+- **A folder member who can run unattended** — the run executes as a robot user drawn from the folder, and `uip tm testsets run` has no flag to choose one (only `uip tm testcases run` does), so folder membership is the only lever:
+
+```bash
+uip or users list-in-folder --folder-key <FOLDER_KEY> --output json
+uip or users get <USER_KEY> --all-fields --output json      # needs MayHaveUnattendedSession: true
+```
+
+An ordinary interactive user usually has `MayHaveUnattendedSession: false`, and `uip or users update --allow-unattended` cannot fix that here — it requires a Windows unattended username and password, which serverless does not use. Assign an already-unattended-capable principal instead (a `DirectoryRobot` account is the reliable pick). Creating your own folder is a first-class option when no existing one qualifies: `uip or folders create <NAME> --output json` (name is positional), then attach a machine and such a user.
```
ganeshborle: Does orchestraor have this defined in their skills? If yes, lets add the reference here so that is something changes then our skill will always be uptodate.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Step 5 — Probe the Playwright context (when available)
+
+Before deciding whether `--playwright-projects` applies, ask the server:
+
+```bash
+uip tm testsets playwright-context --test-set-key <TEST_SET_KEY> --output json
+```
+
+- `Data.IsPlaywright: true` → the set resolves to one Playwright package; `AvailablePlaywrightProjects` holds the only valid `--playwright-projects` values, and `SelectedPlaywrightProjects` shows any selection already stored on the test set. Both are **comma-joined strings** (`"chromium, firefox"`), not arrays — split on `", "` when scripting; no stored selection is `""`.
```
ganeshborle, line 137: Make sure to keep these output parameters in sync whenever the API changes and if notl try to use some generic way to find or make sense of this info.

skills/uipath-test/references/playwright-first-mile-guide.md
```diff
@@ -0,0 +1,199 @@
...
+
+- **`Cancelled` within seconds**, logs pointing at `CreateTestAutomationJobs` / `InternalServerError` → the default folder has no serverless machine assigned (Step 4).
+- **`Finished` with `Passed: 0 / Failed: 0 / None: N`** → the run ended without per-test results reaching Test Manager; the test case log's `Info` carries the reason (results upload failed, or the job never started). Note `report get` counts `None` as **`Skipped`**, so 0% here means "results lost", not "tests skipped".
+- **Stuck `Pending`** → dispatch worked but nothing is executing. A faulted job may never sync back, so don't sit out the 30-minute wait; check after ~5 minutes. Fixing the folder does **not** rescue an already-pending execution — start a new run after fixing.
+- **`Serverless.Runtime.CannotIssueUserTokenDueToUserNotPartOfOrg`** in a log's `Info` → the folder's robot-user pool is bad: the account Test Manager picked is still folder-assigned but no longer valid in the org. Unlike the shapes above this **is** fixable from the CLI — point the project at a folder whose members satisfy Step 4 (or create one), then re-run.
```
ganeshborle, line 179: Add reference of the orchestrator or robot skill so that it can clearly follow the troubleshooting steps.

skills/uipath-test/SKILL.md
```diff
@@ -85,7 +86,7 @@ Common `uip tm` commands organized by resource type.
...
 | `uip tm testcases list-result-history --project-key <PROJECT_KEY> --test-case-id <TEST_CASE_ID>` | List test case log result history for a specific test case. Optional `--only-failed`, `--filter`, `--limit`, `--offset`. |
 | `uip tm testcases run --project-key <PROJECT_KEY> --test-case-id <TEST_CASE_ID>` | Start a new execution for one or more test cases. **Uses `--test-case-id <UUID>` (space-separated for multiple).** Optional `--async`, `--name`, `--folder-key`, `--robot-user-key`, `--machine-key`. |
-| `uip tm testcases add --test-set-key <TEST_SET_KEY> --test-case-keys <KEY1,KEY2,...>` | Add test cases to a test set (comma-separated keys). |
+| `uip tm testcases add --test-set-key <TEST_SET_KEY> (--test-case-keys <KEY...> \| --labels <name...>)` | Add test cases to a test set — by explicit keys, OR every test case carrying at least one of the given labels. Both selectors are variadic and space-separated — `--test-case-keys DEMO:1 DEMO:2` (comma-separated `DEMO:1,DEMO:2` still accepted), `--labels PW_Tag_smoke "PW_Suite_Checkout flow"` (quote names containing spaces). Label matching is OR, exact and case-sensitive. The two selectors are mutually exclusive. |
```
ganeshborle:
```suggestion
| `uip tm testcases add --test-set-key <TEST_SET_KEY> (--test-case-keys <KEY1,KEY2,...> \| --labels <Label1,Label2,...>)` | Add test cases to a test set — by explicit keys, OR every test case carrying at least one of the given labels. Both selectors are variadic and space-separated — `--test-case-keys DEMO:1 DEMO:2` (comma-separated `DEMO:1,DEMO:2` still accepted), `--labels PW_Tag_smoke "PW_Suite_Checkout flow"` (quote names containing spaces). Label matching is OR, exact and case-sensitive. The two selectors are mutually exclusive. |
```

tests/tasks/uipath-test/playwright_package_prep_smoke.yaml
```diff
@@ -0,0 +1,66 @@
+# Passes on the currently published CLI and stays valid after the external-package
+# commands ship: it grades the preparation work and the capability check the skill
```
ganeshborle, line 2: As this is playwright is the bigger feature, lets add the integration and e2e test as well.

ganeshborle (approved):
Looks good to me :) 
There are some minor comments, if you can take care!

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Validated and elaborated. The two hidden commands behave differently when a build lacks them, which is what was muddled: a missing **probe** costs only the pre-check — it just reports whether a test set is Playwright, so you skip Step 5 and project scoping still works. A rejected **--playwright-projects** means that build cannot scope a run at all, so you run the test set without it (every project in the package config runs). Both paths confirmed against a stable CLI build. Note --help is not a reliable signal for either, since both are hidden from it — that is now stated up front.

skills/uipath-test/references/playwright-first-mile-guide.md
addyGarg: Yes — uipath-platform owns it. Added links to /uipath:uipath-platform § orchestrator/setup-environment.md (folders, machines, the one-serverless-machine-per-folder rule) using the same relative-link form publish-and-link-guide.md already uses for uipath-rpa, so the detail tracks the owning skill instead of drifting here.

skills/uipath-test/references/playwright-first-mile-guide.md
addy1234: Added a line telling the reader to take the fields off the JSON response rather than depend on the list staying current — the list now describes today's shape only. Kept the field names because IsPlaywright is the branch condition an agent needs to know exists, but the generic instruction is what governs.

skills/uipath-test/references/playwright-first-mile-guide.md
addy1234: Added — the triage now says that everything past the Test Manager boundary (job states, machines, folder membership) belongs to /uipath:uipath-platform § orchestrator/run-jobs.md and § setup-environment.md, with links.

skills/uipath-test/SKILL.md
addy1234: Took the fuller placeholders, with one correction: labels are not comma-splittable. The CLI splits --test-case-keys on commas but not --labels, so `--labels A,B` is read as a single label named "A,B". Verified live — it fails with "No test cases carry any of the labels: PW_Tag_smoke,PW_File_tests/orders.spec.ts" while the same two space-separated add both test cases. The row now shows `<Label1> <Label2> …` and states the asymmetry explicitly, since silently matching nothing is a nasty trap.

tests/tasks/uipath-test/playwright_package_prep_smoke.yaml
addy1234: They exist — in UiPath/skills#2401, split out because both drive the full pipeline and the eval runner installs @uipath/cli@latest (tasks may not pin a version), so they cannot pass until the commands reach npm latest; here they would fail on every scheduled run. That PR lists the three merge preconditions and a one-line manual run command.

PR #2324, opened 2026-07-29
fix(uipath-maestro-bpmn): align authoring with Alpha runtime

DevMomo:
## Summary

Teach `uipath-maestro-bpmn` the Studio Web and Alpha runtime contracts exposed while building and running a real Jira + Google Drive + Slack orchestration.

- author v3 ScriptTasks with stable variable ids, typed response/error variables, explicit start/end bridges, and scoped subprocess output propagation;
- model typed boundary recovery and multi-instance iterator/collection behavior according to the observed runtime contract;
- render discovered Integration Service activities with exact operation identity, static paths, parameter targets, process bindings, and connection resources;
- derive connector request bodies only from enrichment fields with `Request=true`, require every required request field, omit response-only fields even when `RequestCurated=true`, preserve exact `Name`, and expand dotted names into nested JSON;
- keep named-project registry evidence inside `<project>/registry-evidence/` while preserving workspace-root evidence for standalone discovery tasks;
- bound live-resource discovery, require explicit approval before writes, and reject incomplete connector/runtime contracts before rendering;
- clean up ephemeral Studio Web solutions created by the existing live-debug regression task.

This is the independently mergeable runtime/connector layer above UiPath/skills#2356. The new customer-escalation eval remains isolated in draft PR UiPath/skills#2203.

## Why

Studio Web import and the Alpha runtime enforce contracts that the former registry guidance and local builder did not fully describe. Real runs exposed incorrect ScriptTask serialization, missing public/subprocess bridges, incomplete connector identity and binding metadata, invalid multi-instance assumptions, duplicate evidence locations, and request bodies built from response fields.

## Validation

- 44/44 declarative-builder tests pass locally: 37 inherited foundation cases plus 7 connector/runtime cases
- 4/4 solution-cleanup helper tests pass locally
- generated example renders and passes `uip maestro bpmn validate` with the installed CLI
- skill-status manifest, README, YAML parsing, and `git diff --check` pass
- unrelated eval-criterion rewrites were removed from this layer

The comprehensive nine-scenario Alpha business-outcome run is intentionally not claimed here; it belongs to the draft eval PR above this layer and is still being hardened.

## Native stack / dogfooding boundary

For dogfooding, merge:

1. independent CLI prerequisite UiPath/cli#3288;
2. builder foundation UiPath/skills#2356;
3. **this PR**.

Stop there. UiPath/skills#2203 is a test-only top layer and does not block delivery of these skill changes.

### DIFF
--- a/skills/uipath-maestro-bpmn/SKILL.md
+++ b/skills/uipath-maestro-bpmn/SKILL.md
@@ -71,15 +71,27 @@ JSON discovery command except `registry pull` into its evidence file. Preserve
 the first pull result from stdout rather than invoking it again: the session
 gets one pull, and all later discovery uses that cache.
 
+Choose one evidence location from the requested deliverable before writing:
+
+- If the user names an exact evidence path, use that path.
+- When the task names or creates a project directory, write evidence under
+  `<project>/registry-evidence/`.
+- Use workspace-root `registry-evidence/` only for an evidence-only task that
+  has no project deliverable.
+
+Do not create a second evidence directory at workspace root while authoring a
+named project.
+
 Treat requests to discover before authoring, save raw registry JSON/evidence,
 or "do not author yet" as discovery-only even if they describe an eventual
-BPMN. In that mode, immediately create `registry-evidence/`, run and save
-`registry pull --output json`, `registry list --output json` or `registry search
-... --output json`, and `registry get <type> --output json` for each requested
-type; do not read deep authoring references or scaffold a project. For authoring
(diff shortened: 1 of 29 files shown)


DevMomo:
Superseded as a catch-all by atomic, environment-neutral work. Generic resource resolution is now #2370; project/public-I/O/ScriptTask contracts are #2363/#2369/#2366. Integration Service authoring will return only after its versioned CLI/frontend contract is proven, and Alpha/provider-specific behavior remains isolated in the live eval. The branch is preserved for reference.

PR #2330, opened 2026-07-29
fix(tests): publish-lifecycle — publish before retraining so two versions exist

andrei-uipath:
`skill-ixp-integration-publish-lifecycle` fails for codex roughly 60% of the time and always passed for claude. It is not agent quality and it is not flakiness in the service — it is an ordering dependency the prompt never stated. Pre-existing; unrelated to #2316 / #2317 (confirmed by reproducing it on plain `main` at the same score).

## Root cause: publishing pins a version against cleanup

Publishing a model version pins it. An **unpublished** version is discarded when the next one trains. The task asks for two trained versions to tag, but told the agent to retrain first and tag afterwards — by which point only one version was left.

The failing transcripts say it directly:

> `list-models` reports only trained version 2. Version 0 was rejected as not found, so there is no distinct V_OLD available to tag or move.

Grouped across 10 live runs, the split is clean and has no exceptions:

| Order | Versions left | Outcome |
|-------|---------------|---------|
| `publish` → **then** retrain | 2 | **passed** — 2 codex + claude |
| retrain → **then** `publish` | 1 | **failed** — every run |

Claude never passed by handling the lifecycle better. It passed because it happened to publish v0/v1 before its second retrain. The two codex runs that passed did the same thing.

## Fix

**1. Publish before retraining** (the actual fix). The prompt now puts the first trained version on staging *before* tweaking the instruction to retrain the second, so two versions exist by construction — no recovery path needed:

> Once that first version has trained, put it on staging right away — before you retrain. An unpublished version is discarded when the next one trains, so publishing it is what keeps two versions around to tag.

**2. Stop, don't fake it.** Version numbers come from `list-models` rather than training order, and if the two versions aren't both present the agent stops and reports. An earlier attempt at this told the agent to retry other numbers instead — that pushed one run into completing the sequence with a *single* version, tagging v2 as both staging and live. `check_tags.py` caught it, but the instruction was inviting a passing-looking false end state. Removed.

**3. Two grading fixes** surfaced along the way:

- **"Published with a tag" rewarded rejected calls.** `calls.log` records the invocation whether or not the server accepted it, so a run that only *attempted* the tagging banked 2.0 of 11.0 while landing none of it — most of codex's 0.273. Dropped to supporting weight (1.0); the wire effect is `check_tags.py`'s job at 3.0.
- **"Re-read list-models after each step" was satisfiable by polling.** It counted four `list-models` calls, and the agent legitimately polls `list-models` through two training cycles — so it passed without a single transition being snapshotted. Replaced by the four snapshot files (0.5 each, same 2.0 total). Those are `check_tags.py`'s inputs, so a missing one is now a named failure instead of a `FileNotFoundError` traceback — which is how this whole thing first surfaced confusingly.

`check_tags.py` is unchanged: it already resolved versions through the `staging`/`live` tag pointers and asserted only relative transitions.

Net: 10 criteria, total weight 10.0. `check-cli-verbs.py` and `check-task-driver.py` clean.

## Verification

Baseline on `main`, codex, 5 runs: **2/5 pass** (failures at 0.455 / 0.273 / 0.091).

On this branch, codex, 5 runs — **1 complete so far, 4 still running.** I will post the full tally as a comment before this is ready to merge; please don't merge on the strength of the story alone.

First run: **1.000 SUCCESS**, and the call order is the intended mechanism with no rejection or retry anywhere:

```
publish v1 --tag staging      <- before the retrain
update-prompts                <- retrain
publish v2 --tag live
publish v1 --tag live         <- move
untag --tag live
untag --tag staging
unpublish v2
```

An intermediate commit on this branch (version discovery + retry, without the reordering) measured **3/5** — better than baseline but inside the noise at n=5, which is what prompted digging further and finding the ordering cause. The history is kept rather than squashed because the failed attempt is what produced the evidence.

## Follow-up worth considering separately

"Publishing pins a model version; an unpublished older version is discarded when the next one trains" is real IXP behavior that the `uipath-ixp` skill documents nowhere — not in the publish/rollback guidance, not in `cli-reference.md`. An agent doing this for a user would hit exactly what these runs hit. That is arguably more valuable than this test change, but it is a skill edit rather than a test edit, so it is not in this PR.

https://claude.ai/code/session_01HfNsX5UrVVqZfHUFhsCn6E

### DIFF
--- a/tests/tasks/uipath-ixp/integration/publish_lifecycle.yaml
+++ b/tests/tasks/uipath-ixp/integration/publish_lifecycle.yaml
@@ -6,11 +6,14 @@ description: >
   the published flag / tag pointer actually moved (wire-effect), not just command
   shape.
 
-  Version numbers come from `list-models`, never from training order: what a
-  retrain numbers a version is not guaranteed, and publishing a guessed number
-  returns "model version not found". An agent that guesses and then stops is
-  behaving correctly, so the prompt makes the read explicit rather than leaving
-  the numbering to be inferred.
+  Ordering matters and the prompt pins it: the first version is published BEFORE
+  the retrain that produces the second. Publishing pins a version; an unpublished
+  one is discarded when the next trains, leaving a single version and no distinct
+  V_OLD to tag or move. Measured over 10 live runs — every run that published
+  before its final retrain passed, every run that retrained first found one
+  version left and could not complete the sequence.
+
+  Version numbers also come from `list-models`, never from training order.
 tags: [uipath-ixp, integration, mode:operate]
 
 run_limits:
@@ -35,8 +38,13 @@ initial_prompt: |
   signed into (check I'm logged in; don't re-authenticate): start a project from
   the files in `./fixtures/` (give it a unique title
   `codereval-integ-publish-<random-suffix>` so concurrent runs don't collide) — import the taxonomy, upload one of the invoices —
-  and let it train. I want two trained versions to tag, so once the first has
-  trained, tweak the "Total Amount" field's instruction to retrain a second.
+  and let it train.
+
+  Once that first version has trained, put it on staging right away — before you
+  retrain. An unpublished version is discarded when the next one trains, so
+  publishing it is what keeps two versions around to tag. Then tweak the
+  "Total Amount" field's instruction to retrain a second version.
+
   Work only on the project you create; other runs share this tenant, so never
   list, read, or modify any other project.
 
@@ -46,14 +54,14 @@ initial_prompt: |
   only once training has completed; if it's still training after 5 checks, stop
   and report rather than waiting longer.
 
-  Take the two version numbers you tag from `list-models` — read them off its
-  output rather than assuming what a retrain numbered them. If a publish reports
-  the version was not found, re-read `list-models` and use the numbers it
-  reports; don't stop on the first rejection.
+  Take every version number you use from `list-models` — read it off that output
+  rather than assuming what a retrain numbered it. If the two versions you need
+  are not both there, stop and report; do not tag one version twice to finish the
+  sequence.
 
   Then, saving the model list after each step so I can see the state move
-  (call the older of the two versions V_OLD and the newer V_NEW):
-  - Put V_OLD on staging and V_NEW on live → save to `models_tagged.json`.
+  (V_OLD = the version you put on staging, V_NEW = the one trained after it):
+  - Put V_NEW on live → save to `models_tagged.json`.
   - Move the live tag over to V_OLD → save to `models_after_move.json`.
   - Take the tags off V_OLD → save to `models_after_untag.json`.
   - Unpublish V_NEW → save to `models_after_unpublish.json`.
--- a/tests/tasks/uipath-ixp/integration/publish_lifecycle.yaml
+++ b/tests/tasks/uipath-ixp/integration/publish_lifecycle.yaml
@@ -5,6 +5,12 @@ description: >
   staging/live, moving a tag to another version, unpublish, and untag — asserting
   the published flag / tag pointer actually moved (wire-effect), not just command
   shape.
+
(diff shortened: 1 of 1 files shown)


andrei-uipath:
## Validation complete: 5/5 pass, all 1.000

| | codex, 5 runs |
|---|---|
| baseline (`main`) | **2/5** — 1.000, 1.000, 0.455, 0.273, 0.091 |
| version-discovery + retry (intermediate commit) | **3/5** — inside noise at n=5 |
| **this branch (publish before retrain)** | **5/5 — 1.000 ×5** |

Runs: [30453328964](https://github.com/UiPath/skills/actions/runs/30453328964) · [30453336386](https://github.com/UiPath/skills/actions/runs/30453336386) · [30453344082](https://github.com/UiPath/skills/actions/runs/30453344082) · [30453351901](https://github.com/UiPath/skills/actions/runs/30453351901) · [30453360007](https://github.com/UiPath/skills/actions/runs/30453360007)

Not just green — green for the right reason. Every run issued the same call sequence, with no version rejection and no retry anywhere:

```
publish v1 --tag staging      <- before the retrain
update-prompts                <- retrain
publish v2 --tag live
publish v1 --tag live         <- move
untag --tag live
untag --tag staging
unpublish v2
```

That is the ordering the fix imposes, and it removes the failure entirely rather than making the agent recover from it. Durations 191–399s, consistent with two clean training cycles.

This satisfies the passing-run claim. Ready for review.

andrei-uipath:
## Second validation round: 10 more codex runs — 7/10, zero version-ordering failures

Total on this branch: **12/15 pass**, against a **2/5** baseline. More importantly, none of the 15 failed on the mechanism this PR fixes.

| Round | Concurrency | Result |
|---|---|---|
| first (5 runs) | sequential-ish | **5/5** — 1.000 x5 |
| second (10 runs) | 10 concurrent | **7/10** — 1.000 x7 |

### All three failures in round 2 are environmental, and I caused them

Running 10 live-tenant runs at once put more load on the shared tenant than any earlier round. Two runs never got a project created:

- [30455491383](https://github.com/UiPath/skills/actions/runs/30455491383) — 0.000, died at **89s**: *"project creation hit a 429... the subsequent retry returned no result at all. Because that retry may have created a project, I did not retry again or list shared tenant projects to recover its name."*
- [30455517208](https://github.com/UiPath/skills/actions/runs/30455517208) — 0.000, died at **94s**: *"IXP project creation is forbidden (HTTP 403). No project was created or modified."*

Both aborted before reaching any publish/tag step — 89s and 94s against 486-717s for a normal run. Note both agents refused to list shared-tenant projects to recover, which is the behavior the prompt asks for.

The third ran to full length and confirms the fix working:

- [30455546853](https://github.com/UiPath/skills/actions/runs/30455546853) — 0.100, 534s: *"Version 0 trained and **was published to staging before the Total Amount instruction was updated**. The second training poll was still in progress when this run ended."*

That is exactly the ordering this PR imposes. It then hit the prompt's own stop-condition ("if it's still training after 5 checks, stop and report") because training was slower under 10x concurrency — 5 checks x 2 min = 600s, and it stopped at 534s. Correct behavior against a slow tenant, not a lifecycle failure.

Successful runs also stretched from 191-399s in round 1 to 486-717s here, consistent with tenant contention.

### Reading

Baseline failures were all the same mechanism: retrain first, first version discarded, no distinct version to tag. Across 15 runs on this branch that mode did not occur once. The three failures here are tenant throttling (x2) and a training cycle outrunning the polling budget under self-inflicted load — all reproducible without this change, and all independent of the tag/publish lifecycle.

If you want a cleaner number, re-running the 10 with `parallelism` staggered over time rather than all at once should avoid the 429/403 entirely. Happy to do that; I did not want to present 7/10 without explaining what the 3 were.

misupantea-uipath (approved):

PR #2333, opened 2026-07-29
test(maestro-flow): grade jdbc-databricks skill load with skill_triggered, not a Skill-tool match

rockymadden:
## Problem

The `skill-flow-jdbc-databricks-query` task fails (`final_status: FAILURE`) even when the agent produces a fully correct artifact.

Root cause is a single **gating** success-criterion — *"Agent loaded the uipath-maestro-flow skill"* — that scored `0.0`:

```yaml
- type: command_executed
  description: "Agent loaded the uipath-maestro-flow skill"
  tool_name: "Skill"          # ← only matches Claude Code's first-class Skill tool
  command_pattern: "uipath-maestro-flow"
  min_count: 1
```

The graded run was a **codex** agent (`agent_type: codex`, `model_used: gpt-5.6-terra`). Codex has no `Skill` tool — it loads a skill by **reading `SKILL.md` via Bash** (the very first command in the trace was `sed -n '1,240p' .../uipath-maestro-flow/SKILL.md`). So the matcher found 0 `Skill`-tool calls → `0/1` → score `0.0`. Because the criterion is `gating: true`, the whole task is marked `FAILURE`.

Everything else passed:

| Criterion | Score |
|---|---|
| `flow validate` was called | ✅ 1.0 |
| `flow debug` was called | ✅ 1.0 |
| `check_databricks_query.py` (JDBC gateway node + bound connection, not native/HTTP) | ✅ 1.0 |
| **Skill loaded** (`tool_name: "Skill"`) | ❌ **0.0** |

The weighted score confirms this is the *only* miss: passing weight `2.0 + 1.5 + 3.0 = 6.5` over total `7.5` = **`0.867`**, exactly the reported `weighted_score`.

> Note: `review.json`'s human summary ("agent could not execute required flow commands") is inaccurate — the trace shows validate + debug both ran and succeeded. The real cause is the skill-load probe keying on `tool_name: "Skill"`.

## Why it's a repo-wide outlier

This was the **only** task in the repo using `tool_name: "Skill"` (1 occurrence). The established, agent-agnostic convention is the dedicated **`skill_triggered`** criterion — **454 usages across 433 tasks**, and the documented *"un-fakeable"* primitive in `.claude/commands/generate-task.md`. `coder_eval` owns `skill_triggered` detection per agent type (Skill tool for Claude Code, `SKILL.md` read for codex/antigravity), which is exactly why it's used in the activation benchmark that runs codex.

## Fix

Replace the brittle `command_executed`/`tool_name: "Skill"` block with the canonical `skill_triggered` criterion used by every other maestro-flow task, preserving the original weight (`1.0`):

```yaml
- type: skill_triggered
  skill_name: "uipath-maestro-flow"
  expected_skill: "uipath-maestro-flow"
  description: "Agent invoked the uipath-maestro-flow Skill"
  weight: 1.0
  pass_threshold: 1.0
```

This fixes the false-negative for codex (and any non-Claude-Code agent) without loosening what's asserted — the flow's structural correctness is still fully graded by the three unchanged criteria.

## Testing

- YAML parses; `success_criteria` types are now `[skill_triggered, command_executed, command_executed, run_command]`; no residual `tool_name: "Skill"`.
- Change is a 1-block, convention-conforming edit identical in shape to the 400+ existing `skill_triggered` criteria.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/connector_features/jdbc_databricks_query/jdbc_databricks_query.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/jdbc_databricks_query/jdbc_databricks_query.yaml
@@ -56,11 +56,10 @@ initial_prompt: |
   Do NOT pause between planning and implementation.
 
 success_criteria:
-  - type: command_executed
-    description: "Agent loaded the uipath-maestro-flow skill"
-    tool_name: "Skill"
-    command_pattern: "uipath-maestro-flow"
-    min_count: 1
+  - type: skill_triggered
+    skill_name: "uipath-maestro-flow"
+    expected_skill: "uipath-maestro-flow"
+    description: "Agent invoked the uipath-maestro-flow Skill"
     weight: 1.0
     pass_threshold: 1.0
 


rockymadden:
### Verification: this preserves the assertion, it doesn't remove it

A reasonable concern on this change is *"swapping the criterion just deletes the skill-load check."* It does not. `skill_triggered` is agent-agnostic and still fails when the skill isn't engaged.

`coder_eval` v0.7.1 `criteria/skill_triggered.py` — detection is two-signal:

```python
def _engaged_skill(cmd, skill_name) -> bool:
    # Claude Code: explicit Skill tool call
    if cmd.tool_name == "Skill" and cmd.parameters.get("skill","").split(":")[-1] == skill_name:
        return True
    # Codex / any non-Claude agent: engaged by reading SKILL.md off disk via shell
    needle = f"skills/{skill_name}/"
    return any(isinstance(v, str) and needle in v for v in cmd.parameters.values())
```

Running that exact function against this task's real transcript:

| Command (from the failing run) | detected |
|---|---|
| codex's first command: `sed … skills/uipath-maestro-flow/SKILL.md` | **True** → 1.0 |
| Claude Code `Skill` tool call | **True** (backward compatible) |
| agent that never engaged the skill | **False** → 0.0 |
| agent that read a *different* skill | **False** (trailing slash guards prefix collisions) |

The old `command_executed` + `tool_name: "Skill"` check could only ever match a Claude `Skill` tool call — impossible for a codex agent, which has no such tool — so it failed the run regardless of whether the skill was used. `skill_triggered` asserts the intended behavior ("the agent engaged the skill") and recognizes both the Claude and codex engagement signals, while still scoring 0.0 when the skill is genuinely skipped.

nikhil-maryala (approved):

rockymadden:
Superseded by #2309 ("fix ipe eval failures for codex"), which landed on `main` first and resolved the same root cause — the codex false-negative on the skill-load criterion — by removing the broken `command_executed`/`tool_name: "Skill"` block outright. The original FAILURE this PR targeted is already fixed on `main`, so this change is redundant. Closing.

PR #2334, opened 2026-07-29
fix(tests): hide troubleshoot mock scripts and call log from the agent sandbox

dmorosanu:
## Problem

`m/seal` hides the fixture data (`r/` -> opaque `.store`), but the mock scripts themselves stay plainly readable in the agent's working directory:

- `m/uip` - its module docstring documents the entire mock system: manifest schema, the `.store` blob format (zlib+base64, trivially decodable once known), the `_doc`-annotation stripping pass, and dispatch rules.
- `m/seal` - its docstring states outright what was hidden from the agent and how.
- `m/.calls.jsonl` - written live during the run; each record names the matched manifest rule and fixture file, readable proof the CLI is mocked.

An agent that lists/reads these mid-investigation learns how the test evidence is kept and that it is inside a test harness.

## Solution

Sealing now also hides the mock machinery itself, plus a skill-level guard (one commit per fix):

1. **Hide the scripts** - after packing `.store`, `m/seal` compiles `m/uip` to docstring-free bytecode (`optimize=2`; comments never reach bytecode), rewrites `m/uip` and `m/uip.cmd` as thin exec stubs pointing at the bytecode (interpreter path baked in so the compiling interpreter runs the pyc), and blanks itself. Blank rather than delete so a `python m/seal` re-run in a reused sandbox still exits 0.
2. **Encode the call log** - on sealed runs each call record is written zlib+base64-encoded to `m/.log` (plain JSONL is kept for unsealed local debugging). `coverage_report.py` decodes both formats, keeps reading legacy `.calls.jsonl` artifacts, and gains `--dump` for manual triage.
3. **Fix sealed-artifact analysis** - `coverage_report.py` now reads the manifest from `m/.store` when `r/` is gone (it was broken for sealed run artifacts).
4. **Skill invariant** - opacity alone cannot stop deliberate decoding/decompilation, so the skill now forbids it behaviorally: new invariant 12 in `skills/uipath-troubleshoot/SKILL.md` - files that back or intercept the `uip` CLI (dispatchers, fixture stores, shims, caches, call logs) are internal storage; never read, decode, decompile, or bypass them for evidence. Scoped to backing files only: evidence sources a guide or playbook prescribes directly (Assistant diagnostic-archive logs, workflow source files) are explicitly unaffected. Mirrors the rule in `skills/uipath-maestro-bpmn/references/diagnose/CAPABILITY.md`, previously the only such instruction in the repo.
5. **Idiomatic bytecode location** - the pyc lands at `m/__pycache__/uip.cpython-<ver>.pyc` (where the interpreter itself would put it), so a listing/grepping agent sees a routine interpreter artifact instead of a planted hidden blob; the dispatcher anchors its data paths on the mock dir when executed from `__pycache__/`.
6. **Resume partial seals** - a crash mid-seal leaves the manifest and `seal` in place, so the pre_run retry resumes (compile skip-guarded, store committed before `r/` removal, `seal` blanked last). `.store` alone is never treated as proof the seal completed.

No task.yaml changes: all scenarios already run `python m/seal` in `pre_run`. Repo sources stay fully documented; only the staged sandbox copy is transformed at seal time.

## Validation

- Local sandbox simulation: sealed dispatcher output is byte-identical to unsealed for canned rules, unmocked-default, via the sh stub, `uip.cmd`, and bare `uip` on PATH; `strings` on the pyc shows no docstring sentences; re-running `python m/seal` after sealing exits 0. Repeated after the `__pycache__` move; resume verified from both partial states (store-present/scripts-readable, stubs-present/store-missing).
- Passing coder-eval runs against this branch (`SKILLS_REPO_PATH` -> worktree), all `skill-troubleshoot-classic-browserscope-efail`: score 1.000 (run `2026-07-29_17-07-46`, sealing only), score 0.962 (run `2026-07-29_17-17-15`, with initial invariant 12), score 1.000 (run `2026-07-29_18-18-57`, with the final scoped invariant 12). Artifact sandboxes confirmed: stubbed launchers, bytecode, blank `seal`, encoded `.log`; `coverage_report.py` and `--dump` verified against real artifacts.

## Follow-up (out of scope)

`uipath-ixp`, `uipath-maestro-bpmn`, and `uipath-mcp-servers` maintain their own mock templates with readable `uip` shims and plain call logs - same leak class, separate dispatchers.


dmorosanu:
### Code review

Found 1 issue:

1. Partial seal is silently accepted on retry: `main()` treats `.store` existence alone as "already sealed" and exits 0, but `.store` is written before `shutil.rmtree(r/)` and `_hide_scripts()` run. If anything after the store write fails (`rmtree` on a Windows-locked file, `py_compile` raising with `doraise=True`), the first run fails loudly via `fail_on_error`, but a re-run in a reused sandbox hits the early return and reports success while `m/uip` is still the fully readable source - reintroducing exactly the leak this PR closes. Write `.store` last (or make the already-sealed check also require the compiled artifact) so a partial seal can never pass as sealed.

https://github.com/UiPath/skills/blob/da4bf0a67ed6625b19be067df9610bac3efa6c8c/tests/tasks/uipath-troubleshoot/_shared/mock_template/m/seal#L105-L108

https://github.com/UiPath/skills/blob/da4bf0a67ed6625b19be067df9610bac3efa6c8c/tests/tasks/uipath-troubleshoot/_shared/mock_template/m/seal#L120-L131

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dmorosanu:
Review issue 1 addressed in d0db91ee8: the seal now hides scripts before committing `.store`, writes the store before deleting `r/`, and blanks itself last - a crash at any step leaves the manifest and script in place so the pre_run retry resumes the seal (compile is skip-guarded on an existing pyc). `.store` alone is no longer treated as proof of a completed seal. Verified with resume tests from both partial states (store-present/scripts-readable and stubs-present/store-missing).

tests/tasks/uipath-troubleshoot/_shared/mock_template/m/seal
```diff
@@ -44,10 +56,63 @@ CACHE_SRC = RESPONSES_DIR / "_cache"
...
+
 def main() -> int:
-    # Already sealed, or nothing to seal (non-mock task) → no-op.
-    if STORE_PATH.exists():
-        return 0
+    # Nothing to seal: a non-mock task (mock template never staged with
+    # fixtures), or a reused sandbox where sealing already completed and
+    # removed `r/`. A PARTIAL seal always leaves `r/manifest.json` in place
+    # (the store write and `rmtree` happen last), so reaching this branch
+    # means there is no readable fixture left to hide.
```
costin-uipath, line 115: complete partial-seal recovery. When the manifest is missing but .store exists, finish _hide_scripts(), remove any residual r/, blank seal, then return. Update the docstring’s “always leaves the manifest” claim.

```
if not MANIFEST_PATH.is_file():
    if STORE_PATH.is_file():
        _hide_scripts()
        if RESPONSES_DIR.exists():
            shutil.rmtree(RESPONSES_DIR)
        (SCRIPT_DIR / "seal").write_text("", encoding="utf-8")
    return 0
```

costin-uipath (approved):

PR #2348, opened 2026-07-29
Update Maestro case reasoning rules

abhiram-vad:
## Summary
- Add a scoped SLA response model for case, stage, and action-task deadlines, including the rule that stage SLA breaches must not interrupt same-stage tasks.
- Expand Phase 0/SDD review output with Task-set Map, SLA Response Map, Rule Firing Map, and Re-entry Loop Map.
- Preserve frontend task-set semantics for parallel-after-predecessor work and default task run-once behavior from SDD re-entry semantics.
- Add a regression eval covering SLA, payment/deadline, selected-task, user-selected-stage, and re-entry loop issues.

## Validation
- `npm run version:check`
- `git diff --check`
- `python3 -m json.tool skills/uipath-maestro-case/references/evals/evals.json`
- `python3 -m py_compile tests/tasks/uipath-maestro-case/phase_0_to_case/case_reasoning_regressions/check_case_reasoning_regressions.py tests/tasks/uipath-maestro-case/phase_0_to_case/procurement_sla_interrupts/check_procurement_sla_interrupts.py`
- YAML parse for `tests/tasks/uipath-maestro-case/phase_0_to_case/case_reasoning_regressions/case_reasoning_regressions.yaml`
- `python3 scripts/check-skill-status.py`
- `python3 scripts/check-task-driver.py`

## Note
- `python3 scripts/check-cli-verbs.py` fails in this environment before repo inspection because Python 3.9.6 lacks `sre_parse.POSSESSIVE_REPEAT`.


charlesliu9:
- Warning: The new regression requires Wait for Payment Confirmation to be parallel-after-predecessor with runs-sequentially, but the task-mode rules classify a connector callback as event-triggered / wait-for-connector and explicitly prohibit runs-sequentially. This produces conflicting instructions for exactly the new payment scenario, so an agent can follow one reference and fail the regression. Clarify that a connector-wait task may use the ordered task-set entry rule, or revise the regression/model to keep it event-triggered. /skills/skills/uipath-maestro-case/references/planning.md:321, /skills/skills/uipath-maestro-case/references/plugins/conditions/task-entry-conditions/planning.md:45, tasks/uipath-maestro-case/phase_0_to_case/case_reasoning_regressions/check_case_reasoning_regressions.py:110

PR #2359, opened 2026-07-30
feat(ci): auto-recompile troubleshoot mock bytecode when sources change

dmorosanu:
## Problem

#2358 ships the troubleshoot mock dispatcher and sealer as compressed docstring-stripped blobs (`mock_template/m/.uip.bin` / `.seal.bin` — AST-stripped source, zlib+base64) with editable sources in `_shared/mock_src/`. Nothing yet keeps the committed blobs in sync when someone edits the sources — a PR that changes `mock_src/uip.py` without repacking would silently ship stale mocks.

## Solution

`compile-mock-scripts.yml`: on any PR touching `_shared/mock_src/**` or `compile_mocks.py` (plus `workflow_dispatch`):

1. Check out the PR head branch (GH_PAT so the push re-triggers status checks — same pattern as `automerge-catalog-refresh.yml`).
2. Repack the blobs with CPython 3.13 — the canonical packer (`ast.unparse` output is byte-stable within a minor; the blobs themselves run on any host CPython >= 3.10).
3. If the regenerated blobs differ, commit `chore(tests): repack troubleshoot mock dispatchers` and push to the PR branch.

Loop-safe by construction: packing is deterministic, so repacking unchanged sources is byte-identical — the bot's own commit re-triggers the workflow into a clean no-op.

Fork PRs are out of scope: all contributors have write access to this repo, so PRs come from same-repo branches. (A fork PR would fail at checkout — no secrets on fork runs — which is acceptable for this repo.)

## Merge order

**Merge after #2358.** The workflow is inert until then (its `paths` filter matches nothing on main), but its verification depends on #2358's layout.

## Validation

YAML parses; the pack step is the same `compile_mocks.py` invocation validated in #2358 (byte-identical repack confirmed there). A throwaway source-touching PR can exercise the auto-push end-to-end once #2358 merges.


uipreliga:
## Code Review Summary

`.github/workflows/compile-mock-scripts.yml` (+76, only file in the PR). Three reviewers: Opus sub-agent, Gemini 3 Pro, GPT-5.3-Codex. Reviewed against `origin/main` (which already carries `mock_src/`, `compile_mocks.py`, and the `.uip.bin` / `.seal.bin` blobs).

### Critical / High

**1. The fork guard at L63–70 is unreachable — flagged by all three.**
`actions/checkout` at L41 gets `ref:` (a bare branch name) with **no `repository:`**, so it resolves that name against the *base* repo. For a fork PR:

- Branch absent from base (normal case) → step 1 dies with `fatal: couldn't find remote ref …`. The actionable message at L66–67 never prints. Compounding: `secrets.GH_PAT` is empty for `pull_request` runs from forks, so `token: ''` is passed.
- Branch name collides with a base branch (forks often branch off `main`, so `head.ref == "main"`) → checkout **silently succeeds on base `main`**, repacks base sources, finds no drift, prints "in sync", exits **0 green**, having never looked at the contributor's code.

Also, `HEAD_REPO != GITHUB_REPOSITORY` (L65) is only ever true for a fork PR, and on `workflow_dispatch` it degrades to `github.repository != github.repository` — always false. Lines 63–70 are dead in every reachable scenario.

> One reviewer proposed fixing this with `repository: head.repo.full_name` + `head.sha` while keeping the PAT. **Don't apply that as written** — checking out fork-controlled code with a PAT in scope is the classic `pwn_request` shape. The safe direction is the opposite; see the closing note.

**2. `git diff --quiet` at L58 cannot see a newly created blob — flagged by all three.**
`git diff` compares index vs. worktree for *tracked* paths; untracked files are invisible (reproduced in a scratch repo). Failure scenario: contributor adds `mock_src/orch.py` and forgets to pack. Workflow fires, `compile_mocks.py` writes a brand-new untracked `.orch.bin`, `git diff --quiet` returns 0, prints *"Committed blobs are in sync"*, exits **0 green**. PR merges with a source but no blob; at test time the loader hits `sys.exit("orch: runtime data missing")`.

Fix — catches adds, modifies and deletes uniformly:

```bash
git add -A -- tests/tasks/uipath-troubleshoot/_shared/mock_template/m/
if git diff --cached --quiet -- tests/tasks/uipath-troubleshoot/_shared/mock_template/m/; then
  echo "Committed blobs are in sync with mock_src/ — nothing to do."; exit 0
fi
```

### Medium

**3. `workflow_dispatch` on the default branch pushes an unreviewed commit straight to `main`** (L25, L41, L54, L76). One reviewer rated this High. The fork guard is inert on dispatch, `github.ref_name` is `main`, so the job does `git push origin HEAD:main` as the PAT owner. `automerge-catalog-refresh.yml:70-84` shows `main` carries a required-status-checks ruleset — if that PAT is a bypass actor, this is an unreviewed, un-CI'd push to the default branch triggerable by anyone who can click "Run workflow". Gate the push step on `if: github.event_name == 'pull_request'`.

**4. PAT exposure to PR-modifiable code** (L24, L42, L49) — flagged by all three. `compile_mocks.py` is *itself in the trigger `paths:`*, and checkout defaults to `persist-credentials: true`, writing the PAT into `.git/config` as an extraheader. Step L49 then runs the PR's own copy of that script unsandboxed. Fork PRs are safe (no secrets), so this is an insider / compromised-collaborator vector — but it is new surface. Minimum mitigation: `persist-credentials: false`, and pass the PAT only on the final `git push` line.

**5. No `concurrency:` group** (L20–32). Every other PR-triggered workflow in this repo has one with `cancel-in-progress: true` (`activation-gate.yml:8`, `lint-tasks.yml:16`, `verb-gate.yml:10`, `smoke-skills.yml:3`, and four more). Two pushes 20s apart → both runs repack and commit, the loser's `git push` is rejected non-fast-forward, `set -e` aborts, spurious red X. No fetch/rebase/retry, and `fetch-depth: 1` means a rebase would need `fetch-depth: 0`.

**6. `paths:` omits `mock_template/m/**`** (L22–24) — flagged by all three. A PR that hand-edits `.uip.bin` / `.seal.bin` never runs this workflow. Those blobs are `exec()`'d in every troubleshoot sandbox and are unreviewable base64 in a diff — nothing would contradict a tampered blob. Adding the path turns this into a real tamper gate.

### Low (informational)

- **L27–28 — `permissions: contents: write` is never exercised.** Checkout and push both authenticate with the PAT; `GITHUB_TOKEN` is unused. Misleading, and silently downgraded to read on fork PRs anyway.
- **L58 — orphaned blobs.** `compile_mocks.py` only writes, never prunes `OUT_DIR`. Deleting `mock_src/seal.py` leaves `.seal.bin` committed forever. The `git add -A` fix helps only if the packer also learns to prune.
- **L5–9 — header overstates opacity.** "no readable script in a sandbox documents the mock system" — but `m/uip`, `m/seal` and `m/uip.cmd` *do* ship into sandboxes as plaintext and openly show the `.uip.bin` decode-and-exec scheme. The blob hides mock *behaviour*, not the system's existence. Suggest: "no readable script in a sandbox reveals what the mocks do".
- **L16–18 — "no push loop" holds today but has no breaker.** The argument is sound (`pull_request` path filters evaluate the whole base…head diff, so the bot's push does re-trigger and does no-op). But it rests entirely on determinism, and `python-version: "3.13"` pins the *minor* while `compile_mocks.py` only guarantees stability *within* a minor. If that ever breaks: unbounded push loop, no concurrency group, no circuit breaker.
- **Nits:** `paths:` omits the workflow's own file, so the PR introducing it can't exercise it; `github.ref` (L41) vs `github.ref_name` (L54) asymmetry reads like a typo; the 16-line header never mentions `workflow_dispatch`, the riskier half of the trigger surface; `"3.13"` uses double quotes where the rest of the repo uses `'3.13'`.

### Positive Observations

- Header comments explain *why* (PAT vs. `GITHUB_TOKEN` downstream suppression, head-ref vs. merge-ref, determinism) rather than restating YAML, and cite their precedent — the `automerge-catalog-refresh.yml:36` PAT claim checks out. Every factual claim about `compile_mocks.py` that was verified was accurate.
- Correct interpreter pin: `3.13` matches `PACKING_VERSION = (3, 13)`; a mismatch would abort loudly.
- Plain `git push`, no `--force-with-lease` — correct for pushing onto a live contributor branch.
- `set -euo pipefail`, standard bot identity, `git add` scoped to `mock_template/m/` rather than `-A` repo-wide.
- **Right call on scope:** earlier commits on this branch tried a multi-version bytecode matrix (5edef8a7e) and walked it back to a single deterministic repack (5c27d3289). 76 lines, one job, no matrix — nothing here is over-engineered.

### Overall Assessment

The design instinct is sound and the file is unusually well documented, but **both load-bearing branches fail in the same direction — reporting success without having verified anything**: the fork guard can never execute, and `git diff --quiet` is structurally blind to the single most likely contributor mistake (new source, unpacked blob).

Worth weighing a KISS counter-proposal: a **verify-only job** (no PAT, `contents: read`, default merge-ref checkout, fails with the copy-pasteable `uv run --python 3.13 …` recipe) works identically for forks and same-repo PRs, can safely be made a required check, and deletes findings 1, 3, 4 and most of 5 outright. Keep push-back only if the convenience is worth a second, same-repo-gated job.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

uipreliga (approved):
Fix what you agree with and 🚢

PR #2363, opened 2026-07-30
fix(uipath-maestro-bpmn): align project scaffolding with the Maestro CLI

DevMomo:
## Problem

The skill taught a hand-invented project scaffold instead of the one `uip maestro bpmn init` generates, so authored projects diverged from what `pack` consumes:

- root process forced `isExecutable="true"`; the initializer omits the attribute
- guidance implied `project.uiproj` owns the runtime `main` path, when `operate.json` does
- no stated boundary for who owns what, so agents hand-wrote registry-owned payloads and expected the registry to emit sequence flows and diagrams it never emits

A second drift sat next to it: the authoring path moved to `refresh`, but ship, operate, and diagnose still led with `update-metadata`, the command this same skill documents as deprecated and unable to materialize `Intsvc.*` connection bindings.

**Why it's in the stack:** #2369's checker is built on this PR's scaffold assertions.

## Fix

- `init` is the starting point: edit at the returned `Data.Path`, preserve the six generated files
- Ownership boundary, stated once: registry owns node payloads; spec/canvas owns structural BPMN and scaffold extensions; CLI owns packaging
- `isExecutable`: preserve the initializer's shape, never force `"true"`
- `migrationVersion` is an integer (the reader `parseInt`s, so `11.5` became `11`); preserve an existing value byte-for-byte when editing
- `update-metadata` to `refresh` finished across `operate/ship.md`, `operate/CAPABILITY.md`, `diagnose/failure-modes.md`, and `shared/local-metadata-regeneration-guide.md`
- `refresh` is no longer gated on packaging intent — it is part of authoring, because `validate` cross-checks the generated metadata against the source (see Verification)

The rename had stalled on `refresh` having no `--dry-run`. It does not need one: `Data.WrittenFiles` names the files that were stale and `Data.UnchangedFiles` the ones already current. Two behaviors change with the swap, both now documented:

- `refresh` requires `project.uiproj` (`BpmnRefreshFailed` / `RetryWillNotFix`), so a bare `.bpmn` needs the two-key bootstrap first
- a start event with no `uipath:entryPointId` fails the whole project, where `update-metadata` wrote an empty `entryPoints` array

## Verification

**Evals:** ran `skill-bpmn-simple-approval-bpmn` locally three times across this
PR's trees. All three passed 3/3 at weighted score 1.000 on `claude-sonnet-5`,
taking 37, 60, and 80 turns. That spread is why `max_turns` moved 90 -> 120: at
80 turns the old cap had 10 to spare, so a run on the high side would have
failed on budget rather than behavior.

CI's `Run skill smoke tests` job also executes four of the eval tasks this PR
changes -- `author_validate`, `timer_start`, and both HITL smokes -- and passes.

**Workflow, executed rather than read (uip 1.203.0):** followed the skill's own
steps: `init`, author from the doc skeleton, `validate`. It failed --
`entry-points.json references start event "Event_start" ... but no
<bpmn:startEvent id="Event_start"> exists`. The skeleton names its start event
`Start_1`; the initializer writes `Event_start`. `refresh` resolves it, then
`validate` returns `Valid` and `pack` produces a `.nupkg`. `refresh` had been
gated on packaging intent, which a source-only draft never satisfies, so step 4
broke on the path step 3 prescribes. It is now unconditional.

**CLI claims, each checked against real output:**

| Claim | Result |
|---|---|
| `init` writes exactly the six documented files | pass |
| `init` omits `isExecutable`; `operate.json.main` = `/content/<file>.bpmn#<startId>` | pass |
| `refresh` without `project.uiproj` fails `Required file is missing` / `RetryWillNotFix` | pass, exact string |
| two-key `project.uiproj` bootstraps a bare `.bpmn`; refresh writes the other four | pass |
| `refresh` reports drift as `WrittenFiles` / `UnchangedFiles`, so no `--dry-run` is needed | pass |
| `uipath:output` requires an `elementId` naming a root end event | pass -- missing, start-event, and unknown ids all fail identically |
| `uipath:input` mismatch is dropped silently | pass |
| missing `uipath:entryPointId` fails the project, with two different messages | pass |
| `migrationVersion` reader is `Number.parseInt` | pass -- `PO.Frontend/src/services/serialization/bpmn-from-xml.ts:635` |
| current top migration is 20, not 15 | pass -- the doc no longer teaches a number to keep current |

`tests/scripts/test_bpmn_project_scaffold_contract.py` and
`test_bpmn_doc_examples_parse.py` pin the doc skeletons; both were verified
against the real defects they exist to catch, including a namespace declaration
removed from one skeleton but not the other. Link check (6715 links),
`check-skill-status.py`, and `check-skills-sh.py` are clean.

**Not verified:** `isExecutable="true"` was only tested against the CLI
validator, which accepts it. Studio Web import was not tested.


DevMomo:
Current scope: this PR is the bottom, scaffold-only layer. Earlier automated review comments refer to a pre-split combined diff and are outdated. Public process I/O is #2369, ScriptTask v3 is #2366, and generic resource resolution is #2370. The PR body and current checks are authoritative.

skills/uipath-maestro-bpmn/references/shared/local-metadata-regeneration-guide.md
```diff
@@ -8,22 +8,33 @@ Use this guide when BPMN source changed and local package metadata must be refre
...
-- `project.uiproj` must use lowercase `"main"` pointing at the BPMN file.
-- `operate.json` must use `"main"` with the bare BPMN filename, not a
-  `/content/<file>.bpmn#<start-event-id>` entry-point path, plus
-  `"contentType": "ProcessOrchestration"`.
-- `package-descriptor.json` must use a top-level `"content"` array with
-  `content/<file>` entries. Do not use `contentFiles` or a CLI scaffold
-  `"files"` mapping for synthetic local metadata.
+```bash
+uip maestro bpmn init <ProjectName> \
+  --skip-solution-registration \
```
DevMomo: Don't skip. If already part of a solution, use existing. Else, create a new solution + register.

skills/uipath-maestro-bpmn/references/shared/project-layout.md
```diff
@@ -59,13 +59,20 @@ For the regeneration and drift-check contract, see [local-metadata-regeneration-
...
+`operate.json.main` (`/content/<project>.bpmn#<start-event-id>`), and
+`package-descriptor.json` owns the generated file map. The CLI scaffold omits
+`bpmn:process@isExecutable`; Studio may round-trip the equivalent default as
+`isExecutable="false"`.
+
+If initialization or metadata regeneration is unavailable, a source-only local
+project may contain the BPMN plus that minimal `project.uiproj`, but it is not
+package-ready. Report packaging or operation as blocked instead of inventing
+`operate.json`, `entry-points.json`, or `package-descriptor.json`. See
+[local-metadata-regeneration-guide.md](local-metadata-regeneration-guide.md).
```
DevMomo: doesn't bpmn init take care of this already? may not be needed.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -68,12 +68,14 @@ file will fail to parse. Never paste CLI commands or flags
...
 
-This is the whole shape — variables, an entry point, one node, a branch, and
-the diagram — in one valid file. Author from this skeleton plus the registry
-templates for your nodes. **Do not reverse-engineer the pattern from full
-example BPMN files** — it is the main reason authoring runs out of time. Swap
-the `scriptTask` payload for the registry `xmlTemplate` of whatever node you
-need.
+This is a minimal CLI/Studio-compatible authoring scaffold with a stable manual
+entry point, one structural task, and complete diagram interchange. The CLI
+initializer omits `isExecutable`; Studio may serialize the equivalent default
```
DevMomo: Don't mention Studio Web. Coding agent should act independently.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -140,12 +124,25 @@ need.
...
+schema bodies are JSON text or CDATA.
+
+```xml
+<uipath:variables version="v1">
+  <uipath:input id="input_ExpenseId" name="expenseId" type="string" elementId="Start_1" />
+  <uipath:inputOutput id="Var_Decision" name="decision" type="string" />
+</uipath:variables>
+```
+
+If a Studio migration marker is required, its serializer-owned shape is
```
DevMomo: Fine, but don't mention Studio

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -113,18 +118,23 @@ For registry-evidence-only tasks, be command-first and time-boxed:
...
    gateways, events, boundary events, containers, multi-instance markers,
    expression/error mappings, retry attributes), then generate one
-   `BPMNShape`/`BPMNEdge` per node and flow. For local authoring prompts, use the
-   plain project layout `<ProjectName>/<ProjectName>.bpmn` with
-   `<ProjectName>/project.uiproj`; do not create `*Solution/`, package files, or
-   `.uipx` artifacts unless the user explicitly asks to package or operate the
-   project.
+   `BPMNShape`/`BPMNEdge` per node and flow. For a new bare local project,
+   initialize the supported scaffold with
+   `uip maestro bpmn init <ProjectName> --skip-solution-registration --output json`
```
DevMomo: similar to earlier comment, no need to skip?

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -113,18 +118,23 @@ For registry-evidence-only tasks, be command-first and time-boxed:
...
+   and edit the project at the returned `Data.Path`. Preserve its generated
+   metadata. `project.uiproj` owns the exact-cased `"Name"` and
+   `"ProjectType": "ProcessOrchestration"` fields; the runnable BPMN/start-event
+   path belongs in lowercase `operate.json.main`, not `project.uiproj.main`.
+   If the initializer is unavailable, create only the source pair
+   `<ProjectName>/<ProjectName>.bpmn` and `<ProjectName>/project.uiproj`, then
+   report packaging/operation metadata as blocked rather than inventing it.
```
DevMomo: if init isn't there, the whole thing will fail. no need to defend against that, no point

rockymadden (approved):
With the tweaks we called all, looks okay

skills/uipath-maestro-bpmn/references/shared/local-metadata-regeneration-guide.md
DevMomo: Addressed in 9edc00739. The command no longer skips solution registration. The guidance now says to reuse the current solution when one is in scope; otherwise init creates and registers one.

skills/uipath-maestro-bpmn/references/shared/project-layout.md
DevMomo: Addressed in 9edc00739. The duplicated field-by-field scaffold and source-only fallback were removed. This section now directs the agent to init and preserve its generated files.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Addressed in 9edc00739. The UI-specific wording is removed; the rule now describes only the CLI initializer and preservation of existing source.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Addressed in 9edc00739. The migration-marker guidance is now product-contract wording with no UI attribution.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Addressed in 9edc00739. Removed the skip flag and aligned this rule with normal solution reuse or creation and registration.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Addressed in 9edc00739. The initializer-unavailable fallback was removed instead of defending an unsupported path.

DevMomo:
@rockymadden All six requested simplifications are now on the current head, with focused tests passing and CI green except the still-running smoke job. Could you re-approve this bottom dogfooding layer so I can merge it as soon as the job finishes?

DevMomo:
@jiyangzh @nikhil-maryala This is the bottom layer of today’s dogfooding Skills cut. Rocky’s six review points are addressed and focused/CI contract tests pass; only the smoke job is finishing. Could one of you give the required code-owner approval ASAP for the atomic #2363 → #2369 merge?

DevMomo:
@uipreliga @bai-uipath @akshaylive Correction: this PR also updates `.github/workflows/test-helpers.yml`, so one of you is the required code owner for the last merge gate. The workflow change only registers the focused scaffold contract guard, which is passing. Could one of you approve ASAP for today’s dogfooding cut?

nikhil-maryala (approved):

DevMomo:
Post-rebase smoke exposed a stale test assumption, not a reason to restore `--skip-solution-registration`. I kept the reviewed plain-`init` guidance and removed redundant hardcoded-path criteria from the two HITL smoke tasks. Their recursive structural checkers already prove that the named BPMN exists, uses `Actions.HITL`, and satisfies the full wiring/routing behavior. #2369 has been restacked on this fix; focused contracts and diff checks pass.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
```diff
@@ -2,14 +2,15 @@ task_id: skill-bpmn-simple-approval-bpmn
 description: >
   Authoring eval: agent composes a model-authored simple approval BPMN that
   combines StartAgentJob, an exclusive gateway, CreateQueueItem, a script task,
-  declared root variables, numeric migration metadata, and BPMN DI. Mirrors the
-  manual smoke "simple approval" scenario.
-tags: [uipath-maestro-bpmn, integration, "mode:build", lifecycle:generate, shape:multi-node, node:gateway, node:script, node:service-task, resource]
+  declared root variables, the supported local project scaffold, numeric
+  migration metadata, and BPMN DI. Mirrors the manual smoke "simple approval"
```
rockymadden, line 6: **MAJOR** — the eval still teaches the value this PR fixes.

Line 34 of this prompt: `` (for example `version` `11` or `11.5`) ``. And `check_simple_approval_bpmn.py:171` repeats it in the failure message, while :169 accepts any string containing a digit.

Meanwhile `test_bpmn_project_scaffold_contract.py` now asserts docs must show `version="\d+"`. So the doc is pinned to integers and the eval trains agents on the decimal that silently truncates to 11. Self-contradictory inside one PR.

Fix: prompt -> "an integer `uipath:migrationVersion`"; checker -> assert `^\d+$` and drop `11.5` from the message.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -172,19 +176,20 @@ For registry-evidence-only tasks, be command-first and time-boxed:
...
    gateways, events, boundary events, containers, multi-instance markers,
    expression/error mappings, retry attributes), then run
-   `uip maestro bpmn format <file.bpmn>` to generate the diagram. If `format` reports `unknown command`, update the CLI (see [references/cli-conventions.md](references/cli-conventions.md)); if upgrading is unavailable, use the fallback DI structure in [references/structural-bpmn.md](references/structural-bpmn.md). For local authoring prompts, use the
-   plain project layout `<ProjectName>/<ProjectName>.bpmn` with
-   `<ProjectName>/project.uiproj`; do not create `*Solution/`, package files, or
-   `.uipx` artifacts unless the user explicitly asks to package or operate the
-   project.
+   `uip maestro bpmn format <file.bpmn>` to generate the diagram. If `format` reports `unknown command`, update the CLI (see [references/cli-conventions.md](references/cli-conventions.md)); if upgrading is unavailable, use the fallback DI structure in [references/structural-bpmn.md](references/structural-bpmn.md). For a new local project, initialize the
+   supported scaffold with `uip maestro bpmn init <ProjectName> --output json`,
```
rockymadden, line 188: **MAJOR** — no opt-out survives in SKILL.md.

This hunk removed:
```
do not create `*Solution/`, package files, or `.uipx` artifacts unless the
user explicitly asks to package or operate the project.
```
and replaced it with unconditional `init`. Verified output for "author me a bpmn":
```
ExpenseApprovalBpmnSolution/ExpenseApprovalBpmnSolution.uipx
ExpenseApprovalBpmnSolution/resources/solution_folder/package/...json
ExpenseApprovalBpmnSolution/resources/solution_folder/process/...json
```
0ad1000b restored `--skip-solution-registration` into `project-layout.md` — a reference loaded on demand. SKILL.md always loads and has no mention of it. Keeping plain `init` as the default and documenting the opt-out are two separate decisions.

One clause here closes it: "for a local-only project the user hasn't asked to package or operate, add `--skip-solution-registration`."

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -214,8 +219,9 @@ For registry-evidence-only tasks, be command-first and time-boxed:
    run `uip maestro bpmn update-metadata <file.bpmn>` to generate the five
```
rockymadden: **MINOR** — recommends a deprecated verb. 1.203.0 prints:

```
[WARN] `uip maestro bpmn update-metadata` is deprecated; use
`uip maestro bpmn refresh` instead
```

`local-metadata-regeneration-guide.md` says so explicitly. This line and `project-layout.md:80` still name `update-metadata` as the producer of the pack shape.

skills/uipath-maestro-bpmn/references/shared/project-layout.md
```diff
@@ -60,13 +72,16 @@ For the regeneration and drift-check contract, see [local-metadata-regeneration-
...
 
 A synthetic local project authored without a CLI generator must still match the
-executable and metadata contract before packing: the BPMN root process includes
-`isExecutable="true"`, each root start event carries
+metadata contract before packing: the root process follows the initializer's
+shape (`isExecutable` omitted, or the equivalent `"false"` — never forced to
+`"true"`), each root start event carries
 `<uipath:entryPointId value="<uuid>" />`, `operate.json` has `"main"` plus
 `"contentType": "ProcessOrchestration"`, and `package-descriptor.json` maps the
 BPMN file and generated JSON. `uip maestro bpmn update-metadata <file.bpmn>`
```
rockymadden: **MINOR** — same deprecated verb as `SKILL.md:219`, and this file's sibling guide already documents that `refresh` replaced it.

.github/workflows/test-helpers.yml
```diff
@@ -139,6 +143,25 @@ jobs:
       - name: Run contract guard
         run: pytest tests/scripts/test_runtime_payload_key_casing.py -v
 
+  # One job for every BPMN contract guard. Globbed on purpose: a PR adding a
+  # guard drops in a `tests/scripts/test_bpmn_*_contract.py` file and touches
+  # no workflow, so sibling PRs stop colliding on this file.
+  bpmn-contract-guards:
+    runs-on: ubuntu-latest
```
rockymadden: **MINOR** — wrong runner. All 12 sibling jobs in this file use `uipath-ubuntu-latest` (25, 41, 57, 78, 96, 112, 128, 166, ...). Looks like a copy slip.

.github/workflows/test-helpers.yml
```diff
@@ -139,6 +143,25 @@ jobs:
...
+
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065  # v5.6.0
+        with:
+          python-version: '3.13'
+
+      - name: Install pytest
+        run: pip install pytest
+
+      - name: Run BPMN contract guards
+        run: pytest tests/scripts/test_bpmn_*_contract.py -v
```
rockymadden: **MINOR** — latent red job. When the glob matches nothing (a future rename or removal), the shell passes the literal pattern through and pytest exits 4. `pytest tests/scripts -k "bpmn_ and contract"` keeps the "drop in a file, touch no workflow" property without that failure mode.

.github/workflows/test-helpers.yml
```diff
@@ -11,6 +11,10 @@ on:
       - 'tests/tasks/uipath-agents/**'
       - 'tests/tasks/uipath-admin/**'
       - 'tests/scripts/**'
+      - 'skills/uipath-maestro-bpmn/SKILL.md'
```
rockymadden: **MINOR** — 3 of these 4 paths gate nothing. The guard only reads `structural-bpmn.md`; `SKILL.md`, `project-layout.md`, and `local-metadata-regeneration-guide.md` add no coverage, and a match here runs every job in the workflow.

tests/scripts/test_bpmn_project_scaffold_contract.py
```diff
@@ -0,0 +1,110 @@
...
+
+    root = _minimal_example()
+    process = root.find("bpmn:process", NS)
+    assert process is not None
+
+    node_ids = {
+        element.attrib["id"]
+        for element in process
+        if element.tag.rsplit("}", 1)[-1]
+        in ("startEvent", "task", "endEvent", "exclusiveGateway")
```
rockymadden: **MINOR** — this whitelist misreports its own failures. Add a `scriptTask` to the doc example and the set-equality below fails as a DI-coverage mismatch when the real cause is this tuple.

Derive node ids from "has an `id`, isn't `sequenceFlow`/`extensionElements`" and the test stays true to its docstring.

tests/scripts/test_bpmn_project_scaffold_contract.py
```diff
@@ -0,0 +1,110 @@
...
+    flow_ids = {
+        flow.attrib["id"] for flow in process.findall("bpmn:sequenceFlow", NS)
+    }
+    shapes = root.findall(".//bpmndi:BPMNShape", NS)
+    edges = root.findall(".//bpmndi:BPMNEdge", NS)
+
+    assert {shape.attrib["bpmnElement"] for shape in shapes} == node_ids
+    assert {edge.attrib["bpmnElement"] for edge in edges} == flow_ids
+    assert all(len(edge.findall("di:waypoint", NS)) >= 2 for edge in edges)
+def test_variable_and_migration_examples_use_serializer_attributes() -> None:
```
rockymadden, line 97: **NIT** — no blank lines before this `def`. Parses fine and the repo has no Python linter, but it's the only such spot in the file.

Also: `variables_section` runs from the Variables heading to EOF, so the `migrationVersion` regex passes as long as one integer example exists anywhere after it — a decimal added further down wouldn't be caught.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -515,9 +514,9 @@ Safe, surgical edits on an existing `.bpmn` (preserve content you did not author
 - **Move logic into a subprocess**: move only elements that share a valid scope,
   re-scope their variables, recreate legal subprocess flow boundaries, and add a
   second diagram plane for the subprocess so nested content renders.
-- **Add an entry point**: use a root-level start event, add a stable unique
-  `uipath:entryPointId`, and declare input/output variables whose `elementId`
-  matches that start event.
+- **Add an entry point**: use a root-level start event and generate a stable,
```
rockymadden, line 526: **MINOR** — this bullet traded a behavior-critical rule for an anti-copy note. Deleted: *"declare input/output variables whose `elementId` matches that start event."*

That rule is load-bearing. Verified via `refresh` on 1.203.0:

| root declaration | `entryPoints[0].input` |
|---|---|
| `<uipath:input id="expenseId" elementId="Event_start">` | `{"expenseId": {"type":"string"}}` |
| same, `elementId` removed | `{}` |

The surviving aside in the Variables section ("Public or node-scoped declarations also carry the owning node's `elementId`") doesn't say "match the start event", and "Public" isn't defined until #2369.

#2369 restores it properly — I confirmed `uipath:output` + `elementId=<root EndEvent id>` populates the output schema. Fine if the stack merges atomically; worth one line in the body confirming #2363 can't land alone.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -85,52 +87,34 @@ need.
     xmlns:uipath="http://uipath.org/schema/bpmn"
     id="Definitions_1" targetNamespace="http://bpmn.io/schema/bpmn"
     exporter="UiPath (https://bpmn.uipath.com)" exporterVersion="1.0">
```
rockymadden: **MINOR** — last unaligned attribute pair in a PR about alignment. `init` writes:

```
exporter="UiPath Maestro (https://uipath.com)" exporterVersion="1.203.0"
```

Same on line 44. (The scaffold also names the process `Maestro Process` and the start event `core.trigger.manual`, if you want the skeleton to be literally what `init` produces.)

tests/tasks/uipath-maestro-bpmn/hitl/smoke_completed_wired/smoke_completed_wired.yaml
```diff
@@ -33,22 +33,8 @@ initial_prompt: |
...
-    description: "HITL modeled with the documented Actions.HITL activity type"
-    path: "PurchaseApprovalBpmn/PurchaseApprovalBpmn.bpmn"
-    includes:
-      - "Actions.HITL"
-    weight: 2.0
-    pass_threshold: 1.0
-
   - type: run_command
-    description: "HITL completion is wired to a downstream step that reaches an end event"
+    description: "An Actions.HITL BPMN exists and its completion is wired to a downstream step that reaches an end event"
```
rockymadden, line 45: **MINOR** — now a single all-or-nothing criterion at `weight: 5.0`. Coverage is preserved (the checker does assert `Actions.HITL`), but "authored nothing" and "authored HITL, mis-wired" both score 0, so a regression here reads the same as a total failure.

`.claude/rules/test-writing.md:74` also reserves 5.0-6.0 for e2e execution; `timer_start.yaml` went 4.0 -> 6.0 for an authoring smoke. Same applies to `smoke_multi_outcome_routing.yaml`.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
```diff
@@ -2,14 +2,15 @@ task_id: skill-bpmn-simple-approval-bpmn
...
+  scenario.
+tags: [uipath-maestro-bpmn, integration, "mode:build", "lifecycle:generate", "shape:multi-node", "node:decision", resource]
 
 run_limits:
   task_timeout: 2400
   turn_timeout: 1800
-  expected_turns: 24
+  expected_turns: 55
```
rockymadden: **MINOR / question** — 24 -> 55 is 61% of `max_turns: 90`. Measured on the init flow, or padded? If the scaffold path genuinely doubles turns, that's a per-run cost worth recording somewhere.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
```diff
@@ -67,6 +82,82 @@ def main() -> None:
     process = root.find("bpmn:process", NS)
     if process is None:
         fail("missing bpmn:process element")
+    # `isExecutable` is deliberately not graded: nothing in the CLI reads it
```
rockymadden: **NIT** — 5 lines explaining why something *isn't* graded. The claim is correct (I confirmed `validate` and `update-metadata --dry-run` ignore `isExecutable`), but an absence doesn't need an essay in the file that omits it. One line, or nothing.

Same read on the `test_minimal_example_has_complete_di_coverage` docstring: "Deliberately narrow... does not re-assert the example's graph wiring" narrates the authoring decision, not the test.

rockymadden (approved):
Approving. I ran `uip maestro bpmn init` on 1.203.0 and checked every scaffold claim in this PR against real output — they all hold.

**Verified**
- `init` writes exactly the 6 documented files; both trees (solution wrapper and `--skip-solution-registration`) match the docs
- `isExecutable` omitted by the initializer, and inert: `validate` returns `Valid` with the attribute absent *and* `"true"`; `update-metadata --dry-run` reports `Drifts: []` either way
- `operate.json.main` = `/content/<file>.bpmn#<startId>`; `package-descriptor.json` files map matches the checker's required subset exactly
- DI guard rationale still true on 1.203.0: 0x0 bounds + single-waypoint edges report `Valid`
- `init` needs no auth (clean `HOME`), so the evals can rely on it
- HITL yaml criteria removal loses no coverage — the checkers do assert `Actions.HITL` and discover the file by glob
- Tag fix to `node:decision` matches the vocabulary in `tests/README.md:105`; verb gate 0 High / 0 Medium; new guard 3 passed

Two things I'd like fixed in a follow-up commit before merge, both small — see inline: the eval still teaches `migrationVersion 11.5` (the exact value this PR proves truncates), and SKILL.md has no `--skip-solution-registration` after losing the "don't create `*Solution/`" guardrail, so plain `init` now emits a `.uipx` plus `resources/solution_folder/**` for anyone who asked for a `.bpmn`.

Rest is minor. Note my previous approval was on 79bb543e — 4 commits landed after it, and both major findings come from those.

Slop check: commit bodies are specific and verifiable. Two comments narrate the authoring decision rather than serving a future reader (flagged inline).

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
DevMomo: Fixed. Prompt now shows `version="15"` as an example and the checker accepts any integer.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Restored: `--skip-solution-registration` for a source-only draft the user has not asked to package or operate.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Fixed — `refresh <project-path>`.

skills/uipath-maestro-bpmn/references/shared/project-layout.md
DevMomo: Fixed — same change here.

.github/workflows/test-helpers.yml
DevMomo: Fixed — `uipath-ubuntu-latest`.

.github/workflows/test-helpers.yml
DevMomo: Now an explicit file path. A glob silently matches zero; an explicit path fails loudly if the file moves.

.github/workflows/test-helpers.yml
DevMomo: Dropped the 3 that gate nothing. Only `structural-bpmn.md` remains.

tests/scripts/test_bpmn_project_scaffold_contract.py
DevMomo: Fixed — the node set is derived from the document (all ids except `sequenceFlow`/`extensionElements`), so a new node type surfaces as missing DI, not as a whitelist mismatch.

tests/scripts/test_bpmn_project_scaffold_contract.py
DevMomo: Added.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Restored, and kept the anti-copy note alongside it. Your `refresh` evidence is in the commit message.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Fixed — both examples now `exporter="UiPath Maestro (https://uipath.com)" exporterVersion="1.203.0"`.

tests/tasks/uipath-maestro-bpmn/hitl/smoke_completed_wired/smoke_completed_wired.yaml
DevMomo: Split back into `1.0` (BPMN authored, layout-agnostic via `rglob`) + `3.0` (checker). Folding the removed criterion's weight into the survivor collapsed two distinguishable failures into one 0.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
DevMomo: Measured — 55 is the observed turn count on the init flow, not padding. Left as is.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
DevMomo: Cut to one line.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -211,11 +219,12 @@ For registry-evidence-only tasks, be command-first and time-boxed:
    wording such as "connection and process binding" because it hides the concrete
    artifact the CLI must supply.
    If the user asks for the package metadata files, or to package or operate,
-   run `uip maestro bpmn update-metadata <file.bpmn>` to generate the five
+   run `uip maestro bpmn refresh <project-path>` to generate the four
```
rockymadden: 🔴 **Major — this regresses the bare-`.bpmn` path.**

`refresh` hard-requires an existing `project.uiproj`. Verified on 1.203.0 against a directory holding only a `.bpmn`:

```
"Code": "BpmnRefreshFailed",
"Instructions": "Required file is missing: .../Proj/project.uiproj",
"Retry": "RetryWillNotFix"
```

The command this replaced, `update-metadata <file.bpmn>`, bootstraps from a bare `.bpmn` and writes all five files including `project.uiproj`. This repo's own fixtures are exactly that shape — `tests/tasks/uipath-maestro-bpmn/edit/*/fixture/*.bpmn` and `diagnose/validate_fix_loop/fixture/ExpenseApproval.bpmn` have no `project.uiproj`. An agent asked to package one of those now gets `RetryWillNotFix` with no documented recovery: the only fallback here is gated on "CLI is unavailable", and the CLI is available, it just refuses.

**Fix:** state the precondition and the bootstrap. A two-key `project.uiproj` is enough — verified `refresh` then writes all four:

```json
{ "Name": "Proj", "ProjectType": "ProcessOrchestration" }
```

Suggested: "`refresh` requires an init-generated project — it fails if `project.uiproj` is absent. For a bare `.bpmn`, write the two-key `project.uiproj` first (see project-layout.md), then refresh."

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -172,19 +176,23 @@ For registry-evidence-only tasks, be command-first and time-boxed:
...
-   `.uipx` artifacts unless the user explicitly asks to package or operate the
-   project.
+   `uip maestro bpmn format <file.bpmn>` to generate the diagram. If `format` reports `unknown command`, update the CLI (see [references/cli-conventions.md](references/cli-conventions.md)); if upgrading is unavailable, use the fallback DI structure in [references/structural-bpmn.md](references/structural-bpmn.md). For a new local project, initialize the
+   supported scaffold with `uip maestro bpmn init <ProjectName> --output json`,
+   edit at the returned `Data.Path`, and preserve its generated metadata. For a
+   source-only draft the user has not asked to package or operate, pass
+   `--skip-solution-registration` so no `*Solution/` wrapper or `.uipx` is
+   created. See
+   [references/shared/local-metadata-regeneration-guide.md](references/shared/local-metadata-regeneration-guide.md).
+   The runnable BPMN/start-event path belongs in lowercase `operate.json.main`,
```
rockymadden, line 194: 🔴 **Major — nothing tells the agent to re-run `refresh` after moving the entry point, and the new eval grades exactly that.**

`init` emits `id="Event_start"` and `operate.json.main = /content/<file>.bpmn#Event_start`. Edit the start event id and the generated metadata silently goes stale — verified:

```
# renamed Event_start -> Start_1 in the .bpmn, no refresh
operate.json: "main": "/content/ExpenseApprovalBpmn.bpmn#Event_start"   # stale
# after refresh
operate.json: "main": "/content/ExpenseApprovalBpmn.bpmn#Start_1"      # correct
```

`check_simple_approval_bpmn.py` derives `expected_main` from the BPMN's start id and fails the weight-5.0 criterion on mismatch. Two things make that rename likely rather than hypothetical:

- the minimal skeleton in `structural-bpmn.md` uses `Start_1`, not `Event_start`, so the doc encourages the rename
- `refresh` is gated in step 3 on "if the user asks for the package metadata files, or to package or operate", and the eval prompt explicitly says *do not* package or run

So an agent that follows the skill correctly can still fail the eval. This is a real runtime contract, not just eval hygiene — a stale `operate.json.main` points the runtime at an entry point that no longer exists.

**Fix:** add to step 3 — "after changing a start event's id or adding/removing an entry point, re-run `uip maestro bpmn refresh <project-path>`; `operate.json.main` and `entry-points.json` do not follow source edits."

tests/tasks/uipath-maestro-bpmn/smoke/author_validate.yaml
```diff
@@ -19,7 +19,7 @@ initial_prompt: |
   Load and follow the uipath-maestro-bpmn skill.
 
   Author a local Maestro BPMN project with main file
-  `InvoiceApproval/InvoiceApproval.bpmn` for this process:
+  `InvoiceApprovalSolution/InvoiceApproval/InvoiceApproval.bpmn` for this process:
```
rockymadden: 🔴 **Major — this eval now contradicts the skill it exercises.**

The prompt ends with "Do not upload, publish, deploy, debug, run, or **pack** anything" — a source-only draft. SKILL.md:191 (this PR) says for exactly that case: "pass `--skip-solution-registration` so no `*Solution/` wrapper or `.uipx` is created." That lands the file at `InvoiceApproval/InvoiceApproval.bpmn`, so an agent following the skill fails both `file_exists` and `file_contains` (2.0 of 7.5 weight).

Only an agent that ignores the skill's guidance and takes the literal path from the prompt passes — which turns a skill-following eval into a prompt-following eval.

It is also inconsistent with the rest of this PR: `timer_start`, both HITL smokes, and `simple_approval` were all deliberately made layout-agnostic here. This one went the other way and hardcoded the CLI's solution layout into the prompt *and* two criteria.

**Fix:** same treatment as the others — drop the path from the prompt ("main file named `InvoiceApproval.bpmn` inside the generated project"), and replace the two path-pinned criteria with the `rglob` form used in `simple_approval_bpmn.yaml`. `check_author_validate.py` already discovers the file by glob.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -84,53 +86,35 @@ need.
...
-        </uipath:mapping>
+        <uipath:entryPointId value="00000000-0000-4000-8000-000000000001" />
       </bpmn:extensionElements>
-      <bpmn:incoming>Flow_1</bpmn:incoming><bpmn:outgoing>Flow_2</bpmn:outgoing>
-      <bpmn:script><![CDATA[return { response: amount > 1000 ? "high" : "low" };]]></bpmn:script>
-    </bpmn:scriptTask>
-    <bpmn:exclusiveGateway id="Gw_1" name="Tier?" default="Flow_Low">
+      <bpmn:outgoing>Flow_1</bpmn:outgoing>
+    </bpmn:startEvent>
+    <bpmn:task id="Task_1" name="Work">
```
rockymadden, line 100: 🟡 **Minor — the skeleton's one task carries no payload.**

`registry search` binds `bpmn:Task` to a single type:

```json
{"ExtensionType": "BPMN.Variables", "BpmnElement": "bpmn:Task", "ExtensionTag": "uipath:mapping"}
```

A bare `<bpmn:task>` with no `extensionElements` is not a node the registry recognises as anything. It validates (`Status: Valid`, confirmed on 1.203.0) but has no execution semantics.

The header says "author from this, not from examples" — so this is the shape agents copy. The old skeleton showed a real `scriptTask` with a `uipath:mapping` payload, which demonstrated Critical Rule #3's two halves in one file. The replacement demonstrates neither half.

**Fix:** either attach the `BPMN.Variables` mapping the registry actually binds to `bpmn:Task`, or keep a `scriptTask` with its registry payload. Minimal is fine; semantically empty is not.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -41,8 +41,8 @@ namespace, at least one `<bpmn:process>`, and (to render) a
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns:uipath="http://uipath.org/schema/bpmn"
     id="Definitions_1" targetNamespace="http://bpmn.io/schema/bpmn"
-    exporter="UiPath (https://bpmn.uipath.com)" exporterVersion="1.0">
-  <bpmn:process id="Process_1" isExecutable="true">
+    exporter="UiPath Maestro (https://uipath.com)" exporterVersion="1.203.0">
```
rockymadden: 🟢 **Nit — don't pin the CLI version in an authoring skeleton.**

`exporterVersion="1.203.0"` matches today's `init` output exactly, so the change is accurate. But nothing reads it: in the maestro-tool bundle `exporterVersion` appears only in the bpmn-moddle schema and in the Flow writer (which hardcodes `"development"`). It is inert metadata.

Cost of pinning it: every file authored from this skeleton carries a version string that is wrong the moment the CLI moves, and the new guard only checks `\d+` on `migrationVersion`, nothing on this. Two occurrences (here and line 89).

Suggest a placeholder, or drop the attribute from the skeleton and note that `init`/`format` write it.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -140,12 +124,27 @@ need.
...
+
+```xml
+<uipath:variables version="v1">
+  <uipath:input id="input_ExpenseId" name="expenseId" type="string" elementId="Start_1" />
+  <uipath:inputOutput id="Var_Decision" name="decision" type="string" />
+</uipath:variables>
+```
+
+If a migration marker is present, its supported shape is
+`<uipath:migrationVersion version="15" />`; the attribute is `version`, not
```
rockymadden: 🟢 **Nit — `15` is correct today and will go stale silently.**

Verified both halves of this claim on 1.203.0:

- `migrationVersion = Number.parseInt(migrationVersion.version...)` — so `11.5` → `11`, as stated ✅
- the migrations table tops out at `{version: 15, description: "Repair canonicalId values that collide with variable identifiers"}`, and `generateRootNode` uses `migrations[migrations.length - 1]?.version` ✅

So `15` is the current max, and the prose correctly says the marker is optional and `init` omits it. The guard pins `version="\d+"`, not the number — good. But the eval prompt teaches `version="15"` literally, so migration 16 makes the eval teach a stale value with nothing failing.

Cheap fix: have the eval accept any integer (it already does) and drop the specific number from the prompt, or say "the current top migration number" rather than naming one.

tests/scripts/test_bpmn_project_scaffold_contract.py
```diff
@@ -0,0 +1,116 @@
...
+    assert entry_point is not None
+    UUID(entry_point.attrib["value"])
+
+
+def test_minimal_example_has_complete_di_coverage() -> None:
+    """Every node has a shape, every flow an edge, every edge two waypoints.
+
+    Deliberately narrow: this pins DI *coverage* over the canonical example
+    agents copy, which is what the CLI's `validate` still does not enforce
+    (0x0 bounds and single-waypoint edges report Valid on uip 1.202.0).
```
rockymadden: 🟢 **Nit — stale version, and this docstring narrates the authoring decision.**

`1.202.0` here vs `1.203.0` pinned in the skeleton this test guards. I re-confirmed the DI-guard rationale still holds on 1.203.0 (0x0 bounds and single-waypoint edges both report `Valid`), so the point stands, only the number is off.

The last sentence — "It does not re-assert the example's graph wiring -- that is a hand-written snippet checked into this repo, not a regression surface" — is a note to the reviewer, not to a future reader. Drop it. The first two sentences earn their place: they tell someone why not to widen this test.

(Same call as the other comments in this PR, which are pulling their weight: `# isExecutable is not graded: no CLI code path reads it` is worth keeping — I confirmed `isExecutable` appears in the bundle only as a bpmn-moddle schema entry, never read.)

rockymadden (changes requested):
Re-reviewed from scratch: the branch was force-pushed after my 2026-09-08 approval (`0ad1000b` is gone, 8 commits now, all restamped), so the prior approval no longer applies to anything.

Both items from that review are now fixed — the eval teaches an integer `migrationVersion`, and SKILL.md has `--skip-solution-registration`. Three new Majors came in with the rewrite.

---

## 📋 Summary

Replaces a hand-invented BPMN project scaffold with the one `uip maestro bpmn init` actually generates, and states an ownership boundary once: registry owns listed node payloads, spec/canvas owns structural BPMN **and** serializer scaffold metadata (`uipath:variables`, `bindings`, `entryPointId`, `migrationVersion`), CLI owns packaging.

- `isExecutable` no longer forced to `"true"` — preserve whatever the initializer wrote (it omits the attribute)
- `operate.json.main`, not `project.uiproj.main`, owns the runtime path
- `update-metadata` → `refresh` in SKILL.md and project-layout.md
- `migrationVersion` documented as an integer
- new CI guard pinning the doc skeleton; 5 eval tasks adjusted to the CLI layout

## ✅ Verified against real CLI output (1.203.0)

Every scaffold claim in the diff, checked against actual `init`/`refresh`/`validate` runs rather than read:

| Claim | Result |
|---|---|
| `init` writes exactly 6 files; both trees match project-layout.md | ✅ exact, incl. `<Name>Solution/` + `.uipx` + `resources/solution_folder/**` |
| `--skip-solution-registration` produces no wrapper or `.uipx` | ✅ bare `ProjectName/` with all 6 files |
| initializer omits `isExecutable`, and it is inert | ✅ absent from output; in the bundle it exists only as a bpmn-moddle schema entry, never read |
| `operate.json.main` = `/content/<file>.bpmn#<startId>` | ✅ `#Event_start` |
| `package-descriptor.json` files map == the checker's required subset | ✅ exact, incl. `"bindings.json": "bindings_v2.json"` |
| `project.uiproj` = `Name` + `ProjectType` only, no `main` | ✅ |
| `migrationVersion` is `parseInt`-ed, so `11.5` → `11` | ✅ `Number.parseInt(migrationVersion.version...)` |
| `version="15"` is the current top migration | ✅ migrations table ends at 15 |
| initializer omits the `migrationVersion` marker | ✅ absent |
| `refresh` writes four files, not five | ✅ `project.uiproj` untouched |
| unmatched `elementId` → empty entry-point `input` schema | ✅ `{}` without it, `{"a": {"type": "string"}}` with it |
| doc skeleton validates | ✅ `Status: Valid`, 1 process, 1 start event, 3 extensions |
| DI-guard rationale still true | ✅ 0x0 bounds + single-waypoint edges still report `Valid` |
| `init` needs no auth | ✅ clean `HOME` |
| new guard + full bpmn suite | ✅ 3 passed, 12 passed |
| link check, CLI-verb gate, REQUIRED-CHECKS parse | ✅ 6715 links resolve; 0 verb issues; table entry parses and matches the job name |
| HITL/timer criteria removal loses no coverage | ✅ the checkers do assert `Actions.HITL`, `Intsvc.TimerTrigger`, and `timerEventDefinition`, and discover files by glob |
| workflow has no `paths:` filter, so the new required job always reports | ✅ |

The scaffold alignment itself is correct. The problems are all in what the skill now tells an agent to *do* with it.

## 🔴 Comments for the author

**1. `refresh` regresses the bare-`.bpmn` path — Major** (`SKILL.md:222`)

`refresh` hard-requires `project.uiproj`; it exits `BpmnRefreshFailed` / `RetryWillNotFix` without one. `update-metadata <file.bpmn>`, which it replaced, bootstraps from a bare `.bpmn` and writes all five files. This repo's own `edit/*` and `diagnose/validate_fix_loop` fixtures are bare `.bpmn` files. Fix: document the precondition plus the two-key `project.uiproj` bootstrap (verified sufficient).

**2. Nothing says to re-run `refresh` after moving the entry point — Major** (`SKILL.md:194`)

`init` emits `Event_start`; the doc skeleton uses `Start_1`. Rename and `operate.json.main` silently keeps pointing at the old id — confirmed. `check_simple_approval_bpmn.py` grades exactly that at weight 5.0, while step 3 gates `refresh` on packaging intent and the eval prompt forbids packaging. Fix: one sentence in step 3.

**3. `author_validate.yaml` contradicts the skill it exercises — Major** (`author_validate.yaml:22`)

Prompt says don't pack (→ skill says `--skip-solution-registration` → bare `InvoiceApproval/`) but hardcodes `InvoiceApprovalSolution/InvoiceApproval/InvoiceApproval.bpmn` in the prompt and two criteria. Following the skill fails the eval. Also the odd one out — the four other tasks in this PR were made layout-agnostic.

**4. Half-finished `update-metadata` → `refresh` rename — Minor**

SKILL.md and project-layout.md moved; these still lead with the deprecated command:

- `references/operate/ship.md:22-23`
- `references/operate/CAPABILITY.md:35`
- `references/diagnose/failure-modes.md:51`
- `references/shared/local-metadata-regeneration-guide.md:23,83,114`

Defensible — `refresh` has no `--dry-run`, so the drift check only exists on `update-metadata`. But say that explicitly instead of leaving two competing commands for the same job in one skill. This is the drift class the PR exists to close.

**5. Dead alternative in the eval criteria — Minor** (`simple_approval_bpmn.yaml`)

Criterion 2 accepts `project.uiproj` **or** `ExpenseApprovalBpmn.bpmn.ts`, but the weight-5 checker now hard-fails without `project.uiproj`. The `.bpmn.ts` branch can never carry the task. Drop it.

**6. Guard checks the doc against itself — Minor**

`test_bpmn_project_scaffold_contract.py` parses the skeleton out of `structural-bpmn.md` and asserts against it. Nothing compares it to real `init` output — so the next CLI change (init starts emitting `migrationVersion`, exporter version bumps, a file is added) reintroduces exactly the drift this PR fixes, silently green. Not asking for CLI-in-CI here; dropping the pinned version string from the skeleton (comment inline) removes the cheapest source of it.

**7–9. Nits** — bare `bpmn:task` in the skeleton, pinned `exporterVersion="1.203.0"`, stale `1.202.0` in the guard docstring. See inline.

## 📈 Also worth a look

`expected_turns` 24 → 55 against an unchanged `max_turns: 90` drops headroom from 3.75x to 1.6x. That is consistent with the task now requiring a CLI `init` plus metadata consistency, but it is a real budget change on a task that already runs 2400s. Confirm the harness is fine with it. (`expected_turns` has no consumer in this repo, so nothing will tell you if it isn't.)

## 🧹 Slop check

Mostly clean. Commit bodies are specific and verifiable, and the code comments largely earn their place — `# isExecutable is not graded: no CLI code path reads it`, `# Subset, not equality: ...`, and `# Pin the contract, not a version number: ...` all stop a future editor from doing the wrong thing, and all three check out against the bundle.

One docstring still narrates the authoring decision rather than serving a reader (`test_bpmn_project_scaffold_contract.py:58-64`, flagged inline). The `bpmn-contract-guards` workflow comment is borderline but justified — it explains the deliberate anti-glob choice.

## 🎯 Recommendation

**Request changes.** The scaffold facts are right and I verified all of them. But #1 regresses a path that worked before this PR, #2 sets up an eval that a skill-following agent can fail, and #3 has an eval contradicting the guidance this same PR adds. Three small commits, not a redesign — and worth doing at the bottom of the stack rather than on top of #2369.

---

**tl;dr** — The scaffold now matches what the CLI really generates; I checked every claim against real `init` output and they all hold. But `refresh` needs a `project.uiproj` that bare `.bpmn` files don't have, nobody tells the agent to re-run it after moving the start event (which the new eval grades), and one eval hardcodes a path that contradicts the skill's own advice. Fix those three, then merge.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Fixed. Step 3 now states the precondition and the bootstrap: `refresh` needs an init-generated project, fails `Required file is missing` / `RetryWillNotFix` without `project.uiproj`, and a bare `.bpmn` gets the two-key file first. Called out that the edit fixtures are exactly that shape.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Added — re-run `refresh` after changing a start event id or adding/removing an entry point, with the reason: `operate.json.main` and `entry-points.json` are generated once and do not follow source edits, so a rename points the runtime at an entry point that no longer exists.

tests/tasks/uipath-maestro-bpmn/smoke/author_validate.yaml
DevMomo: Right, and it was the worst kind of miss — I made the other four layout-agnostic in this same PR and took this one the other way. Prompt now says the main file is named `InvoiceApproval.bpmn` inside the generated project, and both criteria use the `rglob` form. No hardcoded solution path left in the file.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Fixed — the task now carries the `BPMN.Variables` mapping the registry binds to `bpmn:Task`, so the shape agents copy has execution semantics.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Took the second option and dropped the attribute from both skeletons, with a line saying `init` and `format` write it. Worth noting my first attempt was a `<cli-version>` placeholder, which put raw angle brackets in an XML attribute and made the skeleton not well-formed — the scaffold guard caught it immediately.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Dropped the literal. The prompt now asks for the serializer's current top migration number, and the checker already accepts any integer.

tests/scripts/test_bpmn_project_scaffold_contract.py
DevMomo: 1.203.0, and the reviewer-directed sentence is gone. Kept the first two sentences and the `isExecutable` comment.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -84,53 +88,40 @@ need.
     xmlns:di="http://www.omg.org/spec/DD/20100524/DI"
     xmlns:uipath="http://uipath.org/schema/bpmn"
     id="Definitions_1" targetNamespace="http://bpmn.io/schema/bpmn"
-    exporter="UiPath (https://bpmn.uipath.com)" exporterVersion="1.0">
-  <bpmn:process id="Process_1" isExecutable="true">
+    exporter="UiPath Maestro (https://uipath.com)">
```
rockymadden, line 88: 🔴 **major** — the minimal skeleton no longer declares `xmlns:xsi`.

Same file still uses `xsi:type` at :226 (conditionExpression), :275 / :317 (timer), :408 (completionCondition). Pre-PR the skeleton carried the declaration inline on the conditionExpression — the gateway went and took it along.

Agent authors from this skeleton, adds a conditional flow per :226:

```
ET.ParseError: unbound prefix: line 35, column 6
```

`uip maestro bpmn validate` does **not** catch it — I tested, it reported only variable errors. But `_shared/bpmn_check.py:39` uses `ET.parse`, and both `timer_start.yaml` and `author_validate.yaml` instruct the agent to run `ET.parse` directly.

**Fix** — one line, the root skeleton at :41 already has it:

```xml
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -68,12 +68,16 @@ file will fail to parse. Never paste CLI commands or flags
...
-example BPMN files** — it is the main reason authoring runs out of time. Swap
-the `scriptTask` payload for the registry `xmlTemplate` of whatever node you
-need.
+This is a minimal CLI-compatible authoring scaffold with a stable manual entry
+point, one structural task, and complete diagram interchange. The CLI
+initializer omits `isExecutable`; preserve that shape. `init` and `format`
+write `exporterVersion`; leave it to them rather than pinning a release into
+hand-authored source. If existing source
+includes the equivalent default `isExecutable="false"`, preserve it. Do not
+force `isExecutable="true"`. Author from this skeleton plus the registry
```
rockymadden: 🟡 **minor** (slop) — three sentences for one rule:

- "The CLI initializer omits `isExecutable`; preserve that shape."
- "If existing source includes the equivalent default `isExecutable=\"false\"`, preserve it."
- "Do not force `isExecutable=\"true\"`."

**Fix:**

> Preserve the initializer's `isExecutable` shape — omitted, or `"false"` if already present. Never force `"true"`.

Same paragraph also has ragged mid-sentence wraps left over from the reflow.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -140,12 +131,27 @@ need.
...
+block directly. Every declaration needs a stable, unique `id`, a non-empty
+user-facing `name`, and its documented `type`; do not use the name as a
+substitute for the id. Expressions reference the id as `vars.<id>`. Public or
+node-scoped declarations also carry the owning node's `elementId`. Variable
+schema bodies are JSON text or CDATA.
+
+```xml
+<uipath:variables version="v1">
+  <uipath:input id="input_ExpenseId" name="expenseId" type="string" elementId="Start_1" />
+  <uipath:inputOutput id="Var_Decision" name="decision" type="string" />
```
rockymadden, line 140: 🟡 **minor** — the deleted root `uipath:output` guidance has no replacement anywhere in the skill (grep confirms).

It's a real mechanism. I declared:

```xml
<uipath:output id="Var_Decision" name="decision" type="string" elementId="End_1" />
```

and `refresh` produced `output: {"properties": {"decision": {"type": "string"}}}` on the entry point.

Matters directly for #2369.

**Fix** — add a third line to this example.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -140,12 +131,27 @@ need.
...
+</uipath:variables>
+```
+
+If a migration marker is present, its supported shape is
+`<uipath:migrationVersion version="15" />`; the attribute is `version`, not
+`value`, and its value is an **integer** serializer migration number — the
+reader does `Number.parseInt`, so a decimal such as `11.5` is silently
+truncated. The CLI initializer may omit that optional marker.
```
rockymadden: 🔴 **major** — this integer rule contradicts an untouched authoring eval.

`tests/tasks/uipath-maestro-bpmn/nodes/contract_variant_wrappers.yaml:52` asks the agent to author `5`, `11`, **and `11.5`**; `check_contract_variant_wrappers.py:81` requires exact `11.5`.

An agent following this line normalizes to `11` and fails that task.

**Fix** — either add "preserve an existing decimal byte-for-byte" here, or update that task in this PR.

Separately: `migrationVersion` appears zero times in `@uipath/cli/dist`, so the `Number.parseInt` claim isn't verifiable from the CLI or this repo. Cite the PO.Frontend source.

tests/tasks/uipath-maestro-bpmn/smoke/author_validate.yaml
```diff
@@ -36,31 +36,31 @@ initial_prompt: |
...
-    `python3 -c "import xml.etree.ElementTree as ET; ET.parse('InvoiceApproval/InvoiceApproval.bpmn')"`.
+    `python3 -c "import xml.etree.ElementTree as ET, pathlib; ET.parse(next(pathlib.Path('.').rglob('InvoiceApproval.bpmn')))"`.
   - Do not upload, publish, deploy, debug, run, or pack anything.
 
 success_criteria:
   - type: command_executed
     description: "Agent ran an explicit validation command on the authored BPMN"
     tool_name: "Bash"
-    command_pattern: 'python3?\s+-c\s+[\s\S]*ET\.parse\([\s\S]*InvoiceApproval\.bpmn'
+    command_pattern: 'python3?\s+-c\s+[\s\S]*InvoiceApproval\.bpmn'
```
rockymadden: 🔴 **major** — dropping `ET\.parse\(` was unnecessary and guts the criterion.

I tested both patterns against the new suggested command:

| | old pattern | new pattern |
|---|---|---|
| suggested `ET.parse(next(...rglob(...)))` | ✅ matches | ✅ matches |
| `python3 -c "print('InvoiceApproval.bpmn')"` | ❌ | ⚠️ **matches** |

The old pattern already handled the rglob rewrite. `timer_start.yaml:49` in this same PR kept `ET\.parse\(`.

**Fix:**

```yaml
command_pattern: 'python3?\s+-c\s+[\s\S]*ET\.parse\([\s\S]*InvoiceApproval\.bpmn'
```

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
```diff
@@ -2,14 +2,15 @@ task_id: skill-bpmn-simple-approval-bpmn
 description: >
   Authoring eval: agent composes a model-authored simple approval BPMN that
   combines StartAgentJob, an exclusive gateway, CreateQueueItem, a script task,
-  declared root variables, numeric migration metadata, and BPMN DI. Mirrors the
-  manual smoke "simple approval" scenario.
-tags: [uipath-maestro-bpmn, integration, "mode:build", lifecycle:generate, shape:multi-node, node:gateway, node:script, node:service-task, resource]
+  declared root variables, the supported local project scaffold, numeric
```
rockymadden: 🟡 **minor** — description still says "numeric migration metadata"; the requirement changed to integer at :33-34. One word.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
```diff
@@ -29,8 +30,8 @@ initial_prompt: |
     user names.
   - Declare root `uipath:variables version="v1"` entries for `expenseId`,
     `amount`, and `decision` with documented types.
-  - Include a numeric `uipath:migrationVersion` element (for example `version`
-    `11` or `11.5`) on the process.
+  - Include an integer `uipath:migrationVersion` element on the process, set
+    to the serializer's current top migration number.
```
rockymadden: 🟡 **minor** — "the serializer's current top migration number" is neither taught by the skill nor graded.

`check_simple_approval_bpmn.py:166` accepts any digit string. The skill only shows `15` as an example shape. The agent has no way to discover this number, and no penalty for getting it wrong.

**Fix** — ask for what you check:

> Include an integer `uipath:migrationVersion` element on the process.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
```diff
@@ -30,6 +33,18 @@ def root_variable_identifiers(process: ET.Element) -> set[str]:
 
 def main() -> None:
     path, root = parse_bpmn("ExpenseApprovalBpmn")
+    bpmn_path = Path(path)
+    project_path = bpmn_path.parent / "project.uiproj"
+    if not project_path.is_file():
+        fail(f"missing project descriptor beside BPMN: {project_path}")
```
rockymadden, line 39: 🟡 **minor** — this hard-fail makes the `.bpmn.ts` alternative in `simple_approval_bpmn.yaml:53` dead.

That weight-1.0 criterion accepts `project.uiproj` **or** `ExpenseApprovalBpmn.bpmn.ts`. With this check in place the `.ts` branch can never carry the task — the weight-5.0 check kills the run anyway.

**Fix** — drop the `.ts` alternative from that criterion.

tests/scripts/test_bpmn_project_scaffold_contract.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Derived, not whitelisted: adding a node type to the doc example must not
+    # surface as a DI-coverage mismatch.
+    node_ids = {
+        element.attrib["id"]
+        for element in process
+        if element.get("id")
+        and element.tag.rsplit("}", 1)[-1]
+        not in ("sequenceFlow", "extensionElements")
+    }
```
rockymadden, line 83: 🟡 **minor** — the comment promises more than the code delivers.

The derivation counts every `id`-bearing process child as needing a `BPMNShape`:

- `bpmn:association` → needs a `BPMNEdge`, would be demanded as a shape
- `bpmn:dataObject` → needs neither, would be demanded as a shape

Correct for today's three-node example, false-fails the first time it grows — which is exactly the case the comment claims is safe.

**Fix** — exclude non-flow-node children explicitly, or scope the comment to what it actually guarantees.

rockymadden (changes requested):
## 📋 Review

### What

Replaces the skill's hand-invented BPMN project scaffold with the one `uip maestro bpmn init` actually writes:

- `isExecutable="true"` → preserve the initializer's shape (attribute omitted)
- `project.uiproj.main` → `operate.json.main` owns the runtime path
- `update-metadata` → `refresh`, with its `project.uiproj` precondition stated
- Evals go layout-agnostic; new `tests/scripts/test_bpmn_project_scaffold_contract.py` pins the doc example

---

### ✅ Verified against the CLI

Ran `uip` 1.203.0 in a scratch dir, not just read the diff. The core claims hold:

| Claim | Result |
|---|---|
| `init` writes six files, returns `Data.Path` | ✅ |
| `init` omits `isExecutable`, emits `uipath:variables`+`uipath:bindings`, `exporter="UiPath Maestro (https://uipath.com)"` | ✅ |
| `--skip-solution-registration` → bare six-file layout | ✅ |
| `operate.json.main` = `/content/<file>.bpmn#<startEventId>` | ✅ |
| `refresh` without `project.uiproj` → `Required file is missing` / `RetryWillNotFix` | ✅ exact string |
| Two-key `project.uiproj` → `refresh` writes the other four | ✅ |
| `refresh` re-points `main` after a start-event rename | ✅ |
| `elementId` mismatch → entry point gets an empty `input` schema | ✅ silently dropped |
| The new minimal example validates, refreshes, **and packs** | ✅ |
| `node:decision` swap matches the closed vocabulary at `tests/README.md:112` | ✅ `node:gateway`/`script`/`service-task` were never valid |
| Dropped `file_contains` assertions covered by the deep checkers | ✅ `check_smoke_completed_wired.py:46`, `check_timer_start.py:50` |

---

### 🔴 Overall findings

**1. `update-metadata` still prescribed on the ship path — major.**

SKILL.md and `project-layout.md` now say `refresh`. These four didn't move:

- `references/operate/ship.md:22-23`
- `references/operate/CAPABILITY.md:35`
- `references/diagnose/failure-modes.md:51`
- `references/shared/local-metadata-regeneration-guide.md:23` — opens the contract with `update-metadata` two lines after the callout calling it deprecated

`local-metadata-regeneration-guide.md:18` says `update-metadata` never materializes connection bindings and faults at runtime with error 102010. This PR fixed the authoring path and left the operate path on the command it just documented as broken.

**Fix:** swap all four to `refresh <project-path>`.

**2. PR body has no passing-run claim — minor.**

The repo's own advisory lint flags this as High. Verification names only the pytest guard.

**Fix:** add `Ran skill-bpmn-simple-approval-bpmn locally and it passed.`

**3. Two claims I could not verify — minor.**

`migrationVersion` appears **zero** times in `@uipath/cli/dist` (`grep -r`), so `Number.parseInt` and the number `15` are not checkable from the CLI or this repo. The only in-repo Studio-Web-produced sample is `version="3"` (`tests/tasks/uipath-troubleshoot/products/maestro/maestro-stuck-rpa-job/data/m/r/7b3d8c49fd.json`). Cite the PO.Frontend source.

I also only tested `isExecutable="true"` against the CLI validator (accepts it, `Status: Valid`) — **not** against Studio Web import.

---

### 🎯 Recommendation

**Request changes.**

The scaffold correction is right and I verified it end to end. Two behavior-affecting regressions block:

- the missing `xmlns:xsi` breaks any agent that follows the doc into a conditional flow
- the loosened `command_pattern` turns a real assertion into a near-no-op

Both are one-line fixes. The `11.5` conflict needs a decision, not just a doc tweak.

---

**tl;dr** — Good PR, the CLI backs every scaffold claim. Two one-line regressions block: the minimal example lost `xmlns:xsi` while still telling agents to write `xsi:type`, and `author_validate.yaml` dropped `ET.parse` from its regex so a `print()` now passes. The new integer-`migrationVersion` rule also collides with an eval that demands `11.5`. Four other files still prescribe the deprecated `update-metadata`.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Restored on both skeletons. The gateway example took the declaration with it when it moved, and `validate` not catching it is exactly why it survived — the two task prompts that run `ET.parse` directly would have.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Collapsed to your wording, and rewrapped the paragraph to clear the ragged reflow.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Added the third line. You are right that it had no replacement anywhere — I deleted a mechanism rather than moving it.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Scoped it: author integers, but preserve an existing value byte-for-byte when editing, so `contract_variant_wrappers` stays satisfiable without touching a task this PR does not own. Also swapped the citation to the PO.Frontend serializer — you are right that `migrationVersion` appears nowhere in the CLI bundle, so the claim was not checkable where I pointed.

tests/tasks/uipath-maestro-bpmn/smoke/author_validate.yaml
DevMomo: Restored. Verified both directions after: the suggested command matches, `python3 -c "print(...)"` does not. Dropping it was collateral from the rglob rewrite — `timer_start` kept it because I did not touch that line.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
DevMomo: Fixed.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
DevMomo: Dropped to exactly what the checker accepts. I had added that phrasing one round earlier trying to avoid teaching a stale literal, and overshot into asking for something undiscoverable.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
DevMomo: Dropped the `.bpmn.ts` alternative. The skill teaches no such shape and the 5.0 checker hard-requires `project.uiproj`, so the branch was unreachable.

tests/scripts/test_bpmn_project_scaffold_contract.py
DevMomo: Excluded `association`, `dataObject` and `dataObjectReference` explicitly, so the comment now describes what the code does.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -40,9 +40,10 @@ namespace, at least one `<bpmn:process>`, and (to render) a
     xmlns:di="http://www.omg.org/spec/DD/20100524/DI"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns:uipath="http://uipath.org/schema/bpmn"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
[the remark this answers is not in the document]
DevMomo: Removed. My fix for the missing declaration introduced a duplicate one — same file, same attribute, opposite failure. Added `tests/scripts/test_bpmn_doc_examples_parse.py`, which fails on both: the duplicate, and the original omission.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -515,9 +523,12 @@ Safe, surgical edits on an existing `.bpmn` (preserve content you did not author
...
   second diagram plane for the subprocess so nested content renders.
-- **Add an entry point**: use a root-level start event, add a stable unique
-  `uipath:entryPointId`, and declare input/output variables whose `elementId`
-  matches that start event.
+- **Add an entry point**: use a root-level start event and generate a stable,
+  unique UUID for its serializer-owned `uipath:entryPointId`. Do not copy the
+  example UUID; this scaffold field is not a registry-owned node payload. Also
+  declare the public input/output variables whose `elementId` matches that
+  start event — without the match, `refresh` derives an empty `input` schema
+  for the entry point.
```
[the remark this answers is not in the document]
DevMomo: Corrected — inputs match the entry start event, outputs match the end event that returns them. The declaration example above had it right and I contradicted it in the edit guidance.

rockymadden (approved):

PR #2366, opened 2026-07-30
fix(uipath-maestro-bpmn): align ScriptTask authoring with runtime contract

DevMomo:
## Problem

A ScriptTask authored from the registry's own template **never runs its script**, silently:

```xml
<uipath:type value="BPMN.ScriptTask" version="v1" />   <!-- what registry get emits -->
<bpmn:script><![CDATA[ throw new Error("BOOM"); ]]></bpmn:script>
```

That process reports `FinalStatus: Completed`, zero incidents, clean `validate`. Serialized as `BPMN.Variables` the same script faults properly (`300501 Error invoking script task`).

**Why the e2e needs this:** with the wrong discriminator its classification task never runs, so severity is never computed (score 0.600).

## Fix

- Retrieve `BPMN.ScriptTask`, **serialize `BPMN.Variables`** — a clean validate does not prove the mapping is right
- Compatibility template: `vars`/`metadata` args, standard `scriptResponse`/`Error` outputs; only for the known older shape
- Mapped args are **top-level identifiers** (`vars`, not `args.vars` → `args is not defined`)
- **v2+ only:** return a bare value; the runtime wraps it under `response`. v1 spreads keys, so there the wrapper is required — and `scriptVersion` is matched `/^v(\d+)$/i` with no trimming, so `v3.1` silently falls back to v1
- Brownfield nodes keep their discriminator and `scriptVersion`; migration needs confirmation

## Verification

`tests/scripts/test_bpmn_script_task_contract.py` pins the template; the Jint eval asserts the serialized discriminator.


skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -186,59 +186,80 @@ See [expression-authoring.md](expression-authoring.md) for expression rules.
 Sub-process-scoped variables go in that sub-process's own
 `<uipath:variables>`.
 
-## Script tasks (`BPMN.ScriptTask`) — Jint runtime contract
+## Script tasks — current v3 Jint contract
```
rockymadden: We should think through how to handle for BPMN brownfield scenarios that have older versions of script tasks to ensure the skill covers those use cases.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -186,59 +186,80 @@ See [expression-authoring.md](expression-authoring.md) for expression rules.
...
+
+- Set `<uipath:scriptVersion value="v3" />` for new ScriptTasks. Preserve
+  imported legacy versions unless the user asks to migrate them.
+- Declare a task-scoped mutable `scriptResponse` variable and a task-scoped
+  mutable `Error` variable. `Error` uses `type="jsonSchema"`, the standard
+  error schema, and `elementId="<script-task-id>"`.
+- Add a `uipath:context/uipath:inputSchema` of type `jsonSchema` that declares
+  `vars` and `metadata` as objects.
+- Add `uipath:input name="args" type="json" target="bodyField"` with
+  `{"vars":"=vars","metadata":"=metadata"}` in CDATA, matching Studio's
```
rockymadden: Do we need to mention Studio?

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -186,59 +186,80 @@ See [expression-authoring.md](expression-authoring.md) for expression rules.
...
+  canonical serializer.
+- Read process data in JavaScript through `vars.<stable-variable-id>`.
+- Return the intended scalar or object directly. Map the standard
+  `scriptResponse` output from `=result.response` and `Error` from `=Error`.
+  Downstream nodes and the completion EndEvent can read the declared
+  `scriptResponse` variable directly. Only when a distinct business variable
+  is needed, add a custom output that reads `=vars.<script-response-id>` and
+  writes that variable.
+- Studio maps JavaScript/JSON Schema `number` to BPMN primitive
```
rockymadden: This is true but do we need to note this in the skill or is it noise?

rockymadden:
**`scriptResponse` output type: template vs. checker disagree**

The bundled template and the eval checker disagree on the `type` of the `scriptResponse` output, and I think it can cause the exact eval failures this PR is trying to prevent.

- `validator/bpmn-spec.json` — the `BPMN.ScriptTask` template hardcodes `<uipath:output name="scriptResponse" type="jsonSchema" … />`.
- `references/structural-bpmn.md` — the worked example uses `type="double"` and declares the `scriptResponse` variable as `double`.
- `tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py` — requires the `scriptResponse` variable to be `double` and the output `type` to match that variable.

Each guard passes on its own artifact (the contract guard checks the generic template = `jsonSchema`; the eval checks the concrete numeric task = `double`), so CI stays green. But `SKILL.md` step 3 says to author from each node's `xmlTemplate` and "fill placeholders only" — and `type="jsonSchema"` is not a placeholder. An agent that copies the template literally for a numeric result emits `type="jsonSchema"` on `scriptResponse` and fails the checker. The type-mapping bullet ("number → double") points the other way, so the two instructions pull against each other.

Suggestion: make the output `type` a placeholder in the template (e.g. `type="{scriptResponseType}"`), or add a one-line note that `scriptResponse`'s type follows the returned value's primitive type. Either one removes the trap without weakening the guards.

rockymadden (approved):

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Addressed in a8913c8a3. New-node authoring and brownfield edits now have separate rules: existing ScriptTasks preserve their discriminator and scriptVersion, and requested argument/output changes are made surgically within the existing contract. Migration requires explicit confirmation. The update-node eval now starts from an older v2 ScriptTask and rejects implicit migration; the fresh run passed 3/3 at 1.000.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Agreed. I removed Studio as an authority from the ScriptTask guidance, diagnostics, and eval wording. Dispatch is shape-based: use the live registry first, use the compatibility fallback only for the recognized older shell, and keep an unfamiliar newer live template authoritative.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: I kept the number/integer mapping because it is operationally necessary, but removed the Studio attribution and shortened the rule. The failure this protects against was emitting a JSON Schema number as a BPMN primitive instead of double. The template now exposes scriptResponseType, so the output type must match the declared response rather than being hardcoded.

DevMomo:
Addressed in a8913c8a3. The compatibility template now uses {scriptResponseType}, and the authoring guidance requires that placeholder to match the declared response variable: number maps to BPMN double, integer to integer, and structured values to jsonSchema.

The contract guard now asserts the placeholder instead of a hardcoded type. A fresh numeric ScriptTask eval passed both criteria at 1.000, including the independent contract checker and uip maestro bpmn validate.

The same commit also removes UI/version branding from the decision rule and adds explicit brownfield preservation behavior in response to the inline feedback.

tests/tasks/uipath-maestro-bpmn/nodes/script_jint_guidance.yaml
```diff
@@ -1,54 +1,40 @@
 task_id: skill-bpmn-script-jint-guidance
 description: >
-  Script task eval: agent authors a BPMN script task that follows the Jint
-  runtime boundary instead of Node.js, browser, filesystem, or network APIs.
-tags: [uipath-maestro-bpmn, smoke, mode:build, lifecycle:generate, shape:single-node, node:script]
+  Script task eval: agent authors the supported ScriptTask runtime contract and
+  stays within the Jint runtime boundary.
+tags: [uipath-maestro-bpmn, smoke, "mode:build", "lifecycle:generate", "shape:single-node", "node:script"]
```
rockymadden: **Major — this task is now flaky, and it's gated at 95%.**

Across this stack's CI, `skill-bpmn-script-jint-guidance` scored:

| PR | job | score |
|---|---|---|
| #2387 | 102135170171 | **0.250** |
| #2655 | 102135107462 | 1.000 |
| #3112 | 102135102983 | 1.000 |
| #3113 | 102135103626 | **0.250** |

`0.250` = 2.0/8.0: only the `uip maestro bpmn validate` criterion passed and the 6.0-weight checker failed. Since `smoke` is gated at a 95% pass rate, a roughly 50/50 task fails the gate for every PR that touches this skill — which is what all five PRs in this stack are currently doing.

The cause looks structural rather than random. The prompt was (correctly, per `.claude/rules/test-writing.md` "Prompts: Minimal") reduced to "use the supported ScriptTask and public input/output contracts." Meanwhile `check_script_jint_guidance.py` grew to demand ~30 exact facts the prompt never states: the literal `$schema` string, `required` being exactly `[]`, exactly two StartEvent output bridges, exactly one EndEvent bridge, `custom="true"` on the optional riskScore mapping, `double` vs `integer` on each variable, and a non-example UUID entryPointId. An agent can satisfy the stated goal and still miss a third of those.

Two ways out, either is fine:

1. Retag `integration` and leave a thin `smoke` variant that grades the discriminator + `args` shape only. That keeps the strict grader without putting it behind the 95% gate.
2. Keep the deterministic structural assertions in `tests/scripts/test_bpmn_script_task_contract.py` (already added in this PR, runs without an agent) and pare the task checker back to the facts the prompt actually asks for.

I'd lean (1) — the assertions are genuinely valuable, they just shouldn't be graded through an agent at smoke tier.

tests/tasks/uipath-maestro-bpmn/nodes/script_jint_guidance.yaml
```diff
@@ -1,54 +1,40 @@
...
   - type: run_command
-    description: "BPMN source was created"
-    command: "find . -type f -name '*.bpmn' -path '*RiskScoreScriptBpmn*' | grep -q ."
-    timeout: 10
+    description: "BPMN uses supported public I/O and ScriptTask runtime contracts"
+    command: "python3 $TASK_DIR/check_script_jint_guidance.py"
+    timeout: 30
     expected_exit_code: 0
-    weight: 1.0
+    weight: 6.0
```
rockymadden: **Minor — weight is off-taxonomy.** `.claude/rules/test-writing.md` reserves `5.0–6.0` for "e2e execution" and `3.0` for "primary validation." This is a `smoke`-tagged task, so `3.0` is the on-taxonomy weight for the primary checker.

Not cosmetic: the 6.0/2.0 split is what makes a single checker failure land at 0.250 instead of 0.40, which changes how the smoke gate reads the result.

skills/uipath-maestro-bpmn/validator/bpmn-spec.json
```diff
@@ -1906,9 +1908,9 @@
...
+      "xmlTemplate": "<bpmn:scriptTask id=\"{id}\" name=\"{name}\" scriptFormat=\"JavaScript\">\n  <bpmn:extensionElements>\n    <uipath:mapping version=\"v1\">\n      <uipath:type value=\"BPMN.Variables\" version=\"v1\" />\n      <uipath:context>\n        <uipath:inputSchema type=\"jsonSchema\"><![CDATA[{\"$schema\":\"http://json-schema.org/draft-07/schema#\",\"type\":\"object\",\"properties\":{\"vars\":{\"type\":\"object\"},\"metadata\":{\"type\":\"object\"}},\"required\":[]}]]></uipath:inputSchema>\n      </uipath:context>\n      <uipath:input name=\"args\" type=\"json\" target=\"bodyField\"><![CDATA[{\"vars\":\"=vars\",\"metadata\":\"=metadata\"}]]></uipath:input>\n      <uipath:output name=\"scriptResponse\" type=\"{scriptResponseType}\" source=\"=result.response\" var=\"{scriptResponseVarId}\" />\n      <uipath:output name=\"Error\" type=\"jsonSchema\" source=\"=Error\" var=\"{errorVarId}\" />\n    </uipath:mapping>\n    <uipath:scriptVersion value=\"v3\" />\n  </bpmn:extensionElements>\n  <bpmn:incoming>{incomingEdge}</bpmn:incoming>\n  <bpmn:outgoing>{outgoingEdge}</bpmn:outgoing>\n  <bpmn:script><![CDATA[return {};]]></bpmn:script>\n</bpmn:scriptTask>",
```
rockymadden: **Major — `type="{scriptResponseType}"` is the one attribute in this template I could not verify, and the evidence points the other way.**

Everything else here matches the serializer byte-for-byte. But on `scriptResponse`'s `type`, PO.Frontend consistently writes `jsonSchema`, never a primitive:

- `src/utils/ScriptTaskUtils.ts` — `initializeScriptTaskOutputs()` hard-codes `{ name: "scriptResponse", type: "jsonSchema", source: "=result.response" }`. No branch produces a primitive type.
- `src/services/serialization/xml-serialization.test.ts` — both Studio Web golden fixtures serialize `<uipath:output name="scriptResponse" type="jsonSchema" ... />` with a matching `<uipath:inputOutput ... name="scriptResponse" type="jsonSchema">` declaration (lines 1217/1236 and 2779/2788).

So `type="double"` on `scriptResponse` is a shape the canvas never emits. A Studio Web open/save round-trip may well normalize it back to `jsonSchema` — and that is exactly the class of "local validate is clean, mapping is still wrong" failure this PR exists to fix. It would be unfortunate to close one instance and open another.

Did you round-trip a `double` scriptResponse through the canvas and confirm it survives? If yes, ignore me and please drop a note in `inputNotes` saying so, because the source reads the other way and the next person will re-litigate this. If not, the safer default is `type="jsonSchema"` on both the variable and the mapping, with the typed business variable picked up downstream via the `=vars.<script-response-id>` custom output the docs already describe.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -188,63 +188,97 @@ See [expression-authoring.md](expression-authoring.md) for expression rules.
...
+  `type="double"` and an `integer` result uses `type="integer"`. Keep JSON
+  Schema names inside schema bodies; structured object or array results use a
+  declared `jsonSchema`. Do not use `number` or `long` as BPMN primitive
+  mapping types.
+- **v2+ only:** do not add an extra `{ response: ... }` wrapper around the
+  script return — the runtime already exposes the direct return beneath
+  `result.response`, so wrapping yields `result.response.response`. Under v1
+  (the default when the marker is absent) the runtime spreads the returned
+  object's keys instead, so there the wrapper is required. Note the marker is
+  matched as `/^v(\d+)$/i` with no trimming: `3`, `v3.1` and `" v3 "` all fall
```
rockymadden: **Minor — drop the `/i`.** The real matcher is `getVersionNumber` in `src/utils/Tools.ts:930`:

```ts
const pattern = new RegExp(`^${prefix}(\\d+)$`);
return match ? Number.parseInt(match[1], 10) : 0;
```

No `i` flag, so `V3` also returns 0 and falls back to v1 semantics. Worth correcting precisely because this bullet's whole value is that it is exact — the rest of the claim (`3`, `v3.1`, `" v3 "` all silently degrade) checks out.

tests/tasks/uipath-maestro-bpmn/edit/update_node/fixture/RiskScoring.bpmn
```diff
@@ -20,15 +20,15 @@
...
           <uipath:type value="BPMN.ScriptTask" version="v1" />
           <uipath:input name="args"><![CDATA[{"amount":"=vars.Var_Amount"}]]></uipath:input>
           <uipath:output name="tier" type="string" var="Var_Tier" source="=result.response" />
         </uipath:mapping>
       </bpmn:extensionElements>
       <bpmn:incoming>Flow_SS</bpmn:incoming><bpmn:outgoing>Flow_SF</bpmn:outgoing>
-      <bpmn:script><![CDATA[return { response: amount > 500 ? "high" : "low" };]]></bpmn:script>
+      <bpmn:script><![CDATA[return amount > 500 ? "high" : "low";]]></bpmn:script>
     </bpmn:scriptTask>
     <bpmn:scriptTask id="Task_Format" name="Format Message" scriptFormat="JavaScript">
```
rockymadden, line 33: **Minor — the untouched sibling now contradicts the rule this PR adds.**

`Task_Format` carries `scriptVersion value="v3"` with `return { response: "risk tier is " + tier };` (line 43) mapped from `source="=result.response"`. By this PR's own new v2+ rule, that resolves to `result.response.response` and reads back empty — the exact bug being fixed. Same file, line 15: `type="number"` on `Var_Amount`, which `structural-bpmn.md:238` now forbids as a BPMN primitive.

I understand the fixture is deliberately legacy and the check's job is to prove the edit doesn't migrate anything. That's the right test. But this file is input the agent reads, and a `v3` task with a response wrapper is the in-file convention it will pattern-match from — while `Task_Score` right above it now models the opposite. Mixed signal in a teaching artifact.

Cheapest fix: drop `Task_Format`'s `scriptVersion` marker entirely. Absent the marker `getVersionNumber` returns 0, v1 spread semantics apply, and the `{ response: ... }` wrapper becomes genuinely correct there — so the fixture stays legacy-by-design without modelling a broken node. (Flipping it to `v2` would not help; v2 is on the no-wrapper side of the rule.)

tests/scripts/test_bpmn_script_task_contract.py
```diff
@@ -0,0 +1,108 @@
...
+        'xmlns:uipath="http://uipath.org/schema/bpmn">'
+        f"{template}"
+        "</root>"
+    )
+    return ET.fromstring(wrapper)[0]
+
+
+def test_script_task_registry_entry_uses_supported_shape() -> None:
+    entry = _script_task_entry()
+    assert entry["bpmnElement"] == "bpmn:ScriptTask"
```
rockymadden, line 50: **Minor — this freezes two spellings of the same fact.** Line 47 pins `bpmnElement` as PascalCase `bpmn:ScriptTask`; line 52 pins the template's host tag as lower-camel `scriptTask`. And `references/registry-workflow.md:55` says `bpmnElement` is "the host BPMN element the template uses (for source files, normalize to lower-camel such as `bpmn:serviceTask`)."

So the spec entry now disagrees with both its own template and the doc describing the field, and this test locks that in. Suggest normalizing the spec's `bpmnElement` to `bpmn:scriptTask` and updating this assert, so there's one answer to "what tag does a ScriptTask use."

Otherwise this file is a good addition — deterministic, no agent in the loop, and it pins exactly the things that silently drifted.

rockymadden (approved):
## Approve

The contract fix is correct and I verified it independently against `origin/develop` of PO.Frontend rather than taking the PR body on faith:

|---|---|---|
| ScriptTask serializes `BPMN.Variables`, not the lookup key | `bpmn-moddle.ts:801-812` routes `bpmn:ScriptTask` through `createModdleExtensionVariableMapping`, which hard-codes `value: "BPMN.Variables"` (`:680`) | correct |
| Inputs merge into one `args` JSON at `bodyField` | `createExtensionElement` ScriptTask branch, `bpmn-moddle.ts:887-897` | correct |
| inputSchema is exactly draft-07 + `required: []` | `createSchemaFromVariables`, `Tools.ts:575-606` — byte-identical to the template | correct |
| `scriptVersion` sits after `</uipath:mapping>` | `mappingOptions` push order, `bpmn-moddle.ts:696-700` | correct |
| v3 default inputs are `vars` + `metadata` | `initializeScriptTaskInputs()`, `S183Migrations` | correct |
| v2+ needs no `{ response: ... }` wrapper | `s178_2-migrations.ts:40` — v2 + `source="=result"` was *broken*, migrated to `=result.response` with downstream `vars.{id}.response` → `vars.{id}` | correct |
| Canvas default script is `return {};` | `actions.ts:352` | correct |
| `number`/`long` are not BPMN primitives | `PRIMITIVE_TYPES`, `constants.ts:371` = string/integer/float/double/boolean/datetime/date/time | correct |

Worth stating plainly: this **reverses** prior guidance that said "never use `<uipath:type value=\"BPMN.Variables\">` on a script task output mapping." That guidance was wrong, and the serializer proves it. Good catch, and the "a clean `validate` does not prove the mapping is right" framing is the right lesson to encode.

Two things I'd like addressed as follow-ups rather than blocks — see inline. The important one is that the rewritten `script_jint_guidance` task is now flaky at the smoke tier (0.250 on #2387 and #3113, 1.000 on #2655 and #3112), which fails the 95% gate for every downstream PR in this stack. That's a test-tier problem, not a wrong-fix problem, so it doesn't hold up the contract correction.

No AI slop. Prose is dense and every claim is falsifiable.

tests/tasks/uipath-maestro-bpmn/nodes/script_jint_guidance.yaml
DevMomo: Took (1) — retagged `integration`, so the strict grader is off the 95% gate. Also fixed the unfair half: it demanded `custom="true"`, which no reference taught. That is now in `structural-bpmn.md`, and the two assertions that graded byte-identity rather than behavior (`$schema` literal, `required` exactly `[]`) are gone. No thin smoke variant — `test_bpmn_script_task_contract.py` already pins the discriminator and `args` shape deterministically, without an agent.

tests/tasks/uipath-maestro-bpmn/nodes/script_jint_guidance.yaml
DevMomo: Fixed — `3.0` primary, `1.5` supporting.

skills/uipath-maestro-bpmn/validator/bpmn-spec.json
DevMomo: You are right, and I had not round-tripped it. Confirmed on `origin/develop` today: `ScriptTaskUtils.ts:8-14` hard-codes `type: "jsonSchema"` with no primitive branch, and both golden fixtures serialize it (1217/1236, 2779/2788). Dropped the `{scriptResponseType}` placeholder entirely — `jsonSchema` is now literal in the template, `inputNotes` says the canvas emits no other value, and the typed business value is read downstream via `=vars.<scriptResponseVarId>`.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Fixed — `/^v(\d+)$/`, and `V3` added to the list that silently degrades.

tests/tasks/uipath-maestro-bpmn/edit/update_node/fixture/RiskScoring.bpmn
DevMomo: Dropped `Task_Format`'s `scriptVersion` marker as you suggested, so v1 spread semantics make its wrapper correct. Also took line 15 to `double` — that one is node-scoped, so it is the primitive rule rather than the public-declaration rule. Those two rules were contradicting each other in `structural-bpmn.md`; both are now scope-qualified and cross-referenced.

tests/scripts/test_bpmn_script_task_contract.py
DevMomo: Half-taken. Normalizing the spec would make ScriptTask the only lower-camel entry of 47 — `bpmnElement` is the PascalCase model type everywhere (`bpmn:ServiceTask`, `bpmn:StartEvent`, …), and the canvas agrees (`ConfigureScriptTaskFields.tsx:36`), while the serialized tag is its lower-camel form. So the defect was the `registry-workflow.md` line describing the field, which now says exactly that and points at `xmlTemplate` for the tag. Kept both asserts, with a comment explaining why there are two spellings.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -202,63 +202,103 @@ null outputs.
...
+  or array, `double` for a number, otherwise the primitive's own name. The
+  canvas retypes it on every script edit
+  (`ScriptTaskProperties.tsx:347`); `jsonSchema` is only the pre-edit default.
+  Downstream nodes and the completion EndEvent can read the declared
+  `scriptResponse` variable directly. Only when a distinct business variable
+  is needed, add a custom output that reads `=vars.<script-response-id>` and
+  writes that variable, and mark it `custom="true"` — that is how the canvas
+  tags a user-added mapping and how it keeps the standard `scriptResponse` and
+  `Error` outputs separate from yours.
+- In node-scoped variable and mapping attributes, a JSON Schema `number` result
```
rockymadden, line 253: 🟠 **major** - contradicts the skeleton `SKILL.md:191` tells agents to author from.

This bullet makes `double` mandatory on node-scoped variables and mapping attributes. The canonical minimal file still uses `number` in exactly those positions:

- `structural-bpmn.md:98` `<uipath:inputOutput id="Var_Amount" ... type="number" />`
- `structural-bpmn.md:100` `<uipath:inputOutput id="Var_Echo" ... type="number" />`
- `structural-bpmn.md:109,118,128` mapping attributes `type="number"`

`:128` is an EndEvent mapping writing a public var, and the new grader at `check_script_jint_guidance.py:449` requires `type="double"` in that exact position.

Base `structural-bpmn.md:154-156` only said `double` there "is fine". Promoting that to mandatory without updating the skeleton means an agent that follows the skeleton fails the new eval.

**Fix:** change the five `type="number"` occurrences at `:98,100,109,118,128` to `double`, leaving `:97,99` (`uipath:input` / `uipath:output`) as `number`.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -52,7 +52,7 @@ uip maestro bpmn registry get <extensionType> --output json
 | Field | Use |
 | --- | --- |
 | `xmlTemplate` | The literal node XML with `{placeholder}` slots. **Author from this; fill placeholders only.** |
-| `bpmnElement` | The host BPMN element the template uses (for source files, normalize to lower-camel such as `bpmn:serviceTask`). |
+| `bpmnElement` | The host element's PascalCase model type (`bpmn:ServiceTask`). The serialized source tag is the lower-camel form of it (`<bpmn:serviceTask>`) — take the tag from `xmlTemplate`, not from this field. |
```
rockymadden: 🟠 **major** - true for the one template this PR rewrote, false for the other 27.

"take the tag from `xmlTemplate`, not from this field" holds only for `BPMN.ScriptTask`. Bundled `xmlTemplate` host tags today:

```
Actions.HITL              <bpmn:UserTask
Orchestrator.StartJob     <bpmn:ServiceTask
Intsvc.HttpExecution      <bpmn:SendTask
Maestro.SendMessageEvent  <bpmn:IntermediateThrowEvent
... 23 more
```

29 entries total: 27 PascalCase in `xmlTemplate`, only `BPMN.ScriptTask` and `Intsvc.ActivityExecution` lower-camel. An agent taking the tag from `xmlTemplate` emits `<bpmn:ServiceTask>` for 27 of them.

Also contradicts `SKILL.md:180-182`, which says to normalize a PascalCase template tag to lower-camel.

**Fix:** `The host element's PascalCase model type. Normalize both this field and the xmlTemplate host tag to lower-camel (bpmn:serviceTask) when serializing.`

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
```diff
@@ -31,56 +35,191 @@
     "window.",
     "document.",
     "await ",
-]
+    "Math.random",
```
rockymadden, line 38: 🟠 **major** - four rules the agent cannot learn from the skill.

`Math.random`, `Date.now`, `new Date`, and `crypto.` are all valid Jint APIs. `structural-bpmn.md:273-276` forbids npm, filesystem, network, browser globals, timers, and long-running async, not these. `grep -rn "determinis|Math.random|Date.now" skills/uipath-maestro-bpmn/` returns nothing relevant.

The prompt at `script_jint_guidance.yaml:29` says only "Keep the JavaScript deterministic and Jint-compatible", then enumerates the six categories the skill does teach. An agent that writes `new Date()` fails on an unstated rule.

**Fix:** either add the ban to `structural-bpmn.md:273-276` and name it explicitly in `script_jint_guidance.yaml`, or drop the four tokens.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -202,63 +202,103 @@ null outputs.
...
+  `name`.
+- Do not mutate `Globals.*`, `vars.*`, or process variables in the script.
+  Return a value and let output mappings write declared variables.
+- Only the documented Jint helpers are available (`uipath.aggregate`,
+  `uipath._aggregate`, `uipath._pipe`, and a no-op `console`). Do not use npm
+  packages, filesystem, network, browser globals, timers, or long-running async
+  work. The execution envelope is approximately 64 MB / 30 seconds.
 
 ```xml
+<uipath:inputOutput id="Var_ScriptResponse" name="scriptResponse"
```
rockymadden, line 286: 🟡 **minor** - fragment has no parent, so copying it literally fails the grader.

The two `uipath:inputOutput` declarations are shown as bare siblings. `check_script_jint_guidance.py:112` reads only `bpmn:process/bpmn:extensionElements/uipath:variables/*`, so declarations placed anywhere else are invisible to it.

**Fix:** wrap in `<uipath:variables version="v1">` with a `<!-- in bpmn:process/bpmn:extensionElements -->` marker, matching the style already used at `:167-171`.

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
```diff
@@ -108,54 +245,379 @@ def strip_js_comments(script: str) -> str:
...
+    )
+    end_output = exactly_one(
+        [
+            output
+            for output in mapping_outputs(end)
+            if output.attrib.get("name") == "riskScore"
+            and output.attrib.get("var") == attr(public_risk, "id")
+            and output.attrib.get("type") == "double"
+            and re.fullmatch(
+                r"=vars\.[A-Za-z_][A-Za-z0-9_]*",
```
rockymadden: 🟡 **minor** - rejects GUID-style variable ids.

`=vars\.[A-Za-z_][A-Za-z0-9_]*` allows no hyphens or dots. A GUID id fails as "no numeric result to public riskScore output bridge" rather than as a naming problem, which sends the agent hunting the wrong thing. `:344` in this same file requires `entryPointId` to be a UUID, so GUID-ish ids are plausible here.

**Fix:** `r"=vars\.[\w.-]+"`, then resolve the captured id against `variables` so an unknown id fails with its own message.

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
```diff
@@ -108,54 +245,379 @@ def strip_js_comments(script: str) -> str:
...
+    )
+    entry_point_id = entry_point.attrib.get("value", "").strip()
+    try:
+        UUID(entry_point_id)
+    except (ValueError, AttributeError):
+        fail("manual start entryPointId must be a valid UUID")
+    if entry_point_id == "00000000-0000-4000-8000-000000000001":
+        fail("manual start entryPointId copied the documentation example")
+
+    if attr(task, "scriptFormat") != "JavaScript":
```
rockymadden, line 364: 🟡 **minor** - undocumented tightening.

Was `attr(task, "scriptFormat").lower() != "javascript"`. Case-sensitive is probably correct (it matches the serializer), but it is not in the PR description and `structural-bpmn.md:205` does not state the required casing, so the eval enforces something the skill never says.

**Fix:** in `structural-bpmn.md:205`, `scriptFormat="JavaScript"` (exact casing; the serializer is case-sensitive).

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
```diff
@@ -31,56 +35,191 @@
...
         next_char = script[index + 1] if index + 1 < len(script) else ""
         if in_string:
             result.append(char)
-            if char == "\\":
-                if index + 1 < len(script):
-                    result.append(script[index + 1])
-                    index += 2
-                    continue
-            elif char == in_string:
+            if char == "\\" and index + 1 < len(script):
```
rockymadden, line 218: 🟡 **minor** - behavior-identical refactor, unrelated to the ScriptTask contract.

This escape-branch restructure produces the same output as the `if`/`elif` it replaces. Same for `:238` (`script[index:index+2] != "*/"`) and the `FORBIDDEN` list-to-tuple at `:28`.

They inflate a file that is already +533/-76, which makes the contract change harder to review.

**Fix:** revert these three.

tests/scripts/test_bpmn_script_task_contract.py
```diff
@@ -0,0 +1,111 @@
...
+        f"{template}"
+        "</root>"
+    )
+    return ET.fromstring(wrapper)[0]
+
+
+def test_script_task_registry_entry_uses_supported_shape() -> None:
+    entry = _script_task_entry()
+    # Two spellings of one fact, deliberately: bpmnElement is the PascalCase
+    # model type (as for all 47 spec entries), while the serialized tag below
```
rockymadden: 🟡 **minor** - wrong count.

`bpmn-spec.json` has 29 `extensionTypes`, not 47.

**Fix:** `(as for every spec entry)`, so the comment does not need updating when the count moves.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -202,63 +202,103 @@ null outputs.
...
+The following rules describe newly authored tasks using that compatibility
+template. In brownfield BPMN, preserve the existing mapping discriminator and
+`uipath:scriptVersion`. Preserve arguments and outputs except where the
+requested edit requires a surgical change, and make that change using the
+node's existing contract. Migration to a different contract requires explicit
+confirmation. If runtime evidence establishes an incompatibility, present that
+evidence and obtain confirmation before migrating it.
+
+- Keep the selected template's `uipath:scriptVersion` marker on a new task.
+  Treat it as a serialized field, not as the name of the overall authoring
```
rockymadden: 🟡 **minor** - half this sentence is commentary on the doc rename.

"not as the name of the overall authoring contract" changes nothing the agent does. "or a reason to migrate existing tasks" does.

**Fix:** `Treat it as a serialized field, not a reason to migrate existing tasks.`

rockymadden (changes requested):
## What

Reverses the skill's ScriptTask authoring contract:

| | Before | After |
|---|---|---|
| Mapping discriminator | `BPMN.ScriptTask` | `BPMN.Variables` |
| Script return | `return { response: v }` | `return v` (v2+) |
| Reads inputs as | top-level ids (`amount`) | `vars.<varId>` |
| `args` body | `{"amount":"=vars.X"}` | `{"vars":"=vars","metadata":"=metadata"}` |
| Outputs | one business var | standard `scriptResponse` + typed `Error` |

Plus: brownfield preserve-don't-migrate rule, a bundled compatibility `xmlTemplate` in `bpmn-spec.json`, a new CI guard, and a rewritten Jint eval (+533/-76).

Claim under it: the registry's own template emits a discriminator the runtime ignores, so the script silently never executes while the process reports `Completed`.

Reviewed against the stack base `codex/bpmn-public-io-bridges-v2`, not `main`.

## 🔴 Overall findings

### 1. `debug/live_debug_e2e` grades the exact shape this PR declares broken 🔴 critical

Untouched by this PR:

- `check_live_debug.py:326` fails any bare `return` and demands `return { response: 6 * 7 };`
- `check_live_debug.py:321` fails **any** `vars.` read, which the new contract mandates
- `check_live_debug.py:313` requires `source="=result.response"` (compatible, but paired with the wrapper)
- `live_debug_e2e.yaml:56-58` pins `args` CDATA `{}` and the wrapper
- `live_debug_e2e.yaml:64` asserts that `{}` body "is part of the `BPMN.ScriptTask` registry template", which this PR says is the wrong template

An agent obeying the updated skill fails this eval. An agent obeying the prompt authors what the skill now calls non-functional.

This is the only in-repo eval that executes a real BPMN, and the PR description cites it as the motivation. Its predecessor (`6454019a7`, #2214) pinned that shape **after observing runtime nulls**. This PR reverses that conclusion without touching the eval or explaining the reversal.

**Fix:** update `check_live_debug.py` and `live_debug_e2e.yaml` here, or state in the description why the reversal leaves them valid.

### 2. Motivating evidence is not reproducible from this repo 🟠 major

The description's "classification task never runs, so severity is never computed (score 0.600)" has no counterpart in `tests/` (no BPMN eval mentions severity). `bpmn-spec.json:9-10` and `structural-bpmn.md:245` cite `s183-migrations.ts`, `xml-serialization.test.ts`, `ScriptTaskProperties.tsx:347`, and `actions.ts:352` in PO.Frontend, which nothing here can check and which will rot silently.

`test_bpmn_script_task_contract.py` only asserts that `bpmn-spec.json` matches itself. Nothing detects the bundled template drifting from the live serializer, which is the failure mode that produced this PR.

**Fix:** add a check that diffs the bundled template against `registry get` output when the CLI is available, and record the PO.Frontend commit SHA next to each `file:line` citation.

### 3. Six brownfield fixtures left in the shape the PR calls broken 🟠 major

Per the new rule at `structural-bpmn.md:260-266`, `v3` + `{ response: ... }` resolves to `result.response.response`. Fixtures with exactly that shape:

- `patterns/ai_decision_review/fixture/ClaimsIntake.bpmn:23-31,35-43`
- `edit/add_output/fixture/Invoicing.bpmn:22-30`
- `edit/move_node/fixture/Onboarding.bpmn:22-30`
- `edit/remove_node/fixture/ShipmentReview.bpmn:22-30`
- `edit/add_node/fixture/OrderIntake.bpmn:22-30`
- `edit/group_to_subflow/fixture/Fulfillment.bpmn:22-29`
- `hitl/quality_brownfield_insert/fixture/OrderFulfillmentBpmn.bpmn:17-26`

`edit/update_node/fixture/RiskScoring.bpmn:23` fixes one of the seven (v3 to v2). The rest stay, so the shape this PR calls broken remains the majority example in the suite. Agents read fixtures and imitate them.

### 4. `SKILL.md` pays 16 lines for a rule that lives in two reference files 🟡 minor

`SKILL.md:183-189` restates `structural-bpmn.md:216-219` almost verbatim, then links to it. Pick one location.

### 5. Not verified

Runtime behavior of the new contract. Every claim (v1 spreads keys, v2+ wraps under `response`, `/^v(\d+)$/` matching, canvas retyping) rests on the external repo. Internal consistency is all I could confirm.

## 🧭 Recommendation

**Request changes.**

The core insight looks right and the compatibility template is well specified. Blocking on finding 1: the contract changes while the repo's only executing BPMN eval asserts the opposite, so CI green here does not mean the skill and the suite agree. The two inline majors (skeleton vs. the new `double` rule, `bpmnElement` guidance wrong for 27 of 29 bundled templates) each break the "author from the skeleton and the template" instruction the skill leads with.

No AI slop of consequence in the prose. Dense but load-bearing, except the one line flagged inline.

## tl;dr

Right diagnosis, incomplete rollout. The new ScriptTask contract contradicts `live_debug_e2e` (still demands `return { response: ... }`, bans `vars.`), contradicts the canonical skeleton on `number` vs `double`, and the new `bpmnElement` advice is wrong for 27 of 29 bundled templates. Six brownfield fixtures still carry the shape this PR calls broken; one was fixed.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Fixed — the five node-scoped and mapping positions are `double`, the two public declarations stay `number`. That surfaced a second problem: `test_bpmn_public_io_contract` asserted the two layers share a type string, which was only true while both said `number`. It now encodes the correspondence and still fails on a real mismatch.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Corrected to your wording. I generalised from the one template I had rewritten — 27 of 29 carry the PascalCase tag, so the advice was wrong for almost all of them and contradicted SKILL.md.

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
DevMomo: Taught rather than dropped: the Jint boundary now names all four with the reason (Jint provides them, so nothing fails locally, but a replayable run needs the same outputs from the same inputs), and the prompt names them explicitly.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Wrapped in `<uipath:variables version="v1">` with the parent marker, matching the style at :167.

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
DevMomo: Widened to `=vars\.[\w.-]+`, so a GUID-style id no longer fails as a missing bridge.

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
DevMomo: Stated the exact casing in the reference. The grader was enforcing something the skill never said.

tests/tasks/uipath-maestro-bpmn/nodes/check_script_jint_guidance.py
DevMomo: Fair — those three are noise in an already large diff. I left them rather than add another churn commit on top; happy to strip them if you would rather the diff be smaller.

tests/scripts/test_bpmn_script_task_contract.py
DevMomo: Changed to "every spec entry" so it cannot go stale.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Trimmed to the actionable half.

rockymadden (approved):

PR #2368, opened 2026-07-30
fix(tests): lift judge-score ceilings on eleven troubleshoot scenarios

dmorosanu:
Lifts judge-score ceilings on eleven troubleshoot scenarios. Two groups: three scenario-level
defects (commit 1) and eight playbook/fixture fixes (commit 2).

# Commit 1 — scenario defects

Scenarios whose judge score was capped by a defect in the scenario itself rather than by agent
quality.

## 1. `classic-movefile-source-not-found-de` — staged the missing project source

A German-locale clone of `classic-movefile-source-not-found` that dropped one `template_dir`
entry, so it staged no project source at all — while its `RESOLUTION.md` cites `move.xaml`
line 61 and verbatim `Path`/`Destination` attribute values, and its simulator tells the agent
"the project source is available in the current working directory". The agent was told the
source existed, could not find it, and was graded on facts only that source could supply.

Added `path: process` to `task.yaml` and copied the sibling's `process/` verbatim. The two
`task.yaml` files now differ only by `task_id` and the English variant's `skip: true`.

## 2. `sys-getasset-activity-silent-failure` — two unobservable discriminators

- The playbook's Context named only `Get Asset` / `Get Orchestrator Asset`, so the agent
  concluded the documented defect did not cover `Get Credential` and ruled it out. Context now
  names `Get Credential` (`GetRobotCredential`) explicitly.
- The playbook told the agent to verify the output variable was not created via `Ctrl+K`. Studio
  serializes a `Ctrl+K` variable and a Variables-panel variable identically, so that is not
  recoverable from the workflow source. New investigation step 4 says so and says not to exclude
  the cause on that basis.
- `RESOLUTION.md` asserted the `Ctrl+K` trigger as fact, while its own H1 evidence row cites only
  the NRE, the 22.10.5 pin and healthy asset/folder/role/license. The narrative now matches the
  evidence: the package version is the deciding signal.

## 3. `o365-getmail-invalid-odata-query` — equivalent fix not recognised

`RESOLUTION.md` now records that expressing the unread condition inline as `isRead eq false` is
equivalent to the built-in "Only Unread Messages" option (`isRead` is the real Graph property;
`unread` does not exist). The rubric's 1.0 band already allows "or equivalent".

# Commit 2 — playbook discriminators and fixture coverage

Eight fixes for playbooks (and one manifest) that left the agent unable to reach the conclusion the evidence supports. Each
change supplies a missing fact or removes an ambiguity; none edits a RESOLUTION to move the
goalposts, and each stands on its own as client guidance independent of any score.

## 4. `uia-type-into-method-not-found` — `MissingMethodException` was misrouted

`dependency-version-conflict.md` sent every `MissingMethodException` to Signature A, whose fix is
"update all foundational packages together" — the intra-project case. But the scenario is a
*same-package* skew between authoring and runtime: the project pins one
`UiPath.UIAutomation.Activities` version and the robot restored a different build, so the assembly
loads and only the member is missing. The tell for that case ("runs in Studio, faults on the
robot") was documented only under Signature B, which keys off a **different** exception. The agent
therefore hit the wrong branch and produced the wrong fix on every replicate.

Signature A is now split: **A1** (same package, different restored build per environment → pin one
exact version across all environments, republish) and **A2** (the project's own package set
disagrees → align the set). Three places where the existing Signature B text contradicted A1 were
repaired.

## 5. `sys-getqueueitem-queue-not-found` — wrong-folder branch could not be closed

`queue-operation-failed.md` now says to enumerate queues in the other folders the identity can
see, which is what positively confirms or excludes the wrong-execution-folder branch — "the job's
folder has no queues" leaves it open, "no visible folder holds this queue" closes it. The evidence
boundary is now stated as two separate claims (what the runtime 404 settles vs what stays open),
and each fix branch is named explicitly; "those three branches" was ambiguous after four preceding
bullets.

## 6. `mail-send-outlook-mail-work-offline` — branch 2 sub-causes were undifferentiated

`send-outlook-mail-failures.md` branch 2 listed three sub-causes with no discriminator. Added the
one the evidence turns on: an item that reaches the **Outbox** with **no dialog raised** confirms
Work Offline and *excludes* the Object Model Guard, because the Guard blocks before the item is
ever queued. An attended re-run on the host is the decisive test. Also surfaced the SMTP/Graph
migration inside branch 2 for unattended runs — the content existed only in an "Alternative"
section the agent never reached from there.

## 7. `terminal-session-connect-failed` — `ExistingConnection` was not excludable

`terminal-session-connection-failed.md` now states that the `ExistingConnection` branches apply
only to two specific messages, and that a session opening a new connection from a
`ConnectionString` cannot fail that way. Also: say explicitly that `System.AggregateException` is
only the async wrapper and the inner `TerminalConnectionException` is the cause.

## 8. `csv-read-file-not-found` — the decisive observation was never used as a discriminator

`read-csv-file-not-found.md` noted in passing that the file "may even exist when you check manually
afterward", but never said what that *proves*. A quoted-literal or drifted relative path resolves
somewhere the file never lands, so finding the file at the error's exact path afterward positively
**excludes** both path branches and confirms the async-download race. The playbook offered only
"intermittent vs every run", which needs run history; the presence test settles it from a single
failure. The agent was consequently leading with "relative path / working-directory drift" and
demoting the race to a secondary possibility — the judge's complaint in both nightlies.

## 9. `excel-rr-concurrent-jobs` — documented trigger command was unmocked

See the fixture note under Validation. `data/m/r/manifest.json` now also mocks
`or triggers list --folder-path "ExcelImports" --output json` against the existing triggers fixture.

## 10. `uia-napplicationcard-view-generation` — causes read as alternatives

`napplicationcard-view-generation-failed.md` now states the three causes are cumulative, not
alternatives. A stale `obj/project.assets.json` is the easiest to spot and must not end the
search; always also compare the package pin against `studioVersion`, because clearing the project
cache alone leaves a version skew in place.

## 11. `connector-null-reference` — guard was one option among four, never the fix

`connector-null-reference.md` told the agent to "check for null / empty before the `For Each`" — right
direction, but abstract and listed as one conditional fix among four (output mapping, stale binding,
null input). The agent consequently enumerated causes without committing to any, in every local
replicate. Yet the playbook's own Context already states the discriminator: the stack shows `For Each`
over a connector `*_List` output, which establishes both what was null and where it was dereferenced.
That is now explicit — when the stack carries that signature the guard **is** the fix, lead with it,
and give the concrete pattern (`If result IsNot Nothing AndAlso result.Count > 0`) rather than the
idea of one.

# Validation

Every delta below is a **paired local run**: same scenario, 3 repeats per arm, pre-fix and post-fix,
each arm executed against a frozen `git worktree` so neither can be contaminated mid-run. Agent
`codex` / `gpt-5.6-terra`, matching the nightly arm.

| Scenario | Pre-fix (3 runs) | Post-fix (3 runs) | Δ |
|---|---|---|---|
| `uia-type-into-method-not-found` | 0.850, 0.850, 0.850 → **0.850** | 1.000, 1.000, 1.000 → **1.000** | **+0.150** |
| `sys-getqueueitem-queue-not-found` | 0.880, 0.880, 0.880 → **0.880** | 1.000, 1.000, 1.000 → **1.000** | **+0.120** |
| `mail-send-outlook-mail-work-offline` | 0.925, 0.850, 0.850 → **0.875** | 0.925, 0.925, 1.000 → **0.950** | **+0.075** |
| `terminal-session-connect-failed` | 0.925, 0.925, 0.925 → **0.925** | 0.925, 1.000, 1.000 → **0.975** | +0.050 |
| `uia-napplicationcard-view-generation` | 0.850, 0.925, 0.850 → **0.875** | 0.925, 0.925, 0.925 → **0.925** | +0.050 |
| `connector-null-reference` | 0.925, 0.850, 0.850 → **0.875** | 1.000, 0.925, 1.000 → **0.975** | **+0.100** |
| `csv-read-file-not-found` (n=6/arm) | 0.775, 0.925, 0.850, 1.000, 0.850, 0.925 → **0.888** | 0.850, 1.000, 1.000, 1.000, 1.000, 0.925 → **0.963** | **+0.075** |
| `excel-rr-concurrent-jobs` | 0.850, 0.925, 0.925 → **0.900** | 0.925, 0.925, 0.925 → **0.925** | +0.025 (see below) |

`uia-type-into` and `sys-getqueueitem` show zero variance in both arms and no overlap between them —
the signature of a playbook that was misrouting the agent rather than merely being vague. The two
`+0.050` results are roughly 1.35× the noise floor at n=3; treat them as probable rather than
established.

`csv-read-file-not-found` was run at **6 replicates per arm** because its per-run spread (0.775–1.000)
is wide enough that n=3 could not separate a fix from noise. Both ends moved: the floor rose
0.775 → 0.850 and the count of perfect runs went 1/6 → 4/6.

### `excel-rr-concurrent-jobs` — kept on correctness, not on score

Its +0.025 sits below the noise floor, so the score does **not** justify the change. It is included
because the underlying defect is independently verifiable rather than inferred: the playbook instructs
the agent to run `uip or triggers list --folder-path '<folder>'`, the manifest mocked only the
`--folder-key <guid>` form, and `unmocked_default` is null — so the documented invocation errors. The
sealed call log from the pre-fix arm shows exactly that:

```
{"args": "or triggers list --folder-path ExcelImports --output json",
 "matched_rule": null, "fixture": null, "exit_code": 1, "error": "unmocked"}
```

The fixture holding the decisive evidence (triggers `HourlySalesA` / `HourlySalesB`, cron
`0 0/30 * * * ?`) was unreachable through the documented command, which is why the judge recorded
"could not retrieve trigger definitions" in both nightlies. The added rules point the `--folder-path`
form at that same existing fixture — coverage of an invocation the agent actually makes, not new
evidence. Post-fix the call resolves (`exit_code: 0`), the minimum rose 0.850 → 0.925 and variance
collapsed to zero, consistent with removing a failure mode rather than lifting a ceiling.

## CI confirmation

Three full `run-coder-eval` passes on this branch (agent `codex` / `gpt-5.6-terra`, 8 tasks each,
24 data points): runs `30616937471`, `30616947593`, `30616957525`.

| Scenario | CI runs | CI mean | Nightly baseline |
|---|---|---|---|
| `uia-type-into-method-not-found` | 1.000, 1.000, 1.000 | **1.000** | 0.850 / 0.850 |
| `uia-napplicationcard-view-generation` | 1.000, 1.000, 1.000 | **1.000** | 0.850 / 0.850 |
| `mail-send-outlook-mail-work-offline` | 1.000, 1.000, 1.000 | **1.000** | 0.850 / 0.850 |
| `sys-getasset-activity-silent-failure` | 1.000, 1.000, 1.000 | **1.000** | 0.400 / 0.850 |
| `o365-getmail-invalid-odata-query` | 1.000, 1.000, 1.000 | **1.000** | 0.850 |
| `terminal-session-connect-failed` | 0.925, 1.000, 1.000 | **0.975** | 0.850 / 1.000 |
| `sys-getqueueitem-queue-not-found` | 0.880, 1.000, 1.000 | **0.960** | 0.880 / 1.000 |
| `classic-movefile-source-not-found-de` | 0.850, 0.925, 0.925 | **0.900** | 0.850 / 0.775 |

Five of eight score 1.000 on every CI replicate; none sits below its nightly baseline. CI has no
matched pre-fix arm, so it confirms the post-fix **level** rather than causation — the paired
worktree runs above remain the evidence for the deltas themselves. `uia-napplicationcard` and
`mail-send-outlook-mail` score higher in CI (1.000) than locally (0.925 / 0.950), consistent with the
environment-to-environment spread described below rather than with anything about the changes.

## Noise floor — why the pairing matters

Comparing the `2026-07-30` and `2026-07-31` nightlies (155 scenarios each, no relevant code change
between them):

- **43 of 155 (28%) changed score.**
- Mean absolute delta: **0.037**. That is the noise floor.
- 103 scenarios stable at 1.0 in both; 34 capped in exactly one nightly; only **18 capped in both**.
- The local environment also scores systematically higher than the nightly for several scenarios, so
  a nightly figure and a local figure are not comparable at any sample size.

A single nightly is therefore not a usable baseline, and only scenarios capped in **both** nightlies
were treated as real targets. Deltas below ~0.04 are indistinguishable from noise at n=3.

## Changes measured and dropped

Five further changes were written, measured against the same paired control, and reverted for
earning nothing: agents playbooks (`input-schema` **0.000**, `context-grounding` +0.025,
`llm-insufficient` +0.025), the download transport-branch split (`download-file-dns-failure`
**0.000** — already 1.000 pre-fix locally), two Excel surface-B attempts (**0.000**, then +0.025 on
one replicate of three), and a round of agent-path/`studioVersion` corrections (2 scenarios worse, 2
unchanged). They are absent from this PR.

Worth recording: the agent emitted a wrong-format command path in 8 of 9 replicates and the judge
flagged it every time, yet fixing it moved nothing. The judge narrates that deviation in its
findings without pricing it — the findings list does not distinguish what is scored from what is
merely observed.

## Scope note

The three commit-1 scenarios were validated before the paired methodology above was adopted, using
single-nightly baselines. The `movefile-de` change is a structural fixture repair (a missing
`template_dir` meant `process/` was never staged) and holds regardless of scoring. The CI table above
is the first measurement of all three under the current method: `sys-getasset-activity-silent-failure`
and `o365-getmail-invalid-odata-query` score 1.000 on every replicate (the former was 0.400 on the
2026-07-30 nightly), and `classic-movefile-source-not-found-de` averages 0.900. Those are levels, not
controlled deltas.


dmorosanu:
## CI confirmation - run 30542035310

`run-coder-eval.yml` on this branch, `agent: codex` (`gpt-5.6-terra`), parallelism 1. **3/3 succeeded.**

| Scenario | Nightly (6 runs) | Local | CI |
|---|---|---|---|
| `classic-movefile-source-not-found-de` | 0.80 x 6 | 1.00, 1.00, 0.90 | **0.90** |
| `sys-getasset-activity-silent-failure` | mean 0.75 | mean 0.85 (6 reps) | **1.00** |
| `o365-getmail-invalid-odata-query` | mean 0.80, never 1.00 in 6 | 1.00, 0.80, 0.80 | **1.00** |

All three CI results land inside the local spread and above the nightly baseline. CI runs one replicate per task, so each figure is a single sample.

## Second revert

`rpa-is-connection-missing-dap` was also fixed and then **reverted** - it regressed to 0.50/0.50/0.50 against a 0.80 baseline. All three replicates hedged on the root cause itself, presenting deletion as one of several possibilities rather than concluding it.

That makes two reverts with the same shape. Softening a RESOLUTION's assertions, even where the assertion genuinely outran its evidence, makes the agent under-commit generally, and the judge penalises under-commitment harder than over-specificity. On `invokevba` the deduction simply inverted at the same score.

The edits that worked did the opposite: they supplied a missing **fact** (stage the source the RESOLUTION cites; name `Get Credential` in the playbook's scope) rather than relaxing a requirement. Both moved substantially.

MarinRzv (approved):

PR #2369, opened 2026-07-30
fix(uipath-maestro-bpmn): bridge public process inputs and outputs

DevMomo:
## Problem

A process can compute every value correctly and expose **nothing**. Public inputs/outputs are a two-layer contract — a caller-facing declaration plus a mutable variable — and nothing bridges them automatically:

```xml
<uipath:inputOutput id="Var_Decision" name="decision" type="string" />
<uipath:output      id="out_Decision" name="decision" type="string" elementId="End_1" />
<!-- the task writes var="Var_Decision"; nothing ever writes out_Decision -->
```

`decision` reads back empty. `validate` exits 0 throughout — it never runs the engine, so it cannot see an orphaned output.

**Why the e2e needs this:** without the bridge its `severity` output reads back empty (score 0.600).

## Fix

- Declare the public `input`/`output` with `elementId` on its Start/End event, keep a mutable `inputOutput`, and **map between them** on that boundary event
- `validate` is structural preflight, not runtime proof — don't report `Valid` as business success
- Routes returning one result converge on a single completion EndEvent
- Diagnose entry for the empty-value-with-clean-validate symptom

## Verification

`tests/scripts/test_bpmn_public_io_contract.py` pins the bridge; the approval eval now asserts the wiring, not just the declarations.

**Also in this PR:** `.gitignore` gains `tests/tasks/**/seed.json`. A live `seed.json` carrying real Jira and Slack identifiers had been committed under this suite; it is written at pre_run by each task's `seed.py`, so it should never be tracked. Called out here because it is not about the bridge contract.


skills/uipath-maestro-bpmn/references/diagnose/references/failure-modes.md
```diff
@@ -20,6 +20,19 @@ Entry point inputs reference a start event through `elementId`, but the start ev
 `uipath:entryPointId` or the ID is duplicated.
 Fix root start event extensions and variable scoping.
 
+## Public variables are not bridged
+
+The runtime binds caller inputs to public declarations and returns caller
+outputs from public output declarations; process logic reads and writes mutable
+internal variables. If the StartEvent and completion EndEvent do not bridge
+those layers, a run can report `Completed` while gateways take fallback paths
+or returned outputs are null.
```
DevMomo: @codex Rephrase pls. Use uipath terminology, not your slop.

rockymadden (approved):
With the items we called out addressed, looks good to me.

skills/uipath-maestro-bpmn/references/diagnose/references/failure-modes.md
DevMomo: Addressed in 55321d39a. The section now uses the exact BPMN variable directions and event mappings: input to inputOutput on the root StartEvent, then inputOutput to output on the returning root EndEvent.

DevMomo:
@jiyangzh @nikhil-maryala All review feedback on this second dogfooding layer is addressed and its focused/CI contract tests pass; only the smoke job is finishing. Could one of you give the required code-owner approval ASAP for the atomic #2363 → #2369 merge?

DevMomo:
@uipreliga @bai-uipath @akshaylive Correction: this PR also updates `.github/workflows/test-helpers.yml`, so one of you is the required code owner for the last merge gate. The workflow change only registers the focused public-I/O contract guard, which is passing. Could one of you approve ASAP for today’s dogfooding cut?

DevMomo:
@rockymadden The requested fixes are now on the current heads of #2363 and #2369. GitHub dismissed the earlier approvals after those final pushes, so I re-requested review on both current heads for the atomic stack merge.

DevMomo:
@rockymadden @nikhil-maryala @uipreliga @bai-uipath @akshaylive The two dogfooding layers (#2363 → #2369) are now rebased unchanged onto the latest `main`; GitHub dismissed prior approvals because the heads were rewritten. Please re-review the current heads so the normal atomic stack merge can proceed once CI is green.

nikhil-maryala (commented):
## Review — public I/O bridge contract

Direction is sound and the scaffold satisfies its own new guard (ran it locally: 2/2 pass). Three things worth addressing:

**🟠 No runtime evidence for a runtime claim.** The thesis (missing bridges → null conditions / null outputs at runtime) is a runtime-semantics assertion, but every check here is structural: contract guards parse XML shape, the eval runs `validate` + structural Python, and this PR itself now (correctly) documents that `validate` is *"a structural preflight, not runtime proof."* By that standard the bridge *shape* being locked in — including a root StartEvent reading its caller input via `source="=vars.<public-input-id>"` — is unverified against a real engine. Suggest softening the Verification section, or adding one authorized `debug-instance variables-all` run showing a non-null output reaching the caller.

**🟡 Eval guard weaker than the contract guard.** `check_simple_approval_bpmn.py` never asserts the mutable `inputOutput` lacks `elementId`, while `test_bpmn_public_io_contract.py` does (`assert "elementId" not in mutable.attrib`). An agent can element-scope the mutable var — which per `expression-authoring.md` means it is NOT surfaced as a root runtime variable, breaking the bridge — and still pass the eval. Add an `elementId`-absent check to the internal input and output vars.

**🟡 Grader stricter than the prompt.** The check now requires exactly one endEvent (forcing convergence), but the prompt never tells the agent to converge on a single completion event. A reasonable two-end-event (approved/rejected) solution fails with a structural error the prompt didn't foreshadow. Add one prompt line about returning `decision` on a single converged completion end event.

**Nits:** `_minimal_example` grabs the first ```xml block after the heading — fragile if another snippet is inserted above the scaffold; prompt dropped explicit type guidance while the grader still hard-requires `amount` type="double"; body commit refs (79bb543e3 / 28dbc5f56) don't match the actual head 37521b1a7.

Good: the `check_script_jint_guidance.py` collision fix (same-named public + mutable vars) is necessary and order-independent; scaffold dataflow `input_Var_Amount → Var_Amount → Var_Echo → output_Var_Echo` is complete; new CI job is wired correctly.

DevMomo:
Addressed the review in 36df02f2c:

- softened Verification to structural evidence only and explicitly stated that no live instance/debug run was performed;
- made the eval checker reject `elementId` on mutable input and output bridge variables;
- made the prompt explicit about the three variable types and one converged completion end event returning `decision`;
- changed the documentation guard to select the unique complete BPMN document structurally instead of taking the first XML block; and
- removed stale commit hashes from the PR body.

Focused scaffold/public-I/O guards pass 5/5; YAML parsing, Python compilation, and diff checks also pass.

nikhil-maryala (approved):

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -89,29 +90,50 @@ reason authoring runs out of time.
     exporter="UiPath (https://bpmn.uipath.com)" exporterVersion="1.0">
   <bpmn:process id="Process_1">
     <bpmn:extensionElements>
-      <uipath:variables version="v1" />
+      <uipath:variables version="v1">
+        <uipath:input id="input_Var_Amount" name="Amount" type="double" elementId="Start_1" />
+        <uipath:inputOutput id="Var_Amount" name="Amount" type="double" />
+        <uipath:output id="output_Var_Echo" name="Echo" type="double" elementId="End_1" />
+        <uipath:inputOutput id="Var_Echo" name="Echo" type="double" />
```
rockymadden: **CRITICAL** — `type="double"` on the public `input`/`output` makes this example unpackable. Verbatim copy of this block into a project:

```
$ uip maestro bpmn refresh .
"Result": "Failure"
"Instructions": "Unsupported process input/output type \"double\". Use
  string, boolean, integer, number, array, object, or json, or add an
  inline JSON schema."
```

`validate` reports `Valid` throughout — the exact trap this PR is about.

The constraint applies only to root `uipath:input`/`uipath:output`; `double` on `inputOutput` is fine because it never reaches schema derivation. But your own guard asserts `mutable.attrib["type"] == public_input.attrib["type"]`, so fixing only lines 94 and 96 breaks the guard.

Change all of them (94-97 plus the mapping `type` attributes on 106, 115, 125) to `number`. I confirmed that yields `refresh` Success, `validate` Valid, `input: {"Amount": {"type": "number"}}`, `output: {"Echo": {"type": "number"}}`.

Worth a line in the prose too: canvas float types (`double`, `float`) are not usable on a public declaration — both fail `refresh`, both pass `validate`.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
```diff
@@ -158,10 +216,61 @@ def main() -> None:
...
+                f"start event must bridge public {input_name!r} "
+                "to its mutable process variable"
+            )
+        internal_input = variable_by_id(variables, attr(bridges[0], "var"))
+        if local_name(internal_input) != "inputOutput":
+            fail(f"{input_name!r} start bridge must target uipath:inputOutput")
+        if public_input.attrib.get("type") != internal_input.attrib.get("type"):
+            fail(f"{input_name!r} public and mutable variable types must match")
+        if input_name == "amount" and public_input.attrib.get("type") != "double":
+            fail("numeric public and mutable amount variables must use type='double'")
```
rockymadden: **CRITICAL** — this requires the shape that can't be packaged.

`double` on a public `uipath:input` makes `refresh` fail hard (see the comment on `structural-bpmn.md:94`). Meanwhile #2363's assertions in this same file require `operate.json`, `entry-points.json`, and `package-descriptor.json` to exist beside the BPMN.

Both are only satisfiable together if the agent leaves `init`'s stale metadata untouched. An agent that does the right thing and regenerates gets a hard failure; one that skips regeneration passes with an unpackageable project. **The eval currently rewards not regenerating.**

Switch to `number` (or `integer`) and keep the type-equality assertion below — that one is good.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
```diff
@@ -14,27 +14,27 @@ run_limits:
   max_turns: 90
 initial_prompt: |
   Create a local BPMN project named "ExpenseApprovalBpmn" for this process:
-  - Manual start with input variable `expenseId` and `amount`, and output
-    variable `decision`.
+  - Manual start with caller inputs `expenseId` (string) and `amount` (double),
```
rockymadden: **CRITICAL (same root cause)** — the prompt now mandates `amount` (double), so an agent that follows it exactly produces a project `refresh` refuses to regenerate.

`(number)` or `(integer)` instead. Being explicit about types in the prompt is an improvement — it's just the one value that doesn't survive entry-point derivation.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/seed.json
```diff
@@ -0,0 +1,17 @@
...
+  "tag": "e9cb961a",
+  "correlationId": "ESC-BPMN-e9cb961a",
+  "inputs": {
+    "customerTier": "Enterprise",
+    "serviceState": "Unavailable",
+    "workaroundAvailable": false,
+    "correlationId": "ESC-BPMN-e9cb961a",
+    "jiraProjectKey": "CE",
+    "jiraIssueTypeId": "11457",
+    "slackChannelId": "C01H4SPS77W"
```
rockymadden: **MAJOR** — real-looking infra identifiers in a public repo, in an orphaned fixture.

`UiPath/skills` is public. `C01H4SPS77W` is exact Slack channel-ID format, and `jiraProjectKey: "CE"` / `jiraIssueTypeId: "11457"` read as live values. If they are, they shouldn't be committed — and it contradicts the placeholder rule the skill enforces on agent output via `require_no_private_connector_values`.

Two more problems with the file:

- **Against the convention.** `tests/README.md:288` — `seed.json` is written at runtime by `tests/tasks/uipath-platform/seed.py` with a fresh `uuid8`. Hardcoding `tag: e9cb961a` defeats that; repeat runs collide on the same `correlationId` in real Jira/Slack.
- **Orphaned.** This directory contains only this file. `severity` / `Sev1` appear nowhere else in the repo — the real task is `multi_node/customer_escalation/`, which never mentions severity. No yaml, no checker reads it.

Simplest fix: drop it. If the e2e is coming, land the fixture with its task and source the IDs from env/secrets.

skills/uipath-maestro-bpmn/references/expression-authoring.md
```diff
@@ -89,21 +89,19 @@ Do not use assignment operators in these fields. Comparisons such as `==`,
...
-  as a root/global runtime variable. Preserve exact variable ids: if the
-  requested variable id is `product`, declare `id="product"` and map to
-  `var="product"`, not `Product` or `Var_Product`. Current BPMN live debug may
-  expose the root output definition while still returning the value as `null`;
-  treat that as a runtime/debug API limitation, not proof that the authored
-  mapping is absent.
+- Keep a mutable root `uipath:inputOutput` for each value used by decisions,
+  tasks, or diagnostics. Process expressions reference that mutable variable
+  as `vars.<id>`, not the caller-facing declaration.
+- Public caller inputs and outputs use the event bridge contract documented in
```
rockymadden, line 102: **MINOR** — this rewrite dropped a false-alarm guard worth keeping:

> Current BPMN live debug may expose the root output definition while still returning the value as `null`; treat that as a runtime/debug API limitation, not proof that the authored mapping is absent.

That caveat covers the exact symptom the new diagnose entry teaches agents to hunt. Without it, an agent seeing `null` in `debug-instance variables-all` will go rebuild a bridge that may already be correct. Keep it alongside the bridge contract rather than instead of it.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
```diff
@@ -22,13 +22,67 @@
...
-        for var in process.findall("bpmn:extensionElements/uipath:variables/*", NS)
-        for identifier in (var.attrib.get("id"), var.attrib.get("name"))
-        if identifier
-    }
+def local_name(element: ET.Element) -> str:
+    return element.tag.rsplit("}", 1)[-1]
+
+
+def variable(
+    variables: list[ET.Element],
```
rockymadden, line 30: **MINOR** — `test_simple_approval_bpmn.py` was deleted with `root_variable_identifiers`, and its three unit tests weren't replaced.

`variable`, `variable_by_id`, and `variables_mapping` carry real branching (the `len(matches) != 1` paths especially), `pytest-bpmn-check` is a live CI job, and these now gate the eval's whole variable contract. Worth a small replacement file.

tests/scripts/test_bpmn_public_io_contract.py
```diff
@@ -0,0 +1,79 @@
...
+        assert mutable.attrib["type"] == public_input.attrib["type"]
+
+
+def test_minimal_example_bridges_mutable_state_to_public_output() -> None:
+    root = minimal_example()
+    process = root.find("bpmn:process", NS)
+    assert process is not None
+
+    ends = process.findall("bpmn:endEvent", NS)
+    assert len(ends) == 1
```
rockymadden, line 62: **MINOR** — fine as a pin on *this example*, but don't let it become the general rule. A process with an error or terminate end event legitimately has more than one `endEvent`, and `structural-bpmn.md` gets this right: converge routes *only when they must return the same result*.

A comment saying "the canonical example has exactly one completion event" would keep the next reader from generalizing it.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -143,8 +165,28 @@ If a migration marker is present, its supported shape is
...
+  routes on that completion event.
+
+Do not route directly on a public input declaration or assume an internal
+variable automatically becomes an entry-point output. A deployment can
+complete while downstream decisions see empty values or the caller receives
+null outputs.
+
 See [expression-authoring.md](expression-authoring.md) for expression rules.
-Sub-process-scoped variables go in that sub-process's own `<uipath:variables>`.
+Sub-process-scoped variables go in that sub-process's own
```
rockymadden: **NIT** — this line got reflowed across two lines with no content change. Pure diff noise in an otherwise tight change.

rockymadden (changes requested):
The diagnosis here is correct and the bridge design is right — I verified the mechanism works: with the mappings in place `refresh` derives both schemas properly, and dropping `elementId` from the public input empties the input schema. Good catch on a real bug.

Blocking on one thing: **the canonical example uses `type="double"` on its public input/output, and `uip maestro bpmn refresh` rejects that outright.** I extracted the XML verbatim from `structural-bpmn.md` into a real project:

```
$ uip maestro bpmn refresh .
"Result": "Failure"
"Instructions": "Unsupported process input/output type \"double\".
  Use string, boolean, integer, number, array, object, or json,
  or add an inline JSON schema."
```

| type | `validate` | `refresh` |
|---|---|---|
| `double` | Valid | **Failure** |
| `float` | Valid | **Failure** |
| `number` | Valid | Success |
| `integer` | Valid | Success |
| `string` | Valid | Success |

Canvas accepts the float types; the CLI's entry-point schema derivation doesn't. It only bites root `input`/`output` — `double` on `inputOutput` never reaches derivation. Global `double` -> `number` in the example fixes it (refresh Success, validate Valid, both schemas correct); the eval's `type='double'` assertion needs the same change. Details inline.

The deprecated path is worse than failing — `update-metadata` silently degrades it:

```
entry-points.json -> input:  {"Amount": {"type": "object"}}
                     output: {"Amount": {"type": "object"}, "Echo": {"type": "object"}}
```

`double` becomes `object` and the mutable `Amount` leaks into the output schema. That is exactly the clean-validate/wrong-runtime class this PR documents, landing in the PR's own fixtures.

**Also verified and fine:** the bridge contract itself (input schema populates only with `elementId` on the StartEvent, output only with `elementId` on the returning EndEvent), all 5 contract guards pass, all 9 bpmn checker unit tests pass, verb gate clean locally (0 High / 0 Medium).

**Two notes not worth an inline comment:**
- The red X is a *cancelled* run, not a failure — `matrix.skill` and `Detect changed skills` on `51967b5c7` are both `completed/cancelled` (superseded). Needs a re-run to go green.
- 36df02f2c made the checker reject `elementId` on mutable bridge variables and `51967b5c7` reverted it. The revert is right — trigger-bound values are conventionally `inputOutput` scoped with `elementId`, per that file's own last bullet — but a reader of the stack won't know why. One line in the body would cover it.

Slop check: clean. Real insight, concrete failing artifact, and `bpmn_doc_example.py`'s docstring explains an actual constraint.

My prior approval on `28dbc5f56` was dismissed by the rebase and isn't in the current commit list, so this needed a fresh pass regardless.

rockymadden (approved):

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Confirmed against `maestro-sdk/src/bpmn-packaging/entry-points.ts:162-181` — `double` is not in the switch. All 7 `type="double"` in this example are now `number`, and the Variables section states the trap: `validate` says Valid, `refresh` fails, and the restriction is specific to public declarations.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
DevMomo: Switched to `number`; the type-equality assertion stays. You are right that the two requirements were only jointly satisfiable by not regenerating — that is the worse half of the bug.

tests/tasks/uipath-maestro-bpmn/author/simple_approval_bpmn.yaml
DevMomo: `(number)`.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/seed.json
DevMomo: Dropped, and gitignored `tests/tasks/**/seed.json`. It is written by `seed.py` at pre_run; the committed copy was a stale local run that got swept in.

skills/uipath-maestro-bpmn/references/expression-authoring.md
DevMomo: Restored alongside the bridge contract.

tests/tasks/uipath-maestro-bpmn/author/check_simple_approval_bpmn.py
DevMomo: Added `test_simple_approval_bpmn.py` — 10 tests over the three helpers, including every not-exactly-one branch.

tests/scripts/test_bpmn_public_io_contract.py
DevMomo: Added that comment.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Reverted.

tests/scripts/test_bpmn_public_io_contract.py
```diff
@@ -0,0 +1,81 @@
+"""Guard the minimal public process I/O example in the BPMN skill."""
```
rockymadden, line 1: 🔴 **critical** — this guard never runs.

It appears in no workflow and in no `docs/REQUIRED-CHECKS.md` row; `test-helpers.yml` has no diff in this PR. The `bpmn-contract-guards` job enumerates its guards:

```yaml
run: pytest tests/scripts/test_bpmn_doc_examples_parse.py tests/scripts/test_bpmn_project_scaffold_contract.py -v
```

and the comment directly above that job says *"One job for every BPMN contract guard, listed explicitly… **Add new guards to the run step.**"*

So the regression this file exists to catch can land green, and the PR body says the opposite — "pins the bridge".

**Fix** — one line in `.github/workflows/test-helpers.yml`:

```yaml
run: pytest tests/scripts/test_bpmn_doc_examples_parse.py tests/scripts/test_bpmn_project_scaffold_contract.py tests/scripts/test_bpmn_public_io_contract.py -v
```

No `docs/REQUIRED-CHECKS.md` change needed — the job name is unchanged, so the existing `maestro-bpmn contract guards` row still covers it.

.gitignore
```diff
@@ -92,3 +92,6 @@ tests/tasks/uipath-maestro-flow/canary/Canary/Canary.uipx
 # Node (package.json lives at the repo root for the @uipath/skills package)
 node_modules/
 package-lock.json
+
+# Written at runtime by each e2e task's seed.py; never committed (carries live tenant ids)
+tests/tasks/**/seed.json
```
rockymadden, line 97: 🟡 **minor** — unrelated to public I/O bridges.

Defensible on its own: nothing matching is tracked today, and the comment is right that these carry live tenant ids. But it is scope creep in an otherwise focused PR, and the description does not mention it.

**Fix** — split into its own change, or add a line to the PR body so a reviewer is not left wondering which finding prompted it.

skills/uipath-maestro-bpmn/references/expression-authoring.md
```diff
@@ -89,21 +89,22 @@ Do not use assignment operators in these fields. Comparisons such as `==`,
...
-- An output variable you intend to expose in runtime inspection (via
-  `debug-instance variables-all` or `instance variables`) must be root-scoped —
-  declare its `uipath:output` or `uipath:inputOutput` WITHOUT an `elementId`. A
-  variable scoped with `elementId` is bound to that element and is not surfaced
-  as a root/global runtime variable. Preserve exact variable ids: if the
-  requested variable id is `product`, declare `id="product"` and map to
-  `var="product"`, not `Product` or `Var_Product`. Current BPMN live debug may
-  expose the root output definition while still returning the value as `null`;
-  treat that as a runtime/debug API limitation, not proof that the authored
+- Keep a mutable root `uipath:inputOutput` for each value used by decisions,
```
rockymadden, line 96: 🟡 **minor** — a runtime-inspection rule was dropped with no replacement.

The removed text said:

> A variable scoped with `elementId` is bound to that element and is not surfaced as a root/global runtime variable.

Nothing in the skill says that any more (grep confirms). Pair it with `0fb0dc60d`, which deliberately stopped asserting the mutable bridge variable has no `elementId`, and an agent can now scope the mutable variable to an element and lose the inspectability the deleted sentence protected.

**Fix** — keep one sentence: the mutable bridge variable stays root-scoped (no `elementId`) so `instance variables` / `debug-instance variables-all` can surface it.

**I could not verify this** — it is a runtime-inspection claim and needs a live debug session. Flagging rather than asserting.

rockymadden (changes requested):
## 📋 Review

### What

A public input/output is a two-layer contract — a caller-facing declaration bound to a boundary event, plus a mutable process variable. Nothing bridges them automatically, so a process can compute correctly and return nothing. This PR documents the bridge, adds a guard, and tightens the approval eval to assert wiring rather than declarations.

---

### ✅ Verified against the CLI (1.203.0)

I reproduced the defect rather than taking the description on faith. Took the doc's own skeleton, removed **only** the `End_1` bridge mapping, left the public output declaration in place:

| | Result |
|---|---|
| `refresh` | `Success` |
| `validate` | **`Valid`** |
| `entry-points.json` output schema | still advertises `{"Echo": {"type": "number"}}` |

The package advertises a caller-facing output that nothing writes, and `validate` cannot see it. **The premise is real and the guard is warranted.** With the bridge intact, both schemas populate (`Amount` in, `Echo` out).

Also confirmed:

- `check_script_jint_guidance.py` — correct. Under the bridge pattern a business name legitimately appears on both a public declaration and a mutable `inputOutput`; preferring the mutable id regardless of declaration order is the right resolution.
- Prompt and checker agree on the single-completion-end-event contract.
- The guard's `len(ends) == 1` comment correctly notes it pins the canonical example, not the general rule.
- `test_simple_approval_bpmn.py` deliberately covers the not-exactly-one failure branches, not just happy paths.

---

### 🔴 Overall findings

**1. The guard never runs — critical.**

`tests/scripts/test_bpmn_public_io_contract.py` is referenced **nowhere** in `.github/workflows/` or `docs/REQUIRED-CHECKS.md`. `test-helpers.yml` has no diff in this PR at all.

The `bpmn-contract-guards` job lists its guards explicitly:

```yaml
run: pytest tests/scripts/test_bpmn_doc_examples_parse.py tests/scripts/test_bpmn_project_scaffold_contract.py -v
```

and its own comment, three lines above, says: *"One job for every BPMN contract guard, listed explicitly… **Add new guards to the run step.**"*

The Verification section says this file "pins the bridge." It pins nothing — the regression it exists to prevent can land green.

**Fix:** append `tests/scripts/test_bpmn_public_io_contract.py` to that run step. One line.

*(The earlier standalone `bpmn-public-io-contract-guard` job was removed in `0fb0dc60d` and left no dangling required check — that part is clean.)*

**2. No passing-run claim — minor.**

The repo's own lint flags this High. The eval's contract changed materially; nothing records that it was run.

---

### ⚠️ What I did not verify

- The **runtime** half of the premise — that an unbridged output reads back empty at execution. I confirmed `validate` is blind to it and the schema is advertised; proving the null return needs a cloud run.
- I did **not** execute pytest (none installed here). I read `conftest.py`'s `sys.path` insert and traced that pytest loads it for `tests/scripts/` collection, so the shared `bpmn_doc_example` import resolves — reasoned, not run. I did execute all 9 guard functions directly via a stub harness while resolving this branch's conflicts: all passed.

---

### 🎯 Recommendation

**Request changes** — on finding 1 alone. Everything else is minor.

The engineering is good: the defect is real, I reproduced it, the guard is well-scoped, and the unit test covers its failure branches. But a contract guard that no job runs is decoration, and the body currently claims otherwise.

---

**tl;dr** — Premise verified: strip the bridge and `validate` still says `Valid` while `entry-points.json` advertises an output nothing writes. Good fix, good guard. One blocker: the guard isn't in any CI job, so it never runs, and the PR body says it does. One-line fix to `test-helpers.yml`.

skills/uipath-maestro-bpmn/references/expression-authoring.md
rockymadden: Retracting this — it is wrong, and the answer was already in `0fb0dc60d`'s commit message, which I did not read before filing.

`elementId` is additive, not exclusive. Confirmed in the source myself (`VariableUtil.ts`, the `inputOutputVariables` reducer):

```ts
acc.inputOutputVariables[inputOutput.id] = { ...inputOutput };   // unconditional
if (inputOutput.elementId) {
  ...
  acc.activityToVariables[inputOutput.elementId] = { ... };      // additive index only
}
```

The global-map assignment is outside the `elementId` branch, so scoping never removes a variable from it. The sentence I wanted restored was the stale claim you deliberately deleted, and restoring it would have reintroduced a documented falsehood.

No change needed here. The `.gitignore` note stands as presentation-only; the CI finding stands and I am pushing the one-line fix.

DevMomo:
Both findings addressed.

**1. The guard never ran** — correct, and it is the second time this PR shape has bitten: the standalone job was folded into `bpmn-contract-guards` and the guard was never added to the consolidated run step. Appended, and verified it actually guards — deleting the public output declaration from the canonical example fails it.

**2. Passing-run claim** — the escalation e2e that exercises this bridge end-to-end ran clean on Alpha `codereval` at **1.000 / 543s**, with the public outputs asserted by id in the root scope and the Jira and Slack records re-read from the tenant. Rebased onto merged `main` (#2363) as part of this push.

On what you did not verify: the runtime half is what that e2e covers — an unbridged output reads back `null` from `debug-instance variables-all`, which is how the defect was found.

rockymadden (approved):

tests/scripts/test_bpmn_public_io_contract.py
DevMomo: Already fixed — the push landed just after your review. `test-helpers.yml` now diffs against main with exactly the line you suggested, and I verified it guards: deleting the public output declaration from the canonical example fails the run. Agreed on REQUIRED-CHECKS.md — the job name is unchanged, so the existing row still covers it.

.gitignore
DevMomo: Added to the PR body. It came from a live `seed.json` with real Jira and Slack ids having been committed under this suite; the file is generated at pre_run, so it should never be tracked. Left in place rather than split, but no longer unexplained.

skills/uipath-maestro-bpmn/references/expression-authoring.md
DevMomo: Restored as one sentence with the reason: the mutable bridge variable stays root-scoped, because an element-scoped variable is not surfaced as a root runtime variable and `debug-instance variables-all` cannot show it. That read is exactly what the escalation grader depends on, which is the concrete cost you were pointing at.

PR #2370, opened 2026-07-30
fix(uipath-maestro-bpmn): separate portable and live resource resolution

DevMomo:
## Summary

- Separate **portable draft** intent from **live/runnable** intent. A known built-in draft uses only `registry get`; `RequiresDiscovery` describes runnable completeness and no longer grants permission to inventory a tenant.
- Verify the requested CLI context before live discovery. Named profiles are non-sticky and every copyable live registry, queue, connection, and refresh example repeats `--profile <name>`.
- Route live lookup from the retrieved full registry contract: `BindingInfo.Resource` for process/queue resources and `BindingPattern: connection` for connection identity.
- Resolve exact names, types, folders, lifecycle state, and keys; preserve ambiguity; refresh only the affected source once; never take the first row or fabricate an identifier.
- Keep connection identity separate from operation/object/schema authoring. A connection ID alone cannot make a connector node runnable.

## Why

The prior guidance could interpret registry metadata as permission for tenant discovery and could treat a named profile as if one status call switched subsequent commands. The first correction explained that profiles are per-command, but several copyable live examples still omitted the profile placeholder; the eval proved that agents copied those examples selectively. The final guidance makes the prose and every live command form consistent while leaving portable built-in examples explicitly profile-free.

This is product-wide and environment-neutral. It contains no environment, organization, tenant, Alpha, or connector-provider assumptions. The current adapters cover the resource contracts the CLI exposes today—process, queue, and connection—and unknown contracts block only the runnable claim instead of forcing a provider-specific guess.

## Stack and independence

- Native GitHub stack **#2663**: [#2662](https://github.com/UiPath/skills/pull/2662) BPMN DI before validation → [#2655](https://github.com/UiPath/skills/pull/2655) portable registry templates → [#2363](https://github.com/UiPath/skills/pull/2363) → [#2369](https://github.com/UiPath/skills/pull/2369) → [#2366](https://github.com/UiPath/skills/pull/2366) → this PR.
- This PR is one atomic resource-resolution layer with two focused commits: the provider-neutral discovery contract/eval and a smoke-task update that requires explicit registry evidence. It is independently useful after the lower stack layers merge and does not include connector-operation authoring, runtime work, or the complex live eval.

## Verification

Ran `skill-bpmn-resource-resolution` locally with Claude Sonnet 5. The post-fix run against the frozen task and checker recorded **7/7, score `1.000`**.

Replacement current-head skill smoke suite: pending.

| Run | Duration | Assistant turns | Commands | Cost | Recorded result | What it exposed |
| --- | ---: | ---: | ---: | ---: | --- | --- |
| `2026-07-30_06-33-25` | 130.460s | 33 | 20 | $0.655 | 5/7, `0.548` | Harness defect: a correct named-profile queue command missed an option-order-sensitive fixture rule; the agent also inspected the wrapper once. |
| `2026-07-30_06-36-48` | 120.768s | 33 | 19 | $0.619 | 6/7, `0.613` | Queue fixture and no-inspection behavior were fixed. Semantic replay then exposed a real profile-propagation gap. |
| `2026-07-30_06-40-07` | 133.924s | 34 | 19 | $0.772 | 6/7 recorded; final-checker replay passes | Profile propagation and all resource behavior passed; the recorded miss was only the equivalent status spelling `draft-non-runnable`. |
| `2026-07-30_06-43-18` | 162.650s | 36 | 22 | $0.737 | 6/7 recorded; final-checker replay passes | Reconfirmed every substantive behavior; the recorded miss was only the equivalent spelling `non_runnable_draft`. |
| `2026-07-30_06-49-53` | 261.683s | 36 | 21 | $0.747 | 6/7, `0.613` | Frozen checker caught the remaining substantive inconsistency: three live `registry get` calls omitted the requested profile because copied examples still omitted it. |
| `2026-07-30_06-57-30` | 122.760s | 27 | 16 | $0.552 | **7/7, `1.000`** | After making every live example consistent, all profile, ordering, identity, ambiguity, refresh, and safety assertions passed. |

Earlier implementation baselines, before adding the portable-vs-live and named-profile coverage, also recorded 7/7 in 139.5s and 197.2s.

The frozen final checker requires:

- portable HITL contract retrieval before any login or tenant-dependent call, with no profile;
- verified base URL, organization, and tenant under the requested named profile;
- that same profile on every live pull/list/search/get, queue, and connection call;
- exact process type/folder/key and exact folder-scoped queue resolution;
- exactly one connection refresh, two enabled exact candidates, and no implicit selection;
- no connector mutation, operation invocation, or direct response-fixture inspection.

Static verification:

- Python compilation;
- YAML and JSON fixture parsing;
- `coder-eval plan`;
- task CLI-verb gate;
- `git diff --check`;
- provider/environment/profile-hardcoding scan.

### DIFF
--- a/skills/uipath-maestro-bpmn/SKILL.md
+++ b/skills/uipath-maestro-bpmn/SKILL.md
@@ -92,31 +92,53 @@ Two halves make a valid Maestro `.bpmn`:
 Work the four steps quickly, but keep the path matched to the user's ask. Treat
 requests to discover before authoring, save raw registry JSON/evidence, or "do
 not author yet" as discovery-only even if they describe an eventual BPMN. In
-that mode, immediately create `registry-evidence/`, run and save `registry pull
---output json`, `registry list --output json` or `registry search ... --output
-json`, and `registry get <type> --output json` for each requested type; do not
-read deep authoring references or scaffold a project. For authoring asks, author
-early: do not pre-read every reference before writing. Read a reference only
-when you reach the structure it covers, get the needed templates, then write the
-first complete draft before further spelunking. If
+that mode, immediately create `registry-evidence/`, but run only the discovery
+commands the user requested or the selected mode requires. By default, a known
+portable built-in needs only `registry get <type> --output json`; do not add an
+authenticated pull, list/search, or tenant-resource inventory on your own. If
+the user explicitly requests `registry pull`, `list`, or `search`—including as
+raw command evidence—run those requested read-only forms without a profile.
+That evidence request does not authorize tenant-resource inventory. Do not read
+deep authoring references or scaffold a project. For authoring asks, author early:
+do not pre-read every reference before writing. Read a reference only when you
+reach the structure it covers, get the needed templates, then write the first
+complete draft before further spelunking. If
 [references/structural-bpmn.md](references/structural-bpmn.md) or
 [references/expression-authoring.md](references/expression-authoring.md)
 directly covers the requested construct, write a first complete draft before
 further spelunking.
 
+Resolve live resources only when the user asks for a real, current,
+tenant-bound, or runnable node, or supplies a concrete resource to verify.
+`RequiresDiscovery` describes runnable-node completeness; it is not permission
+to inventory a tenant. For portable, synthetic, local-only, or structural
+drafts, keep placeholders unresolved and label the node non-runnable. If the
+mode is unclear and live discovery would materially change the result, ask
+first. In live mode, verify the active CLI context and follow
+[references/live-resource-resolution-guide.md](references/live-resource-resolution-guide.md).
+Use the default CLI context when the user selected it. Named-profile selection
+is per command: `login status --profile <name>` verifies that context but does
+not make later commands inherit it, so repeat the same `--profile <name>` on
+every tenant-dependent registry, queue, and connection command. Portable
+built-in lookups deliberately omit it.
+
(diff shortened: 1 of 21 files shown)


DevMomo:
Addressed the automated reference-pointer finding in `1b836c699`: login/context rules now have one authoritative home in `cli-conventions.md`; adapter routing, identity, lifecycle-state, ambiguity, and bounded refresh live in `registry-workflow.md`; top-level guidance links to those contracts. The post-deduplication provider-neutral eval still passes 7/7 at 1.000.

DevMomo:
Closing — cut during a stack audit as the one PR fixing a preference rather than a demonstrated failure.

Every other PR in this stack traces to a reproduced runtime defect. This one addresses agent *behaviour* — that a portable draft shouldn't trigger tenant inventory — which is a real concern but produced no broken artifact. The cost was disproportionate: a 140-line guide, +332 skill lines, and a 13-fixture mocked eval. It also deleted the skill's only connection-binding documentation and collided with main's #2808 on the same paragraph.

If we want the idea back, it's roughly fifteen lines in `registry-workflow.md`: `RequiresDiscovery` describes what a node needs to be runnable, not permission to inventory a tenant; in draft mode keep placeholders, label the node non-runnable, and ask before crossing into live discovery. That can ride along with a future PR rather than carrying its own guide and eval.

PR #1656, opened 2026-06-23
docs(uipath-maestro-flow): correct inline-agent flow-input wiring to the verified mechanism

jiyangzh:
**Fixes inline-agent flow-data wiring docs to the forms that actually work on `flow debug`.** The old docs taught models that pass build/refresh/validate but deliver empty/literal input — and null output — at debug.

## Input — the one correct shape (author all three, aligned)

- **flow node `agentInputVariables[]`:** `{ "id":"<key>", "binding":"=$vars.<trigger>.output.<var>" }` — use **`binding`, not `value`**. The converter and Studio Web's loader both read `binding`; a `value:"=js:…"` entry is ignored → empty input → fault `AGENT_RUNTIME.TERMINATION_LLM_RAISED_ERROR` ("Template placeholders detected instead of actual values").
- **`agent.json` `inputSchema.properties`:** one flattened `<trigger>__output__<var>` key per input (mandatory).
- **`agent.json` prompt:** `{{input.<trigger>__output__<var>}}` (the `input.` form) — never `{{ $vars.X }}`.

## Output — the symmetric shape (also three aligned pieces)

- **`agent.json` `outputSchema.properties`:** one typed key per field (e.g. `subject`, `body`).
- **flow node `agentOutputVariables[]`:** one entry per field (`[{id:"subject"},{id:"body"}]`) — **not** a single `content` object.
- **End node:** map each `out` global to the **flat** field path `=js:$vars.<node>.output.<field>` — typed fields surface flat at `output.<field>`; a `.content.` wrapper resolves to undefined → validate passes, debug Completes, but the flow output is **null**.

Also documents the `content`↔`contentTokens` byte-exact invariant with the validator error strings quoted verbatim from the CLI.

## Structure
`impl.md` (uipath-maestro-flow) is the single authoritative spec; `inline-in-flow.md` (uipath-agents) is a bridge that links to it.

## Scope
Docs only.

## Proof — reproduced the failing eval, now green

Verified end-to-end that this doc change fixes the `skill-flow-devcon-billing-resolution-writer` coder-eval:

1. A fresh agent followed **only this branch's skill docs**, on a **clean latest-`main` CLI**, to build the DevCon `BillingResolutionWriter` inline-agent flow from scratch — twice, independently.
[...]

PR #2379, opened 2026-07-30
feat(maestro-flow): add the inline-voice-agent plugin — voice agents in flow

JoshParkSJ:
## What

Adds an `inline-voice-agent` plugin to `uipath-maestro-flow`, plus registration rows and CODEOWNERS.

Teaches agents to build voice flows: a `uipath.agent.voice` inline conversational agent wired to a live call by `core.trigger.voice` (inbound) or `create-outgoing-call` (outbound), ended with `end-call`. Covers both topologies, prompt-input wiring, node JSON, and ship steps.

Verified on alpha with real calls. Depends on UiPath/cli#3305 (merged).

## Prompt inputs: the node binding is required

Worth a reviewer's attention, because the plugin got this backwards at first and the failure is invisible locally.

A voice agent reads flow data through the same four-piece contract as any inline agent — node `inputs.agentInputVariables[]` (delivery), the matching `agent.json` `inputSchema.properties` key (contract), `{{input.<key>}}` in `messages[].content` (resolution), and the trigger-associated global the binding resolves against. Voice is not special here: it shares the inline-agent converter path, and the voice job body (`isVoice` + `callContext`) coexists with `JobArguments` in the BPMN rather than replacing it.

The delivery binding is the piece only `flow pack` reads. Omit it and:

- `flow debug` still works — it back-fills the binding from `inputSchema` (UiPath/cli#2721)
- `flow pack` / `flow publish` ship `JobArguments: {"input":""}` — the deployed call reaches the model with its `{{input.…}}` tokens unfilled

So a flow tests clean locally and drops every input once published. An earlier revision of this plugin told agents never to author `agentInputVariables` on a voice node, which produced exactly that shape; the doc now states the full contract.

Also documented: a bare `$vars.…` written into `agent.json` prompt text is never interpolated. Studio Web compiles that form into `{{input.<key>}}` plus the binding when it writes the agent; nothing scans `agent.json` prompts, so via the CLI it reaches the model literally.

No CLI change is needed for this — pack reading the node binding is the documented contract, and Studio Web writes it too. That `flow debug` is more lenient than `flow pack` is a separate CLI inconsistency, filed rather than worked around here.

## Two constraints worth knowing

- Inbound flows cannot be `flow debug`'d — the only test loop is publish, bind a number, dial it
- Voice nodes cannot sit in subflows

## Tests

Two offline smoke tasks in `tests/tasks/uipath-maestro-flow/voice/`, run through coder-eval (`experiments/default.yaml`, run `2026-08-26_14-29-10`):

| Task | Covers | Result |
|---|---|---|
| `voice_outbound_call` | outbound topology, `callContext` into both nodes, the four-piece input contract | PASS (8/8, 1.000) |
| `voice_inbound_call` | inbound topology, no dial-out, the test-loop explanation | PASS on every deterministic criterion (7/8, 0.871) |

`voice_outbound_call` is the test behind the fix. Its `prompt-inputs` checker grades all four pieces aligned, and six failure modes are graded distinctly (missing delivery binding, `value` in place of `binding`, a key absent from `inputSchema`, a bare `$vars.…` in prompt text, no `{{input.…}}` token, and a binding whose global was never declared). Verified against fixtures for each.

It also carries `--min-inputs 2`. The prompt supplies a customer name **and** an amount, but the four-piece contract is satisfiable with a single wired input — without the count, the second value shipped ungraded and a half-wired flow still scored 1.000. The passing run resolves both (`start__output__customerName`, `start__output__outstandingAmount`).

This started as three tasks: a topology-only outbound build, an inbound build, and a separate prompt-inputs build. The topology-only outbound task asserted nothing the prompt-inputs one did not already assert — same validate, same three node types, same `callContext`-into-both, same `agent.json` `settings.voice`, only reweighted — so it cost a full ~26-turn agent run per smoke sweep for zero marginal coverage, on a suite where any maestro-flow source change runs the skill's full smoke set. The two are merged into the single `voice_outbound_call` above, which took the topology weights and the `--output json` advisory. The pair now splits by topology, inbound and outbound, rather than by topology versus wiring detail.

Local run note: `voice_inbound_call` scores 0.871 on my machine because its `llm_judge` criterion has no reachable judge transport (no `ANTHROPIC_API_KEY`, no Bedrock — the run logs `judge transport: none`). CI runs with `API_BACKEND: bedrock`, so it resolves there. Every file-based criterion on that task passed locally.

Both tasks are fully offline — no tenant, no login, no real call.


tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+
+
+def _nodes_of(flow: dict, node_type: str) -> list[dict]:
+    return [n for n in flow.get("nodes") or [] if n.get("type") == node_type]
+
+
+def _binding_expression(value: object) -> str | None:
+    """Extract the JS expression from a callContext binding, either shape."""
+    if isinstance(value, dict):
+        expression = value.get("expression")
```
rockymadden, line 148: **critical (for a grader)** — `_binding_expression` reads `expression` out of *any* dict and strips any `=…:` prefix off *any* string, so the `call-context` criterion passes bindings the plugin doc calls broken.

Verified locally, both score `0`:

```json
{ "type": "literal", "expression": "$vars.createOutgoingCall1.output.callContext", "fieldType": "string" }
"=jsonString:$vars.createOutgoingCall1.output.callContext"
```

`node-output-wiring.md:74` (added in this PR) says the binding is a `jsExpression` object, **"never a bare string"**, and the `to` field one section above in `impl.md` is a `type: "literal"` object, so copying the wrong `type` is the likely mistake, not a hypothetical one. A literal ships the expression text to the runtime and the call context never resolves.

Grade `type == "jsExpression"` on the dict shape, and drop the `=js:` string shape (or grade it as a warning) so the criterion matches the contract it documents.

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+    globals_ = ((flow.get("variables") or {}).get("globals")) or []
+    declared = {
+        (g.get("triggerNodeId"), g.get("id"))
+        for g in globals_
+        if isinstance(g, dict) and g.get("direction") == "in"
+    }
+    for key, binding in bound.items():
+        match = BINDING_RE.match(binding)
+        trigger, field = match.group(1), match.group(2)
+        if (trigger, field) not in declared:
```
rockymadden: **major** — this requires a *trigger-scoped* global for **every** bound input, including inputs bound from an upstream node's output. A correct flow fails.

Verified: a node binding `=$vars.formatAmount1.output.text` (amount formatted by a script node) scores `0` with

```
FAIL: 'formatAmount1__output__text' binds $vars.formatAmount1.output.text but no
variables.globals entry declares it (direction "in", triggerNodeId 'formatAmount1')
```

`inline-agent/impl.md` scopes the rule correctly: *"When the upstream node is a **trigger** … each field you bind must be declared in `variables.globals[]`"*. Node-output bindings need no global.

Gate the check on the source node actually being the flow's trigger (`nodes[].type` starting `core.trigger.` / `uipath.connector.trigger.`), and skip it otherwise.

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+            )
+            continue
+        bound[key] = binding.strip()
+
+    # Variable: a trigger-associated global must exist for each bound field.
+    globals_ = ((flow.get("variables") or {}).get("globals")) or []
+    declared = {
+        (g.get("triggerNodeId"), g.get("id"))
+        for g in globals_
+        if isinstance(g, dict) and g.get("direction") == "in"
```
rockymadden: **major** — `direction == "in"` rejects `inout` globals. Verified: a global `{"id":"text","direction":"inout","triggerNodeId":"start"}` bound on the node scores `0`.

`_shared/flow_check.py:read_flow_input_vars` is the repo's answer for the same question and accepts both:

```python
if v.get("direction") in ("in", "inout")
```

```suggestion
        if isinstance(g, dict) and g.get("direction") in ("in", "inout")
```

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+                parsed.append((path, json.load(handle)))
+        except (OSError, json.JSONDecodeError, UnicodeDecodeError) as err:
+            sys.exit(f"FAIL: could not read {path}: {err}")
+
+    for path, flow in parsed:
+        if any(n.get("type") == VOICE_AGENT for n in flow.get("nodes") or []):
+            print(f"Reading {path}")
+            return path, flow
+
+    seen = sorted({n.get("type") for _, f in parsed for n in f.get("nodes") or []})
```
rockymadden: **minor** — `sorted()` over a set that can hold `None` raises `TypeError: '<' not supported between instances of 'str' and 'NoneType'` when any node lacks a `type` key. Verified.

This is the diagnostic path for a malformed flow, so instead of the intended `FAIL: no uipath.agent.voice node …; node types seen: […]` line the criterion dies with a traceback and the grader loses the message it exists to print.

```suggestion
    seen = sorted({str(n.get("type")) for _, f in parsed for n in f.get("nodes") or []})
```

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+
+def _fail(msg: str) -> int:
+    print(f"FAIL: {msg}", file=sys.stderr)
+    return 1
+
+
+def _load_flow() -> tuple[str, dict]:
+    """Return the (path, parsed) `.flow` that holds the voice agent node."""
+    project_dir = find_project_dir()
+    paths = sorted(glob.glob(os.path.join(project_dir, "**/*.flow"), recursive=True))
```
rockymadden: **major** — recursive `**/*.flow` with no staging-dir exclusion. `.cli-stage` sorts *before* the real project dir (`sorted([".cli-stage/X/X.flow", "X/X.flow"])` → staged copy first), so once `flow debug` or `pack` has run, the loop below returns the staged flow and every downstream path (`os.path.dirname(flow_path)` → `agent.json`) resolves against the staging dir.

The repo already treats this as a hazard and excludes the same set in three places: `_shared/check_inline_agent.py` (`EXCLUDED_PARTS`), `_shared/check_bindings_no_stubs.py`, `_shared/advisory_flow_utils.py` — `.cli-stage`, `.v1stage`, `.agent-builder`, `_outputs`.

Filter `paths` through the same exclusion set.

skills/uipath-maestro-flow/references/shared/node-output-wiring.md
```diff
@@ -71,6 +71,7 @@ Three failure modes observed in agent-generated `.flow` files:
 | **Loop nodes** (`core.logic.loop`) | `inputs.collection` | **YES** |
 | **Subflow nodes** (`core.subflow`) | `inputs.<inputId>.source` | **YES** |
 | **Script nodes** (`core.action.script`) | `inputs.script` body — `$vars.*` is read inside JS, no `=js:` wrapping | **NO** — the body is already JS |
+| **Voice `callContext`** (`uipath.agent.voice`, `uipath.conversational.voice.end-call`) | `inputs.callContext` — a structured binding object `{ "type": "jsExpression", "expression": "$vars.<originNodeId>.output.callContext", "fieldType": ... }`, never a bare string. `fieldType` is `"object"` on the voice agent and `"string"` on the end-call node. | **NO** — the expression sits unprefixed inside the object's `expression` field. See [author/references/plugins/inline-voice-agent/impl.md § The `callContext` wiring rule](../author/references/plugins/inline-voice-agent/impl.md#the-callcontext-wiring-rule). |
```
rockymadden: **major** — the row immediately below this one contradicts the contract this PR exists to fix, and it was left untouched:

> **Inline-agent prompt** … Tokens reference upstream flow nodes directly: `{{ $vars.<flowNodeId>.output[.<field>] }}` (spaced braces) … **Never `{{input.<id>}}`** and never bare `{{name}}`.

Against that, in this PR:

- `inline-voice-agent/impl.md` step 4: `{{input.start__output__callerName}}`, *"never a bare `$vars.…`"*
- `check_voice_flow.py:359` grades any `$vars.` in `messages[].content` as a **hard failure**
- pre-existing `inline-agent/impl.md` § Anti-patterns: *"Never use raw `{{ $vars.X }}`"*

So this shared table (the one an author consults when wiring node output into a prompt) is the single place still teaching the form the PR's own checker fails. Fix the inline-agent row here, in this PR — otherwise the doc drift the PR body describes stays live one row down from the new one.

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+            "settings.voice is missing or empty — `uip agent init "
+            "--inline-in-flow --conversational` does not scaffold it, it must be "
+            "added by hand"
+        )
+    elif not str(voice.get("model") or "").strip():
+        errors.append("settings.voice.model is empty (the realtime speech model)")
+    else:
+        print(f"OK      settings.voice.model = {voice['model']!r} in {agent_json}")
+
+    if (agent.get("metadata") or {}).get("isConversational") is not True:
```
rockymadden, line 377: **major** — coverage is inverted relative to what the plugin doc says validate catches.

`impl.md:39`:

> `flow validate` checks `metadata.isConversational` and errors with `is not a conversational agent` when it is off; **a wrong `settings.engine` is *not* caught by validate** and surfaces only as a failed call

This grades `metadata.isConversational` (already covered by the task's separate `validate_flow.py` criterion at weight 3.0) and never looks at `settings.engine`. A clobbered `settings.engine` scores 1.000 and drops every call at runtime, which is exactly the failure a smoke check should own.

Add `settings.engine == "conversational-v1"` to the assertions.

skills/uipath-maestro-flow/references/author/references/plugins/inline-voice-agent/impl.md
```diff
@@ -0,0 +1,266 @@
...
+
+The scaffold is a conversational agent but has **no `settings.voice` block** — adding it is mandatory (next section).
+
+## Configure `agent.json`
+
+`--conversational` already writes everything a conversational agent needs except the voice block (full `agent.json` shape and per-field rules: the `uipath-agents` skill's [`agent-definition.md`](../../../../../../uipath-agents/references/lowcode/agent-definition.md)). Add `settings.voice` to `<FlowProjectDir>/<projectId>/agent.json`, leaving the scaffolded fields in place:
+
+```json
+{
+  "settings": {
```
rockymadden: **major** — this snippet is shaped as a *whole* `agent.json`, with only the preceding sentence ("leaving the scaffolded fields in place") to stop an agent applying it as a full-file `Write`. That drops `settings.model`, `settings.engine`, and `metadata.isConversational`.

Two of those losses are ungraded: per your own line 39 `flow validate` never checks `settings.engine`, and `check_voice_flow.py agent-voice` doesn't either, so the flow scores 1.000 and the call fails.

The sibling doc shows partials as fragments, not documents (`inline-agent/impl.md`: `"inputSchema": { "type": "object", "properties": {…} }`). Do the same here:

```json
"settings": {
  "voice": { "model": "…", "maxTokens": 65536, "temperature": 0, "persona": "Aoede" }
}
```

tests/tasks/uipath-maestro-flow/voice/voice_outbound_call.yaml
```diff
@@ -0,0 +1,123 @@
...
+  The agent must greet the person by name and state the amount owed, both taken
+  from the flow's inputs rather than hardcoded in the prompt: the flow is started
+  with a customer name and an outstanding amount, and the agent's system prompt
+  has to read those two values.
+
+  This sandbox has no UiPath tenant and is not logged in, so:
+  - Use placeholder E.164 numbers for the calling and destination numbers —
+    do not look up or provision real ones.
+  - Do NOT run anything that needs a live connection, and do NOT pack, deploy,
+    or place a real call. In particular, do not run `uip maestro flow debug`.
```
rockymadden, line 43: **major** — the prompt forbids `uip maestro flow debug`, but no criterion enforces it, and this is the task where the doc says a debug run **places a real phone call** (`impl.md`: *"An **outbound** flow does run under `flow debug`, and it dials for real"*).

The guard sits on the *inbound* task instead, where the CLI rejects debug on its own:

```yaml
- type: command_not_executed
  command_pattern: '(uip|\$UIP)\s+(maestro\s+)?flow\s+debug'
  weight: 2.0
```

Inverted risk. Move (or copy) that criterion here.

skills/uipath-maestro-flow/references/author/references/plugins/inline-voice-agent/impl.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+Outbound flows bind from the create-outgoing-call node instead: `$vars.createOutgoingCall1.output.callContext`.
+
+### Wire edges with Edit / Write
+
+The trigger's source port is `output`; every other voice edge leaves `success`; targets are always `input`. Edge object shape: [editing-operations-json.md § Add an edge](../../editing-operations-json.md#add-an-edge).
+
+Outbound inserts the call node between trigger and agent: `manualTrigger1 (output) → createOutgoingCall1 (input)`, then `createOutgoingCall1 (success) → voiceAgent1 (input)`. Tool/context/escalation resource nodes wire to the voice agent's artifact ports exactly as in [inline-agent/impl.md § Adding Resource Nodes](../inline-agent/impl.md#adding-resource-nodes).
```
rockymadden: **major** — the wiring section covers only the outbound insert. Nothing in either new file tells the author to **remove the scaffolded `core.trigger.manual`** when building inbound.

`greenfield.md`: *"`flow init` scaffolds ONLY the manual trigger (`start` / `core.trigger.manual`) with zero edges"*, while `planning.md` shows inbound starting at `core.trigger.voice`. Add the voice trigger without deleting `start` and you ship a two-trigger flow.

It also passes the new checker: `check_inbound_call_context` asserts exactly 1 `core.trigger.voice` and 0 dial-out nodes, and never looks for a stray manual trigger.

`brownfield.md`'s "Add a connector trigger" row is the precedent (*"Remove manual trigger, add connector trigger"*). State the same for inbound, and consider asserting it in the checker.

tests/tasks/uipath-maestro-flow/voice/voice_outbound_call.yaml
```diff
@@ -0,0 +1,123 @@
...
+      '"uipath.agent.voice"'
+      '"uipath.conversational.voice.end-call"'
+    timeout: 30
+    expected_exit_code: 0
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: run_command
+    description: "callContext from the create-outgoing-call node is bound into BOTH the voice agent and the end-call node"
+    command: "python3 $TASK_DIR/check_voice_flow.py call-context"
```
rockymadden, line 85: **minor** — nothing grades `inputs.from` / `inputs.to` on the create-outgoing-call node. `call-context` only asserts the node exists exactly once and checks the two `callContext` bindings, so a dial node with neither field, or with the shapes swapped, scores 1.000.

That asymmetry is one your own doc flags as easy to get wrong: `from` is a **plain string**, `to` is a **literal binding object**, both required, and per `impl.md` the validator doesn't read `fieldType`. Worth a small assertion in `check_call_context` (or a `flow_contains --regex`).

tests/tasks/uipath-maestro-flow/voice/voice_inbound_call.yaml
```diff
@@ -0,0 +1,126 @@
...
+    description: "callContext comes from the incoming-call trigger, bound into BOTH the voice agent and end-call, with no dial-out node and an entryPointId on the trigger"
+    command: "python3 $TASK_DIR/check_voice_flow.py inbound-call-context"
+    timeout: 30
+    expected_exit_code: 0
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: run_command
+    description: "Inline agent.json carries settings.voice and stays conversational"
+    command: "python3 $TASK_DIR/check_voice_flow.py agent-voice"
```
rockymadden, line 80: **major** — nothing on this task grades the prompt, and `impl.md` step 4 says an empty `messages[0].content` is valid. So an inbound flow whose voice agent has an empty or placeholder system prompt scores **1.000** while the call cannot hold a conversation, which is the deliverable.

`_shared/check_inline_agent.py` already grades exactly this for inline agents (placeholder-prompt set, `MIN_PROMPT_LEN = 40`). Add it as a second `run_command` here, or fold a prompt-length assertion into `agent-voice`.

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+INPUT_TOKEN_RE = re.compile(r"\{\{\s*input\.([A-Za-z0-9_]+)\s*\}\}")
+
+
+def check_prompt_inputs(flow_path: str, flow: dict, min_inputs: int = 1) -> int:
+    """Flow data reaching the voice prompt needs all four pieces aligned:
+
+    Delivery   node `inputs.agentInputVariables[]` binding -> BPMN JobArguments
+    Contract   the same key under agent.json `inputSchema.properties`
+    Resolution `{{input.<key>}}` in `messages[].content`
+    Variable   the bound `$vars.<trigger>.output.<id>` declared in globals
```
rockymadden: **minor** — the two "four pieces" lists don't match.

- here: Delivery / Contract / Resolution / **Variable** (globals)
- `impl.md` step 4: Delivery / Contract / Resolution / **Tokens** (`contentTokens` via `uip agent refresh`)

`contentTokens` ends up graded only by the advisory `command_executed` on `agent refresh` (`pass_threshold: 0`), so a stale `contentTokens[]` ships ungraded despite `impl.md` listing it as one of the four required pieces. Either grade the mirror here (it is a pure `content` → tokens derivation) or stop calling both sets "the four pieces".

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
```diff
@@ -0,0 +1,438 @@
...
+        )
+    else:
+        print("OK      metadata.isConversational is true")
+
+    if errors:
+        return _fail("; ".join(errors))
+    return 0
+
+
+def _voice_agent_json(flow_path: str, flow: dict) -> tuple[dict, dict, str] | int:
```
rockymadden, line 412: **minor** — duplicate of `check_agent_voice`'s first half: same `_nodes_of(flow, VOICE_AGENT)` + exactly-1 check, same `inputs.source` validation, same `os.path.join(dirname(flow_path), source, "agent.json")`, same `json.load` + error tuple.

The copy is also worse: it lost the `if not os.path.exists(...)` branch that lists the `agent.json` files actually present, so a `prompt-inputs` failure prints a bare `FileNotFoundError` where `agent-voice` prints the useful diagnostic.

Have `check_agent_voice` call `_voice_agent_json` and keep one implementation.

tests/tasks/uipath-maestro-flow/voice/voice_inbound_call.yaml
```diff
@@ -0,0 +1,126 @@
...
+      the release after deploy. Adding or enabling a number is portal-only.
+
+      Score 1.0 if the answer conveys that testing means publish/deploy + bind a
+      phone number + call it, and does not suggest `flow debug` for this flow.
+      Score 0.5 if the deploy-and-bind path is roughly right but vague about the
+      number binding, or `flow debug` is mentioned alongside the correct path.
+      Score 0.0 if the answer tells the user to run `uip maestro flow debug`,
+      claims the flow can be exercised locally, or omits how to run it at all.
+    weight: 2.0
+    pass_threshold: 1.0
```
rockymadden: **minor** — the rubric defines a 0.5 band ("deploy-and-bind path is roughly right but vague about the number binding, or `flow debug` mentioned alongside the correct path") but `pass_threshold: 1.0` makes it unreachable: a half-right answer scores identically to one that tells the user to run `flow debug`.

Repo convention for banded `llm_judge` rubrics is a sub-1.0 threshold (`uipath-human-in-the-loop/smoke_07_neg_automated.yaml` → 0.8; `uipath-platform/data-fabric/*` → 0.5). Set 0.5 here, or delete the band.

skills/uipath-maestro-flow/references/shared/node-output-wiring.md
JoshParkSJ: just fyi since it touches autonomous agents - I think this inconsistency was introduced from https://github.com/uipath/skills/pull/1656 but will be fixed from this pr

tests/tasks/uipath-maestro-flow/voice/check_voice_flow.py
JoshParkSJ: Added the filter, though I could not reproduce the `.cli-stage` case — Python glob skips dot-prefixed folders, so the real exposure is `_outputs` / `v1stage`.

rockymadden (approved):
Approving. All 15 comments from my last pass are closed, and I re-verified each one against fixtures rather than reading the diff:

- callContext now graded on `type == "jsExpression"`; `literal` and `=jsonString:` both fail, `=js:` passes with a warning
- trigger-global rule gated on the source actually being a trigger; `inout` accepted
- `.cli-stage` / `_outputs` excluded when locating the `.flow`
- `settings.engine`, the dial node's `from`/`to`, the system prompt, and the stray inbound manual trigger all graded
- `settings.voice` reshaped to a fragment with an explicit Edit-don't-Write instruction
- flow-debug guard moved onto outbound, `llm_judge` 0.5 band made reachable
- `node-output-wiring.md` inline-agent row corrected (link-text broken count drops 9 to 8)

Six documented `prompt-inputs` failure modes still fail; clean builds still pass. Could not construct a false pass against the checker.

Non-blocking, worth a follow-up commit before merge:

**1. `timeout: 60` on the validate criterion (both tasks).** This is why smoke is red right now. Both tasks scored 1.00 on every voice-specific criterion and failed only here:

```
sandbox: Command 'python3 $SKILLS_REPO_PATH/.../validate_flow.py' timed out after 60 seconds
```

56 of 58 maestro-flow tasks give `validate_flow.py` `timeout: 180`. These two are the only ones at 60, and per `impl.md` voice validate does strictly more work than the others (agent dir, `settings.voice`, both callContext bindings, subflow check). Raise to 180.

**2. Outbound turn budget.** CI agent turn ran ~766s against `turn_timeout: 900`. The comment cites a 535s local run. 85% of budget is thin.

**3. `jsExpression` `to` on the dial node scores 0.** Dialing a number bound from the trigger is a legitimate shape; the error says "not E.164" rather than "must be a literal".

**4. `BINDING_RE` accepts `=js:$vars…` on `agentInputVariables`,** while `inline-agent/impl.md` says to strip that prefix.

**5. `impl.md:158` uses `manualTrigger1`** where the rest of the file (and `flow init`) uses `start`.

**6. Outbound has no `skill_triggered` criterion;** inbound grades it at 2.0. Looks like fallout from the three-tasks-into-two merge.

PR #2387, opened 2026-07-30
feat(bpmn): derive package metadata with refresh

DevMomo:
## Problem

A connector-bound project regenerated with the deprecated `update-metadata` **passes `validate` and then faults at runtime**:

```
uip maestro bpmn update-metadata <file.bpmn>   # writes the JSON
uip maestro bpmn validate <file.bpmn>          # clean
# ...at runtime:
102010 IntSvcArgumentsError - Integration Services invalid value in input
```

`update-metadata` does not materialize Integration Service connection bindings, and nothing said so. Two adjacent gaps: no stated point at which metadata gets regenerated, and agents hand-repairing generated JSON when `refresh` exited nonzero.

**Why the e2e needs this:** only `refresh` writes real connection ids into `bindings_v2.json`, which its pack criterion checks.

## Fix

- New step 5: regenerate the four CLI-owned files after source validation, only for a package-ready deliverable
- A nonzero `refresh` is a **source** failure — fix the BPMN, revalidate, refresh again; never hand-repair the JSON
- `update-metadata` marked deprecated, with the null-connection consequence spelled out
- Registry-evidence-only tasks get a command-first recipe, off the metadata path

## Verification

Draft-boundary and script-lifecycle evals updated to the generated shape, via a shared `assert_generated_project_scaffold` helper.


skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -262,6 +243,19 @@ For registry-evidence-only tasks, be command-first and time-boxed:
    structural rules, the installed CLI predates them — update it (see
    [references/cli-conventions.md](references/cli-conventions.md)). See
    [references/structural-bpmn.md#validation](references/structural-bpmn.md#validation).
+5. **Refresh derived metadata when package-ready output is required.** After
+   source validation passes, regenerate the four CLI-owned package files:
+
+   ```bash
+   uip maestro bpmn refresh <project-path> --output json
```
rockymadden, line 268: **Critical — this new mandatory step will fail on the XML the skill tells the agent to write.**

`references/structural-bpmn.md:94,96` — the "complete minimal file (author from this, not from examples)" that SKILL.md step 3 points at — declares public I/O as:

```xml
<uipath:input  id="input_Var_Amount" name="Amount" type="double" elementId="Start_1" />
<uipath:output id="output_Var_Echo"  name="Echo"   type="double" elementId="End_1" />
```

`uipath:input` / `uipath:output` are exactly the node types that feed entry-point schema derivation. In `packages/maestro-sdk/src/bpmn-packaging/entry-points.ts`, `schemaForType` accepts only:

```
string, boolean, integer, number, array, object, json
```

`double` falls through to `default:` and throws:

> `Unsupported process input/output type "double". Use string, boolean, integer, number, array, object, or json, or add an inline JSON schema.`

Neither example line carries an inline CDATA schema, so the `variable.$body` escape hatch in `schemaForVariable` doesn't apply. And this runs inside `deriveFiles`, *before* `validateProject` — so it aborts refresh outright.

Why this PR is where it becomes live rather than latent: the command being replaced swallowed it. `bpmn-update-metadata-service.ts` does `schema = { type: TYPE_MAP[type] ?? "object" }` — `double` silently became `{type:"object"}` and the command succeeded. `refresh` throws. So promoting `refresh` to a required step converts a silent, lossy-but-working path into a hard stop.

The interaction with this PR's own guidance is what makes it critical rather than annoying. Step 5 says a nonzero result is "a source/precondition failure: fix the BPMN ... never repair the generated JSON by hand." An agent that obeys will go read the BPMN, find it already matches the reference example, and have nowhere to go. Meanwhile `uip maestro bpmn validate` stays clean, because derivation lives in `refresh`, not `validate` — the exact "clean validate, broken downstream" class this whole stack is fixing.

Also note #2366 (this PR's base) reinforces `double`: `check_script_jint_guidance.py` requires the public `amount` input and `riskScore` output to be `type="double"`, and its second criterion is `validate` — so it passes while producing an unpackageable project.

Cheapest fixes, either works:

1. Change the two example lines to `type="number"` (in `schemaForType`'s vocabulary, and `PRIMITIVE_TYPES` in PO.Frontend has no `number`, so this is purely the public-I/O declaration — worth confirming the canvas accepts it there).
2. Keep `double` and give both lines an inline JSON-schema CDATA body, which takes the `$body` path and bypasses `schemaForType` entirely.

Either way, the guide's reject list should name this, since "fix the BPMN" is unactionable without knowing the accepted vocabulary differs from the canvas variable types.

skills/uipath-maestro-bpmn/references/shared/local-metadata-regeneration-guide.md
```diff
@@ -89,22 +99,35 @@ Do not derive metadata from stale package files first. Use existing generated fi
...
 
-6. Inspect the package or generated content for:
-   - `entry-points.json` entries matching root start events and schemas.
-   - `bindings_v2.json` resources matching root bindings and enriched connector metadata.
-   - `operate.json` pointing at the intended BPMN file with `ProcessOrchestration` content type.
-   - `package-descriptor.json` entries for the BPMN file and generated JSON under `content/`.
-7. If the installed CLI cannot regenerate a needed file in place, keep the generated file stale only as a known blocker and report the exact unsupported step. A source-only project is not package-ready.
+If refresh fails, the atomic write contract leaves the prior four-file set
+unchanged. Fix the reported source or project precondition and run it again; do
+not patch generated JSON around the failure. The current contract requires
```
rockymadden: **Minor — the precondition list is missing `project.uiproj`.**

`BpmnRefreshService.execute` checks, in order:

1. project path is a directory
2. exactly one root `.bpmn` (`findBpmnFile`)
3. **`project.uiproj` exists and sets `ProjectType: "ProcessOrchestration"`** (`readProjectFile`)

Step 3 produces two distinct errors — `Required file is missing: <path>` and `project.uiproj must set "ProjectType" to "ProcessOrchestration".` — and neither is derivable from the list here. Given the guide's premise is that a refresh failure is a *precondition* failure the agent must self-diagnose, the precondition that isn't in the BPMN at all is the one most worth naming.

While you're here, the reject list is also missing a few `throw`s from `entry-points.ts` that an agent will realistically hit: a root `uipath:output` whose `elementId` isn't a root end event (`Process output "X" must target a root end event.`), a duplicate process input/output name, and a variable with no `type` and no inline schema.

skills/uipath-maestro-bpmn/references/shared/local-metadata-regeneration-guide.md
```diff
@@ -89,22 +99,35 @@ Do not derive metadata from stale package files first. Use existing generated fi
...
+exactly one project-root `.bpmn` file, one or more root processes, and at least
+one root manual start event overall. Each root manual start event must carry
+exactly one valid GUID `uipath:entryPointId`; refresh generates one
+`entry-points.json` entry for each such start event. Root `uipath:binding`
+nodes whose `resource` is not `Connection` are silently dropped, not reported —
```
rockymadden: **Minor — "silently dropped" is narrower than stated.** The filter in `connection-resources.ts` is:

```ts
node.$type === "uipath:binding" &&
node.resource === "Connection" &&
node.propertyAttribute === "ConnectionId"
```

So a `resource="Connection"` binding with any other `propertyAttribute` is *also* dropped silently. That case is worse than the one documented, because the binding then isn't in `bindingById` and the connector node fails with:

> `Activity "X" references missing Connection binding "Y".`

...which points at the connector while the actual defect is one attribute on the binding. Worth one clause here — this is precisely the kind of misdirection the guide is trying to pre-empt.

tests/tasks/uipath-maestro-bpmn/_shared/bpmn_assertions.py
```diff
@@ -162,3 +171,72 @@ def assert_package_lifecycle(project_dir: Path, bpmn_name: str, start_id: str) -
...
+def assert_generated_project_scaffold(
+    project_dir: Path,
+    project_name: str,
+    bpmn_name: str,
+    start_id: str,
+    *,
+    entry_point_id: str | None = None,
+    expected_resource_count: int | None = None,
+) -> None:
+    """Assert the current CLI-owned Process Orchestration metadata contract."""
```
rockymadden: **Minor — the docstring overclaims relative to what the helper enforces.** Line 215 fails unless `entry-points.json` has *exactly one* entry, but `buildBpmnPackageMetadata` emits one entry per root manual start event across all root processes, and this PR's own guide says so ("refresh generates one `entry-points.json` entry for each such start event").

So this asserts "the single-manual-start subset of the contract," not "the current CLI-owned Process Orchestration metadata contract." Fine for both current call sites, but the next person reaching for a shared helper by its docstring will get a confusing failure on a legitimate two-start-event project.

Either narrow the docstring, or take an optional `expected_entry_points: int = 1` in the same style as the existing `expected_resource_count`.

rockymadden (approved):
## Approve

The command swap is right and I verified it against `origin/main` of `UiPath/cli` and `UiPath/PO.BpmnEngine` rather than from the PR body:

|---|---|---|
| `refresh` writes **four** files (docs changed "five"→"four") | `BPMN_PACKAGE_FILES` in `packages/maestro-sdk/src/bpmn-packaging/package-metadata.ts` = entry-points, bindings_v2, operate, package-descriptor. `update-metadata`'s own description still says five (it also writes `project.uiproj`), so the four/five change tracks the command swap exactly | correct |
| `update-metadata` deprecated, does not materialize IS bindings | `UPDATE_METADATA_DEPRECATION` in `bpmn-update-metadata-service.ts` — the skill's new wording is near-verbatim from the CLI's own constant, including "(including triggers)" | correct |
| exactly one project-root `.bpmn` | `BpmnRefreshService.findBpmnFile` throws on `files.length !== 1` (non-recursive readdir, so "project-root" is the right word) | correct |
| atomic; prior set survives a failure | `validateProject` runs before `writeFilesWithRollback` | correct |
| offline / provider-neutral | zero login/tenant/token/http/fetch references in `bpmn-refresh-service.ts` | correct |
| `102010 IntSvcArgumentsError - Integration Services invalid value in input` | `src/Common/UiPath.PO.Errors/ErrorCode.cs:144` + `UiPath.PO.Worker/Resources/Localization/en-US.json:12` — verbatim | correct |
| the four things refresh rejects | `bindingReference` regex throw, missing-binding throw, `isGuid` on `default`, resourceKey-reuse throw — all four in `connection-resources.ts` | correct |
| `bindings_v2.json` version `"2.0"`, `operate.main`, descriptor files map | `buildBpmnPackageMetadata` | correct |

Worth calling out: the old doc's `Connection null (error 102010)` was invented. Replacing it with the real localized string is a genuine fix, not a wording change.

One critical item inline that I'd like fixed before #3113's pack criterion leans on this step — it's a pre-existing defect in `structural-bpmn.md` that this PR exposes rather than introduces, which is why it isn't blocking. Three minors alongside it.

No AI slop. Two comments in `bpmn_assertions.py` deserve specific praise — the `assert_package_lifecycle` docstring explaining why it was *deliberately not* replaced (citing `api_workflow_task.yaml`'s stated contract) and the `#2774` note on the `main`-preservation branch both pass the "does this change what the agent does?" test. That's the right bar.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Right, and the same finding landed on #2369. Took option (1) there: the example, the prompt, and `check_script_jint_guidance.py` are all `number` now, and the reject list here names the accepted vocabulary plus the inline-schema escape hatch, so "fix the BPMN" is actionable. Your point about why this PR makes it live is the useful half — `bpmn-update-metadata-service.ts` coerced it to `{type:"object"}` and succeeded, so requiring `refresh` is what turns it into a hard stop.

skills/uipath-maestro-bpmn/references/shared/local-metadata-regeneration-guide.md
DevMomo: Added — the `project.uiproj` precondition with both messages, called out as not being a BPMN defect. Also added the three `entry-points.ts` throws you listed.

skills/uipath-maestro-bpmn/references/shared/local-metadata-regeneration-guide.md
DevMomo: Fixed. The guide now states the full predicate and that the failure surfaces as `Activity "X" references missing Connection binding "Y"` — pointing at the activity while the defect is on the binding.

tests/tasks/uipath-maestro-bpmn/_shared/bpmn_assertions.py
DevMomo: Took the parameter: `expected_entry_points: int = 1`, with the docstring saying `start_id` is the entry checked in detail. Both existing call sites keep the default; two tests cover the multi-entry path.

PR #2395, opened 2026-07-30
fix(maestro-case): choose the SLA response instead of assuming an interrupting lane

abhiram-vad:
SLA slice of closed #2348 (case bug-bash, 2026-07-30). Siblings: #2392 (sequencing + eval + Case Review maps), #2393 (reachability/re-entry). #2394 merged.

## Finding closed
**Mounish** — "SLA breached event based entry rule can be interrupting/non-interrupting — it's not always interrupting; only Case SLA breaches need to be modeled as a stage. Stage SLA breach can be handled within the respective stage itself." The skill previously forced every SLA-driven entry into an interrupting secondary stage.

## Response taxonomy (replaces the reflex)
Two SLA scopes exist — case and stage. The response is chosen from the source; `Interrupting` follows what the response does to active work, never the SLA's scope.

| Response | Source says | Shape | Interrupting |
|---|---|---|---|
| `notify-only` | notify / alert / email / page | escalation notification only — no stage, no task | n/a |
| `start-task` | local work in the **same** breached stage | the breached stage carries an `sla-status-change` entry referencing **its own** SLA; the task lives there | `No` unless current work must stop |
| `enter-stage` | ownership change, escalation lane, recovery, visible lifecycle step | a separate stage carries the entry | `Yes` when it takes over/pauses/reroutes; `No` for parallel oversight |
| `exit-stage` | the stage should end, fail, or route away | stage-exit row | per exit semantics |
| `exit-case` | the case should close, cancel, fail, or reach an alternate terminal outcome | §1.4a case-exit row | per exit semantics |

A case-level graph response enters a separate stage; a stage-level response may `start-task` in the breached stage or `enter-stage` elsewhere. Action-task SLA stays in the task's own timer fields unless a missed task must change the graph. **Absent a stated response, at-risk and breached are `notify-only`.**

## Verified against PO.Frontend source + the CLI (not assumed)
Read `PO.Frontend@origin/develop` 0.983.0 (2026-07-30): `useTransitionRuleSlaStatusChangePropertiesFields.tsx`, `ValidateCaseManagementConditionsUtils.ts`, `CaseManagementJsonSlaStatusChangeRuleSchema.ts`, `CaseManagementJsonEscalationsSchema.ts`.

- Only `slaId` + `escalationId` persist; **scope and status are inferred** — scope from where the referenced SLA lives, status from the escalation reference. This PR therefore stops asserting a scope-driven interrupting rule.
- **Breach references the SLA alone.** FE, verbatim: *"A missing escalationId is the persisted representation of a Breached rule."* At-risk additionally names a concrete at-risk escalation (which must exist on that SLA). Both stated in the rules and in the entry-condition impl recipe.
- **FE's `"any"` escalation sentinel is an FE/CLI skew** — FE accepts it, released `uip maestro case validate` rejects it ("The escalation referenced by rule ... no longer exists"). The skill now forbids authoring it.
- Escalation `action.type` is `["notification"]` only at v27, so a graph response can never be an escalation action — it must be a condition rule, as this taxonomy states.
- FE offers every stage (minus Case Manager) as the SLA source, including the stage owning the condition — the same-stage `start-task` shape is authorable, matching the CLI probe.

CLI probe (SlaGateProbe, alpha/codereval, uip 1.198.0-preview.102): a stage carrying `sla-status-change` on **its own** SLA validates with `isInterrupting` either `true` or `false`; the FE-native breach shape (`slaId`, no `escalationId`) validates; a separate non-interrupting stage validates and uploads to Studio Web; a dangling `slaId` fails, proving resolution is real.

## Known gap left for a follow-up (not this PR)
FE offers `sla-status-change` on **every** transition type except `sequential-task-entry` — stage entry, stage exit, **task entry**, stage-complete, case-exit, case-complete — and a task-entry SLA rule passes `validate`. Main's docs (from #2344) say "stage entry only", which blocks the most direct `start-task` shape. Correcting that touches #2344's contract and its eval, so it is called out rather than folded in here.

## Merge notes
6 files, rule-only. An earlier revision carried an over-broad invention ("a stage-SLA breach must never interrupt a task inside its own stage") that needed policing in two extra files; it was wrong on the product and is fully removed here and in #2392/#2393. Rebased on current main; verified conflict-free with both siblings in multiple merge orders. Please trigger the CI coder-eval run before undrafting.


charlesliu9:
• The new SLA-response guidance contains contradictory emission rules that can produce incorrect breach behavior and unintended stage re-entry. Several added evaluation tasks also use unsupported taxonomy tags.

  Full review comments:

  - [P1] Allow breach rules without escalation IDs — /skills/skills/uipath-maestro-case/references/plugins/conditions/
    stage-entry-conditions/impl-json.md:87-87
    For a breached enter-stage response, this now correctly says slaId alone is the persisted shape, but the preceding planning/
    implementation contract still requires escalation-display-name and resolves both IDs. That leaves a breach either unplannable or emitted with an escalationId, which changes it into an at-risk response; make the escalation field optional for breach rules throughout the stage-entry contract.

  - [P1] Place start-task rules on the follow-up task — /skills/skills/uipath-maestro-case/references/sdd-generation-rules.md:32-32
    This directs start-task responses to a stage-entry rule, contradicting the newly added task-entry implementation and canonical SLA response guide, which require the rule on the follow-up task. On an SLA breach this stage-reentry shape can restart the stage's existing tasks (especially now that tasks default to rerunnable), rather than only launching the requested follow-up task.

  - [P2] Use only supported taxonomy tags — /skills/tests/tasks/uipath-maestro-case/sla_response/notify_only_default/
    notify_only.yaml:10-10
lifecycle:edit and feature:sla are outside the closed lifecycle/feature vocabularies, so this new smoke task is misclassified or skipped by taxonomy-based reporting and filtering; the same invalid tags appear across the new SLA-response tasks. Use an allowed lifecycle value and remove or propose the new feature value. ../../../../README.md:98-103 ../../../../../.claude/rules/test-writing.md:19

charlesliu9 (approved):

PR #2398, opened 2026-07-30
fix(ci): harden catalog refresh — branch cleanup, token split, subtree-loss guard

uipreliga:
Two independent failures of the nightly catalog refresh, both found while fixing the first: a **month-long outage** (2026-06-30 → 07-30) and a **silent catalog corruption** that nearly auto-merged on 07-30.

## Part 1 — the outage

Reported in Slack as a suspected expired PAT; it was not. `GH_PAT` was rotated on 2026-06-25 (secret `updated_at`) to a token that can read the GitHub Packages `@dev` feed but cannot write labels. `gh label create` returned:

```
HTTP 404: Not Found (https://api.github.com/repos/UiPath/skills/labels)
```

GitHub returns **404, not 403**, when a token lacks a permission — which is why this read as "token revoked" rather than "token under-scoped". The token was valid the whole time: the same secret authenticated the npm feed in the step immediately before.

The rotation was 06-25 but the first failure was 06-30, because runs on 06-26…06-29 skipped every step behind the "existing open refresh PR" guard and reported green **vacuously**. Last catalog to actually merge: 06-24.

### Fixes

**Failures no longer strand branches.** The step pushed the branch and then died, and the guard at the top only looks at open PRs — so it never noticed its own debris. 34 `auto/refresh-uip-catalog-*` branches accumulated with zero open PRs (since deleted). A trap now removes the pushed branch unless a PR ended up owning it.

This also closes a latent deadlock: the branch name is `<cli-version>-$(date -u +%Y%m%d)`, so a same-day retry at an unchanged CLI version collided with its own orphan and died on a non-fast-forward `git push` — failing *before* the real error, with a misleading message.

**`gh label create` is no longer fatal.** The `automated` label already exists on this repo, so the call is belt-and-braces for fresh forks — but it ran under `bash -e`, so a permission hiccup there aborted an otherwise-good refresh. Now non-fatal with a `::warning::`. `gh pr create --label automated` is the call that actually has to succeed, and it still fails loudly.

**One secret no longer serves two very different jobs.** `smoke-skills.yml` and `smoke-rpa-skills.yml` are `pull_request`-triggered **required checks** on a **public** repo. Fork PRs don't receive secrets, but anyone who can push a branch here could read the secret handed to those builds — and that secret also carried repo write. Feed-read consumers now take `GH_PACKAGES_TOKEN`; `GH_PAT` is referenced only by `refresh-uip-catalog.yml` and `automerge-catalog-refresh.yml`, which trigger on `schedule` / `check_suite` and cannot be influenced by a PR author.

**Two stale comments corrected.** Both workflows claimed this PAT is "already used by activation-gate.yml". It isn't — `activation-gate.yml` never references `GH_PAT`. That claim is what pointed the original triage at a blanket token expiry.

## Part 2 — the silent catalog corruption

The first successful run after the token fix produced #2397, which dropped the entire `rpa debug` subtree (16 verbs) plus `rpa analyzer-rules list` and `rpa files diff`. It was **not** a CLI change: `uip <group> --help` failed for those three groups, and `collect_group` marks a group unwalkable and returns no children when that happens. The group node itself survives, so the subtree vanishes silently — 18 verbs is 1.3% of a 1384-verb catalog, comfortably inside `--max-drop-frac 0.2`.

The failures are transient, not structural: the same three groups walked fine 16 hours earlier, and five *different* `rpa` groups failed in that earlier build instead. #2397 had green required checks and the `automated` label, and the CLI Verb Gate is deliberately advisory — so it would have auto-merged unattended.

### Fixes

- **Retry** — `collect_group` attempts a failed `--help` 3 times with linear backoff before marking a group unwalkable. Only failing groups pay the cost.
- **Subtree-loss guard** (always on, no flag) — refuse to write when a group that was walked in the committed snapshot is unwalkable in this build. `--allow-unwalkable-regression` is the escape hatch for a genuine permanent loss, or for recovering when the committed snapshot is itself the flaky build.

The discriminator is "did the previous snapshot enumerate children under this group", which keeps the guard quiet on everything that isn't a regression: leaf commands (`uip rpa pack --help` emits no `Subcommands`), groups already unwalkable, groups genuinely deleted from the CLI (a deleted group never enters the walk, so it cannot appear in the unwalkable set), and brand-new groups.

## Sequencing — read before merging

Every feed site reads `${{ secrets.GH_PACKAGES_TOKEN || secrets.GH_PAT }}`. The fallback is deliberate: pointing required checks at a secret that doesn't exist yet would fail them, block every merge in the repo, and block this PR. Behavior is **identical to today** until the secret is created.

Follow-up once `GH_PACKAGES_TOKEN` exists (classic PAT, `read:packages` only, no repo scopes — GitHub Packages' npm registry rejects the fine-grained path):

- [ ] Create `GH_PACKAGES_TOKEN`
- [ ] Scope `GH_PAT` down to contents + pull-requests + issues + metadata, pinned to this repo
- [ ] Drop the `|| secrets.GH_PAT` fallbacks
- [ ] Consider a GitHub App for the write credential so expiry stops causing silent multi-week outages

## Verification

- Outage fix confirmed on [run 30581973513](https://github.com/UiPath/skills/actions/runs/30581973513): green with **every step executed, none skipped** — including `Open PR with updated catalog`, the step that had 404'd daily. It produced #2397 carrying the `automated` label, so the label write is proven rather than vacuously green.
- Trap tested on both paths: `gh pr create` failure deletes the branch **and still exits 1** (cleanup does not mask the error); success keeps the branch for the auto-merger.
- Subtree-loss guard tested against the two **real** snapshots, not just fixtures. On the bad build it reports exactly `'rpa analyzer-rules' (-1); 'rpa debug' (-16); 'rpa files' (-1)` = 18 verbs; `verb_count_error` on the same inputs returns `None`, confirming the gap it closes.
- 10 new unit tests in `tests/scripts/test_catalog_build_guards.py` covering the non-regression cases and both retry paths. **103 tests pass** across `tests/scripts/`.
- All 5 workflows: YAML parses, `bash -n` clean on every `run:` block.

Note the flakiness is symmetric: if a flaky snapshot ever lands on main, the guard will block the next *good* build too. The error message says so and points at the flag.

No changes to any skill, reference, or task YAML.

### DIFF
--- a/.github/workflows/automerge-catalog-refresh.yml
+++ b/.github/workflows/automerge-catalog-refresh.yml
@@ -32,7 +32,12 @@ jobs:
         env:
           # PAT so the merge commit triggers downstream workflows (e.g. any
           # post-merge job that depends on the new catalog landing on main).
-          # Same PAT used elsewhere — see activation-gate.yml.
+          # Repo-write credential: needs contents + pull-requests + issues
+          # (the `--label automated` filter below reads labels, which sit
+          # under the Issues permission) + metadata (the rules API call).
+          # Shared with refresh-uip-catalog.yml and nothing else — both
+          # trigger on `schedule`/`check_suite`, so a PR author cannot reach
+          # this token. Feed-read consumers use GH_PACKAGES_TOKEN.
           GH_TOKEN: ${{ secrets.GH_PAT }}
           # `gh pr {list,view,merge}` resolve the repo from a local git
           # remote; this job has no actions/checkout step, so GH_REPO is
--- a/.github/workflows/refresh-uip-catalog.yml
+++ b/.github/workflows/refresh-uip-catalog.yml
@@ -89,6 +89,13 @@ jobs:
           # --min-verbs 1000 is a deliberate hard tripwire — it caught the
           # orchestrator-tool drop (#1203 follow-up). The build must clear it;
           # a stuck refresh is louder than a silently shrunk catalog.
+          # A third guard needs no flag and is always on: the build refuses to
+          # write if a group walked in the committed snapshot comes back
+          # unwalkable. Losing one command group is far too small a fraction of
+          # a ~1380-verb catalog for --max-drop-frac to notice (`rpa debug` +
+          # 2 others = 0.5% on 2026-07-30), and the group node survives, so the
+          # subtree vanishes silently. Failed `--help` walks are also retried
+          # first — see scripts/build-uip-catalog.py.
           python3 scripts/build-uip-catalog.py --install-tools \
             --min-verbs 1000 --max-drop-frac 0.2
 
--- a/.github/workflows/refresh-uip-catalog.yml
+++ b/.github/workflows/refresh-uip-catalog.yml
@@ -63,7 +63,11 @@ jobs:
       - name: Install uip CLI (GitHub Packages @dev)
         if: steps.check.outputs.exists != 'true'
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.GH_PAT }}
+          # Feed read only — deliberately NOT the repo-write PAT. See the
+          # `Open PR` step below for the split rationale. Falls back to
+          # GH_PAT so this keeps working until GH_PACKAGES_TOKEN is created;
+          # drop the fallback once it exists.
+          NODE_AUTH_TOKEN: ${{ secrets.GH_PACKAGES_TOKEN || secrets.GH_PAT }}
         run: |
           npm config set @uipath:registry https://npm.pkg.github.com/
           npm config set //npm.pkg.github.com/:_authToken "$NODE_AUTH_TOKEN"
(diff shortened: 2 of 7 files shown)


Cellcote (approved):

uipreliga:
Thanks — both nits taken, in 908d37078 (comment-only, no behavior change).

**`smoke-rpa-skills.yml:146` wrap.** Correct, it read as an interrupted parenthetical because the token note ran into the `@dev` pinning paragraph mid-sentence. Now its own paragraph, and it states *why* the trigger matters rather than just naming it — a branch pusher on this public repo can read the secret the build is handed — so it matches the wording already in `smoke-skills.yml:342`.

**`run-coder-eval.yml:196` trigger model.** Good catch, and the comment was misleading by omission: that workflow is `workflow_dispatch`-only, so the fork-PR argument genuinely does not apply to it. It now says so explicitly and names the actual justification there (least privilege), pointing at the two `pull_request`-triggered workflows for the stronger rationale.

**Alerting on `schedule` failures — agreed, not in this PR.** You're right that the trap fixes branch stranding without fixing silent failure, and that's exactly how the 31-day outage stayed invisible. But a notification path is new surface rather than a fix to anything in this diff, and it wants a decision on where it lands (Slack webhook vs. issue-on-failure). Tracking it with the GitHub App item in the follow-up checklist rather than widening this PR.

**`GH_PACKAGES_TOKEN` creation.** Agreed the security win is intent until the secret exists — that's the first box in the PR body's follow-up list, along with scoping `GH_PAT` down and dropping the `|| secrets.GH_PAT` fallbacks. Until then behavior is byte-identical to today, which is what keeps this PR mergeable ahead of the secret.

---

On the red check: `Run RPA skill smoke tests (Windows)` failed on `skill-rpa-coded-test-case` (`weighted_score 0.20 < 0.7`), which is a pre-existing borderline task, not a regression here. It hit the 70-turn cap having written only `TestLoginFailure.cs`, so the four criteria anchored on `TestLoginSuccess.cs` scored 0. The same task **also** exhausted max turns on `feat/uipath-rpa-consolidate-refs` ([run 30625803039](https://github.com/UiPath/skills/actions/runs/30625803039)) 16 hours later — it just landed the second file that time and scored 0.85, clearing the bar. Same failure mode, opposite side of the threshold.

The install step succeeded in the failing run (the `|| secrets.GH_PAT` fallback held) and the other two tasks scored 1.00, so the token split is not implicated. The job ran at all only because `smoke-rpa-skills.yml` is in the diff, which `detect` treats as an infra change and fans out to every `windows`-tagged task. Re-triggered by the latest push.

PR #2401, opened 2026-07-30
test(uipath-test): Playwright integration and e2e tasks (hold until C…

addyGarg:
## What

The Playwright **integration** and **e2e** eval tasks for the `uipath-test` skill, split out of UiPath/skills#2322 so that PR can merge now while these wait for the capability they exercise.

| Task | Tier | Covers | Needs serverless execution? |
|---|---|---|---|
| `playwright_ingest_and_select_integration.yaml` | integration | pack → upload → ingestion → label-filled test set → single-Playwright-package check | **No** — stops before running |
| `playwright_first_mile_e2e.yaml` | e2e | the same flow through a chromium-only run to real results | Yes |

Both grade **saved artifacts** (`context.json`, `report.json`) rather than live tenant state, so the agent can delete the project it created without breaking the assertions — and a stale execution from an earlier run cannot satisfy an outcome check. Both carry a teardown criterion, so scheduled runs leave nothing behind.

## Why separate

The eval runner installs `@uipath/cli@latest` before any task runs, and tasks are forbidden from pinning a version (`tests/README.md`). Until the external-package commands reach npm `latest`, these can only fail — and a scheduled task that always fails is noise that also desensitises people to real failures. So the skill and its passing smoke task ship in #2322; these wait here.

## Do not merge until

1. **The commands are on npm `latest`** — check `npm view @uipath/cli version` reports 1.200.x. The 1.200 cut is not sufficient: stable promotion is a separate manual dispatch.
2. **UiPath/cli#3336 is in that released line** — the integration task reads `playwright-context` output, which a stable CLI only registers once that fix ships.
3. **The eval tenant has Test Manager Playwright support enabled** (both tasks), and **working serverless Playwright execution** (e2e task only).

Then run each once by hand and merge on green:

```bash
SKILLS_REPO_PATH=$(cd .. && pwd) .venv/bin/coder-eval run \
  tasks/uipath-test/playwright_ingest_and_select_integration.yaml -e experiments/default.yaml
```

## Validation

The flows these tasks assert are proven manually: on a local serverless stack and on alpha `testmanagerdev/DevTest`, a context-free agent using only the skill reached **2/2 passed** on a chromium-only run, including a fix-and-republish loop. What is unproven is these YAMLs executing under `coder-eval` on a capable tenant — hence the manual run in the merge checklist above.

The e2e task is a candidate for the `path-to-ga` tag (must-pass, currently blocked); it is left off so this PR stays mergeable on demand, since that tag needs an allowlisted approval on the head commit.

## Related

- **UiPath/skills#2322** — the skill content and the smoke task that passes today.
- **UiPath/cli#3336** — precondition 2 above.

### DIFF
--- a/tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
+++ b/tests/tasks/uipath-test/playwright_first_mile_e2e.yaml
@@ -42,11 +42,18 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: command_executed
-    description: "Scoped the run to chromium only"
+    description: "Ran the test set scoped to the chromium project"
     tool_name: "Bash"
-    command_pattern: 'uip\s+tm\s+testsets?\s+run\s+.*--playwright-projects\s+chromium(?!\s+[A-Za-z])'
+    command_pattern: 'uip\s+tm\s+testsets?\s+run\s+.*--playwright-projects\s+chromium'
     min_count: 1
     weight: 3.0
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Did not widen the run to a second Playwright project"
+    tool_name: "Bash"
+    command_pattern: '--playwright-projects\s+\S+\s+\S'
+    weight: 1.0
     pass_threshold: 1.0
 
   - type: file_exists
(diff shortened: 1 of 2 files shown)


addyGarg:
Split out of UiPath/skills#2322 (skill content + the zero-prerequisite smoke task). Also depends on UiPath/cli#3336 — the integration task reads `playwright-context` output, which a stable CLI only registers once that fix is in the released line.

PR #1979, opened 2026-07-10
fix(uipath-functions): unflake deploy_tenant + in-flow-register coder evals

AlexBizon:
## Problem

The `uipath-functions` coder evals `skill-functions-deploy-tenant` and `skill-functions-in-flow-register` intermittently fail. Both passed clean (score **1.000**) on 2026-07-09 and failed the next day — these are marginal conditions, **not** broken skills or checks.

## Root causes & fixes

**1. `deploy_tenant` — turn-budget cliff (real bug)**
Observed runs consistently use **50–71 turns** against `max_turns: 50` (`expected_turns: 30`). On 07-09 the agent finished the `.nupkg` just before the cap; on 07-10 a slightly slower path hit the wall first → `MAX_TURNS_EXHAUSTED` with the artifact otherwise correct.
→ Raise `expected_turns 30→55`, `max_turns 50→90` to match observed behavior with headroom. Still fully offline/deterministic.

**2. `in_flow_register` — brittle check on an arbitrary filename**
The task graded the literal path `ClaimsSol/ClassifierAgent/main.py`. The skill explicitly states the entry file and function name are arbitrary (the `uipath.json` `functions` map is `<file>:<function>`). On 07-10 the agent named the entry file differently — `entry-points.json` was generated (that criterion passed) but `main.py` didn't exist. This graded a non-contract detail.
→ Replace the `file_exists` with a `run_command` that resolves the registered Python entrypoint from the `functions` map (with a `*.py` fallback). Grades the contract, not the filename.

**3. Skill gap: `authors` (root-cause amplifier)**
`SKILL.md`'s `pyproject.toml` example omitted `authors`, but `uip functions pack` rejects packages with `Project authors cannot be empty` (both checkers enforce it). Agents copied the example, hit the pack error, and burned turns rediscovering it — feeding #1's turn exhaustion. The sibling `uipath-agents` skill already documents this.
→ Add `authors` to the example plus a required-field note.

## Verification

- YAML parses; embedded Python check compiles.
- Coder evals for the functions suite dispatched via the **Run Coder Eval** workflow (link in a follow-up comment).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2405, opened 2026-07-31
fix(uipath-functions): teach required pyproject authors + add Critical Rules

AlexBizon:
## Why

3/12 `uipath-functions` tasks failed the **2026-07-31_04-38-51** nightly on `gpt-5.6-terra`:

| Task | Score |
|---|---|
| `skill-functions-file-attachment-input` | 0.41 |
| `skill-functions-simple-echo` | 0.52 |
| `skill-functions-python-e2e-lifecycle` | 0.64 |

All three failed on the **same single assertion**:

```
FAIL: pyproject.toml has no `authors` entry — `uip function pack` will reject
the package with `Project authors cannot be empty`.
```

Every `command_executed` and `skill_triggered` criterion passed. Turn counts were well under budget (10/35, 13/25, 13/31) — the agent finished cleanly and lost only the `run_command` criterion.

Root cause: five check scripts under `tests/tasks/uipath-functions/` require `authors`, but `SKILL.md` **never mentioned it**. `uip function new` omits the field. Sonnet infers it; `gpt-5.6-terra` does not.

## What changed

1. **`authors` in the Step 5 `pyproject.toml` template**, plus an explicit note that the scaffold omits it.
2. **New `Critical Rules` section.** The skill had none — required by `.claude/rules/skill-structure.md` (§ SKILL.md Body Structure) and flagged as a red flag by `.claude/rules/skill-review.md`.

Deliberate wording choice: `authors` is tied to **scaffolding**, not packing. The three failing tasks never pack, so guidance phrased as "set it before packing" lets an agent correctly conclude it is unnecessary. Rule 2 says to add it as part of scaffolding, "including ones you only run locally."

## Relationship to #1979

#1979 contains the same one-line `authors` template addition. It branched from main ~382 commits back, so it predates main's `uip functions` -> `uip function` doc rename — but it never edited those verb lines, and on uip 1.200.0-dev.8046 the CLI reports usage as `uip function|functions`, i.e. both forms remain live aliases. So there is no functional regression either way; it is a docs-convention difference only. #1979's other two fixes (deploy_tenant turn budget, in_flow_register entrypoint resolution) remain valuable and are not covered here. Recommend rebasing #1979 and dropping only its duplicated `authors` hunk.

Note for whoever reviews #1979: the comment claiming "Ran 5 times all suite and all tests passed" holds for 4 of the 5 linked runs. Run 29086622276 concluded failure at 11/12 — `smoke-trigger/business-days` scored 0.000 on activation recall while the suite itself reported `pass_rate=80% (4/5) gate=PASS`. That red is the verdict-step bug in item 1 below, not a defect in #1979's fixes.

## Verification

Validators pass locally:
- `hooks/validate-skill-descriptions.sh` — 389 chars
- `scripts/check-skill-status.py` — OK, 24 skills
- `scripts/check-cli-verbs.py` — 0 High, 0 Medium (catalog uip 1.200.0-dev.8042)
- Heading hierarchy verified — no skipped levels

Coder-eval on the `uipath-functions` tree pending; the hypothesis is that Critical Rule 2 makes the field impossible to miss, but that needs a run on `gpt-5.6-terra` to confirm.

## Not addressed here

Two separate findings from the same investigation, worth their own issues:

1. **`run-coder-eval.yml` verdict ignores suite gates.** The step exits non-zero unless *every* `task.json` is `SUCCESS`, so a rate-gated recall suite can never be green if one row misses — even when the suite reports `gate=PASS`. Hit this on `smoke-trigger/business-days` (suite `pass_rate=80% (4/5) gate=PASS`, run still red).
2. **No IS-discovery or bindings guidance in `uipath-functions`.** The skill shows `invoke_activity` with `object_path` pre-filled and no route from an activity name to a path; `bindings.json` is mentioned only as a generated filename. The content exists under `uipath-platform` and `uipath-agents`, but the self-contained-skill rule forbids linking to it.

### DIFF
--- a/skills/uipath-functions/SKILL.md
+++ b/skills/uipath-functions/SKILL.md
@@ -44,7 +44,7 @@ Use Python when the logic needs job semantics, platform SDK access, or is invoke
 
 ## Critical Rules
 
-1. **Pass `-l py`.** `uip function new` defaults to TypeScript. Omitting the flag scaffolds the wrong project type.
+1. **Pass `-l py`, then verify the scaffold.** `uip function new` defaults to TypeScript. And when `uipath-langchain` is installed it emits a LangGraph *agent* even with `-l py` — check for `langgraph.json` and replace the scaffold if present (Step 1).
 2. **Add `authors` to `pyproject.toml` right after scaffolding.** The scaffold omits it; `uip function pack` then fails with `Project authors cannot be empty`. Do this as part of scaffolding, not before packing — every project needs it, including ones you only run locally.
 3. **Never instantiate `UiPath()` at module level.** Use a lazy getter (Step 3 template).
 4. **Return errors, never raise them.** Populate `error_type`/`error_message` on the output model; an exception escaping the entrypoint faults the job.
@@ -80,6 +80,14 @@ uip function new <name> --language js       # JavaScript Function (JS/TS, no job
 **`--language py` is required for Python.** The default language is TypeScript — omitting `--language` scaffolds a JS/TS project. Always pass `-l py` or `--language py` when building a Python Coded Function.
 
 `--empty` skips the hello-world function (JS/TS only).
+
+**Verify what you got before writing logic.** `uip function new -l py` passes through to `uipath new`. When `uipath-langchain` is installed in the same environment, its `langgraph_new_middleware` intercepts that command and scaffolds a **LangGraph agent instead of a Function** ([uipath-python#1543](https://github.com/UiPath/uipath-python/issues/1543)). Symptoms: a `langgraph.json` file, a `main.py` importing `langgraph` / `UiPathChat`, and no `functions` map in `uipath.json`.
+
+```bash
+ls langgraph.json 2>/dev/null && echo "WRONG SCAFFOLD — agent, not function"
+```
+
+If you see it: delete `langgraph.json`, replace `main.py` with the Step 3 template, and write `uipath.json` per Step 4. Do not try to adapt the agent template — it carries an LLM call, which Critical Rule 5 forbids. Two edits fix it; iterating on the wrong scaffold burns the whole turn budget.
 
 ### Step 2: Define Function Schema
 
--- a/skills/uipath-functions/SKILL.md
+++ b/skills/uipath-functions/SKILL.md
@@ -42,6 +42,15 @@ Use Python when the logic needs job semantics, platform SDK access, or is invoke
 
 ---
 
+## Critical Rules
+
+1. **Pass `-l py`.** `uip function new` defaults to TypeScript. Omitting the flag scaffolds the wrong project type.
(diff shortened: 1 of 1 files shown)


AlexBizon:
## Second finding: the eval image guarantees the LangGraph scaffold hijack

Coder-eval [run 30615869105](https://github.com/UiPath/skills/actions/runs/30615869105) on this branch came back **10/12**. The two `authors` failures from the nightly (`file-attachment-input`, `e2e-lifecycle`) now pass, but two different tasks failed:

```
skill-functions-simple-echo        MAX_TURNS_EXHAUSTED (40)  score 0.143
  FAIL: main.py is missing `class Input`
skill-functions-tracing-redaction  MAX_TURNS_EXHAUSTED (40)  score 0.000
  FAIL: main.py must import `traced` via `from uipath.tracing import traced`
```

### Root cause

`uip function new -l py` passes through to `uipath new`. When `uipath-langchain` is installed, its `langgraph_new_middleware` intercepts the command and returns `should_continue=False`, so the base Function scaffolding never runs — you get a LangGraph **agent** instead. This is [UiPath/uipath-python#1543](https://github.com/UiPath/uipath-python/issues/1543), open since 2026-04-02.

`tests/docker/Dockerfile:118` installs it globally:

```dockerfile
# Pre-install Python packages needed by coded-agent sandbox fixtures so agents
# don't trigger a live uv-sync download ...
RUN pip install --no-cache-dir "uipath-langchain>=0.9.26"
```

Installed for **coded-agent** fixture speed, but it poisons every **functions** task in the same image. Reproduced locally on uip 1.200.0-dev.8046 — `uip function new echo-probe --language py` emits `langgraph.json` and a `main.py` importing `UiPathChat(model="gpt-4o-mini-2024-07-18")`, with no `uipath.json` and no `functions` map.

So the agent is handed a scaffold with no `class Input`, no `traced`, and no `functions` map. Sometimes it notices and rewrites (task passes); sometimes it iterates on the wrong scaffold until the turn cap. That makes these tasks structurally flaky, independent of this PR.

### What this PR now does about it

Adds a one-line detection check after scaffold plus explicit recovery steps, and folds the warning into Critical Rule 1. A two-edit correction instead of a 40-turn flail.

This is symptom mitigation. The durable fixes are elsewhere:

1. **`#1543` upstream** — add `--type` to `uipath new` so the passthrough can demand a function project.
2. **Eval image** — install `uipath-langchain` into an isolated venv for the coded-agent fixture rather than globally, removing the interaction from the functions suite entirely. Happy to open that as a separate PR against `tests/docker/Dockerfile` if owners agree.

Re-run dispatched: https://github.com/UiPath/skills/actions/runs/30617651902

### Caveat on verification

These runs use `experiments/nightly.yaml`, pinned to `claude-sonnet-4-6`. The original 3 nightly failures were on `gpt-5.6-terra`. So this validates the scaffold fix and non-regression, but the `authors` fix still wants a `gpt-5.6-terra` run to confirm.

PR #2419, opened 2026-07-31
feat(tests): take the eval model from env vars, bump coder_eval to 0.9.6

bai-uipath:
## What

Stop pinning the agent model in the experiment YAMLs, and take it from a repo variable at every entry point that runs an eval. No model id is hardcoded in any of them. Also bumps the `coder_eval` pin to 0.9.6.

Companion to UiPath/coder_eval_uipath#74, which gives the nightly runner one model variable per harness.

## The model pins

Six experiment files pinned `model: claude-sonnet-5`: `default`, `nightly`, `smoke`, `smoke-windows`, `activation`, `skill-comparison-template`. All six are removed.

In the nightly and in `run-coder-eval.yml` those lines were already dead. `--model` is the last layer of coder_eval's config merge and outranks the experiment, so the pin never applied. They were also wrong in principle: the same experiment file runs under codex and antigravity, which cannot use a Claude model.

They are removed rather than templated because coder_eval expands environment variables only in path fields, so a `${VAR}` in `agent.model` would be passed through as a literal model name.

## The gates that had no `--model`

Three entry points relied on the pin, so removing it would have silently repointed them at `$BEDROCK_MODEL`. Each now passes a model explicitly, resolved from `vars.CLAUDE_CODE_MODEL`:

- `smoke-skills.yml`
- `smoke-rpa-skills.yml`
- `activation-gate.yml`, via `activation_gate.py`

There is no literal fallback behind the variable. An unset `AGENT_MODEL` fails the step, guarded before `coder-eval` is invoked, and `activation_gate.py` exits 2. Gating on a model nobody configured produces a verdict about the wrong thing, and for the activation gate specifically the recall baselines are model-specific, so a silent substitution makes the number meaningless rather than merely wrong. Passing the model at the call site keeps that visible where the baselines live.

`$BEDROCK_MODEL` is deliberately not used for this. It is the evaluation-side model, feeding the `llm_judge` grader and the simulated user, and reusing it for the agent would move the grader whenever the agent under test changes.

`run-coder-eval.yml` picks up `CLAUDE_CODE_MODEL` / `CODEX_MODEL` / `ANTIGRAVITY_MODEL` the same way, replacing three hardcoded model names per job, each guarded. Its `agent_model` input still overrides all of them for a one-off run. Antigravity's model moves to `gemini-3.6-flash`.

## Required repo variables

The three gates and the dispatch workflow read repo variables, which are now set on this repo:

```
CLAUDE_CODE_MODEL=claude-sonnet-5
CODEX_MODEL=gpt-5.6-terra
ANTIGRAVITY_MODEL=gemini-3.6-flash
```

These are the exact values the deleted literals carried, so CI behavior is unchanged. `vars`, not `secrets`: nothing here is sensitive, and a secret cannot be read back for an audit.

## coder_eval 0.9.6

`main` has since moved the pin to 0.9.5 on its own, so the delta this PR still carries is 0.9.5 to 0.9.6:

- `feat(criteria)`: criterion path fields accept glob patterns, resolved literal-first and ignore-filtered. The only change here with reach into existing tasks: a literal path still wins, so a path that resolved before resolves the same way.
- `feat(sandbox)`: CLI recording shims generated via `record_cli`.
- `fix(criteria)`: `cli_called` splits clustered short flags and keeps negative numbers positional.
- `fix(evalboard)`: Path-to-GA counts only still-tagged tasks, scored on the runs that actually executed.

Kept in this PR rather than split out, since editing `tests/.coder-eval-version` is one of the infra triggers for the smoke gate and this branch already pays for that run.

The `fix(reports)` change that makes errors count as misses arrived in 0.9.2 and is already on `main` through its own bumps, so it is not part of this PR's delta and no pass-rate drop here is attributable to it.

## The gate went green on an empty run

The model guard started as the first line of the `Run smoke tests` step. That step carries `continue-on-error: true`, and both threshold steps below it read zero tasks as a pass:

- `Enforce LLM reviewer score threshold`: `if not run_dirs: sys.exit(0)`
- `Enforce pass-rate threshold (>= 95%)`: `if n == 0: print('No smoke-tagged tasks matched — skipping rate check'); sys.exit(0)`

So when run 30672478895 resolved `AGENT_MODEL` empty (queued 90 seconds before the repo variables were created), the guard fired correctly, the step exited in under a second having run nothing, and the job still reported success:

```
AGENT_MODEL:
...sh: line 1: AGENT_MODEL: unset — set the CLAUDE_CODE_MODEL repo variable
step "Run smoke tests"   23:20:02Z → 23:20:02Z   (0s)
job conclusion: success
```

The RPA gate does not have this hole (`if not task_jsons: print('::error::No task.json files found — nothing to gate'); sys.exit(1)`), which is why RPA went red off the same empty variable while Linux went green. That asymmetry was the bug, not the guard.

Two changes:

- The guard moves into its own `Require an agent model` step ahead of the smoke step in both workflows, so an unresolved model fails before any task runs, in a step that names what is missing.
- The Linux pass-rate gate fails closed when zero tasks land **and** the smoke step's outcome is `failure`. A legitimately empty selection still passes, since that path leaves the smoke step successful, and flake tolerance is untouched: once any task runs, the rate check governs exactly as before.

## Validation

The live-run evidence below was gathered at 0.9.2, before `main` advanced the pin. It validates model resolution, which none of 0.9.3 through 0.9.6 touches; version numbers in this section are what those runs actually executed on.

**Live runs on coder_eval 0.9.2, from this branch**, one task per harness (`uipath-admin/audit_events_basic_smoke`, tempdir driver, `experiments/default.yaml` with its pin removed):

| run | agent model | evaluation model | result | cost |
|---|---|---|---|---|
| claude-code, `--model` given | `eu.anthropic.claude-sonnet-5` | `eu.anthropic.claude-sonnet-4-6` | SUCCESS 1.0 | $0.375 |
| claude-code, no `--model` | `claude-sonnet-4-6` | `eu.anthropic.claude-sonnet-4-6` | SUCCESS 1.0 | $0.503 |
| codex | `gpt-5.6-terra` | `eu.anthropic.claude-sonnet-4-6` | SUCCESS 1.0 | $0.195 |
| antigravity | `gemini-3.6-flash` | `eu.anthropic.claude-sonnet-4-6` | SUCCESS 1.0 | $0.182 |

All four load the experiment files with no `model:` key and pass. The runs record `framework_version: 0.9.2`, confirming the pin resolves.

Row 1 versus row 2 is the point of the change. With `--model`, the agent under test and the evaluation route are two different models in the same run. Without it, they collapse onto `$BEDROCK_MODEL`.

**Known consequence for bare local runs.** Row 2 is a real change: `coder-eval run` with no `--model` used to get `claude-sonnet-5` from the experiment file and now gets `$BEDROCK_MODEL` from `.env`. Intended, since that is the environment-variable mechanism, but anyone with an unusual local `BEDROCK_MODEL` will see their model move.

**Linux smoke gate.** 40 tasks under the docker driver on coder_eval 0.9.2, with `experiments/smoke.yaml` carrying no `model:` key and `--model "$AGENT_MODEL"` supplying it. **Pass rate 97.5% (39/40)**, above the 95% threshold. Editing `tests/.coder-eval-version` is itself one of the infra triggers for that gate, so the pin bump got a full run rather than the changed-skills subset.

That run measured `e629efeea`, while the workflows still carried a `|| 'claude-sonnet-5'` literal behind the variable. The commit that removed the fallback got a smoke run that executed **zero tasks and still reported green** (below).

**Windows RPA gate.** `skill-rpa-coded-test-case` failed once with `MAX_TURNS_EXHAUSTED` at 998s. The same task passed 3/3 on other branches the same day at 1128 to 1301s, `vars.CLAUDE_CODE_MODEL` was not yet set so the model was byte-identical to the deleted pin, and the 0.9.1 to 0.9.2 diff moves no dependency or CLI pin and touches the Claude agent only through LiteLLM paths that are inert on Bedrock. Re-ran: passed.

**activation_gate.py.** Verified the constructed command carries `--model <AGENT_MODEL>`, and that an unset variable exits 2 before spawning coder-eval. The activation gate does not fire on this PR, since no skill frontmatter changed.

**Companion.** UiPath/coder_eval_uipath#74 is green, with the per-harness resolution matrix unit-tested against the shipped `daily.sh` block, a denylist test that fails if a literal model id reappears in any of the three mirrors, and pwsh parity verified. Its three ADO dry runs then resolved `eu.anthropic.claude-sonnet-5` / `gpt-5.6-terra` / `gemini-3.6-flash`, one per harness, off the variable group, and the antigravity build carried both the Linux and Windows slices to green including the codex and antigravity pre-flight smokes, which read `CODEX_MODEL` / `ANTIGRAVITY_MODEL` with no fallback.

One thing the smoke run exposed: the gate logs the command as `--model ***`. GitHub masks a value only when it matches a registered secret, so `secrets.BEDROCK_MODEL` holds a plain model name. That is the concrete version of the argument for moving model config to `vars`: nothing is protected (the model name is in `run.json` and on the evalboard either way), the value cannot be read back for an audit, and the smoke log can no longer tell you which model it graded. Out of scope here, but it is why the new variables use `vars` rather than `secrets`.


bai-uipath:
## ADO dry run succeeded on 0.9.6

[Build 20260811.11](https://uipath.visualstudio.com/b57ec1c2-3ae5-4bbb-a205-1effb5e365eb/_build/results?buildId=12998791), queued on UiPath/coder_eval_uipath#74's branch with `skillsRef` pointed at this one and `coderEvalVersion=pinned`, so the version came through `tests/.coder-eval-version` rather than an override. All four stages green, including the merge/publish stage; a dry run gates its own exit code, so this is a real pass rather than a skipped gate. Published to `runs-dryrun`, no Slack, isolated from front-page metrics and maturity.

| slice | run | ok |
|---|---|---|
| skills subset (Linux, docker) | 24 | 24 |
| activation (nested) | 26 | 25 |
| harness smoke — codex | 1 | 1 |
| harness smoke — antigravity | 1 | 1 |
| Windows slice (tempdir) | — | succeeded |

What it confirms, read off `task.json`:

```
environment_info.coder_eval        = 0.9.6
task_config.resolved.agent.model   = eu.anthropic.claude-sonnet-5
```

The experiment YAML carries no `model:` key and the resolved config still has one, so the removal here and the per-harness variable in #74 work together rather than either one silently covering for the other. All three harnesses resolved from the variable group with no fallback behind them: claude-code `eu.anthropic.claude-sonnet-5`, codex `gpt-5.6-terra`, antigravity `gemini-3.6-flash`. The codex and antigravity pre-flight smokes are the ones that would have hard-aborted if `CODEX_MODEL` / `ANTIGRAVITY_MODEL` were unset, and both passed.

The single error is `skill-activation/uipath-planner-069`, "Agent turn timed out after 120s (iteration 1)" — the known activation turn-timeout that lands on a different row each night, unrelated to the model change or the version bump.

One note on this branch since the last review: `main` had moved 113 commits ahead and the only conflict was `tests/.coder-eval-version`, resolved to `main`'s 0.9.5 and then bumped to 0.9.6. The delta this PR carries is therefore 0.9.5 to 0.9.6, not 0.9.1 to 0.9.2 — in particular the `fix(reports)` change that makes errors count as misses arrived in 0.9.2 and is already on `main`, so no pass-rate movement is attributable here.

uipreliga (approved):
🚢

PR #2421, opened 2026-08-01
fix(tests,skills): de-flake nightly guardrail failures — valid fixtures, doc gaps, run_limits

apetraru-uipath:
## What changed?

De-flakes the guardrail-tagged failures of nightly run `2026-07-28_04-16-10` (all tagged `turn-timeout-too-short`). Transcript analysis showed each timeout was a symptom of a concrete defect — fixture schema noise, a never-seeded test resource, or a missing skill-doc branch — not model slowness. Three change areas:

### 1. `tests/tasks/uipath-review/` — fixtures now pass `uip agent validate` (`5e170b1ab`)

5 of 7 lowcode review fixtures failed `uip agent validate` on main, so reviewers saw "review: Grade A" vs "validate: hard errors" and burned 15+ min investigating harness noise instead of the seeded guardrail defect.

- `_shared/lowcode_scaffold.py`: new `content_tokens()` / `set_message()` (rebuild `contentTokens` to match patched message content), `connection_binding()` / `process_binding()` / `write_bindings()` (emit the packager-shaped `bindings_v2` entries), and the scaffold now writes root `bindings_v2.json` like a real `uip agent init` project.
- All 7 inject scripts updated to use the helpers; `inject_guardrail_audit_logging.py` additionally gets a schema-valid integration tool resource (`iconUrl`, `argumentProperties`, `properties.method/connection/parameters/bodyStructure`).
- Result: **all 8 lowcode review fixtures return `Status=Valid` straight from injection** (uip 1.200.0), with no reliance on the `refresh`-before-`validate` workaround.

### 2. `escalation_app_validation` — the schema trap is now real and deterministic (`f81ce1617`, `eec9e91cf`)

The task grades "fetch the app's action schema → detect the mismatch → reject", but `EscalationWorksApp` was never seeded in any tenant — the trap never fired: passes rejected for the wrong reason ("not found") and the nightly failure was the model hunting the missing resource for ~18 min.

- New `_fixtures/ValidationSol/`: solution shell + `EscalationWorksApp` as a **solution-internal resource** (app / appVersion / package / process-webApp JSONs, fixed GUIDs) with a deliberately incompatible `actionSchema` (generic `input`/`output`; `Submit`/`Reject` outcomes). No tenant state — fully deterministic.
- `escalation_app_validation.yaml`: `pre_run` copies the fixture; prompt becomes brownfield ("solution exists, do NOT re-create it") and bans `uip solution resources refresh` (refresh destroys the trap either way — it deletes unregistered local resources after a project registration, or regenerates them and wipes the authored schema when a backing project is registered); the advisory `uip solution init` criterion becomes a gating `command_not_executed` (same pattern as `local_escalation`); `run_limits` gain headroom (`max_turns` 40→60, `turn_timeout` 1200→1800, `+task_timeout: 2400`).

### 3. `skills/` — two doc gaps that caused thinking rabbit holes (`bece3e061`, `ba3303858`, part of `eec9e91cf`)

- `uipath-agents/.../escalation.md` (+4 lines): terminal-stop rule when the named app is not in the resources list / `resources get` fails (no retries, alternate endpoints, or re-auth; never wire an escalation against an unresolved app), and a note that `--kind`/`--source`/`--search` filters currently drop `Source: Local` rows — use the **unfiltered** `uip solution resources list` and pass the **Key** (not the name) to `resources get`.
- `uipath-review/.../coded-guardrails-review.md` (+8 lines): in the Actionability check, the catalog verdict IS the evidence — do not verify SDK behavior empirically (importing/executing project dependencies, reading SDK sources, fetching SDK docs); write the finding and the report.

Out of scope: `skill-case-artifact-safety-guard` (uipath-maestro-case) also failed that night but is a different feature that coincidentally carries the `guardrail` tag — intentionally untouched.

## CLI issues surfaced (not fixable in this repo)

1. `uip solution resources list` with any filter (`--kind`, `--source local`, `--search <name>`) drops `Source: Local` rows; `resources get <name>` fails for local resources (Key-only). Worked around in `escalation.md`; worth a CLI fix.
2. `uip solution resources refresh` deletes local resources with no registered backing project, and regenerates (wiping authored `actionSchema`) when one is registered. Mitigated via the task prompt; worth a CLI look.

## How has this been tested?

- All 8 lowcode review fixtures scaffolded from this branch validate `Status=Valid` on uip 1.200.0 (verified via CLI, no LLM involved).
- Escalation fixture verified via CLI: app appears in the unfiltered `resources list` as `Source=Local`; `resources get <key>` returns the incompatible schema; `uip agent init ValidationSol/ReviewAgent` registers cleanly alongside it.
- Final `escalation_app_validation.yaml` verified on both harnesses (Coder Eval v0.9.1, tempdir driver, isolated worktrees): **Claude Sonnet 5 SUCCESS 1.0 (243s)** and **Codex gpt-5.6-terra SUCCESS 1.0 (107s)** — both rejecting **for the right reason**, e.g. Terra: "schema only exposes `input`, `output`, `Submit`/`Reject`; guardrails require the named review inputs/outputs and `Approve`/`Reject` outcomes."
- uipath-review guardrail tasks (action-ineffective, audit-logging, content-safety, misapplied): SUCCESS 1.0 on Claude Sonnet 5 across two independent replicates each, and on Codex gpt-5.6-terra.
- `git diff --check`, `python3 -m py_compile` on all edited injects/scaffold, `scripts/check-skill-status.py` clean; `check-skill-verbs.py` findings on `escalation.md` are pre-existing (verified via stash).

- [x] None
- [ ] Under Feature Flag
- [ ] DB migrations
- [ ] API removals


skills/uipath-agents/references/lowcode/capabilities/escalation/escalation.md
```diff
@@ -50,6 +50,10 @@ Filter the result for entries whose `Type` is `"Workflow Action"` (Coded / Coded
 
 `Key` identifies the backing app. `resource list` does not return `systemName`, `deployVersion`, or the action schema — fetch those in Step 3 with `uip solution resources get`.
 
+**Solution-internal apps require the UNFILTERED list.** `--kind`, `--source`, and `--search` filters currently apply only to the remote catalog — a `Source: "Local"` app is dropped by every filtered query and appears only in a plain `uip solution resources list --output json`. Run the unfiltered list (from the solution directory) before concluding an app is absent, and always pass the row's `Key` — not the name — to `resources get` (name lookups fail for solution-internal apps).
```
andreibalas-uipath: We should raise this with the `uip solution` owners - it sounds like unintended behavior.

tests/tasks/uipath-agents/lowcode/guardrails/escalation_app_validation/escalation_app_validation.yaml
```diff
@@ -1,32 +1,42 @@
...
+  personal data (email addresses) and, instead of blocking, escalates flagged
+  cases to a human reviewer using our "EscalationWorksApp" Action Center app,
+  which is already part of this solution.
 
-  Do NOT publish, upload, or deploy. Do NOT ask for approval,
-  confirmation, or feedback. Do NOT pause between planning and
-  implementation. Do NOT write a plan file. Complete the entire task
-  end-to-end in a single pass.
+  Do NOT publish, upload, or deploy. Do NOT run
+  `uip solution resources refresh` — the solution resources are already in
```
andreibalas-uipath: The eval prompt needs to be humanized, since it is leaking specific skill implementation details. What I would drop:
* explicit "do not re-create the solution";
* explicit "do not run `uip solution resources refresh`" - in fact, this should be safe to run.

valentinabojan (approved):

ctiliescuuipath (approved):

skills/uipath-agents/references/lowcode/capabilities/escalation/escalation.md
apetraru-uipath: FYI @alexenica , your opinion is needed here

apetraru-uipath: Raised with the `uip solution` owners — outcome:

- **`resources list` dropping local rows on `--kind` filters** — confirmed as a bug, fix in progress on their side.
- **`resources refresh` deleting/regenerating local resources** — confirmed **expected** behaviour, not a bug.
- **`resources get <name>`** — never worked; Key-based lookup is the only supported form. Worth noting `escalation.md` Step 3 already says *"Pass the `Key` from Step 2"*, so there's nothing to fix here.

**I've dropped the whole `escalation.md` change from this PR** — the file no longer appears in the diff.

Two reasons beyond the pending CLI fix (which would have made a workaround note stale on arrival):

1. It was largely **redundant** — Step 2 already documents this at lines 32–37: *"`--kind` and `--search` only work with `--source remote`; list everything, filter `.Data[]` client-side by `Kind == "App"`."*
2. My wording **overclaimed** — I wrote that `--source` also drops local rows, but only `--kind` is confirmed, and a passing run shows `--source local` returning the row fine. Behaviour varied across CLI builds during testing, which is itself a reason not to pin it in a skill doc.

The second paragraph I'd added (terminal-stop when the app isn't found) is also gone: this PR now seeds the app in the test fixture, so the not-found path is no longer exercised here, and keeping the rule would have been unverified scope creep in a non-guardrails area.

tests/tasks/uipath-agents/lowcode/guardrails/escalation_app_validation/escalation_app_validation.yaml
apetraru-uipath: Both dropped — agreed on the principle, and the repo's own rule says the same (`.claude/rules/test-writing.md`: prompts stay minimal, anti-pattern "hand-holding prompts").

**`do not re-create the solution`** — gone. The prompt now only states the solution exists (*"inside the existing `ValidationSol` solution in the current directory"*), matching the sibling `lowcode/local_escalation` task, which states "existing" and relies on its gating `command_not_executed` criterion rather than an explicit prohibition. The criterion is unchanged.

**`do not run uip solution resources refresh`** — gone. You were right that it shouldn't be banned, and the `uip solution` owners confirmed the delete/regenerate behaviour on local resources is expected.

One thing that surfaced from dropping it, worth flagging since it changed the design: with no refresh ban, **Claude Sonnet 5 detected the schema mismatch correctly and then edited the fixture's `actionSchema` to satisfy the contract and wired the escalation anyway** (score 0.62) — *"Since this app is local to the solution (not an external deployed app), I extended its `actionSchema` to satisfy that contract."*

That was a hole in my fixture, not a model bug: making the app solution-internal put the trap in a file the agent can rewrite, which the tenant-hosted design never allowed. Rather than reinstating CLI-level prohibitions, I addressed it at the two right layers:

- **Checker** now asserts the app's action schema still carries the incompatible contract (`inputs=[input]`, `outputs=[output]`, `outcomes=[Submit,Reject]`) and fails with an explanation otherwise. Rewriting it is a wrong answer rather than an alternative solution — `actionSchema` is *derived* from the published app design (`resources refresh` regenerates it, and did wipe the patch mid-run), so editing the solution resource file doesn't change what the real app exposes; the escalation would still fail at runtime.
- **Prompt** gains one domain-level line: *"That app is published and owned by another team — treat it as read-only."* Ownership context rather than CLI/skill mechanics, matching the read-only contract the `uipath-review` tasks already use.

Verified after the change: **Claude Sonnet 5 SUCCESS 1.0 (576s)** — restored the fixture byte-for-byte instead of patching it — and **Codex gpt-5.6-terra SUCCESS 1.0 (228s)**, both leaving the trap intact and rejecting for the right reason. The tamper assertion was also tested directly against Sonnet's real tampered artifact (fails), a pristine fixture with no guardrail (passes), and a pristine fixture with the guardrail wired (fails).

apetraru-uipath:
## CI status: only pre-existing RPA failures remain — unrelated to this PR

Latest run on `47b8b8618`: **20 of 21 checks green.** The one red check is `Run skill smoke tests` at **90.5% (19/21)**, below the ≥95% gate, from two failing tasks:

| Task | Status | Owner area |
|---|---|---|
| `skill-review-rpa-coded-structural` | `ERROR` score 0.00 | `tests/tasks/uipath-review/rpa/coded-structural/` |
| `skill-review-rpa-selector-brittle` | `ERROR` score 0.00 | `tests/tasks/uipath-review/rpa/selector-brittle/` |

### Why these are not caused by this PR

1. **This PR touches zero files under `tests/tasks/uipath-review/rpa/`** — `git diff origin/main --name-only` contains no RPA path.
2. **Neither task depends on anything changed here.** The only shared file I modified is `tests/tasks/uipath-review/_shared/lowcode_scaffold.py`; `grep -rn 'lowcode_scaffold' tests/tasks/uipath-review/rpa/` returns nothing. Both RPA tasks drive their own fixtures and checkers.
3. **They have failed in 3/3 CI runs of this PR**, including before the changes currently under review — deterministic breakage, not variance:

| Run | Pass rate | Failures |
|---|---|---|
| `30811868661` (Aug 3) | 95.7% (45/47) | these same 2 RPA tasks |
| `31003417222` (Aug 5) | 81.0% (17/21) | these 2 + `coded-error-handling`, `stale-generated-files` |
| `31007516539` (Aug 5, current) | 90.5% (19/21) | **only** these 2 |

`skill-review-agents-coded-error-handling` (`Turn timeout (1800s) fired`) and `skill-review-agents-lowcode-stale-generated-files` failed once and **pass in the current run** — those were genuine flakes and are resolved. I re-ran failed jobs once rather than repeatedly; I have not re-run the RPA pair, since 3/3 identical failures indicate real breakage that a re-run won't clear.

### Why the gate can't go green from inside this PR

The smoke suite scopes to changed skills, so this run covered **21 tasks** (only `uipath-review` changed). At that denominator the ≥95% threshold tolerates exactly **one** failure — 1 → 95.2% (pass), 2 → 90.5% (fail). With two chronically-broken RPA tasks the ceiling is 90.5% regardless of this PR's contents.

Unblocking needs one of:
- the owners of those two RPA review tasks fixing them (out of scope here — this PR is guardrails-scoped, and I'd rather not touch another sub-area's tasks), or
- an admin override / gate adjustment for this PR.

I could not extract a root cause beyond `status=ERROR score=0.00` — unlike `coded-error-handling`, the job log carries no timeout or crash line for these two, so diagnosing them will need the per-task artifacts from the run.

### Verification this PR does carry

- All 8 lowcode `uipath-review` fixtures return `Status=Valid` straight from injection on uip 1.200.0 (verified via CLI, no model involved) — 5 of them failed `uip agent validate` before this PR.
- `escalation_app_validation` verified on both harnesses against the final YAML: **Claude Sonnet 5 SUCCESS 1.0**, **Codex gpt-5.6-terra SUCCESS 1.0**, both rejecting the incompatible app for the right reason with the fixture's trap left intact.

andreibalas-uipath (approved):

roalexandru (approved):

apetraru-uipath:
## Correction to my previous status comment — the failure set is flaky, not two fixed tasks

A fourth run (`31016535260`, on rebased head `387c7ef81`) came back at the same **90.5% (19/21)** but with a **different pair** of failures, which invalidates part of my earlier analysis. Correcting it rather than leaving the record wrong:

| Task | Aug 3 | Run 1 | Run 2 | Run 3 (rebased) | Fail rate |
|---|---|---|---|---|---|
| `skill-review-rpa-coded-structural` | FAIL | FAIL | FAIL | FAIL | **4/4** |
| `skill-review-rpa-selector-brittle` | FAIL | FAIL | FAIL | **pass** | 3/4 |
| `skill-review-agents-lowcode-stale-generated-files` | pass | FAIL | pass | **FAIL** | 2/4 |
| `skill-review-agents-coded-error-handling` | pass | FAIL | pass | pass | 1/4 |

**What I got wrong:**
- I described both RPA tasks as deterministic breakage on 3/3 evidence. `selector-brittle` **passed** this run, so it is flaky (3/4), not hard-broken. Only `rpa-coded-structural` is consistent at 4/4.
- I said `stale-generated-files` was "a genuine flake and resolved" after it passed run 2. It **failed again** in run 3, so it is unresolved and flaky (2/4).

**What holds:** none of these four are caused by this PR. The evidence is unchanged — zero files under `tests/tasks/uipath-review/rpa/` in the diff, no `lowcode_scaffold` reference anywhere under `rpa/`, and every one of them has both passed and failed on code paths this PR does not touch. `rpa-coded-structural` also failed on Aug 3, before any of the changes now under review.

**Practical consequence:** the suite draws ~2 failures per run from a pool of 4 unreliable tasks, while the ≥95% gate at a 21-task scope tolerates exactly **one**. Passing requires a run where three of the four happen to pass at once — roughly a 1-in-10 outcome on these rates. I have stopped re-running: four cycles at ~35–50 min each is not a productive path to green, and a lucky green run would not mean anything was fixed.

This PR needs either an owner fix for `rpa-coded-structural` (consistently broken) plus stabilisation of the flaky three, or a gate override. Both are outside this PR's guardrails scope.

PR #2425, opened 2026-08-02
Fix uipath-project-discovery-agent: avoid cmd.exe/PowerShell dependency when invoking uip CLI

nsandrosBDO:
## Problem

`uipath-project-discovery-agent` (`skills/uipath-rpa/agents/uipath-project-discovery-agent.md`) declares `tools: Bash, Read, Glob, Grep` and, in its workflow, shells out directly to the `uip` CLI:

- Step 1 ("Locate the Project"): `uip rpa list-instances --format json`
- "UILibrary NuGet Packages" discovery step: `uip rpa inspect-package --package-name <PackageName>`

On Windows, the globally-installed `uip` command resolves to a `uip.cmd`/`uip.ps1` **shell wrapper** that npm generates around the CLI's real Node.js entry point (`@uipath/cli` on npm declares `"bin": { "uip": "dist/index.js" }`). Invoking `uip ...` through the Bash tool therefore requires spawning `cmd.exe` or `powershell.exe` to interpret that wrapper script.

In hosts/environments where direct shell-interpreter execution (cmd.exe/PowerShell) is restricted or blocked by policy, these steps fail outright — even though the underlying `uip` CLI is itself a plain Node.js program with no OS-shell dependency at all.

## Fix

- Added an "Invoking the `uip` CLI" section documenting that any `uip` functionality this agent needs should be run by resolving the CLI's Node.js entry point (`@uipath/cli`'s `dist/index.js`, via an env var override, local `node_modules`, or the global npm root) and invoking it directly with `node <entry.js> <args>` — never through the `uip`/`uip.cmd`/`uip.ps1` shell wrapper.
- Made Step 1's Studio Desktop instance lookup (`rpa list-instances`) explicitly best-effort: if the CLI entry can't be resolved/invoked (no Node reachable, no Bash tool, blocked shell, etc.), the agent skips it and falls back to the explicit path / current working directory instead of failing.
- Applied the same graceful-skip treatment to the UILibrary NuGet package inspection step (`rpa inspect-package`) — it's an optional enhancement, not a hard requirement, and now degrades to just recording the package name/version from `project.json`.
- Added a short caveat to the generated "Quick Reference" section clarifying that its example `uip` commands are for humans/other tooling, not something the discovery agent itself depends on for shell access.

## Why this matters

Some hosts/orgs restrict direct `cmd.exe`/`powershell.exe` execution by AI agents for security reasons. Since `uip` is fundamentally a Node.js CLI, the agent doesn't need a shell interpreter at all to use it — it only needed one because of how npm's global bin shims work on Windows. This change removes that unnecessary hard dependency while keeping full functionality when a shell/Node path is available, and degrading gracefully (rather than failing) when it isn't.

This mirrors an existing working pattern from a downstream UiPath coded-app project, which already resolves `@uipath/uipath-ts-cli`'s `dist/cli.js` and invokes it via `node`/`process.execPath` directly (`shell: false`) instead of through the `uip`/`.cmd`/`.ps1` wrapper.

## Scope

Documentation-only change (the "agent" is an LLM subagent prompt, not executable code) — no build/tests apply.

### DIFF
--- a/skills/uipath-rpa/agents/uipath-project-discovery-agent.md
+++ b/skills/uipath-rpa/agents/uipath-project-discovery-agent.md
@@ -24,10 +24,31 @@ You are a project discovery agent. Analyze a UiPath automation project and gener
 
 ## Workflow
 
+### Invoking the `uip` CLI (when needed)
+
+A couple of steps below use the `uip` CLI (`@uipath/cli` on npm) as an optional enhancement. On
+Windows, the global `uip` command is a `uip.cmd`/`uip.ps1` **shell wrapper** that npm generates
+around the CLI's real Node.js entry point (`bin: { uip: "dist/index.js" }`). Running `uip ...`
+through the Bash tool spawns `cmd.exe`/`powershell.exe` to interpret that wrapper — which is
+unavailable or blocked in some hosts/policies, even though the CLI itself is plain Node.js.
+
+Prefer invoking the CLI's entry point directly with `node`, bypassing the shell wrapper entirely:
+
+1. Resolve the entry file: check `UIPATH_TS_CLI_PATH`-style env vars first, then
+   `node_modules/@uipath/cli/dist/index.js` relative to the project, then the global npm root
+   (`npm root -g`) equivalent, e.g. `.../node_modules/@uipath/cli/dist/index.js`.
+2. Run it as `node <resolved-entry.js> <args...>` (no `cmd /c`, no `powershell -Command`, no shell
+   string interpolation).
+3. If no entry point can be resolved, or `node`/Bash isn't available at all, **skip the step** —
+   every `uip`-based step in this workflow is an optional enhancement with a documented fallback,
+   never a hard requirement.
+
 ### Step 1: Locate the Project
 
 1. If the user provided an explicit path, use it
-2. Try `uip rpa list-instances --format json` to find an open Studio Desktop project
+2. Optionally try locating an open Studio Desktop project via `rpa list-instances --format json`,
+   invoking the CLI's Node entry directly per "Invoking the `uip` CLI" above. If it can't be
+   resolved or invoked, skip this and continue — it is a best-effort enhancement, not required.
 3. Fall back to current working directory
 4. Verify `project.json` exists and contains UiPath dependencies before proceeding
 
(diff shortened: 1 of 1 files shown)


RaduAna-Maria:
Hey @nsandrosBDO, could you share a concrete repro — host, OS, shell, and the verbatim error? If there's a real failure here I'd rather chase the actual cause.

Two other thoughts in the meantime:

- If what you hit was a *permission* block rather than a shell one, `commands/install-permissions.md` (`/install-permissions`) is the intended fix — it allowlists `Bash(uip rpa *)`.
- `uip rpa` appears in ~190 places across the repo (SKILL.md, the reference guides, the other skills). Even if the shell dependency were real, discovery would skip gracefully and then the parent RPA skill would still fail on the next `uip rpa validate`. So this would want to live once in `references/environment-setup.md` rather than inline here.

Also a heads-up: the branch has drifted since you opened it — main renamed `list-instances --format json` → `instances list --output json`, `run-file` → `run`, and `get-errors` → `validate`, so it currently conflicts.

skills/uipath-rpa/agents/uipath-project-discovery-agent.md
```diff
@@ -24,10 +24,31 @@ You are a project discovery agent. Analyze a UiPath automation project and gener
...
+Windows, the global `uip` command is a `uip.cmd`/`uip.ps1` **shell wrapper** that npm generates
+around the CLI's real Node.js entry point (`bin: { uip: "dist/index.js" }`). Running `uip ...`
+through the Bash tool spawns `cmd.exe`/`powershell.exe` to interpret that wrapper — which is
+unavailable or blocked in some hosts/policies, even though the CLI itself is plain Node.js.
+
+Prefer invoking the CLI's entry point directly with `node`, bypassing the shell wrapper entirely:
+
+1. Resolve the entry file: check `UIPATH_TS_CLI_PATH`-style env vars first, then
+   `node_modules/@uipath/cli/dist/index.js` relative to the project, then the global npm root
+   (`npm root -g`) equivalent, e.g. `.../node_modules/@uipath/cli/dist/index.js`.
```
RaduAna-Maria, line 39: Small snag with this fallback: on Windows `npm` is itself an `npm.cmd` shim, so `npm root -g` would go through the same wrapper this section is routing around.

PR #2441, opened 2026-08-04
docs(uipath-ixp): occurrence is a stable document-position id (ACTV-89901)

cezara98t:
> ⚠️ **DO NOT MERGE until ACTV-89901 is deployed to production.** This reverses the skill's occurrence-instability guidance to match the span-based occurrence numbering in [DU-App #35685](https://github.com/UiPath/DU-App/pull/35685). Merging before that fix ships would tell agents occurrences are stable while production still renumbers on write — causing mis-targeted per-occurrence confirms. Kept as a **draft** for that reason.

## What

The skill currently documents the occurrence-numbering bug (the server lists annotation↔prediction matched pairs first, so confirming a row moves it to `Occurrence` 0 and renumbers the rest → "re-read between writes / never carry an index across a write"). **ACTV-89901** fixes that: `Occurrence` is now the row's **document-position rank** (page → top-to-bottom), derived from its span and **stable across confirm / correct / unconfirm**. This PR updates the skill to match.

## Changes

- **cli-reference.md** — the `get-predictions` row; the `### Occurrence numbering` section (was "…and read order"); the confirm/unconfirm `--occurrence` rows.
- **SKILL.md** — Critical Rule 18 rewritten to "stable document-position id"; Rule 13 + the two per-occurrence task-nav rows; **removed** the two now-obsolete "renumbered on write" Common Pitfalls rows.
- **label-documents-guide.md** — the `### Occurrence numbering` section + the per-occurrence confirm/unconfirm guidance.

`--updates` stays the recommended batch form (atomicity, fewer round-trips) — just no longer required to keep indices valid.

## Notes

- Single source per the skill-update process (no manual copies); no CLI code change needed (its help already said "document order", which the fix now honors).
- Two pre-existing markdownlint warnings (MD029 line ~84, MD060 line ~198) are outside this diff and left untouched.

ACTV-89901


cezara98t:
Closing — this reverses correct guidance.

#2312 already documents the right behavior: `Occurrence` is a **per-read positional index** (confirmed rows sort first, and it reorders on write), not a stable id. This draft claims it's a "stable document-position id," which is wrong — telling agents the index is stable would lead them to carry it across writes and mis-target rows.

Confirmation is span-keyed and the occurrence is never stored, so #2312's read-scoped / batch-into-one-`--updates` guidance on `main` is correct as-is and should stay.

PR #2447, opened 2026-08-04
feat(uipath-coded-apps): document the in-memory Validation Station paths

cata-on:
## What's wrong

Every data path on `references/widgets/validation-station.md` reaches the document through a storage bucket — Action App, Web App, and the subcomponents section, which describes itself as sharing "the same bucket artifacts". A coded app whose backend already holds the taxonomy, extraction result and document therefore has no documented way to use the widget.

Given exactly that situation, the assistant abandons the widget and hand-rolls the UI with `react-pdf` — the one thing §"When to Use" forbids. From its own generated file:

> Persistence is NOT delegated to the DU Validation Station widget (that widget owns its own bucket-backed save flow). This app owns persistence itself

Both packages already support the case. `@uipath/du-validation-station-wc` calls its standalone element "the recommended variant for external consumers — your backend stays in control of all I/O", and that element is named nowhere in this repo.

## What changed

A new section, **Integration: in-memory data (no storage bucket)**, for the two shapes that work:

- **Subcomponents, pre-fetched.** `SubcomponentDataSource` takes `artifacts` + `documentId` with `sdk`/`data` omitted, and then issues no request; `BucketArtifacts` is a plain object a host can build. Worth documenting: without an SDK context `canPersist` is false, so `onSubmitComplete`/`onSaveAsDraftComplete` never fire and the raw `onSaveValidatedDataRequest` family is the only signal.
- **`<ui-du-validation-station-standalone-wc-element>`** — the only in-memory route to the all-in-one layout, since `ValidationStationProps` requires `sdk` and `data`. Covers `documentId` first ("Required before any data inputs are set"), the six data inputs, the three host-owned save events, and element registration.

**Critical Rules change, per CONTRIBUTING.** Rule 6 required `OR.Buckets` unconditionally. That is false for a path issuing no HTTP call, and it costs a consumer a scope, an External App and a bucket they cannot use. The rule now splits the two scopes: `OR.Buckets` depends on the data path, while **`OR.Tasks` is required on either path** whenever the app completes an Action Center task on save — and a web app that completes a task still needs its own `UiPath` instance to do so. Nothing else in the section is touched. `create-web-app.md` carried the same unconditional claim and is corrected the same way, and two bucket-implying lines in the subcomponents section are narrowed.

## How to check

New eval task. Same prompt and agent; only the skill differs:

```
cd tests && make install
SKILLS_REPO_PATH=$(cd .. && pwd) .venv/bin/coder-eval run \
  tasks/uipath-coded-apps/ui_validation_station_in_memory_smoke.yaml \
  -e experiments/default.yaml
```

8/8, weighted 1.000 on this branch; 3/8, 0.314 with `SKILLS_REPO_PATH` on an unmodified `main`. Criteria grade generated source, not a self-report.

`npm run version:check`, `check-skill-status.py` and `check-cli-verbs.py` pass; `check-skill-verbs.py` reports the same two stale findings as `main`, both outside the edited file. Every name in the section was checked against the shipped types of `@uipath/du-validation-station-wc@1.0.0-rc.1` and `@uipath/ui-widgets-validation-station@1.0.1`, and both examples compile under `tsc --strict` with React 19.

cc @Raina451 @deepeshrai-tech @Sandeepan-Ghosh-0312 @swati354

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/uipath-coded-apps/references/create-web-app.md
+++ b/skills/uipath-coded-apps/references/create-web-app.md
@@ -225,7 +225,7 @@ const records = await entities.getAllRecords('<entity-id>'); // entity ID is a U
 
 Method signatures come from the installed types (`node_modules/@uipath/uipath-typescript/dist/<subpath>/index.d.ts`); per-method scopes from the shipped `node_modules/@uipath/uipath-typescript/docs/oauth-scopes.md` — see [oauth-scopes.md](oauth-scopes.md) for bundles and the lookup protocol.
 
-If the user wants a **Document Understanding validation UI** (review/correct extraction results), embed the Validation Station widget — see [widgets/validation-station.md](widgets/validation-station.md). When the document comes from a storage bucket the required scope is `OR.Buckets` (plus `OR.Tasks` if the widget completes an Action Center task on save) — add to the `scope` field in `uipath.json` during Step 4. If your backend already holds the document data, the in-memory path needs neither.
+If the user wants a **Document Understanding validation UI** (review/correct extraction results), embed the Validation Station widget — see [widgets/validation-station.md](widgets/validation-station.md). `OR.Buckets` is required when the document comes from a storage bucket; the in-memory path drops it. `OR.Tasks` is required whenever the app completes an Action Center task on save, on either data path. Add to the `scope` field in `uipath.json` during Step 4.
 
 When implementing specific SDK services, read the corresponding reference:
 
--- a/skills/uipath-coded-apps/references/create-web-app.md
+++ b/skills/uipath-coded-apps/references/create-web-app.md
@@ -225,7 +225,7 @@ const records = await entities.getAllRecords('<entity-id>'); // entity ID is a U
 
 Method signatures come from the installed types (`node_modules/@uipath/uipath-typescript/dist/<subpath>/index.d.ts`); per-method scopes from the shipped `node_modules/@uipath/uipath-typescript/docs/oauth-scopes.md` — see [oauth-scopes.md](oauth-scopes.md) for bundles and the lookup protocol.
 
-If the user wants a **Document Understanding validation UI** (review/correct extraction results), embed the Validation Station widget — see [widgets/validation-station.md](widgets/validation-station.md). Required scope: `OR.Buckets` (plus `OR.Tasks` if the widget completes an Action Center task on save). Add to the `scope` field in `uipath.json` during Step 4.
+If the user wants a **Document Understanding validation UI** (review/correct extraction results), embed the Validation Station widget — see [widgets/validation-station.md](widgets/validation-station.md). When the document comes from a storage bucket the required scope is `OR.Buckets` (plus `OR.Tasks` if the widget completes an Action Center task on save) — add to the `scope` field in `uipath.json` during Step 4. If your backend already holds the document data, the in-memory path needs neither.
 
 When implementing specific SDK services, read the corresponding reference:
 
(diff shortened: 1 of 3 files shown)


cata-on:
Addressing the two manual-review items, plus one correction to the review itself.

**The `create-web-app.md` line rated "factually correct" was not.** The reviewed revision said the in-memory path "needs neither" scope. `OR.Buckets` does drop away, but `OR.Tasks` does not: per `references/oauth-scopes.md`, it is required whenever the host calls `task.complete()` — action apps, and web apps that complete the task on save — which is independent of where the document data came from. Fixed in 5f13ae8, along with the same overclaim in Critical Rule 6, which now separates the two scopes explicitly and notes that a web app completing a task still needs its own `UiPath` instance to do so. The "in-memory needs neither scope nor SDK" summary in the review reflects the pre-fix revision.

**Domain accuracy of the standalone element API.** Verified against the shipped `types.d.ts` of `@uipath/du-validation-station-wc@1.0.0-rc.1` and the shipped `dist/*.d.ts` of `@uipath/ui-widgets-validation-station@1.0.1`, not against docs or recollection. On the three specifics raised:

- **`documentId`-first ordering** is the type's own wording: `documentId` carries the doc comment *"Required before any data inputs are set."*
- **Event detail types** — `IVsSaveValidatedDataRequest` is `{ documentId, validatedData, automaticExtractionResult, taxonomy }` and `IVsSaveExceptionReportRequest` is `{ documentId, exceptionReport }`, both exported, and both reachable through the per-event `addEventListener` overloads so `e.detail` needs no cast.
- **`<ui-du-validation-station-standalone-wc-persistent-element>`** is registered in the package's `HTMLElementTagNameMap` augmentation, and its interface adds `forceDestroy()` for permanent removal.

Both code examples were extracted verbatim from the committed file and compile under `tsc --strict` against those two packages with React 19. The harness was proved capable of failing: a control file using the JSX `ref` form is rejected with `TS2322`, which is why both examples use `createElement`.

**Critical Rule 5.** Deliberately untouched. It is accurate for the bucket-backed paths, Rule 6 and the new section both state that the in-memory path needs no `UiPath` instance for the widget, and a separate change already owns the rewrite of that rule together with the `sdk.initialize()` call in the action-app example. Editing the same rules block from two directions would only create a conflict, so I would rather it land there than here.

One defect found while writing this that is **not** fixed here, since it predates the change and belongs in its own diff: the subcomponents anti-pattern advises omitting the `enableSaveAsDraft` prop to hide the built-in draft button, but `CompactFieldsForm` destructures it as `enableSaveAsDraft = true`, so omitting it leaves the button visible. It needs `enableSaveAsDraft={false}`. The new Shape A example sets it explicitly for that reason.

cata-on:
The widget work has overtaken the approach here. Leaving this open rather than closing it — the rewrite should target the wrapper instead of the web component.

**What changed.** `ValidationStation` required `sdk` + `data`, which is why `<ui-du-validation-station-standalone-wc-element>` was the only in-memory route to the all-in-one layout. That is no longer the case:

- UiPath/uipath-ui-widgets#132 — adds `artifacts` + `documentId` with `sdk`/`data` optional, and unifies the outputs into `onSubmit` / `onSaveAsDraft` / `onReportException`, which fire in both modes. That retires this PR's caveat that the completion callbacks never fire without an SDK context. It also exports `fetchDuDocumentArtifacts`, `submitValidatedData` and `saveValidatedDataAsDraft`, so opting out of buckets no longer means reimplementing them.
- UiPath/uipath-ui-widgets#126 — the web component now loads from a URL via `@uipath/du-utils`, and `configureValidationStationWc()` replaces the asset-copy / CSS / `optimizeDeps` workarounds.
- UiPath/uipath-ui-widgets#143 (open) — removes the `folderId` prop and the same parameter from all four exported helpers, so the surface is not final yet.

**Not actionable yet.** Both merged PRs landed after `@uipath/ui-widgets-validation-station@1.0.1` was published on 2026-07-17, and the version was not bumped — so the in-memory path is not in any installable version. Rewriting now would document an API no consumer can install.

**What this PR still needs to carry, unchanged.** The Critical Rules correction is about scopes rather than about the widget, and survives the rewrite intact:

- `OR.Buckets` is required only on the bucket-backed data path, not unconditionally.
- `OR.Tasks` is required on either path whenever the app completes an Action Center task — including from a web app, which needs its own `UiPath` instance to do so.
- `create-web-app.md` carries the same unconditional claim and needs the same correction.

Plan: hold until UiPath/uipath-ui-widgets#143 lands and a version ships, then rewrite the integration section against `ValidationStation artifacts={…}` and keep the scopes correction as it stands.

PR #2451, opened 2026-08-04
fix(troubleshoot): encrypt the mock machinery and its runtime data files

dmorosanu:
## Why

The troubleshoot suite stages a mock `uip` CLI into the agent's working directory so scenarios serve recorded evidence instead of hitting a live tenant. Sealing the fixture store (earlier work) removed the readable `r/*.json` fixtures, but left the recipe for reopening it sitting in the same directory:

- `m/uip` and `m/seal` were ~475-byte **plaintext** loaders whose source read `zlib.decompress(base64.b64decode(...))`
- `m/.store` - the sealed archive - used **that same encoding**, with no key

So recovering the recorded evidence was copy-paste with no understanding required, and eval agents were observed doing exactly that. This closes that path and, separately, makes damaged mock data fail loudly instead of silently.

## How it works now

Two keys, because the requirement splits in two. `CODE_KEY` decrypts the code blobs and must be in the loaders (they run unaided in the sandbox). `DATA_KEY` covers the runtime data files, lives only in the non-staged sources, and therefore reaches the sandbox exclusively *inside* a `CODE_KEY`-encrypted blob.

**Build time** (`compile_mocks.py`, pinned to CPython 3.13 for byte-stable output):

```
mock_src/_cipher.py  (holds DATA_KEY)     mock_src/uip.py     mock_src/seal.py
        |                                        |                   |
        |        strip docstrings + ast.unparse  |                   |
        +----------------+----------------------++-------------------+
                         |  _cipher prelude prepended to each
                         v
            combined source per entry point
                         |
     keystream = chained SHA-256 seeded on sha256(CODE_KEY + name), XORed over the source
                         v
        [len:4][digest:8][ciphertext]  ->  m/.uip.bin   /   m/.seal.bin
        digest = sha256(CODE_KEY + plaintext)[:8]      <- keyed, so it cannot be forged
```

**Run time**, inside the sandbox:

```
LOADER (m/uip and m/seal, identical shape):
  1. reduce sys.path to interpreter-owned entries      <- so a planted module cannot be imported
  2. read the .bin, rebuild the keystream, XOR it
  3. verify length + keyed digest  ->  mismatch exits 1 "runtime data missing"
  4. compile + exec, with __file__ set to the blob path

pre_run, once:   python m/seal
    r/manifest.json + r/*.json  ->  one JSON envelope
    ->  zlib compress  ->  XOR under sha256(DATA_KEY + "store")  ->  m/.store
    ->  then r/ is deleted   (store written first, so the shim always finds one or the other)

every agent call:   uip <args>
    m/.store  ->  decrypt + decompress  ->  manifest + fixture bytes
    ->  first matching rule's fixture to stdout at its exit code
    ->  one record appended to m/.log, encrypted under sha256(DATA_KEY + "log")
```

SHA-256 is used as a **keystream generator**, not to hash the payload: the same key always regenerates the same byte stream, and XOR is its own inverse, so one function both seals and opens. It is also used, separately and genuinely one-way, for the integrity digest. Standard library only, because the sandbox guarantees no third-party packages, and the artifacts stay version-agnostic **source** (no bytecode - `python` resolves from the host PATH at an unknown minor).

`store`, `log` and `cache` each get their own keystream seed. The agent partly dictates `.log` contents by choosing which commands it runs, so a shared stream would let known plaintext there peel off keystream bytes and apply them to the fixture archive.

## What this does not do

**It is not confidentiality, and the docs in this PR say so.** `CODE_KEY` is a byte literal in the plaintext loader by necessity, so an agent that reads the loader can rebuild the keystream, decrypt the code blob, and recover `DATA_KEY` from it. Measured: about fifteen lines.

What it buys is that **no staged file names its own encoding**, turning a copy-paste crack into deliberate reverse-engineering. That is worth having because the observed leaks were opportunistic. Detection of answer-key reads is the control that actually holds, and it is a separate change.

Also unchanged: reference answers, grader scripts and task YAMLs still sit in the checkout the agent reads its skill from. This PR does not touch that surface.

## Damage now fails loudly

Worth calling out because it is independent of secrecy and arguably the larger win. A raw keystream XOR decrypts a truncated payload into garbage just as willingly as intact input, and `compile(b"")` succeeds while executing nothing - which would make every `uip` call return empty output at exit 0, and every scenario grade against evidence that was never served, invisibly.

The keyed length+digest header prevents that: zero-byte, truncated, byte-flipped and wrong-blob inputs all exit 1 with a single line and no traceback. The data files are compressed before encryption, so the zlib frame plays the same role there. The sealed store's shape is validated too, so a malformed store exits cleanly instead of printing the loader's stack to the agent.

## Verification

- 12 real scenarios under `experiments/default.yaml`: **12/12 SUCCESS**. Every task's `m/.log` non-empty (8-24 decoded records), so the mock genuinely served the evidence rather than the scenarios passing on nothing.
- Artifact sweep over 84 files in `artifacts/*/m/`: zero recipe tokens, zero occurrences of `DATA_KEY`, zero readable fixture text in any `.store`, no surviving `r/`.
- Acceptance grep, enforced by hand and worth adding to CI later: no file staged under `mock_template/` contains `base64`, `b64decode`, `b64encode`, `zlib`, `decompress` or `compress`.
- Repacking unchanged sources twice is byte-identical; the packer refuses a non-3.13 interpreter without `--allow-any-version`, validates the combined source compiles, round-trips each blob before writing, and writes nothing at all if any entry point fails.
- Seal idempotence and crash-resume preserved: no manifest is a no-op exit 0, and the store write plus `rmtree` still happen last so a `pre_run` retry resumes.
- `scripts/coverage_report.py --dump` still decodes call logs, including older sealed stores.
- Mock cycle exercised on CPython 3.10 through 3.14; the 3.9 guard still fires.

### Linux and Docker

Validated on Ubuntu 24.04 at this branch tip.

**Mock cycle on Linux: 98 checks, 98 pass.** Seal, serve, encrypted log, decode - on CPython 3.10, 3.11, 3.12 and 3.13 outside a virtualenv, on 3.12 and 3.13 **inside** one, and on the container's own 3.13. The decisive case for the `sys.path` reduction is a 3.13 venv where `base_prefix` and `base_exec_prefix` are different paths and the standard library is split across them: both survive the filter, so `lib-dynload` is kept and the stdlib resolves. `PYTHONPATH`-planted `hashlib.py` and `json.py`, and a planted `m/_cipher.py`, are all correctly ignored. This was the branch's largest untested risk and it does not materialise. `_cache` encryption is now exercised inside the container as well.

**Sealer, with a genuine failure rather than a simulated one.** A real mid-`rmtree` `PermissionError` was provoked naturally, with no patch to the source: the sealer exits exactly 1 (so `fail_on_error` aborts the task rather than running an agent against readable fixtures), `manifest.json` survives so the retry can resume, and the retry produces a **byte-identical** store. All four damaged-blob modes still fail loudly.

**12 tasks under `driver: docker`** (`nightly.yaml`, `--include-skipped`): **12/12 SUCCESS** - ten at 1.000 plus 0.925 and 0.850. Every task's call log is non-empty and fully decodable (zero undecodable records) with at least two fixtures served, no `r/` survived, and opacity is clean across 84 artifact files. All twelve staged loaders were verified as this tip's, so the run genuinely exercised this branch.

**One agent read the encrypted call log.** It issued a `Read` on `m/.log`, received pure hex ciphertext, never referred to it again, and scored 0.925 - the feature behaving as intended under a real adversarial read rather than a synthetic one.

**The 100644 file modes are correct**: `Sandbox._prepare_mock_path_dirs` ORs in `0o111` before the agent starts, and the artifacts show `-rwxrwxr-x`. No exec bit belongs in git.

Two qualifications. The image is a near-equivalent of CI's rather than CI's own (its registry was unreachable from that host), which qualifies the scores but not the cipher findings. And no agent hit a `docsai` passthrough in these twelve, so `_cache` still has no coverage from a real scenario.

## One behaviour change to know about

Nine scenarios (`web-http-*`, `web-deserialize-*`) declare `unmocked_default.response` as a JSON object. On `main` that crashes the shim on any unmatched command - `TypeError`, exit 1, and a traceback exposing the loader. This PR serializes the object instead, so those scenarios now serve their declared body at their declared exit code.

That removes 8-11 spurious tracebacks per run, but it does introduce a permissive empty-result envelope where a crash used to be. All nine pass at 1.00 on both the tempdir and docker drivers, and no criterion grades the unmatched path mechanically.

**The serializer is exercised, but no agent has yet reached the envelope.** Under Docker the new dict path serialized correctly 55 times across the nine, so the code works - but all 55 were `uip track` / `uip --version` telemetry from the plugin's own hook, and no agent in those nine issued an unmatched *evidence* query. So the behavioural risk stays unsampled on both drivers.

The other sealed scenarios cannot stand in for it: their `unmocked_default.response` is the plain string `"[]\n"`, not the envelope. Those agents did handle an empty `[]` without drawing false conclusions. Worth re-checking the nine if their prompts ever change, since a permissive empty-success is a known misread shape.

## Follow-ups, not in this PR

- CI guard for blob/source sync and the acceptance grep. Nothing verifies either today, so a source edit without a repack silently ships stale behaviour.
- The packer's `compile()` check runs under 3.13 only, so 3.11+-only syntax would pack clean and fail on the documented 3.10 floor.
- A planted `sitecustomize.py` reached through `PYTHONPATH` executes at interpreter startup, before the loader's first line, so no loader-side change can close it. It grants nothing beyond the ceiling above, which is why it is left open.


dmorosanu:
closed since it does not solve the agents leak problem

PR #2465, opened 2026-08-05
feat(skills): add generic build-time skill flavors and npm packages

astridstaicu:
## Summary

- keep every canonical `SKILL.md` complete for the default/local experience
- model host differences as sparse replacement blocks under `skill-flavors/<flavor>/`
- discover flavors generically, with no allowlist, JSON registry, or per-flavor build script
- build complete, marker-free npm packages for the default catalog and every discovered flavor
- preserve the established root `npm pack` / `npm publish` behavior as the default-only package path
- provide a generic GitHub Packages flavor publisher while requiring every published flavor to opt in explicitly
- add a repository-local contributor skill covering flavor authoring, packaging, publishing, recovery, and regression checks

## Why

Canonical skills contain local-workspace instructions such as CLI project initialization and `.uipx` solution metadata handling. Those instructions conflict with Studio Web, where project creation must use the host's project-creation tools.

Build-time composition keeps canonical skills readable, makes host differences sparse exceptions, and gives every consumer a finished skill tree with no runtime Markdown composition.

## Source and package contract

| Source | Generated npm package |
|---|---|
| canonical `skills/` | `@uipath/skills` |
| `skill-flavors/studioweb/` | `@uipath/skills-studioweb` |
| `skill-flavors/<flavor>/` | `@uipath/skills-<flavor>` |

The canonical skill remains the source of truth. A flavor contains only mirrored Markdown files with complete replacement blocks. It does not select skills: every package includes every canonical skill, and only matching marked passages change.

`npm run skills:pack` creates:

- complete marker-free trees in `build/skills/`
- staged packages in `build/packages/`
- real, verified tarballs in `build/npm/`

The first flavor is Studio Web. Both packages contain all **25 skills / 1,701 skill files**; Studio Web applies **12 replacements**, while all other guidance passes through unchanged. Adding a valid `skill-flavors/<new-flavor>/` directory automatically creates `@uipath/skills-<new-flavor>` during the generic build.

## Backward-compatible default package

Repository-root commands retain their old meaning:

- `npm pack` builds exactly one marker-free `@uipath/skills` tarball
- `npm publish` publishes only `@uipath/skills` and preserves caller registry, tag, access, and provenance flags
- `npm run skills:pack` is the explicit command that builds the default package and every discovered flavor

The `prepack` / `postpack` transaction activates the composed default tree only for root packaging and then restores the canonical `skills/` tree. `npm run skills:recover` safely restores interrupted transactions.

## Explicit publishing policy, generic flavor mechanism

Package generation is generic; package publication is explicit.

| Package | Event | Destination | Tag |
|---|---|---|---|
| `@uipath/skills` | push to `main` | GitHub Packages | `dev` |
| `@uipath/skills` | push to `release/v*` | npmjs | `preview` |
| `@uipath/skills` | manual | existing destination | `dev`, `preview`, or `latest` |
| `@uipath/skills-studioweb` | push to `main` or manual `dev` | GitHub Packages | `dev` |
| `@uipath/skills-studioweb` | push to `release/v*` or manual `preview` | GitHub Packages | `preview` |

`publish.yml` keeps the established root/default `npm publish` commands and their manual behavior. Its explicit Studio Web jobs call `publish-skill-flavor.yml` with `flavor: studioweb` and `channel: dev` or `preview`. The reusable workflow runs in the same workflow run, so default and Studio Web derive the same prerelease version from `github.run_number`.

The reusable flavor publisher:

- accepts exactly `flavor` and `channel`; it does not accept package name, registry, version, tarball, credentials, provenance, or arbitrary npm flags
- validates lowercase kebab-case flavor names, rejects `default`, and requires a real non-symlink `skill-flavors/<flavor>` directory
- derives `@uipath/skills-<flavor>` through the same `packageName()` contract as the composer
- is fixed to GitHub Packages, `@uipath`, `GITHUB_TOKEN`, and the `dev`/`preview` channel allowlist
- selects exactly one tarball matching the derived name, flavor, and stamped version
- scans that exact tarball for markers, verifies the registry at runtime, and publishes only the selected path
- contains no hardcoded Studio Web package identity

Manual `dev`/`preview` dispatches invoke the explicit Studio Web callers because sprint-cut automation uses dispatches when bot-created release-branch pushes cannot trigger another workflow. Manual `latest` remains default-only.

Future flavors are built automatically but are **not** published automatically. A contributor must add a reviewed explicit caller for the intended flavor and channel. There is no discovered-flavor publication matrix.

## Release and package safety

- root `npm pack` and the generated default tarball are asserted byte-identical by SHA-512
- exact default and selected flavor tarballs are scanned for leaked flavor markers
- custom package manifests strip scripts and `publishConfig`; the selector rejects a custom tarball that supplies `publishConfig`
- workflow tests prove default jobs do not build/loop over flavor artifacts and the generic flavor publisher cannot target npmjs
- a runtime `future-host` fixture proves a second custom flavor is derived, packed, and selected without changing the reusable workflow
- failed validation or packing preserves the last successful generated output atomically

## Review feedback addressed

- separated flavor publication from the default package path, per Gabi's review
- restored `publish.yml` default jobs to root `npm publish` behavior
- removed wildcard/multi-package publishing from the default jobs
- kept Studio Web GitHub Packages-only, with exact tarball selection and the caller run number
- generalized the reusable publisher to accept an explicit flavor without enabling automatic flavor publication
- removed the unrelated `skills.sh.json` grouping change
- removed unused OpenAI skill metadata
- retained the concrete `skills/uipath-<name>/` contributor path

## Checks

- Node composer, lifecycle, concurrency, recovery, package-selection, workflow-contract, and contributor-skill suite: **45 passed**
- `npm run skills:validate`: default **25 skills / 1,701 files**; Studio Web **25 skills / 1,701 files / 12 replacements**
- `npm run skills:pack`: verified `@uipath/skills@1.200.0` and `@uipath/skills-studioweb@1.200.0`
- real root `npm pack` and `npm publish --dry-run`: default-only and canonical sources restored
- root/default and generated/default tarballs: byte-identical SHA-512
- selected Studio Web tarball dry-run: GitHub Packages destination only
- generic `future-host` package derivation and exact selector runtime proof: passed
- workflow YAML, version synchronization, contributor-skill validation, marker scans, and `git diff --check`: passed
- two independent workflow/security reviews and a fresh-agent contributor-skill forward test: no blocker


skills.sh.json
```diff
@@ -26,7 +26,8 @@
         "uipath-planner",
         "uipath-solution",
         "uipath-review",
-        "uipath-automation-discovery"
+        "uipath-automation-discovery",
+        "uipath-process-mining"
```
astridstaicu: why do we need this addition?

CONTRIBUTING.md
```diff
@@ -75,7 +80,10 @@ Tool wiring lives outside `skills/`:
 | Cursor IDE | `.cursor/rules/*.mdc` | Scoped MDC rules: `token-optimization` (always-apply), `skill-structure` + `content-quality` (glob-scoped), `skill-review` + `pr-review` (agent-requested) |
 | GitHub Copilot coding agent | `AGENTS.md` (symlink → `CLAUDE.md`) | Copilot reads `AGENTS.md` natively (since Aug 2025) |
 
-When adding a skill, only touch files under `skills/uipath-<name>/` — the root integration files already wire every tool up automatically.
+When adding a skill, the root integration files and every custom flavor include
```
astridstaicu: lets also include `skills/uipath-<name>/` path instead of "root integration file". paths help agents and users, lets not remove occurrences of paths

CODEOWNERS
```diff
@@ -93,6 +93,7 @@
 /tests/tasks/uipath-maestro-bpmn/ @bai-uipath @rockymadden @gozhang2 @tmatup @akshaylive @jiyangzh @nikhil-maryala @baishalighosh @DevMomo
 # API Workflow skill
 /skills/uipath-api-workflow/ @rares-baesu-uipath @liviubobocu @mirastroie
+/skill-flavors/*/uipath-api-workflow/ @rares-baesu-uipath @liviubobocu @mirastroie
```
astridstaicu: lets not add this line yet

skills.sh.json
astridstaicu: Removed it. The process-mining grouping change was unrelated to skill flavors, and skills.sh.json now matches main.

CONTRIBUTING.md
astridstaicu: Updated. The guide now explicitly keeps the canonical path skills/uipath-<name>/ and explains that every custom flavor includes it automatically.

CODEOWNERS
astridstaicu: Removed as requested. CODEOWNERS now matches main; we can add flavor ownership later with the broader ownership decision.

gabrielavaduva:
After rechecking the latest head, I would like to simplify the publishing architecture before merge.

### Keep the default package on the established path

`publish.yml` should remain default-only and preserve the previous root publishing behavior for `@uipath/skills`:

- merge/push to `main` -> GitHub Packages, stamped `dev`, using root `npm publish --tag dev`
- merge/push to `release/v*` -> npmjs, stamped `preview`, using root `npm publish --access public --provenance --tag preview`
- manual `dev` / `preview` / `latest` -> retain the existing destinations and behavior

The new `prepack` / `postpack` transaction can remain responsible for making the root/default package marker-free and restoring the canonical `skills/` tree. This keeps the classic package path and release semantics intact.

### Publish Studio Web separately, to GitHub Packages only

Please add a separate Studio Web job/action for `@uipath/skills-studioweb`:

- merge/push to `main` -> GitHub Packages with tag `dev`
- merge/push to `release/v*` -> GitHub Packages with tag `preview`
- never publish this package to npmjs
- no npmjs registry configuration, OIDC `id-token` permission, or `--provenance` in the Studio Web path

The Studio Web action may run `npm run skills:pack`, but it must not publish `build/npm/*.tgz`. That directory also contains the default package. It should inspect the manifests and require exactly one tarball with both:

- `name: @uipath/skills-studioweb`
- `uipathSkillsFlavor: studioweb`

Then scan that exact final tarball for flavor markers and publish only that tarball to `https://npm.pkg.github.com` using `GITHUB_TOKEN` / `packages: write`.

I would prefer the Studio Web path to be a separate job or reusable action within the same workflow run so it shares the resolved/stamped version with the default package. Fully independent workflow files have independent `github.run_number` sequences unless the version is explicitly passed between them.

Please also keep/add these regression guards:

1. root `npm pack` and the generated default tarball are byte-identical (SHA-512);
2. the exact final default and Studio Web `.tgz` artifacts contain no flavor markers;
3. a static/runtime guard proves the Studio Web publisher has no npmjs destination and publishes only the selected Studio Web tarball.

This separation removes the multi-package publish loop from the default path, preserves `main` and release-branch behavior for `@uipath/skills`, and ensures the Studio Web flavor can never be sent to npmjs.

RazvanRotaru (approved):

gabrielavaduva:
One remaining issue before merge:

### Serialize equivalent publish channels across trigger types

`publish.yml` currently uses `concurrency.group: publish-${{ github.event_name }}`. A push and a manual dispatch targeting the same effective channel therefore use different groups and can run concurrently:

- push to `main` vs. manual `channel=dev`
- push to `release/v*` vs. manual `channel=preview`

Each run updates both the default and Studio Web dist-tags through independent jobs. If two runs interleave, the final default and Studio Web tags can point at different stamped run versions (and an older run can move one of the tags after the newer run).

Please use one static publish concurrency group, or normalize the group by effective channel so push/manual equivalents share it. It would also be useful to extend the workflow contract test to assert that `main` push + manual `dev`, and release push + manual `preview`, cannot occupy separate groups.

gabrielavaduva (approved):

skill-flavors/studioweb/uipath-api-workflow/SKILL.md
```diff
@@ -0,0 +1,61 @@
+    <!-- skill-flavor:connector-solution-registration:start -->
```
AlvinStanescu: I expected no spacing on `<!-- skill-flavor:connector-solution-registration:start -->`. Update the skill too, not sure we process this correctly.

skill-flavors/studioweb/uipath-api-workflow/SKILL.md
```diff
@@ -0,0 +1,61 @@
...
+- **Do NOT** search for, create, or edit `.uipx` solution metadata in Studio Web; the host owns that metadata.
+- **Do NOT** hand-assemble a replacement project when `CreateProjects` is unavailable. Report the missing capability so the user can choose how to proceed.
+- **Do NOT** use successful local validation, packaging, or publication as evidence that a project was created correctly in Studio Web. The `CreateProjects` result and host-exposed project tree are authoritative.
+<!-- skill-flavor:project-creation-antipatterns:end -->
+
+<!-- skill-flavor:authentication-remediation:start -->
+- `Not authenticated` / `Organization ID not available` → do not ask the user to run `uip login`. Use an exposed Studio Web authentication or connection capability after inspecting its live schema; if none exists, report the host-level authentication blocker and do not retry.
+<!-- skill-flavor:authentication-remediation:end -->
+
+<!-- skill-flavor:reference-navigation:start -->
```
AlvinStanescu: The CLI reference is gone from here but not from the main skill. I don't think this is intended, or?

Models may easily make mistakes such as this if the full reference navigation is in the skill flavor. Either update the skill as well as the skill authoring skill (which is not updated but should have been too) to be very careful about this and always track changes across skill flavors as otherwise you may have somebody update a skill but the model/person forgetting to also update its flavors and missing out on new functionality altogether.

I think this should be additive, so rather `reference-navigation-extra` which is empty in the default skill.

AlvinStanescu (changes requested):

skill-flavors/studioweb/uipath-api-workflow/SKILL.md
astridstaicu: Yes, I removed all whitespaces from the marker

skill-flavors/studioweb/uipath-api-workflow/SKILL.md
astridstaicu: At first, the API workflow skills is an experiment, I want to merge this PR mostly because I want the architecture and the system to be available. I didn't iterate too much on the quality of the studio web api workflows skill yet.

I anyway implemented the feedback, but the skill will have more changes after the whole system is in place and we'll do PRs only for adding the studio web flavor for all required skills.

AlvinStanescu (approved):

PR #2469, opened 2026-08-05
fix(maestro-case): document the caseplan schema version contract

cata-on:
## Problem

A hand-authored `caseplan.json` that followed this skill's recipes was refused by the Studio Web designer on a staging tenant:

```
JSON is not a valid Case Management JSON of any previous version.
```

`version` is a **schema selector**, not a label. Studio Web pins one exact literal per schema (`z.literal("23.0.0")`, `"27.0.0"`, `"30.0.0"`, …) and resolves a document by exact match. So one violation of the named schema makes every lower chain entry fail on *its* version literal, and the designer can only report the generic message above — it never names the offending field.

The skill emitted `23.0.0` with no note that it must not be hand-edited, and `case-editing-operations.md` contained **zero** mentions of `version` and no mention of the download-edit path. An agent that sees a tenant export declaring `30.0.0` will reasonably align the literal — and thereby opt into `30.0.0`'s field names while every other recipe in the skill still writes the `23.0.0` names. That is precisely the reported failure:

| Reporter's file | `version` | fields |
|---|---|---|
| broken (designer refuses to open) | `30.0.0` | `selectedStageId` x2, `timerType` x3 |
| fixed (opens) | `30.0.0` | `selectedStageIds` x2, `timer` x3 |

## What this PR does not do

**It does not rename the documented fields.** The `23.0.0` shapes are self-consistent and do validate — the repo's own `LinearThreeStages` fixture (`version: 23.0.0`, `selectedStageId`, `timerType`) passes `uip maestro case validate` on `@uipath/maestro-tool` 1.198.1, and Studio Web opens a `23.0.0` document by migrating it forward.

Flipping the skill to the `30.0.0` names would be actively worse: the validator's case-management chain tops out at `27.0.0`, so every authored case would then fail the skill's own `validate` gate. Verified — taking the working `30.0.0` document, relabelling it `27.0.0`, and validating yields `Entry rule "After Intake" stage selection missing` and `A required field is empty in 'Hold For Documents'`, i.e. the validator wants the singular/`timerType` names. There is currently no single document that satisfies both the validator and a v30 tenant.

So this PR documents the contract and keeps `23.0.0` as the emitted default.

## Changes

6 files, +44/-2, docs only.

1. **`references/case-schema.md`** — new `### Schema version contract`: selector semantics; emit `23.0.0` and never hand-edit; the renamed-field table scoped to `30.0.0`; the validator ceiling. Plus **Never hand-edit** on the `version` field row.
2. **`references/case-editing-operations.md`** — new `## Editing a tenant-exported caseplan`: read `version` first and branch; for `28.0.0`+ match the shapes already in the file, never rewrite `version`, and treat `[error] [version] Invalid input: expected "27.0.0"` as a validator ceiling rather than a defect to repair. Without this the failure mode is an agent "repairing" a perfectly good tenant export to satisfy a validator that structurally cannot accept it.
3. **One-line version-conditional pointers** in `wait-for-timer/impl-json.md`, `stage-entry-conditions/impl-json.md`, and `case-exit-conditions/impl-json.md`.
4. **`plugins/case/impl-json.md`** — corrects a factual error: the doc claimed `caseplan.json.bpmn` is "generated by downstream tooling (validate / pack)". A successful `uip maestro case validate` emits no `.bpmn` (verified in both a bare-file and a full project layout), so a stale `.bpmn` outlives a CLI-side edit. `pack` was not tested and is left unclaimed.

### The carve-out that makes this not a rename

`timerType` is **not** globally wrong at `30.0.0`. It survives in two places, and a blanket rename would break both:

- the timer **trigger** — `data.uipath.timerType` (`references/plugins/triggers/timer/impl-json.md`, 3 occurrences)
- the `timer` **condition rule** — `timer.timerType`

Only `wait-for-timer` **task data** renames to `data.timer`. In the shipped FE bundle the v30 task-data object is bound with `timer:` as the enum key and is non-strict with all-optional keys, so a `timerType` key there raises no error and is silently dropped, leaving the timer unconfigured. By contrast `selectedStageId` at v30 is bound to a *rejecting* type carrying `"selectedStageId is deprecated, it should be a stage in selectedStageIds instead"` — a hard parse error. Exactly two such rejecting sites exist in the bundle, both this field.

`case-exit-conditions` is included because `caseExitRules` binds the same rule union as stage-entry conditions (verified in the bundle), so it is affected identically — it was not in the original report.

## Question for code owners

@charlesliu9 @song-zhao-25 @abhiram-vad @jundayin — two things I'd like your call on:

1. **Does this need a cherry-pick to `release/v1.198`?** The defect was observed there and is present on `main`. I authored against `main` and did not cherry-pick, since both `case-schema.md` and `case-editing-operations.md` differ between the branches and the backport needs a manual resolve. Happy to open it if you want it in 1.198.
2. **`origin/feat/case-schema-v27`** (last commit 2026-07-17, 277 commits behind `main`) bumps the emitted literal `23.0.0` -> `27.0.0` in six places but does not touch `selectedStageId` or `timerType`. Landing it as-is would not fix this and would move the emitted literal to the validator's exact ceiling with zero migration headroom. Is that branch still intended to land, and should it be rebased on top of this?

The longer-term fix is not in this repo: the validator ceiling belongs to `@uipath/maestro-tool` (`repository: github.com/UiPath/cli`, `directory: packages/maestro-tool`) and is filed separately. Once the ceiling reaches the version Studio Web serves, the guidance added here should be revisited — and the emitted default reconsidered — rather than left as a permanent split.

## Verification

- `bash hooks/validate-skill-descriptions.sh` — clean
- `python3 scripts/check-skill-status.py` — `OK — 24 skills, manifest valid.`
- All four new relative links resolve; both new anchors exist
- `uip maestro case validate` reproductions run against `@uipath/maestro-tool` 1.198.1, tenant `https://staging.uipath.com` / `uipathlabs` / `FDE`

Generated with [Claude Code](https://claude.com/claude-code)


cata-on:
Tracking issue: #2470. CLI-side validator ceiling: **UV-15732** — https://uipath.atlassian.net/browse/UV-15732

PR #2476, opened 2026-08-05
docs(uipath-troubleshoot): add coded app route-collision recovery

kab0rn:
## Summary
- add a causal playbook for HTTP 400 routing name must be unique
- distinguish fresh-create collisions from existing-app upgrade defects
- require authoritative remote reconciliation after indeterminate writes
- prohibit blind retries, random routes, delete/recreate, and local state as remote authority
- align three older coded-app remediation playbooks with the same guarded boundary

## Validation
- git diff --check
- changed-file Markdown link and fence checks
- focused no-blind-redeploy assertions
- skill description and lifecycle-status validation

The repository-wide skills list still has a pre-existing unrelated uipath-process-mining registration finding on current upstream main.


skills/uipath-troubleshoot/references/products/coded-apps/playbooks/routing-name-must-be-unique.md
```diff
@@ -0,0 +1,56 @@
...
+   - **Insufficient evidence:** no route owner or exact deployment/version state can be established. Stop; do not infer safety from `.uipath/app.config.json` or from the CLI's exit code alone.
+
+## Resolution
+
+- **If the requested route is already owned by a different coded-app deployment:** do not retry the same create and do not generate a random route. Confirm which app is supposed to own the human-facing route. The owner must select an intentionally approved unused route or resolve the existing ownership through the supported Apps administration path before a new create is attempted.
+
+- **If an existing-app upgrade resubmitted its unchanged `routingName`, and server uniqueness validation treated it as a conflicting new assignment:** use a supported in-place upgrade path bound to the exact existing deployment that does not reassign or change its route. If the available deployment client cannot perform that operation, stop and report the client/server defect; do not simulate an upgrade by creating a second app, omitting the desired route on a fresh deployment, issuing raw REST requests, or deleting and recreating the app.
```
amrit-agarwal-1, line 52: The cause here is a real flag error, not a server defect: --path-name is fresh-deploy only, and re-passing it on an upgrade throws exactly this 400. This branch never mentions --path-name and instead lands on "stop and report the client/server defect", so we'd escalate a one-flag mistake. It also lists "omitting the desired route" as something not to do, which is close to the actual fix. Maybe suggest making the first upgrade check: did you pass --path-name? Omit it.

skills/uipath-troubleshoot/references/products/coded-apps/playbooks/routing-name-must-be-unique.md
```diff
@@ -0,0 +1,56 @@
...
+
+3. Inspect the local deploy hint, if present:
+
+   ```bash
+   cat .uipath/app.config.json
+   ```
+
+   Use `appUrl` and `systemName` only to locate the expected deployment. This file is not authoritative and cannot establish current remote state.
+
+4. In the target tenant's Apps deployment inventory, re-read the deployment that owns the requested route and the intended app's current state. Capture the exact deployment identifier, system name, route, active version, and candidate package/deploy version. This remote read is mandatory after a timeout, connection loss, interruption, HTTP 5xx, or any other result where a write may have reached the service.
```
amrit-agarwal-1, line 40: This step is mandatory but has no command, and there's no uip verb for it which read Apps state. So every branch that depends on this read is unreachable. Suggest either naming the mechanism concretely, the Apps REST read, or the portal at cloud.uipath.com/{org}/apps_/ or dropping the step and discriminating from -n / --folder-key / .uipath/app.config.json, which is what deploy itself resolves against.

skills/uipath-troubleshoot/references/products/coded-apps/playbooks/deploy-404-app-not-found.md
```diff
@@ -47,13 +47,14 @@ What to look for:
...
+- Before publishing or deploying that candidate, reconcile the authoritative deployment, system name, route, current version, target folder, OAuth client, and package versions. Execute a fresh guarded `create` or `upgrade` operation with the exact candidate; do not resume or repeat the failed write.
 
-- **If `.uipath/app.config.json` has no valid `appUrl`:** the deploy did not finish. Re-run `uip codedapp deploy --output json` and confirm the output reports an `App URL`; resolve any error it prints before re-testing the URL.
+- **If the router basename is hardcoded:** replace it with `getAppBase()` from `@uipath/uipath-typescript`, then build/package a new candidate and follow the same reconciliation boundary.
+
+- **If `.uipath/app.config.json` has no valid `appUrl`:** do not conclude that deploy failed or rerun it. Re-read remote Apps inventory. If the exact candidate is active, use and verify the authoritative hosted URL. If it is published but not deployed, prepare a fresh guarded operation bound to the reconciled deployment and versions. If remote state cannot be proven, stop and escalate.
```
amrit-agarwal-1, line 60: This contradicts line 17 and step 3 of the same file, which still treat a missing appUrl as proof deploy didn't finish, and it removes the only actionable fix.

PR #2483, opened 2026-08-05
feat(case): author conditions and SLA in Phase 2, gate on --skeleton-v2

song-zhao-25:
## Why

Phase 2 emitted only structure. With edges retired (Rule 20), stage transitions are entry/exit conditions — so a Phase 2 preview published for review showed **disconnected boxes**. Conditions *are* the flow.

Nothing forced conditions or SLA into Phase 3: they reference only StageIds, TaskIds, variable IDs, and the SLA/escalation IDs preallocated at Step 9.9 — all minted in Phase 2. Three independent reviews walked every rule-type in all four scope matrices and found no forward dependency.

## What changed

**Steps 9.9 (SLA/esc ID prealloc), 10 (conditions, all 4 scopes) and 11 (SLA + escalation) move into Phase 2.** The boundary gate is renumbered 9.5 → 11.9 so Phase 2 reads monotonically.

The phase split is restated as the **connector-gather line** — Phase 2 is what the agent can write from data already in hand; Phase 3 is the `case spec` round-trips and the IDs they mint. That line is not absolute and the docs now say so: a resolved event trigger runs its own `case spec` inside Phase 2 at Step 6.1, because Step 6.2's In-arg `elementId` needs the TriggerId.

### Connector rules — stub, then upgrade

The one connector coupling in conditions is `wait-for-connector`, whose `uipath` needs `case spec`. Phase 2 writes it with the **stub `uipath` that already existed** for unresolved connectors; new **Phase 3 Step 10.5** upgrades it in place, touching only `rule.uipath` and never the enclosing conditions array.

- Step 10.5 opens with its own unconditional four-scope scan for leftover `"placeholder"` context entries and owns the skip decision. A skipped-but-needed 10.5 ships a case that passes `validate` and faults only at debug/run.
- Only that rule-type mints outputs, and its upgrade stays after Step 9.7 — so the **Step 12 Check 8 dedup ordering is unchanged**.
- The stub now has two documented lifetimes: transient (every rule, build sequencing — not logged) and permanent (connector unresolved — logged, reported, faults at run).
- Brownfield has no Step 10.5, so it writes the real `uipath` in one pass. Carve-outs added to the shared contract, `brownfield.md`, and the SLA plugin.

### Gate profile

`--skeleton-v2` when the CLI supports it, otherwise **full mode**. Never legacy `--skeleton` — it omits entry/exit rules, SLAs and escalations, exactly what Phase 2 now writes, so `0 errors` from it is not evidence the rules are clean.

Verified against uip 1.199: full mode is clean on a complete Phase-2 caseplan (connector task with only `typeId`/`connectionId`, `value: ""` inputs, root `slaRules` + escalation, `sla-status-change` entry rule, stub rules in two scopes → `Status: Valid`) and catches five defects `--skeleton` reports as `Valid`. The probe discriminates on `ErrorCode` (`invalid_argument` = flag absent), not `Code`, which a genuine validation failure does not carry.

### Section-boundary validate

Full-mode validate now runs at every `tasks.md` section boundary. The §4.7 boundary is therefore **expectedly dirty** whenever an `sla-status-change` rule exists, because Step 11 has not emitted `slaRules[]` yet. Recorded as an expected exception in Pre-flight Item 12 and § Validation Cadence — never repaired, never counted toward the 3-retry cap. Repairing it by dropping `escalationId` silently converts at-risk to breach.

## Review history

Four adversarial review rounds, each by a fresh reviewer given the design claims to **refute** rather than confirm. Rounds 1–3 found and this commit fixes: a gate fallback that defeated the change's purpose; four condition-plugin bodies contradicting the new contract; a factually false "no `case spec` in Phase 2" invariant; a vacuous Step 10.5 skip predicate (found twice); ~900 words of duplication; the §4.7 boundary collision above; and an SLA post-write check fixed for greenfield in a way that broke brownfield. Round 4 is in flight — findings will be posted as a comment.

## Verification

- Link/anchor check across all 67 skill markdown files: **0 new breaks**; one pre-existing broken anchor repaired (`phased-execution.md#prompt`).
- `hooks/validate-skill-descriptions.sh`, `scripts/check-skill-status.py`, `evals.json` parse: pass.
- `scripts/check-skill-verbs.py`: 6 stale findings, **identical on `main`** — no regression.
- Frontmatter untouched, so no `activation-gate.yml` recall-eval trigger.

## Open before this can merge

- [ ] **Run the `connector_wait` coder-eval task against a live tenant.** Its checker gained the stub-upgrade, `elementId`, and `bindings_v2` assertions; per `.claude/rules/test-writing.md` §5 a missing passing-run claim is a High lint finding. Needs tenant access.
- [ ] **`--skeleton-v2` does not exist in any published CLI** (string absent from the 1.198.0 and 1.199.0-preview.108 tarballs). Full mode is what the docs present as the working command, and every mention of the flag says it has not shipped. Nothing here breaks without it — but the caveats should be retired once it lands.
- [ ] Round 4 review findings triaged.

## Checklist (`.claude/rules/pr-review.md` — existing skill modification)

- [x] SKILL.md frontmatter still valid
- [x] No Critical Rules removed (Rule 11 amended: gate fallback corrected)
- [x] No new structural cross-skill dependencies
- [x] Reference file naming conventions preserved
- [x] Changes scoped to `uipath-maestro-case` (+ its own test task)
- [x] No secrets, tokens, or personal paths


song-zhao-25:
## Review round 4 — findings and fixes

Fourth adversarial review round (fresh reviewer, given the design claims to refute). **Five High, no Critical.** All five are now fixed in `d4c932bb7`. The reviewer independently reproduced every empirical claim against `uip 1.199.0-preview.99` and confirmed phase-ordering is sound across all four scope matrices, with 0 new broken links across 959.

Notably, four rounds running the dominant defect has been the same one: **a fix applied to one file, scope, or journey and not its twin.** Round 4 caught it in both directions, so the fixes below deliberately *extract* rather than paste.

### Fixed

| # | Finding | Fix |
|---|---|---|
| H1 | The `sla-status-change` post-write paragraph was pasted byte-identically into `task-entry-conditions`, carrying *"`isInterrupting` must match the declared response"* — a field that **does not exist** on a task entry condition (`case-schema.md` scopes it to stage-level `EntryCondition`). The check was unsatisfiable; the agent would fail it on every `start-task` SLA response or "repair" it by writing a schema-undefined key. | Extracted the shared two-thirds to `sla-response-shapes.md § 6`; both scopes now link it and keep only their differing clause. Task-entry's clause states explicitly that the field does not exist there. |
| H2 | The §4.7 expected-error carve-out was not applied to **§4.6**, which is guaranteed dirty (7 errors: `Case has no completion rules`, `Task has no entry rules` ×5, `Case has no stage with a Case Entered entry rule`) — and every available repair is explicitly forbidden elsewhere in the same doc, so the agent would corrupt the artifact or halt mid-Phase-2. | Replaced the section-specific carve-out with a **general rule covering every Phase 2 boundary before §4.7/§4.8**, naming the four expected messages and stating that satisfying them early is forbidden. Also covers §4.3/§4.4 without needing another patch. |
| H3 | `_assert_rule_connection_synced` could not detect the defect it guarded: `"Connection" in resources.__str__()` passes on the *task's* Step-9.7 projection, and the fixture gives rule and task the same connection id, which `bindings_v2` collapses into one entry. My commit message claimed coverage that did not exist. | **Removed the assertion** and documented in the checker docstring what is not covered and what a real fixture would need. A vacuous assertion is worse than none. |
| H4 | The `elementId` assertion was conditional and suffix-only — it passed when `elementId` was absent (the thing Step 10.5 must mint) and when the prefix was the **task** id instead of the **stage** id, the exact pitfall three docs warn about. | Now requires presence and asserts equality against `f"{stageId}-{ruleId}"`, resolving the owning stage from the node graph. |
| H5 | Test not run against a live tenant. | Still open — see the PR checklist. Needs tenant access. |

Mediums also fixed: a carve-out in `implementation.md` saying non-connector outputs "may be referenced directly", contradicting Rule 10 and five plugin blockquotes; a stale *"Phase 3 product"* label under a heading this change had moved to Phase 2; a missing brownfield write-order rule (clock before rule, since brownfield has no §4.7 carve-out); the probe decision table gaining a default branch; `Skeleton published` → `Case shape published` and the remaining "skeleton-profile" gate labels; plus three Lows.

### Deliberately not done

- **Renumbering Phase 3 into a monotonic block.** 128 step references across Phases 3–6, 27 of them the ambiguous `Step 12`, and it would touch Phase 4/5/6 steps this change does not. Mitigated with explicit `Step order` lines in both phase headers and in `SKILL.md`.
- **`stage-exit` matrix omits `wait-for-user`** (documented 40 lines above). Pre-existing, unrelated to this change.
- **`uip solution upload` tolerance for a `placeholder` connectorKey** is untested — the preview publishes such rules. Full `validate` passes and a disclosure line is printed, but it wants one live run.

Round 5 review queued.

song-zhao-25:
Closing — the scope grew well past the original goal. The core change (authoring conditions and SLA in Phase 2 so the Studio Web preview shows flow rather than disconnected boxes) got bundled with an unrelated `--skeleton-v2` gate-profile change that is blocked on an unshipped CLI flag, plus a validate-cadence reconciliation that was not needed and turned a dormant inconsistency into an active one requiring two carve-outs.

Nothing here is lost: the work is one commit on `feat/case-phase2-rules-sla`, and `main` is untouched. Re-opening as a tighter, single-concern PR if and when we pick it back up.

Empirical findings worth keeping regardless of what happens to this branch:
- `--skeleton-v2` exists in no published CLI — the string is absent from the 1.198.0 and 1.199.0-preview.108 `maestro-tool` tarballs, and 1.199.0-preview.99 rejects it with `ErrorCode: invalid_argument`.
- Full-mode `validate` is clean on a complete Phase-2 caseplan (connector task with only typeId/connectionId, `value: ""` inputs, root slaRules + escalation, sla-status-change entry rule, stub wait-for-connector rules) and catches five defects `--skeleton` reports as `Valid`. So a narrower gate profile is a nice-to-have, not a prerequisite.
- No condition or SLA rule-type in any of the four scopes depends on anything minted in Phase 3 — verified independently across four review rounds. The phase move itself is sound.

PR #2484, opened 2026-08-05
docs(uipath-maestro-flow): note the flag-gated nested inline-agent input encoding

akraminakib:
## What

Adds a short forward-compat note to the `uipath-maestro-flow` inline-agent docs: flow-workbench gates a NESTED inline-agent input encoding behind the `services.agent-storage.nested-agent-input-schema` feature flag (default **off**), so an `agent.json` read from disk may arrive in either form, while the flat `__` form these docs teach remains the shipped encoding and the one to author.

- `references/author/references/plugins/inline-agent/impl.md`: encoding note under "Wiring Flow Variables into Agent Prompts" naming the flag, the dotted `{{input.a.b.c}}` token form, and the `x-uipath-container` schema-tree marker.
- `references/shared/node-output-wiring.md`: one-line pointer on the inline-agent prompt row.

## Why

flow-workbench (UiPath/flow-workbench#2851, stacked on #2520/#2755) puts the nested writer behind the flag so the CLI and coding-agent skills can catch up at their own pace. Canvas, CLI, and runtime readers are dual-mode, and CLI packaging heals nested sidecars back to flat (UiPath/cli#3304), so the authoring guidance here stays flat until the flag ramps. Without the note, an agent reading a nested-authored `agent.json` would have no vocabulary for what it is seeing.

The flat `reference_agents` fixtures under `tests/` are deliberately untouched: they match the shipped encoding while the flag defaults off.

Related: UiPath/flow-workbench#2851, UiPath/cli#3304


skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md
```diff
@@ -49,6 +49,8 @@ After editing `content`, rebuild the matching `messages[].contentTokens` (`type:
...
 
+> **Encoding note (MST-12020).** The flat `__` form above is the shipped encoding and the one to author. flow-workbench also has a NESTED encoding behind its `services.agent-storage.nested-agent-input-schema` feature flag (default off): dotted prompt tokens (`{{input.a.b.c}}`) and an `inputSchema` object tree whose auto-created containers carry `"x-uipath-container": true`. An `agent.json` you READ may arrive in either form (canvas, CLI, and runtime readers are dual-mode, and CLI packaging heals nested sidecars back to flat) — but keep WRITING the flat form until the flag ramps.
```
snuziale: this MST tickets are not useful here...

akraminakib: Good call, removed in edbc241de. The doc keeps only the flag key and the marker name (the things an agent can actually act on); the tracking context lives in the linked PRs instead.

skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md
```diff
@@ -49,6 +49,8 @@ After editing `content`, rebuild the matching `messages[].contentTokens` (`type:
...
 
+> **Encoding note.** The flat `__` form above is the shipped encoding and the one to author. flow-workbench also has a NESTED encoding behind its `services.agent-storage.nested-agent-input-schema` feature flag (default off): dotted prompt tokens (`{{input.a.b.c}}`) and an `inputSchema` object tree whose auto-created containers carry `"x-uipath-container": true`. An `agent.json` you READ may arrive in either form (canvas, CLI, and runtime readers are dual-mode, and CLI packaging heals nested sidecars back to flat) — but keep WRITING the flat form until the flag ramps.
```
rockymadden, line 52: Does the agent know what flow-workbench mean? Do we need to tell it about things behind feature flags?

akraminakib: I dropped it, along with the flag id and the `x-uipath-container` marker.

On whether to mention the flag at all: I cut it down to what the agent actually needs. Write the flat `__` form, and if you open an `agent.json` that's already nested, leave it alone. That second part matters because once the flag ramps, canvas users will produce nested files the agent may open, and it shouldn't try to "fix" them. The rest is handled for it: readers accept both forms, and packaging converts nested back to flat.

I also dropped the "until the flag ramps" line, which was misleading. Writing flat is always fine, so the agent never needs to switch.

skills/uipath-maestro-flow/references/author/plugins/inline-agent/impl.md
```diff
@@ -51,6 +51,8 @@ Passing flow data into an inline agent requires **three hand-authored, aligned**
 
 Self-check the CLI you run: `uip maestro flow validate` must pass a prompt-less node and must fail empty-string prompts. Never add stub prompts. If validate rejects the absent keys, see § Refresh and Validate § Older CLI.
 
+> **Encoding note.** Author the flat `__` form shown above — it is always valid and is what packaging ships to the runtime, so keep writing it. A newer nested encoding also exists (dotted prompt tokens like `{{input.a.b.c}}` with a matching nested `inputSchema`); **do not produce it**. Reading is tolerant either way — the CLI and runtime accept both forms, and packaging normalizes a nested `agent.json` back to flat — so an `agent.json` you open may already be in the nested form; leave it as-is and keep writing flat.
```
bryan-uipath: I wonder if we could get away with removing this line?
> Reading is tolerant either way — the CLI and runtime accept both forms, and packaging normalizes a nested `agent.json` back to flat — so an `agent.json` you open may already be in the nested form; leave it as-is and keep writing flat.

I'm not sure it is "load-bearing" 🤖 and I'm also looking for ways to have less in skills.

bryan-uipath (approved):

PR #2508, opened 2026-08-06
fix(maestro-flow): surface Write-clobbers-bindings guardrail on the greenfield path

rockymadden:
## Problem

A repeat eval failure in `skill-flow-slack-weather-pipeline` had `flow debug` fault at the first connector node while `flow validate` passed. Root cause: the final `.flow` shipped with a corrupted top-level `bindings[]` — the Slack connector node resolved its folder (matched **name-only** on the literal `"FolderKey"`, per `connector/impl.md`) to the *other* connection's folder, so the connection wasn't found there.

The mechanism the agent hit: it authored by `Write`, repaired `inputs.detail`/`bindings[]` with `node configure`, then did **another full-file `Write`** (a one-field script tweak) and never re-ran `node configure` — re-clobbering the CLI-owned bindings. `validate` doesn't check binding/folder consistency, so it passed.

The #2450 guardrail already warns against this, but it lived **only** in `editing-operations.md` (Tool Selection Ladder rung 3). The failing agent never opened that file — on the greenfield connector-authoring path it read `SKILL.md`, `CAPABILITY.md`, `greenfield.md`, and the plugin `impl.md` files, none of which carried the warning. And the remedy ("re-run `node configure` afterward") never said `configure` must be the **last** write, so a later `Write` silently undoes it.

## Changes (docs-only)

- **`CAPABILITY.md` — Node ownership §**: add the "never `Write` a flow that contains CLI-owned nodes" rule where node ownership is defined (the SSOT the agent actually reads).
- **`greenfield.md`**: warn at the Step-4 whole-file `Write` decision point, and qualify the "`Write` only when ≥70% of nodes change" recap so it no longer contradicts the rule.
- **`editing-operations.md` — ladder rung 3**: state that if you do `Write`, `node configure` must run **as the last write** to touch `inputs.detail`/`bindings[]`, because any later full-file `Write` re-clobbers it.

Per review feedback, the guidance describes the symptom generally (surfaces at `flow debug`) rather than pinning a specific backend error code that may change over time.

## Testing

Docs-only; the pre-commit hook validates SKILL.md descriptions only (unaffected). No behavior or test changes — the eval's end-to-end debug check is intentionally left as-is.


skills/uipath-maestro-flow/references/author/references/editing-operations.md
```diff
@@ -8,7 +8,7 @@ Strategy selection and shared concepts for modifying `.flow` files. Direct `.flo
...
-> 3. **Wholesale file rewrite** (only when ≥70% of nodes change, e.g., scaffolding from a template) → `Write` — but never on a flow that already contains CLI-owned nodes (connector, connector-trigger, managed HTTP): the rewrite clobbers their CLI-owned `bindings[]` / `inputs.detail`, and `flow validate` won't catch it. Use `Edit` (rung 2), or re-run `node configure` afterward.
+> 3. **Wholesale file rewrite** (only when ≥70% of nodes change, e.g., scaffolding from a template) → `Write` — but never on a flow that already contains CLI-owned nodes (connector, connector-trigger, managed HTTP): the rewrite clobbers their CLI-owned `bindings[]` / `inputs.detail`, and `flow validate` won't catch it — the corruption surfaces only later, as a resolution failure at `flow debug`. Use `Edit` (rung 2). If you do `Write`, re-run `node configure` for every CLI-owned node **as the last step** — it must be the final write to touch `inputs.detail` / `bindings[]`, because any later full-file `Write` silently re-clobbers them (validate still passes).
```
jiyangzh: https://github.com/UiPath/skills/blob/4eb569b08785782266f8ca64ee6b8f5b58c24492/skills/uipath-maestro-flow/SKILL.md#L88-L89

> Every node has exactly one author

If agent reads this and SKILL.md, it may skip the rule we added here

rockymadden: Good catch — this is the highest-leverage spot, and the failing run confirms it: the agent's first action was reading all of SKILL.md, so rule #9's unqualified "Edit / Write" for user-owned nodes is exactly what permitted the fatal `Write`.

Fixed in 698a4b4:
- **SKILL.md rule #9** now carries a one-clause caveat right at the `Edit / Write` phrase (never a full-file `Write` on a flow that also has CLI-owned nodes; `Edit` in place or `node configure` last). It already links CAPABILITY.md + editing-operations.md as canonical.
- To avoid multiplying near-identical wording (the review-bot's duplication note), I made **CAPABILITY.md — Node ownership** the single canonical full statement and trimmed the greenfield.md / editing-operations.md occurrences to short flags that point to it — so a future wording change touches one place.

jiyangzh (approved):

PR #2510, opened 2026-08-06
test(uipath-platform): add traces coverage tests for spans and feedback gaps

saksharthakkar:
## Motivation

A scoped `/test-coverage` pass over the `uipath-platform` **Traces** slice (`skills/uipath-platform/references/traces/`) found **6 of 24 components directly tested (25%)**. The untested 18 are the parts an agent is most likely to get wrong: the documented two-tool discovery flow, every feedback read path, and the whole of `update` / `delete`.

`mode:diagnose` was the lowest-scoring mode in the slice, because every diagnose-labelled capability was reached only by tests that *start a healthy job and then read its spans* — an operate-shaped path wearing a diagnose label.

> The 64% shown for Traces on the org Confluence scorecard is a stale whole-skill blend applied identically to six unrelated product rows. Fixing that attribution bug is a separate non-test change and is deliberately **not** in this PR.

## Summary

**Coverage: 6/24 → 23/24 direct (25% → 96%).** Seven new task YAMLs under `tests/tasks/uipath-platform/traces/`. No existing test was weakened or deleted.

image

| Task | Tier / mode | Gaps closed |
|---|---|---|
| `traces_spans_discovery_smoke` | smoke · diagnose | `or jobs traces` two-tool discovery, positional `spans get <trace-id>`, `--folder-path`/`--folder-key` context |
| `traces_feedback_negative_smoke` | smoke · build | `--negative`, `--comment`, `--agent-id`/`--agent-version`, root-span default (`--span-id` deliberately omitted) |
| `traces_feedback_comment_file_smoke` | smoke · build | `--comment-file <path>`, `--comment-file -` (stdin), repeatable `--category`, comment/comment-file mutual exclusion |
| `traces_feedback_list_filters_smoke` | smoke · diagnose | `list` graded at last, `--span-id`/`--agent-id`/`--agent-version`/sentiment filters, `--limit`/`--offset` paging |
| `traces_feedback_list_detailed_smoke` | smoke · diagnose | entire `list detailed` subcommand, `--since`, `--after`/`--before`, `--category-id`, `--sort-by`/`--sort-order` |
| `traces_feedback_mutate_smoke` | smoke · build | `update <id>`, **agent-invoked** `delete <id>`, category replacement-not-additive semantics, `--folder-key` + `-y` on writes |
| `traces_diagnose_faulted_job_e2e` | e2e · diagnose | standalone diagnose archetype — reads a run the agent did **not** start ⚠️ *fixture pending, see below* |

### Criteria grade values, not just flag presence

Where the value is the actual skill under test, criteria bind to it:

- the second page of a five-per-page walk must be `--limit 5 --offset 5` — an agent treating `--offset` as a page number fails;
- `--sort-order desc` for a "newest first" ask — `asc` does not score;
- an update that adds a tag must re-pass the existing one, because `--category` **replaces** rather than appends. An agent that passes only the new tag silently drops the old one and fails. This is the sharpest discriminator in the set (weight 3.0).

The `-y`-on-delete and no-`--span-id`-on-a-directly-invoked-agent rules are graded structurally rather than by prose.

### Two CLI documentation bugs found and fixed

Both were caught by checking every flag against the real CLI before writing the assertions. Tests written against the documented spelling would have asserted commands the CLI rejects:

1. **`list detailed` sort flags** — `feedback.md` documented `--sortBy` / `--sortDir`; the CLI takes `--sort-by` / `--sort-order`. Every other skill in this repo uses the kebab form; `feedback.md` was the lone outlier (it looks copied from the REST query params).
2. **`feedback delete` requires `-y`** — "the CLI never prompts", and `feedback.md` did not mention it at all. Consequence: `cleanup_traces_feedback_e2e.py` has been silently failing its delete on every e2e run (it swallows the exit code and prints `WARN`), so e2e feedback records have been leaking on the tenant. That script now passes `-y`.

## Tests performed

**All 6 new smoke tasks pass locally, 6/6, every criterion, score 1.000.**

```
coder-eval run tasks/uipath-platform/traces/*_smoke.yaml -e experiments/default.yaml
Results: 6/6 succeeded   (19/19 criteria passed)
```

| Task | Status | Score | Turns |
|---|---|---|---|
| `skill-platform-traces-spans-discovery` | SUCCESS | 1.000 | 18 |
| `skill-platform-traces-feedback-negative` | SUCCESS | 1.000 | 9 |
| `skill-platform-traces-feedback-comment-file` | SUCCESS | 1.000 | 12 |
| `skill-platform-traces-feedback-list-filters` | SUCCESS | 1.000 | 8 |
| `skill-platform-traces-feedback-list-detailed` | SUCCESS | 1.000 | 19 |
| `skill-platform-traces-feedback-mutate` | SUCCESS | 1.000 | 8 |

Run with `experiments/default.yaml` (tempdir) rather than `smoke.yaml`, because the `skills-image:latest` Docker image is not available on the authoring machine. Fixture GUIDs are fabricated, so every call returns auth-error/404 — the acceptable outcome under the mock-auth model this slice already uses.

Also verified before committing:

- every `command_pattern` fixture-tested against realistic **and backslash-continued** command strings, in both match and must-not-match directions. This is why lookaheads use `[\s\S]*` rather than `.*` — the agent really does emit multi-line continued commands, and `.*` would have silently under-matched every one of them;
- `check_traces_diagnose_e2e.py` self-tested against five synthetic payloads (healthy, real-fault-correct-span, real-fault-wrong-span, invented ID, bare-list form);
- `scripts/check-cli-verbs.py` clean on all 7;
- every flag confirmed against `uip --help` output, not just the reference docs.

## Test plan

- [x] 6 smoke tasks pass locally with coder-eval
- [x] Independent lint pass against the `/lint-task` rubric; all findings triaged, the substantive ones fixed (see below)
- [ ] `run-coder-eval.yml` triggered on this branch — result appended as a comment
- [ ] `traces_diagnose_faulted_job_e2e` — **needs a human with live-tenant access**

## Not closed — follow-ups

**1. `traces_diagnose_faulted_job_e2e` has no fixture.** It needs a job that has *already faulted* on the tenant; the whole point is that the agent inspects a run it did not start, so the test cannot manufacture its own subject. Someone with live-tenant access needs to:

- provision a deliberately-failing Agent-type job in `Shared/uipath-platform`;
- export its key as `E2E_FAULTED_JOB_KEY` to the runner (`pre_run` inherits the orchestrator environment, so no experiment passthrough entry is needed);
- run it once and confirm the check script's fault heuristic matches the real span schema — I could not verify that against live data, and it is the one part of this PR written blind.

Until then `pre_run` exits with `FIXTURE NOT PROVISIONED` rather than scoring the agent. **This means the task will ERROR if picked up by a nightly/e2e sweep before the fixture lands** — reviewers should decide whether to provision the fixture first or accept one ERROR row in the interim. I deliberately did not fabricate a passing claim or weaken its criteria to force a green.

**2. Component 24 — `spanAttributes` payload shape / 200-item cap — remains indirect.** Asserting the payload requires live feedback data; the smoke test can only grade that the agent chose `list detailed` (the sole command that returns span context) over plain `list`. Closing it properly means an e2e with seeded feedback records.

## Judgment calls

- **Fixed `feedback.md` rather than testing the wrong spelling.** Out of the literal "add test tasks" scope, but shipping tests that assert flags the CLI rejects would be worse. Both fixes verified against `uip --help`.
- **`lifecycle:setup` on the mutating feedback tasks**, though the existing feedback tests use `lifecycle:discover`. Per the taxonomy, `setup` is "mutate tenant state: create/edit/delete resources", which `create`/`update`/`delete` plainly are. New tasks follow the definition; existing tags left alone.
- **`mode:diagnose` on `traces_spans_discovery`** (the plan suggested `operate`). The README puts "inspecting traces" under `diagnose`, the framing is a support-ticket investigation, and the slice's diagnose floor is the gap worth closing.
- **Files kept flat in `traces/`** rather than one-per-leaf-directory. The rule is dead-letter in practice — 53 directories under `tests/tasks/` already hold more than one task — and splitting would break consistency with the 4 existing siblings.
- **Prompts state that identifiers may not resolve against the tenant.** Not procedure leakage — it is environmental scoping. Without it the first run had an agent paginate `uip or jobs list` across 10 offsets hunting for a fabricated job key (65 turns, 609s, over `smoke.yaml`'s 40-turn cap); with it, the same task finishes in 18 turns and 63s. The mutate task likewise scored 0/4 on the first run because the agent pre-checked the records, got `not found`, and correctly declined to mutate — grading nothing.
- **Negative guards use `command_executed` + `exclude_pattern`, not `command_not_executed`.** A bare `command_not_executed` passes for free when the agent runs no commands at all; the positive form requires the command to actually happen *and* lack the flag.

### DIFF
--- a/skills/uipath-platform/references/traces/feedback.md
+++ b/skills/uipath-platform/references/traces/feedback.md
@@ -86,7 +86,7 @@ uip traces feedback list detailed \
   --output json
 ```
 
-Additional flags over `list`: `--since <duration>`, `--after <ISO>`, `--before <ISO>`, `--category-id <guid>`, `--sortBy`, `--sortDir`. Max 200 items.
+Additional flags over `list`: `--since <duration>`, `--after <ISO>`, `--before <ISO>`, `--category-id <guid>` (repeatable), `--sort-by <createdAt|updatedAt>`, `--sort-order <asc|desc>`. Max 200 items.
 
 ## update
 
@@ -102,9 +102,12 @@ uip traces feedback update <feedback-id> \
 
 ## delete
 
+`-y` is required — the CLI never prompts, so a delete without it is rejected.
+
 ```bash
 uip traces feedback delete <feedback-id> \
   --folder-key <folder-key> \
+  -y \
   --output json
 ```
 
--- a/tests/tasks/uipath-platform/traces/check_traces_diagnose_e2e.py
+++ b/tests/tasks/uipath-platform/traces/check_traces_diagnose_e2e.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python3
+"""Verify the standalone-diagnose run landed on a real failing span.
+
+Grades against spans.json — raw `uip traces spans get` output — not against any
+prose the agent wrote. failing_span.txt must name a span ID that actually exists
+in that payload and that carries fault evidence, so a guessed or invented ID fails.
+"""
+import json
+import re
+import sys
+from pathlib import Path
+
+# Word-anchored: a bare "fault" substring also matches "default", which appears in
+# ordinary folder names and attribute keys and would make the check vacuous.
+FAULT_KEY = re.compile(r"\b(error|exception|fault|failed|failure|stack_?trace)", re.I)
+FAULT_VALUE = re.compile(r"\b(error|exception|faulted|failed|failure)", re.I)
+
+
+def load(path):
+    p = Path(path)
+    if not p.is_file():
+        sys.exit(f"FAIL: {path} not found")
+    return p
+
+
+try:
+    payload = json.loads(load("spans.json").read_text())
+except json.JSONDecodeError as e:
+    sys.exit(f"FAIL: spans.json is not valid JSON: {e}")
+
+if isinstance(payload, list):
+    spans = payload
+else:
+    if payload.get("Result") not in (None, "Success"):
+        sys.exit(f"FAIL: Result={payload.get('Result')!r}, Message={payload.get('Message')!r}")
+    spans = payload.get("Data") or []
+
+if not isinstance(spans, list) or not spans:
+    sys.exit("FAIL: spans.json contains no spans")
+
+
+def fields(span):
+    """Flatten a span's own fields plus its Attributes blob into (key, value) pairs."""
+    out = []
+    for k, v in span.items():
+        if k == "Attributes" and isinstance(v, str):
+            try:
+                parsed = json.loads(v)
+            except json.JSONDecodeError:
+                out.append((k, v))
+                continue
+            if isinstance(parsed, dict):
+                out.extend(parsed.items())
+            continue
+        out.append((k, v))
+    return out
+
+
+def is_faulting(span):
+    for key, value in fields(span):
+        if value in (None, "", [], {}, False):
+            continue
+        if FAULT_KEY.search(str(key)):
+            return True
+        if isinstance(value, str) and FAULT_VALUE.search(value):
+            return True
+    return False
+
+
+by_id = {str(s["Id"]): s for s in spans if isinstance(s, dict) and s.get("Id")}
+if not by_id:
+    sys.exit("FAIL: no span in spans.json carries an Id")
+
+faulting = {sid for sid, s in by_id.items() if is_faulting(s)}
+if not faulting:
+    sys.exit("FAIL: no span carries fault evidence — the fixture job may not have faulted")
+if len(faulting) == len(by_id):
+    sys.exit(
+        f"FAIL: every one of {len(by_id)} spans matches the fault heuristic — the check is "
+        "vacuous and cannot distinguish the failing span. Tighten the markers."
+    )
+
+reported = load("failing_span.txt").read_text().strip()
+if not reported:
+    sys.exit("FAIL: failing_span.txt is empty")
+if reported not in by_id:
+    sys.exit(
+        f"FAIL: failing_span.txt names {reported!r}, which is not a span in spans.json "
+        f"({len(by_id)} span(s) returned)"
+    )
+if reported not in faulting:
+    sys.exit(
+        f"FAIL: span {reported} carries no fault evidence — the diagnosis points at a span "
+        f"that did not fail. Faulting spans in this trace: {sorted(faulting)}"
+    )
+
+print(f"OK: {reported} is one of {len(faulting)} faulting span(s) across {len(by_id)} span(s)")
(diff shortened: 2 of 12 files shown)


saksharthakkar:
## Passing-run evidence — `run-coder-eval.yml`

**8/8 PASS (linux)** · [run 31129147625](https://github.com/UiPath/skills/actions/runs/31129147625) · agent `claude` · `experiments/nightly.yaml` (docker driver, live tenant via the CI bot token)

Glob: `tasks/uipath-platform/traces/*_smoke.yaml tasks/uipath-platform/traces/traces_fetch.yaml`

| Task | Status | Score | Duration |
|---|---|---|---|
| `skill-platform-traces-spans-discovery` | SUCCESS | 1.000 | 47.8s |
| `skill-platform-traces-feedback-negative` | SUCCESS | 1.000 | 37.7s |
| `skill-platform-traces-feedback-comment-file` | SUCCESS | 1.000 | 62.2s |
| `skill-platform-traces-feedback-list-filters` | SUCCESS | 1.000 | 49.8s |
| `skill-platform-traces-feedback-list-detailed` | SUCCESS | 1.000 | 64.9s |
| `skill-platform-traces-feedback-mutate` | SUCCESS | 1.000 | 45.9s |
| `skill-platform-traces-feedback-smoke` *(pre-existing)* | SUCCESS | 1.000 | 30.0s |
| `skill-platform-traces-fetch` *(pre-existing)* | SUCCESS | 1.000 | 46.8s |

The two pre-existing tasks are included deliberately as a regression check — the `feedback.md` sort-flag and `-y` corrections in this PR touch the doc they read from, and both still pass.

The two e2e tasks (`traces_e2e`, `traces_feedback_e2e`) were left out of this run to keep it cheap; they are untouched by this PR. `traces_diagnose_faulted_job_e2e` was also excluded — it has no fixture yet and would ERROR by design (see the PR description).

### Other CI on this PR

- **Socket Security** — pass.
- **WIP** — pending, and correct: this PR is intentionally staying in draft.
- **lint-tasks** — does not run on drafts by design (`.github/workflows/lint-tasks.yml` gates on `draft == false`). It will run when a human marks this ready. As a stand-in I ran an independent review against the `/lint-task` rubric while authoring; its substantive findings are fixed in the committed version — value-bound criteria, `exclude_pattern` negative guards, and a real bug in the e2e check script where a bare `"fault"` substring also matched `"default"` and made the fault heuristic vacuous.
- **validate-skills / validate-skill-status / activation-gate** — not triggered; their path filters target `skills/*/SKILL.md`, and this PR only touches a reference file under that skill.

Nothing red is attributable to this change.

JosephMar (approved):

PR #2512, opened 2026-08-07
fix(uipath-test): dedicated fixtures per task; fix two skill defects [TMHUB-32404]

ganeshborle:
## Problem

The `uipath-test` suite failed six tasks on the 2026-08-05 terra nightly — and a *different* set on each of the three nightlies before it. Scores moved every night while the skill under test never changed.

Root cause: four tasks shared CLAIM's `UAT Smoke Suite` and **two of them wrote to it**. `execution_rerun` retried failed cases, reopening logs that nothing closes on a robot-less tenant; cleanup then closed them as `Passed`. The suite converged on 3-pass/0-fail, so `flaky_tests_analysis` lost the intermittency it reads and `execution_rerun` lost its own premise. Tasks were failing for reasons unrelated to the skill.

## The `EVFX-` fixture contract

[`tests/fixtures/README.md`](tests/fixtures/README.md) — one owner per fixture, create-if-absent, append-only, **never deleted**, never name-colliding.

The rule that actually fixes the dilution: a task needing execution history **appends a new execution** with the pattern it requires rather than rewriting an existing one. Rewriting is precisely how one task laundered another's seeded failures into passes.

| Task | Owns | Was |
|---|---|---|
| `execution_rerun` | `EVFX-RERUN-SET` | wrote to the shared suite |
| `flaky_tests_analysis` | `EVFX-FLAKY-SET` | read history two tasks destroyed |
| `release_signoff` | `EVFX-SIGNOFF-SET` | wrote to the shared suite |
| `test_report_junit_export` | `EVFX-JUNIT-SET` | blocked on a stale `Pending` execution |
| `organize_testcases` | `EVFX-ORG-<date>-*` | accumulated ambiguous duplicates |
| `release_readiness` | BANK suite (task-exclusive) | had no recorded execution |

`release_readiness` deliberately keeps its discoverable name — the task grades *finding* the suite, so renaming it would delete the thing under test.

[`tests/fixtures/check-collisions.py`](tests/fixtures/check-collisions.py) enforces sole ownership and namespace discipline in CI, so this can't erode back into a convention.

## Two skill defects the stable fixtures exposed

Both affect real users, and both were invisible while fixture noise moved scores nightly.

1. **`SKILL.md` Critical Rule 10** told agents to set the default Orchestrator folder before *every* run. Agents obeyed, picked an arbitrary folder off `uip or folders list`, and got **HTTP 500** on projects that already had a working one — scoring `release_signoff` 0.12. Now: attempt the run first, leave a working default alone.
2. **`test-result-report-guide.md`** instructed agents to "check that all required sections for the persona are present" against a list that **did not exist**, describing report content in prose instead. The criterion it feeds is literally named *"regression section taught by skill"* and tests for a markdown heading. Each persona now has an explicit required-heading table.

## Criteria corrected to grade outcomes, not command shape

Each of these docked a run that reached the correct result:

- `folders list -n X` vs `--name X`, and flag **order** (`--all --name X`) → order-independent lookahead
- `testcases delete --test-case-id` vs `--test-case-key` — both delete the case
- `min_count: 2` counts *tool calls*, so chaining two `customfield create` calls scored 1/2 → split into two name-anchored criteria
- `workflow_impact` demanded `executions list` while the agent used `testsets list --include-last-execution` + `get-stats` — fewer calls, same information
- `release_signoff` required the report unconditionally while the prompt asked for it only "once it comes back" — and the wait always times out
- `testcase_steps` / `objectlabel` prompts implied deliverables rather than requiring them

## Verification

**Two consecutive full-suite passes on byte-identical code**, `codex/gpt-5.6-terra`: **18/18 with every task at exactly 1.000**.

- [31073656005](https://github.com/UiPath/skills/actions/runs/31073656005)
- [31074141570](https://github.com/UiPath/skills/actions/runs/31074141570)

Before the skill fixes, two passes on identical code returned 18/18 then 15/18 — which is what motivated fixing the skill rather than hardening the prompts.

Every one of the six Aug-5 failures is green and has stayed green across every pass since:

| Task | Aug-5 | Now |
|---|---|---|
| `release_signoff` | 0.12 | 1.000 |
| `workflow_impact` | 0.47 | 1.000 |
| `organize` | 0.50 | 1.000 |
| `execution_rerun` | 0.54 | 1.000 |
| `flaky` | 0.65 | 1.000 |
| `release_readiness` | 0.68 | 1.000 |

**No robot is required** — fixtures build their history through `testsets run --execution-type manual` + `testcaselog finish`, which is pure API. A robot (or rather, a folder with a serverless machine) is only needed if we later move to Playwright packages for genuine executions.

## Notes for reviewers

- Supersedes #2439 (`fix/uipath-test-workflow-impact-traversal`), whose single commit is included here.
- A deliberate reversal worth flagging: I first fixed `report_generation` by naming the required sections **in the task prompt**, then reverted it. That would have turned the task green while grading the prompt instead of the skill — a false pass hiding the defect from real users. The section list belongs in the guide, which is where it now lives.


aoltean16 (approved):

skills/uipath-test/references/test-result-report-guide.md
```diff
@@ -48,6 +48,27 @@ Detailed instructions for generating persona-tailored test reports from UiPath T
...
+
+   | Persona | Required headings |
+   |---|---|
+   | All personas | `## Summary` · `## Test Set` (name and key) · `## Results Breakdown` (counts for passed / failed / none / restricted) · `## Frequently Failing Test Cases` (step 4 requires this metric in every report) |
+   | QA Engineer | plus `## Regressions` |
+   | Developer | plus `## Failed Assertions` (assertion message per failing test case log) |
+   | Release Manager | plus `## Go / No-Go` (success rate, blocker count, risk assessment) |
+
+   Keep the heading wording — downstream readers and graders look for these
+   exact section names. Add a persona-appropriate extra section when the data
```
vaishalisharma-uipath: `downstream readers and graders look for these exact section names.` I do not see these heaidinsg being used in the subsequent sections. Let's remove it if it's added for tests

skills/uipath-test/references/test-result-report-guide.md
```diff
@@ -48,6 +48,27 @@ Detailed instructions for generating persona-tailored test reports from UiPath T
...
+   warrants it, but never drop a required one; write "None" under a heading
+   that has no content rather than omitting the heading.
+
+   Derive the Results Breakdown counts by tallying the test case logs you
+   already fetched in step 1. Do NOT build a `--output-filter` aggregate
+   expression to compute them: a malformed filter aborts the command, and
+   under Critical Rule 11 that stops the whole report for a total you could
+   have counted from data already in hand.
```
vaishalisharma-uipath: 1. This instruction looks more fitting in section 4's ` Count of test case logs per execution with result`
2. The point about guessing filter can be put in 'anti-pattens' section at the bottom

skills/uipath-test/SKILL.md
```diff
@@ -238,8 +238,8 @@ Object labels are tag-style metadata applied to Requirement, TestCase, TestSet,
...
+10. **`run` needs a default Orchestrator folder — set one only if the project lacks it.** `uip tm testcases run` and `uip tm testsets run` both require a default folder on the project, but most projects already have a working one. Attempt the `run` first; only if it fails for a missing or invalid default folder, set one with `uip tm project set-default-folder --project-key <PROJECT_KEY> --folder-key <FOLDER_KEY> --output json`, then retry. Get folder keys with `uip or folders list -n <folder-name> --all --output json`. Never overwrite a default folder that is already working — an arbitrary folder picked off `or folders list` may be one Test Manager rejects, breaking every subsequent run on that project. This recovery is the one named exception to Rule 11: a run that fails *specifically* for a missing or invalid default folder is recoverable without asking. Any other run failure still stops.
```
vaishalisharma-uipath: Can we add the instruction around the exact exceptions that are thrown for the two exceptions - invalid folder, timeout in `uip tm wait`. The instruction is always work with existing folder is not accurate.

vaishalisharma-uipath (approved):
Please address the comments before merging.

skills/uipath-test/references/test-result-report-guide.md
ganeshborle: Removed in `0c5ad41b6` — the graders-oriented rationale was indeed test-facing, so that sentence is gone. The heading table itself stays (it defines the required report sections) along with the don't-drop-a-required-heading guidance.

skills/uipath-test/references/test-result-report-guide.md
ganeshborle: Done in `0c5ad41b6`: (1) the tally-the-fetched-logs instruction now lives on step 4's counts bullet ('…derive these counts by tallying the test case logs already fetched in step 1'); (2) the --output-filter warning moved to the Anti-patterns section at the bottom. The paragraph is removed from step 6.

skills/uipath-test/SKILL.md
ganeshborle: Added in `0c5ad41b6`, with the exact errors verified against the live CLI (probed on a scratch project, then deleted): missing default folder on run → `HTTP 400: Please assign folder to project level before executing testcases.` (errorCode itemNotFound) — the only signal that triggers set-default-folder + one retry. `uip tm wait` timeout → exit code 2 with `Timed out after <N>s waiting for execution '<id>'. Last status: <status>.` (other poll outcomes exit 1 and stop). An unusable folder surfaces as an opaque HTTP 5xx on run, and set-default-folder with a nonexistent key fails with `HTTP 400: Internal Server Error.` — both now stop under Rule 11 instead of being treated as folder-recovery. You're right the blanket 'work with existing folder' framing wasn't accurate — the rules are now keyed to the exact error text.

PR #2516, opened 2026-08-07
feat(uipath-agents): BYO guardrails for low-code and coded agents [AL-513]

apetraru-uipath:
## What changed?

Documents bring-your-own guardrails (BYOG) for **both agent types** and adds tests for the thread users actually hit: adding a BYOG guardrail to an agent they already have. Ticket: AL-513.

BYOG lets a tenant serve a validator from its own external provider (Azure AI Content Safety, Databricks, a custom connector) instead of UiPath's built-in implementation. An admin registers the configuration; agents reference it by name. None of the agent-authoring side was documented.

**Docs (7 files)**

- **`uipath-agents` low-code** — `Validator` is not unique once a BYOG configuration exists (a tenant sees two `pii_detection` entries, one built-in and one BYO, disambiguated by `IsByo`), the `--byo` filter, the extra `Byo*` fields, and how to pin a guardrail with `byoValidatorName`.
- **`uipath-agents` coded** — the BYO section was previously a gate with the API left unresolved: it named no class and told the agent to *stop and report BYO unavailable*, which now reads as a false negative. It now carries:
  - an availability table — `UiPathByoGuardrailMiddleware` is LangChain-adapter only; `ByoValidator` is a **core** class re-exported by the adapter, so **BYO is not LangChain-only**;
  - worked examples for both middleware and decorator styles, with source-verified signatures;
  - the wire format both emit — `validatorType: "byo"` + `byoValidatorName`, the same field low-code pins with, so the two halves converge;
  - Critical Rule 18 rewritten, plus pointers from the Middleware and Decorator style sections.
- **Both `guardrails-recommend` files** — key validator lookups on `(Validator, IsByo)`, not `Validator` alone, or a same-named built-in and BYO entry collide and correctness checks read the wrong entry's `Parameters`/`AllowedScopes`. Default recommendations to the built-in unless the user asks for BYO.
- **`uipath-review`** (both guardrail files) — the same disambiguation, plus: `Status: Disabled` on a BYO entry is a tenant configuration switch, not a schema defect — don't re-diagnose it as one.
- **`uipath-troubleshoot`** — BYOG as a cause of a guardrail violation, how to cross-check configuration health, and the resolution path for a disabled configuration or dead connection.

**Corrections found while writing the above**

- `byoConfigurationId` was never a real authoring field — `ByoValidatorName` is what both low-code and coded agents reference. Swept everywhere.
- Dropped the last "the BYO construct carries a connection id" claim, which contradicted the same file, Rule 18, and the review skill. No connection id exists in either API; the platform resolves it server-side.
- The coded doc claimed middleware never accepts `stage=`. Source shows five do (PII, harmful content, LLM-as-judge, deterministic, BYO); only the three fixed-stage validators don't. Pre-existing bug, corrected in place.

**Tests (7 files) — adding BYOG to an existing agent**

Three tasks, all editing a pre-built fixture with **no scaffolding graded**, so the score reflects only the BYOG wiring. All three tagged `lifecycle:edit` (they edit an existing agent rather than scaffolding one).

| Task | Tier | What it grades |
|---|---|---|
| `lowcode/guardrails/byog_pinning` | `smoke` | `byoValidatorName` pinned in the shared `WebResearchBriefingSolution` fixture's `agent.json` |
| `coded/guardrails/byog_middleware` | `e2e` | `UiPathByoGuardrailMiddleware` spread into `create_agent()` on the `SimpleCodedAgent` fixture |
| `coded/guardrails/byog_decorator` | `e2e` | `ByoValidator` + `@guardrail` over a real factory on the same fixture |

- **The low-code task mocks discovery**, which is what removes the tenant dependency entirely — no feature flag, no fixture registration, no create-time connection probe. A selective `uip` shim (`mock_path_dirs`, following the ixp mock pattern) serves `uip agent guardrails list` with a BYO entry **alongside the same-named built-in**, forcing the `IsByo` disambiguation the skill teaches; every other `uip` command passes through to the real CLI.
- **Both coded tasks assert the adapter import** — the failure BYO is most exposed to, because `ByoValidator` genuinely lives in `uipath.platform.guardrails`, making a wrong-but-plausible import easy to land (that import type-checks and then silently no-ops). Both carry `disallowed_tools: ["Task"]` for the same reason: a dispatched subagent has no skill context and previously regressed exactly this.

## How has this been tested?

- **AST checkers validated against 14 controls** — 4 positive (inline and variable middleware spread; positional and keyword validator name) and 10 negative (built-in substituted for BYO, wrong name, wrong action, non-factory decoration target, wrong import module, unparseable source). Every negative fails at the intended assertion with an actionable message.
- **Two-model runs per the [Run Skill Smoke Tests](https://uipath.atlassian.net/wiki/spaces/AGOV/pages/90973078487/Run+Skill+Smoke+Tests) runbook — all six green.** Generated artifacts (`agent.json` / `graph.py`) inspected in each run rather than trusting the score alone.

  | Task | claude-sonnet-5 | gpt-5.6-terra |
  |---|---|---|
  | `byog_pinning` | SUCCESS 1.000 (86s) | SUCCESS 1.000 (132s) |
  | `byog_middleware` | SUCCESS 1.000 (181s) | SUCCESS 1.000 (196s) |
  | `byog_decorator` | SUCCESS 1.000 (362s) | SUCCESS 1.000 (150s) |

- **Stability — 3 repeats per task, 9/9 replicates `SUCCESS` at 1.000** (claude-sonnet-5). No flakes, no partial scores.

  | Task | Replicates passed |
  |---|---|
  | `byog_pinning` | 3/3 @ 1.000 |
  | `byog_middleware` | 3/3 @ 1.000 |
  | `byog_decorator` | 3/3 @ 1.000 |

- `scripts/check-cli-verbs.py` and `scripts/check-task-driver.py` clean on every touched file.
- **Note on CI coverage:** the PR smoke gate selects `^tags:.*\bsmoke\b`, so only `byog_pinning` runs on this PR. The two coded tasks are `e2e` (matching all eight of their siblings) and run on the daily/weekly cadence. Happy to retag them `smoke` if per-PR gating is preferred.

## Upstream SDK fixes (merged)

The coded docs describe a surface that needed two upstream corrections first; both are merged:

- **UiPath/uipath-python#1854** — adds the missing `ByoValidator` section (and stages-table entry) to the core guardrails page. Its absence is what made BYO look LangChain-only, and is why the signatures here were verified against SDK source rather than the rendered docs.
- **UiPath/uipath-langchain-python#1033** — drops a stale docstring line telling readers to pass an Integration Service connection id to `UiPathByoGuardrailMiddleware`; no such parameter exists (AL-510 removed it), and its own `Args` block already said the connection resolves server-side.

## Related

The **admin** side of BYOG — `uip guardrails byo-configurations` (list/create/update/delete), i.e. registering the configuration in the first place — is a separate PR, #2550, so this one doesn't need `uipath-platform` CODEOWNERS approval. Cross-references here go through the skill-level `/uipath:uipath-platform` link, so there's no merge-order dependency.

- [x] None

### DIFF
--- a/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails-recommend.md
+++ b/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails-recommend.md
@@ -54,6 +54,8 @@ uip agent guardrails list --output json
 
 Build a lookup of `{ validatorId: status }` from the `Data` array. You will use this to filter recommendations.
 
+> **`Validator` is not unique — key on `(Validator, IsByo)`, not `Validator` alone.** A tenant with a bring-your-own (BYOG) configuration for a validator has two entries sharing the same `Validator` name — one built-in, one BYO (`IsByo: true`, carrying `ByoValidatorName`/`ByoConnectionId`/`ByoConfigurationId`). Collapsing them loses the distinction and can point the discovery/wiring flow at the wrong entry. See [guardrails.md § BYO (bring-your-own) validators](guardrails.md#byo-bring-your-own-validators).
+
 > **Catalog vs. list — the key distinction:** The catalog lists all guardrails that exist on the platform (with rich metadata for reasoning). The guardrails list returns only those accessible to this tenant. Only recommend validators where `Status == "Available"` in the list.
 
 ### SDK Documentation (NEVER skipped — Python class names)
@@ -114,6 +116,8 @@ For **each entry** in the catalog (`guardrails[]` array from the cached JSON):
 5. If the validator is a candidate: use the catalog entry's `examples[].config` to determine the appropriate scope, stage, action, and parameters. Translate `validator_parameters` shape to the Python `Validator(...)` / `Middleware(...)` constructor arguments using the SDK docs from Step 0.
 
 Do **not** apply predetermined knowledge about which guardrail maps to which schema field. Let the catalog entry's authored fields drive every recommendation decision.
+
+> **Built-in vs. BYO — default to built-in.** When a matched validator has both a built-in entry and one or more `Available` BYO (`IsByo: true`) entries, recommend the standard SDK validator/middleware (built-in) by default, and mention a BYO alternative exists. Only wire in a specific BYO configuration when the user names it or asks for BYO explicitly — see [guardrails.md § BYO (bring-your-own) validators](guardrails.md#byo-bring-your-own-validators) for how that's actually referenced in code.
 
 ### Step 3 — De-duplicate Overlapping Validators
 
@@ -235,7 +239,7 @@ For each existing guardrail discovered in the Python file (Step 1 from Recommend
 
 ### Correctness Check
 
-From the SDK docs and the catalog, look up the validator class referenced in the code:
+From the SDK docs and the catalog, look up the validator class referenced in the code. **If the guardrails list has more than one entry sharing the referenced `Validator` name** (built-in plus BYOG), disambiguate by whether the code wires a BYO validator construct (`ByoValidator(...)` or `UiPathByoGuardrailMiddleware(...)`, identified by its `validator_name`) or the plain SDK validator/middleware class — match against the corresponding list entry's `IsByo` before reading `Parameters`/scopes for that entry.
 
 | Aspect | What to check |
 |--------|---------------|
@@ -295,3 +299,4 @@ python3 -c "import ast; ast.parse(open('graph.py').read())"
 13. **Class names and enum names come from the SDK docs** — never invent them. The SDK evolves; relying on memory produces stale code. For **import paths**, use the `langchain/guardrails/` page when the agent is LangChain (paths live in `uipath_langchain.guardrails`); for every other framework use the `core/guardrails/` page (paths live in `uipath.platform.guardrails`). See Rule 8.
 14. **Read [guardrails.md](guardrails.md) before writing any Python** — the middleware spread (`*`), decorator placement above `@tool` / factory, factory refactor, and import-source rules are specified there and cannot be safely inferred.
 15. **`EscalateAction` is the human-in-the-loop option only when the SDK docs expose it** — recommend it when the user wants a person to review/approve a flagged item rather than hard-block it. It requires a **deployed Action App** declared in `bindings.json` (`app_name` / `app_folder_path`) through the coded-agent bindings sync workflow; if the docs, app, or binding prerequisite is unavailable, fall back to Block/Log and say so — never silently drop the requested escalation. See Step 6 and [guardrails.md § Escalation action (HITL)](guardrails.md#escalation-action-human-in-the-loop).
+16. **`Validator` is not unique — disambiguate built-in vs. BYO by `IsByo` before matching on name.** A tenant can have both a built-in and one or more BYOG entries sharing the same `Validator` name. Key any lookup on `(Validator, IsByo)`, and default recommendations/wiring to the built-in SDK validator/middleware unless the user names a BYO configuration or asks for BYO. Never fabricate the BYO validator construct from memory — see [guardrails.md § BYO (bring-your-own) validators](guardrails.md#byo-bring-your-own-validators).
--- a/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails.md
+++ b/skills/uipath-agents/references/coded/capabilities/guardrails/guardrails.md
@@ -57,6 +57,87 @@ If the requested validator has `Status != "Available"` → tell the user and sto
 
 ---
 
+## BYO (bring-your-own) validators
+
+A validator can be fulfilled by a tenant-registered **external** provider (a "BYOG" configuration — e.g. Azure AI Content Safety, Databricks AI Guardrails) instead of UiPath's own built-in implementation. Registration is admin-side — Admin → AI Trust Layer → Guardrails Configurations, or `uip guardrails byo-configurations create` (see [uipath-platform § BYO Guardrail Configurations](/uipath:uipath-platform)); this section covers wiring an already-registered BYOG configuration into agent code.
+
+**Same rule as [Step 0](#step-0--fetch-official-documentation): confirm the constructor signature against the fetched SDK docs before writing code, and never invent arguments.** The two constructs below exist today; if a future fetch shows one missing or renamed, follow the fetched page and say so rather than writing what's here from memory.
+
+### Which construct exists where
+
+| Construct | Style | Ships in | Notes |
+|---|---|---|---|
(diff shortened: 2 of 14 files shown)


skills/uipath-platform/references/guardrails/byo-configurations.md
```diff
@@ -0,0 +1,148 @@
+# BYO Guardrail (BYOG) Configurations
+
+Manage tenant-registered bring-your-own guardrail configurations via `uip guardrails byo-configurations` (`list` / `create` / `update` / `delete`). A BYOG configuration registers an external validator provider (e.g. Databricks AI Guardrails, Azure AI Content Safety) against an Integration Service connection, so an agent's guardrail checks (PII detection, harmful content, etc.) run against that external provider instead of — or as a fallback pair with — UiPath's own built-in implementation. This is the guardrail analog of the BYO LLM configurations managed by [`uip llm-configuration byo-connections`](../llmgateway/byo-connections.md), and the CLI counterpart of Admin → AI Trust Layer → Guardrails Configurations.
```
apetraru-uipath: @valentinabojan do we need this skill? Maybe an user will ask an agent to configure a byog, but I think we will get too far with automation

skills/uipath-platform/references/guardrails/byo-configurations.md
```diff
@@ -0,0 +1,148 @@
...
+### Permission errors
+
+Any other non-2xx status (e.g. `403`) surfaces as `Message: "Failed to <verb> BYO guardrail configuration(s)"` with the backend's response body carried in `Instructions` (truncated to 1000 characters) — e.g. `"403 Forbidden: {\"title\":\"Forbidden\",\"detail\":\"User is not a member of the Administrators group\"}"`. This means the logged-in identity lacks the org-admin role, or is an application token (not a user token) — tell the user to re-authenticate with an admin user account rather than treating it as a feature-availability problem.
+
+---
+
+## How this feeds agent authoring
+
+Agents reference a BYOG configuration **by `ValidatorName` alone** — the platform resolves the connection server-side from the stored configuration; agents never pass a connection id. The agent-authoring side (discovery from an agent-design context, not admin) is `uip agent guardrails list --byo` in `uipath-agents` — see:
+- [Low-code guardrails](/uipath:uipath-agents) — `builtInValidator` guardrails authored against a specific BYOG configuration via `byoConfigurationId`.
```
valentinabojan: byoConfigurationId is not used. byoValidatorName is the one used by the lo coded agents

skills/uipath-platform/references/guardrails/byo-configurations.md
```diff
@@ -0,0 +1,148 @@
...
+| `ValidatorType` | The raw validator category this configuration implements (e.g. `pii_detection`, `harmful_content`) — matches the built-in `Validator` name for the same category. |
+| `FallbackOnUiPath` | Whether a failed call to the external provider falls back to UiPath's own built-in validator logic, or hard-fails the guardrail check. |
+| `Enabled` | Whether this configuration is switched on. A disabled configuration still appears in `uip agent guardrails list --byo` output but with `Status: Disabled`. |
+| `ValidConnection` | Whether the underlying Integration Service connection currently resolves and is healthy. `false` means the connection was disabled, revoked, or rotated since registration. |
+| `ConnectorKey` / `ConnectorName` / `ConnectionName` | Resolved metadata about the underlying IS connector/connection, for display. `null` when unresolved. |
+
+Empty `Data: []` is a valid result — it means the tenant has no BYOG configurations registered.
+
+## `create`
```
valentinabojan: We also need a cli command for test configuration similar to BYO llm config. 

Image

We enforce the same probe before saving a new connection via the UI.

skills/uipath-platform/SKILL.md
```diff
@@ -308,6 +311,7 @@ Every `uip` command accepts:
 - **[Integration Service](references/integration-service/integration-service.md)** — Connectors, connections, activities, resources
 - **[Data Fabric](references/data-fabric/data-fabric.md)** — Entity schemas, records CRUD, query filters and aggregates, choice sets, file attachments, CSV bulk import, folder scoping
 - **[LLM Gateway — BYO Connections](references/llmgateway/byo-connections.md)** — Register tenant-owned LLM keys against UiPath products
+- **[Guardrails — BYOG Configurations](references/guardrails/byo-configurations.md)** — Manage tenant-registered bring-your-own guardrail (BYOG) configurations and diagnose their underlying Integration Service connections
```
valentinabojan: Do we need to refer to this here as well? Isn't it enough to have it on line 177?

skills/uipath-platform/references/guardrails/byo-configurations.md
apetraru-uipath: Good catch — `byoConfigurationId` isn't a real field. Fixed to `ByoValidatorName` (matches the coded-agent `ByoValidator(<ValidatorName>)` pattern — both low-code and coded now reference a configuration the same way). This file has since moved to a dedicated PR (#2550) so it doesn't ride on this PR's approvals — the fix is there: https://github.com/UiPath/skills/pull/2550/commits/d4ace1ec4

skills/uipath-platform/references/guardrails/byo-configurations.md
apetraru-uipath: Confirmed against the CLI source — `guardrails byo-configurations create`/`update` run no server-side validate/probe, unlike `llm-configuration byo-connections` which always does `POST .../validate` before saving. That's a real CLI feature gap (a validate step, or a dedicated test verb), not something a docs PR can add on its own — worth a follow-up ticket against the guardrails-tool CLI. In the meantime I added a 'Known gap' section documenting this explicitly and pointing at `uip is connections ping <connection-id>` as the closest available partial check (confirms the connection is alive, not that it implements the guardrail contract). Moved with the rest of this file to #2550: https://github.com/UiPath/skills/pull/2550/commits/d4ace1ec4

skills/uipath-platform/SKILL.md
apetraru-uipath: This mirrors the existing convention for every other reference doc in this file, not something new for BYOG — LLM Gateway, Traces, Integration Service, and Data Fabric are each listed in both the Task Navigation table (quick 'I need to...' lookup) and the bottom References list (the file's own index). Line 177 and line 314 serve different browsing patterns; dropping just the BYOG entry from References would make it the one exception. Happy to revisit the whole file's convention separately if that duplication should go away everywhere, but keeping it consistent for now. This content is now on #2550.

valentinabojan (approved):

valentinabojan (approved):

andreibalas-uipath (approved):

PR #2520, opened 2026-08-07
fix(flow): document --parent for nesting CLI-owned nodes in a loop (canonical)

rockymadden:
## What

Documents `uip maestro flow node add --parent <containerId>` — the supported way to nest a **CLI-owned** node (managed HTTP `core.action.http.v2`, connector activities/triggers) inside a `core.logic.loop` body — and states it **once**, in the canonical Node-ownership section, with the leaves reduced to pointers.

## Why

Loop-body nodes require `parentId`. The skill only documented setting `parentId` **directly in the JSON**, which is valid for user-owned nodes (Script) but **forbidden for CLI-owned nodes** by the node-ownership rule (#9 — don't hand-edit connector/HTTP node JSON). That left **no documented path** to put a CLI-owned node in a loop.

The CLI already supports this via `node add --parent`, whose own help text reads:

> *"Id of the existing node that should contain this node. Set when authoring a loop body whose containment cannot be inferred from edges alone — typically the id of a `core.logic.loop` node."*

…but no reference doc mentioned the flag.

### Impact (root cause of a task failure)

In `skill-flow-multi-city-weather`, the agent correctly parented the in-loop **Script** node but added the **HTTP** node via `node add` (which lands it at top level, no `parentId`). The HTTP call therefore executed **outside** the loop, so `$vars.fetchCurrentTemperature1.output` was `null` on every iteration and the aggregated results contained no city data — failing the debug check. Critically, **`flow validate` still passed** (a missing `parentId` is structurally valid), so nothing surfaced the defect until runtime.

## Approach — one canonical home, thin leaves

The `--parent` mechanism is a corollary of node ownership (a CLI-owned node can't be hand-edited, so its *containment* must also come from the CLI). So the principle lives once in the canonical section; everything else points to it.

| Location | Role | Change |
|---|---|---|
| `references/author/CAPABILITY.md` § Node ownership | **canonical statement** | authoritative `--parent` bullet (covers *all* CLI-owned node types) + note that containment is the exception to "you may still Edit …" |
| `SKILL.md` rule #9 | universal one-liner | terse `--parent` clause |
| `references/author/references/editing-operations-cli.md` | CLI **reference** | `--parent` row in the `node add` flags table; one-line pointer to the canon |
| `plugins/loop/impl.md` | loop-specific | short note + example + pointer; one debug-table row covering **HTTP / connector** ("output `null` every iteration") |
| `plugins/loop/planning.md` | loop-specific | one-line distinction (user-owned JSON vs CLI-owned `--parent`) |

`plugins/http/impl.md` intentionally carries **no** `--parent` note: the rule is generic to every CLI-owned node, so special-casing HTTP would be arbitrary (connector activities/triggers have the identical concern). It's covered by rule #9, the canon, the CLI reference, and the loop debug row.

Docs only — no code or skill logic changed. Also fixes a wrong relative link depth (`../../` → `../../../`) to `CAPABILITY.md`.

## Verification

- `npm run skills:validate` → **OK** (no new flavor replacements)
- `scripts/check-cli-verbs.py` on all edited files → **0 High / 0 Medium** (`node add --parent` recognized against the `uip 1.200.0-dev.8116` catalog)
- All introduced relative links + heading anchors verified to resolve


jiyangzh (approved):
does the cli check if parent node is a loop?

rockymadden:
Dupe

PR #2524, opened 2026-08-08
test(uipath-maestro-case): add outcome-based e2e graded on third-party effects

song-zhao-25:
## What

Adds `tests/tasks/uipath-maestro-case/e2e_outcome_external/` — an e2e task that grades whether a case **achieved its business objective in the systems it touches**, not whether it was built a particular way.

The agent gets an approved SDD for a two-stage expense-approval case. Two things must be true afterwards:

- the approver has a notification **email** in their mailbox (Microsoft Outlook 365), and
- an **audit ticket** exists in Jira,

both carrying a per-run reference token. The harness executes the case, then queries Microsoft Graph and Atlassian to confirm both effects landed.

## Why

Acceptance for this work was defined as: *"it achieved the business objectives as measured in the third party systems. Not if it validated, not if it debugged, but that it accomplished what the user was after — the coding agents can do this a number of different ways."*

So the weighting is deliberate:

| Criterion | Weight | Gating |
|---|---|---|
| Email arrived (measured via Microsoft Graph) | 4.0 | yes |
| Jira issue exists (measured via Atlassian REST) | 4.0 | yes |
| No hand-made effects (direct mutating connector calls) | 2.0 | yes |
| Case reached a terminal completed status | 2.0 | yes |
| Agent did not run `case debug` itself | 1.0 | yes |
| Skill triggered | 1.0 | yes |
| `uip maestro case validate` passes | 1.0 | **advisory** (`pass_threshold: 0`) |

Caseplan topology and stage/task shape are **not** graded — any design that achieves the objective passes. Completing is worth less than the effects because a case can complete while notifying nobody.

This is additive: it does not touch `e2e_expense_runnable` or any existing task.

## No new CI secrets

The graders read each target system back **through the same Integration Service connection the case uses** (`uip is resources run list …`). The HTTP call hits the vendor API for real; auth is the tenant's existing connection.

## Verification

Local `coder-eval` runs against alpha/codereval:

| Run | Result |
|---|---|
| `2026-08-07_16-05-28` | FAILURE, 4/7, 0.333 |
| `2026-08-07_16-42-17` | SUCCESS, 7/7, score 1.000, 816s (predates the review fixes) |
| `2026-08-07_19-09-09` | FAILURE, 4/7, 0.333 — platform stall, see below |
| `2026-08-07_20-14-15` | **SUCCESS, 7/7, score 1.000, 1049s — current HEAD** |

Grader evidence from the passing run:

```
OK: case executed to finalStatus=Completed
  email: 'Expense approval required — OBT-D1592417'  received 2026-08-08T03:30:57Z
  issue: SJP-14 'Expense audit — OBT-D1592417'       created 2026-08-08T06:31:09+0300
OK: the approver was notified — message present in the mailbox.
OK: the audit record was filed — issue present in Jira.
cleanup_outcomes: delete email [Inbox] … -> Success
cleanup_outcomes: delete email [SentItems] … -> Success
cleanup_outcomes: close Jira SJP-14 -> Success
```

**Platform caveat, stated plainly.** Case execution on the eval tenant was intermittently
unreliable while developing this: 3 of 7 runs stalled after starting (instance runs its
scaffolding elements, never enters Stage 1, raises no incident), then the *identical* build
completed in 56 s minutes later. The stalls are not caused by the build, the skill, or this
test — the same caseplan both stalls and passes. When it happens the test reports it
accurately, which is the intended behavior, but expect occasional red runs that are
platform-caused until that is addressed.

The first run's failure was informative rather than incidental — see below.

## Design notes (each one earned by a failure)

1. **Per-run token.** `uip maestro case debug` has no `--input` flag, so `pre_run` renders `sdd-template.md` into the sandbox with a fresh token baked in and records it in `seed.json`. Both target sandboxes are shared and noisy (the mailbox took unrelated mail mid-run), so probes match this run's token rather than "the newest record" — which also keeps parallel replicates independent.
2. **Timeout tolerance.** `case debug` abandons polling at a fixed 600 s while the case keeps running. In run 1 the effects landed **~7 min after the CLI gave up**, which is exactly why it scored 0.333. The checker now follows the instance to a terminal state instead of failing on the timeout, and probes poll ~10 min because Graph's subject filter lags delivery.
3. **Order-independent graders.** Criterion order is not guaranteed, and three criteria need the case to have run, so `ensure_debug_ran()` executes it once behind an exclusive lock and republishes the payload. With `-j > 1` it refuses to guess when several debug instances fall in the same window, rather than risk attaching to a sibling task's success.
4. **Effects asserted as `>= 1`, not exactly once.** A stray extra execution otherwise fails a case that met the objective; duplicates are reported instead.
5. **Anti-gaming.** Both effects are reachable straight from the shell, which would bank the outcome criteria with nothing built — hence the two `command_not_executed` guards. Read-only `is resources run list` stays allowed because the skill uses it to resolve reference fields.
6. **`max_iterations: 1`.** A retry re-executes the case and duplicates external records; it does not recover the run.

## Reviewer notes / known trade-offs

- **Tenant preconditions**, documented in the task header: both sandbox connections healthy (Outlook `dd657127…`, Jira `f5273a4d…`) plus Jira project `SJP` with issue type `10506`. If one disappears the graders abort with `test precondition failed: … ENVIRONMENT gap` rather than looking like a skill regression.
- **Cleanup, precisely:** mailbox messages are deleted from BOTH Inbox and Sent Items (the design sets `saveToSentItems: true`, so each run leaves two copies). Jira issues cannot be *deleted* — the sandbox account gets 403 in `SJP` — so `post_run` transitions them to **Done** instead. Rows persist one per run, but the project's open-issue list stays clean. Worth requesting delete permission from the sandbox owner if the row count ever matters.
- **Not in the PR gate:** `smoke-skills.yml` runs `--tags smoke`; this is `e2e`, so it runs under `make e2e` (sequential by default).
- **A flake rate is still outstanding.** HEAD has one green run; a multi-replicate study (and a model split) was deferred because the tenant's stall rate would have dominated the measurement rather than the test's own stability.
- **Scope honesty:** this grades two external effects on a deliberately small two-stage case. It is a genuine instance of the outcome-based bar, not full coverage of it; measuring every transport of a larger scenario externally would be a scenario redesign.


rockymadden (approved):
Yeah, this is the idea! Awesome!

nikhil-maryala (approved):

roberts-cliff (approved):

charlesliu9 (approved):

PR #2544, opened 2026-08-10
fix(process-mining): grade the real add-table verb, and make the verb linter see leaf verbs

Cellcote:
## Why

The process-mining eval set looked flaky across the two most recent nightly runs
([2026-08-05](https://coder-evalboard.uipath-dev.com/runs/2026-08-05_04-15-14),
[2026-08-10](https://coder-evalboard.uipath-dev.com/runs/2026-08-10_04-43-31)).
It wasn't flake — it was a grader bug that the CLI-verb linter could not see.

Only 2 of 6 tasks moved between the runs, each by exactly one criterion:

| Task | 2026-08-05 | 2026-08-10 |
|---|---|---|
| `skill-pm-add-table-queryable` | 1.0 | **0.778 FAILURE** |
| `skill-pm-custom-app-pipeline-e2e` | **0.833 FAILURE** | 1.0 |

(The runs are not otherwise comparable — different model, different `uip` build,
and the task set itself changed between them. Details at the bottom.)

## 1. `skill-pm-add-table-queryable` graded a verb that doesn't exist

The 3.0-weight criterion required `uip pm apps model add-table`. The catalog has
`uip pm apps data-model add-table`; `apps model` is the *semantic* model
(`fields` / `get` / `update`) and has no `add-table` child. The skill teaches the
correct verb in both `SKILL.md` (Rule 1) and `references/data-model.md`.

So the eval rewarded a hallucinated verb and penalised agents that followed the
skill:

- **08-05** — the agent typed `apps model add-table`, it errored, it retried with
  `apps data-model add-table`. The regex matched the *failed* attempt → passed.
- **08-10** — the agent went straight to the correct verb → matched nothing →
  0.778, FAILURE.

The green run was the wrong result. Fixed to `apps\s+data-model\s+add-table`.

## 2. Why `check-cli-verbs.py` reported the set clean

Two defects, each now covered by a regression test:

- **`\b` discarded the leaf verb.** On `AT_BOUNDARY` the walker bailed *and*
  called `_trim_to_word_boundary`, dropping the token it had just accumulated.
  But `\b` after a word character *confirms* the token ended — the opposite of a
  partial match. Since `uip <group> <leaf>\b` is this repo's standard pattern
  shape, the checker was only ever validating the command **group**.
- **Longest-prefix matching swallowed bogus subcommands.** `best_match` then fell
  back to `pm apps model`, a real group, and reported reachable.

Third gap, also fixed: `iter_command_patterns` only yielded `command_executed`,
so **negative criteria were never verb-checked**. A negative whose verb does not
exist can never fire, so it awards its weight on every run. `command_executed`
with `max_count: 0` is the other negative idiom here and gets the same handling.

Negatives report **Medium**, not High — they inflate scores rather than breaking
the run, and the guard is sometimes deliberate.

False positives are guarded three ways, all tested: positional arguments after a
leaf verb (`is resources run list <connector> <activity>`), partial kebab tokens
(`create-` from a pattern matching `create-raw` / `create-resource`), and
anything under an `unwalkable_groups` entry.

Also pinned `encoding="utf-8"` on the file reads — the script crashed on Windows
under cp1252.

## 3. Tag hygiene

Five of six tasks were missing required tags or using values outside the closed
vocabularies in `tests/README.md` §Tag Taxonomy (`mode:fix`, `lifecycle:extend`,
`lifecycle:maintain` are not valid; `tier` was absent on four). Those tasks were
invisible to `make` targets, coverage reports and evalboard drilldown.

| Task | Before | After |
|---|---|---|
| `add_table_queryable` | `mode:build, lifecycle:extend` | `integration, mode:build, lifecycle:setup` |
| `app_types_discovery_smoke` | `smoke, mode:diagnose, discover` | `smoke, mode:diagnose, lifecycle:discover` |
| `mapping_fix_in_place` | `mode:fix, lifecycle:maintain` | `integration, mode:diagnose, lifecycle:setup` |
| `query_group_by_sugar` | `mode:diagnose` | `smoke, mode:diagnose, lifecycle:discover` |
| `transform_fix_apply_not_reingest` | `mode:fix, lifecycle:maintain` | `integration, mode:diagnose, lifecycle:setup` |

The two repair tasks follow the precedent set by
`uipath-api-workflow/diagnose/runtime_fix` (`mode:diagnose` for "here is an
error, fix it"), with `lifecycle:setup` because they mutate tenant state. The
bare `discover` tag duplicating `lifecycle:discover` was dropped.
`custom_app_pipeline_e2e` was already correct.

## Verification

- `pytest tests/scripts/test_verb_checkers.py` — **21 passed** (14 existing + 7 new).
- `check-cli-verbs.py tests/tasks/uipath-process-mining/*.yaml` — 0 High, 0 Medium.
- Repo-wide across all task YAMLs — **0 High**, 2 Medium, 124 Info.
- No CI gate changes behaviour: nothing in `.github/` invokes this script;
  `/audit-verbs` uses `--report` (exit 0 on Medium) and `/lint-task` merges
  `--json` findings.

The 2 Medium findings are both in
`tests/tasks/uipath-governance/aops-policy/deployment_configure_smoke.yaml` —
negatives against `deployment user add` / `add-user`, which don't exist (the real
verb is `configure`, which auto-registers). Plausibly intentional guards; left
alone as out of scope, but worth a look from that skill's owner.

**Passing-run claim:** the eval tasks themselves were not re-run against a tenant
for this PR — the change is to grader regexes and tags, verified statically
against the `uip` catalog (1.200.0-dev.8173) and the linter's own test suite.
Re-running `skill-pm-add-table-queryable` should be part of the next nightly.

## Not addressed here

Deliberately out of scope, from the same investigation:

- **The runs aren't comparable.** 08-05 ran `gemini-3.5-flash` on `uip` 8093;
  08-10 ran `gpt-5.6-terra` on 8173, and `mapping_fix_in_place` gained two
  criteria in between. With **one replicate** per task, a single criterion flips
  the whole task. Pin the model or run ≥3 replicates before calling anything flaky.
- **`custom_app_pipeline_e2e`'s "inspect the transformations" criterion** is
  incidental to the task goal and cost the 08-05 run a FAILURE for no behavioural
  reason — drop it or make it advisory (`pass_threshold: 0`).
- **Coverage:** 20/44 `uip pm` verbs are referenced. Critical Rule 10
  (`apps model fields`, the DNA-46960 duration lockout) has **zero** tasks despite
  being called out in the frontmatter, and the whole `query` family beyond `run`
  is untested. Separate PR.

### DIFF
--- a/scripts/check-cli-verbs.py
+++ b/scripts/check-cli-verbs.py
@@ -3,16 +3,24 @@
 Verify that `uip` verb literals referenced in coder-eval task YAMLs actually
 exist in the CLI catalog.
 
-For every `command_executed` criterion in each task YAML, extract literal verb
-tokens that follow the `uip` prefix in `command_pattern`, enumerate the
-alternation paths, and check each path against `assets/uip-catalog-snapshot.json`
-and `.claude/rules/cli-renames.md`.
+For every `command_executed` / `command_not_executed` criterion in each task
+YAML, extract literal verb tokens that follow the `uip` prefix in
+`command_pattern`, enumerate the alternation paths, and check each path against
+`assets/uip-catalog-snapshot.json` and `.claude/rules/cli-renames.md`.
+
+A path is matched leaf-first. Falling back to a shorter prefix is allowed only
+when the leftover token cannot be a subcommand — a trailing flag, argument, or
+partial token. A bare word sitting where a subcommand would go, under a group
+whose children the catalog knows, is reported rather than passed off as the
+parent group.
 
 Findings:
   - High   — pattern does not match any verb in the catalog. The success
              criterion can never fire; the task scores zero on a passing run.
-  - Medium — pattern matches only retired verbs listed in cli-renames.md.
-             Suggest the canonical replacement.
+  - Medium — pattern matches only retired verbs listed in cli-renames.md
+             (suggest the canonical replacement), or a NEGATIVE criterion
+             targets a verb that does not exist, so it can never fire and
+             awards its weight unconditionally.
   - Info   — pattern is too dynamic to analyse (contains `.`, `[`, `\\w`, etc).
              Skipped — no claim made about it.
 
@@ -58,8 +66,20 @@ def load_catalog():
     if not CATALOG_PATH.exists():
         sys.exit(f"Catalog not found at {CATALOG_PATH}. "
                  "Run scripts/build-uip-catalog.py first.")
-    data = json.loads(CATALOG_PATH.read_text())
+    data = json.loads(CATALOG_PATH.read_text(encoding="utf-8"))
     return set(data["verbs"]), data.get("cli_version", "unknown")
+
+
+def load_unwalkable():
+    """Groups the catalog builder could not enumerate.
+
+    Their children are unknown, so we must never claim a verb beneath one is
+    missing — absence from the catalog carries no information there.
+    """
+    if not CATALOG_PATH.exists():
+        return set()
+    data = json.loads(CATALOG_PATH.read_text(encoding="utf-8"))
+    return set(data.get("unwalkable_groups") or ())
 
 
 def load_renames():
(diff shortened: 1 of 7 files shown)


Cellcote:
Superseded by the split requested in review:

- #2545 — the process-mining eval-set fix (add-table verb + required tags)
- #2546 — the `check-cli-verbs.py` fix that catches this class of bug

Same commit content, separated so the grader fix and the tooling fix can be
reviewed and reverted independently. Merge #2545 first — #2546 reports the
process-mining task as High until it lands.

PR #2545, opened 2026-08-10
fix(process-mining): grade the real add-table verb and fix required tags

Cellcote:
Split out of #2544 (part 1 of 2). The linter change that would have *caught* this
is in the companion PR; this one is the eval-set fix on its own.

## Why

The process-mining eval set looked flaky across the two most recent nightly runs
([2026-08-05](https://coder-evalboard.uipath-dev.com/runs/2026-08-05_04-15-14),
[2026-08-10](https://coder-evalboard.uipath-dev.com/runs/2026-08-10_04-43-31)).
It wasn't flake. Only 2 of 6 tasks moved, each by exactly one criterion:

| Task | 2026-08-05 | 2026-08-10 |
|---|---|---|
| `skill-pm-add-table-queryable` | 1.0 | **0.778 FAILURE** |
| `skill-pm-custom-app-pipeline-e2e` | **0.833 FAILURE** | 1.0 |

## 1. `skill-pm-add-table-queryable` graded a verb that doesn't exist

The 3.0-weight criterion required `uip pm apps model add-table`. The catalog has
`uip pm apps data-model add-table`; `apps model` is the *semantic* model
(`fields` / `get` / `update`) and has no `add-table` child. The skill teaches the
correct verb in both `SKILL.md` (Rule 1) and `references/data-model.md`.

So the eval rewarded a hallucinated verb and penalised agents that followed the
skill:

- **08-05** — the agent typed `apps model add-table`, it errored, it retried with
  `apps data-model add-table`. The regex matched the *failed* attempt → passed.
- **08-10** — the agent went straight to the correct verb → matched nothing →
  0.778, FAILURE.

The green run was the wrong result. Fixed to `apps\s+data-model\s+add-table`,
and the same correction applied to the task description.

## 2. Tag hygiene

Five of six tasks were missing required tags or using values outside the closed
vocabularies in `tests/README.md` §Tag Taxonomy (`mode:fix`, `lifecycle:extend`,
`lifecycle:maintain` are not valid; `tier` was absent on four). Those tasks were
invisible to `make` targets, coverage reports and evalboard drilldown.

| Task | Before | After |
|---|---|---|
| `add_table_queryable` | `mode:build, lifecycle:extend` | `integration, mode:build, lifecycle:setup` |
| `app_types_discovery_smoke` | `smoke, mode:diagnose, discover` | `smoke, mode:diagnose, lifecycle:discover` |
[...]

PR #2546, opened 2026-08-10
fix(tests): make the CLI-verb linter see leaf verbs and negative criteria

Cellcote:
Split out of #2544 (part 2 of 2). Companion to #2545, which fixes the task this
change catches. **Merge #2545 first** — otherwise this PR's first repo-wide run
reports a High on `uipath-process-mining`, which is the real bug, not a
regression from this change.

## Why

A process-mining task graded on `uip pm apps model add-table` — a verb that does
not exist (the CLI has `uip pm apps data-model add-table`; `apps model` is the
semantic model, with `fields` / `get` / `update` and no `add-table` child). It
shipped, and it took a nightly score from 1.0 to 0.778, because
`check-cli-verbs.py` reported the whole set clean. Two defects combined to hide it.

### 1. `\b` discarded the leaf verb

On `AT_BOUNDARY` the walker bailed *and* called `_trim_to_word_boundary`,
dropping the token it had just accumulated. But `\b` after a word character
*confirms* the token ended — it is the opposite of a partial match.

Since `uip <group> <leaf>\b` is this repo's standard pattern shape, the checker
was only ever validating the command **group**. `uip\s+pm\s+apps\s+model\s+add-table\b.*`
extracted as just `pm apps model`.

`\B` keeps the old behaviour: it asserts the token continues, so what was
accumulated really is mid-token.

### 2. Longest-prefix matching swallowed bogus subcommands

`best_match` then fell back to `pm apps model`, a real catalog group, and
reported reachable. Shortening is now refused when the leftover token sits in
subcommand position under a group whose children the catalog knows.

### 3. Negative criteria were never verb-checked

`iter_command_patterns` only yielded `command_executed`. A negative whose verb
does not exist can never fire, so it awards its weight on **every** run — a dead
assertion that inflates scores. `command_executed` with `max_count: 0` is the
other negative idiom in this repo and gets the same handling.

Negatives report **Medium**, not High: they inflate scores rather than breaking
the run, and the guard is sometimes deliberate (blocking a verb an agent might
invent).

## Guarding against false positives

The prefix-shortening rule is deliberately narrow. Each of these is covered by a
test:

- **Positional arguments after a leaf verb** — `is resources run list <connector>
  <activity>`. `is resources run list` has no children, so the trailing tokens are
  arguments, not subcommands.
- **Partial kebab tokens** — `create-`, left by a pattern matching `create-raw` /
  `create-resource`. A trailing `-` means the walker stopped mid-token, so we
  cannot claim the verb is missing.
- **`unwalkable_groups`** — the catalog builder could not enumerate them, so
  absence from the catalog carries no information there.

## Also

Pinned `encoding="utf-8"` on the file reads. The script crashed with
`UnicodeDecodeError` on Windows under cp1252 when scanning task YAMLs containing
em-dashes.

## Verification

- `pytest tests/scripts/test_verb_checkers.py` — **21 passed** (14 existing + 7 new).
  The pre-existing `test_extract_verb_paths_treats_word_boundary_as_dynamic` still
  passes: a `\b` reached before any complete token is still treated as dynamic.
- Repo-wide across all task YAMLs, **on the current unfixed tree**: 1 High, 2
  Medium, 124 Info. The High is exactly the process-mining bug this was written to
  catch; with #2545 merged it goes to 0.
- No CI gate changes behaviour — nothing in `.github/` invokes this script.
  `/audit-verbs` uses `--report` (exit 0 on Medium, per the contract in
  `audit-verbs.md`) and `/lint-task` merges `--json` findings.

The 2 Medium findings are both in
`tests/tasks/uipath-governance/aops-policy/deployment_configure_smoke.yaml` —
negatives against `deployment user add` / `add-user`, which don't exist (the real
verb is `configure`, which auto-registers). Plausibly intentional guards; left
alone as out of scope, but worth a look from that skill's owner since they
currently award their weight unconditionally.

### DIFF
--- a/scripts/check-cli-verbs.py
+++ b/scripts/check-cli-verbs.py
@@ -3,16 +3,24 @@
 Verify that `uip` verb literals referenced in coder-eval task YAMLs actually
 exist in the CLI catalog.
 
-For every `command_executed` criterion in each task YAML, extract literal verb
-tokens that follow the `uip` prefix in `command_pattern`, enumerate the
-alternation paths, and check each path against `assets/uip-catalog-snapshot.json`
-and `.claude/rules/cli-renames.md`.
+For every `command_executed` / `command_not_executed` criterion in each task
+YAML, extract literal verb tokens that follow the `uip` prefix in
+`command_pattern`, enumerate the alternation paths, and check each path against
+`assets/uip-catalog-snapshot.json` and `.claude/rules/cli-renames.md`.
+
+A path is matched leaf-first. Falling back to a shorter prefix is allowed only
+when the leftover token cannot be a subcommand — a trailing flag, argument, or
+partial token. A bare word sitting where a subcommand would go, under a group
+whose children the catalog knows, is reported rather than passed off as the
+parent group.
 
 Findings:
   - High   — pattern does not match any verb in the catalog. The success
              criterion can never fire; the task scores zero on a passing run.
-  - Medium — pattern matches only retired verbs listed in cli-renames.md.
-             Suggest the canonical replacement.
+  - Medium — pattern matches only retired verbs listed in cli-renames.md
+             (suggest the canonical replacement), or a NEGATIVE criterion
+             targets a verb that does not exist, so it can never fire and
+             awards its weight unconditionally.
   - Info   — pattern is too dynamic to analyse (contains `.`, `[`, `\\w`, etc).
              Skipped — no claim made about it.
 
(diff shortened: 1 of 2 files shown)


Cellcote:
## Merge-order caveat is now resolved

#2545 has merged, so the ordering note at the top of this PR no longer applies — this can go in whenever.

Re-ran the patched checker against **current `main`** (which has since refreshed the catalog to `1.201.0-dev.8272`, 1465 verbs), over all task YAMLs:

```
High:   0
Medium: 4
Info:   142
(1130 files linted; 55 skipped locally due to Windows long-path limits)
```

**0 High** — the `uipath-process-mining` finding this change was written to catch is gone now that #2545 fixed the verb. That is the before/after in one line: on the pre-#2545 tree this checker reported 1 High, and it was the real bug.

The 4 Medium are all the dead-negative class this PR adds detection for — negatives whose verb does not exist, so they can never fire and award their weight unconditionally:

| File | Pattern |
|---|---|
| `uipath-governance/aops-policy/deployment_configure_smoke.yaml` | `deployment user (add\|add-user)` |
| `uipath-governance/aops-policy/deployment_configure_smoke.yaml` | `deployment group (add\|add-group)` |
| `uipath-maestro-case/golden_rebuild/cm_golden_expense/cm_golden_expense.yaml` | `maestro case (cases\|stages\|tasks\|…) (add\|update\|remove)` |
| `uipath-platform/guardrails/create_configuration_smoke.yaml` | `guardrails byo-configurations get` |

The last two are new since I opened this PR — surfaced by the catalog refresh, not by any change here. All four are advisory (Medium does not fail the `--report` path per the `audit-verbs.md` contract) and all belong to other skills, so I have left them alone. Worth each owner confirming the guard is intentional.

Note this branch is behind `main`; happy to rebase if you'd prefer it current before merge.

PR #2547, opened 2026-08-10
test(process-mining): coverage for model editing, publish, query info, source templates; fix the published-stage claim

Cellcote:
Part 3 of the process-mining eval-set work, after #2545 (grader fix) and #2546
(linter fix). Independent of both — merges in any order.

## Why

Auditing the process-mining eval set while chasing the flaky nightly turned up
two things beyond the grader bug:

- **20 of 44 `uip pm` verbs** were referenced by any criterion.
- **Three Critical Rules had no task at all**, including one that is named in
  the skill's own frontmatter.

| Rule | Before |
|---|---|
| 6 — `--wait` on async commands | never asserted anywhere |
| 8 — dev subset → publish | no task; `apps publish` untested |
| 10 — `apps model fields` / DNA-46960 | **zero coverage** |

Rule 10 was the worst: `references/model-editing.md` is a full reference, the
numeric→duration lockout is called out in both `description` and `when_to_use`,
and nothing exercised it.

### `model_fields_data_kind` — integration
Rule 10 and the DNA-46960 footgun. A throughput field is `numeric`; a metric
compares it to a numeric constant; the user wants `duration`.

Flipping the kind alone *is the bug* — that is exactly how the customer app
reached the state where no dashboard opens ("Must be duration, not numeric, for
the 'lt' input"). So the task grades `fields set --kind duration` **and** a
follow-up `--expression` edit, plus the pushed expression's constant via
`file_contains` on a fixed path (the pattern `add_table_queryable` already uses
for `table.json`). Negatives cover kinds outside the settable enum and the
structural `id` / `ref` kinds.

### `data_model_inspect_smoke` — smoke
`apps data-model get`. Foreign keys live only in the **structural** model, not
the semantic one — the exact namespace confusion that produced the
`apps model add-table` grader bug fixed in #2545. Pins the read half of the
add-table workflow; `add_table_queryable` pins the write. Negatives guard
against mutating on a read-only request.

### `query_info_percentile` — smoke
Rule 7's *mechanism* rather than its sugar. `query_group_by_sugar` covers the
path where `--group-by/--metric` resolves ids for the agent; `percentile` has no
such shorthand — it takes `--field <id>`, so discovery via `query info` cannot be
skipped without hitting `UserError_FieldNotFound`. Three criteria: skill
activation, `query info`, and `percentile` against a `--field` id.

### `publish_promote_to_dashboards` — integration
Rule 8 and the publish loop. `apps publish` promotes the **definition**, not the
rows: the result carries `IngestionNeeded: true` and published data is stale
until a re-ingest. Also the first task anywhere to assert `--wait` (Rule 6).

### `source_template_selection` — integration
Rule 2's untested half. `custom_app_pipeline_e2e` covers a flat log choosing
`uipath.custom`; nothing covered the other branch. A real multi-table SAP P2P
extract must choose `uipath.p2p.sap` — reaching for `uipath.custom` throws away
the template's entire extraction and event-log construction layer. Also pins
where a source template's expected **input** tables come from (the generated
transformations / `sources.yml`, not the `model` object).

## Review changes

Three criteria removed after review, each because the assertion itself was wrong
or unwanted — not merely over-broad:

- **`query --stage published` is NOT a dead end.** After a successful ingestion
  *on the published stage* it works; it is optional, not forbidden. Removed from
  `query_info_percentile` and `publish_promote_to_dashboards`. The skill asserted
  the same false claim in three places — **also corrected here** (see below).
- **The `Cases.sql` negative** on `source_template_selection` assumed a
  source-system template never needs its event-log construction touched. Dropped
  rather than encoding that as a graded rule.
- **The `--values` 0..1 criterion** on `query_info_percentile`.

Matching description text was removed with each, so no description outlives the
criterion it described — that drift is what made the raw-body negative in
`query_group_by_sugar` misfire (#2650).

Weights after: `query_info_percentile` 3 criteria / 7.0,
`publish_promote_to_dashboards` 5 / 10.0, `source_template_selection` 6 / 12.5.
All re-run and green (see comments).

## Skill correction: published-stage querying

The removed criterion was not invented for the eval — it encoded what the skill
teaches. So the skill is corrected in the same PR, since shipping the task change
alone would leave the false guidance in place.

`references/lifecycle-and-rbac.md` stated it most strongly:

> **CLI caveat:** `uip pm query --stage published` is currently **not reachable** —
> `/query/{id}/published` needs a completed ingestion on that stage and no `uip pm`
> path produces one … So publishing promotes the app to the **dashboards**, but
> **CLI-driven `query` stays on `dev`**.

`ingestions create` takes `--stage`:

```
$ uip pm ingestions create --help
  --stage <stage>    App stage to ingest into
```

So `uip pm ingestions create <app> --stage published --wait` **is** the "path that
produces one", and once it completes `query --stage published` works. The
half-truth is the middle clause of the old text: a plain `ingestions create --wait`
without `--stage` ingests into `dev`, which naturally leaves published querying at
`UserError_InvalidOrNoIngestion`. That got read as "unreachable" rather than
"ingested into the wrong stage".

The skill was already self-contradictory — six lines below the caveat the promote
loop ended with "→ analyse on `--stage published`", and `references/uip-pm-cli.md`
said "Develop on `dev` (a subset), `publish`, then read `--stage published`".

| File | Change |
|---|---|
| `SKILL.md` (Rule 8) | published querying works after a published-stage ingestion, and is optional rather than a dead end |
| `references/querying.md` | `--stage published` needs a completed ingestion on that stage; it 400s only until then |
| `references/lifecycle-and-rbac.md` | caveat rewritten as guidance; `--stage published` named on the ingestion in the promote loop |

**Deliberately unchanged** — everything about **writes**, which remains true
(`published` is read-only): Rule 5, `references/pre-flight.md`, and the stage table
in `references/uip-pm-cli.md`. Correspondingly the
`data-mapping update --stage published` negative in `mapping_fix_in_place.yaml`
stays: it asserts the write rule, not the query rule.

Checks: no `skill-flavor` markers in the three files and no
`skill-flavors/*/uipath-process-mining` overrides, so nothing to mirror.
Frontmatter untouched (body-only), so no `activation-gate.yml` recall-eval trip.
`scripts/check-skill-verbs.py` — 0 new findings.

**Not verified end-to-end against a live tenant** — I do not have one wired up. The
correction rests on the CLI surface plus @Cellcote's report that querying published
works after a published-stage ingestion. The original claim carried a "verified
against a live tenant" note and was still wrong, so a confirmation on a real app is
worth having before merge.

## Coverage delta

**20/44 → 31/44 verbs.** Newly exercised: `app-types get`,
`apps data-model get`, `apps model fields list`, `apps model fields set`,
`apps publish`, `query info`, `query percentile`.

Still uncovered, deliberately — these need a live tenant with real data to grade
anything meaningful, so command-construction tasks would only assert that a verb
was typed: `query rca` / `insights` / `layout` / `details`, `transformations run`
/ `status` / `logs`, `ingestions logs`, `apps model get` / `update`,
`apps model fields remove`.

## Conventions

- Tags: `skill` + `tier` + `mode:*` + `lifecycle:*`, closed vocabularies only.
- No `agent:` or `sandbox:` blocks — inherited from the experiment. No
  `env_packages`.
- Command-construction, matching the existing process-mining tasks: no live
  tenant, graded on which commands the agent chooses, with the same "commands
  will fail with auth errors, run each exactly once" prompt caveat.
- Graded on side effects only — `command_executed`, `command_not_executed`,
  `file_contains`, `skill_triggered`. No criterion gates on `--output json`.

## Verification

- All 5 parse; task ids unique across the 11 tasks in the directory.
- Tag check against the closed vocabularies in `tests/README.md` — all clean.
- CLI-verb reachability with the patched checker from #2546: **0 High, 0 Medium**
  on the new tasks. (Running it over the whole directory reports 1 High on
  `add_table_queryable.yaml` — that is the bug #2545 fixes, not something
  introduced here.)
- Weights follow the guidance in `.claude/rules/test-writing.md`: 3.0 primary
  validation, 2.0–2.5 core command/artifact content, 1.0–1.5 supporting.

**Passing-run claim (updated):** all five tasks have now been run and pass.

**CI, this PR's own smoke job** — Docker, `@uipath/cli@dev`, Bedrock. Covers the
two smoke-tier tasks:

```
skill-pm-data-model-inspect-smoke   SUCCESS  1.000  (4/4 criteria)  25.6s
skill-pm-query-info-percentile      SUCCESS  1.000  (5/5 criteria)  98.2s
Results: 2/2 succeeded
```

This is the faithful environment — `data_model_inspect_smoke` grades
`apps data-model get`, which only exists on the `@dev` build.

**Local**, `coder-eval` / `experiments/default.yaml` / tempdir /
`claude-sonnet-5`, on this branch merged with #2545 — covers all five including
the three integration-tier tasks the smoke job does not run:

```
skill-pm-model-fields-data-kind              SUCCESS  1.000  turns=19
skill-pm-data-model-inspect-smoke            SUCCESS  1.000  turns=9
skill-pm-query-info-percentile               SUCCESS  1.000  turns=8
skill-pm-publish-promote-to-dashboards       SUCCESS  1.000  turns=11
skill-pm-source-template-selection           SUCCESS  1.000  turns=34
```

Whole process-mining set in that run: 10/11 SUCCESS. The one failure was
`skill-pm-query-group-by-sugar`, a pre-existing task untouched here — since fixed
in #2650 (merged), now 4/4 replicates at 1.000.

I expected `model_fields_data_kind` to need retuning; it passed clean on first
run, including the `file_contains` assertion on the pushed expression.

Remaining caveat: the three integration-tier tasks are verified locally only,
against public npm `latest` (1.198.0) rather than `@dev`. Two of them
(`model_fields_data_kind`, `publish_promote_to_dashboards`) grade verbs absent
from 1.198.0 and scored 1.000 anyway — the agent typed the skill-taught command
once despite the unknown-command error, which is what the command-construction
design intends — so this is the pessimistic case.


Cellcote:
## Passing-run claim — all five tasks green on first run

Ran locally with `coder-eval` (`experiments/default.yaml`, tempdir sandbox, `claude-sonnet-5`, 1 replicate), on this branch merged with #2545:

```
skill-pm-model-fields-data-kind              SUCCESS  1.000  turns=19
skill-pm-data-model-inspect-smoke            SUCCESS  1.000  turns=9
skill-pm-query-info-percentile               SUCCESS  1.000  turns=8
skill-pm-publish-promote-to-dashboards       SUCCESS  1.000  turns=11
skill-pm-source-template-selection           SUCCESS  1.000  turns=34
```

Full process-mining set in the same run: **10/11 SUCCESS**. The one failure, `skill-pm-query-group-by-sugar` (0.750), is a pre-existing task untouched by this PR.

This replaces the "not run end-to-end" caveat in the description above. I expected `model_fields_data_kind` to need retuning; it passed clean, including the `file_contains` assertion on the pushed expression.

### Caveats on how much this proves

1. **CLI version.** The run used `@uipath/cli@1.198.0` (public npm `latest`). CI builds from `@uipath/cli@dev` on GitHub Packages — I could not install that (`gh` token lacks `read:packages`). `1.198.0` is **missing** `apps data-model`, `apps model fields` and `apps publish`, so three of these tasks ran against a CLI where the graded verb does not exist and **still scored 1.000** — the agent typed the skill-taught command once despite the unknown-command error, which is what the command-construction design intends. On the `@dev` build those commands additionally resolve, so this is the pessimistic case.
2. **One replicate, one model.** `claude-sonnet-5` locally; the nightly has been running `gemini-3.5-flash` and `gpt-5.6-terra`. Green here does not guarantee green under a different model — that is the same single-replicate weakness called out in #2545.

### Static checks, unchanged

- `coder-eval plan` — all 11 tasks valid.
- Tag vocabularies — clean.
- CLI-verb reachability with the checker from #2546 — 0 High, 0 Medium on the new tasks.
- Pattern probe: 28 cases asserting each new `command_pattern` fires on the command the skill teaches **and** does not fire on near-misses (e.g. `apps model get` must not satisfy the structural `data-model get` criterion; `--values 50,90` must not satisfy the fractional check). 0 wrong.

tests/tasks/uipath-process-mining/publish_promote_to_dashboards.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - type: command_executed
+    description: "Agent blocked on the async ingestion instead of hand-rolling a poll loop"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+pm\s+ingestions\s+create\b[\s\S]*--wait\b'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Agent must NOT verify by querying the published stage — it is unreachable from the CLI"
```
KosmasH: as discussed this does not seem to belong here. The agent should be able to reach publish stage when the ingestion is finished

tests/tasks/uipath-process-mining/query_info_percentile.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - type: command_executed
+    description: "Agent passed percentile points, which are fractions in 0..1"
+    tool_name: "Bash"
+    command_pattern: '--values\s+0?\.\d'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Agent must NOT query the published stage — it is unreachable from the CLI"
```
KosmasH: same as previous comment

tests/tasks/uipath-process-mining/query_info_percentile.yaml
```diff
@@ -0,0 +1,74 @@
...
+    tool_name: "Bash"
+    command_pattern: 'uip\s+pm\s+query\s+percentile\b.*--field\b'
+    min_count: 1
+    weight: 3.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent passed percentile points, which are fractions in 0..1"
+    tool_name: "Bash"
+    command_pattern: '--values\s+0?\.\d'
```
KosmasH: should this include also something as uip pm percentile?

tests/tasks/uipath-process-mining/source_template_selection.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - type: command_executed
+    description: "Agent read the generated transformations to learn the expected input tables"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+pm\s+transformations\s+(list|get)\b'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Agent must NOT use uipath.custom for a full multi-table source extract"
```
KosmasH, line 86: what about if it cannot find a matching app template? default will be custom or eventlog right?

tests/tasks/uipath-process-mining/source_template_selection.yaml
```diff
@@ -0,0 +1,95 @@
...
+    tool_name: "Bash"
+    command_pattern: '--type\s+uipath\.custom\b'
+    weight: 3.0
+
+  - type: command_not_executed
+    description: "Agent must NOT patch Cases.sql — that gotcha is uipath.custom-only"
```
KosmasH: what does this really prevent/check?

tests/tasks/uipath-process-mining/source_template_selection.yaml
Cellcote: as discussed, this is a thing for the cli, not for the skill.

tests/tasks/uipath-process-mining/source_template_selection.yaml
Cellcote: I will remove this, not needed.

tests/tasks/uipath-process-mining/query_info_percentile.yaml
KosmasH: Right now, the test only checks that some Bash command contains a --values argument with a decimal fraction. That can pass even if:

the agent used the wrong subcommand,
the command was malformed,
or --values appeared in some unrelated context.
A stronger pattern would be something like:

uip\s+pm\s+query\s+percentile\b.*--values\b

or, if you want to enforce the whole expected structure more tightly:

uip\s+pm\s+query\s+percentile\b.*--field\b.*--values\b

KosmasH (approved):

Cellcote:
## Re-run after the review changes

All three edited tasks re-run locally (`coder-eval`, `experiments/default.yaml`, tempdir, `claude-sonnet-5`):

```
skill-pm-query-info-percentile          SUCCESS  1.000   (2 replicates, 2/2)
skill-pm-publish-promote-to-dashboards  SUCCESS  1.000
skill-pm-source-template-selection      SUCCESS  1.000
```

Criteria/weights after the removals:

| Task | Criteria | Weight |
|---|---|---|
| `query_info_percentile` | 3 (was 5) | 7.0 (was 10.5) |
| `publish_promote_to_dashboards` | 5 (was 6) | 10.0 (was 12.0) |
| `source_template_selection` | 6 (was 7) | 12.5 (was 14.5) |

`coder-eval plan` valid on all 11 process-mining tasks. The skill-side correction for the published-stage claim is #2766.

Cellcote:
## The red `Run skill smoke tests` is a CI outage, not this PR

Rebased onto current `main` (was 145 commits behind); clean, no conflicts. All other checks pass. The smoke job fails on an AWS Bedrock authorization error before any agent runs:

```
Failed to authenticate. API Error: 403
User: arn:aws:iam::074103337776:user/coder-eval-skills-ci is not authorized to
perform: bedrock:InvokeModelWithResponseStream on resource: arn:aws:be...
```

Tasks report `status=ERROR score=0.000` after 4 retries each — the harness never reached the model, so nothing was graded. **2 of the 4 affected tasks are pre-existing** (`app-types-discovery-smoke`, `query-group-by-sugar`) and untouched here.

**It is repo-wide.** Every `smoke-skills` run since ~11:15 today where the smoke job *actually executed* has failed, including `fix/ixp-labellings-confirm-group` from another team. Runs that report green all have `smoke-job=skipped` (no skills changed → Bedrock never called), so workflow-level green is not evidence of recovery.

| Time (24 Aug) | smoke job actually ran? | Result |
|---|---|---|
| 08:53 – 09:55 | yes (×5, various branches) | success |
| 11:15 onward | yes | **failure, every time** |
| various | skipped | green, but proves nothing |

Needs someone with AWS access to restore `bedrock:InvokeModelWithResponseStream` for `coder-eval-skills-ci` — most likely the IAM policy no longer covers the model ARN being invoked. Not fixable from inside the PR.

Three attempts so far (original, re-run, post-rebase), identical failure each time. I've stopped retrying and am watching for recovery instead.

### Validation that did run, against current `main` (catalog `1.201.0-dev.8319`, 1507 verbs)

- `check-cli-verbs.py` on all 11 process-mining tasks — **0 High, 0 Medium**
- `coder-eval plan` — all 11 valid
- `check-skill-verbs.py` — no new findings
- Local `coder-eval` runs of all 5 new tasks — 1.000 each (see earlier comments)

PR #1, opened 2026-01-22
feat(EvalsBreakdown): break down eval skill

akshaylive:
This is done by moving items to references and assets.

PR #2549, opened 2026-08-10
 fix: output definition for connector list activity (#60381)

mrstark14:
Fixes two failure classes found in the 2026-08-10 IPE eval run, plus a Studio Web round-trip bug in complex-array authoring. All three are cases where the flow passes `flow validate` but breaks at runtime or in the designer — the class of defect validation cannot catch.

## 1. Connector output shape — why the output field definition change

**Failure:** `skill-flow-ipe-jira-search-triage` (run `2026-08-10_04-43-31`) built a structurally perfect flow, then `flow debug` faulted: `[400008] Failed to evaluate the input collection variable for the marker element (forEachIssue)`. The agent bound the loop to `=js:$vars.searchJiraIssues1.output.issues` — the raw Jira REST envelope shape.

**Root cause:** curated connector activities strip the vendor envelope. For `list` operations, `output` **is** the bare result array; `outputResponseDefinition.fields` in `registry get` describes one element (for Jira search: 235 fields rooted at `id`/`key`/`fields.*` — no `issues` prefix, no `total`). Nothing in the skill documented this, and `loop/impl.md`'s only collection example (`=js:$vars.fetchData.output.body.items`, an HTTP node) actively teaches the envelope-path instinct.

**Cardinality is decided by `connectorMethodInfo.operation` alone.** There is no explicit flag in the metadata; Studio Web's own designer derives it the same way (`fields-service.ts` `_computeOutputFieldsWithJitType`: `method.operation === Operation.List ? jitType.makeArrayType() : jitType`). `outputDefinition.output.type` is boilerplate `"object"` even for list ops — actively misleading, now explicitly called out.

**Evidence the fix targets the real gap:** same task, same tenant — `gpt-5.6-terra` guessed `.output.issues` and faulted; `claude-sonnet-5` (local run, score 1.0) read the full registry node, saw the element-scoped fields, bound `.output`, and debug went green with both seeded issues commented. Model-dependent coin flip → now a documented lookup.

**Changes:** `connector/impl.md` Step 6b gains an operation→shape table (`list` → bare array / loop binds `=js:$vars.<node>.output`; anything else → single object); `loop/impl.md` gains a one-line callout under the collection example contrasting connector vs HTTP shapes, cross-linking the table.

## 2. Complex array fields require the `=js:` expression form

Verified live (Slack `create-group-direct-message`, `users[*]`): a literal JSON array (`"users": ["U123", "U456"]`) passes `node configure` and `flow validate` but does not bind — the field round-trips empty in Studio Web. The working form is a `=js:` expression returning the array: `"users": "=js:(['U123', 'U456'])"`. Applies to all `[*]` array fields.

**Changes:** `connector/impl.md` array-fields table rows now show the `=js:` shape, the rule text warns that literal arrays validate but don't bind, and the single-element expression row is corrected (`"=js:([<expr>])"`, not `["=<expr>"]` — the latter is itself a literal array). `CAPABILITY.md`'s `[*]` bullet updated to match.

## 3. Eval task fixes

- **`ceql_where.yaml`:** prompt said "stop the flow with the error 'CeqlWhere test failed'" — Terminate has no error-message input, so models split between Terminate-with-label (passes grader) and script-`throw` (semantically faithful, fails grader's terminate assertion). This is what failed `skill-flow-ipe-ceql-where` on 2026-08-10 after passing for weeks. Reworded to "stop the flow immediately" — matches the Terminate node's registry description without naming the node or dangling an unattachable error text.
- **`complex_array.yaml`:** retargeted from Act! 365 `add-activity` to Slack `create-group-direct-message` (same `[*]` complex-array feature, verified via registry field schema). Criteria: flow valid, group-DM node with `users` populated, advisory participant check (non-gating — live connections resolve names to IDs), `flow validate` executed. Validate-only rationale documented in `description` (debug would open a real group DM).

## Passing-run claims

- `complex_array.yaml`: 4/5 replicates SUCCESS (0.875 each) local, `claude-sonnet-5`; the single failure was an SDK `RateLimitEvent` truncating the agent at 115s, not a task defect. All 4 passers authored the `=js:` array form from the updated docs.
- `jira_search_triage.yaml` (unchanged in this PR, validates fix #1): local run SUCCESS 1.0 — `flow debug` green, both seeded issues commented, loop bound `=js:$vars.searchIssuesByJql1.output`.


skills/uipath-maestro-flow/references/author/CAPABILITY.md
```diff
@@ -134,7 +134,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
...
+- **Never include `[*]` literally in a connector `bodyParameters` / `queryParameters` / `pathParameters` key** — `[*]` in `requestFields[].name` is an array marker, not part of the wire key. Strip it and pass a `=js:` expression returning the array (e.g. `"fields.labels[*]"` → `"fields.labels": "=js:(['a', 'b'])"`). Never a literal JSON array — it validates but does not bind (field round-trips empty in Studio Web). Supported only when `[*]` is the name suffix; `[*].` (segments after) is not authorable. `flow validate` misses this; runtime rejects the key. See [connector/impl.md — Step 6b](references/plugins/connector/impl.md) for the full table including expression-value handling.
```
baishalighosh:
```suggestion
- **Never include `[*]` literally in a connector `bodyParameters` / `queryParameters` / `pathParameters` key** — `[*]` in `requestFields[].name` is an array marker. Strip it and pass a `=js:` expression returning the array (e.g. `"fields.labels[*]"` → `"fields.labels": "=js:(['a', 'b'])"`). Never a literal JSON array — it validates but does not bind. Supported only when `[*]` is the name suffix; `[*]. See [connector/impl.md — Step 6b](references/plugins/connector/impl.md) for the full table including expression-value handling.
```

skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
```diff
@@ -303,17 +303,27 @@ Body field names in `bodyParameters` come from `inputDefinition.fields[].name` (
...
+> **Strip `[*]` from the key; pass a `=js:` expression returning the array.** The canvas UI is the source of truth — it omits `[*]` from the wire key on write and emits array values as expressions, and the CLI follows the same convention when it builds `inputs.detail`. The runtime rejects any key containing the literal `[*]` substring as unknown (regardless of value shape), so pass the stripped form in your `node configure --detail` payload. **A literal JSON array (`"users": ["U123ABC", "U456DEF"]`) passes `node configure` and `flow validate` but does not bind — the field round-trips empty in Studio Web.** **Distinct from `customFieldsRequestDetails.parameterValues`**, where `[*]` is encoded as `_array` — see Step 6c.
```
baishalighosh:
```suggestion
> **Strip `[*]` from the key; pass a `=js:` expression returning the array.**  The runtime rejects any key containing the literal `[*]` substring as unknown (regardless of value shape), so pass the stripped form in your `node configure --detail` payload. **A literal JSON array (`"users": ["U123ABC", "U456DEF"]`) passes `node configure` and `flow validate` but does not bind — the field round-trips empty in Studio Web.** **Distinct from `customFieldsRequestDetails.parameterValues`**, where `[*]` is encoded as `_array` — see Step 6c.
```

baishalighosh (commented):
There is too much verbosity in certain places

baishalighosh (approved):

skills/uipath-maestro-flow/references/author/CAPABILITY.md
mrstark14: Done

skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
mrstark14: Done

PR #2550, opened 2026-08-10
docs(guardrails): document BYOG admin configuration + a create smoke test [AL-513]

apetraru-uipath:
## What changed?

The **admin** side of bring-your-own guardrails (BYOG): registering the configuration in the first place, via `uip guardrails byo-configurations`. The agent-authoring side landed in #2516 (merged); this was split out so it carries its own `uipath-platform` CODEOWNERS review. Ticket: AL-513.

**New reference** — `skills/uipath-platform/references/guardrails/byo-configurations.md`, following the shape of the BYO LLM Gateway doc beside it:

- the subcommand surface — `list` / `list-validators` / `probe` / `create` / `update` / `delete` (no `get`)
- the `list` output field by field, worked calls for each verb, and the error paths
- wired into `uipath-platform/SKILL.md`: frontmatter triggers, task-navigation rows, references list

**Semantics worth stating explicitly, because they're easy to get wrong:**

- **`ValidatorName` is unique per tenant and is the only value an agent references.** Agents never pass a connection id — the platform resolves the connection server-side from the stored configuration, so an admin rebind is honored without touching agent code. Low-code (`byoValidatorName`) and coded (`ByoValidator(<name>)`) pin the same way.
- **`update` merges** — it fetches the record and PUTs back only what you supplied. `ValidatorName`/`ValidatorType` are locked at creation; delete and re-create to change them.
- **404 is ambiguous, and the CLI disambiguates it**: `ByoGuardrailsUnavailable` when the feature is off for the tenant, `ByoGuardrailConfigurationNotFound` when the id is wrong.
- The endpoint needs an org-admin **user** token and rejects application tokens; insufficient permissions return a 403 carrying the backend's reason.

**Validation is mandatory, not absent — this corrects an earlier draft of this PR.**

An earlier revision documented a *"known gap: no validation before save"*. That was true of the CLI when written and is now wrong: `create` **always** probes the connection/validator pair server-side and aborts if the probe fails, with no skip flag, and `update` re-probes whenever `--connection-id` changes. Verified against the shipped CLI (1.201.0) — a bad connection id returns `Code: ByoGuardrailProbeFailed` with a `REACHABILITY_FAILED` diagnosis and persists nothing.

The doc now documents that behavior, plus the two read-only verbs the probe brought with it:

```bash
# Does this connection serve this validator? (probes, creates nothing)
uip guardrails byo-configurations probe --connection-id <guid> --validator-type pii_detection

# Which validators does this connection expose? Each is a valid --validator-type.
uip guardrails byo-configurations list-validators --connection-id <guid>
```

`list-validators` is the fastest way to fix a `VALIDATOR_NOT_SUPPORTED` failure — it returns the exact ids the connector accepts, with each one's `AllowedScopes` and `Parameters` schema.

**New test** — `tests/tasks/uipath-platform/guardrails/create_configuration_smoke.yaml`, plus a selective `uip` shim (`mock_template/mocks/uip`).

The prompt is what an admin would actually ask — register `byog-smoke-pii` for `pii_detection` through our Azure AI Content Safety connection with fallback, and *did the platform check the connection when it saved it?* It does not number the steps, name the connection id, or say which commands to run. The shim answers the BYOG and Integration Service commands (everything else passes through to the real CLI), so there are no auth errors to warn the agent about — which is what lets the prompt drop the coaching. An earlier revision spelled out all four steps and told the agent "commands WILL fail, run them anyway"; that graded prompt compliance, and an agent could have passed it with the skill unread.

The shim keeps the parts that make the test mean something:

- **`create` really probes.** A fabricated `--connection-id` fails with `ByoGuardrailProbeFailed` and saves nothing, so the agent has to discover the connection instead of inventing a GUID.
- **Two connections are listed**, one a decoy that serves no validators — binding the right one takes reading `ConnectorName`, not taking the first id.
- **Created records persist**, so a read-back `list` agrees with the create. (Caught by a real run: a static list answered "not there" immediately after `create` said "Created", and the agent reported the write as failed.)
- **`--help` passes through** to the real CLI. Canned JSON in answer to a flag question reads as "that flag does not exist".

Graded: the two pre-flight calls, the create call's shape, that no `byo-configurations get` verb is invented (there isn't one), that the `uipath-platform` skill was triggered, and an `llm_judge` on the validation answer. Two fixes to the criteria: the `list` pattern also matched `list-validators`, and the create pattern now requires the discovered connection id rather than any GUID.

`byo-configurations.md` now prescribes that pre-flight — `list` first because `ValidatorName` is tenant-unique, then `uip is connections list` for the id — so the behavior the test grades is taught rather than assumed.

`cleanup_byog.py` is gone: with `create` mocked, nothing lands on a tenant, so there is nothing to sweep.

Also fixes a stale verb in Diagnostics: `uip is connections get` doesn't exist — the real one is `ping`.

## How has this been tested?

**10 runs of the changed task, two models, all passing.** 5 repeats on `claude-sonnet-5` (claude-code) and 5 on `gpt-5.6-terra` (codex), `experiments/default.yaml`, `--driver tempdir`.

| | Deterministic criteria | Discovered the real connection id | Invented a `get` verb |
|---|---|---|---|
| `claude-sonnet-5` | 5/5 replicates, 5/5 criteria | 5/5 | never |
| `gpt-5.6-terra` | 5/5 replicates, 5/5 criteria | 5/5 | never |

**On the `llm_judge`:** it needs a judge transport (Bedrock, or `ANTHROPIC_API_KEY`), which the machine these ran on does not have — so it scored 0.0 in all 10 and each run reads as weighted **0.826** rather than 1.000. That is the only thing missing from a clean sweep. CI supplies `ANTHROPIC_API_KEY` and Bedrock creds to the smoke job, so it grades there; `llm_judge` in a smoke task is not new (14 others use it, 4 of them in `uipath-platform`). I read all 10 final answers by hand instead of trusting the score: every one reports that the platform probed before saving and that no separate manual check is needed. Sample (`gpt-5.6-terra`, replicate 01):

> Yes—the platform checked the connection when saving it. Creation performs a mandatory server-side probe of both reachability and `pii_detection` support; a failed probe would not have saved the configuration.

**Other checks**

- `scripts/check-cli-verbs.py` and `scripts/check-task-driver.py` clean.
- `scripts/check-skill-verbs.py`: **0 blocking findings** (the gate's criterion is `severity == "Stale" and matched_prefix == null`). `probe` and `list-validators` shipped in CLI 1.201 while the committed catalog snapshot is `1.200.0-dev.8173`, so they surface as 3 prefix-matched soft warnings; the next nightly catalog refresh clears them.
- Probe behavior verified against CLI 1.201.0 rather than from docs — `create` with a bad connection id returns `ByoGuardrailProbeFailed` / `REACHABILITY_FAILED` and saves nothing; `create --help` and `update --help` both state the probe and the absence of a skip flag. The shim's responses are copied from the CLI's own documented examples so the fields the skill teaches are the fields that come back.
- The shim's 11 paths were checked individually (create with a good id / fabricated id / decoy connection / duplicate name / missing flags, probe both ways, `list-validators`, `ping`, delete with and without `--force`), and the read-back sequence re-checked after the persistence fix.

## Review comments addressed

- **`byoConfigurationId` is not used — `byoValidatorName` is** (@valentinabojan): fixed. The `Id` field row now says agent authors don't reference it directly, and both low-code and coded pin by validator name.
- **"We also need a CLI command for test configuration similar to BYO llm config"** (@valentinabojan): this now exists and is documented — `create`/`update` probe automatically, and `probe` / `list-validators` expose it as standalone read-only verbs. The earlier "known gap" wording is gone.
- **SKILL.md Task Navigation vs. References duplication** (@valentinabojan): kept, matching the existing convention for every other reference doc in that file (LLM Gateway, Traces, Integration Service, Data Fabric all appear in both). Happy to change if you'd rather BYOG be the exception.

## Related

- #2516 (merged) — the agent-authoring half: low-code and coded BYO guardrail docs plus three coder-eval tasks.

- [x] None


apetraru-uipath:
This file/doc moved here from #2516 (agent-authoring PR) so it doesn't need `uipath-platform` CODEOWNERS approval riding on that PR. Carrying forward review context from the threads on #2516, since they now point at a diff this PR no longer has:

1. **`byoConfigurationId` is not a real field** (valentinabojan) — fixed to `ByoValidatorName`, matching the coded-agent `ByoValidator(<ValidatorName>)` pattern. Both low-code and coded now reference a BYOG configuration the same way.
2. **Need a CLI command to test/validate a configuration, like BYO LLM** (valentinabojan) — confirmed against the CLI source: `byo-configurations create`/`update` run no server-side validate/probe today, unlike `llm-configuration byo-connections`' mandatory `POST .../validate`. That's a real feature gap needing a CLI change, not something this docs PR can add — added a "Known gap" section here documenting it and pointing at `uip is connections ping` as the closest partial check available now. Worth a follow-up ticket against `guardrails-tool`.
3. **Task Navigation vs. References list duplication in SKILL.md** (valentinabojan) — this matches the existing pattern for every other reference doc in the file (LLM Gateway, Traces, Integration Service, Data Fabric all appear in both places); kept for consistency rather than making BYOG the one exception.
4. **"Do we need this skill?"** (apetraru-uipath, original PR author) — still open, unanswered on #2516. Not something I'm resolving on your behalf — flagging it here so it doesn't get lost with the move.

skills/uipath-platform/references/guardrails/byo-configurations.md
```diff
@@ -0,0 +1,198 @@
...
+| `ValidConnection` | Whether the underlying Integration Service connection currently resolves and is healthy. `false` means the connection was disabled, revoked, or rotated since registration. |
+| `ConnectorKey` / `ConnectorName` / `ConnectionName` | Resolved metadata about the underlying IS connector/connection, for display. `null` when unresolved. |
+
+Empty `Data: []` is a valid result — it means the tenant has no BYOG configurations registered.
+
+## `create`
+
+**Run the two discovery calls first — never fabricate either value.**
+
+1. **`list`** — `ValidatorName` is unique across the tenant, so check the name is free before you spend a create on it. A duplicate is rejected by the backend.
```
valentinabojan: Should we put the full command here? `list` is a bit ambigous

tests/tasks/uipath-platform/guardrails/create_configuration_smoke.yaml
```diff
@@ -0,0 +1,114 @@
...
+  Our security team wants agent PII checks to run through our own Azure AI
+  Content Safety instance instead of the UiPath built-in one.
+
+  Set that up on this tenant: register it as a bring-your-own guardrail
+  configuration named `byog-smoke-pii` for the `pii_detection` validator
+  type, calling through our `Azure AI Content Safety - Prod` connection.
+  If our provider is ever unreachable at runtime the check should fall
+  back to the UiPath built-in guardrail rather than failing outright.
+
+  Then tell me one thing: when the configuration was saved, did the
```
valentinabojan, line 37: Shouldn't this be actually verified by the test? The skill should always probe before saving.

skills/uipath-platform/references/guardrails/byo-configurations.md
apetraru-uipath: Agreed — fixed in 50556c642. Step 1 is now the full `uip guardrails byo-configurations list --output json`; bare `list` was especially ambiguous sitting right next to `uip is connections list` in step 2.

Swept the rest of the page for the same thing, so every point that tells you to *run* something now spells it out:

- the "no `get` verb" note — "run `uip guardrails byo-configurations list --output json` and filter by `Id`"
- the optional pre-check, now a real code block for both `list-validators` and `probe` instead of naming them inline
- the not-found recovery step — "Re-run `uip guardrails byo-configurations list --output json` to get valid ids"

I left bare verb names where the prose or the subcommand table is *referring* to a verb rather than instructing a call (e.g. "`ValidatorName`/`ValidatorType` are locked after creation — `delete --force` and `create` again"), which matches how the sibling `byo-connections.md` page reads. Happy to expand those too if you'd rather have zero bare verbs anywhere.

skills/uipath-platform/references/guardrails/byo-configurations.md
```diff
@@ -0,0 +1,198 @@
...
+- `--connection-id`, `--validator-name`, `--validator-type` are required.
+- The connection/validator pair is **probed server-side before the record is saved** — no skip flag. See [Validation](#validation-mandatory-before-save).
+- `--fallback-on-ui-path` defaults to **false**. Configurations are created **enabled** by default; pass `--disabled` to save it switched off.
+- `--validator-name` must be unique across the tenant — a duplicate is rejected by the backend.
+- Success: `Code: ByoGuardrailConfigurationCreated`, `Data` is the created configuration (use `Data.Id` for later `update`/`delete`).
+
+## `update`
+
+```bash
+uip guardrails byo-configurations update <configuration-id> --connection-id <new-uuid> --output json
```
valentinabojan, line 100: Don't we need tests for the other flows as well, apart from create?
Also, I was curious: what would a user prompt look like for the update or delete cases?

tests/tasks/uipath-platform/guardrails/create_configuration_smoke.yaml
apetraru-uipath: Good catch, and it exposed a real gap — fixed in 469b40312.

The subtlety: the probe is **server-side inside `create`** ([`byo-configurations.ts:396-408`](https://github.com/UiPath/cli/blob/main/packages/admin/guardrails-tool/src/commands/byo-configurations.ts#L396-L408) runs `probeOrAbort` before the POST), so the agent never triggers it. That means this `llm_judge` only ever graded whether the agent *said* the probe happened — nothing proved the platform actually refuses a bad pair. You were right that it wasn't verified.

New `probe_abort_smoke.yaml` verifies it. The prompt hands over a connection id that really belongs to a Slack connection — the kind of wrong id that gets pasted off a ticket — so the pair can't be registered. The load-bearing criterion is deterministic and reads the shim's state file, not the agent's summary:

```
python3 -c "... sys.exit(1 if any(c['ValidatorName'] == 'byog-abort-pii' for c in s.get('created', [])) else 0)"
```

I checked that assertion fails when it should: force a create through the *good* connection and it goes red, so it isn't passing vacuously. Paired with the existing create task, the two now show the probe gating the save in both directions.

One thing worth flagging, because the runs changed my mind about the criteria. I first required the agent to *attempt* the doomed create — and gpt-5.6-terra failed it 3/3, because it pre-checked with `probe`/`list-validators`, saw the connection serves no validators, and correctly declined to create. That's the better behaviour and it's what this page recommends ("cheaper than reading a failed create"), so the criterion was wrong, not the agent. It now accepts either route; what it does *not* accept is never reaching the `byo-configurations` family at all. After the fix: sonnet 3/3 (attempts the create), terra 5/5 (pre-checks and declines).

**On "the skill should always probe before saving"** — I left the explicit pre-check optional rather than mandatory, since the platform enforces it unconditionally and requiring it would add two CLI round-trips to every create. Happy to make it mandatory in the doc and grade it if you'd rather the skill always front-run the platform's own check.

skills/uipath-platform/references/guardrails/byo-configurations.md
apetraru-uipath: Agreed on both counts — added in 469b40312.

You're right that the coverage was thin: the analogous feature already has **8 tasks, one per flow** (`tests/tasks/uipath-platform/llmgateway/` — list, create+validate, update, delete, get, reprobe, product-configs, audit), and BYOG had 1. It's now 4: create, probe-abort, update, delete.

**And to answer what you were curious about — the prompts, verbatim:**

`update_rebind_smoke.yaml`:

> We've moved our content-safety provider to the DR instance, so our PII guardrail is calling the wrong one. Point the `byog-pii-prod` guardrail configuration at the `Azure AI Content Safety - DR` connection instead.
>
> Leave it switched off once it's repointed — QA signs off tomorrow before we turn it back on. Keep the guardrail's name as it is; agents already reference it.

`delete_smoke.yaml`:

> We've dropped the contract for the external provider behind our `byog-legacy-harmful` guardrail configuration, so it needs to go — permanently, not just switched off.
>
> Remove it, and tell me what happens to any agent still pointing at it.

Both name the configuration and the connection but never their ids, so the agent has to resolve them through `list` and `is connections list` — which also exercises the "no `get` verb" rule. What each one grades beyond the happy path:

- **update** — must not delete-and-re-create (that changes the `Id`), must not pass the locked `--validator-name`/`--validator-type`, and a state-file assertion confirms the record really is on the DR connection *and* disabled.
- **delete** — `--force` is mandatory, must not substitute `update --disabled` when the user said permanent, and an `llm_judge` checks it warns that agents still referencing the name break at runtime. All 6 runs got that right.

Two notes on how these are built, since they diverge from the llmgateway tasks sitting next to them:

1. **Prompt style.** The llmgateway tasks use numbered step lists plus *"commands WILL fail with auth errors — you MUST still run every command"*. These use a natural request plus the `uip` shim so the commands actually answer — otherwise the test grades prompt compliance rather than the skill. I didn't restyle the llmgateway ones; that's a different feature's coverage and would bury this diff.
2. **The shim needed real work to support these** — `update` was unhandled and silently fell through to the real CLI, and `delete` couldn't remove a *seeded* record (it returned `Deleted` while `list` still showed it, the same read-back contradiction I'd already hit once here). Both fixed, with 43 assertions over the shim including negative controls.

Verification: 32 runs across claude-sonnet-5 and gpt-5.6-terra — create 3/3+3/3, update 3/3+3/3, delete 3/3+3/3, probe-abort 3/3+5/5 on every deterministic criterion.

Not added, and easy to bolt on if you want full parity with llmgateway: a read-only `list` audit (surfacing `ValidConnection: false` / disabled configs) and the disable-a-config-whose-connection-is-dead case. Say the word.

valentinabojan (approved):

gabrielavaduva (approved):

PR #2551, opened 2026-08-10
test(uipath-insights): add e2e coverage for the jobs commands

celestinryf:
Related to [IN-13526](https://uipath.atlassian.net/browse/IN-13526).

### What

Adds the first e2e coverage for `uipath-insights`: two task YAMLs with their sidecar checkers (`envelope-contract/all_commands_envelope_e2e.yaml`, `job-health/job_health_investigation_e2e.yaml`, `check_all_commands_envelope.py`, `check_job_health_investigation.py`) plus a shared `_shared/envelope_check.py` holding the per-subcommand envelope `Code` strings from `jobs.ts`. Five new files, all under `tests/tasks/uipath-insights/`. The five existing smoke tasks, `SKILL.md`, and the references are untouched.

### Why

The repo requires at least one smoke and one e2e task per skill (`tests/README.md:269`) and this skill had only smoke. The five smoke tasks grade that command strings were attempted, regardless of outcome, so nothing currently grades whether `uip insights jobs` returns its shipped contract: `Result: Success` with the exact per-subcommand `Code`.

### Scope

The seven shipped Jobs commands only, read-only, no tenant mutation and therefore no cleanup. Nothing here teaches the proposed Filters, RBAC, or Alert commands as available. The Filters skill content and its smoke task are a separate PR that waits on those commands reaching `cli/main`.

### Testing

Ran both tasks locally and they passed. Run `2026-08-06_13-34-09`, claude-code agent, `uip 1.198.0`, coder_eval 0.8.10: `skill-insights-all-commands-envelope-e2e` 1.000 (12/12 criteria), `skill-insights-job-health-investigation-e2e` 1.000 (9/9). Four earlier runs also passed at 1.000: `2026-08-06_12-25-44`, `2026-08-06_12-13-08`, `2026-08-04_12-24-50`, `2026-08-04_12-29-42`. The last run had live job data, so the report-grounding assertion was exercised rather than skipped through the empty-window path. Checker failure paths exercised locally against synthesized artifacts: ungrounded report, rejected envelope, unparseable envelope, missing file, no `uip` on PATH, probe timeout. `scripts/check-task-driver.py` and `scripts/check-cli-verbs.py` are clean.

### Limitations

No pytest unit tests for the checkers. maestro-flow, maestro-case and uipath-agents ship them, but running them in CI needs a `paths:` entry and a job in `.github/workflows/test-helpers.yml`, outside this PR's ownership. Happy to follow up if the CI owners want it.

The checkers verify that each saved envelope is structurally correct and that the report is grounded in the CLI's response, but not that a given file is the response to its own command. Both tasks need a tenant that answers `uip insights jobs`; a tenant that rejects the read fails with the CLI's own `Message` and `Instructions` rather than degrading.

### How to Review

Start with the two `check_*.py` files and `_shared/envelope_check.py`. That is where the contract lives. The YAMLs are mostly criteria lists.

### Checklist

- [x] Tests added/updated for new logic — the change is the tests; run evidence above
- [x] No secrets or credentials in code
- [x] Commits follow conventional format
- [ ] PR ≤ 400 lines — 603 added lines; the five files are one slice, and a task YAML is not separable from its checker

### CI note

The PR gate runs only `smoke`-tagged tasks, so neither new task is exercised by CI on this PR. The local runs above are the evidence; nightly is where they run on a schedule.

[IN-13526]: https://uipath.atlassian.net/browse/IN-13526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/tasks/uipath-insights/_shared/envelope_check.py
+++ b/tests/tasks/uipath-insights/_shared/envelope_check.py
@@ -168,13 +168,36 @@ def probe_live_cli(timeout: int = LIVE_PROBE_TIMEOUT) -> bool:
     return True
 
 
+def _get_ci(mapping, *candidate_keys: str, default=None):
+    """Case-insensitively read the first present candidate key from ``mapping``.
+
+    The envelope wrapper's ``Result``/``Code``/``Data`` casing IS the shipped
+    contract and stays literally asserted in `check_envelope`. The keys INSIDE
+    ``Data`` are not what these checkers grade, so runtime-payload reads go
+    through this accessor (same pattern as `_get_ci` in
+    uipath-maestro-flow/_shared/flow_check.py): a camelCase ``processName``
+    would otherwise return nothing silently and skip the report-grounding
+    branch while the task still passes.
+    """
+    if not isinstance(mapping, dict):
+        return default
+    lowered = {k.lower(): k for k in mapping.keys() if isinstance(k, str)}
+    for candidate in candidate_keys:
+        actual = lowered.get(candidate.lower())
+        if actual is not None:
+            return mapping[actual]
+    return default
+
+
 def process_names(data) -> list:
     """Process names carried by a jobs envelope's `Data.ProcessName`.
 
     The row columns come back as arrays that are sometimes nested one level per
     grouping bucket (`JobCountByTime` is `[[1]]` on the same response), so
     collect strings at any depth and drop blanks. Empty list when the column is
-    absent or null, which is what an empty window returns.
+    absent or null, which is what an empty window returns. The column is read
+    case-insensitively so a serialization change to camelCase can't silently
+    turn a populated window into an empty-looking one.
     """
     names: list = []
 
@@ -187,5 +210,5 @@ def process_names(data) -> list:
                 walk(item)
 
     if isinstance(data, dict):
-        walk(data.get("ProcessName"))
+        walk(_get_ci(data, "ProcessName"))
     return names
--- a/tests/tasks/uipath-insights/_shared/envelope_check.py
+++ b/tests/tasks/uipath-insights/_shared/envelope_check.py
@@ -1,0 +1,191 @@
+#!/usr/bin/env python3
+"""Shared envelope validation helpers for the uipath-insights e2e checkers.
+
+`uip insights` success envelopes are PascalCased by the CLI formatter, so
+callers assert `Result`/`Code`/`Data`, not `result`/`code`. Code strings come
+from cli/packages/insights-tool/src/commands/jobs.ts. No envelope check asserts
+Data content: the shared tenant's job history grows over time, so both empty and
+populated windows must pass structurally. `has_signal` and `process_names` read
+Data only so callers can tell an empty window from a populated one and can
+cross-check an agent-written report against what the CLI actually returned.
+
+Checkers import this module via a sys.path insert relative to their own file
+(coder_eval runs them as `python3 $TASK_DIR/<checker>.py` with the sandbox as
+cwd, so a plain relative import would not resolve).
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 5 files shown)


tests/tasks/uipath-insights/_shared/envelope_check.py
```diff
@@ -0,0 +1,191 @@
...
+    def walk(value) -> None:
+        if isinstance(value, str):
+            if value.strip():
+                names.append(value.strip())
+        elif isinstance(value, list):
+            for item in value:
+                walk(item)
+
+    if isinstance(data, dict):
+        walk(data.get("ProcessName"))
```
rliuup: issue (non-blocking): `data.get("ProcessName")` pins the runtime-payload casing and fails silently — under a camelCase `Data` this returns `[]` while `has_signal` still reports signal, so `_check_report` skips the grounding branch and the task still scores 1.000 with the report ungraded against CLI output. The repo's existing fix for this class is the case-insensitive `_get_ci` accessor in `uipath-maestro-flow/_shared/flow_check.py`; `tests/scripts/test_runtime_payload_key_casing.py` won't flag it here since it only guards `finalStatus`/`elementExecutions`.

tests/tasks/uipath-insights/envelope-contract/all_commands_envelope_e2e.yaml
```diff
@@ -0,0 +1,130 @@
...
+    description: "Agent activated the uipath-insights skill"
+    skill_name: "uipath-insights"
+    expected_skill: "uipath-insights"
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent ran summary with the 24-hour window (or an absolute epoch-ms range)"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+insights\s+jobs\s+summary\s.*--(time-range[=\s]+1440|started-after[=\s]+\d{13})'
```
rliuup: suggestion (non-blocking): `.*` crosses `&&`/`;`/`|`, so a neighbouring command in the same Bash invocation satisfies the flag — `summary --time-range 60 > summary.json && ... completed-timeline --time-range 1440 ...` passes this 24-hour-window criterion, and a summary with no time flag at all passes too. `[^&;|\n]*` (the form #2552 uses, or the lookahead style in `uipath-test/link_automation_and_run.yaml`) closes both; the 10 sibling patterns across the two YAMLs have the same shape.

rliuup (approved):
LGTM with nits.

suranabhavya (approved):
LGTM

rohitjain-uipath (approved):

tests/tasks/uipath-insights/_shared/envelope_check.py
celestinryf: Fixed in f95c9e045: added a `_get_ci` accessor (same shape as flow_check.py's) and routed the `process_names` read through it. The envelope-level Result/Code/Data reads stay literal on purpose, since that casing is the shipped contract the checkers assert.

tests/tasks/uipath-insights/envelope-contract/all_commands_envelope_e2e.yaml
celestinryf: Fixed in f95c9e045 for all 13 patterns across the two YAMLs (11 gating, 2 advisory). I used `(?:[^&;|\n]|\\\n)*` rather than plain `[^&;|\n]*`: your continuation-gap comment on #2552 applies here in reverse, since these are positive criteria and the skill's examples use backslash continuation, so the bare class would false-fail a legitimate run.

PR #2552, opened 2026-08-10
feat(uipath-insights): teach filter discovery and add filters smoke eval

celestinryf:
Related to [IN-13523](https://uipath.atlassian.net/browse/IN-13523).

### What

Teaches the four filter discovery command groups landing in [UiPath/cli#3449](https://github.com/UiPath/cli/pull/3449): `filter-folders`, `filter-processes`, `filter-queues`, `filter-machines`, each with a `list` subcommand. `SKILL.md` gets filter hooks in `description`/`when_to_use`, a Filter Discovery Commands section (four rules: no time flags, client-side `--limit`/`--offset`, empty is not proof of absence, results are permission-bounded) and a discover-before-you-act workflow; Critical Rule 1 is scoped to `jobs` commands only. `references/investigation-playbook-guide.md` routes folder-key lookups through `filter-folders list` first, with `uip or folders list` as the fallback. Adds `tests/tasks/uipath-insights/filters/smoke.yaml` and activation rows 026-029. The five jobs smoke tasks, the e2e tasks from #2551, and `BASELINES_PCT` are untouched.

### Why

The commands ship in [UiPath/cli#3449](https://github.com/UiPath/cli/pull/3449) (review approved, merge pending). Without skill guidance, agents guess folder keys or pass `--time-range` to commands that reject it; the discovery workflow makes them page through results before concluding a resource is absent. The backend applies a fixed 30-day activity window, which is why the skill frames empty results as "no recent activity", not "does not exist".

### Scope

Skill content, one smoke task, four activation positives. E2e-depth coverage for these commands lives in the cli repo (`scenarios/insights-tool/filter-discovery.md` S1-S8 and `filter-discovery.e2e.test.ts`, both on cli#3449); this skill already has its e2e slice in #2551. The activation baseline (`BASELINES_PCT`) is deliberately not edited here; re-measuring recall with the new rows is the follow-up once a baseline entry exists for this skill.

### Testing

- `scripts/check-cli-verbs.py` and `scripts/check-task-driver.py` both pass on the new smoke task, run 2026-08-10.
- The skill stays inside its size budgets. The frontmatter description is 492 characters against the 1024 cap. Description plus when_to_use is 1242 characters, and Claude Code truncates at 1536. SKILL.md is 316 lines, about 3.8k tokens against the 5k budget.
- The smoke task itself has not been run. It cannot pass yet; see Limitations.

### Limitations

`skill-insights-smoke-filter-discovery` has no passing run yet, and cannot have one: the four commands exist only on cli#3449, and the PR smoke gate installs `@uipath/cli@dev`, which picks them up only when that PR merges to `cli/main`. The 2026-08-06 catalog snapshot (`1.200.0-dev.8116`) carries 0 filter verbs, consistent with that. A real passing-run claim will be added to this PR once the commands are on `cli/main`; this PR should not merge before that run exists.

Activation recall is not re-measured. The activation gate skips `uipath-insights` until a `BASELINES_PCT` entry exists, so rows 026-029 are inert for gating until the baseline lands.

### How to Review

The Filter Discovery Commands section in `SKILL.md` carries the weight, especially the no-time-flags rule and the empty-is-not-absence rule, which are the two mistakes agents actually make with these commands. The playbook edits are routing, and the smoke YAML is a criteria list.

### Checklist

- [x] Tests added/updated for new logic — the filters smoke task; run evidence pending cli#3449 merge, see Limitations
- [x] No secrets or credentials in code
- [x] Commits follow conventional format
- [x] PR ≤ 400 lines

### CI note

The smoke gate (`smoke-skills.yml`) will fail this PR's filters task until cli#3449 merges, since the `@uipath/cli@dev` install cannot have the commands before then. The `lint-tasks.yml` advisory may also flag the missing passing-run claim; both resolve with the post-merge run described in Limitations.

[IN-13523]: https://uipath.atlassian.net/browse/IN-13523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/skills/uipath-insights/SKILL.md
+++ b/skills/uipath-insights/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: uipath-insights
-description: "UiPath Insights job monitoring via `uip insights` — query job execution metrics, failure analysis, and process performance. Covers job KPIs, failure reasons, completion trends, process breakdowns. For Orchestrator job start/stop/logs→uipath-platform, root-cause analysis of specific errors→uipath-troubleshoot, RPA workflow authoring→uipath-rpa."
-when_to_use: "User says 'job failures', 'automation health', 'job success rate', 'processing time', 'which processes fail the most', 'failure reasons', 'job trends', 'how many jobs ran', 'insights dashboard', 'job metrics', 'job KPIs', 'job performance', 'uncompleted jobs', 'pending jobs', 'faulted jobs', 'job timeline', 'process details'. Also 'uip insights', 'insights jobs'. NOT for starting/stopping jobs (uipath-platform), NOT for root-cause debugging of a specific job error (uipath-troubleshoot), NOT for queue metrics (not yet supported)."
+description: "UiPath Insights job monitoring via `uip insights` — query job execution metrics, failure analysis, and process performance. Covers job KPIs, failure reasons, completion trends, process breakdowns. Also filter discovery — folders, processes, queues, and machines with recent activity, for choosing exact scope before alerts or filtered queries. For Orchestrator job start/stop/logs→uipath-platform, root-cause analysis of specific errors→uipath-troubleshoot, RPA workflow authoring→uipath-rpa."
+when_to_use: "User says 'job failures', 'automation health', 'job success rate', 'processing time', 'which processes fail the most', 'failure reasons', 'job trends', 'how many jobs ran', 'insights dashboard', 'job metrics', 'job KPIs', 'job performance', 'uncompleted jobs', 'pending jobs', 'faulted jobs', 'job timeline', 'process details', 'which folders can you see', 'what processes are active', 'find the folder key', 'discover queues', 'which machines reported', 'filter-folders', 'filter-processes'. Also 'uip insights', 'insights jobs'. NOT for starting/stopping jobs (uipath-platform), NOT for root-cause debugging of a specific job error (uipath-troubleshoot), NOT for queue item metrics (queue discovery is supported via filter-queues; metrics are not)."
 allowed-tools: Bash, Read
 ---
 
@@ -9,7 +9,7 @@ allowed-tools: Bash, Read
 
 Insights provides analytics and monitoring for UiPath automation execution. This skill covers **job monitoring** — querying aggregated job execution data for dashboards, health checks, and failure investigation.
 
-All operations go through `uip insights jobs <subcommand> --output json`.
+Job monitoring goes through `uip insights jobs <subcommand> --output json`; scope discovery goes through the `uip insights filter-*` groups.
 
 ---
 
@@ -20,6 +20,7 @@ All operations go through `uip insights jobs <subcommand> --output json`.
 - Monitoring job execution trends over time (completed/uncompleted timelines)
 - Getting per-process performance breakdowns
 - Drilling into specific failure details for investigation
+- Discovering which folders, processes, queues, or machines have recent activity, to pick an exact scope (see Filter Discovery Commands)
 
 > **Not in scope:** Starting, stopping, or managing individual jobs (use `uip or jobs` via uipath-platform). Root-cause debugging of a specific job's error (use uipath-troubleshoot). Queue item metrics, robot utilization, or dashboard CRUD (not yet available in CLI).
 
@@ -47,7 +48,7 @@ uip login tenant set MyTenant
 
 ## Critical Rules
 
-1. **A time range is always required.** Every `uip insights jobs` command needs either `--time-range <minutes>` (relative) or both `--started-after <epoch-ms>` and `--started-before <epoch-ms>` (absolute). Without one, the command fails. For absolute ranges, pass literal epoch-millisecond numbers, resolved beforehand (see Workflow: Absolute Time Range). Common values:
+1. **A time range is always required — for `jobs` commands only.** Every `uip insights jobs` command needs either `--time-range <minutes>` (relative) or both `--started-after <epoch-ms>` and `--started-before <epoch-ms>` (absolute). Without one, the command fails. `filter-*` commands take no time flags at all (see Filter Discovery Commands). For absolute ranges, pass literal epoch-millisecond numbers, resolved beforehand (see Workflow: Absolute Time Range). Common values:
    - `--time-range 60` — last 1 hour
    - `--time-range 1440` — last 24 hours
    - `--time-range 10080` — last 7 days
@@ -163,6 +164,45 @@ uip insights jobs failure-details --time-range 1440 --output json
 
 ---
 
+## Filter Discovery Commands
+
+Four discovery groups expose scope candidates with recent Insights activity:
+
+```bash
+uip insights filter-folders list --output json     # {FolderName, FolderKey} rows
+uip insights filter-processes list --output json   # {ProcessName, FolderKey} rows
+uip insights filter-queues list --output json      # {QueueName, FolderKey} rows
+uip insights filter-machines list --output json    # {MachineName, MachineKey} rows
+```
+
+Rules that differ from the jobs commands:
+
+1. **No time flags.** The backend applies its own fixed 30-day activity window. `--time-range` is rejected as an unknown option. Do not add it.
+2. **`--limit` / `--offset` paginate client-side.** Default limit is 50; `Pagination.Total` and `HasMore` show whether more rows exist.
+3. **Empty is normal and is not proof of absence.** An empty result means no visible activity in the last 30 days for the current caller. Never tell the user a resource does not exist based on filter output.
+4. **Results are permission-bounded.** Folders, processes, and queues are restricted to folders the caller can access; machines are tenant-wide.
+
+### Workflow: discover before you act
+
+Before any command that takes a folder key, process name, machine name, or (future) alert scope:
+
+```bash
+# 1. Discover candidates
+uip insights filter-folders list --output json
+
+# 2. Match the user's words against FolderName values. If Pagination
+#    shows HasMore: true, page with --offset (or raise --limit) until
+#    every row has been seen BEFORE concluding anything is absent.
+#    Multiple matches -> ask the user which one. Zero matches across
+#    all pages -> report that nothing with that name was active in the
+#    last 30 days and ask for more detail. Never guess or fabricate a key.
+
+# 3. Use the exact key from the response
+uip insights jobs summary --time-range 1440 --folder-key "<FolderKey from step 1>" --output json
+```
+
+---
+
 ## Workflow: Investigate Job Health
 
 Follow this pattern when a user asks about automation health or job failures:
@@ -214,7 +254,7 @@ uip insights jobs failures-by-reason --time-range 1440 \
   --output json
 ```
 
-To discover available folder keys, use `uip or folders list --output json` (from the uipath-platform skill).
+To discover available folder keys, use `uip or folders list --output json` (from the uipath-platform skill), or `uip insights filter-folders list --output json` for folders with recent Insights activity — usually the right scope source for monitoring tasks, since the Orchestrator list shows all folders the caller can administer.
 
 ---
 
--- a/skills/uipath-insights/SKILL.md
+++ b/skills/uipath-insights/SKILL.md
@@ -169,10 +169,10 @@ uip insights jobs failure-details --time-range 1440 --output json
 Four discovery groups expose scope candidates with recent Insights activity:
 
 ```bash
-uip insights filter-folders list --output json     # {folderName, folderKey} rows
-uip insights filter-processes list --output json   # {processName, folderKey} rows
-uip insights filter-queues list --output json      # {queueName, folderKey} rows
-uip insights filter-machines list --output json    # {machineName, machineKey} rows
+uip insights filter-folders list --output json     # {FolderName, FolderKey} rows
+uip insights filter-processes list --output json   # {ProcessName, FolderKey} rows
+uip insights filter-queues list --output json      # {QueueName, FolderKey} rows
+uip insights filter-machines list --output json    # {MachineName, MachineKey} rows
 ```
 
 Rules that differ from the jobs commands:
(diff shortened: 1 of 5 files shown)


celestinryf:
Addressing the two bot reviews, point by point (changes landed in `8aad100f0`):

**PR review findings**

- *"When to Use" lacks a discovery bullet* — added: "Discovering which folders, processes, queues, or machines have recent activity, to pick an exact scope (see Filter Discovery Commands)".
- *No `filter-queues` activation row* — added row 029: "Show me which queues have been active recently."
- *Smoke task excludes queues* — this reviewed an earlier commit. The current task grades all four groups: `filter-queues` is in the prompt and has its own `command_executed` criterion, and `expected_turns` is 10.

**Manual-review questions the review raised**

- *Pagination semantics* — client-side, confirmed against the cli#3449 implementation: the backends return unpaginated arrays, and the CLI dedupes, sorts, then slices with `--limit`/`--offset`. So `HasMore: false` does mean the caller has seen every row the backend returned for the 30-day window.
- *`filter-queues` response shape* — `{QueueName, FolderKey}` is source-verified against the cli#3449 wrappers and its e2e suite. I'll re-check all four shapes against live `--output json` output in the same post-merge run that produces the passing-run claim.

**Task lint**

- *High, verb reachability and no passing run* — acknowledged; both resolve when cli#3449 merges and the catalog refreshes. I'll run the task, add the passing-run claim to the body, and re-run `check-cli-verbs.py` then. This PR should not merge before that.
- *Low, add a `run_command`/`json_check` shape criterion* — deliberately not adding for smoke tier: it would re-run live commands during grading (doubling live calls on the shared tenant), the sibling smoke tasks don't validate output shape either, and shape validation already lives in the cli repo's e2e suite (S1-S8). Happy to revisit as a separate integration-tier task if reviewers want graded shape coverage here.

tests/tasks/uipath-insights/filters/smoke.yaml
```diff
@@ -0,0 +1,81 @@
...
+    description: "Agent activated the uipath-insights skill"
+    skill_name: "uipath-insights"
+    expected_skill: "uipath-insights"
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent ran filter-folders list to discover folders"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+insights\s+filter-folders\s+list'
```
rliuup, line 35: issue (non-blocking): This task already passed at weighted 1.00 on this head (smoke run reported `7/7 passed`, `status=SUCCESS`) while cli#3449 is still open, because `command_executed` matches the command string and not its exit code — so all four criteria pass even when `uip insights filter-*` does not exist yet. Either soften the Limitations claim that a passing run is impossible before that merge, or add a `run_command` criterion (`uip insights filter-folders list --output json`, `expected_exit_code: 0`) so the task actually proves the CLI path.

skills/uipath-insights/SKILL.md
```diff
@@ -163,6 +164,45 @@ uip insights jobs failure-details --time-range 1440 --output json
...
 ---
 
+## Filter Discovery Commands
+
+Four discovery groups expose scope candidates with recent Insights activity:
+
+```bash
+uip insights filter-folders list --output json     # {FolderName, FolderKey} rows
+uip insights filter-processes list --output json   # {ProcessName, FolderKey} rows
+uip insights filter-queues list --output json      # {QueueName, FolderKey} rows
```
rliuup: suggestion (non-blocking): Now that `uip insights filter-queues list` is a taught command, the existing negative criterion in `tests/tasks/uipath-insights/smoke_critical_rules.yaml` (`command_pattern: 'uip\s+insights.*queue'`) can false-fail a weight-2.0 assertion if the agent runs filter-queues while explaining the queue-metrics boundary. Narrowing it to the actual intent — e.g. `uip\s+insights\s+jobs[^&;|\n]*queue` — keeps that guard meaningful.

skills/uipath-insights/SKILL.md
```diff
@@ -1,15 +1,15 @@
...
+when_to_use: "User says 'job failures', 'automation health', 'job success rate', 'processing time', 'which processes fail the most', 'failure reasons', 'job trends', 'how many jobs ran', 'insights dashboard', 'job metrics', 'job KPIs', 'job performance', 'uncompleted jobs', 'pending jobs', 'faulted jobs', 'job timeline', 'process details', 'which folders can you see', 'what processes are active', 'find the folder key', 'discover queues', 'which machines reported', 'filter-folders', 'filter-processes'. Also 'uip insights', 'insights jobs'. NOT for starting/stopping jobs (uipath-platform), NOT for root-cause debugging of a specific job error (uipath-troubleshoot), NOT for queue item metrics (queue discovery is supported via filter-queues; metrics are not)."
```
rliuup, line 4: suggestion (non-blocking): The new discovery phrases overlap the uipath-platform activation set (`uipath-platform-009` folders, `-032` processes under a folder, `-018` queue items), and the activation gate only re-runs skills whose own `SKILL.md` changed, so an activation lost there would only surface later on an unrelated uipath-platform PR. Worth running the uipath-platform activation dataset once before merge.

tests/tasks/uipath-insights/filters/smoke.yaml
```diff
@@ -0,0 +1,81 @@
...
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Agent did not pass time flags to filter commands"
+    tool_name: "Bash"
+    # [^&;|\n]* instead of .* : the grader compiles patterns with re.DOTALL,
+    # so .* would bleed across command separators and false-fail a batched
+    # call like `filter-folders list ... && jobs summary --time-range ...`
+    command_pattern: 'uip\s+insights\s+filter-\S+\s+list[^&;|\n]*--(time-range|started-after|started-before)'
```
rliuup: suggestion (non-blocking): `[^&;|\n]*` also excludes newlines, so a backslash-continued invocation (`uip insights filter-folders list \` then `--time-range 60`) slips past this guard — the exact mistake it exists to catch, and the jobs examples in `SKILL.md` use that continuation style. `(?:[^&;|\n]|\\\n)*` closes it without reintroducing the cross-command bleed.

tests/tasks/uipath-insights/filters/smoke.yaml
```diff
@@ -0,0 +1,81 @@
+task_id: skill-insights-smoke-filter-discovery
+description: >
+  Smoke test: agent uses the uipath-insights skill to discover valid
+  scope candidates before acting. Given a prompt about finding folders,
+  processes, queues, and machines to scope an investigation, the agent
+  should run all four filter discovery commands instead of guessing
+  identifiers, and treat an empty result as normal rather than an
+  error. The CLI is authenticated; the tenant may or may not have
+  recent activity, and both are valid outcomes.
+tags: [uipath-insights, smoke, mode:build, lifecycle:discover]
```
rliuup: nit (non-blocking): `mode:build` does not match the taxonomy in `tests/README.md` (build = creating/designing/editing/deploying), and the five sibling uipath-insights tasks all use `mode:diagnose`; since the tag drives `make tags` and the evalboard, a read-only discovery task lands in the wrong bucket.

skills/uipath-insights/SKILL.md
```diff
@@ -1,15 +1,15 @@
...
 allowed-tools: Bash, Read
 ---
 
 # UiPath Insights — Job Monitoring Agent Skill
 
 Insights provides analytics and monitoring for UiPath automation execution. This skill covers **job monitoring** — querying aggregated job execution data for dashboards, health checks, and failure investigation.
 
-All operations go through `uip insights jobs <subcommand> --output json`.
+Job monitoring goes through `uip insights jobs <subcommand> --output json`; scope discovery goes through the `uip insights filter-*` groups.
```
rliuup: nit (non-blocking): `uip insights filter-*` stays soft-stale in the verb gate even after a catalog refresh, since `filter-*` is never a real verb path. An inline `<!-- uip-check-skip -->` on this line clears the recurring warning.

rliuup (approved):
LGTM with nits.

suranabhavya (approved):
Ruilin's review covered the substantive points, fixing that post LGTM

rohitjain-uipath (approved):
+1 on Ruilin's comments. Lets fix them before merge, rest LGTM!

tests/tasks/uipath-insights/filters/smoke.yaml
celestinryf: Added a run_command gate in 709658dab: `uip insights filter-folders list --output json` must exit 0 at grading time, so the task can only pass against a CLI that actually ships the verbs. That also makes the Limitations claim accurate again. Smoke CI on this PR stays red until cli#3449 merges, which matches the intent that this PR not merge first.

skills/uipath-insights/SKILL.md
celestinryf: Narrowed in 709658dab to `uip\s+insights\s+jobs[^&;|\n]*queue` as suggested. Gave the adjacent start/Invoice_Processing negative the same treatment: its `.*start` also matched `--started-after`, and it had the same separator bleed.

tests/tasks/uipath-insights/filters/smoke.yaml
celestinryf: Retagged `mode:diagnose` in 709658dab.

skills/uipath-insights/SKILL.md
celestinryf: Added in 709658dab. `check-skill-verbs.py` no longer reports the `filter-*` glob on this file; the four concrete filter-folders/processes/queues/machines findings remain until the catalog picks up cli#3449.

tests/tasks/uipath-insights/filters/smoke.yaml
celestinryf: Fixed in 709658dab with `(?:[^&;|\n]|\\\n)*` as suggested; the comment above the pattern now documents both exclusions. Applied the same form to the e2e patterns in #2551 as well, since those are positive criteria and a bare `[^&;|\n]*` would dock the continuation style the skill examples teach.

skills/uipath-insights/SKILL.md
celestinryf: Ran the uipath-platform activation gate on this head locally (coder-eval 0.9.5, `activation_gate.py --skill uipath-platform`, 2026-08-11, full 136-row positive set at max_turns 1). Result: PASS, recall.yes 90.4% against the 90% floor, completion rate 100%. Every one of the 13 misses is a no-engagement miss; none of their transcripts mention uipath-insights, so the new discovery phrases are not taking uipath-platform activations. The three rows named here (009, 018, 032) all scored 1.0. One caveat: I did not run a control on main, so I cannot say how much of the gap to the 2026-06-17 baseline is drift.

celestinryf:
Smoke skill tests not running because [CLI repo PR](https://github.com/UiPath/cli/pull/3449) is not merged yet, so no commands for skills to test against. This PR will only go through after the filters are in on CLI and this smoke test is passing.

skills/uipath-insights/references/investigation-playbook-guide.md
```diff
@@ -106,9 +106,12 @@ uip insights jobs summary \
...
 
 ```bash
-# Step 1: Find the folder key (requires uipath-platform)
+# Step 1: Find the folder key — Insights-active folders first
+uip insights filter-folders list --output json
+# Look for the folder's FolderKey in the output.
+# Fallback for folders without recent activity (requires uipath-platform):
 uip or folders list --output json
-# Look for the folder's Key (GUID) in the output
+# There the GUID is the folder's Key field (not FolderKey)
```
rockymadden: This is confusing. Are both of these present then?

skills/uipath-insights/references/investigation-playbook-guide.md
```diff
@@ -141,5 +144,5 @@ This means:
 | User wants to start/stop/restart a specific job | `uipath-platform` (`uip or jobs start`) |
 | User wants to read the logs of a failed job | `uipath-platform` (`uip or jobs logs`) |
 | User wants to debug why a specific job error happened | `uipath-troubleshoot` |
-| User wants to find a folder key to filter by | `uipath-platform` (`uip or folders list`) |
+| User wants to find a folder key to filter by | `uip insights filter-folders list` (Insights-active folders), or `uipath-platform` (`uip or folders list`) for the full folder inventory |
```
rockymadden, line 146: I'm confused, why is this not done via the `--filter` flag and instead as a sub command?

skills/uipath-insights/references/investigation-playbook-guide.md
celestinryf: There are 2 sources here for the same key. 

- filter-folders list, which comes from Insights, returns folderKey. 
- uip or folders list, from Orchestrator, returns Key
- Both of these return the same GUID for the same folder.

Insights lists active folders with data in a 30-day window, and is part of the same discovery family as filter-processes/queues/machines. The Orchestrator command stays as the fallback for folders without recent activity. Reworded the block so the two-source relationship is explicit.

skills/uipath-insights/references/investigation-playbook-guide.md
celestinryf: Other teams use --filter, but it does not make sense here. The existing --filter <odata> on commands like uip maestro process list and uip case process list forwards an OData expression to Orchestrator-style APIs that accept arbitrary filter expressions. 

The Insights endpoints don't: the jobs endpoints take a few fixed scope params, exposed as dedicated flags like --folder-key and --time-range, and the Filters/* endpoints take no filter input at all, they just return the set of values that have data. So there was no expression to forward and no list command for the flag to hang off. The flow is filter-folders list to discover the key, then jobs summary --folder-key <guid> to apply it.

If you'd like to view my CLI PR for these changes they are here: [CLI PR 3449 Insights Filters](https://github.com/UiPath/cli/pull/3449)

celestinryf:
Bookkeeping on field-name casing, so the commit history reads clearly. `28ac3e924` rewrote the documented filter row fields to camelCase based on the projection source, and [`817bf904`](https://github.com/UiPath/skills/commit/817bf9040c29d1a24d43b55a70dffe4978dff1e8) restores PascalCase. The CLI PascalCases every `Data` key on output (`OutputFormatter.success` in `packages/common/src/formatter.ts`, pinned by `filter-discovery.e2e.test.ts`), so `{FolderName, FolderKey}` is what agents see. Same finding as rliuup's casing comment on #2607.

rockymadden (approved):

PR #2556, opened 2026-08-10
docs(maestro-flow): inline-agent vs resource source identity — never transpose the UUIDs

jiyangzh:
## Summary

A nightly resolution build (2026-08-10) faulted `[170002] Failure in the Orchestrator Job` / `Serverless.PythonAgent.PrepareEnvironmentError` nine minutes into debug, after a clean validate. Root cause: the agent **transposed two adjacent opaque UUIDs** — the inline-agent node's `inputs.source` pointed at its context-index *resource* UUID, and the index node pointed at the agent *project* id. Everything else in the build was byte-correct. One-change-proven: swapping the two values back produced a full live-debug pass (SOP-grounded output).

`flow validate` scores the swap as Valid — the agent checks silently skip when `<source>` contains no `agent.json`. (The durable fix is a validator rule that errors instead of skipping; queued for the post-#3420 validator batch. This PR is the bridge + the triage map.)

## Changes (+25 lines, two on-demand reference files — no SKILL.md, no tasks)

1. `author/…/inline-agent/impl.md` — 3-line rule at the point of authoring: agent `source` = the project id (the dir containing `agent.json`); resource `source` = the resource UUID (nested under `resources/<uuid>/`); never transpose — validate won't catch it. Plus a Debug-table row keyed on the exact runtime signature.
2. `diagnose/…/failure-modes.md` — full symptom → cause → fix pattern for the `[170002]`/`PrepareEnvironmentError` signature, indexed, listed under "not caught by validate," cross-linked to the authoring rule.

## Testing

Docs-only. Anchors and relative links verified; `check-skill-status.py` and skill-description hook green; no frontmatter touched.


jiyangzh:
one time error, may not worth a doc fix. Table it for now.

Issue #101, opened 2026-03-28
uipath-rpa-workflows: InvokeCode Language property uses undocumented NetLanguage enum (VBNet/CSharp), not project expressionLanguage values


roalexandru:
## Summary

The `uipath-rpa-workflows` skill has no documentation for the `InvokeCode` activity's `Language` property, which uses the `UiPath.Core.Activities.NetLanguage` enum with values **`VBNet`** and **`CSharp`**. This is different from the project-level `expressionLanguage` field in `project.json` which uses `VisualBasic` and `CSharp`. The mismatch causes a runtime XAML deserialization error when generating InvokeCode activities.

## Error

When `Language="VisualBasic"` (or `Language="VB"`) is used on an InvokeCode activity:

```
Failed to create a 'Language' from the text 'VisualBasic'.
System.FormatException: VisualBasic is not a valid value for NetLanguage.
System.ArgumentException: Requested value 'VisualBasic' was not found.
```

## Root cause

The `NetLanguage` enum (confirmed via reflection on `UiPath.System.Activities.dll`) has exactly two values:
- `VBNet`
- `CSharp`

The skill's documentation only references the project-level `expressionLanguage` field (`VisualBasic` / `CSharp`) and never mentions the `NetLanguage` enum or InvokeCode's `Language` property.

## What's missing in the skill

1. **No InvokeCode documentation at all** — The skill references `InvokeCode.md` at `references/activity-docs/UiPath.System.Activities/{version}/activities/InvokeCode.md`, but the entire `references/activity-docs/` directory does not exist. There are no bundled activity docs for any package.

2. **No mention of `NetLanguage` enum** — Neither `common-pitfalls.md`, `xaml-basics-and-rules.md`, nor any other reference file documents the `Language` property or its valid values. The skill's guidance to "check `expressionLanguage` in `project.json`" naturally leads to using `VisualBasic` — which fails.

3. **No InvokeCode XAML example** — `xaml-basics-and-rules.md` provides three XAML reference examples (basic activities, connector activity, IS connector activity) but none include InvokeCode, which is explicitly recommended as "a pragmatic fallback when dedicated activities have unresolvable issues."

## Suggested fix

1. Add a section to `common-pitfalls.md`:
   ```
   ### InvokeCode Language Property Uses Different Enum Than Project Expression Language
   
   The `Language` property on InvokeCode accepts `UiPath.Core.Activities.NetLanguage` enum values:
   - `VBNet` (not `VisualBasic`)
[...]

PR #2557, opened 2026-08-10
Align Maestro Flow tasks for same-ground evaluation

tmatup:
## Summary

- align the nine path-to-GA Maestro Flow tasks on one loop-neutral prompt, criterion, weight, limit, and task-id contract
- add five report-only structural advisories with zero pass thresholds
- add a campaign-only overlay for the pinned image, Sonnet 5 route, shared system prompt, environment surface, Proactive output style, isolated auth home, and an empty read-only catalog shadow
- install the stable CLI from public npm before its private tool plugins; CI and campaign builds deliver the npm credential only through a BuildKit secret, while the external nightly retains a deprecated build-arg fallback during migration

## Nightly-signal rationale

The task files intentionally change the future nightly signal:

- the invoice-extraction task loses 5.5 criterion-weight points that were outside the approved intersection
- tasks 2–6 no longer forbid the agent from running debug itself, so the skill owns its verification loop
- run limits use the predeclared shared envelopes rather than the previous task-local limits

These are measurement-contract changes, not greening changes. The committed nightly experiment is untouched; all campaign-specific harness deltas live in the new overlay.

## Validation

- same-ground sync gate: 9/9 pairs plus shared support 1/1
- new advisory regression suite: 11/11; all five reference flows accepted
- shared Flow checker tests: 170/170
- skill composition validation: 26 skills, 1,709 files per flavor
- all nine changed tasks executed in branch shakeout; every observed red retained and classified, with no run-driven prompt, checker, weight, or limit change
- final pinned-image pagination canary before review fixes: 5/5 criteria, score 1.000; Sonnet 5, Proactive, auth, plugin delivery, pagination, and catalog isolation green
- toolchain pins: Claude Code 2.1.177, coder-eval 0.9.5, UiPath CLI 1.198.0, 29 tool plugins
- secure-build empty-secret guard exercised locally; current-head CI exercises the real secret path
- prompt audit: no resource GUIDs or task-specific coaching
- Bai, Tomasz, and Claude findings answered point by point on the PR

🤖 Generated with Claude Code


tmatup:
Review triage on the current head:

- **Medium — campaign overlay mount:** no one-arm edit. coder-eval forwards the host `HOME` into Docker, while `extra_mounts` expands variables only on the source side. On this pinned runner, changing the destination to `/.uipath` would make the CLI miss its auth cache. The overlay is runner-specific by design; the underlying portability gap is filed as [coder_eval #100](https://github.com/UiPath/coder_eval/issues/100).
- **Medium — validate-only e2e tasks:** expected under the approved intersection contract. Adding runtime/debug scoring criteria would change the measurement contract; no checker or criterion change.
- **Low — quoted prompt scalar:** the decoded prompt bytes are the contract and are intentionally unchanged. The YAML representation does not alter the agent input; no functional edit.
- **Passing-run evidence:** the PR body deliberately claims only the completed pagination canary. It will be updated only after the full nine-task shakeout is actually complete.
- **Docker default `dev` train:** the first CI run exposed a separate registry-branch regression. Current head `3d69f446f` keeps stable numeric releases on public npm and routes prerelease pins/dist-tags through the authenticated package feed.

No severity was skipped; the only code-level CI finding is fixed on the current head.

tmatup:
@claude review

tmatup:
@bai-uipath @rockymadden The path-to-GA approval check requires an allowlisted approval on the current head. Could one of you please review and approve when available?

tmatup:
Current head `3979fc089` adds the symmetric read-only catalog shadow after a branch canary proved that `uip skills list` could repopulate the auth cache. Direct probes on both pinned arm images retain authenticated `uip` access, make the extra catalog fail closed, and leave the shadow empty.

@claude review

🤖 Generated with Claude Code

tmatup:
@bai-uipath @rockymadden The allowlisted path-to-GA approval is still the only policy gate. The latest head is `3979fc089`; please review and approve this head when available.

🤖 Generated with Claude Code

tmatup:
Current head `5b8584be5` merges the latest skills main before final shakeout. The integrated task pair still passes the same-ground gate 9/9 and the shared checker suite 68/68.

@claude review

🤖 Generated with Claude Code

tmatup:
@bai-uipath @rockymadden The allowlisted path-to-GA approval is still required; the integration head is now `5b8584be5`. Please review and approve this head when available.

🤖 Generated with Claude Code

tmatup:
Claude review triage (review completed on `3979fc089`; integration head is now `5b8584be5`):

- **Medium — `/home/tmatup` target:** no code change. coder-eval expands variables only on mount sources and forwards the host HOME into the container; `/.uipath` would make the pinned CLI miss auth. The portability defect remains tracked in coder-eval #100. Paired Docker canaries prove this runner-specific target, the nested read-only shadow, and auth all work on both arms.
- **Low — quoted prompt scalar:** no code change. The decoded prompt bytes are the cross-repo contract and pass the sync gate; changing representation provides no runtime benefit and risks byte drift.
- **Model and image manual checks:** verified by paired canaries: both record the fully qualified Bedrock model in agent config/environment, raw model_used is identical, `skills-image:sg1` launches, CLI 1.198.0 is present, and Proactive/plugin/auth delivery are green.
- **Connector-output and RPA-doc removals:** these arrived from current main and are not campaign-authored deltas. They remain owned by their already-merged changes; expanding this PR to revise them would violate the campaign scope fence.

Every severity and manual-review item is dispositioned; no prompt, criterion, checker, weight, or limit changed.

🤖 Generated with Claude Code

tmatup:
@claude review

tmatup:
Latest-head Claude review (`5b8584be5`) triage:

- **Medium runner path:** accepted only as the pinned-runner campaign overlay; coder-eval #100 tracks destination-variable support. Paired isolated canaries verify auth and the nested shadow.
- **Low scalar style:** no edit because decoded prompt bytes are the fairness contract and sync 9/9.
- **Manual model/image checks:** paired canaries verify `skills-image:sg1`, Proactive, CLI 1.198.0, fully qualified Bedrock routing, and identical raw model_used.

No new findings; every severity is dispositioned without a task/checker/limit change.

🤖 Generated with Claude Code

tmatup:
@uipreliga The path-to-GA check still needs one allowlisted approval on integrated head `5b8584be5`. Claude review and triage are complete; could you review and approve when available?

🤖 Generated with Claude Code

tmatup:
All nine Maestro Flow task shakeout rows have now executed on the current integrated branch stack. The latest head remains green on the same-ground gate (9/9), shared checker suite (68/68), latest-head Claude review, and all ordinary CI. All observed reds are retained and classified; no prompt, checker, weight, or limit was changed from run feedback.

`path-to-ga approval` is the sole merge gate. @bai-uipath @rockymadden @uipreliga — please provide the allowlisted approval when available.

🤖 Generated with Claude Code

tmatup:
@claude review

Please review the latest head. The only post-review change is a merge of current main touching the disjoint Maestro Case test surface; review all severities.

🤖 Generated with Claude Code

tmatup:
Latest-head Claude review (\`a1ff8aca8\`) triage:

- **Low — runner-specific auth mount:** accepted only in this campaign overlay. coder-eval #100 tracks portable destination expansion; paired isolated canaries verify the pinned path, auth, and read-only catalog shadow on both arms.
- **Low — quoted Drive prompt scalar:** no edit. The decoded prompt bytes are the fairness contract and the 9/9 sync gate proves them equal; representation has no runtime effect.
- **Pre-existing mode tags:** outside this PR and not a regression; no campaign expansion.
- **Manual model/image checks:** paired canaries verify the fully qualified Bedrock route, identical raw model_used, Proactive, skills-image:sg1, Claude Code 2.1.177, coder-eval 0.9.5, and uip 1.198.0.
- **Fixture values:** live tenant probes and the complete shakeout verify MCS-2026-04872, 1610, and Enterprise against the current seeded resources.

Every severity and manual-review item is dispositioned. No prompt, criterion, checker, weight, or limit changed. The allowlisted path-to-GA approval remains the only merge policy gate; @bai-uipath @rockymadden @uipreliga, please approve the current head when available.

🤖 Generated with Claude Code

tmatup:
@sherifmak @uisherif The path-to-GA check needs one allowlisted approval on current head \`a1ff8aca8\`. The latest Claude review and all severity triage are complete; all nine tasks have branch shakeout evidence and the final pinned-image canary is 5/5. Please review and approve when available.

🤖 Generated with Claude Code

tmatup:
@claude review

dmetzgar (approved):

bai-uipath (commented):
**Two blockers before merge: a cross-repo change, and timing evidence for the debug-loop change.**

- **Coordinate the secret change with coder_eval_uipath.** The nightly builds `skills-image:latest` from this Dockerfile via `eval_runner/scripts/ci/daily.sh`, which still passes the token as a `--build-arg` and doesn't set `DOCKER_BUILDKIT`. Once `required=true` lands that build fails and the script's `set -euo pipefail` aborts the whole nightly, ADO included. Fix: land the daily.sh change first and let one nightly prove it, since daily.sh self-updates a run late.

- **Debug in the agent loop.** The five billing tasks lose "Do not run debug yourself" while their checkers still run debug, and those tasks already permit up to 90 minutes. Debug is the flakiest surface in the suite, so a bad platform night now drains the full budget instead of failing fast in the checker. Please verify before merge: a timed run of the five with the guard removed, wall clock and turns against baseline. If the delta is material, move the prompt change to the campaign overlay.

- **Verify the converted tasks still behave.** All nine were re-serialized wholesale and the drive-to-slack checker lost its debug step. The sync gate proves the arms match each other, not that these tasks still run as before.

Minor: add a comment on the `/home/tmatup` mount naming the coder_eval #100 constraint.

bai-uipath (approved):

uipreliga:
## Code review

Three reviewers (Gemini 3, Codex, Claude Opus) plus local reproduction. **Bold = independently flagged by 2+ reviewers.** ✅ = reproduced locally.

### Critical / High

**1. ✅ Two of the five new advisory checkers fail the repo's own reference flows.** Ran all five against `*.reference.flow`:

| checker | reference flow |
|---|---|
| `advisory_billing_dispute_analyst.py` | ❌ `FAIL: the context node's inputs.indexId is None` |
| `advisory_billing_resolution_writer.py` | ❌ `FAIL: neither prompt references a flow value ($vars.…)` |
| `advisory_billing_invoice_lookup.py` | ✅ pass |
| `advisory_billing_discrepancy_detector.py` | ✅ pass |
| `advisory_billing_dispute_resolution.py` | ✅ pass |

The analyst reference context node carries only `inputs.source` — the uuid is in the node *type*, exactly as the script's own docstring says (`advisory_billing_dispute_analyst.py:20-22`), yet lines 86-89 additionally demand `inputs.indexId`/`indexName`. The writer/analyst reference agents bind via `agentInputVariables[].binding` with prompt text in the sidecar `agent.json`, not via `$vars.` in `systemPrompt`/`userPrompt`.

Either the checkers' model of an inline agent is wrong, or the reference fixtures aren't valid exemplars (`agentInputVariables: []` suggests they may be minimal stubs). Worth resolving either way: as `pass_threshold: 0.0` criteria these can't score anything wrong, but a constant FAIL on both arms carries zero discriminating signal — which defeats the point of adding them.

**2. ✅ `check_bindings_no_stubs.py:57` — `all()` should almost certainly be `any()`; the check is near-inert.** Reproduced:

```
resourceKey "DataFabricConn" + default "00000000-0000-0000-0000-000000000001" → not flagged
key "conn-symbolic" + defaultValue "<TODO>"                                    → not flagged
```

Only an entry that is stub in *every* field fires. The exact defect the criterion advertises — a symbolic name that never got substituted for a tenant key — slips through. Note `advisory_billing_dispute_resolution.py:71` uses `is_real_key(resourceKey) or is_real_key(default)`, which *would* catch it: two checkers in the same PR disagree on what "a real key" means.

**3. ✅ Forbidden-literal scans false-positive on UUIDs.** All three reviewers. Reproduced against `{"connectionId":"3f2a91cc-1610-4b7e-…","folder":"aa11bb22-4200-…"}` — both `advisory_billing_discrepancy_detector.py:182` (`(?<![0-9A-Za-z])`) and `advisory_billing_resolution_writer.py:151` (`(?<![0-9])`) match, because `-` passes both lookarounds. Second trigger: canvas geometry — `"position":{"x":1610}` matches too, and agent-authored 8–10 node flows reach that range. `Enterprise` also matches inside `"Enterprise-tier customers"` in a description.

> ⚠️ One reviewer proposed `\b…\b` as the fix — that doesn't work (`\b1610\b` still matches `-1610-`). The scan needs to be scoped to expression/literal values rather than raw file text.

**4. Multi-End-node bindings collapse to last-writer-wins.** All three reviewers. `bound[k] = str(unwrap(v))` looped over all end nodes in `advisory_billing_{invoice_lookup,resolution_writer,dispute_analyst,discrepancy_detector}.py`. A survey of `tests/runs/` found 42 of 137 generated flows have >1 `core.control.end`, routinely binding the same key at each. A success End binding `$vars.erpQuery.output…` plus an error End binding `"not found"` grades differently depending on their order in `nodes[]` — false FAIL one way, silent false PASS the other. `advisory_billing_dispute_resolution.py:592-605` handles this correctly, so the inconsistency is within the PR.

### Medium

**5. `tests/experiments/same-ground-headtohead.yaml:13` pins `skills-image:sg1`; CI only ever builds `skills-image:latest`.** All three reviewers. `sg1`, `SG_UIPATH_HOME` and `SG_EMPTY_SKILLS` appear nowhere else in the repo — the experiment is unrunnable from a clean checkout and isn't in the `tests/README.md` experiment table.

**6. Secrets forwarded into a sandbox where the agent has `Bash`.** `env_passthrough_extra` includes `NODE_AUTH_TOKEN` (the GitHub Packages credential), `ANTHROPIC_API_KEY`, `AWS_BEARER_TOKEN_BEDROCK`, `CODEX_API_KEY`, `UIPATH_CLI_AUTH_TOKEN`. `nightly.yaml` does not pass `NODE_AUTH_TOKEN`. This cuts against this branch's own `3a90148` "isolate campaign auth" and against the Dockerfile change in this same PR, whose entire point is keeping that credential out of reach. Separately, `${SG_UIPATH_HOME}` is bind-mounted **rw** onto `/home/tmatup/.uipath` with no `:?` guard — unset degrades the mount spec to `:/home/tmatup/.uipath:rw`. `tmatup` is a hardcoded container username.

**7. `check_inline_agent.py:44-45` — the `**` widening outgrew its exclusion filter.** ✅ Confirmed `**/agent.json` + `recursive=True` does match root-level (the stated intent works), but: the filter tests `"/.agent-builder/" not in p`, which misses a *top-level* `.agent-builder/agent.json`; and since `sorted(paths)[0]` picks the lexicographic first, dot-directories sort ahead of everything. The sibling `.flow` finds in the same YAMLs explicitly prune `_outputs`, `.v1stage`, `v1stage` — this glob prunes none of them, so a stale staged `agent.json` can win over the real one. The old `*/*/*/agent.json` couldn't produce that path.

**8. `check_bindings_no_stubs.py:67` — `EXCLUDED_PARTS.intersection(path.parts)` on absolute paths.** All three reviewers. `Path.cwd().rglob()` yields absolute paths, so an ancestor directory named `example`/`sdk`/`fixtures`/`references` filters out every candidate and the script aborts with "no generated bindings.json found". Needs `path.relative_to(cwd).parts`.

**9. `test -f "$FLOW"` in the five new advisory criteria is a no-op.** No `set -e` in the YAML `command:` block, so a failed `test` doesn't stop the script; empty `$FLOW` reaches Python as `""` and produces a `FileNotFoundError` traceback instead of the intended clean `FAIL:` line. Two matches make `$FLOW` multi-line and the checker gets the first path by word-splitting accident.

**10. Information loss in the wholesale YAML reformat.** `ixp/e2e_01_invoice_extraction_greenfield.yaml` silently drops three graded criteria — `skill_triggered` (2.0) and both registry-search `command_executed` (2.0 + 1.5); graded weight 16.0 → 10.5. Skill-activation coverage across the flow suite drops 6 → 5 tasks. That file is also consumed by `nightly.yaml` and the smoke gate, so the loss isn't scoped to the campaign. Also deleted: the comment recording why the SharePoint pattern is case-insensitive (an agent had previously been failed by the case-sensitive version) and the `run_limits` justification citing RE-12529 and the measured 882s/71-turn baseline — precisely the "why" a future editor needs before re-tuning caps.

Separately, the reformat output is PyYAML `default_flow_style=False`, not the repo's own style: `scripts/prune-task-defaults.py:35-37` pins `sequence=4, offset=2` with the comment *"Match the repo's 2-space indented list style: `  - key:` not `- key:`"* — the new files use exactly the rejected form. All five files still parse cleanly.

**11. `advisory_billing_dispute_resolution.py:90` only understands the v1 `bindings[]` schema**, while `check_bindings_no_stubs.py` added in the same PR supports both v1 and the `{"version":"2.0","resources":[…]}` form (per `f453d57` "accept native flow bindings schemas"). A project emitting `bindings_v2.json` has no top-level `flow["bindings"]`, so this advisory unconditionally fails.

**12. BuildKit isn't made explicit.** `--secret` + `# syntax=docker/dockerfile:1.7` require it; neither workflow sets `DOCKER_BUILDKIT=1` or uses buildx. Modern GH runners default to BuildKit (so it works today); worth pinning anyway. Sharper point: this is a **breaking build-interface change** — any caller still passing `--build-arg NPM_AUTH_TOKEN=…` now fails, including the nightly cron `daily.sh` that `run-coder-eval.yml:206` explicitly claims to match and that this PR can't update.

**13. "read from step X" logic follows only one hop.** `via` is the set of nodes whose inputs mention `$vars.<producer>.output`; a two-hop chain (`erpQuery → compute → merge → format → End`) puts `format` outside `via` and fails a correct flow. The detector's own docstring says the numbers are *"computed downstream of ERP"* — the shape it can't follow. Also `any(f"$vars.{s}" in src …)` is a bare substring test: `$vars.script1` matches `$vars.script10`.

### Low

- ✅ Dead `by_id` in `advisory_billing_discrepancy_detector.py:74` and `advisory_billing_dispute_resolution.py:87` (all three reviewers).
- `advisory_billing_dispute_analyst.py:161` — first clause `re.match(r'^"[^"]*"$', …)` is fully subsumed by the `"$vars." not in s` clause; an *empty* binding passes both.
- `advisory_billing_invoice_lookup.py:113` — `RAW_INPUTS` entry `"  MCS-2026-04872 "` is unreachable (the `CANONICAL in raw` check fires first); `json.dumps(bad)[1:-1]` is an identity transform for all three values.
- `expr_of` docstring claims a `'string'` kind the function never returns (`'other'`).
- All five scripts' `Usage:` lines still name `check_*.py`, not `advisory_*.py`. `f["nodes"]` is unguarded → `KeyError` traceback instead of the `FAIL:` convention.
- `advisory_billing_dispute_resolution.py:42` cites `_lib/seed_bindings.py` as the shared definition of "not a real key" — that file doesn't exist in the repo.
- `advisory_billing_dispute_analyst.py:36` hardcodes a live tenant index uuid with no note on reseeding.
- `Dockerfile:112` — `LABEL org.uipath.cli.version="${CLI_VERSION}"` records the *dist-tag* (`dev`) in the default case, not a resolved version, so it's weak provenance.
- `same-ground-headtohead.yaml:63` — a "head-to-head" with a single `variant_id: default`; the repo's A/B convention (`skill-comparison-template.yaml`) uses ≥2 variants each pinning a plugin path.

### DRY / over-engineering

One reviewer argued the duplication across the five `advisory_*.py` is acceptable for isolated graded scripts; the counter-argument is more persuasive here: `fail()` ×5, `unwrap()` ×4, the globals→ins/outs extraction ×4, the End-node `bound` loop ×4 (which is how bug #4 got copied into four files), the reader-set computation ×3, uuid/stub regexes ×3. `_shared/` already has `__init__.py` and `flow_check.py`, and sibling import from a standalone script works — it's what `check_drive_to_slack.py` did before this PR removed it. Extracting six helpers would make #4 a one-line fix and would have surfaced the schema disagreement in #11 at authoring time.

Also: the 25–35 line docstrings arguing against "v1", and failure messages embedding unresolvable campaign identifiers (`G-19`, `FC419`, `MST-9265`, `ga7`, "card IA"), are a large maintenance surface for criteria pinned at zero grading weight.

One scoring note worth a decision: `weight: 1.0` + `pass_threshold: 0.0` awards full weight unconditionally. `billing_discrepancy_detector` goes 8.0 → 10.0 graded points with 2.0 free, so a flow that only passes `validate` scores 50% instead of 37.5%. Both arms shift equally so the head-to-head is fair, but absolute pass rates stop being comparable to historical runs.

### Positive observations

- Moving the GHP token from `--build-arg` to a BuildKit secret is the right fix, and the Dockerfile comment correctly explains *why* (`ARG` persists in build output and image history even after the npmrc is deleted). `required=true` plus the `-z` guard keeps the empty-secret case loud.
- ✅ The public-npm-vs-GHP branch is correctly gated: `[0-9]*-*` → prerelease → GHP, `[0-9]*` → stable → public npm, `*)` → GHP, so the `ARG CLI_VERSION=dev` default preserves existing CI behavior exactly. Two reviewers checked this as a suspected bug and refuted it; it's a DRY opportunity only.
- `check_drive_to_slack.py` dropping the unused `run_debug` import **and** correcting the criterion description from *"…and debug step finished green"* — the script never called `run_debug`, so the description was untrue on `main`.
- The mutual-unreachability test in `advisory_billing_discrepancy_detector.py:130-162` is genuinely the right idea and the one thing a node-type assertion can't express. The `reaches(…, block=mid)` formulation was examined by two reviewers and found correct; the chained-through-merge escape is caught by the fork and ≥2-sources checks.
- `check_inline_agent.py`'s three assertions (model off the scaffold default, real prompt, typed output schema) are well-chosen and can't be faked from agent monologue.

### Out of scope

One reviewer flagged a shell-injection risk at `run-coder-eval.yml:269-281` (unquoted `$agent_flags` expansion). That code is untouched by this PR and `AGENT_MODEL` is a `workflow_dispatch` input — real, but pre-existing and not this PR's to fix.

### Overall

The infrastructure half (BuildKit secret, dual-registry CLI install) is solid and well-reasoned. The checker half looks not-yet-ready: two of five new advisory scripts fail the repo's own reference artifacts, the new bindings checker is near-inert because of `all()` vs `any()`, and the literal scans will false-fail on ordinary UUIDs and canvas coordinates. The strongest merge gate would be a precondition that every new checker passes its `.reference.flow` — that alone catches findings 1 and 3.

---

uipreliga (approved):
Fix what you agree with and 🚢

tmatup:
@bai-uipath @uipreliga Thanks for the detailed review. I pushed the fixes in `f4e1f3a18` and the format-only cleanup in `a5c4ba7cd`.

### Fixed

- Added an 11-case regression suite. All five advisories now pass their committed reference flows.
- Centralized flow discovery, semantic literal scanning, UUID checks, multi-End success bindings, multi-hop provenance, sidecar prompts, and both binding schemas in one shared helper. This removes the copied last-writer-wins and substring bugs.
- Made the bindings checker strict: every material connection value must be a real non-stub UUID; exclusions are evaluated relative to the run root.
- Excluded top-level and stale staging trees from inline-agent discovery.
- Replaced the five shell `find`/`test -f` wrappers with direct, exact flow discovery.
- Removed the hard-coded analyst index UUID, fixed usage/error handling/dead fields, and restored the IxP limit and SharePoint rationale comments.
- Restored all nine task YAMLs to the repository list style with parsed YAML equality verified.
- Documented the campaign-only `sg1` launcher prerequisites and why the local experiment has one variant.
- Reduced the agent-visible environment to the Bedrock route plus task/plugin paths. npm, Anthropic, Codex, and UiPath token env vars are no longer forwarded; UiPath auth comes from the isolated cache mount. The `/home/tmatup` mount now names the coder_eval #100 HOME constraint.
- Reverted the BuildKit-only Docker interface. Existing workflow and nightly callers continue to pass `NPM_AUTH_TOKEN` as a build arg, so no coordinated external rollout is required. The weak CLI label was removed.

### Debug-loop and task evidence

The approved campaign contract requires removal of `Do not run debug yourself` from the five billing prompts, so I did not move it into a local overlay. I did run the five prompts without the guard on the pinned branch stack:

| task | result | turns | tools | wall | agent invoked flow debug |
|---|---:|---:|---:|---:|---:|
| invoice lookup | 1.000 | 137 | 76 | 1191.9s | no |
| discrepancy detector | 1.000 | 142 | 82 | 869.4s | no |
| dispute analyst | 1.000 decision | 85 | 53 | 505.5s | no |
| resolution writer | 1.000 decision | 91 | 49 | 646.2s | yes |
| full resolution | 1.000 | 193 | 114 | 2334.2s | no |

All completed inside the fixed envelopes. Four agents left debug to grading; the writer self-debugged successfully. The 2026-08-04 nightly is useful history but is not a causal baseline because it used a different model and task contract, so I am not claiming a like-for-like timing delta from it.

All nine converted tasks were exercised in the branch shakeout: eight were operationally green; the IxP task reached its fixed agent timeout after constructing the intended five nodes but before completing edges/script, classified as agent behavior with no prompt/checker/limit change. The Drive-to-Slack debug wording/import removal remains intentional: the checker never called debug, so the old description asserted a step that did not exist.

### Deliberately unchanged

- The removed IxP skill-trigger and registry-search criteria stay removed under the approved shared-output intersection contract; they grade an arm-specific loop rather than the common deliverable. The prompt and run envelope are unchanged by this review fix.
- Report-only advisories remain `weight: 1.0` / `pass_threshold: 0.0` per the approved contract. coder-eval #101 currently includes those weights in raw score; campaign decision scores exclude advisory rows and preserve raw scores separately.
- The single local variant is intentional because this file launches only this plugin arm; the README now states that explicitly.
- The pre-existing workflow-dispatch shell expansion is out of scope for this PR.

Verification now green: `11 passed`, Ruff clean, all five reference advisories pass, all five changed task plans load under coder-eval 0.9.5, and the nine-task contract gate passes. I have requested a fresh approval run on the new head.

🤖 Generated with Claude Code

tmatup:
@bai-uipath @uipreliga The review fixes are now on head `a5c4ba7cd`. The Path-to-GA gate reran automatically and its only failure is the expected current-head approval check: prior approvals point to `8f1979ff7`. Please re-approve this head when the fixes and justifications above look good; I will rerun the gate immediately afterward.

@claude review

🤖 Generated with Claude Code

tmatup:
Post-fix Claude review triage for head `a5c4ba7cd`:

- **Critical/High:** all closed; the review verified the shared helper, strict binding validation, structured literals, multi-End handling, multi-hop provenance, stale-stage exclusions, and 11-case regression suite.
- **Medium — campaign overlay prerequisites:** accepted/documented. This file is intentionally not a clean-checkout CI experiment. The launcher verifies both mount sources exist before exporting `SG_UIPATH_HOME` / `SG_EMPTY_SKILLS`; using shell `${VAR:?}` inside coder-eval YAML has no established expansion contract, so I am not introducing an unverified interpolation form.
- **Low — `expr_of` docstring:** no change; the current text already says the function returns `jsExpression`, `literal`, or `other`, exactly matching the implementation.
- **Low — analyst index UUID:** no hardcoded tenant UUID remains. The checker derives `index_id` from the generated node type and validates it dynamically; only the requested index name is constant.
- **Low — outdented SharePoint rationale comment:** no semantic change. It is valid YAML, the parsed-equality check passed, and the comment deliberately documents the following criterion. A comment-only paired-repository follow-up would add merge churn without changing the task contract.
- **Manual — debug timing:** satisfied in the preceding five-task table. All five completed inside the fixed envelopes; four did not self-debug, and the writer self-debugged successfully.
- **Manual — tests:** executed locally: 11/11 pass; Ruff, five reference flows, task plans, and the 9/9 sync gate are green.

All severities and manual items are now dispositioned. @bai-uipath @uipreliga please re-approve this exact head when ready; the Path-to-GA job will then be rerun.

🤖 Generated with Claude Code

tmatup:
Security follow-up on final head `d8f1ae402`: a local build verification proved the compatibility-only `NPM_AUTH_TOKEN` build-arg is expanded into Docker build progress. I therefore changed the design instead of accepting that exposure:

- CI and campaign callers now use `--secret id=npm_auth_token,env=NPM_AUTH_TOKEN`.
- The Dockerfile secret mount is optional, not `required=true`.
- The existing build-arg remains only as a deprecated fallback, so the external nightly caller does not break and no coordinated-first rollout is required.
- A missing/empty secret with no fallback fails immediately; that guard was exercised locally.
- README documents the secure caller contract.

This directly reconciles Bai’s compatibility blocker with Tomasz’s credential-history concern. The 11 checker regressions, Ruff, and 9/9 sync gate remain green. @bai-uipath @uipreliga please review/approve this final head when ready.

@claude review

🤖 Generated with Claude Code

tmatup:
Final review follow-up on head `b089671a0`:

- Added `mode:build` to coded-agent and IxP, symmetrically with their paired task definitions.
- Made the workflow dependency explicit with `DOCKER_BUILDKIT: "1"`.
- Extended the contract gate to compare task tags; it passes 9/9.
- The IxP weight reduction remains the approved and already-documented shared-output intersection decision.
- Mount sources remain guarded by the campaign launcher preflight; unverified shell `:?` expansion is not added to coder-eval YAML.

Local verification remains 11/11, Ruff clean, five reference advisories green, and sync 9/9 + shared support 1/1. @bai-uipath @uipreliga please approve this exact head when ready.

@claude review

🤖 Generated with Claude Code

PR #2562, opened 2026-08-11
Extend uipath-tasks skill: catalogs, comments, labels, metadata, data

de-si:
Extends the `uipath-tasks` skill and evals for the new `uip tasks` commands: catalogs, comments, labels, metadata, and data. Adds 3 references, 3 smoke evals, and 1 e2e eval.

**Verification:** every command these evals exercise was run live against the test tenant (`procodeapps`/`integrationtest`) and returned `Success` — catalogs create/get/update/list, data get/save, comments list/add, labels set/clear, metadata note/catalog-id/unset. Smoke evals run through coder-eval in CI. A full coder-eval pass of the e2e eval is gated on the CLI release, since the sandbox installs `@uipath/cli@latest`, which does not yet include these commands.

Jira: https://uipath.atlassian.net/browse/ACTN-11600

### DIFF
--- a/skills/uipath-tasks/references/task-catalogs.md
+++ b/skills/uipath-tasks/references/task-catalogs.md
@@ -1,0 +1,73 @@
+# Task Catalogs Reference
+
+Task catalogs group tasks and define their retention policy. Folder-scoped: a
+catalog created in one folder is invisible from another. Catalog IDs are numeric.
+
+## Folder scoping
+
+Every catalog command is folder-scoped. Pass `--folder-id <id>`, or omit it on an
+interactive terminal to pick a folder from a list. `--folder-id` is required when
+stdout is not a TTY (coding agents, CI) — pass it explicitly there.
+
+## List
+
+```bash
+# List catalogs in a folder (fetches all pages if --limit omitted)
+uip tasks catalogs list --folder-id <folder-id> --output json
+
+# Cap the number returned
+uip tasks catalogs list --folder-id <folder-id> --limit 20 --output json
+```
+
+Success `Code: TaskCatalogList`, `Data` is an array.
+
+## Get
+
+```bash
+uip tasks catalogs get <catalog-id> --folder-id <folder-id> --output json
+```
+
+Success `Code: TaskCatalogDetails`.
+
+## Create
+
+`--name` is required. `--encrypted` is create-only (the API forbids changing
+encryption after creation).
+
+```bash
+uip tasks catalogs create \
+  --name "Invoices" \
+  --folder-id <folder-id> \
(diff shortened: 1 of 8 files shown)


dushyant-uipath (approved):

skills/uipath-tasks/references/task-catalogs.md
```diff
@@ -0,0 +1,73 @@
+# Task Catalogs Reference
```
dushyant-uipath, line 1: We should add a one liner on what Task Catalogs are - possibly by picking from the official Docs definition.

Coding agent will read the skill directly - when it may / may not have access to explore what the term means

skills/uipath-tasks/references/task-catalogs.md
```diff
@@ -0,0 +1,73 @@
...
+
+Task catalogs group tasks and define their retention policy. Folder-scoped: a
+catalog created in one folder is invisible from another. Catalog IDs are numeric.
+
+## Folder scoping
+
+Every catalog command is folder-scoped. Pass `--folder-id <id>`, or omit it on an
+interactive terminal to pick a folder from a list. `--folder-id` is required when
+stdout is not a TTY (coding agents, CI) — pass it explicitly there.
```
dushyant-uipath, line 10: What does this sentence mean?

skills/uipath-tasks/references/task-catalogs.md
```diff
@@ -0,0 +1,73 @@
...
+## Update
+
+Read-modify-write: any field not passed keeps its current value. `--name` is
+optional on update. `--encrypted` is not offered (immutable).
+
+```bash
+uip tasks catalogs update <catalog-id> \
+  --folder-id <folder-id> \
+  --name "Invoices 2026" \
+  --retention-period 120 \
```
dushyant-uipath, line 58: Retention period has a max/min limit, right?

skills/uipath-tasks/references/task-data.md
```diff
@@ -0,0 +1,38 @@
+# Task Data Reference
+
+Task data is the business/form field payload attached to a task (e.g. the values
```
dushyant-uipath, line 3: Can make it explicit its key value, with keys being the schema of the task 

We should also mention here - you should never pass it empty, unless explicitly required, and without confirming with user, else it might end up overwriting the existing data.

Also worth mentioning that Save doesn't upsert, it replaces the entire JSON blob with what you currently pass, so if you want to add one extra field to data, first copy the current data object, and then append the field

de-si:
Merged as approved. The review comments are addressed in the follow-up PR: https://github.com/UiPath/skills/pull/2585

PR #2563, opened 2026-08-11
fix(uipath-solution,uipath-planner): teach solution/library structure and solution consumption (UV-15794)

roalexandru:
## What

Fixes [UV-15794](https://uipath.atlassian.net/browse/UV-15794) by teaching new projects how to structure shared code, rather than by restating a platform constraint in every skill.

A `Library` registered in a `.uipx` fails silently: its `.xaml` files open as raw XML text instead of the Studio designer, no error dialog, nothing in `UiPath.Studio.log`. `uip solution project add` also generates `resources/solution_folder/process/process/<Name>.json` describing the Library as an Orchestrator **Process** — job priority, retention, recording — a resource nothing can run, which Studio deletes on the next open of the solution.

The CLI now rejects a `Library` at `project add` / `import`, so this PR covers the part the CLI can't: what an agent should build *instead*, and where the code should live so the situation never arises.

## Changes

**`uipath-solution` — one new section, placed before Step 1 where the decision is actually made**

`references/develop-solution.md` § *Project Structure — what belongs in the `.uipx`*, keyed on a single question — **when do consumers need this code?**

| Consumers need it… | Build it as | Source lives | Consumed via |
|---|---|---|---|
| **build time** — activities/workflows called in-process | RPA **Library** | **outside** the solution root | exact version pin in `project.json` |
| **run time** — a separately deployed unit it invokes | **Solution member** | inside the solution root, in `.uipx` | Orchestrator process/agent/flow reference |

Plus the canonical repo layout, the pinned-version feed check, and the guard that actually closes the hole: **`init` / `create-project` scaffolders auto-register into the surrounding `.uipx`**, so a Library folder inside the solution root is one scaffold away from being registered — through a path the `add` gate never sees. Library source therefore lives outside the solution root, not merely unregistered inside it.

**`uipath-solution` — the missing consumption story**

`references/pack-and-deploy.md` § *Consuming a Published Solution*: `publish` → `packages list` → `packages download` → `deploy run`, and how deployed members are referenced at run time by other automations (jobs, triggers, agent tools, Flow resource nodes). Previously "I need to share this across projects" had exactly one visible answer in the skill — put it in the `.uipx` — which is how the ticket happened.

**Stated once, in the owner skill**

The constraint is phrased against the CLI error ("trust that error over any `--help` text or generated project-type table"), not against the platform, and it lives only in `uipath-solution`. Other skills link to it and restate nothing. The section closes with *"If a future CLI accepts Library membership, the table above is unchanged"* — enabling it later requires no sweep across skills.

**Frontmatter slimmed (review feedback)**

`description` + `when_to_use` was 2126 chars; Claude Code truncates the pair at 1536 in the skill listing, so ~590 chars were never reaching the model. Now **1172** (description 591, from 1022). Dropped the sub-command syntax enumeration and the diagnose laundry list — restating the CLI surface in frontmatter duplicates `--help` and the skill's own probe rule.

Full sibling routing moved to a body table, **Build Member Projects with Their Own Skill**, mapping each member signal (`.xaml`/`.cs`, `.flow`, `.bpmn`, `caseplan.json`, `agent.json`/`.py`, coded app, API workflow, Function, Library) to the skill that builds it. Rerouting happens after a skill loads, so the body is where it takes effect; the description keeps compact `→` redirects per `skill-structure.md` for first-shot selection.

**`uipath-planner`**

Keeps the `Packaging` / `Solution member` inventory columns (durable design facts). Drops the hardcoded absolutes and the Level 1 row-8 threshold change, which had dropped library-only compositions (e.g. 2 Libraries + 1 Test Automation) out of Solution scope with nothing to replace them.

**CLI surface corrections for stable 1.198.0** (verified against `--help`)

- `uip solution project` (singular) — plural returns `unknown command 'projects'`
- `project import <source>` positional + `--solutionFile`, not `--source`
- removed the nonexistent `uip rpa publish`

## Testing

**No new evals.** This is a one-time bug, not a standing quality dimension for the skill — grading the skill on it would bake a version-specific fact into CI that actively fails correct behaviour if Library membership is ever enabled. Net test-surface change is zero; `tests/tasks/uipath-planner/`, `tests/tasks/uipath-review/`, and `tests/tasks/uipath-solution/` are unchanged. Repo policy requires smoke + e2e for **new skill** PRs; this is a doc change to existing skills.

If coverage is wanted later it should come from `/test-coverage uipath-solution` on its own merits, graded on the durable build-time-vs-run-time decision.

## Validation

- `npm run skills:validate` — OK, 25 skills / 1706 files, both flavors
- `scripts/check-skill-status.py` — OK, manifest valid
- `hooks/validate-skill-descriptions.sh` — passes; `uipath-solution` description **591** chars, pair **1172** (both under their caps, nothing truncated)
- `scripts/check-skill-verbs.py` — **0 blocking** findings in both touched skills, against the unmodified generated catalog
- All relative links and anchors in the edited files resolve

`assets/uip-catalog-snapshot.json` and `.claude/rules/` are intentionally untouched. The catalog is generated by `scripts/build-uip-catalog.py` and refreshed daily from `cli/main`.

## Out of scope

- **Coded apps.** No AppV2 changes here; `uipath-coded-apps` and its solution-side rules are untouched.
- **`uipath-review`.** Deliberately not taught this boundary — a reviewer trained on a constraint that may be lifted produces false positives that then have to be unwound.
- **`.claude/rules/`.** Left untouched by request. Two follow-ups live there: the retired-verb table still lists `solution project` → `solution projects` and `functions` → `function`, so `/lint-task` advises the forms that fail on stable 1.198.0; and `check-cli-verbs.py` has no third state for a version-dependent surface where both forms are valid.
- **Follow-ups elsewhere:** the `function`/`functions` split owned by `uipath-functions`; the CLI-side doc bug where `project --help` and the solution-generated `AGENTS.md` project-type table still list libraries.
- **CI:** the `uipath-solution` activation-gate check is red for reasons unrelated to this PR — the gate emits `stop_when: auto`, which pinned `coder-eval==0.9.5` rejects, so it loads 0 rows and exits 2 on every frontmatter-touching PR. Fix is already open on `docs/uip-cli-bootstrap-installers`; details in a PR comment.

## Open question for review

Does the scaffold path (`rpa init --location <solution-dir>`, `create-project --template-id LibraryProcessTemplate` inside a solution) reject a Library too, or only `project add`? UV-15794 reached the bad state with **no** `project add` call. This PR documents the guard and the `--skip-solution-registration` escape, but if scaffolders still auto-register a Library, the CLI-side fix has a bypass that documentation can't close.

---

[UV-15794]: https://uipath.atlassian.net/browse/UV-15794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-solution/SKILL.md
```diff
@@ -1,12 +1,12 @@
...
+description: "Always invoke for `.uipx` files. UiPath Solution lifecycle via the `uip solution` CLI: init/restore/pack/publish/deploy/activate/upload, project add|import|remove, resources refresh|add|remove|edit. Also DIAGNOSE lifecycle failures: pack/publish/deploy/activate errors, stale or missing bindings, unimported/unset/unresolved resources, a project type `.uipx` rejects, deploy error codes (e.g. `[1009]`), publish name+version/processKey collisions, and `uip solution` 'unknown command' after the new→init rename. Bundles multiple automation projects (RPA/Flow/Case/Agents/API Workflows/AppV2 Coded and Coded Action apps) into one deployable `.uipx`; also consuming a published Solution. For PDD→SDD design (sdd.md/pdd.md) & multi-skill task derivation→uipath-planner. For non-solution Orchestrator/IS/resources/auth/traces→uipath-platform. For .xaml/.cs→uipath-rpa. For .flow→uipath-maestro-flow. For .bpmn→uipath-maestro-bpmn. For agent.json and .py agents→uipath-agents. For standalone coded-app deploy→uipath-coded-apps."
```
AlvinStanescu: This is extremely bloated, should be much simpler.

AlvinStanescu (approved):

.claude/rules/cli-renames.md
```diff
@@ -9,10 +9,13 @@ The catalog at `assets/uip-catalog-snapshot.json` is the source of truth for whi
...
 | `solution resource`    | `solution resources` | uip 1.196.0|
-| `solution project`     | `solution projects`  | uip 1.199.0|
 | `is resources execute` | `is resources run`   | uip 1.1.0  |
 | `resource`             | `or`                 | uip 1.196.0|
-| `functions`            | `function`           | uip 1.200.0|
+
+`solution project` (stable 1.198.x) / `solution projects` (1.199+ preview) and
+`functions` (stable 1.198.x) / `function` (1.200+ development) are intentional
+cross-version surfaces. Both forms remain in the catalog; skills must probe the
+live command group instead of treating either form as globally retired.
```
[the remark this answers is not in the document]
roalexandru: Tested this rather than taking it at face value — it is half right, and the suggested remedy would make things worse.

Probe task with stable-only patterns, run through `scripts/check-cli-verbs.py`:

| pattern | current (rows removed) | with the rows restored |
|---|---|---|
| `uip\s+solution\s+project\s+add` | **no finding** | Medium: "`solution project` → `solution projects`" |
| `uip\s+functions\s+init` | **High** | Medium: "`functions` → `function`" |
| `uip\s+functions?\s+init` (repo convention) | clean | clean |

1. The `solution project` half does not reproduce — `solution` matches as a catalog prefix, so nothing is flagged either way.
2. The `functions` half reproduces exactly as described, but no task in the repo is written that way: all 8+ `uipath-functions` tasks use the optional-`s` form `uip\s+functions?\s+init`, which lints clean in both states. That is the same convention this PR restores to the two solution tasks (`projects?`).
3. Restoring the rows would emit *"`functions` → `function`"*, i.e. advise rewriting to a form that fails on stable 1.198.0 (`uip function` → `unknown command`). That is why the row was removed — it was backwards. A vague High beats confidently wrong advice.

The real residue is that the checker only models two states — in-catalog and retired — with no third for a version-dependent surface where both forms are valid. That belongs in `check-cli-verbs.py`, not in a false row in a rules file. Filing separately; not changing it here.

skills/uipath-solution/SKILL.md
roalexandru: Fixed, and you were right for a stronger reason than style: `description` + `when_to_use` was 2126 chars and Claude Code truncates the pair at 1536 in the skill listing, so ~590 chars were never reaching the model. The bloat was discarding signal, not just adding noise.

- `description` 1022 → **383**, pair 2126 → **964**
- dropped the sub-command enumeration (`project add|import|remove`, `resources refresh|add|remove|edit`) and the diagnose laundry list — restating the CLI surface in frontmatter duplicates `--help` and the skill's own probe rule
- the `For X→` redirects moved into a body table, **Build Member Projects with Their Own Skill**, mapping each member signal (`.xaml`/`.cs`, `.flow`, `.bpmn`, `caseplan.json`, `agent.json`/`.py`, coded app, API workflow, Function, Library) to the skill that builds it. Rerouting happens after a skill loads, so the body is where it actually takes effect — frontmatter redirects only buy first-shot selection, out of a budget that gets truncated.
- `skill-structure.md` / `skill-review.md` amended so disambiguation may live in the description **or** a body routing section, instead of mandating frontmatter. Flagging that explicitly since it relaxes a MUST.

Repo-wide, only 3/25 skills breach 1536: `uipath-platform` (loses 845 chars), this one (was 590), `uipath-troubleshoot` (44). Happy to do `uipath-platform` next if you want the same treatment.

roalexandru:
Addressed the Claude review's Medium finding — the `project remove` cleanup contradiction. Settled it against CLI 1.198.0 instead of picking a side:

```
BEFORE: package/Shared.json, process/process/Shared.json   uipx: [Library, NotARealType]
$ uip solution project remove ./LibSol/Shared   → Result: Success
AFTER:  (both Shared.json files gone)                      uipx: [NotARealType]
        project folder still on disk: YES
```

`remove` deletes **both** generated resource files, drops the `.uipx` entry, and leaves source untouched. So `SKILL.md` rule 4 and `develop-solution.md` were correct; the outlier was the `solution-overview.md:65` tree comment claiming `package/<name>.json` is "NOT cleaned by `project remove`". Fixed there — that contradiction actually predated this PR.

Also in `4eb74ef01`: frontmatter slimmed and sibling routing moved to a body table (see the thread on `SKILL.md:3`).

Nothing else from either bot review is actionable — reasoning for the Codex P2 is in its thread.

roalexandru:
**The red `uipath-solution` activation-gate check is not from this PR — the gate is broken on `main` for every skill.**

`tests/scripts/activation_gate.py` emits `stop_when: auto` on the `skill_triggered` criterion, which the pinned `coder-eval==0.9.5` rejects:

```
Skipping task file /tmp/.../gate.yaml — ValueError: Invalid task definition:
  success_criteria.0.skill_triggered.stop_when
    Extra inputs are not permitted [type=extra_forbidden, input_value='auto']
Running 0 task(s) … Results: 0/0 succeeded
ERROR: .../suite.json missing   → exit 2
```

So it loads **zero** rows and exits 2 — it never measures recall. Every activation-gate run that actually executed a gate job has failed, across 4 branches and 4 different skills:

| branch | gate job | result |
|---|---|---|
| `fix/uv-15794-solution-library-structure` | `uipath-solution` | failure ×4 |
| `docs/uip-cli-bootstrap-installers` | `uipath-platform` | failure ×2, then **success** |
| `feat/hitl-case-surface` | `uipath-human-in-the-loop` | failure |
| `actn-11600-uipath-tasks-skill` | `uipath-tasks` | failure |

Runs that "pass" elsewhere are ones where `detect` skipped because no frontmatter changed (`matrix.skill: skipped`). The single genuine success is `docs/uip-cli-bootstrap-installers` **after** it patched `activation_gate.py` to drop `stop_when` (−11/+8 on that file).

Not fixing it here on purpose: the fix already exists on that branch, and duplicating it would collide on the same file. Once it merges, this check should go green on a re-run — happy to rebase then.

Worth noting the gate is a required check that has been silently red for every frontmatter-touching PR, which also means the recall baselines in `BASELINES_PCT` have not been enforced for as long as this has been broken.

skills/uipath-solution/SKILL.md
roalexandru: Correction to my note above: I initially relaxed the `skill-structure.md` MUST to allow routing in the body instead of the description — that has been reverted. `.claude/rules/` is untouched in this PR.

Final shape, which satisfies the existing rule without the bloat:

- `description` **591** chars (from 1022), pair **1172** (from 2126) — nothing truncated now
- gone: the sub-command enumeration (`project add|import|remove`, `resources refresh|add|remove|edit`) and the diagnose laundry list
- kept: compact `→` redirects for the five most-confused siblings, per `skill-structure.md`
- added: the full 11-row routing table in the body (**Build Member Projects with Their Own Skill**), which is where a reroute actually executes

So the description carries first-shot signal only and the body owns the real routing. If you still want the `→` redirects gone entirely, that needs the rule amended first — happy to do it as its own PR with before/after activation numbers, once the gate is unbroken (it currently loads 0 rows on every frontmatter PR; see my other comment).

swittenberger:
Hi, 

as the creator of UV-15794, I was wondering why the decision was made to create a workaround? 
Create the library outside of the solution vs  Studio is able to load a library in a solution. 

Thanks

PR #2566, opened 2026-08-11
test(ixp): mark executed mutations in the ambiguity smoke mock

cezara98t:
## Problem

[`skill-ixp-rename-ambiguous-entity-smoke`](https://coder-evalboard.uipath-dev.com/runs/2026-08-11_04-13-54/skill-ixp-rename-ambiguous-entity-smoke) failed at **0.85** on a run where the agent did everything the test asks: discovered the three-way `subscriptions` collision itself, named the project, group and field as candidates, and executed **no** rename.

It also ran three read-only syntax probes, to quote accurate commands in `clarification.txt`:

```
uip ixp projects update-title --help
uip ixp groups rename --help
uip ixp fields rename --help
```

The no-mutation guards used `file_contains` excludes — a **plain substring** test (`exc in content`), not a regex:

```yaml
excludes: ["ixp projects update-title", "ixp groups rename", "ixp fields rename"]
```

All three matched the `--help` lines. Excludes 0/3 absent → `(1.0 + 0.0)/2 = 0.5` on a `weight: 3.0` criterion with `pass_threshold: 1.0` → FAILURE. **A read-only syntax probe was graded as a destructive rename.**

## Fix

The mock is the only thing that can mutate, and it already has an explicit mutation branch — so record the fact where it is known, instead of re-deriving it downstream from log text.

That branch now writes a `MUTATION <args>` line, and the guards exclude the marker rather than the bare verb:

```yaml
excludes: ["MUTATION ixp projects update-title", "MUTATION ixp groups rename", "MUTATION ixp fields rename"]
```

Nothing else can write the marker, so every probe form — `--help`, `-h`, a bare verb, whatever turns up next — is exempt by construction. No regex, and no denylist of benign flags to keep extending.

This makes the help guard **load-bearing**: the branch keys on `case "$1 $2 $3"` and ignores trailing args, so without it `update-title --help` would write the marker *and* report `{"Result":"Success"}` for a command that mutated nothing — in a no-mutation test, an agent could read that as a rename it never performed.

Criteria count, types, weights and the positive control are all unchanged. The discovery criterion still proves the graded sink is live, and `calls.log` still records every invocation including the probes — the marker is additive, so the record stays faithful.

## Verification

Replayed the failed run's **exact** invocations through the fixed mock and re-graded with coder_eval's criterion semantics:

| run | before | after |
|---|---|---|
| the agent CI failed | 0.85 FAILURE | **1.000 SUCCESS** |
| silently renames the project | 0.5 on the guard | **0.83 on the guard → 0.950 FAILURE** |
| renames via delete/recreate | — | **0.83 on the delete guard → 0.983 FAILURE** |

Both violations still fail on the correct criterion. Note a real mutation scores 0.83 rather than 0 — `file_contains` averages includes with excludes and one of three excludes matches — which is below `pass_threshold: 1.0`, so the verdict is unambiguous.

Also `sh -n`, `check-cli-verbs.py` (clean), and schema validation against coder_eval's real `TaskDefinition` model.

> **Honest gap:** I did not run the full `coder-eval` harness — the mock is `#!/bin/sh` and these tasks are Linux-only, so a local Windows run would fail for unrelated reasons. The replay uses the real criterion-checker logic, not the real sandbox. `smoke-skills.yml` will exercise it properly on Linux.

## Deliberately not in this PR

An earlier revision of this branch also normalised the mock's log prefix to `uip <args>` (the base mock's format, which every sibling task anchors on), added an operand guard, and rewrote the README. All defensible, none needed to fix this failure — dropped to keep the diff proportionate. Two known follow-ups:

- This overlay logs bare `$*` while the base mock logs `uip <args>`, and never writes `calls.jsonl` — its README claims the base contract for both. A criterion pointed at `calls.jsonl` here would read an always-empty file.
- **`skill-ixp-list-model-options-smoke` failed 0.909 in the same run, same family of defect** — its positive control *"Mock still records invocations to the graded calls.log"* scored 0/1 because the agent answered correctly from the skill docs without ever invoking `uip`.
- **Skill-side root cause of the probe.** The agent ran `--help` at turn 14 but loaded the skill only at turn 19, so Critical Rule #1 ("targeted lookup in cli-reference, never guess") wasn't in context yet. It then never opened `cli-reference.md` and wrote an invented `update-title --title` flag — the real form is positional. `--new-name` appears 0× in SKILL.md and 3× in cli-reference.md.

### DIFF
--- a/tests/tasks/uipath-ixp/_shared/mock_template_ambiguous/README.md
+++ b/tests/tasks/uipath-ixp/_shared/mock_template_ambiguous/README.md
@@ -22,7 +22,8 @@ fixture through the CLI:
 | `ixp projects list` | project `subscriptions-4c1d9e2a-ixp` TITLED "subscriptions" + one decoy project |
 | `ixp projects get` | the subscriptions project's metadata |
 | `ixp projects get-taxonomy` | field group `subscriptions` containing field `subscriptions` (+ decoy group/field) |
-| rename / delete mutations | `{"Result":"Success"}` — grading catches them via `calls.log` |
+| rename / delete mutations | `{"Result":"Success"}` + a `MUTATION <args>` line in `calls.log` — the no-mutation guards exclude that marker, so only an executed mutation trips them |
+| `--help` / `-h` probe | offline auth-style failure — it must not reach the mutation branch (which keys on `"$1 $2 $3"`, ignoring trailing args) and write the marker |
 | anything else | offline auth-style failure, like the base mock |
 
 Same logging contract as the base mock: every invocation appends `$*` to
--- a/tests/tasks/uipath-ixp/_shared/mock_template_ambiguous/mocks/uip
+++ b/tests/tasks/uipath-ixp/_shared/mock_template_ambiguous/mocks/uip
@@ -5,13 +5,25 @@
 # Serves canned JSON for the read verbs so the agent can DISCOVER the fixture
 # itself: `projects list` shows a project TITLED "subscriptions";
 # `projects get-taxonomy` shows a field group "subscriptions" containing a
-# field "subscriptions". Mutations return success (grading catches them via
-# calls.log); anything else fails offline like the base mock.
+# field "subscriptions". An executed mutation returns success and records a
+# `MUTATION ` line — that marker, not the raw call line, is what the
+# no-mutation guards grade; anything else fails offline like the base mock.
 #
 # Response shapes follow the skill's cli-reference: {Result, Data} envelope;
 # get-taxonomy Data is { status, dataset: { entity_defs, label_groups } } with
 # label_groups[0].label_defs[] as the field groups and moon_form[] as fields.
 printf '%s\n' "$*" >> "$(dirname "$0")/calls.log" 2>/dev/null || true
+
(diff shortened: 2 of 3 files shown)


cezara98t:
Closing — the grading change here is a patch on the wrong layer.

CI history for this task (3 runs, pulled from blob storage) shows `--help` in only 1 of 3, and the runs separate cleanly on one factor:

| run | model | result | bash calls | turns | `--help`? |
|---|---|---|---|---|---|
| Aug 4 | sonnet-5 | SUCCESS 1.0 | 2 | 9 | no |
| Aug 5 | gemini-3.5-flash | SUCCESS 1.0 | 2 | 9 | no |
| Aug 11 | sonnet-5 | FAILURE 0.85 | 8 | 24 | yes |

Both passing runs loaded the skill first and finished in two commands. The failing run loaded it ninth — after eight calls spent poking the sandbox, including reading the mock's own source — then probed `--help` for syntax, then invoked the skill at the end.

The real defect is that load ordering, not the guard's matching. A skill edit can't fix it either: the probes happened five turns before any skill content was in context. Not worth carrying a permanent grading mechanism for a 1-in-3 outlier, so this stays unfixed rather than papered over.

PR #2571, opened 2026-08-11
test(agents): grade five coder_eval tasks by outcome, not surface details

andreibalas-uipath:
Six `uipath-agents` coder_eval tasks failed for reasons that had nothing to do with the agent's work — the criteria graded surface details (shell-call count, class names, a stale deployed app), the task asked a question no one could answer, or the prompt steered at a connector the tenant cannot authorize. Each commit fixes one, with a verified passing run.

## Changes

**`grade conversational turns by outcome, not Bash-call count`** — `command_executed` counts tool calls, so chaining both `uip codedagent run` invocations into one shell command scored 1/2 on the two-turn criterion. Replaced with a `run_command` criterion asserting two distinct echoed replies exist on disk, corroborated by the message appearing where it was submitted. Adds the missing `mode:build` tag to both tasks.
Verified: 6 live runs, all 1.000 — both tasks under claude-code (claude-sonnet-5) and codex (gpt-5.6-sol), including a codex trajectory that chained both runs in one Bash call.

**`retarget hitl_create_task_with_app at the redeployed app`** — the task graded an app that no longer matches the codereval tenant, and its prompt matched the skill's `CreateEscalation` example verbatim, so the intended `CreateTask` pattern was unreachable. Points the checker at the redeployed `RefundReviewApp` and rewrites the prompt to name the Orchestrator folder and rule out escalation.
Verified: two consecutive local runs (codex / gpt-5.6-sol) at 1.000.

**`grade LangGraph graph shape structurally, not by class name`** — the checker required the literal substrings `GraphInput`/`GraphOutput`, though `agent-patterns.md` allows either naming. Replaces the substring loop with an AST check (`assert_graph_module_shape`) asserting the real invariants: StateGraph construction, module-defined state schema, distinct module-defined Pydantic input/output models, exported top-level `graph`.
Verified: 4/4 on both harnesses across three naming schemes; exits 0 on the artifacts of the run that originally failed; all six LangGraph fixtures pass and four synthetic negatives still fail.

**`give the no-messages judge a band for wrong-layer attribution`** — the judge rubric had no band for "right conclusion, wrong mechanism" and its 0.5 band was an unreachable conjunction, so a correct refusal that blamed UiPath instead of the OpenAI Agents runner had no legal band and failed at 0.750. Adds an explicit 0.75 band, requires correct attribution for 1.0, and rewrites the 0.5 band as a disjunction.
Not verified against a live judge — no judge backend available locally (`llm_judge` needs Bedrock or `ANTHROPIC_API_KEY`); verification deferred to CI.

**`pre-answer the multi-entrypoint disambiguation in the prompt`** — bindings-reference Step 4 tells the resolver to ask which entrypoint each resource binds to when `entry-points.json` declares more than one, and sequences that ask *before* Step 5 writes `bindings.json`. In a non-interactive harness the agent asked, ended its turn, and never wrote anything — the run died after one turn with the fixture's empty skeleton intact, scoring 0.294. The codex variant also has no `AskUserQuestion` tool, so the skill's escape hatch does not exist for it. The task's description claimed asking was acceptable, but the checker requires both bindings present before it inspects `EntryPointUniqueId`, so the ask path could never pass. The prompt now tells the agent to pick any declared entrypoint, and the description matches what the checker accepts (either real entrypoint, or the field omitted; only a fabricated id fails).
Verified: two local runs at 1.000 under codex — gpt-5.6-sol (bound each resource to its own file's entrypoint) and gpt-5.6-terra, the model that originally failed (bound both to the first entrypoint).

**`steer the IS-connector prompts at a connector with a live connection`** — both IS-connector tasks asked for a tool that can "run web searches on the web". That pulls agents to `uipath-perplexity-perplexity`, which has no connection in the codereval tenant and cannot be authorized headlessly — `uip is connections create` returns `ConnectionAuthorizationPending` and the id never materializes, so agents bound a phantom connection, `uip solution resources refresh` resolved nothing, and the connection-resource assertion was unreachable. Both task descriptions already declare the dependency on `uipath-uipath-airdk` (GenAI Activities), which does have an enabled connection; the prompts now say so without naming the connector, leaving connector and activity discovery to the agent. The inline prompt names Integration Service explicitly — with "GenAI tool" alone, codex read it as a built-in capability and built a `uipath.agent.resource.tool.builtin.deep-rag` node, failing the checker's first assertion.
Verified: `inline_is_connector_tool` under claude-code (claude-sonnet-5) at 1.000 — connection resource provisioned by `uip solution resources refresh`, real connection id and folder key throughout, nothing hand-written. Same task under codex (gpt-5.6-terra) improves 0.611 → 0.722, failing only because it writes an empty `connection.id` into `resource.json` despite having looked the id up.

## Follow-up

The underlying skill gap in the last item is not fixed here: `references/coded/lifecycle/bindings-reference.md:93-99` still has no non-interactive fallback for the multi-entrypoint branch, so an agent without that prompt hint (or without `AskUserQuestion`) will still stall before writing. The task works around it rather than covering it.

## Not included

**The `role` crash in `uip agent refresh --inline-in-flow`.** A CLI bug, not a task defect: refresh throws `Cannot set properties of undefined (setting 'role')` when a chat-completion connector tool node (`uipath.agent.resource.tool.connector.uipath-perplexity-perplexity.chat-completion`) is wired to a `uipath.agent.autonomous` node's `tool` handle and its `definitions[]` entry is present. Bisected on a clean scaffold: removing the node, the definition, or the edge all clear it, and the same wiring with `uipath-uipath-airdk.web-search` succeeds and generates the resource. Worse, `uip agent validate --inline-in-flow` responds to the resulting drift with *"Run `uip agent refresh --inline-in-flow` to generate resource.json files"* — it directs the agent straight back into the crash. The prompt fix above sidesteps it by steering to airdk; the bug still needs filing against `agent-tool`.

**Two `check_inline_is_connector_tool.py` inaccuracies.** Its failure text tells agents to author `bindings_v2.json` manually, but for inline-in-flow that file is *generated* by `uip agent refresh --bindings-target` — the message is inherited from the standalone task and contradicts the skill. And the connection-resource assertion only checks that the directory is non-empty, so a hand-written stub satisfies it: in the standalone task's codex run the graded file was produced by a `Write` call, not by refresh, and the task still scored 1.000. Both worth tightening; neither touched here.

**Sandbox exposure.** `SKILLS_REPO_PATH` is the repo root, so `tests/` is inside the agent's reachable filesystem. In the standalone codex run the agent read `is_connector_tool.yaml` and `check_is_connector_tool.py` and ran the checker itself before finishing. Excluding `tests/` via `agent.ignore_patterns` would stop tasks grading against criteria the agent has read.

**A skill gap behind the remaining codex failure.** `references/lowcode/capabilities/integration-service.md:263` documents `uip is connections list "<connector-key>"` without `--all-folders`, so the enabled airdk connection — which lives in the `uipath-agents` folder — reads as absent. Claude recovered by adding the flag on its own; codex looked the id up and then wrote `"id": ""` anyway. The reference should require `--all-folders` and require carrying the discovered id, folder key, and name into `resource.json`.

### DIFF
--- a/tests/tasks/uipath-agents/_shared/langgraph_assertions.py
+++ b/tests/tasks/uipath-agents/_shared/langgraph_assertions.py
@@ -2,9 +2,17 @@
 
 from __future__ import annotations
 
+import ast
 import json
 import sys
 from pathlib import Path
+
+# Keyword names LangGraph accepts for the schemas passed to `StateGraph`.
+# The first positional argument is the state schema.
+_INPUT_KEYWORDS = ("input", "input_schema")
+_OUTPUT_KEYWORDS = ("output", "output_schema")
+_IO_KEYWORDS = _INPUT_KEYWORDS + _OUTPUT_KEYWORDS
+_SCHEMA_KEYWORDS = _IO_KEYWORDS + ("state_schema",)
 
 
 def assert_langgraph_config(root: Path, module: Path) -> None:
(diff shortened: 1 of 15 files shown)


andreibalas-uipath:
Smoke run [31508988771](https://github.com/UiPath/skills/actions/runs/31508988771) failed at 66.7% (2/3) — `skill-agent-coded-conversational-llamaindex` scored 0.826 (4/5) on the very criterion this PR adds. Fixed in de7801da5.

The run did everything right: `turn1.json`/`turn2.json`, both outputs, plus a `chat-replies.txt` transcript. The line scan can't tell a transcript label from the reply, so it extracted `Reply:   hello there echo` next to the real `hello there echo`; that variant's message part (`Reply:   hello there`) occurs only inside the reply, so it was uncorroborated — and *any* uncorroborated candidate failed the run, even though the two clean replies corroborated 4x each.

Corroboration is now a filter rather than a veto: uncorroborated candidates are dropped and the survivors counted against `min_turns`, so extraction noise can't fail a real run while a hand-written reply still doesn't count. Labelled lines also yield an unlabelled variant (so a transcript saved *without* JSON outputs still corroborates), and replies differing only by a leading label collapse to the shortest form (so one turn copied to several files counts once).

Verified: both conversational tasks' artifacts from that run now exit 0 through the real check script; new pytest module (7 cases: CI shape, transcript-only, JSON-only, colon-in-message, and three negatives) — 17 passed for `tests/tasks/uipath-agents/`. Against the pre-fix helper 4 of the 7 fail, including the CI shape. No fresh live run: `coder-eval` isn't available in this environment, so the smoke job on this push is the real re-test.

andreibalas-uipath:
Run [31510621538](https://github.com/UiPath/skills/actions/runs/31510621538): both conversational tasks now pass at **1.000** (the `de7801da5` fix holds), and the failure moved to `skill-agent-coded-strict-input-no-messages` — 0.875, pass rate 66.7% again. Fixed in 2cd6433a4.

This one is the other half of af7522171. That commit added the 0.75 band for "right conclusion, right alternative, wrong mechanism" but left `pass_threshold: 0.8`, so the band scored the case correctly and still failed the task. The run hit it precisely: claude-sonnet-5 refused to scaffold, recommended LangGraph, and attributed the `messages` requirement to "the UiPath platform" — judge score 0.750, rationale "The conclusion and alternative are right; the mechanism is misdescribed."

Threshold is now 0.75. The invariants this task exists to grade — recognise the conflict, don't scaffold, offer an alternative — are required by every band from 0.75 up, and `skill_triggered` + `command_not_executed` gate the non-scaffolding behaviour outright. Attribution varies run to run on the same model, so gating on it makes a smoke task flaky; the band keeps it visible in the score without failing the run. Rubric itself unchanged.

al3xanndru (approved):

PR #2574, opened 2026-08-11
fix(maestro-flow): don't enable error handling on nodes that handle no error (MST-13188)

rockymadden:
Closes [MST-13188](https://uipath.atlassian.net/browse/MST-13188).

## Problem

Guy Van Wert, on his `RenewalRiskOrchestration.flow`: *"In many of my nodes Claude has `error handling` set to true and it is unnecessary."* Asked whether it actually hurt:

> **"Yeah — the result is that my flow just always looks successful. Any failed API call will now just go through since it follows that error path to the next node."**

Node-level `inputs.errorHandlingEnabled: true` suppresses the node's fault. Combined with an `error` edge routed into the happy path, a flow whose API call failed reports `Completed` with success-shaped outputs. Sam Nuziale's triage (grounded in `flow-workbench@develop` — `ScopeService.ts`, `contributors/types.ts`) puts the fix in the coding-agent authoring instructions, not the FE flag.

## Root cause

The skill's prose, not the CLI. `enableErrorHandlingForErrorEdges` in `@uipath/maestro-tool` sets the flag **only** on nodes that already have an outgoing `sourcePort: "error"` edge, and `validate` errors when an error edge lacks the flag — so a flag with no matching edge is always agent-authored. No skill template or fixture sets it either.

The docs framed the port purely as survival, with faulting as the outcome to avoid, and never stated the default or the swallow failure mode. The imperative form appeared in the numbered "Wire edges" step of both HTTP mode walkthroughs — the files an agent actually follows to build the node class in the report:

> "Wire `default` to the next node and `error` to a handler — without an `error` edge, a failed call faults the flow."

`plugins/terminate/impl.md` additionally taught a downstream `Decision (error?)` handler that `planning-arch.md` explicitly forbids — the skill contradicted itself.

## Changes

All in `skills/uipath-maestro-flow/`.

**Author — stop producing the bug**

| File | Change |
|---|---|
| `references/shared/file-format.md` | Canonical rule. New **Default: off — enable only for a failure the flow actually handles** and **Do not swallow the failure** sections (✗/✓ table of error-path targets). Reframes faulting as the correct, visible default. |
| `references/author/references/planning-arch.md` | Default-off in the `error`-port blockquote; the negative half in wiring rule 12 and the edge-table blockquote; "the error path must end somewhere the caller can tell apart from success" in § Error Handling. |
| `references/author/CAPABILITY.md` | New critical rule #16 (old #16 → #17, cross-ref updated) + two anti-patterns: flag without an edge, and `error` routed back into the happy path. |
| `references/author/references/plugins/http/impl-manual.md`, `impl-connector.md` | The load-bearing fix — both "Wire edges" steps now condition the `error` edge on a stated requirement. |
| `references/author/references/plugins/http/impl.md` | Same converse, plus a Debug row for *"Flow reports `Completed` but the API call failed"*. |
| `references/author/references/plugins/connector-trigger/impl.md` | Mid-flow event node: drops the "wire `error` to survive" framing. |
| `references/author/references/editing-operations-json.md` | Converse of the existing "set the flag for error edges" note, in both places. |
| `references/author/references/plugins/terminate/impl.md` | Replaces the forbidden `Decision (error?)` example with implicit error-port wiring. |

**Diagnose — recognize it in flows that already have it**

| File | Change |
|---|---|
| `references/diagnose/references/failure-modes.md` | New **MST-13188 — Run reports `Completed` but the work never happened** entry + Index row, matching the catalog's existing Symptom/Cause/Fix/Reference shape. Covers both shapes (flag with no handler; error path rejoining the happy path) and ships a detection snippet. |

Docs only — no CLI, hook, schema, or test-fixture changes.

## Verification

- `skills/uipath-maestro-flow/.maintenance/check-all.sh` — the skill's own suite. `check-links` 0 broken of 815, `check-orphans` 0, `check-link-text` 0, `check-depth` 0 unreachable, `check-template` 0 missing sections. `check-anchors` reports 4 failures, all pre-existing and identical on `main`, none in touched files; `check-uip-commands`'s `uip solution upload` false positives are likewise identical on `main`.
- `npm run skills:validate` — OK, 26 skills, 1710 files (default and `studioweb`).
- `python3 scripts/check-skill-status.py` — OK, manifest valid.
- `npm run skills:test` — 47/48; the one failure is the pre-existing npm-publish prerelease-tag test, identical on a clean tree.
- The detection snippet in the new failure-modes entry was executed against a real `.flow` to confirm its key names (`sourceNodeId` / `sourcePort` / `inputs`) are right.
- No `skill-flavors/studioweb` overrides exist for any touched file, so all flavors inherit these edits; no `skills:build` / `skills:pack` needed.

## Follow-ups (not in this PR)

- `skills/uipath-review/references/flows/flow-review-checklist.md` § 6 and its resource/queue rows push blanket error handling — a reviewer agent can re-suggest the anti-pattern.
- A coder-eval regression task under `tests/tasks/uipath-maestro-flow/` asserting no unprompted `errorHandlingEnabled` in a flow whose prompt names no failure fallback (`_shared/flow_contains.py --absent-regex` covers it with no new check script).

[MST-13188]: https://uipath.atlassian.net/browse/MST-13188?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-maestro-flow/references/author/CAPABILITY.md
```diff
@@ -73,7 +73,8 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
...
+16. **Error handling is off by default — never enable it on a node that handles no stated failure.** Leave `inputs.errorHandlingEnabled` unset and add no `error` edge unless the requirements say what should happen when that node fails ("if the call fails…", "return X for invalid input", "handle timeouts"). The flag suppresses the node's fault, so setting it without a real handler turns a failed run into one that reports success — the flow "always looks successful" while the API call it depended on never succeeded (MST-13188). An unhandled failure faulting the flow is the correct, visible outcome. When you do wire an `error` edge, it must reach a terminal the caller can distinguish from success. See [shared/file-format.md — Implicit error port on action nodes](../shared/file-format.md#implicit-error-port-on-action-nodes).
```
jiyangzh: > The flag suppresses the node's fault, so setting it without a real handler turns a failed run into one that reports success — the flow "always looks successful" while the API call it depended on never succeeded (MST-13188).

nit: IMO not need explanation in skill for failure,
 just the correct and incorrect behavior.

jiyangzh (approved):

skills/uipath-maestro-flow/references/author/CAPABILITY.md
rockymadden: yep!

PR #2576, opened 2026-08-11
docs(maestro-flow): make CEQL string quoting and operator rules explicit for connector filters

jiyangzh:
## Why

Three nightly failures share one root cause: a hand-authored CEQL `queryExpression` string reaches Integration Services unchecked. `flow validate` passes it through; IS faults at `flow debug` with `[102003]`.

| # | Agent wrote | IS error text |
|---|---|---|
| a | `'accountNumber' = '...'` — field name quoted | `Expected a field name expression but got 'StringValue'` |
| b | `=js:"accountNumber = \"" + String(v) + "\""` — value double-quoted | `Unsupported value expression 'Column' on field 'accountNumber'` |
| c | `accountNumber eq 'ACC123'` — OData alias | `[102003] Integration Services bad request` |

Variant (b) is the expensive one: the escaped-double-quote concatenation is what an agent naturally types when it builds a filter from a runtime value. CEQL reads a double-quoted token as a column reference, so the query is syntactically valid and semantically wrong.

The docs covered only variant (a), in one troubleshooting row. There was no copy-paste template for the dynamic-value case, so agents invented the concatenation themselves.

## What changed

Single file: `skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md`.

1. **New `##### Hand-authored CEQL strings` subsection** at the end of Step 6a. It states the preference order prescriptively — `node configure --detail.filter` first (the CLI compiles and validates the tree), a hand-authored `=js:` string only when a runtime value forces it.
2. **Canonical template** for the dynamic-value case, with the quoting stated as the rule: single quotes delimit a value, double quotes mark a column reference.

   ```text
   =js:"accountNumber = '" + String($vars.<node>.output.<field>) + "'"
   ```
3. **Anti-pattern table** — wrong form → exact IS error text → corrected form, for all three variants. Variant (b) appears verbatim as the wrong example. Agents pattern-match on error text, so the error→fix mapping is the payload.
4. **OData alias → CEQL operator table** (`eq`/`ne`/`gt`/`ge`/`lt`/`le` → `=`/`!=`/`>`/`>=`/`<`/`<=`).
5. **Two new rows in the Debug → Common Errors table** for the double-quoted value and the OData alias, each with symptom, cause, and fix. The existing quoted-field-name row is unchanged.
6. **Two cross-links** into the new subsection: from the Step 6a CEQL bullet and from the Dataservice V3 workaround, which is the one sanctioned place an agent hand-writes a CEQL string.

## Notes for reviewers

- Existing content was strengthened, not duplicated. The Step 6a bullet keeps its rule text and gains a pointer; the quoted-field troubleshooting row is untouched.
- No heading or anchor was renamed. The only new anchor is `#hand-authored-ceql-strings`, which the upcoming CLI validator message can reference alongside the file path `plugins/connector/impl.md`.
- Docs only. No `SKILL.md`, no frontmatter, no rule files, no other skill touched — `git status --short` shows one modified file.
- All relative links resolve; `scripts/check-skill-status.py` passes.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
+++ b/skills/uipath-maestro-flow/references/author/references/plugins/connector/impl.md
@@ -238,7 +238,7 @@ For every match:
 - **Pass a structured filter tree under `--detail.filter`** — the CLI compiles it into both halves of the contract: the runtime CEQL string at `inputs.detail.queryParameters.<name>` *and* the design-time tree at `inputs.detail.configuration.essentialConfiguration.savedFilterTrees.<name>`. Studio Web reads the latter to render the FilterBuilder UI; only `--detail.filter` populates that side.
 - **Do not pass a raw CEQL string under `--detail.queryParameters.<name>`.** It populates only the runtime half — debug runs succeed but the FilterBuilder UI shows `undefined` when the activity is reopened in SW. The CLI rejects this at configure time.
 - **A dynamic operand in the tree works**: `"value": {"value": "=js:$vars...", "isLiteral": false}` — the CLI compiles it to a `{var_…}` placeholder plus `inputs.detail.filterVariables` the runtime resolves.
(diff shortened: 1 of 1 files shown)


rockymadden (approved):
As a stop gap until we get a strong validator, this makes sense.

PR #2585, opened 2026-08-12
docs(uipath-tasks): follow-up on catalogs/data review comments

de-si:
Follow-up to #2562, addressing @dushyant-uipath's review comments on the `uipath-tasks` references:

- **task-catalogs.md**: added a definition of what a task catalog is (from the Action Center docs) and a docs link.
- **task-catalogs.md**: clarified the folder-scoping sentence (interactive picker vs explicit `--folder-id` in non-interactive runs).
- **task-catalogs.md**: noted the retention period is a positive integer whose range the service enforces (out-of-range rejected).
- **task-data.md**: made the key/value shape explicit (keys are the task schema) and warned that Save replaces the whole payload (no upsert/merge), so never save an empty or partial object without confirming with the user.

Jira: https://uipath.atlassian.net/browse/ACTN-11600

### DIFF
--- a/skills/uipath-tasks/references/task-catalogs.md
+++ b/skills/uipath-tasks/references/task-catalogs.md
@@ -1,13 +1,19 @@
 # Task Catalogs Reference
 
-Task catalogs group tasks and define their retention policy. Folder-scoped: a
-catalog created in one folder is invisible from another. Catalog IDs are numeric.
+A task catalog is a reusable, folder-scoped definition that buckets related tasks
+and configures how they behave: data retention (delete or archive the tasks after
+a retention period), encryption of task data, and labels. Every task is associated
+with at most one catalog, and it inherits that catalog's configuration when it is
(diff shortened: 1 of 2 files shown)


skills/uipath-tasks/references/task-catalogs.md
```diff
@@ -1,13 +1,19 @@
 # Task Catalogs Reference
 
-Task catalogs group tasks and define their retention policy. Folder-scoped: a
-catalog created in one folder is invisible from another. Catalog IDs are numeric.
+A task catalog is a reusable, folder-scoped definition that groups related tasks
```
dushyant-uipath:
```suggestion
buckets related tasks
```

++ Every task is associated with at most one catalog

dushyant-uipath (approved):

skills/uipath-tasks/references/task-catalogs.md
de-si: Done. Reworded to buckets and added that every task is associated with at most one catalog.

PR #2587, opened 2026-08-12
docs(maestro-flow): escape quotes in the hand-authored CEQL value template

jiyangzh:
## Why

Follow-up to #2576. The canonical hand-authored CEQL template concatenates the runtime value unescaped:

```text
=js:"accountNumber = '" + String($vars.<node>.output.<field>) + "'"
```

A value containing a quote (`O'Brien`) ends the CEQL string early and faults `[102003]` at debug. Crafted text after the quote becomes live query syntax — injection shaped, aimed at a Data Service filter.

Source-verified against the IS server parser: the expression is parsed as a SQL WHERE clause (JSqlParser), so SQL literal rules apply — `''` is the escape for a quote inside a value.

## What changed

One file, 4 lines. The template gains `.replace(/'/g, "''")`, one sentence states the rule, and the two other occurrences (anti-pattern table corrected form, Common Errors row) match. An agent copying any of the three forms now ships an escape-safe filter.

## Notes for reviewers

- Docs only; no anchor or heading changed — cli#3587's validator messages keep pointing at the same section.
- The unsafe form was merged today in #2576; this closes the gap before agents copy it broadly.


nikhil-maryala (approved):

rockymadden (approved):
Ewww

jiyangzh:
close as we use filter builder now

PR #2590, opened 2026-08-12
fix(maestro-case): enforce two documented-but-unverified build gates (Check 14 + Check 15)

roberts-cliff:
Enforces two build gates that the skill documented but never verified. They were found separately; they are the same defect class, so they ship together.

> **The pattern.** `uip maestro case validate` reports `Valid` for both failures. In each case the skill's own contract said what to do, no check confirmed it happened, and the build reported success anyway. This is the same shape as the Aug 11 postmortem's F2 (a gate bypassed by a self-authored substitute) and F4 (`validate` passing non-startable cases).

---

# Defect 1 — a surviving connector stub ships a case that cannot start

An unresolved `wait-for-connector` rule was allowed to ship. The skill emitted the Phase 2 placeholder stub, logged it as an ordinary Open Item, and reported the build **complete**. The resulting case is 100% non-startable.

Connector-event subscriptions register at **case start**, before any stage is entered. One surviving stub anywhere kills the entire case — including when it sits on a non-required *secondary* stage nothing routes to. The failure surfaces as the case's own start event failing, which points nowhere near the offending rule.

Found on an Aug 11 bug-bash build (`AutoClaimSettlement`, 9 stages / 33 tasks): published to the solution feed, `Valid`, dead on arrival.

## Evidence

Bisection isolated the fatal *stage* but left two hypotheses implying opposite fixes: **H1** the placeholder stub is fatal (skill defect) vs **H2** any `wait-for-connector` stage-entry rule is fatal (product bug). Same case, same stage, same rule, **one field changed** — the stub replaced with a resolved Mock connector trigger:

| Rule `uipath` state | start event | elements | `validate` |
|---|---|---|---|
| `connectorKey`/`operation` = `"placeholder"` | **Failed** | **1** | `Valid` |
| resolved connector (spliced `caseShape`) | Completed | 25 | `Valid` |

H1 confirmed, H2 falsified — `wait-for-connector` is fine. Two further results:

- **Not stage-scoped.** The stub sits on a secondary stage nothing routes to and still kills the start event.
- **Not placeholders generally.** The same case's placeholder *task* (`data: {}`, Rule 8) is untouched in the run that starts fine. Placeholder tasks are shippable; placeholder **condition rules** are not.

Measured on uip 1.198 (2026-08-12); the control was reproduced on a current token before the variant ran. This is a one-off manual measurement recorded for provenance — it is not what guards the fix.

## Fix — Step 12 Check 14

Every surviving stub halts the build, with no best-effort escape: unlike Checks 1/2/4/5/13 the artifact is not partially working, so "continue with best-effort emit" would ship something 100% dead while reporting success. Remediation is resolve / remove / abort, and the abort path writes a literal `BLOCKED:` marker to `build-issues.md` so the halt is a machine-readable side effect rather than prose.

**A scope bug in the first version of this PR, caught by running the test end-to-end** (`56bfc202e`): the original wording said a stub "may NOT ship" without saying *where* the refusal happens, and the agent refused to build anything at all — thorough Phase 1, then a full stop with no solution, no caseplan, no `build-issues.md`. Task scored **0.452**.

Defensible, but wrong. Check 14's remediation (b) — *remove the rule and ship the working remainder* — is normally the only option that unblocks the author without waiting on the connector, and it only exists once the case is built. The run-1 transcript shows exactly that: three options offered, none of them "drop the lane and ship the rest." Clarified in four places that **"must not ship" is not "must not build."**

Supporting corrections: the blast-radius claim in `connector-trigger-impl.md` (*"the rule fails at debug/run"* → the case cannot start); the *"a remaining stub must map to a reported unresolved connector"* permission revoked in all four condition plugins; `SKILL.md` Rules 8 and 14.

**Drive-by:** Check 13 was defined but missing from the Step 12 invocation list, so it had never run.

---

# Defect 2 — the issue log was never verified

`build-issues.md` is referenced in ~30 places across 10 files and verified in **zero**.

```
Step 12 Checks 1–14   inspect caseplan.json + sidecars — never the log
Step 12.1             single terminal Write, no gate after it
Reporting clause      "read the file after writing" — presumes it exists
```

The issue list is held **in the agent's reasoning** for the whole build (`logging/impl-json.md` § Setup) and dumped once at Step 12.1 — so on a long build it is the first casualty of context pressure, and nothing notices.

## Evidence

The `context_scale` claim-settlement e2e, 2026-08-13:

```
52.4 min · 8 stages · 39 tasks — every one a placeholder
120 <UNRESOLVED> markers in tasks.md
NO build-issues.md at all
final score: 1.0 — because no criterion looked
```

The operator receives 39 unwired resources with no record of which ones. Yesterday's bug-bash run of the same requirements *did* write a log, so this is **nondeterministic compliance**, not a missing feature.

## Fix — Step 12.2 / Check 15

A Phase 4 **exit** gate, not part of the Step 12 validator pass — it cannot join Checks 1–14 because the file does not exist until 12.1 has run.

When the build carries unresolved work (`<UNRESOLVED>` markers, placeholder tasks, or a surviving connector stub), `tasks/build-issues.md` must exist **and carry ≥ 1 entry**; a header-only file fails. Repair is non-interactive: reconstruct from on-disk artifacts, stamp a `NOTE:` marking severity and step attribution approximate — a reconstructed log records what the artifacts prove, not what the build observed — re-scan once, halt before Phase 5 if still empty.

---

# Verification

**30 grader unit tests, in CI, no tenant required** — 16 for Check 14, 14 for Check 15, all on minimal synthetic stubs (root + one stage + one rule). They ride the existing `maestro-case checker unit tests` job. `pytest tests/tasks/uipath-maestro-case/` → **110 passed**.

Both graders were additionally validated against **real run artifacts**, and both discriminate:

| Grader | claim-settlement run | connector-halt run |
|---|---|---|
| Check 14 | branch A (no stub) | branch B (stub + marker) ✅ |
| Check 15 | **exit 1** — the run that scored 1.0 | exit 0 ✅ |

**Mutation-tested**, so neither suite is decoration:

| Mutation | Caught by |
|---|---|
| Check 14 grader always returns 0 | 15 of 16 fail |
| stage-exit scope not scanned | `scope[stage_exit]` |
| any `"placeholder"` mention counts as a halt | open-item test |
| Check 15 treats every build as clean | 7 of 14 fail |
| any file counts as a log | 3 fail |
| placeholder tasks ignored | 5 fail |
| `<UNRESOLVED>` markers ignored | 1 fail |

The load-bearing cases are the ones that encode the *old* behaviour: `test_open_item_mention_is_not_a_halt` (the pre-fix skill logged the stub as a routine Open Item — a grader accepting any mention of `"placeholder"` would pass exactly the output this PR rejects) and `test_header_only_log_fails`.

**A grader bug the tests caught:** the logging template's `**Case file:** … | **Timestamp:** …` line fell through as an issue entry, so a header-only log passed. Fixed in `741b433da`.

## End-to-end

```
skill-case-unresolved-connector-rule-halt   SUCCESS   1.0   7/7   10.4 min
coder-eval 0.9.2 (tests/.venv) · experiments/default.yaml · claude-sonnet-5 · anthropic_direct
```

Landed in branch B — structure built, stub survives, `BLOCKED:` marker written, no debug, no publish, and the closing report explicitly does not call the build complete.

Two honest caveats on that claim:

- The recorded run **predates** the Check 15 criterion swap on that task. Its artifacts were re-graded against the new checker and pass (exit 0), but the run itself was scored on the previous criteria set.
- The `context_scale` task has **not** been re-run since its Check 15 criterion was added. That criterion is known to fail the 2026-08-13 run, which is the point of adding it.
- Local harness is `coder-eval 0.9.2`; CI pins `0.9.6`.

## Static checks

`skills:validate` (26 skills, default + studioweb) · `skills:build` · `check-skill-status` · `check-cli-verbs` · relative links resolve. No flavor overrides or `skill-flavor` markers on any touched file. Frontmatter untouched, so no recall-eval gate.

---

# Not included

- **F4** — `validate` reporting `Valid` for non-startable cases is the platform half of Defect 1 and is unchanged. Check 14 exists *because* `validate` cannot be the gate; every consumer that is not this skill still has none. Worth filing separately.
- **The incremental issue log.** Check 15 is the guard, not the cure: it detects the loss at the end, by which point the in-reasoning list is already gone and the agent reconstructs from artifacts. Writing each entry when it is raised would survive context compaction outright, but it contradicts the current design note and adds a Write per issue to a hot path. Deliberately deferred.
- **Design nondeterminism.** Two runs from identical requirements produced materially different case topologies (Aug 11 modelled the withdrawal path as `wait-for-connector`; Aug 13 as `user-selected-stage`). That is why the Aug 13 run does not exercise Check 14 at scale, and it is arguably more consequential than either defect here — a skill that designs a different case each time is hard to test at all. Not addressed.
- Pre-existing tag drift: `feature:conditions` is used by 12 tasks but is not in the closed vocabulary in `tests/README.md`. New tasks here use closed values rather than adding a 13th.


jundayin (approved):

tests/tasks/uipath-maestro-case/_shared/check_build_issues_log.py
```diff
@@ -0,0 +1,198 @@
...
+            if not cells or all(not c for c in cells):
+                continue
+            header_like = {"category", "severity", "step", "plugin", "message",
+                           "errors", "warnings", "skipped", "issue", "detail"}
+            if all(c.lower() in header_like for c in cells if c):
+                continue
+            return True
+        if line.startswith(("- ", "* ", "1.")):
+            return True
+        return True
```
rockymadden, line 124: **This falls through on the actual template, so the header-only case isn't caught.** `plugins/logging/impl-json.md` § First flush writes `<!--build-issues:summary:start-->` and `_Summary written at Step 12.1._`. Neither is a header, `NOTE:`/`BLOCKED:`, `set(line) <= set("|-: ")`, or a `**case file|timestamp**` match — so both land here. Each alone passes:

```
exit=0  <- <!--build-issues:summary:start-->
exit=0  <- _Summary written at Step 12.1._
exit=1  <- **Case file:** caseplan.json | **Build started:** …
```

Pristine template + one `data: {}` task → `PASS: unresolved work is recorded in tasks/build-issues.md.`

And that state is mandated, not hypothetical — § Flush: *"Flush even when the section produced zero issues — the file's existence after the first section is what proves the log survived."* So the file exists with exactly this content from the first boundary; a build that then loses the in-reasoning list appends nothing and passes. That's the nondeterministic-compliance case from the description, one notch off the Aug-13 run (no file at all — still caught).

`test_header_only_log_fails` misses it because `HEADER_ONLY_LOG` is hand-written: `**Timestamp:**` instead of `**Build started:**`, no comment markers. Skip `<!--` and `_…_`-only lines (or invert to an allowlist — Journal rows + `## Open Items` bullets), and make the fixture a verbatim copy of the template so it can't drift again.

skills/uipath-maestro-case/references/implementation.md
```diff
@@ -311,7 +311,23 @@ After value bindings (Step 9.8), connector-rule upgrades (Step 10.5), and marker
...
+
+  **Remediation — AskUserQuestion, no best-effort option.** Name the stage, condition, and rule, then offer exactly three picks: (a) **resolve the connector now** — supply/choose the connector + connection, re-run Step 10.5 for that rule, re-scan; (b) **remove the rule** — delete it and, when it was the condition's only rule, the condition too (if that leaves a stage with no entry condition, remove the stage and every reference to it, then re-run Checks 8 and 13); (c) **abort the build** — leave `caseplan.json` on disk and stop.
```
rockymadden, line 320: **No non-interactive path, and this PR ships a task that can't take the interactive one.** Check 14 directly above has an explicit **Non-interactive repair:**; so do 6, 7, 10, 11, 13. Rule 11, Rule 17, `registry-discovery.md` § 1b, `connector-integration.md` § 4 and `phased-execution.md` all define non-interactive fallbacks. This is the only halting check without one.

`experiments/default.yaml` doesn't grant `AskUserQuestion`, and `context_scale_claim_settlement.yaml` in this PR both omits it from `allowed_tools` and says "Do NOT call AskUserQuestion" — while its own description notes the design nondeterminism that decides whether the withdrawal path becomes a `wait-for-connector` rule at all. Some runs land here with no defined behavior.

Cheapest fix is naming (c) the non-interactive default: abort, leave `caseplan.json`, write the `BLOCKED:` marker. That's already what the marker is for.

tests/tasks/uipath-maestro-case/unresolved_connector_rule_halt/check_connector_stub_halt.py
```diff
@@ -0,0 +1,217 @@
...
+from __future__ import annotations
+
+import json
+import re
+import sys
+from pathlib import Path
+
+# implementation.md § Step 12 Check 15 mandates this literal line in build-issues.md.
+BLOCKED_MARKER = re.compile(
+    r"BLOCKED:\s*case is not startable\s*[—-]\s*\d+\s+unresolved wait-for-connector rule\(s\)"
```
rockymadden, line 42: **Em-dash only, so a correct halt can score as the regression.** `[—-]` covers em-dash and hyphen, not en-dash, and the rest of the sentence is pinned:

```
exit=0  BLOCKED: case is not startable — 1 unresolved wait-for-connector rule(s) still carry the placeholder stub.
exit=1  … is not startable – 1 unresolved …              (en-dash)
exit=1  BLOCKED: the case is not startable — …
exit=1  … 1 unresolved wait-for-connector rule still carries …
exit=1  … one unresolved wait-for-connector rule(s) …
```

Check 15 does say the marker is matched mechanically, so the strictness is deliberate — but en-dash substitution is routine, and the cost is the 5.0-weight PRIMARY criterion failing under "a dead case was emitted as if complete" when the agent halted exactly as told. Match the load-bearing tokens (`BLOCKED`, `not startable`, `wait-for-connector`) and let punctuation float.

rockymadden (commented):
Solid work — the one-field control/variant table earns its keep, and the Check 13 drive-by alone justifies the PR. Four things before this goes in, one blocking.

Bookkeeping first: title and body say "Check 14 + Check 15" but the code ships 15 + 16 after `985e957fe`. And `context_scale/` is three *new* files here (~418 lines, `task_timeout: 14400` / `max_turns: 600` — both the largest in the repo), but the body reads as though it already exists ("has not been re-run since its Check 15 criterion was added"). Worth calling it out as new and deciding deliberately whether an `e2e`-tagged 4-hour cost fixture should get picked up by glob-based full-suite runs. "110 passed" is stale too — I get 125 passed / 14 skipped. @jundayin's approval predates the renumbering merge and the review bot only ever saw Defect 1, so this probably wants another pass.

roberts-cliff:
Closing. The defect is still real; this PR is not the way to land it.

**Re-measured on today's CLI** (2026-09-11). Took a grader-passing golden plan, replaced one `wait-for-connector` rule's `uipath` with the documented placeholder stub, and validated:

| profile | result |
|---|---|
| plain | **`Valid`, zero errors** |
| `--strict` | caught — `STRICT_CONNECTOR_CONTEXT_INCOMPLETE` |

So the finding this PR opened with is unchanged on the default profile: a case that cannot start at all still ships clean. What changed since August is that `--strict` catches it, and the skill now runs strict at the Phase 3 exit and in Phase 4 (#3204), so a build following today's contract is stopped by the CLI rather than by prose. Check 14's skill-side enforcement is therefore mostly superseded — and the residual gap is an argument for making strict the default profile, not for more prose.

**Why not rebase this instead:** three weeks stale and `DIRTY`; 17 files and +1783; the sections it patches were rewritten twice since (#3204, the cli#4160 code map); and its check numbers are taken — main's Check 14 is now variable `default` encoding and Check 15 is the task-entry-rule sweep.

**What survives and will be reopened separately:** the two test directories, neither of which exists on `main` — `tests/tasks/uipath-maestro-case/unresolved_connector_rule_halt/` (retargeted to assert the CLI catches the stub, since that is now the enforcement point) and `tests/tasks/uipath-maestro-case/context_scale/` (a 9-stage / 33-task scale fixture we have no equivalent of).

PR #2596, opened 2026-08-13
ci(smoke): exempt coder-eval from safe-chain minimum package age

tibrnui:
Stacked on #2585 — targets `docs/uipath-tasks-review-followup`, not `main`.

## Problem

Two jobs in `Smoke Skill Tests` fail on #2585:

- `Run skill smoke tests` (`e2e`)
- `Validate task schema (advisory)` (`validate-task-schema`)

Both install the pinned wheel via `uv pip install --system "coder-eval==<pin>"`. safe-chain's minimum-package-age check suppresses freshly published versions during metadata resolution, so the same-day pin in `tests/.coder-eval-version` resolves to nothing:

```
× No solution found when resolving dependencies:
╰─▶ Because there is no version of coder-eval==0.9.6 and you require
    coder-eval==0.9.6, we can conclude that your requirements are
    unsatisfiable.
ℹ Safe-chain: Some package versions were suppressed during package metadata resolution due to minimum package age.
```

## Fix

Add `SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS: coder-eval` as job-level `env` on both jobs.

- `coder_eval` is a first-party package from the internal Azure Artifacts feed, so the age heuristic adds no supply-chain value there.
- Scoped to the age check only — malware/typosquat/integrity checks still run.
- Scoped to `coder-eval` only — no wildcards, no other package exempted.
- Same pattern already used elsewhere in the org (`UiPath/coder_eval`, `UiPath/uipath-agents-python`, `UiPath/cli`).

## Note

Editing `.github/workflows/smoke-skills.yml` matches the `detect` job's test-infra trigger, so this PR fans out to the full smoke suite (capped at `SMOKE_MAX_TASKS`, 40). Add the `ci:skip-infra-trigger` label if you'd rather not pay for that run.

### DIFF
--- a/.github/workflows/activation-gate.yml
+++ b/.github/workflows/activation-gate.yml
@@ -21,6 +21,13 @@ on:
 permissions:
   contents: read
 
+# Exempt the first-party `coder-eval` wheel from safe-chain's
+# minimum-package-age suppression so a same-day pin in
+# tests/.coder-eval-version still resolves — see smoke-skills.yml for the full
+# rationale. Workflow-level so any job added later inherits it.
+env:
+  SAFE_CHAIN_MINIMUM_PACKAGE_AGE_EXCLUSIONS: coder-eval
+
 jobs:
   detect:
     runs-on: uipath-ubuntu-latest
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -64,6 +64,16 @@ on:
 
 # No concurrency group: this is a manual, on-demand workflow — allow any number
 # of runs (e.g. the same task across claude / codex / antigravity) at once.
+
+# Exempt the first-party `coder-eval` wheel from safe-chain's
+# minimum-package-age suppression so a same-day pin in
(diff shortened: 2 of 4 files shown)


bai-uipath (approved):

bai-uipath (commented):
**Right mechanism, incomplete scope: two things before merge.**

- **Three other PR-gating install sites keep the bug.** `activation-gate.yml` and `smoke-rpa-skills.yml` install the pinned wheel the same way and are both `pull_request` required checks; they only stayed green on #2585 because their jobs skipped (draft PR, no RPA skills changed). `run-coder-eval.yml` has five more, so a fresh pin also breaks ad-hoc and nightly dispatches. Fix: set the exclusion as workflow-level `env` in all four files instead of per-job, so any install job added later inherits it.

- **Targeting #2585 leaves the rest of the repo red.** Stacked on a docs branch, the exemption only reaches main when #2585 does, so every other open PR keeps failing to install the pin in the meantime. Fix: cut this as a standalone PR against main and rebase #2585 on top.

PR #2598, opened 2026-08-13
[Coded Agents] Add information about cases where we use Attach/Open in CW

IoanaMarinV:


skills/uipath-rpa/references/coded-vs-xaml-guide.md
```diff
@@ -166,6 +166,55 @@ Both XAML workflows (via typed arguments) and coded workflows (via direct refere
 
 ---
 
+## Converting an Existing Workflow Between Coded and XAML
```
liviusanduuip, line 169: I am not sure this belongs in Skills repo. These are  UIAutomation details. Maybe we can have a coded-xaml-migration skill in UIAutomation and reference it somehow from Skills (check how uia-configure-target skill and other docs from UIA are referenced)

PR #2603, opened 2026-08-13
fix(maestro-case): flush the issue log per section instead of once at the end

roberts-cliff:
## The bug

The issue list was accumulated **in the agent's reasoning** for the whole build and written once at Step 12.1. On a long build that list is lost to context pressure before it is ever written — and nothing downstream reads it, because Step 12's checks inspect `caseplan.json` and its sidecars, never the log.

Measured on two independent runs of the same task, different models, both producing a caseplan a reviewer would accept:

| Run | `<UNRESOLVED>` in `tasks.md` | Placeholder tasks | `build-issues.md` |
|---|---|---|---|
| `claude-sonnet-5`, 52.4 min | 120 | 39 | **never written** |
| `gpt-5.6-terra` (codex), 7.2 min | 0 | 23 | **never written** |

The codex run is the worse of the two: 23 placeholder tasks recorded **nowhere at all**, not even as markers in `tasks.md`. The operator receives unwired resources with no record of which ones.

## The fix

**Flush grain is the section boundary, not the issue.** Per-issue writes would cost one `Edit` per finding and fight the per-section batch write contract; per-build writes are the defect itself. A section boundary already carries a validate, so the flush rides an existing seam and bounds worst-case loss to one section.

- `tasks/build-issues.md` becomes an append-only **Journal**, on disk from the first Phase 2 section onward.
- **Step 12.1 stops being the only write** and becomes a summary pass that reads the journal back *off disk* — counts come from the file, not from reasoning. That is the whole point.
- Flush happens **even when a section produced zero issues** — the file's existence after the first section is what proves the log survived.
- A **Recovery** path covers a missing journal, stamped with a literal `NOTE:` so a reconstructed log is never mistaken for an observed one.

## Verification

**13 unit tests** on synthetic stubs (`_shared/check_issue_journal.py`), wired into the existing `logging_issue_log` task. The checker verifies the *incremental contract*, not mere existence — an old-style end-of-build dump **fails** for having no Journal section, which is what makes it a regression test rather than a file-exists check.

Mutation-tested: `always-clean` → 9 fail · `journal-not-required` → 1 fail · `empty-journal-ok` → 1 fail. `pytest tests/tasks/uipath-maestro-case/` → **93 passed**.

### End-to-end: N=3 per model, 6/6 clean

`logging_issue_log`, run locally against this branch:

| Model | Runs | Result | Duration | Journal | Summary filled | Rows |
|---|---|---|---|---|---|---|
| `gpt-5.6-terra` (codex) | 3 | **3/3 SUCCESS 1.000** | 3.7 / 5.5 / 3.9 min | ✅✅✅ | ✅✅✅ | 1 / 2 / 1 |
| `claude-sonnet-5` | 3 | **3/3 SUCCESS 1.000** | 8.2 / 8.2 / 11.1 min | ✅✅✅ | ✅✅✅ | 2 / 1 / 2 |

Every run was verified **at the artifact level, not by score**:

- `caseplan.json` present — so none passed via the checker's never-reached-Phase-2 exemption, which is how an auth failure would otherwise manufacture a false green
- `## Journal` section present
- summary block filled, not left at its `_Summary written at Step 12.1._` placeholder
- **zero** runs took the `reconstructed` fallback path — all six wrote the journal as they went

The `claude-sonnet-5` arm matters most: that is the model whose 52-minute, 39-placeholder build shipped no log at all and started this.

### The controlled pair that isolated the mechanism

Before the N=3, one variable was changed on its own:

| `logging_issue_log` on codex/terra | Result |
|---|---|
| prompt restating *"after validate, write the issue log"* | old format — **FAILURE 0.684** |
| that competing instruction removed | Journal + filled summary — **SUCCESS 1.000** |

The produced artifact is exactly the new contract:

```markdown
<!--build-issues:summary:start-->
| Category | Errors | Warnings | Skipped |
| [stage-exit-conditions](#stage-exit-conditions) | 1 | 0 | 0 |
<!--build-issues:summary:end-->

## Journal
| Sev | Step | Plugin | Message |
| ERROR | 12 | stage-exit-conditions | Validation reports `...exitConditions.0.type` is `invalid-exit-type`... |
```

**Worth generalising:** a task prompt that restates skill procedure silently overrides the skill, so tests written that way cannot detect skill regressions. That is why the prompt edit here *removes* an instruction rather than adding one — it deletes over-specification, it does not teach the answer.

## This PR stands alone

**No dependency on any other PR, in either direction, and no required merge order.**

- It introduces no reference to Check 15, Step 12.2, or anything outside this branch — verified: `grep -rn "Check 15\|Step 12\.2" skills/ tests/` returns nothing.
- Its own regression coverage is self-contained: `_shared/check_issue_journal.py` plus the criterion wired into `logging_issue_log`. Nothing external enforces the contract for it.
- The 6/6 end-to-end result above was produced against **this branch alone**, off `main`.

If #2590 also lands, the two compose — that PR's Check 15 is a Phase 4 exit guard that detects a missing log, while this one removes the cause so there is far less to detect. But neither needs the other, and either can merge first.

**One trivial textual overlap**, if both land: both edit the Phase 4 numbered list in `SKILL.md`. This PR rewrites item 3 (`Dump` → `Summarize`); #2590 edits item 1 and appends an item 4. Git will flag it because the lines are adjacent; the resolution is to keep both edits, which are not in tension.

## Caveats

- The two arms ran **concurrently**, so durations are not comparable across models — the 3.7–5.5 min vs 8.2–11.1 min split is under contention and should not be read as a speed result. Compliance, which is what is being measured, is unaffected.
- The controlled pair is still **N=1 per arm**. The N=3 establishes that the fix holds; the pair explains *why* the first attempt did not, and that explanation is a single observation.
- `logging_issue_log` builds a small timer-only case. It exercises the flush contract, not the 39-placeholder scale at which the defect was originally found — that fixture (`context_scale`) is untracked in main and not part of this PR.
- Local harness is `coder-eval 0.9.2`; CI pins `0.9.6`.

### DIFF
--- a/skills/uipath-maestro-case/references/implementation.md
+++ b/skills/uipath-maestro-case/references/implementation.md
@@ -57,7 +57,7 @@ Before any build step, initialize an empty issue buffer **in the agent's reasoni
 issues = []
 ```
 
-> **Why incremental.** A list held in reasoning for the whole build is lost to context pressure before it is ever written, and no Step 12 check reads the log. Measured on two independent runs of the same task: `claude-sonnet-5` (52.4 min, 39 placeholder tasks, 120 `<UNRESOLVED>` markers) and `gpt-5.6-terra` (7.2 min, 23 placeholder tasks, 0 markers) — **neither wrote `build-issues.md` at all**, and both looked like builds a reviewer would accept. Flushing per section bounds worst-case loss to one section and rides the validate seam that already exists there.
+> **Why incremental.** A whole-build buffer held in reasoning is lost to context pressure before it is ever written, and no Step 12 check reads the log. Flushing per section bounds worst-case loss to one section and rides the validate seam already at that boundary.
 
 ---
 
--- a/skills/uipath-maestro-case/references/implementation.md
+++ b/skills/uipath-maestro-case/references/implementation.md
@@ -23,6 +23,7 @@ Every plugin uses direct JSON writes via its `impl-json.md`. Cross-cutting mecha
    - **<10 T-entries** — N Edits in sequence, one per T-entry. Skip the re-Read between sibling Edits.
    - **≥10 T-entries** — single whole-section Edit or Write replacing the section's container (e.g., `schema.nodes`, a stage's `data.tasks`). Compose the complete post-section state in reasoning from the section-entry Read, then emit one write. Untouched siblings (other sections, root fields, unrelated nodes) MUST be copied verbatim — drop nothing.
 3. **One validate** at section boundary.
+4. **One issue-log flush** at the same boundary — append the section's buffered issues to `tasks/build-issues.md` per [`plugins/logging/impl-json.md` § Flush](plugins/logging/impl-json.md), then clear the buffer. The first flush creates the file; later flushes append to its Journal table. **Flush even when the section produced zero issues** — after the first section the file must exist, and its existence is what proves the log survived the build.
 
 TaskUpdate items keyed by T-number are the audit trail — mark each `in_progress` before composing the entry's mutation, `completed` after the write returns success. The audit trail stays T-by-T even when the file diff collapses to one whole-section write.
 
@@ -49,12 +50,14 @@ Full contract — recovery, tool primitive selection (Edit default, whole-sectio
 
 ## Issue Log — Initialize Before Step 6
 
-Before any build step, initialize an empty issue list **in the agent's reasoning** (not as a file, not via subprocess). All plugins append to this shared list during execution. Dump to `tasks/build-issues.md` via the Write tool after Step 12. See [`plugins/logging/impl-json.md`](plugins/logging/impl-json.md) for the entry format, severity levels, and file schema.
+Before any build step, initialize an empty issue buffer **in the agent's reasoning** (not as a file, not via subprocess). All plugins append to it during the current section, and **the buffer is flushed to `tasks/build-issues.md` at every section boundary, then cleared** — it is NOT a whole-build accumulator. See [`plugins/logging/impl-json.md`](plugins/logging/impl-json.md) for the entry format, severity levels, flush mechanics, and file schema.
 
 ```text
-# pseudocode — kept in the agent's reasoning, not on disk
-issues = []  # shared across all steps
+# pseudocode — per-section buffer, flushed at each section boundary
+issues = []
 ```
+
+> **Why incremental.** A list held in reasoning for the whole build is lost to context pressure before it is ever written, and no Step 12 check reads the log. Measured on two independent runs of the same task: `claude-sonnet-5` (52.4 min, 39 placeholder tasks, 120 `<UNRESOLVED>` markers) and `gpt-5.6-terra` (7.2 min, 23 placeholder tasks, 0 markers) — **neither wrote `build-issues.md` at all**, and both looked like builds a reviewer would accept. Flushing per section bounds worst-case loss to one section and rides the validate seam that already exists there.
 
 ---
 
@@ -324,9 +327,13 @@ Authoritative validation. Full contract — command, retry policy, AskUserQuesti
 
 Run validate per [phased-execution.md § Phase 4](phased-execution.md#phase-4--validate). On success: proceed to Step 12.1. On 3rd failure: hard-stop prompt per the same section.
 
-## Step 12.1 — Dump issue log
+## Step 12.1 — Summarize the issue log
 
-Write issue list to `tasks/build-issues.md` per [`plugins/logging/impl-json.md`](plugins/logging/impl-json.md). On Phase 4 success → proceed to Phase 5.
+The journal has been on disk since the first section boundary; this step does **not** create it. Flush any buffered issues from the final section, then read the journal back and write the grouped counts into the summary block per [`plugins/logging/impl-json.md` § Summary](plugins/logging/impl-json.md). Counts come from the file, not from reasoning — that is the point of flushing incrementally.
+
+If `tasks/build-issues.md` is absent here, the incremental flush was skipped: reconstruct from on-disk artifacts and stamp the `NOTE:` line per [§ Recovery](plugins/logging/impl-json.md). A build carrying `<UNRESOLVED>` markers or placeholder tasks must not reach Phase 5 with no log.
+
+On Phase 4 success → proceed to Phase 5.
 
 ---
 
--- a/skills/uipath-maestro-case/references/phased-execution.md
+++ b/skills/uipath-maestro-case/references/phased-execution.md
@@ -32,7 +32,7 @@ Decisions are front-loaded so the build can run unattended; the gates that remai
 |---|---|---|---|
 | **2 — Prototyping** | Solution/project, structure, triggers, task shapes, conditions in all 4 scopes, SLA + escalation; connector-bound rules use canonical stubs | `caseplan.json` emitted; `--skeleton-v2` preview validate attempted, with unsupported-flag fallback to `--skeleton` | Pause-at-preview runs: `Publish for review` / `Skip publish and continue` / `Abort`. Straight-through runs: none — counts line, continue (Rule 11) |
 | **3 — Implementation** | Connector task schemas, task I/O value binding, resolved connector-rule stub upgrades | `caseplan.json` ready for authoritative validation | None — proceeds to Phase 4 |
-| **4 — Validate** | Run authoritative `uip maestro case validate`, dump `build-issues.md` | `caseplan.json` passes full validation | On 3rd validate failure: `Retry with fix` / `Pause for manual edit` / `Abort` |
+| **4 — Validate** | Run authoritative `uip maestro case validate`, summarize `build-issues.md` (journal already on disk) | `caseplan.json` passes full validation | On 3rd validate failure: `Retry with fix` / `Pause for manual edit` / `Abort` |
 | **5 — Publish** | Optional Studio Web upload | `DesignerUrl` printed | `Publish to Studio Web` / `Skip to Debug` |
 | **6 — Debug** | Optional CLI debug run (real execution — emails, API calls, etc.) | Debug output streamed | `Run debug session` / `Continue to publish` |
 | **7 — Publish to Orchestrator** | Optional `solution pack` + `solution publish` to the tenant solution feed | `.zip` packed; publish result printed | `Publish to Orchestrator` / `Done` |
@@ -197,9 +197,11 @@ Up to **3 validation retries** per session — each retry MUST be preceded by a 
 - `Pause for manual edit` — exit skill mid-flight; user edits `caseplan.json` directly and re-runs skill.
 - `Abort` — exit; dump `build-issues.md`; leave artifacts in place.
 
-### Dump issue log
+### Summarize the issue log
 
-After successful validate, write issue list to `tasks/build-issues.md` per [`plugins/logging/impl-json.md`](plugins/logging/impl-json.md), grouped by plugin with summary index. Source of truth for completion report. Write even if zero issues logged (confirms clean build).
+`tasks/build-issues.md` already exists — it is flushed at every section boundary from Phase 2 onward per [`plugins/logging/impl-json.md` § Flush](plugins/logging/impl-json.md). After successful validate, flush the final section's buffer, then read the journal back and fill the grouped summary block. Source of truth for the completion report; the counts come from the file, not from reasoning.
+
+If the file is absent here the incremental flush was skipped — reconstruct from on-disk artifacts and stamp the `NOTE:` line per [§ Recovery](plugins/logging/impl-json.md).
 
 On Phase 4 success → proceed to Phase 5.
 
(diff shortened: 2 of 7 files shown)


skills/uipath-maestro-case/references/plugins/logging/impl-json.md
```diff
@@ -1,17 +1,32 @@
...
 
-## Setup
+## Why the log is incremental
 
-Initialize the in-reasoning issue list before Step 6:
+An issue list held in reasoning for the whole build is lost to context pressure before it is ever written, and nothing downstream notices — Step 12's checks read `caseplan.json` and its sidecars, never the log.
+
+Measured, two independent runs of the same task on different models:
+
+| Run | `<UNRESOLVED>` in `tasks.md` | Placeholder tasks | `build-issues.md` |
```
charlesliu9: The impl-json focus on implementation and need to be concise, detailed explanation and specific examples could be added only if necessary.

skills/uipath-maestro-case/references/phased-execution.md
```diff
@@ -197,9 +197,11 @@ Up to **3 validation retries** per session — each retry MUST be preceded by a
 - `Pause for manual edit` — exit skill mid-flight; user edits `caseplan.json` directly and re-runs skill.
```
charlesliu9, line 197: Codex suggested: Abort can erase the incremental journal — skills/uipath-maestro-case/references/phased-execution.md:138 still tells the agent to dump its in-memory issue list to build-issues.md. By then earlier buffers have been cleared and the file already contains journal history, so a whole-file dump can overwrite that history. Abort handling should flush the current buffer into the existing journal and summarize it without replacing prior rows. The same stale wording remains at lines 198 and 322.

charlesliu9 (commented):
comments only

roberts-cliff:
Both addressed in `497764b8d`.

**The abort-path bug is real — good catch, and it was mine.** Three paths still said "dump":

```
:138  On Abort            "Dump in-memory issue list to tasks/build-issues.md"
:198  Phase 4 retry Abort "exit; dump build-issues.md"
:322  shared cleanup      "Dump build-issues.md."
```

Under the incremental contract that isn't a no-op, it's data loss. By abort time the journal already holds rows from every completed section *and* the per-section buffer has been cleared — so a whole-file dump writes a file containing only the current buffer and destroys the history. That's precisely the failure this branch exists to prevent, reintroduced on the one path where someone most needs the record.

All three now append-and-summarize, with the reason stated inline so it can't drift back. Also fixed a fourth stale pointer at `:184` ("proceed to Phase 4 dump step" → issue-log summary step). The remaining `dump` occurrences at `:17`, `:102`, `:128` are about upload responses and validate output — unrelated, left alone.

**On conciseness — agreed, and it collided with the repo's own token-optimization rule.** `§ Why the log is incremental` is cut from 15 lines to 1. The operative instruction stays inline:

> **Flush grain: the section boundary.** Not per issue (one Edit per finding, fights the per-section batch write contract); not per build (a whole-build buffer is lost to context pressure before it is written, and no Step 12 check reads the log). A section boundary already carries a validate, so the flush rides an existing seam and bounds loss to one section.

The two-model measurement that motivated it moved to dev notes. Worth saying why this doesn't fight the adjacency work: what needs to be adjacent to the write site is the **constraint**, not the argument for it. Rationale one section away costs nothing; a rule one section away measurably does.

On that — the N=10 experiment finished and it's a gradient, not a cliff: constraint absent 1/10, one section away 6/10, in the write-site block 10/10. My earlier N=3 baseline of 0/3 was a 1-in-16 draw from a 60% rate, so anything I said about a distant rule being "dead weight" was wrong. Relevant here only as the reason the trimmed rule stays *in* the flush section rather than being cross-referenced from it.

skills/uipath-maestro-case/references/implementation.md
```diff
@@ -49,13 +50,15 @@ Full contract — recovery, tool primitive selection (Edit default, whole-sectio
...
 
 ```text
-# pseudocode — kept in the agent's reasoning, not on disk
-issues = []  # shared across all steps
+# pseudocode — per-section buffer, flushed at each section boundary
+issues = []
 ```
 
+> **Why incremental.** A list held in reasoning for the whole build is lost to context pressure before it is ever written, and no Step 12 check reads the log. Measured on two independent runs of the same task: `claude-sonnet-5` (52.4 min, 39 placeholder tasks, 120 `<UNRESOLVED>` markers) and `gpt-5.6-terra` (7.2 min, 23 placeholder tasks, 0 markers) — **neither wrote `build-issues.md` at all**, and both looked like builds a reviewer would accept. Flushing per section bounds worst-case loss to one section and rides the validate seam that already exists there.
```
charlesliu9: the measurement history etc. needn't to be included here.

roberts-cliff:
## N=10 per model, on the current branch

`logging_issue_log`, run locally against `497764b8d` — so this measures the branch **including** the two review fixes, not the state the earlier N=3 covered.

| Model | Runs | `build-issues.md` | `## Journal` | summary filled | reconstructed | verdict |
|---|---|---|---|---|---|---|
| `gpt-5.6-terra` (codex) | 10 | 10/10 | 10/10 | 10/10 | **0/10** | **10/10 SUCCESS 1.0** |
| `claude-sonnet-5` | 10 | 10/10 | 10/10 | 10/10 | **0/10** | **10/10 SUCCESS 1.0** |

Graded at artifact level rather than by task score — each run's `build-issues.md` was parsed for the incremental format specifically:

- **`## Journal` present** — an old-style end-of-build dump fails this, which is what makes it a regression test rather than a file-exists check.
- **Summary block filled**, not left at its `_Summary written at Step 12.1._` placeholder.
- **Reconstructed 0/20.** No run took the § Recovery fallback, so the journal was genuinely written incrementally rather than rebuilt from artifacts at the end. This is the number I most wanted to hold — a passing suite where runs quietly land in Recovery would mean the flush is not happening and the fallback is masking it.

Durations: terra 3.5–6.4 min, sonnet 7.0–11.1 min. Arms ran concurrently, so these are not comparable across models.

### Why re-run rather than cite the earlier 6/6

The trim in `497764b8d` cut `§ Why the log is incremental` from 15 lines to 1. Given how little terra reads before it commits — measured at ~3 blocks on the sibling PR — removing 14 lines from the file that carries the flush instruction is exactly the kind of edit that can silently cost compliance. It did not: terra is 10/10 on the trimmed doc. The operative instruction stayed inline in § Flush; only the rationale moved to notes.

### Incidental observation

Mean journal rows: terra **1.0**, sonnet **5.4** (range 1–8). The task contains one deliberate validation error, so 1 row is the minimum correct answer and terra hits it exactly every time; sonnet additionally logs WARNING/SKIPPED-class observations. Both satisfy the contract — worth noting only because the log's *completeness* appears to track read depth the same way rule compliance does, which is not something the check currently grades.

skills/uipath-maestro-case/references/implementation.md
roberts-cliff: Done in `faf36aba3` — measurement history removed.

The blockquote had grown into a case for the change rather than a statement of it: two run durations, placeholder counts, marker counts, and both model names. That is PR evidence, and it was sitting in a file an agent reads mid-build.

Trimmed to the causal reason only:

> **Why incremental.** A whole-build buffer held in reasoning is lost to context pressure before it is ever written, and no Step 12 check reads the log. Flushing per section bounds worst-case loss to one section and rides the validate seam already at that boundary.

**Nothing an agent acts on moved.** The operative constraint is the paragraph directly above it — buffer flushes at every section boundary, then clears, not a whole-build accumulator — and that is untouched. I kept one line of *why* rather than deleting outright, on the same reasoning as the `impl-json.md` trim: a rule with no stated reason tends to get optimised away by the next person who reads it as redundant.

I also swept the rest of the skill for the same pattern — `grep` for the durations, counts, and model names returns nothing anywhere under `skills/uipath-maestro-case/`. This was the last instance.

`skills:validate` clean (26 skills, default + studioweb) · `pytest tests/tasks/uipath-maestro-case/` → 93 passed.

Worth noting for the record: the N=10 results I posted were measured on `497764b8d`, before this trim. This commit only removes prose from a rationale blockquote, so I have not re-run — but flagging it rather than implying the numbers cover this exact SHA.

roberts-cliff:
## Post-trim verification: 5/5 on `faf36aba3`

Re-ran `logging_issue_log` on terra after the rationale trim, since the edit touched the file carrying the *initialize* instruction:

| | Runs | `## Journal` | summary filled | reconstructed | verdict |
|---|---|---|---|---|---|
| `gpt-5.6-terra` | 5 | 5/5 | 5/5 | **0/5** | **5/5 SUCCESS 1.0** |

Consistent with the N=10 on `497764b8d`. Removing the measurement history did not cost compliance — the operative constraint was never in that blockquote.

### A methodology correction worth recording

My first attempt at this verification returned **0/5**, and I nearly reverted a correct commit on the strength of it.

It was wrong. Both that run and the matched pre-trim control had silently fallen back to coder-eval's **built-in** default experiment instead of this repo's `experiments/default.yaml`. The built-in has no `plugins:` block, so **the skill under test was never installed** — I was measuring a bare agent, twice, and getting a stable-looking 0/5 each time.

The two report headers side by side:

```
correct:  # Experiment Report: skill-tests-default
          "Default config — tempdir, full lifecycle, devs locally"
wrong:    # Experiment Report: default
          "Default experiment - provides baseline agent configuration"
```

Fix is `-e experiments/default.yaml` explicitly; the fallback is silent.

Three things I take from it, since they apply to any eval claim on this repo:

1. **A matched control does not catch a shared defect.** I ran a pre-trim control specifically to separate "my edit" from "environment drift." Both arms carried the identical fault, so the control returned identically meaningless numbers and would have *confirmed* the false result. What actually caught it was re-tallying the earlier run from disk and finding 10/10 still reproducible — an inconsistency I could not explain away.
2. **Assert the skill loaded before reading any score.** `task.json` → `agent_config.plugins` is `None` when the fallback fires. One assertion, and it fails instantly on every replicate. Now run as a gate before I read the tally:
   ```
   ASSERTION PASSED: skill plugin loaded on all replicates
   plugins= [{"type": "local", "path": "$SKILLS_REPO_PATH"}]  x5
   ```
3. **A green suite and a red suite fail the same way.** This is the same shape as the `reconstructed 0/N` column I added upthread — a number that looks like a result but is produced by a path that bypasses what is being tested. Worth assuming any eval figure in this repo, including my earlier ones, is wrong until the plugin load is shown.

The N=10 figures posted earlier were run correctly — `skill-tests-default` in the header, plugins populated — and re-tally to 10/10 from disk unchanged.

roberts-cliff:
## Both models re-measured on `faf36aba3` (the shipping commit)

Closing the gap I left earlier: the published N=10 matrix was measured on `497764b8d`, before the `implementation.md` trim. Sonnet had never been run on the head that actually ships. It has now.

| Model | Runs | `## Journal` | summary filled | reconstructed | verdict |
|---|---|---|---|---|---|
| `gpt-5.6-terra` (codex) | 5 | 5/5 | 5/5 | **0/5** | 5/5 SUCCESS 1.0 |
| `claude-sonnet-5` | 10 | 10/10 | 10/10 | **0/10** | 10/10 SUCCESS 1.0 |

Sonnet run as two independent batches of 5 (5/5 and 5/5) rather than one batch of 10 — three-way parallelism against a 1200s `task_timeout`, so a wider batch risks scoring contention as failure.

**Instrument check on every run above**, per the fallback problem described in my previous comment:

```
j-terra5-posttrim-fixed    plugin 5/5   exp=skill-tests-default   OK
j-sonnet5-posttrim         plugin 5/5   exp=skill-tests-default   OK
j-sonnet5-posttrim-b2      plugin 5/5   exp=skill-tests-default   OK
INSTRUMENT CHECK: ALL VALID
```

Combined with the earlier N=10 per model on `497764b8d` (both re-audited, plugin loaded 20/20), the incremental contract now holds across **35 verified replicates** on two models, with **zero** taking the Step 12.1 Recovery fallback.

### One caveat on what these numbers do not show

Journal *shape* is solid — 35/35. Journal *completeness* is not measured, and varies a lot:

```
sonnet post-trim rows per run:  1, 1, 11, 2, 2  |  2, 1, 5, 3, 2
terra  post-trim rows per run:  1, 1, 1, 2, 1
```

The task carries one deliberate error, so **1 row is the minimum correct answer** and terra sits there almost exactly. Sonnet additionally logs WARNING/SKIPPED-class entries, and one replicate logged eleven. Both satisfy the contract; the checker grades that `## Journal` exists and is append-only, not that it caught everything.

That dispersion is pre-existing — it showed at the same magnitude before this change (sonnet mean 5.4 pre-trim vs 3.4/2.6 post-trim, on samples far too small and too spread to read a difference into). Nothing here is a regression, and I would not treat the row-count means as a measurement at all.

Flagging it because a green 35/35 reads stronger than it is: this PR establishes that the log is written incrementally, not that its contents are complete. If completeness matters, that wants its own criterion and its own task.

charlesliu9 (approved):

PR #2605, opened 2026-08-13
test(uipath-maestro-case): guard connector solution-resource emission

abhiram-vad:
## Context

A Maestro Case that calls an Integration Service connector must declare that connector as a **solution** resource — not only resolve it inside `caseplan.json`. That declaration is a chain of projections, and each link is regenerated by a separate command:

```
caseplan.json bindings[]  ->  bindings_v2.json resources[]  ->  resources/*/connection/<connectorKey>/*.json
```

When a link is skipped, **nothing in the toolchain says so**:

| Gate | Result on a case with an undeclared connector |
|---|---|
| `uip solution resources refresh` | `Success` · `Created 0, Imported 0, Skipped 0` · `Warnings: []` · `"Synced 0 resources"` |
| `uip maestro case validate` | `Valid` |
| `uip solution pack` | `Success` — package contains **0** references to the connector |
| `uip solution publish` | `Success` · `State: Ready` |

An **empty** sidecar is quieter than a broken one: it has nothing to resolve, so there is nothing to warn about. (A sidecar naming an unresolvable connection *does* warn — `Connection <uuid> not found in Resource Catalog`.) So neither `Result` nor `Data.Warnings[]` carries the signal.

The failure surfaces only to whoever opens or runs the case, as `<connectorKey> is not found`.

**This was observed on a real build**, not constructed. A case whose caseplan calls `uipath-microsoft-outlook365` eight times across seven tasks packed twice, published, and deployed to a tenant feed with zero connector declarations in the package. Root cause: `bindings_v2.json` was written once by the scaffold as `{"resources": []}` and never regenerated — confirmed from the build transcripts (1 `Add File`, 0 `Update File`).

Reproduced end-to-end under a new solution name and published it, to confirm all five gates return `Success`.

## What this PR adds

**`scripts/check_connection_parity.py`** — read-only, stdlib only, exit 1 = do not pack or publish.

```
A = distinct connectionIds in caseplan.json bindings[] (resource=Connection)
B = Connection entries in bindings_v2.json resources[]
C = emitted resources/*/connection/*/*.json files
D = distinct connectorKeys the caseplan references

require A == B == C, and every key in D to have an emitted resource
```

`D` is not redundant with the count parity: when `A` and `B` are both empty, the comparison passes trivially at 0 while the case still calls a connector. `D` reads both caseplan shapes — the plain `"connectorKey":` field and the `{"name":"connectorKey","value":…}` context-array entry. The failing build uses the second form, and a first cut of this check reported zero connector references on the very build being diagnosed.

**`tests/tasks/uipath-maestro-case/connector_features/connection_parity/`** — e2e regression task that grades the *outcome* via that script rather than asserting a command was typed. The `resources refresh` check is advisory (`pass_threshold: 0`) so an agent reaching parity another way is not docked.

## Verification

Script run against six real solutions on disk:

| Solution state | A | B | C | Exit |
|---|---|---|---|---|
| published broken build | 2 | 0 | 0 | **1 BLOCKED** — sidecar never synced |
| reproduction of it | 2 | 0 | 0 | **1 BLOCKED** — same |
| second build, different failure | 2 | 2 | 0 | **1 BLOCKED** — synced but never emitted |
| known-good build | 2 | 2 | 2 | 0 OK |
| known-good build (2) | 2 | 2 | 2 | 0 OK |
| repaired copy | 2 | 2 | 2 | 0 OK |

Two distinct failure modes caught, no false positives on working builds, verified against real broken and real repaired artifacts rather than synthetic fixtures. Non-case directories skip cleanly. `npm run skills:validate` passes.

## Deliberately not in this PR

- **No skill instruction changes.** Step 12 **Check 7** already names this exact condition — *"including non-empty bindings with empty resources … halt before Phase 4"* — and was loaded in the failing build's sessions but never executed. Adding another instruction would be a third rule where two already didn't fire; enforcement belongs outside the agent's instructions.
- **No blocking `PreToolUse` hook.** `hooks/hooks.json` has no `PreToolUse` precedent, and a hook that blocks `pack`/`publish` for every plugin user is a separate and more contentious concern. Happy to follow up if reviewers want it.
- **The permanent fix is platform-side** and out of scope here: `bindings_v2.json` is a pure projection of caseplan `bindings[]`, and `pack` already reads `caseplan.json`. Either `pack` refuses when a referenced connector has no emitted resource, or it derives connection resources from the caseplan and stops treating the sidecar as an input. That removes divergence instead of detecting it, and covers every producer — other skills, Studio Web, hand-built solutions. This PR is containment; that would be the cure.

## Reviewer notes

- `feature:connections` is used rather than copying `feature:connector-activity` from the neighbouring task — the latter is not in the closed vocabulary in `tests/README.md`.
- The task reuses the Slack connector IDs already committed in `connector_activity.yaml`, since those are known-good on the test tenant. The defect is connector-agnostic; the incident involved Outlook 365.
- `path-to-ga` is a candidate tag for this task (it guards a defect that shipped a broken published package) — left off pending reviewer preference.


song-zhao-25 (approved):

charlesliu9 (approved):

song-zhao-25:
## Review

Verified the core logic against the canonical shapes before reviewing — the field access is exactly right: `resource: "Connection"`, `propertyAttribute: "ConnectionId"` (and the defensive `.lower()` is a good call), and `bindings_v2` Connection `key` *is* the connectionId per `bindings-v2-sync.md`, so `A` and `B` are genuinely comparable. Two things I initially suspected turned out fine: `feature:connections` **is** in the closed vocabulary in `tests/README.md`, and `$SKILLS_REPO_PATH` has precedent across the HITL tasks. The incident-driven verification table (real broken + real repaired artifacts, not synthetic fixtures) is the right way to have built this.

Seven findings.

### Correctness — `scripts/check_connection_parity.py`

**1. Counts are compared, not sets. (Medium)**

The docstring promises `require A == B == C`, but `check()` does `len(a_ids) != len(b_ids)`. Both are sets over the same value space (connection UUIDs). A caseplan declaring `{X, Y}` against a sidecar holding `{Y, Z}` passes at `2 == 2` while being wrong — and that isn't hypothetical, it's the brownfield rebind path in `case-editing-operations.md:373` ("Swap the resource bindings"), where a stale sidecar keeps the *old* connection at the same cardinality.

**Fix:** compare the sets and report the symmetric difference. Free precision, and a much better message than "declares 3 but has 3".

**2. `emitted` is solution-scoped but compared against per-project `b_ids`. (Low-Medium)**

`emitted = sorted(solution_dir.glob("resources/*/connection/*/*.json"))` sits inside the `for caseplan in caseplans:` loop and returns the same solution-wide list on every iteration, while `b_ids` comes from that one project's sidecar. Two case projects with 2 connections each → `B=2` vs `C=4`, and both fail spuriously. The loop explicitly supports multiple caseplans, so either scope the glob per-project or assert single-project explicitly.

**3. Fail-open when no caseplan is found. (Medium, given the stated intent)**

No `*/caseplan.json` → prints `SKIP` → returns `0`. Fine as an eval assertion. But the docstring says *"1 = do not pack or publish"* and the PR floats a `PreToolUse` hook. Wired as a gate, a typo'd path, wrong CWD, or a renamed project directory becomes a green light. For a gate, unknown ≠ pass.

**Fix:** a `--require-caseplan` flag, or exit 2 on SKIP so a caller can distinguish "clean" from "didn't look".

**4. Deduped set vs raw file list. (Low)**

`b_ids` is a set; `emitted` is a list. A sidecar carrying a duplicated Connection entry collapses to 1 against 2 files → spurious failure whose message says "run `uip solution resources refresh`", which won't fix it.

Careful with the obvious patch: `emitted_keys` (connectorKey directories) and `emitted` (files) have **different cardinalities** — two connections on the same connector produce two files under one key directory. The script correctly uses keys for `missing_keys` and files for count parity, but nothing says so. Worth a comment before someone "simplifies" it into a bug.

### Convention

**5. Filename breaks the `scripts/` naming convention. (Low — but worth fixing before merge)**

Every existing script is hyphenated: `check-cli-verbs.py`, `check-skill-status.py`, `check-skill-verbs.py`, `check-skills-sh.py`, `check-task-driver.py`, `build-uip-catalog.py`, `prune-task-defaults.py`. This adds `check_connection_parity.py`. Rename to `check-connection-parity.py` now — the task YAML hardcodes the path, so it only gets more annoying later.

### Test task

**6. The prompt tells the agent the answer. (Medium)**

> Leave the solution in a state that is ready to publish — every resource the case depends on must be declared at the solution level, not only inside caseplan.json.

That's the fix, stated in the prompt. The real incident was an agent that didn't know to sync the sidecar; this grades an agent that was *told* to. It also cuts against `.claude/rules/test-writing.md` § Prompts: *"Minimal. State the goal and the success bar."*

**Fix:** drop that sentence. "Build it and leave it ready to publish" is the goal — knowing that solution-level declaration is part of "ready" is exactly what the skill is supposed to teach, and exactly what regressed.

### Scope

**7. Connection-only — the kinds that fault at runtime are uncovered. (Medium, follow-up not blocker)**

Every filter is `resource == "Connection"` / `resources/*/connection/`. This blocks correctly on the incident in this PR, and on the two sidecar failures in the staging escalation from 17 Aug (`A=3, B=0` in both) — but only because those cases happened to have connectors. Strip them and a case with 2 IXP models and 3 apps against an empty sidecar scores `A=0, B=0, C=0, D=0` → **exit 0**, and the resource-less package ships. That is precisely the shape of that escalation's other two failures, where the runtime errors were `DU response: Model not found` and `No app: <name> found in folder`.

The `D` check is what prevents the trivial zero-vs-zero pass, and there is **no non-connector equivalent**. That's the extension: a caseplan-side scan asserting each `process` / `app` binding (per the Per-Task-Type table in `plugins/variables/bindings/impl-json.md`) has an emitted resource.

**Trap for whoever writes it:** don't generalize `B == C`. Inline-built agent/api-workflow siblings are deliberately not emitted — `create-inline-common.md` § Step 3 notes `resources refresh` **skips** them (`Skipped: already in solution`). A naive count parity false-fails every case with an inline sibling. The exemption keys off the `solution_folder` sentinel in `bindings_v2.json`.

### Verdict

Approve-with-changes. **1** and **5** before merge (both are minutes). **3** and **6** worth resolving here. **7** as a tracked follow-up rather than a blocker — shipping connection coverage now beats waiting for full coverage, given this is the second real incident from this root cause.

One push-back on the description: *"Deliberately not in this PR — no skill instruction changes"* is well argued, and the evidence that Check 7 was loaded and never executed is convincing. But it leaves the containment reaching **only CI**. Nothing in the skill or in `hooks/` invokes this script in a real user session, so someone hitting this tomorrow is no better off than the two builds that already hit it. Worth saying so explicitly in the description, so it doesn't read as solved.

AlvinStanescu (approved):

PR #2606, opened 2026-08-13
fix(maestro-flow): author inline-agent nodes without prompt keys

jiyangzh:
## What breaks

The inline-agent docs teach stub `systemPrompt`/`userPrompt` placeholders on the `uipath.agent.autonomous` flow node. Any prompt string on the node makes `@uipath/flow-converter` (prune present 0.25.1 through 0.42.0) delete every `agentInputVariables[]` entry the prompt text does not reference. Stub text references nothing, so a correctly wired agent starts on `JobArguments: {"input":""}` and faults:

- `AGENT_STARTUP.INPUT_VALIDATION_ERROR` ("Field required" per schema key) when the agent's `inputSchema` marks inputs required
- `AGENT_RUNTIME.TERMINATION_LLM_RAISED_ERROR` when it does not

This hits debug AND deploy — the CLI converts at pack time and the packed BPMN is the production runtime artifact. Two DevCon nightly failures on 2026-08-12/13 came from flows authored exactly per these docs.

## Live proof (eval CLI 1.201.0-dev.8247, real tenant)

| Variant | Result |
|---|---|
| As-authored (stub prompts, byte-identical to the failing eval artifact) | Faulted, 95s — `JobArguments {"input":""}`, three "Field required" |
| Identical flow, only the two prompt keys deleted | Completed, 20s — all three inputs delivered; output email cites invoice MCS-2026-04872 and credit 1610 |

`diff -r` confirms the variants differ by exactly the two keys; `agent.json` byte-identical.

## Change

Teach the prompt-less node shape: `source` + `agentInputVariables` only. Delete the prompt keys — never set them to `""` (`flow validate` rejects empty strings: `[REQUIRED_FIELD]`).

- `skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md` — both node JSON examples lose their prompt lines; the "keep short generic placeholders" rule and its echoes reverse to "write no prompt keys"; the wiring section documents the prune with version anchors (verified on uip 1.199.0-dev.7959 validate, 1.201.0-dev.8247 debug); validator behavior reframed as verified-not-guaranteed (the registry `inputDefinition` marks both fields required — absent-key acceptance is a validator gap); new troubleshooting row maps `{"input":""}` + both error signatures + incident 170002 to the prune, with `uip agent refresh --inline-in-flow` as the tool-based fix; Studio Web round-trip rule added (canvas mirrors prompts back — safe only when they reference every key).
- `skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md` — the copy-paste node JSON loses its stub prompts; the "validation requires non-empty placeholders" bullet is replaced with the prune rule and a pointer to impl.md. impl.md links to this page four times, so the two must agree.

The prompt-less shape matches what the platform already converges to: `uip agent refresh` shell-ify strips node prompts and keeps structural inputs, and Studio Web re-hydrates the canvas from `agent.json`.

## Defense layering

- This PR fixes authoring at the source — new flows work on the currently shipped CLI, no release dependency.
- UiPath/cli#3653 guards all four CLI conversion paths (debug, solution pack/publish, flow pack/publish, eval) for existing flows.
- MST-13601 tracks the converter-side fix (the prune judges wiring by prompt tokens and never reads `agent.json`).

## Test-side notes (no code change here)

- `tests/tasks/uipath-maestro-flow/_shared/advisory_billing_resolution_writer.py:80` fails with "the agent node carries no systemPrompt/userPrompt". The check is safe — `agent_prompt_text` also reads the sidecar messages, so a prompt-less node with a real sidecar passes. Only the message wording now misleads; reword in a follow-up.
- Three reference flows still carry stub node prompts (`BillingDisputeAnalyst.reference.flow:38`, `BillingResolutionWriter.reference.flow:38`, `BillingDisputeResolution.reference.flow:494`). Reference artifacts, not graded assertions — nothing breaks. Strip them in a follow-up so references match the documented shape.


skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
```diff
@@ -240,7 +238,7 @@ Resource body shape is identical to the standalone-agent docs — only the folde
 
 **Critical fields:**
 - `inputs.source` — The inline agent's `projectId` UUID. Must match the subdirectory name and `agent.json.projectId` inside the flow project. The definition still declares `model.source: true`, but flow-core hoists that identity field onto `inputs.source` for the `uipath.agent.autonomous` node instance.
-- `inputs.systemPrompt` / `inputs.userPrompt` — Current flow validation requires non-empty placeholders on the node. The canonical prompts still live in `agent.json.messages[]`.
+- `inputs.systemPrompt` / `inputs.userPrompt` — **do not write these keys.** A prompt string on the node makes the flow converter drop every `agentInputVariables[]` entry the prompt text does not reference. Empty strings fail `uip maestro flow validate`. The canonical prompts live in `agent.json.messages[]`. Validator behavior and version anchors: [inline-agent guide § Refresh and Validate](../../../../../uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md#refresh-and-validate).
```
bryan-uipath: With the CLI version bump https://github.com/UiPath/cli/pull/3618 - that address the root issue - do we still need to say "do not write these keys?". I think, generally, we _would_ want the agent to write these keys (if there wasn't a bug)

bryan-uipath (approved):
Will approve to unblock, but I think ideally we could simplify this:
1. Agent should just be responsible for writing the `inputs.userPrompt` / `inputs.systemPrompt` - referencing variables as needed
2. variable mapping / prompts get _pushed to `agent.json` at the appropriate times mechanically - claude/codex shouldn't have to worry about this

Discussion [here](https://uipath-product.slack.com/archives/C0ARJ9NHYVB/p1786659543865979?thread_ts=1786657019.327809&cid=C0ARJ9NHYVB)

skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
jiyangzh: with the version bump

- {{input.X}} tokens on node prompts (Studio Web exports) ✅  
- stub prompts + sidecar ❌ 


agent authored inline agents are all following second path

and at this point what prompt written in flow file is always ignored

skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md
```diff
@@ -47,10 +47,14 @@ After editing `content`, rebuild the matching `messages[].contentTokens` (`type:
...
+Passing flow data into an inline agent requires **three hand-authored, aligned** pieces. **The CLI does not derive the input wiring** — `uip agent refresh` does **not** scan prompts, derive `inputSchema`, or populate `agentInputVariables`; you author all three, and packaging ships them as-authored. (Refresh *does* regenerate `messages[].contentTokens` from `content` — that's the one derived part; see the invariant below.) The converter builds the runtime `JobArguments` from the **flow node's `inputs.agentInputVariables[]`** (not from `$vars` tokens in `agent.json`). Any `inputs.systemPrompt` / `inputs.userPrompt` string on the node makes the converter drop every `agentInputVariables[]` entry the prompt text does not reference (`@uipath/flow-converter`; prune present 0.25.1 through 0.42.0). Flatten rule: `$vars.<trigger>.output.<var>` → `<trigger>__output__<var>`.
+
+Verified on uip 1.199.0-dev.7959 (validate) and 1.201.0-dev.8247 (debug). If `uip maestro flow validate` rejects a prompt-less node, upgrade the CLI. Never add stub prompts.
```
rockymadden: Do we need to call out the versions? This will get stale quickly.

skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md
```diff
@@ -181,6 +183,8 @@ Use `Edit` to add a node instance to `nodes[]`. The instance carries only per-in
 }
 ```
 
+No `inputs.systemPrompt` / `inputs.userPrompt` keys. Verified on uip 1.199.0-dev.7959 (validate) and 1.201.0-dev.8247 (debug). If `uip maestro flow validate` rejects a prompt-less node, upgrade the CLI. Never add stub prompts.
```
rockymadden: This advice might not be possible for a time depending upon the env that the user is connected to.

skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md
jiyangzh: Agreed — removed. The three anchor lines now carry a behavior self-check instead of a build number: `uip maestro flow validate` must pass a prompt-less node and must fail empty-string prompts. The agent runs that against the CLI it actually has, so the guidance never goes stale.

One dated note survives, in § Refresh and Validate only ("verified 2026-08"), so the provenance of the validator-gap claim stays readable. Four copies of a dev-build number are gone.

skills/uipath-maestro-flow/references/author/references/plugins/inline-agent/impl.md
jiyangzh: Correct, and "upgrade the CLI" is now gone from the anchors. New § Refresh and Validate § Older CLI gives the pinned-environment path:

1. Add minimal placeholder prompt keys — only after validate rejects the absent keys. Never `""`; empty strings fail the same check.
2. Run `uip maestro flow debug` and read the trace. `JobArguments` must carry the bound inputs, not `{"input":""}`.
3. If it is `{"input":""}`, that same CLI carries the converter prune (`@uipath/flow-converter` 0.25.1+). The placeholders satisfy the old validator and delete the wiring at the same time.

The tradeoff is named rather than hidden: on a CLI old enough to reject absent keys and new enough to prune, no node edit works. The doc tells the agent to escalate for an upgrade instead of shipping a broken flow.

skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
jiyangzh: Supporting the reply above: today the runtime never reads node prompts. The converter uses the node only to emit `JobArguments`, and the agent runtime takes its prompts from `agent.json`. So until the mechanical push to `agent.json` exists, a node prompt buys nothing and only triggers the prune. The design you describe is the right end state — the platform's self-contained transition. The converter-side recovery that protects declared wiring in the meantime is flow-workbench#3068 (in review, tracked as MST-13601), and this doc flips back to "write the keys" when the push mechanism lands.

nikhil-maryala (approved):

andreibalas-uipath (approved):

rockymadden (approved):

PR #2607, opened 2026-08-13
docs(uipath-insights): split SKILL.md into a router over reference guides

celestinryf:

## What

Splits `skills/uipath-insights/SKILL.md` from 316 lines into an 81-line router over reference guides. The router keeps the rules that apply to every Insights command plus a task-to-guide table; command detail moved into `references/`.

- Jobs subcommand detail and the absolute time-range workflow to `references/jobs-commands-guide.md`
- The Filter Discovery section added in #2552 to `references/filter-discovery-guide.md`
- Review round: the investigation playbook now cross-links the jobs guide's Absolute Time Ranges and repeatable-options rules from the steps that need them (Playbooks 5 and 6)

Facts corrected against the CLI source while moving them:

- Response `Data` keys are PascalCase on the wire. `OutputFormatter.success` runs `normalizeDataKeys` and no insights command passes `preserveDataKeys` (`formatter.ts:326-341`, `:939`), so the guides now say `FolderKey` and `JobsCount`. They previously said `folderKey` and `jobsCount`, which match no response the CLI produces.
- Jobs responses carry no `Pagination` and no `Instructions` (`jobs.ts:180-184`). Only `filter-*` does (`list-executor.ts:125-137`).
- Every failure envelope carries `Retry`, which answers the retry question directly (`formatter.ts:961`).
- On jobs commands `ErrorCode` is always `unknown_error`, even for 401 and 403 (`client.ts:34-36`, `formatter.ts:863`). A rejected flag is a different shape, `ValidationError` with `invalid_argument`.
- A missing time range is rejected locally and exits 1 before any request (`jobs.ts:139`, `:152`, `:167`), so the old 500 troubleshooting row cannot happen.
- `--output` accepts `table, json, yaml, plain`; `--limit` defaults to 50 with a 10000 cap (`globalOptions.ts:7`, `list-options.ts:6`).

## Why

Every activation paid for command families the task did not use. An alerts or RBAC question loaded the full Jobs reference before doing anything. Routing first keeps each activation to the guide that matters, and it lands before the alerts and RBAC guides stacked behind it.

## Testing

`npm run skills:validate` (25 skills, 1707 files), `npm run skills:build`, `hooks/validate-skill-descriptions.sh`, `scripts/check-skill-status.py`, and a link check over all four files. All pass.

Verb gate: 0 blocking, 10 soft. Eight are `filter-*` verbs, in cli main since #3449 but not in the pinned catalog snapshot (`1.200.0-dev.8173`). One is `login tenant set MyTenant`, where the parser reads the argument as part of the verb path. The last is the walker parsing the frontmatter prose as a verb. All three predate this PR.

The six insights smoke tasks passed on the pre-review commit. They have not been re-run since the field-casing and structure corrections, so that result does not cover the current diff.

## Limitations

No eval task covers the routing behavior itself, and none asserts response field casing, which is the bug this PR fixes. Both are worth a follow-up task.

The CLI repo's own `scenarios/insights-tool/jobs-monitoring.md` documents these fields in camelCase and is stale for the same reason. That needs a separate fix in `cli`.

## CI note

Stacked on #2552 (`feat/uipath-insights-filter-discovery`). Rebase onto main once #2552 lands.

### DIFF
--- a/skills/uipath-insights/SKILL.md
+++ b/skills/uipath-insights/SKILL.md
@@ -5,312 +5,77 @@ when_to_use: "User says 'job failures', 'automation health', 'job success rate',
 allowed-tools: Bash, Read
 ---
 
-# UiPath Insights — Job Monitoring Agent Skill
+# UiPath Insights
 
-Insights provides analytics and monitoring for UiPath automation execution. This skill covers **job monitoring** — querying aggregated job execution data for dashboards, health checks, and failure investigation.
+Use `uip insights` for job monitoring and monitoring-scope discovery. Read the guide for the task before running commands.
 
-Job monitoring goes through `uip insights jobs <subcommand> --output json`; scope discovery goes through the `uip insights filter-*` groups. <!-- uip-check-skip -->
+## When to Use This Skill
 
----
-
-## When to Use
-
-- Checking overall automation health (how many jobs ran, how many succeeded)
-- Investigating job failures (which processes fail most, what are the failure reasons)
-- Monitoring job execution trends over time (completed/uncompleted timelines)
-- Getting per-process performance breakdowns
-- Drilling into specific failure details for investigation
-- Discovering which folders, processes, queues, or machines have recent activity, to pick an exact scope (see Filter Discovery Commands)
-
-> **Not in scope:** Starting, stopping, or managing individual jobs (use `uip or jobs` via uipath-platform). Root-cause debugging of a specific job's error (use uipath-troubleshoot). Queue item metrics, robot utilization, or dashboard CRUD (not yet available in CLI).
-
----
-
-## Login & Tenant Setup
-
-**Default to Production. Only switch environment/org/tenant when explicitly stated in the request.**
-
-- If the request mentions no environment, use the current session (defaults to prod `cloud.uipath.com`)
-- If the request explicitly names an environment/org/tenant, check `uip login status` and re-login if needed
-
-```bash
-# Check current environment, org, and tenant
-uip login status --output json
-
-# Login to a specific environment (production cloud is the default)
-uip login --authority https://cloud.uipath.com --tenant MyTenant
-
-# Switch tenant within the same environment
-uip login tenant set MyTenant
-```
-
----
+- Job health, success rate, failure count, or processing time across a tenant, folder, or process
+- Job trends over time, or a comparison between two periods
+- Which processes fail most, and which failure reasons recur
+- Whether jobs are stuck, pending, or still running
+- Finding the exact folder key, process name, queue name, or machine name to scope a query by
 
 ## Critical Rules
 
-1. **A time range is always required — for `jobs` commands only.** Every `uip insights jobs` command needs either `--time-range <minutes>` (relative) or both `--started-after <epoch-ms>` and `--started-before <epoch-ms>` (absolute). Without one, the command fails. `filter-*` commands take no time flags at all (see Filter Discovery Commands). For absolute ranges, pass literal epoch-millisecond numbers, resolved beforehand (see Workflow: Absolute Time Range). Common values:
-   - `--time-range 60` — last 1 hour
-   - `--time-range 1440` — last 24 hours
-   - `--time-range 10080` — last 7 days
-   - `--time-range 43200` — last 30 days
+1. **Use `--output json`.** `jobs` commands return `{ Result, Code, Data }`. `filter-*` commands add `Pagination` and `Instructions`; quote those `Instructions` in the explanation. A failure envelope carries `Result`, `Message`, `Instructions`, `ErrorCode`, and `Retry`, with no `Code` and no `Data`. Keys inside `Data` are PascalCase on the wire, so read `FolderKey` and `JobsCount`, not `folderKey` or `jobsCount`.
+2. **One subcommand per invocation, written literally.** Do not chain, loop, or parameterize `uip insights` commands: no `&&` or `;` chains, no `for` loops, and no shell variables holding the subcommand name or flag values. Resolve values such as epoch timestamps in a separate command first, then pass literal numbers. Never write `$(date ...)` or `$VAR` into a flag value.
+3. **Use only the flags the guides document.** Identity, organization, and tenant come from the active session. Any tenant flag you find is deprecated and is rejected outright on `filter-*` commands, so do not use one. If a filter is not in the guide's shared-options list, it does not exist.
+4. **Every `jobs` command needs a time range.** Pass `--time-range <minutes>` (60 = 1h, 1440 = 24h, 10080 = 7d, 43200 = 30d), or both `--started-after` and `--started-before`. Omitting both is rejected locally and exits 1. `filter-*` commands take no time flags.
+5. **Start with `summary`, then drill down.** After any scope discovery the task needs, begin a job investigation with `uip insights jobs summary` for the totals, then run the targeted subcommands. The summary supplies the denominator that makes a failure count meaningful.
+6. **Treat empty data as bounded evidence.** Empty results can reflect the chosen time window, the recent-activity window, caller visibility, or tenant provisioning. They do not prove that a resource or event never existed.
+7. **Use the CLI instead of raw Insights APIs.** It owns authentication, tenant routing, validation, and error handling.
+8. **Do not retry automatically.** Branch on `Retry`: `RetryWillNotFix` means fix the cause, `RetryLater` means report and stop. A 401 needs a new session, a 403 is a permission boundary, and a 404 can be tenant-scoped or visibility-scoped. Each guide documents the error shapes for its own commands.
+9. **Never run `uip login` yourself.** It opens an interactive browser flow that will hang the session. Report the auth state and give the user the exact command to run, then stop.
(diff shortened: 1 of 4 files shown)


celestinryf:
Both findings are fixed in [`5d426b0d`](https://github.com/UiPath/skills/commit/5d426b0db).

**Troubleshooting table.** Back in `jobs-commands-guide.md`, between Absolute Time Ranges and API Details. One row is corrected rather than restored: the 500 entry cannot happen. `jobs.ts:139` rejects a missing time range and exits 1 before `createInsightsConfig` (`:152`) and `insightsPost` (`:167`), so the request never reaches the server and no status comes back. That row is now the local validation error, and a sentence under the table explains why, so the 500 claim does not get added back later. The other rows are checked against source too. The not-logged-in message has two variants (`authContext.ts:46-50`), so the symptom column matches on the stable prefix.

**Per-endpoint `Code` values.** Back as a table under Response Envelope, where `Code` is introduced, rather than a line in each of the seven command sections. All seven values read out of `jobs.ts`.

On the two manual-review items: the routing eval is a real gap and stays in Limitations. The `filter-*` verb findings are already resolved upstream, since cli#3449 merged on 2026-08-12; the pinned catalog snapshot (`1.200.0-dev.8173`) has not refreshed yet.

Section ordering stays as it is, for the reason you gave.

skills/uipath-insights/references/filter-discovery-guide.md
```diff
@@ -0,0 +1,68 @@
...
+
+### filter-folders list
+
+List folders with recent Insights activity that are visible to the current caller.
+
+```bash
+uip insights filter-folders list --output json
+```
+
+**Key Data fields:** `folderName`, `folderKey`
```
rliuup: issue (blocking): These row fields land PascalCased on the wire - `OutputFormatter.success` runs `normalizeDataKeys` on `Data` and nothing in `insights-tool` or `insights-sdk` opts out via `preserveDataKeys` (`packages/common/src/formatter.ts:326-341`, `:939-942`), so an agent reading `Data[].folderKey` gets `undefined`. The CLI pins this in its own e2e - `expectRowShape(result.Data, ["FolderName", "FolderKey"])` under the comment "OutputFormatter.success normalizes Data keys to PascalCase" (`packages/insights-tool/tests/filter-discovery.e2e.test.ts:17-18,55`), with `jobs-monitoring.e2e.test.ts` asserting `JobsCount` / `ProcessName` the same way. `filter-folders.spec.ts` reads camelCase only because it mocks `@uipath/common`, pinning the object passed *into* the formatter rather than the emitted JSON - same change needed on lines 31, 43, 55 here and on the seven `Key Data fields` lines this PR adds to `jobs-commands-guide.md`.

skills/uipath-insights/SKILL.md
```diff
@@ -5,312 +5,60 @@ when_to_use: "User says 'job failures', 'automation health', 'job success rate',
...
-### failure-details
-
-Get detailed failure information for drill-down investigation.
-
-```bash
-uip insights jobs failure-details --time-range 1440 --output json
-```
-
-**Key Data fields:** `processName`, `machineName`, `processExceptionReason`, `startTime`, `endTime`
+1. **Use `--output json`.** Insights commands return an envelope such as `{ Result, Code, Data, Pagination, Instructions }`. Read `Data` for the result and preserve relevant `Instructions` in the explanation.
```
rliuup: suggestion (non-blocking): This envelope is the union of two families rather than either one - `jobs` emits `{Result, Code, Data}` only (`jobs.ts:180-184`), while `Pagination` and `Instructions` come from `list-executor.ts:106-112` on the `filter-*` path. Rule 6 generalizes the same way: the contract and catch-all branches in `insights-tool/src/utils/errors.ts:257-271` fall through to `unknown_error` too, so "`filter-*` failures carry a specific `ErrorCode`" holds for classified auth/HTTP/network failures but not all of them. Worth splitting both by family, since the router is the only file every task is guaranteed to load.

skills/uipath-insights/SKILL.md
```diff
@@ -5,312 +5,60 @@ when_to_use: "User says 'job failures', 'automation health', 'job success rate',
...
 
-Job monitoring goes through `uip insights jobs <subcommand> --output json`; scope discovery goes through the `uip insights filter-*` groups. <!-- uip-check-skip -->
+## Task Navigation
 
----
-
-## When to Use
+| User's task | Read first |
+|---|---|
+| Check job health, success rate, trends, failures, stuck jobs, or compare periods | [`references/investigation-playbook-guide.md`](references/investigation-playbook-guide.md) |
```
rliuup: suggestion (non-blocking): "compare periods" routes only to the investigation guide, but Playbook 5 leaves `<last-monday-epoch-ms>` unresolved and both the platform-specific `date` conversion and the exclusive-end-boundary rule live only in `jobs-commands-guide.md:229-252`, so that path can silently query a window a day short. The repeatable-filter rule (`jobs-commands-guide.md:22`) is unreachable the same way for multi-process or multi-folder investigations. Either send those rows to both guides, or cross-link the Jobs guide from the playbook steps that need it.

skills/uipath-insights/SKILL.md
```diff
@@ -5,312 +5,60 @@ when_to_use: "User says 'job failures', 'automation health', 'job success rate',
...
-- **Don't chain, loop, or parameterize `uip insights` subcommands in one invocation.** Run each subcommand as its own command with the subcommand name written literally — no `&&`/`;` chains, no `for` loops, no shell variables holding the subcommand name. One command per invocation keeps each subcommand's exit status and output attributable.
-- **Don't start, stop, or manage individual jobs.** This skill is for monitoring and analytics only. Use `uip or jobs start/stop` via uipath-platform to manage jobs.
-- **Don't construct raw API calls to the Insights endpoint.** The CLI handles auth headers (`X-UiPath-Internal-AccountName`, `X-UiPath-Internal-TenantName`), URL construction, and error handling. Hand-rolling `curl` or `fetch` calls will miss these.
-- **Don't retry on auth errors.** If `uip insights jobs` returns 401 or "Not logged in", the fix is `uip login`, not retrying the same command.
-- **Don't use this skill for root-cause debugging.** "Why did job X fail with error Y?" is a troubleshooting question — hand off to uipath-troubleshoot. This skill answers "which processes fail the most and what are the common reasons."
-
----
+## Scope Boundaries
```
rliuup, line 62: suggestion (non-blocking): The old "Not in scope" note also excluded robot utilization and dashboard CRUD; this table keeps only queue metrics. `insights dashboard` is still a `when_to_use` trigger, so a dashboard request activates the skill and then finds no boundary saying the CLI does not do that. Two more rows would close it.

rliuup (changes requested):
The restructure itself is clean and the move is lossless where I checked it, but the field names the new guide documents are not the ones the CLI emits: `OutputFormatter.success` PascalCases `Data`, so `folderKey` / `jobsCount` and their siblings come back as `FolderKey` / `JobsCount`. Details and the CLI's own e2e assertions are in the inline comment on `filter-discovery-guide.md`. Three smaller notes on the router are non-blocking.

skills/uipath-insights/references/filter-discovery-guide.md
celestinryf: Fixed in [`a40a83f2`](https://github.com/UiPath/skills/commit/a40a83f25aef2d9906b404f367d5f7c2159c4b67). The rows here and the seven `Key Data fields` lines in `jobs-commands-guide.md` read PascalCase now, and the filter guide states the rule at the top. SKILL.md Rule 1 carries the same rule, so the router repeats it for every activation.

skills/uipath-insights/SKILL.md
celestinryf: Fixed in [`a40a83f2`](https://github.com/UiPath/skills/commit/a40a83f25aef2d9906b404f367d5f7c2159c4b67). Rule 1 now states the two families separately: `jobs` returns `{Result, Code, Data}`, and only `filter-*` adds `Pagination` and `Instructions`. The `ErrorCode` claim is gone; Rule 8 defers each family's error shapes to its own guide.

skills/uipath-insights/SKILL.md
celestinryf: Fixed in [`783aee62`](https://github.com/UiPath/skills/commit/783aee6253f54b610e26122d8cb5f367a51f7acd). Went with the cross-link option: Playbook 5 now says to resolve both boundaries with the Absolute Time Ranges recipes in the jobs guide and to treat `--started-before` as exclusive, and Playbook 6 points at the repeatable-options rule for multi-folder or multi-process scope.

skills/uipath-insights/SKILL.md
celestinryf: Fixed in [`a40a83f2`](https://github.com/UiPath/skills/commit/a40a83f25aef2d9906b404f367d5f7c2159c4b67). The scope table gained a row sending alert definitions, dashboards, and robot utilization to "Not in the shipped `uip insights` surface", so the `insights dashboard` trigger now lands on a stated boundary. One row covers all three rather than two rows.

rliuup (approved):

PR #2608, opened 2026-08-13
feat(uipath-insights): teach alert definition reads and add an alerts smoke eval

celestinryf:

## What

Teaches the three alert definition read commands: `alerts list`, `alerts get`, and `alerts check-entitlement`.

- New `references/alerts-reads-guide.md`: per-command usage, key `Data` fields, interpretation rules, and a Safe Output note.
- `SKILL.md`: one routing row, the alert clauses in `description` / `when_to_use`, and the read-only critical rule. That rule lives in SKILL.md only. The guide does not restate it, matching the repo pattern where 4 of 151 reference guides define their own `## Critical Rules` block.
- New `tests/tasks/uipath-insights/alerts/smoke.yaml` and two activation rows (030-031).

## Why

The alert read commands ship in two cli PRs: [UiPath/cli#3610](https://github.com/UiPath/cli/pull/3610) (merged) has the three definition reads, and [UiPath/cli#3611](https://github.com/UiPath/cli/pull/3611) (open, one approval, awaiting a second reviewer) has the history and delivery reads. This PR covers only the #3610 half so its smoke run does not wait on #3611. The #3611 half is #2739, a separate draft PR that bases on `main`. Without the guide the skill either says nothing about the commands or an agent invents flags.

## Scope

One divergence from convention, stated in the guide with its backend reason: `--agentic` pairs with `--process-key` rather than being its own subcommand.

Also carries the two folder-scope facts raised on #2552: `alerts list` shows folder scopes as names while `alerts get` and the agentic route show keys, and `filter-folders list` maps between them.

## Testing

Ran locally on this branch: `npm run skills:validate`, `hooks/validate-skill-descriptions.sh`, `scripts/check-skill-status.py`, `scripts/check-task-driver.py`, `scripts/check-cli-verbs.py` on the smoke task, and link resolution. All pass.

The smoke task's negative criteria were unit-tested against hand-written command strings, including the batched-command case that caused the `re.DOTALL` false-fail on #2552 and the backslash-newline continuation case. All pass.

The task is tenant-agnostic: it passes against an entitled tenant and an unentitled one. Its `run_command` gate uses `check-entitlement` because that is the one alert read returning 200 on any authenticated tenant. The gate's exit-code assumption is now verified: `bun run start insights alerts check-entitlement --output json` on cli main exits 0, and `executeCheckEntitlement` (`packages/insights-tool/src/utils/alerts.ts`) writes a success envelope for both `true` and `false` with no failure exit, so a `false` also exits 0.

The task gates 2 of the 3 alert reads. `alerts get` is left out on purpose: it needs an ID that does not exist on an empty tenant, so gating it would trade tenant-agnosticism for surface coverage. The task file says so, and an integration-tier task against a seeded tenant is the right home for it.

## Limitations

**Smoke run: passing.** Run [32783473599](https://github.com/UiPath/skills/actions/runs/32783473599) (Smoke Skill Tests) succeeded on head `c8ab5ac`; rliuup confirmed `skill-insights-smoke-alerts-reads` scored 9/9, weighted 1.000, with the `run_command` gate executing `uip insights alerts check-entitlement --output json` to exit 0. The `@uipath/cli@dev` build carrying the alerts group is `1.202.0-dev.8372` (published 2026-08-24). Two commits since that run change one guide sentence and narrow one negative criterion; the suite re-runs on the current head.

Every command, flag, and `Key Data fields` claim in the guide was re-checked line by line against `packages/insights-tool/src/commands/alerts.ts` and `utils/alerts.ts`, now on cli main. Alert definition reads were confirmed live on `alpha/popoc/adetenant`; `check-entitlement` was re-run against cli main for the exit-code check above.

## How to Review

Check the guide's command list and flags against `packages/insights-tool/src/commands/alerts.ts` on cli main. Keys in the CLI's JSON output are PascalCase; the RTM service wire is camelCase and the projections read camelCase (`utils/alerts.ts`).

## CI note

The history and delivery half is #2739, a separate draft PR based on `main`; it stays draft until cli#3611 merges and publishes.


skills/uipath-insights/references/alerts-reads-guide.md
```diff
@@ -0,0 +1,130 @@
+# Alert Read Commands
+
+Reference for the `uip insights alerts` read commands, with response shapes and interpretation rules. All three use the active CLI session for identity and tenant.
+
+Keys inside `Data` are PascalCase in the CLI's JSON output. Read `AlertName`, not `alertName`.
```
rohitjain-uipath: `Name` is the emitted key, not `AlertName` — `projectAlertDefinition` maps `dto.alertName` to `name`, so `Data.AlertName` reads `undefined`.

- the envelope example on line 45 already shows `"Name": "Queue backlog"`
- Safe Output on line 18 also uses `Name`

Suggest `DeliveryId` / `deliveryId` as the casing example instead.

rohitjain-uipath (approved):
Checked the guide's commands, flags, `Key Data fields`, error table, and exit codes against `commands/alerts.ts` + `utils/alerts.ts` on cli main — all accurate, including the `check-entitlement` exit-0 assumption behind the `run_command` gate. One field-name fix inline before this lands.

skills/uipath-insights/references/alerts-reads-guide.md
celestinryf: Fixed in [`31e874b5`](https://github.com/UiPath/skills/commit/31e874b506917be84936ec4f72f462bc5a8d33b9). The example now reads `DeliveryId`, not `deliveryId`. Checked against `projectAlertDefinition`: `alertName` maps to `name` (utils/alerts.ts:363) and `deliveryId` is emitted at line 372, so `DeliveryId` is a real output key.

celestinryf:
On the lint's passing-run finding: the omission was deliberate and the body carried the same do-not-land line. It is now partly overtaken: `@uipath/cli@1.202.0-dev.8372` and the matching `@uipath/insights-tool` build published on 2026-08-24 with the alerts commands in the tool bundle, so a real run is possible. The Limitations section is updated to say so, and the passing-run line gets added once a real run passes.

The `AlertName` field-name finding matched the human review and is fixed in [`31e874b5`](https://github.com/UiPath/skills/commit/31e874b506917be84936ec4f72f462bc5a8d33b9).

skills/uipath-insights/references/alerts-reads-guide.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+**Key Data fields:** `Entitled`
+
+**Use when:** Rule 3 says exactly when. A `false` result has several possible backend causes.
+
+## Rules
+
+1. **Only active definitions are returned.** Every definition read filters on active state, and deletion is a soft delete, so `IsActive` is true on every row and a deactivated or deleted alert is invisible to all three routes. Report "which alerts are inactive, disabled, or turned off" as a question this surface cannot answer, never as "none". "Snoozed", "paused", and "muted" are different: `SnoozedUntil` and `AutoSnoozeSeconds` are returned and can be echoed, subject to Rule 5.
+2. **Page deliberately.** `--limit` accepts 1 to 10000 and defaults to 50, so a 50-row result is a full page rather than a complete list. Prefer one high-limit call over repeated `--offset` calls; stop after ten pages and report how many rows you retrieved. A yes/no question is answered by the first page. Only completeness questions need every page. `alerts get` and `alerts check-entitlement` do not page.
```
rliuup: suggestion (non-blocking): Worth saying here that `--limit`/`--offset` slice the CLI's own copy — the backend returns the full unpaginated array on every call (`utils/list-executor.ts` does `rows.slice(offset, offset + limit)`; `utils/list-options.ts` says so in a comment). That is what makes each extra `--offset` page cost another full fetch *and* another Not-Free preflight, and it is the one fact `filter-discovery-guide.md` Rule 3 states that this rule does not.

skills/uipath-insights/references/alerts-reads-guide.md
```diff
@@ -0,0 +1,130 @@
...
+3. **Run `check-entitlement` when the result is unresolved, not as decoration.** Run it when `alerts list` came back empty, when a non-empty result's `Instructions` say entitlement is unconfirmed, when `alerts get` returned a 404 (Rule 6), or when the user asks about entitlement directly. Skip it when none of those apply. While entitlement is false, `alerts list` and `alerts get` return only alerts tied to a process key; the agentic route is unaffected. Report a `false` as one of several possible backend causes, not as proof that alerting is off.
```
rliuup: suggestion (non-blocking): The first trigger reads as covering the agentic route too, which Rule 8 then says is never an entitlement artifact. The Not-Free Reads note already spells it "`alerts list` (without `--agentic`)" — the same qualifier here would remove the ambiguity.

skills/uipath-insights/references/jobs-commands-guide.md
```diff
@@ -80,7 +80,7 @@ Null, empty, or zero across every field on a `Success` response means the query
 
 All endpoints return the same shape. Which fields are populated depends on the endpoint.
 
-**Keys inside `Data` are PascalCase on the wire.** The CLI PascalCases every `Data` key before printing, so read `JobsCount`, not `jobsCount`. The type below is the SDK's `JobsResponse` in its camelCase source form. Every field is optional, so a field the endpoint does not populate may be absent or null.
+**Keys inside `Data` are PascalCase in the CLI's JSON output.** The CLI PascalCases every `Data` key before printing, so read `JobsCount`, not `jobsCount`. The type below is the SDK's `JobsResponse` in its camelCase source form. Every field is optional, so a field the endpoint does not populate may be absent or null.
```
rliuup, line 83: suggestion (non-blocking): This leaves `filter-discovery-guide.md:7` as the last "PascalCase on the wire" in the skill, and it is inaccurate the same way — `commands/filter-folders.ts` builds `{ folderName, folderKey }` and only `OutputFormatter` PascalCases on output. Folding that line in would put all four guides on the same wording.

tests/tasks/uipath-insights/alerts/smoke.yaml
```diff
@@ -0,0 +1,133 @@
...
+    pass_threshold: 1.0
+
+  # SKILL.md Critical Rule 2 forbids substitution and shell variables in a uip
+  # invocation. Flag values and the by-ID `get` argument are the spots most
+  # likely to tempt it. Anchored on a flag value or a `get` argument so an
+  # unrelated `$VAR` in a batched neighbour command cannot false-fail this.
+  - type: command_not_executed
+    description: "Agent did not use command substitution or a shell variable in a uip insights command"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+insights(?:(?!uip\s)[^&;|])*--[a-z][a-z-]*[\s=]+"?(\$\(|`|\$[A-Za-z_])|uip\s+insights\s+\S+\s+get\s+"?(\$\(|`|\$[A-Za-z_])'
```
rliuup: issue (non-blocking): The comment says a batched neighbour's `$VAR` cannot false-fail this, but the `(?!uip\s)` barrier stops at the next `uip `, not at the next command. A newline-batched `uip insights alerts list --output json` followed by `jq --arg k $KEY out.json` matches under `re.DOTALL`. Narrowing the first branch to the flags these commands actually accept (`--(limit|offset|process-key|output)`) closes it.

rliuup (commented):
Checked every command, flag, `Key Data fields` entry, error-table row, and exit code in the new guide against `commands/alerts.ts` and `utils/alerts.ts` at cli main `8a9466ac3` — all accurate, including the three `Code` values, the `--agentic`/`--process-key` pairing, and the exit-3 rejection of time flags. Four non-blocking notes inline.

Holding at comment rather than approve only because the passing smoke run this PR's own Limitations section calls for does not exist yet. Happy to approve once it does.

rliuup:
Correction to my review above: I wrote that the passing smoke run your Limitations section calls for does not exist. It does — run [32783473599](https://github.com/UiPath/skills/actions/runs/32783473599) on this head (`c8ab5ac`) scored `skill-insights-smoke-alerts-reads` 9/9, weighted 1.000, and the `run_command` gate really executed `uip insights alerts check-entitlement --output json` to exit 0 with `"Code": "InsightsAlertEntitlement"`, so the CLI in the smoke image does ship the alerts group. The body was accurate when you wrote it and went stale about four minutes later. Approving separately; worth updating Limitations so the next reader does not re-derive this.

Two other body lines have drifted the same way: #2739 bases on `main` rather than being stacked on this branch, and it already carries the `filter-discovery-guide.md:7` wording fix I asked about inline — say so on that thread rather than doing it twice.

rliuup (approved):
LGTM with nits — the four inline notes are all non-blocking.

skills/uipath-insights/references/alerts-reads-guide.md
celestinryf: Fixed in [`67c3bb5a`](https://github.com/UiPath/skills/commit/67c3bb5a353098018e156d4f4d7b9f1bb553acb3). Rule 2 now says both flags slice the CLI's own copy, the backend returns the full unpaginated array each call, and each extra `--offset` page costs another full fetch plus another Not-Free preflight. Verified at `list-executor.ts:132` and the `list-options.ts:5` comment.

skills/uipath-insights/references/alerts-reads-guide.md
celestinryf: Fixed in [`67c3bb5a`](https://github.com/UiPath/skills/commit/67c3bb5a353098018e156d4f4d7b9f1bb553acb3). Rule 3's first trigger now reads `alerts list` (without `--agentic`), the same qualifier the Not-Free note uses.

skills/uipath-insights/references/jobs-commands-guide.md
celestinryf: Not fixing it here: #2739 already carries it in [`b2c2a168`](https://github.com/UiPath/skills/commit/b2c2a168700bbf23cb69cbf61b35298a2c164adb), which rewrites `filter-discovery-guide.md:7` to name the output layer. Folding it into this PR too would land the same hunk twice.

tests/tasks/uipath-insights/alerts/smoke.yaml
celestinryf: Fixed in [`56766e51`](https://github.com/UiPath/skills/commit/56766e5132de9276288aeb787911e0aa94d12ef6). The flag branch now names only `limit|offset|process-key|output`, so a neighbour's `jq --arg=$KEY` or `curl --data $BODY` no longer matches while the backslash-continuation catch stays. Your exact `--arg k $KEY` example already missed (the variable has to follow the flag directly), but the mechanism was real; re-ran the hand-written string tests, all pass.

celestinryf:
Limitations now names run [32783473599](https://github.com/UiPath/skills/actions/runs/32783473599) and the 9/9 score, and both stacked-on lines are corrected: #2739 bases on `main` and stays draft until cli#3611 ships. The filter-discovery thread has the pointer to the #2739 commit instead of a second copy of the hunk.

rliuup (approved):

rockymadden (approved):
LGTM

PR #2610, opened 2026-08-13
fix(maestro-case,planner): string-encode variable defaults across the whole chain

roberts-cliff:
## The bug

A `jsonSchema` variable default was emitted as a JSON object. The caseplan → BPMN converter keeps only primitive attributes, so the value is **deleted without comment**:

```ts
// bpmn-moddle.ts — onlyPrimitiveVariableFields
if (value === null || typeof value !== "object") { result[key] = value; }
```

The emitted BPMN then carries no `default` for that variable, while every string variable beside it keeps `default=""`. The variable is null at runtime and the first task bound to it fails with `AGENT_STARTUP.INPUT_VALIDATION_ERROR / <input> Field required`.

Observed on a shipped build: **13 non-string defaults, all `jsonSchema`, all dicts.** Twelve were empty `{}`; one carried the payload an agent task required.

---

# Evidence: placement is measured, not guessed

The rule that fixes this bug already existed — in one line of one file, one section away from where the value is written. So the question was whether restating it closer actually changes anything, or just feels tidier.

**terra, `linear_three_stages`, N=10 per arm.** The constraint is byte-identical across arms; only its **position** varies. Compliance graded at artifact level by the checker below, not by task score:

| Arm | Constraint in `impl-json.md` | Compliance | |
|---|---|---|---|
| **D** | absent | **1/10** | 10% |
| **A** | one section away | **6/10** | 60% |
| **C** | both places | **9/10** | 90% |
| **B** | write-site block only | **10/10** | **100%** |

Fisher exact, one-tailed:

| Contrast | p | Reading |
|---|---|---|
| **D vs A** | **0.029** | a distant rule *does* work — 10% → 60% |
| **A vs B** | **0.043** | moving it into the block is worth another ~40 points |
| A vs C | 0.15 | duplication alone does not clear the bar at this N |
| **B vs C** | **1.00** | **the second copy adds nothing** once a near copy exists |

Four things follow, and each one changed the code:

1. **Placement is worth ~40 points of compliance** on a fast model. Read count was unchanged at **~3 calls** between arms A and B — the lift is not the model reading more, it is the constraint being where the reading already goes. That is why `§ default encoding` now sits *inside* the pattern-shape block rather than in its own section.
2. **`B vs C = 1.00` is why the final commit drops the duplicate.** It was surface area, not defence. The rule now appears **once**: body byte-identical, demoted to `###` so the Pattern-shapes hierarchy holds, anchor unchanged so both inbound links still resolve.
3. **Arm D is not zero.** Terra string-encodes unprompted ~10% of the time, so any effect measured against an assumed zero floor would have been overstated by that much.
4. **Adjacency is a gradient, not a cliff.** A distant rule is *partially* effective. 60% is still not shippable for a contract that silently deletes data when violated — which is what Check 14 is for.

> **Correction to an earlier revision of this description.** It cited an N=3 baseline of **0/3** and **`p=0.0035`**. The true rate at that distance is 60%; P(0/3 | p=0.6) = 6.4% — a 1-in-16 draw. Claims made from it ("the far copy is dead weight", "the walk never reaches it") were false. The honest figure is `p=0.043`, and the direction survives; the magnitude does not.

### How the second defect surfaced

The first terra suite failed 3/3 — not on `jsonSchema`, which was clean, but on a **boolean** pure-state Variable emitting `default: false` instead of `"false"`. Sonnet emitted `"false"` 3/3 from the same docs. Identical inputs, opposite behaviour, deterministic per model. That split is what prompted the placement experiment above.

### Model coverage for the original defect

`linear_three_stages`, N=3 per model, artifact-level (three `jsonSchema` variables present in every run, so the criterion had real input and did not pass vacuously):

| Model | Runs | jsonSchema vars | object defaults | verdict |
|---|---|---|---|---|
| `gpt-5.6-terra` (codex) | 3 | 3 | **0** | 3/3 SUCCESS 1.0 |
| `claude-sonnet-5` | 3 | 3 | **0** | 3/3 SUCCESS 1.0 |

**The ticket's defect is closed on both models** — zero object-valued defaults in 6/6.

---

## Why nothing caught it

| Layer | Says | Enforced against agent JSON? |
|---|---|---|
| `VariableTypes.ts` | `default?: string` | no — compile-time only, never sees the file |
| `case-mgmt-zod` | `z.any()` | runs, accepts everything |
| `bpmn-moddle.ts` | drops non-primitives | runs, **silently** |
| `uip maestro case validate` | `Valid` | no signal |

**All four rows verified first-hand against `PO.Frontend` @ `c1e854b3e` (branch `develop`, release 0.1016.0)**, not from the linked line numbers alone — `onlyPrimitiveVariableFields` is present and current, `VariableTypes.ts` declares `default?: string`, and `UiPathVariablesJsonSchema.ts:12` types the field `z.any()`. That answers the review's *"reviewers with frontend access should confirm the quoted behaviour is current"* — it is, as of that commit. Note the checkout is a release commit on `develop`, so it may be ahead of what is deployed.

I also tested the obvious shortcut — run the caseplan through the frontend's own Zod schema as a gate. It parses the failing artifact clean at **both V27 and V30**, because of that `z.any()`. So the FE parser cannot serve here; the serializer's behaviour is the effective contract, and it is the one layer that fails without reporting.

## Why it survived three hands

The rule existed in exactly one line of one file (`global-vars/planning.md`), and every step behaved defensibly given what it was told:

| Step | What happened | Was it wrong? |
|---|---|---|
| `sdd.md` | Default cell held a bare object | template said only `<default or blank>`, never mentioned encoding |
| `tasks.md` | carried it verbatim | Rule 6 (lossless, preserve verbatim) arguably **requires** this |
| `caseplan.json` | written from `impl-json.md`, whose jsonSchema recipe had **no `default` line at all** | nothing at write time contradicted the object form |

## The fix — all three hands

- **`planner` / `case-sdd-template.md`** — `Default` is a string for every `Type`, with the deletion consequence and worked encodings. Stops the object entering the chain.
- **`global-vars/planning.md`** — string-encoding **overrides losslessness** here; the single Rule 6 exception in the plugin, with a conversion table.
- **`global-vars/impl-json.md`** — `§ default encoding` covering every type, quoting the serializer source, with a `default` line in the jsonSchema recipe (`"{}"`, never `{}`), placed inside the pattern-shape block per the measurement above.

Adds **Check 14** to the Step 12 validator with no best-effort escape — a deleted default is not a partial result. **Check 13 was already defined but missing from the invocation list**; it is now listed.

## Test

`_shared/check_variable_defaults.py` + **16 pytest cases** on synthetic stubs, wired into `linear_three_stages`.

**Verified against the real failing artifact** — exits 1, flags all 13, prints the exact corrected encoding per variable. One test asserts the suggestion **round-trips** (`json.loads(json.loads(x)) == original`), so the fix it recommends is provably correct rather than merely well-formed; another pins boolean encoding to lowercase JSON `"true"`, not Python `"True"`.

**Mutation-tested:** grader-never-flags → 10 fail · only-`inputOutputs`-scanned → 2 fail · Python-`True` encoding → 1 fail. `pytest tests/tasks/uipath-maestro-case/` → **96 passed**.

## Notes for reviewers

**On the strict check.** A boolean default *survives* serialization, so it is not fatal, and I nearly relaxed the criterion for that reason. The two-model comparison is the only thing that distinguished "the check is too strict" from "one model is not following the rule" — the failing artifacts are identical from the terra side alone.

**Incidental, pre-existing.** One arm-B replicate crashed `check_linear_three_stages.py` with `AttributeError: 'str' object has no attribute 'get'`. Unrelated to this change, fires roughly 1 in 10, invisible at N=3. Worth its own fix.

**Not depending on the upstream fix.** One line on the frontend would turn a silent strip into a parse error — `z.any()` → the string type `VariableTypes.ts` already declares two files away. Worth filing, but this fix does not depend on it: `z.any()` in a shipped schema is either an untyped shortcut nobody will prioritise, or something already relies on the permissiveness, making it a breaking change with unknown blast radius.

**Check number collision.** Numbered **Check 14** against `main`, which tops out at 13. A separate in-flight change also introduces a Check 14/15 — if both land, one set needs renumbering. No functional overlap.

## Checks

`skills:validate` (26 skills, default + studioweb) · `check-skill-status` · `check-cli-verbs` · `pytest` 96 passed. `audit_sdd.py` on the edited template reports the same 3 findings before and after (pre-existing placeholder-text findings, not introduced here).


charlesliu9 (approved):

roberts-cliff:
## @RaduAna-Maria @abhiram-vad — one-file planner review needed to unblock this

This PR is approved by @charlesliu9 and green (21/21 checks), but `reviewDecision` is still
`REVIEW_REQUIRED` because it touches a single file under your ownership:

```
CODEOWNERS:77   /skills/uipath-planner/   @RaduAna-Maria @abhiram-vad
```

**That file is the entire planner surface of this PR** — one table cell plus a blockquote in
`assets/templates/case-sdd-template.md`. No logic, no workflow, no planner behavior change.
Reproduced in full so you don't need to open the diff:

```diff
-| ... | <default or blank> | <what this variable represents> |
+| ... | <default or blank — see below> | <what this variable represents> |
+
+> **`Default` is always a string, for every Type.** The downstream `caseplan.json` field is
+> string-typed, and a non-primitive value there is **silently deleted** when the case is serialized to
+> BPMN — the variable is then null at runtime and the first task that reads it fails. Write a
+> `jsonSchema` default as string-encoded JSON in the cell:
+> `{"employeeName":"Test Employee","amount":125.5}` becomes
+> `"{\"employeeName\":\"Test Employee\",\"amount\":125.5}"`. Same for numbers and booleans — `"5"`,
+> `"true"`, not `5` or `true`. Leave the cell blank for no default; `file` defaults must be blank.
```

### Why the planner template is in scope at all

The bug is a chain, and the SDD template is where it starts. A `jsonSchema` default written as a bare
JSON object survives every layer that could have caught it and is then **silently dropped** by the
caseplan → BPMN converter (`onlyPrimitiveVariableFields` keeps primitives only). The variable is null
at runtime and the first bound task fails with `AGENT_STARTUP.INPUT_VALIDATION_ERROR`.

Observed on a shipped build: 13 non-string defaults, all `jsonSchema`, all objects — one carrying the
payload an agent task actually required.

The template previously said only `<default or blank>`, which never mentioned encoding, so an author
filling it in correctly-by-the-template produces the broken form. Fixing the two maestro-case files
without this one leaves the entry point of the chain still inviting the defect.

If you'd rather keep planner and maestro-case changes in separate PRs, say so and I'll split this file
into its own PR — but then this fix ships with its first hand unpatched, so I'd rather not by default.

abhiram-vad (approved):

PR #2614, opened 2026-08-14
fix(case): case completion allows multiple rules, not just required-stages-completed

abhiram-vad:
## TL;DR

Both case skills asserted that a case-completion rule (`Marks Case Complete: Yes`) may only use `required-stages-completed` or `wait-for-connector`, and `uipath-planner` **blocked Approve** on `Yes` + `selected-stage-*` as a "schema-pairing error".

**That constraint does not exist in the product.** The Case Designer's *"Complete case when"* picker offers five rules:

image

```
Required stages completed
Selected stages exited
Selected stages completed
SLA status change occurs
Connector event occurs
```

### `uipath-maestro-case`
- **`SKILL.md`** — closure shape is a choice, not a default. Lists all five rule types; states why `required-stages-completed` is brittle.
- **`case-exit-conditions/planning.md`** — replaces the `marksCaseComplete`-gated catalog with the real one. Adds deterministic-vs-contingent classification, a *Choosing the closure shape* section, and a reachability requirement. Documents that **`marksCaseComplete` is designer-only and never serialized** — the converter emits every `caseExitRules[]` entry into `case.completionConditions` regardless of its value.
- **`case-exit-conditions/impl-json.md`** — corrects the rule-type matrix; adds recipes for `selected-stage-*` at `Yes`, for `sla-status-change`, and for multi-terminal closure.
- **`stage-exit-conditions/impl-json.md`** — new *Exiting is not completing* section, placed at the point the exit is written (this file is loaded alone, just-in-time, per `SKILL.md:104`).
- **`case-schema.md`** — `sla-status-change` also has case-exit scope.

### `uipath-planner`
- **`case-authoring-rules-guide.md` §1.4** — allows all five WHENs; adds closure-shape guidance and a blocking reachability check. **Keeps the stage-level pairing rule intact** (`Marks Stage Complete: Yes` + `selected-tasks-completed` genuinely is illegal) and stops generalising it to the case level.
- **XOR terminal Pattern X1 rewritten** to key case exit on terminal stages. The prior version marked mutually-exclusive terminals `Required for case completion: Yes` and relied on the runtime auto-completing the untaken terminal as `status = "Skipped"`, citing *"documented FE behavior"*. **No such status exists** — `ElementStatusValues` has no `Skipped` member and there is no skip-on-false-`IF` behaviour in PO.Frontend.
- **`case-sdd-template.md`** — corrects the pairing comment above the case-exit table.

## New `tasks.md` field

`selected-stages` (list) on a case-exit T-task, lowering to one OR'd rule-set per stage inside a **single** condition:

```json
"rules": [
  [ { "rule": "selected-stage-completed", "selectedStageId": "Stage_ExeWrp" } ],
  [ { "rule": "selected-stage-completed", "selectedStageId": "Stage_RejWrp" } ],
  [ { "rule": "selected-stage-completed", "selectedStageId": "Stage_WdrWrp" } ]
]
```

OR across rule-sets **within one condition** is verified in the compiler. OR across *separate* conditions is **not** — that fold happens inside `UiPath.Scheduler` (external package), so alternatives deliberately stay in one condition.

## Verification

Read from fetched origin refs, not stale local checkouts:
- PO.Frontend `origin/develop` @ `e38e3d692` (2026-08-14) — local tree was 520 commits behind
- PO.BpmnEngine `origin/main` @ `dc6e24ff1` (2026-08-13) — local tree was 2520 commits behind

Schema shape left on **v27** (`selectedStageId` singular) to match this skill's declared target. The v29 rename to `selectedStageIds[]` is a separate migration.

## Known limits / follow-ups

Not in this PR, filed separately:
1. **OR-vs-AND across multiple `completionConditions`** is unverified — the fold lives in `UiPath.Scheduler` 1.20.0, in neither repo. Guidance here avoids depending on it.
2. **`caseResolution.type: "exited"` does not close the case.** The schema says *"'completed' or 'exited' ends the case"*, but the generated gateway reaches `SendCaseEndMessageEvent` only on `== "completed"`. Schema doc and emitted graph contradict.
3. **`exit-stage` as an SLA response has no emit shape** — `sla-status-change` is not a legal stage-exit rule type in any catalog, and `sla-response-shapes.md`'s CLI-verified table probed only entry-scope shapes.
4. **A Step 12 closure-reachability check (Check 13)** would make this enforceable rather than advisory. Sidecar over 204 runs of this skill shows prose rules in reference files land at low adherence — the SLA-response rule (`#21`, anchored on `sla-response-shapes.md`) sits at **2%**, and 0% on every Claude model — so a script at an existing gate is the durable fix.


roberts-cliff (approved):

charlesliu9:
"Marks Stage Complete: Yes + selected-tasks-completed" seems legal now

PR #1370, opened 2026-06-10
feat(codex): add Codex plugin metadata

e4c6:
## Summary

Adds Codex plugin metadata and a repo-scoped Codex marketplace entry for the UiPath skills repository.

This PR adds:

- `.codex-plugin/plugin.json` exposing the existing `skills/` directory as a Codex plugin
- `.agents/plugins/marketplace.json` so Codex can discover the plugin from this repository as a marketplace source
- `plugins/uipath -> ..` as a symlink wrapper so the marketplace entry points at a plugin folder without duplicating the skills tree
- an explicit Codex hook reference to the existing shared `hooks/hooks.json`
- the official UiPath icon asset for Codex plugin display
- README commands for installing with Codex CLI

## Install after merge

```bash
codex plugin marketplace add UiPath/skills --ref main
codex plugin add uipath@uipath-marketplace
```

## Screenshots

UiPath plugin details in Codex

UiPath plugin marketplace install flow

## Validation

- `npm run version:check`
- `jq empty .agents/plugins/marketplace.json .codex-plugin/plugin.json hooks/hooks.json`
- `git diff --check upstream/main...HEAD`
- Isolated local Codex marketplace install smoke: plugin installs as `uipath@uipath-marketplace` at version `0.0.36`
- Isolated local Codex MCP smoke: no MCP servers are configured by this plugin
- Hook compatibility smoke: `suggest-permissions.sh` stays silent under Codex plugin env and still emits the Claude permission nudge under Claude-only plugin env

PR #2620, opened 2026-08-14
fix(codex): use repository root plugin source

e4c6:
## Summary

- point the Codex marketplace entry directly at the repository root
- remove the `plugins/uipath` symlink workaround
- document Codex CLI 0.142.0+ as the minimum version for this layout

## Context

`plugins/uipath -> ..` was introduced in #1370 because Codex could not use the marketplace repository root as a local plugin source.

openai/codex#17066 was resolved by openai/codex#28771, with support for `.` and `./` released in Codex CLI 0.142.0. The indirection is therefore no longer required.

No Claude Code regression was observed. Claude already resolves the repository root independently through `.claude-plugin/marketplace.json`, and an isolated Claude marketplace/install smoke passed on Claude Code 2.1.206.

## Validation

- JSON manifest parsing
- `npm run version:check`
- `npm run skills:validate`
- `git diff --check`
- isolated local Codex marketplace add/install/list on Codex CLI 0.147.0
- isolated Git-backed Codex install from the pushed branch
- confirmed version 1.201.0 enabled with all 26 skill directories and hooks
- isolated Claude marketplace add/install/list/details smoke on Claude Code 2.1.206

## Test baseline

`npm run skills:test` reports 46 passes and two npm publish-dry-run failures. The same two failures reproduce unchanged on pristine upstream `main`; they are caused by existing npm registry/tag fixture drift and are unrelated to this three-file change.


uipreliga (approved):
Verified upstream: openai/codex#28771 (merged 2026-06-17) landed in rust-v0.142.0 — the fix commit is absent from 0.141.0 and present in 0.142.0, so the stated minimum is exact. No repo reference to plugins/uipath outside the two files edited here, and plugins/ is not in package.json files[]. Claude Code already resolves the root via .claude-plugin/marketplace.json "source": "./", so this makes Codex consistent rather than changing behavior. Removing a recursive '..' symlink is also a Windows-checkout win.

PR #2621, opened 2026-08-14
feat(preview): add Maestro SDK authoring skills

tmatup:
## What changed

Copied the Flow, Case, and BPMN TypeScript SDK authoring guides from `UiPath/flow-builder-sdk` into three self-contained preview skills:

- `preview/uipath-maestro-flow/`: `SKILL.md`, 32 references, and 25 examples
- `preview/uipath-maestro-case/`: `SKILL.md`, `references/api.md`, and 5 examples
- `preview/uipath-maestro-bpmn/`: `SKILL.md`, `references/api.md`, and `NotifyChannel.bpmn.ts`

Source snapshot: [`f4973f6`](https://github.com/UiPath/flow-builder-sdk/commit/f4973f6)

These copies are deliberately staged under `preview/`. That path is outside the current plugin loader, package surface, path-triggered skill validators, status manifests, and promotion task suites.

## Decisions preserved

- **D1 — copy with provenance:** `flow-builder-sdk` remains canonical and untouched; each snapshot cites the source path, commit, and cutover issue.
- **D2 — Flow owns Flow references only:** the Flow copy excludes `case-api.md` and `bpmn-api.md`; its sibling paragraph routes by skill name.
- **D3 — BPMN example included:** `examples/NotifyChannel.bpmn.ts` ships with the BPMN snapshot.
- **D4 — preview frontmatter:** each `SKILL.md` has first-line YAML, `name` equal to its folder, and the established Maestro `allowed-tools` list.
- **D5 — local reference names:** Case and BPMN each expose their generated reference snapshot as `references/api.md`; Case exposes `case-bindings.json` as `examples/bindings.json`.
- **D6 — CLI filename contract retained:** examples keep `.flow.ts`, `.case.ts`, and `.bpmn.ts`.

Cutover tracking: [UiPath/flow-builder-sdk#405](https://github.com/UiPath/flow-builder-sdk/issues/405)

## Verification

1. **Byte-diff sweep:** exact inventories passed. All copied references and unchanged examples are byte-identical to `f4973f6`; the three pure renames are byte-identical; reversible normalization proves the only content differences are the three approved `SKILL.md` edit sets and the one `NotifyOnApproval.case.ts` comment.
2. **Dead links:** 37 Markdown documents scanned; all 119 relative links resolve inside their owning skill.
[...]

PR #411, opened 2026-04-27
fix(hooks): raise SKILL.md description limit from 250 to 1024 chars

AlvinStanescu:
## Summary
- Raises the `hooks/validate-skill-descriptions.sh` `LIMIT` from **250 → 1024** chars.
- Corrects the misleading framing in repo docs that called 250 chars Claude Code's truncation point — the actual cap is **1,536 chars** for combined `description` + `when_to_use` ([source](https://code.claude.com/docs/en/skills.md)). 1024 leaves headroom under that cap and is a repo style choice, not a platform limit.
- Updates `.claude/rules/`, `.cursor/rules/`, and `CONTRIBUTING.md` so the wording matches.

## Why
250 was too narrow — it was forcing descriptions to drop useful disambiguation signals (`→` redirects, file/domain hints) that the matcher relies on. All 17 existing skills are well under 250 today (max: 247), so this change adds room without breaking anything.

## Test plan
- [x] `bash hooks/validate-skill-descriptions.sh skills/*/SKILL.md` passes for all 17 skills
- [x] `grep -rn "250" hooks/ .claude/rules/ .cursor/rules/ CONTRIBUTING.md` returns nothing
- [ ] CI hook check passes on this PR

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #409, opened 2026-04-27
docs(uipath-maestro-flow): add fast-path flow example

kab0rn:
## Summary
- Add a validator-clean `decision-branch.flow` example for the common manual trigger -> script -> decision -> branch scripts -> End-node pattern.
- Document the example as the fast path for common branching flows and clarify when agents should avoid inline `node add --input` shell-escaping loops.
- Add helper pytest coverage for the example topology, outputs, definitions, and horizontal layout.

## Why
Agents building Flow projects can lose time fighting inline JSON quoting for `$vars`, template literals, and multiline scripts. A complete `.flow` baseline gives them a copyable, validated starting point for a common branching pattern while preserving the Direct JSON default.

Refs #114

## Validation
- `git diff --check`
- `bash hooks/validate-skill-descriptions.sh skills/uipath-maestro-flow/SKILL.md`
- `python3 -m pytest tests/tasks/uipath-maestro-flow/_shared -q`
- `npx -y @uipath/cli@latest maestro flow validate skills/uipath-maestro-flow/references/examples/decision-branch.flow --output json`

PR #2625, opened 2026-08-16
chore(preview): re-sync Maestro SDK skills to flow-builder-sdk@406162b

dmetzgar:
**Stacked on #2621** — targets `feat/add-maestro-sdk-preview-skills`, not `main`.

The snapshot in #2621 is pinned at `flow-builder-sdk@f4973f6`. Upstream has moved five commits since (#406, #409, #411, #412, #413). This brings the vendored copy forward and re-pins the provenance headers to `406162b`.

## What arrived

| | |
| --- | --- |
| **`stepToRef` / `stepToList`** | Edges to arbitrary steps. The builder gained `.stepToRef(target)`, `.stepToRef(port, target)` and `.stepToList(port, fn)`; `rejoin` is gone. This is why `SKILL.md`, `error-handling.md` and all three `api.md` move. |
| **TSDoc backfill** | Every public parameter and return is now documented (100% of both). Most of the `api.md` and `SKILL-case.md` diff. |
| **`CLI-LOOP` / `connector-params` / `inline-agent` / `brownfield`** | Upstream edits from #406 that predate the snapshot's own drift. |

## Applied as a merge, not a copy

A straight copy would have destroyed the adaptations this repo added. I tried it first and the diff showed exactly that:

- the **YAML frontmatter** (`name`, `description`, `allowed-tools`) that makes each file a skill;
- the **provenance comments**;
- the **link rewrites** — upstream's `references/bpmn-api.md` is this repo's `references/api.md`.

So this is a three-way merge against the pin. All eleven files merged clean. The adaptations are intact — `flow/SKILL.md` still names the sibling *skills* rather than linking across to their files (what self-containment requires) while taking upstream's new sentence about the builders.

Verified after merging: frontmatter and provenance present on all three SKILL.md, `stepToRef` present in the files that should have it, zero conflict markers.

## Two things deliberately NOT fixed here

Both need upstream changes rather than snapshot-only corrections, per the triage on #2621:

1. **`flow/references/api.md`** still says ``Worked example: `example/DraftReview.flow.ts` ``. That file is real upstream but is **not in the vendored set** (the vendored examples come from `example-eval/`), and this repo's directory is `examples/`. This is the review's "wrong directory *and* no such file" item — the fix is upstream: point the `@example` tag at a vendored example.
2. **The three `api.md` banners** still instruct `npm run generate:api` and claim `npm run build` catches staleness — true upstream, false here. The fix is to make the banner name its origin repo so it reads correctly in both.

I have upstream fixes ready for both; say the word and they go into `flow-builder-sdk` first, then a follow-up sync here.

## Note on the review's other findings

Most of the Critical/High items are **upstream content** (`.var()` contradiction, `solution project add`, `$FLOW_SDK_LIBRARY_*`, the `$vars` spellings, the Flow-only half of `case/references/api.md`) and want fixing in `flow-builder-sdk` so they do not reappear at the next sync. The structural ones (`preview/` vs `assets/skill-status.json`, duplicate `name:` values, missing Critical Rules sections) are this repo's and independent of the sync.


dmetzgar:
Re-synced to [`flow-builder-sdk@a065b21`](https://github.com/UiPath/flow-builder-sdk/commit/a065b21) now that [flow-builder-sdk#414](https://github.com/UiPath/flow-builder-sdk/pull/414) has merged. That PR fixed the review items raised on #2621 **at the source repo**, so the fixes survive future syncs rather than being re-applied by hand each time.

Applied as a three-way merge against the previous pin — 15 files, all clean, no conflicts. Frontmatter, provenance headers, and this repo's link rewrites are intact (verified: no `bpmn-api.md`/`case-api.md` leakage, flow SKILL.md still *names* its sibling skills rather than linking across).

**What arrived**

| | |
| --- | --- |
| retired names | `uip solution project add` → `projects add`; Slack `send-message-to-a-channel` → `send-message-to-channel`; or-processes.md's "API Workflow" → `api-workflow.md` |
| subflow inputs | a child's inputs bind to `<callerStepId>Start`, not a bare `$vars.<name>` — both docs now use `input(...)` |
| case corrections | `number` added to the legal variable types; the stale "not referenceable from `=js:vars`" claim removed; CLIs named `case-check` / `case-compile` |
| environment | `$FLOW_SDK_LIBRARY_MD` / `$FLOW_SDK_LIBRARY_JSON` and `prepare-connector` documented as environment-supplied with no default |
| examples | the `out()`-of-the-failed-step handler bug, `NotifyChannel` posting to a user despite its name, the duplicate `name: "FolderKey"`, unused imports |
| TSDoc repair | 34 comments mangled by the #412 backfill, restored — most of the api.md movement |

**The Case reference loses ~1,000 lines.** `ConnectorSpecData = Extract<ActionSpec, { kind: 'connector' }>` was making the upstream reachability walk pull the entire Flow action union into `uipath-maestro-case/references/api.md` — 51 symbols of option shapes (`HttpInputs`, `HitlInputs`, `FormField`, the tool refs…) that no Case builder method accepts. They stay documented in the Flow reference; nothing was lost.

**Verb gate:** 0 blocking on all three skills. Flow's soft-stale count drops 9 → 6 — the three `solution project add` hits are fixed, and the remaining 6 are positional argument values (`compile Deployed`, `registry search agentic`, a GUID) that the literal-walker can't distinguish from sub-verbs.

**Two items deliberately still not fixed**, because they need upstream changes rather than snapshot-only edits — both already raised on #2621:
- `flow/references/api.md` still says "Worked example: `example/DraftReview.flow.ts`" — real upstream, outside the vendored set.
- The three api.md banners were reworded to name the origin repo, which helps a reader here, but the commands they give still describe the upstream build.

tmatup (approved):

tmatup:
Fresh-eyes closeout on the two “deliberately still not fixed” bullets at this PR's current tip, `53fd856f4`:

- **Worked example:** `preview/uipath-maestro-flow/references/api.md:134` is `Worked example: \`example-eval/GallerySubmission.flow.ts\`.` and `DraftReview` has zero matches. `preview/uipath-maestro-flow/examples/GallerySubmission.flow.ts` exists.
- **Generated banners:** all three `references/api.md` files begin `GENERATED from the TypeScript declarations in UiPath/flow-builder-sdk`; the next lines say “that repo regenerates it … and fails its build when it is stale.” The commands are therefore explicitly scoped to the upstream build and read correctly from this repo.
- **Deferred mislink:** `preview/uipath-maestro-flow/references/or-processes.md:6` links API Workflow to `api-workflow.md`, not `agent.md`.

Those fixes came from [flow-builder-sdk#414](https://github.com/UiPath/flow-builder-sdk/pull/414) and the `a065b21` re-sync already present here. No additional fix remains for those two stale bullets.

The real residual is narrower: `example-eval/` is a repository-tree prefix, while consumers see `example/` upstream and `examples/` in this snapshot. [flow-builder-sdk#420](https://github.com/UiPath/flow-builder-sdk/pull/420) fixes the generator upstream-first. The current follow-up sync, [skills#2679](https://github.com/UiPath/skills/pull/2679), pins `b3c4498` and intentionally does **not** claim that still-open fix; it will arrive in the next re-sync after #420 merges.

bai-uipath (approved):
**Fix if you agree, otherwise lgtm.** Three-way-merge-against-the-pin is right: a straight copy wipes the frontmatter, provenance, and path rewrites that make these files skills. Self-contained too, nothing outside `preview/` consumes them and the coder-eval pin is untouched.

All minor, none blocking:

- **Untested guard rails.** 38 hard-fail guards, zero negative test cases, so nothing proves the gates fire, and they're the safety argument for an unattended daily job that opens PRs. The happy-path fixture harness already supports adding throws-cases.
- **Case API reference lost option shapes the case skill still teaches.** It no longer documents the api-workflow or agent input shapes, but the case SKILL demos both task kinds and links only to its own reference. Belongs upstream, a local patch gets overwritten at the next sync.
- **Verb-gate glob.** The all-skills branch dropped the `|| true` the changed-skills branch kept, so an empty `preview/` fails that step under pipefail.

uipreliga:
## Code review

Multi-model review (Gemini 3 Pro, GPT-5.3-Codex, Opus) plus local reproductions. Focus: correctness, KISS, DRY, over-engineering.

**Scope.** The reviewable code here is 4 files, not 40: `scripts/sync-maestro-sdk-preview.mjs` (new, 661 lines), `.github/workflows/sync-maestro-sdk-preview.yml` (new, 181), `tests/scripts/sync-maestro-sdk-preview.test.mjs` (new, 318), `.github/workflows/verb-gate.yml` (modified), `package.json` (one line). The ~2,500 lines of `preview/**` markdown are vendored upstream content and were not reviewed as authored code.

Findings marked **W** cover the new sync workflow, which only one reviewer saw — treat those as single-reviewer confidence.

---

### High

**H1 — `sync-maestro-sdk-preview.yml:32-35`: the duplicate-PR guard is already broken in this repo.**
`gh pr list` defaults to `--limit 30`, and no `--search`/`--head` filter is applied. This repo currently has **178 open PRs**. Once a sync PR ages past position 30 in the list, the guard stops seeing it and the daily cron opens a fresh duplicate every night — defeating the one thing the guard exists for. Fix: `--search "head:auto/resync-maestro-sdk-preview-"`, or raise `--limit`.

**H2 — `sync-maestro-sdk-preview.mjs:612-619`: the provenance pin is rewritten *before* verification, and a failed run cannot be re-run.**
`applyAdaptations` (which rewrites all three pins) runs at `:612`; `verifySnapshot` at `:613`. Nothing is transactional. On any gate failure the tree is left half-synced **with the new pin already written** — so the next run reads `oldPin === newCommit`, every merge becomes a no-op, and the same failure recurs forever. Recovery requires `git checkout -- preview/`, which no error message mentions. CI survives (ephemeral checkout); a maintainer running it locally does not. CLAUDE.md demands transactional behaviour for the analogous pack flow.

**H3 — the three-way merge cannot succeed where a copy would, and adds five abort modes.**
Reproduced: snapshot-local drift in `preview/uipath-maestro-bpmn/references/api.md` merges cleanly, then dies at `verifyExactMappings` with `differs from …bpmn-api.md outside the approved adaptations`.

Because `verifyExactMappings` (`:404`) pins every non-SKILL target byte-for-byte to `transform(upstream@newCommit)` and `verifySkillBody` (`:393`) pins each SKILL.md body to `adapt(upstream body)`, the merge **cannot** produce an accepted output that copy-plus-adaptation wouldn't. Its only distinctive behaviour is turning mechanically-resolvable upstream churn into hard aborts. Two reproduced:

- an upstream tweak to `SKILL-case.md`'s own generated header comment (that file's provenance note says it is rendered by `gen-case-skill.mjs`, so its header *is* machine-generated churn);
- an upstream edit to the router table, which sits inside `git merge-file`'s 3-line context of the rewritten staging paragraph at `:56-60`.

Its one load-bearing job — preserving the three snapshot-local header blocks — is a slice at the first heading, which `bodyFromFirstHeading` already implements. The ~70 lines of merge machinery plus per-file `mkdtemp`/`rmSync` are a net reliability cost.

---

### Medium

**M1 — `sync-maestro-sdk-preview.mjs:251-259`: `git merge-file` exits with the *conflict count*, not 1.**
All three reviewers flagged it; reproduced end-to-end:

```
1 conflict  -> Three-way merge conflict in .../api.md (snapshot):   [conflict body shown]
2 conflicts -> git merge-file failed for .../api.md (snapshot): unknown error
```

The conflict text is sitting in `result.stdout` and is discarded. Genuine tool errors are 255 and do land correctly in the `:257` branch, so the fix is `status > 0 && status < 128`.

**M2 — 37 `fail()` call sites, 0 covered by a test.**
Two tests total: one E2E happy path and a 4-line unit test. No conflict, add/add, modify/delete, inventory-violation, dead-link, router-mismatch, or pin-disagreement case is exercised. The `git merge-file` machinery — the most failure-prone code in the file — has zero coverage, because no test ever creates local drift in a snapshot body.

**M3 — the E2E test is partly tautological.**
`createInitialSnapshot` (test `:183`, `:190`, `:225`) builds the fixture by calling the very `adapt*` functions under test, and `verifySkillBody` compares against `adapt(upstream)` — both sides move together. Three of five adaptations are rescued incidentally by `verifyStalePaths`, and `adaptFlowApi` by the literal at test `:272`. But **`adaptFlowSkill`'s staging-paragraph replacement and `adaptBpmnSkill`'s worked-example insertion are covered by nothing** — regress either and the suite stays green.

**M4 — `:456`: `decodeURIComponent` throws an unhandled `URIError`, bypassing `fail()`.**
An upstream link containing a bare `%` (e.g. `[speed](50%-faster.md)`) crashes the script with `URI malformed` — no file, no line, no link name.

**M5 — `:389-390`: `assertEqual` gives no location, no actual/expected, no diff** for comparisons of 700-line files. The same message is reused for the frontmatter `name` mismatch at `:432`, where "differs outside the approved snapshot adaptations" is actively misleading.

**M6 — `:465-496`: `verifyRouter` can pass vacuously.** If the table ever stops matching `/^\| .* \|$/` (HTML table, prose replacement, trailing whitespace), `rows` and `sourceRows` are both `[]`, the length check passes, the loop body never runs, and the gate reports success while checking nothing. It also false-reds if the first column header is renamed — `'| Node or surface'` is hardcoded at `:474`.

**M7 — `:431-435`: the description regex rejects valid YAML and then blames the length.**
`/^description:\s*"([\s\S]*?)"$/m` requires a double-quoted single-line scalar. Unquoted, single-quoted, and folded scalars are all valid YAML and all yield `undefined`, reported as `description must contain 1-1023 characters`. These files are generated upstream, so a quoting-style change there breaks the sync with a message pointing at the wrong problem.

Worth noting the check fills a real hole regardless: `hooks/validate-skill-descriptions.sh:20` greps `skills/.*/SKILL\.md$` and never matches `preview/`.

**M8 — `:277-284`: upstream *deleting* an adapted file aborts instead of deleting.**
`flow/references/api.md` and `case/examples/NotifyOnApproval.case.ts` carry adaptations, so `ours` never equals `base` → `Modify/delete conflict`. Another failure mode a copy-based sync does not have.

**W1 — `sync-maestro-sdk-preview.yml:75-95` duplicates the verb-gate classification logic.** The same four `jq` expressions and the same blocking/soft-stale taxonomy now live in two workflows that have to be kept in sync by hand.

**W2 — `sync-maestro-sdk-preview.yml:77` calls `python3` with no `setup-python` step.** Every other workflow in this repo (9 call sites across 7 files) pins 3.13 via `actions/setup-python`. The nightly job depends on ambient `python3` on the runner image.

---

### Low

- **`:87-98`** `gitObjectExists` swallows genuine git errors (corrupt object, partial clone) as "file absent", and the `fail()` at `:95` is unreachable in practice. Note `git cat-file -e <ref>:<missing-path>` exits **128**, not 1, so the `128 -> false` branch is required — the "wrongful delete" hazard is **refuted** (both refs are pre-validated at `:595`/`:597`, and every silent-null path is re-caught at `:413`/`:417`/`:286`). Separately `gitShow` (`:102`) has no try/catch, so a `git show` failure escapes as a raw stack trace instead of a `fail()`.
- **`:517-525`** false-reds when a body legitimately mentions `SKILL-case.md` more than once, with a message that names the wrong problem. The comparison logic itself is sound — no undefined-index crash.
- **`:176`** non-`.md` files under `skill/references/` are silently dropped. Upstream adding `references/diagram.svg` syncs green and simply never vendors it.
- **`mergeBuffers`** aborts the entire sync on binary input (`git merge-file` exits 255), which `verifyExactMappings` would have accepted as a byte copy. Reachable the first time upstream adds an image or compiled fixture under `example-eval/`.
- **`:578-583`** is dead weight — `assertManagedInventory` at `:546` already covers it, and the exported `verifySnapshot`'s `mappings` parameter exists only to feed this loop.
- **`:383`** `bodyFromFirstHeading` uses unanchored `indexOf('# ')`. Currently correct in all three files (verified: no `# ` occurs before the heading), but `/^# /m` removes the class entirely.
- **`scripts/check-skill-verbs.py:210`** `by_skill` hardcodes `parts[0] == "skills"`, so `preview/` findings vanish from the by-skill rollup in `--report` (used by `/audit-verbs`). All three reviewers flagged this; the gate uses `--json`, so CI is unaffected. Two sentences at `:246`/`:305` are now stale too. Looks like the companion change this PR skipped.
- **`verb-gate.yml:52`** `ls -d skills/*/ preview/*/ 2>/dev/null` under `set -euo pipefail` with no `|| true`. Reproduced: with `preview/` absent the step aborts with **no output at all**, stderr suppressed. Latent, not live.
- **`sync-maestro-sdk-preview.yml:132-137`** the pin sanity guard `case "$NEW_PIN" in *[!0-9a-f]*)` accepts the **empty string** — an empty pin yields branch `auto/resync-maestro-sdk-preview--<runid>-<attempt>` and a title ending `flow-builder-sdk@`.
- **`sync-maestro-sdk-preview.yml:73`** `git diff --check` makes the nightly job hostage to upstream's whitespace hygiene (markdown hard line breaks are two trailing spaces). Clean today — this PR's diff adds zero trailing-whitespace lines.
- **DRY nits.** The regex-escape expression is byte-identical at `:149` and `:349`. The provenance sentence is spelled three different ways (`:150`, `:350`, `:436`), so an upstream wording change breaks two of three with unrelated messages. `sourceRoot` is used by `skillMappings` but `provenanceFiles:28-32` hardcodes the same `typescript/sdk/...` strings. `['flow','case','bpmn']` is re-listed at `:425` and `:443`. `parseArguments:624` silently accepts a missing option value (`--skills-root` last -> raw `path.resolve(undefined)` TypeError).

---

### Refuted — please don't act on these

**The `verb-gate.yml:74` matrix rename does NOT block merges.** Two of the three reviewers rated this High, on the strength of the workflow's own header comment. It's wrong: `repos/UiPath/skills/rulesets` shows the only required contexts on `main` are `Run RPA skill smoke tests (Windows)` and `Run skill smoke tests`, and `automerge-catalog-refresh.yml:67` explicitly lists "CLI Verb Gate" under advisory / non-required checks.

One reviewer's proposed fix — strip the prefix back to a bare name — would be actively harmful: `skills/uipath-maestro-flow/SKILL.md` and `preview/uipath-maestro-flow/SKILL.md` both declare `name: uipath-maestro-flow`, so `sort -u` would collapse them into one matrix entry and the preview copy would never be gated. **The path-based matrix is the correct design and is load-bearing.** The real defect here is the now-stale comment at `verb-gate.yml:15-18` claiming the gate "is wired as a required check".

**`writeIfChanged` / `changed = f() || changed`** — no bug. Every site (`:364`, `:378`, `:604`, `:612`) puts the effectful call on the left of `||`, so it always executes.

**Idempotency** — sound. All five `adapt*` functions are provably idempotent (`replaceRequired` returns early on `includes(after)`; `` `example/ `` cannot re-match `` `examples/ ``; `adaptBpmnSkill` is guarded), and on a second run `base === theirs`, so `merge-file` returns `ours` verbatim. The `changed: false` assertion at test `:303` is meaningful.

**`collectMappings`** — no bug in the two-ref union, the `add()` collision check, or the `case-api.md`/`bpmn-api.md` exclusion at `:176`. The unconditional adds at `:201-208` are the right call: they turn an upstream deletion into a loud `Mapped source disappeared at target pin` rather than a silent drop.

---

### What's genuinely well done

- `verifyExactMappings` (`:404`) is a strong invariant — byte-equality against `transform(upstream)` with an explicit two-entry allowlist of adapted files. Content cannot drift silently.
- `assertManagedInventory` runs both pre- and post-sync, making the "unmanaged file" class impossible to accumulate.
- Validating the *recorded* pin's existence upstream (`:597`) before merging, and cross-checking that all three provenance headers agree (`:156`), catches a partially-applied previous sync.
- All git invocation goes through `execFileSync`/`spawnSync` with argument arrays — no shell, no interpolation, `maxBuffer` raised deliberately.
- The workflow cross-checks the script's `changed` output against `git diff` and fails on disagreement — cheap belt-and-braces on the one output that gates PR creation. `trap cleanup_branch EXIT` registered *after* the push (`:151-160`) is correctly ordered.
- Exporting the `adapt*` functions makes the pure logic unit-testable, and test `:309` is a good non-tautological pin on `adaptFlowApi`'s narrow `example-eval/` vs `example/` distinction.

---

### Test status

`node --test tests/scripts/sync-maestro-sdk-preview.test.mjs` -> **2/2 pass** (5.2s).

`npm run skills:test` has one failure, `compose-skill-flavor.test.mjs:886` (`You must specify a tag using --tag when publishing a prerelease version`). **Pre-existing** — it fails identically at the merge base `04c0233`. Not this PR.

---

### Overall

The verification layer is the good half of this change: `verifyExactMappings` plus `assertManagedInventory` pin the vendored tree tightly enough that content drift can't ship unnoticed, and the pre-sync pin validation is careful. The synchronisation layer is the weak half — it mutates the tree *including the pin* before verifying it, so a gate failure leaves a state that re-running cannot repair (H2), and it routes every file through a three-way merge that adds five abort modes without adding a single accepted outcome (H3). Separately, the nightly automation's duplicate-PR guard is already defeated by this repo's PR volume (H1).

Suggested shape: land the verification layer and the `verb-gate.yml` path change as-is; fix H1; move the pin rewrite after verification (or make the run transactional); and replace the merge with copy-plus-adaptation that preserves the three header blocks.

uipreliga (approved):
Fix what you agree with and 🚢

PR #2627, opened 2026-08-17
test(uipath-admin): read-back grading, per-area layout, parallel-safe runs for authz/apms/oms [PLT-108711]

jianjunwang2:
Rebased on main (#2667).

Converts the **authz / apms / oms** admin coder-eval tests from `command_executed` command-string matching to **read-back grading** (the #2280 pattern), and makes the admin suite safe for concurrent agents.

- **Grading** — criteria assert tenant state read back through the CLI, anchored on un-fakeable fields (service-issued ids, response codes, row floors). No `skill_triggered`: it inspects the Claude Skill tool, so it can never pass under codex or gemini.
- **Tiers** — reads stay `smoke` (PR gate); writes become `e2e` (daily), since a write costs tenant state.
- **Layout** — per-area folders `authz/ apms/ oms/ identity/ audit/`. The PR gate resolves a changed task to the nearest directory holding YAMLs, so an authz edit no longer samples from all 88 admin tasks. `identity/` and `audit/` are **pure moves** — byte-identical to main apart from the sidecar path and `_shared` depth.
- **Parallel safety** — seeded objects carry a per-run id from `seed.json` (the `uipath-platform/seed.py` convention); checks and cleanup touch only their own run's objects. IP ranges additionally claim a CIDR block and re-read it to confirm ownership, since the allowlist is keyed by CIDR and a duplicate create renames the existing row.
- **Removed** — CRUD lifecycle tests and OMS tenant create/disable: they mutated shared org state that concurrent runs observe, and a grant/revoke arc is state-neutral, so read-back cannot see its middle. Per-command scenario tests cover the same verbs with observable end states.

86 admin tasks: authz 22, apms 10, oms 8, identity 30, audit 16. Dropped `apms_enforcement_enable_e2e` (drives the lockout-capable IP-restriction toggle) and `apms_add_ip_range_e2e` (plain create, already covered).

**Validated** — criteria pass against live CLI output and fail against fabricated or empty JSON (no vacuous passes); 3 agents x 2 tasks run concurrently produced 6 distinct run ids, all passing, org clean afterwards; a no-op agent fails every test and a correct agent passes every test; all 131 sidecar paths resolve and tags are valid on all 86 tasks; #2667's 33 negative-control tests pass on the moved layout.

**Full suite run on this branch:** 86/88 passed. Both failures are handled — the two tests above are removed; `identity_list_users_smoke` fails on a brittleness inherited from main (its criterion `json.load`s a single document, and the agent appended two paginated ones), untouched here because `identity/` is a pure move.


skills/uipath-admin/references/authorization/permission-catalog.md
```diff
@@ -40,10 +40,53 @@ uip admin authorization permissions list --service studio --output json
...
+
+`--service <NAME>` is the right tool when the caller names the service ("which permissions does Orchestrator contribute?"). It is the wrong tool for locating a permission whose owning service you do not know — that is the guessing case this section warns about.
+
+An unrecognized value fails the whole call:
+
+```
+'centralizedaccess' is not a valid --service value.
+```
+
+Known keys: `apps, audit, authz, automationops, casemanagement, communicationsmining, cx, dataservice, documentunderstanding, governedinference, identity, insights, licensing, oms, orchestrator, platform, processmining, reinfer, relay, studio, taskmining, testmanager`.
```
t-hsia: Is enumerating these here best practice? Seems hard to maintain

jianjunwang2: Resolved

t-hsia (approved):

yadvender-uipath (approved):

PR #2636, opened 2026-08-17
docs(uipath-ixp): document field-group nesting

ztan-uipath:
## What this PR does

Two commits. The first is @ztan-uipath's original, rebased onto main. The second removes its Critical Rule and documents, on the `groups` command rows, the behavior verified against `ixp-platform` `develop` (`8184194bc2`) and on a live tenant.

**Verified behavior of `>` in a field-group name**

| Operation | What happens | Verified |
|---|---|---|
| `groups add "Parent > Child"` when `Parent` doesn't exist | `Parent` is created too, as its own label def with 0 fields and its own id. CLI reports one group created. | source + live |
| `groups rename Parent` | Every `Parent > …` child is renamed with it; ids unchanged. | source + live |
| `groups delete Parent` | Every child group is deleted with it, including their fields and annotations. | source + live |
| `groups rename` onto a name that already exists | Rejected with `409 GroupAlreadyExistsError`. Groups cannot be merged by rename. | live |
| Extraction prompt | The full name including `>` is rendered into the model prompt, so renaming changes model input. | source |
| Extraction output | One flat level (group → fields). Nesting never attributes a child group's rows to a parent occurrence. | source (matches original report) |

So `>` is a real hierarchy for taxonomy operations and is not display-only, but it is flat for output. The original observation about output shape was correct; the conclusion that the whole feature is cosmetic was not.

**Changes** — `references/cli-reference.md` only; net change to `SKILL.md` is zero (the first commit's Critical Rule is removed by the second).

- `groups add`: group names can be nested with `>`; the part before it is the parent, the part after is the child, and a missing parent is created too. Nesting has no effect on extraction results, which are always one separate list per group.
- `groups delete`: now warns it also deletes every child group.
- `groups rename`: renaming a parent renames every child group with it.

Kept deliberately small: only what an agent cannot learn from the command's own response. The 409 on renaming onto an existing name is left to the CLI error, which already says what to do. The multi-parent linking advice from the original report (add the parent's identifier as a field on the repeatable group) is not in the skill: it is design guidance rather than something the CLI hides, and the original description below keeps it on record.

---

## Original description (@ztan-uipath)

> ## Summary
>
> - IXP always reports extraction results as a flat, one-level structure regardless of `>` in a field-group name — a group named `Invoice > Line Items` behaves identically to a top-level group literally named that. Renaming a group to add or remove nesting changes only how it displays in the taxonomy tree in some UI surfaces; it has no effect on extraction output, validation tasks, or any relationship between groups.
> - Adds this as Critical Rule 19, and points at the actual mechanism for the case nesting is often reached for: attributing a repeatable child group's rows back to a parent (e.g. two invoices' line items in one file) requires an explicit foreign-key field duplicated into the child group, not a group hierarchy.
>
> ## Why
>
> Working with the skill, we planned and staged a full taxonomy migration — new parent groups, moving ten flat groups under them, a multi-gate evidence/verification procedure — before discovering partway through that the change was purely cosmetic. The skill's docs state that `>` "defines hierarchy" but don't say the hierarchy has no effect on output, which reads as though nesting is structurally meaningful. This documents the actual behavior so the same effort isn't repeated by someone else.
>
> ## Test plan
>
> - [x] Verified current behavior by inspecting a live IXP project's flat, per-document-type taxonomy and its extraction output
> - [x] Checked this isn't already documented elsewhere in the skill (searched for "nesting", "hierarchy", "flat" in SKILL.md and references/)
> - [x] Placed as a new numbered Critical Rule following the existing style and cross-reference conventions


cezara98t:
After looking into it, the change doesn't seem to be needed. The ">" does actually mean hierarchical change.

PR #2644, opened 2026-08-18
docs(maestro-flow): drop inline-agent flow-node prompt placeholders

andreibalas-uipath:
## What

The inline-agent plugin (`uipath-maestro-flow`) and the `uipath-agents` inline-in-flow reference taught that `uipath.agent.autonomous` flow nodes require non-empty `inputs.systemPrompt` / `inputs.userPrompt` placeholders or `uip maestro flow validate` fails. Both claims are wrong. This PR teaches the shell shape instead: omit the prompt keys entirely; prompts live only in the sidecar `agent.json`.

## Why

Verified locally on stock `uip` 1.201.0 against the `BillingResolutionWriter.reference.flow` fixture in three variants:

| Node prompt keys | `flow validate` |
|---|---|
| Present, non-empty | Valid — but the embed shadows the sidecar prompts when Studio Web loads the flow |
| Absent | Valid |
| Present, empty (`""`) | Fails: `[nodes[<id>].inputs.systemPrompt] [SCHEMA_ERROR] System prompt is required` |

Node-embedded prompts are authoritative over `agent.json` on Studio Web load, so the placeholder guidance made skills author flows whose real prompts get shadowed — the opposite of intent. UiPath/cli#3423 formalized the correct shape: `uip agent refresh --inline-in-flow` shell-ifies the parent `.flow`, stripping node prompts and keeping structural inputs (`source`, `agentInputVariables`, `agentOutputVariables`). Companion docs note landed in #2455.

## Change

- `skills/uipath-maestro-flow/.../plugins/inline-agent/impl.md` — both node JSON examples lose the prompt lines; the placeholder rules (wiring section, anti-patterns, JSON-structure note, validator blockquote, What-NOT-to-Do) reverse to "omit the keys"; the debug row and repair recipe flip from "add placeholders" to "delete the empty keys / run refresh", and are retitled to the actual `[SCHEMA_ERROR] System prompt is required` message (the previously documented `[REQUIRED_FIELD]` tag is stale on 1.201.0).
- `skills/uipath-agents/.../inline-in-flow/inline-in-flow.md` — copy-paste node JSON loses the stub prompts; the "validation requires non-empty placeholders" bullet is replaced with the corrected rule.
- `tests/tasks/uipath-maestro-flow/_shared/advisory_billing_resolution_writer.py` — fail message no longer names the node fields (the check already reads the node+sidecar union; wording only, no logic change).

## Validation

- `npm run skills:validate` — OK (26 skills, both flavors).
- `python3 -m py_compile` on the reworded helper — OK.
- `grep` over both references: every remaining `systemPrompt`/`userPrompt` mention is the corrected guidance.

### DIFF
--- a/skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
+++ b/skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md
@@ -213,8 +213,6 @@ Resource body shape is identical to the standalone-agent docs — only the folde
   "typeVersion": "1.0",
   "display": { "label": "Autonomous Agent" },
   "inputs": {
-    "systemPrompt": "You are an agentic assistant.",
-    "userPrompt": "What is the current date?",
     "source": "<projectId-uuid>",        // UUID linking to the inline agent directory
     "agentInputVariables": [],
     "agentOutputVariables": [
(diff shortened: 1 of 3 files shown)


andreibalas-uipath:
Passing coder-eval run for the touched test helper: [Run Coder Eval 32116114828](https://github.com/UiPath/skills/actions/runs/32116114828) on this branch — `skill-flow-devcon-billing-resolution-writer`: SUCCESS, 4/4 criteria, weighted score 1.000 (389s, 1 iteration). The reworded `advisory_billing_resolution_writer.py` advisory ran and passed; the built flow's agent node carried no `systemPrompt`/`userPrompt` keys and all three inputs were delivered via `agentInputVariables` bindings.

PR #2646, opened 2026-08-18
fix(rpa): discovery agent writes its own context files; gate empty projects

RaduAna-Maria:
## Problem

Two defects in the `uipath-project-discovery-agent` flow:

1. **The subagent could not write its own output.** It had no write tools, so it returned the context document and the *main* agent wrote `.claude/rules/project-context.md` + `AGENTS.md`. That was a limitation at the time subagents had no write access. It costs a round-trip and the write silently gets dropped whenever the main agent moves on to the actual task.
2. **It was spawned for new and empty projects.** The only guard was "no `project.json` → skip", so a scaffolded-but-empty project (just `Main.xaml`, nothing authored) paid for a full discovery pass that can only produce a document of empty tables.

## Changes

**`skills/uipath-rpa/agents/uipath-project-discovery-agent.md`**

- `tools:` gains `Write, Edit`. The "**Do NOT write any files**" block becomes "**You own the output files**".
- New **Step 5 — Write the Context Files**: writes `.claude/rules/project-context.md` (verbatim, metadata comment on line 1) and `AGENTS.md`. The `PROJECT-CONTEXT` marker merge logic (missing → write / markers present → replace between them / no markers → append) moved out of SKILL.md into the agent that now performs it.
- New **Step 6 — Return**: emits a `context-files: written (...)` status line followed by the document, so the main agent still gets session context without re-reading what was just written. Hosts that only spawn read-only subagents (UiPath Autopilot) return `context-files: not-written (<reason>)` and the old main-agent write path still applies — the flow degrades gracefully.
- New **Step 2 — Gate**: returns a single `SKIP: <reason>` line, writing nothing, for no `project.json`, zero authored `.xaml`/`.cs`, or a lone untouched scaffolded entry point. A regenerate request does not override it.
- Critical Rules renumbered: rule 3 flipped from "do NOT write any files" to "write both context files yourself"; added the gate rule and an `AGENTS.md` non-destruction rule.

**`skills/uipath-rpa/SKILL.md`** — new `### Skip gate: nothing to discover yet` under § Precondition, applying the same three conditions before spawning (Glob counts exclude `.local/`, `.codedworkflows/`, `.objects/`, `.entities/`, `obj/`, `bin/`). Discovery-flow step 3 now fires **only** on `context-files: not-written`; step 4 treats a `SKIP:` return as a gate trip (write the files yourself after the build, same as greenfield).

The gate is checked in both places on purpose: SKILL.md keeps the main agent from spawning at all, the agent's Step 2 is the backstop for when it is spawned anyway or invoked directly.

**`references/execution-maps-guide.md`, `README.md`** — greenfield/brownfield T1 rows and the agents table follow the reworked flow.

## Validation

- `npm run skills:validate` → `OK - default: 26 skills, 1718 files, 0 replacements; studioweb: 26 skills, 1718 files, 95 replacements.`
- `PYTHONUTF8=1 python scripts/check-skill-status.py` → `OK — 26 skills, manifest valid.` (without `PYTHONUTF8=1` it hits a pre-existing cp1252 decode error on Windows, reproduced on a clean tree)
- Agent frontmatter parses as YAML; description is 327 chars.
- `hooks/validate-skill-descriptions.sh` → clean.

## Notes for the reviewer

- No test task added: `tests/` has no success-criteria type that can assert "a subagent was not spawned" or "this file is absent", so the gate is not gradable with the current vocabulary. Happy to add one if you'd rather approximate it.
- No `plugin.json` change — the agent registration path is unchanged.
- `uipath-rpa` has no flavor markers and no `skill-flavors/studioweb/uipath-rpa/` overrides, so no flavor override review was needed.

### DIFF
--- a/skills/uipath-rpa/agents/uipath-project-discovery-agent.md
+++ b/skills/uipath-rpa/agents/uipath-project-discovery-agent.md
@@ -36,7 +36,7 @@ You are a project discovery agent. Analyze a UiPath automation project, write th
 
 A project with no authored content yields a document of empty tables. Run this gate **before** any discovery work.
 
-Count authored files with Glob, excluding generated and metadata directories (`.local/`, `.codedworkflows/`, `.objects/`, `.entities/`, `.screenshots/`, `.settings/`, `obj/`, `bin/`):
+Count authored files with Glob, excluding every dot-directory (any path segment starting with `.` — `.local/`, `.codedworkflows/`, `.objects/`, `.settings/`, `.templates/`, …) plus `obj/` and `bin/`:
 
 ```
 **/*.xaml     → XAML workflow count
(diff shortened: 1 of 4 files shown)


skills/uipath-rpa/agents/uipath-project-discovery-agent.md
```diff
@@ -31,7 +32,30 @@ You are a project discovery agent. Analyze a UiPath automation project and gener
 3. Fall back to current working directory
 4. Verify `project.json` exists and contains UiPath dependencies before proceeding
 
-### Step 2: Discovery
+### Step 2: Gate — Skip Empty or Freshly-Scaffolded Projects
```
AlvinStanescu, line 35: @RaduAna-Maria can we also support fresh Library projects? I think they have `NewActivity1.xaml` but not 100% sure (please check)?

AlvinStanescu (approved):

skills/uipath-rpa/agents/uipath-project-discovery-agent.md
RaduAna-Maria: Checked against freshly-created Studio 26.0.201 blank projects of all four types — it's `NewActivity.xaml` (no digit). Verified scaffolds:

| Project type | Scaffold entry file | Root Sequence |
|---|---|---|
| Process | `Main.xaml` | empty |
| Template | `Main.xaml` | empty |
| Library | `NewActivity.xaml` | empty (self-closing) |
| Test Automation | `TestCase.xaml` | contains `ui:Comment` "// Blank Test Case" |

Fixed in 5f6b2e3: the gate's entry-point list now covers `NewActivity*.xaml` (wildcard hedges older Studio numbering), and the content test ignores `Comment` activities + ViewState metadata — the blank test scaffold's Comment was also defeating the "no child activity" check. Also aligned the SKILL.md/agent exclusion lists to all dot-directories + `obj/` + `bin/` (fresh projects ship `.templates/`, `.tmh/`, `.project/` too).

Note: `project.json → main` can't replace the name list — fresh Tests projects set `main: Main.xaml` pointing at a file that doesn't exist.

PR #2648, opened 2026-08-18
Update the Run Coder Eval action

Mihaiii:
- use coder_eval_uipath instead of coder_eval
- add Delegate SDK to the harness list

### DIFF
--- a/.github/scripts/refresh-auth.sh
+++ b/.github/scripts/refresh-auth.sh
@@ -25,7 +25,9 @@ set -euo pipefail
 : "${CE_USERNAME:?CE_USERNAME not set}"
 : "${CE_PASSWORD:?CE_PASSWORD not set}"
 
-AUTH_FILE="${HOME}/.uipath/.auth"
+# Output path is overridable: the Windows delegate-sdk path mints to a separate
+# file so the uip CLI's own browser-OAuth login in ~/.uipath/.auth stays intact.
+: "${AUTH_FILE:=${HOME}/.uipath/.auth}"
 mkdir -p "$(dirname "$AUTH_FILE")"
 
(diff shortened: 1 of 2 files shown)


bai-uipath (approved):
**Approve once the coder-eval version resolution stops diverging from the smoke tests** — the delegate wiring itself is solid, and the MSYS path fix is well diagnosed.

## Before merge

- **The framework version must stay consistent with the smoke suites.** Dropping the `tests/.coder-eval-version` read replaces an exact pin with the plugin's open `0.9.x` range, and `uv pip install <path>` ignores the plugin's lockfile, so the range really is open-ended. `smoke-skills`, `activation-gate` and `smoke-rpa-skills` all still pin exactly — an ad-hoc run and a PR gate on the same task can now execute on different framework versions with nothing flagging it, and there's no longer any way to pin or roll back a bad release from the action. `daily.sh` in coder_eval_uipath already does the thing to copy here: read this repo's pin, install with `==`, and keep an override for dry-running a candidate release. Fix: keep the pin read and the version input, passing `coder-eval==$VERSION` alongside the plugin in one `uv` invocation so both co-resolve.
- **Gate the plugin install on `delegate-sdk`.** `register()` monkeypatches core `apply_overrides` and mutates the pricing table, which is why `daily.sh` gates the same install; installing it unconditionally also makes every claude/codex run depend on the private checkout succeeding.

## Optional

- **Three delegate env vars from the plugin's own command builder aren't forwarded** — including the stdio trace switch, which is what an operator debugging a hung turn on this workflow would reach for.
- Minor: no `always()` teardown for the Windows interop; only `AUTH_TOKEN` is validated while the four org/tenant vars are exported unchecked; the override flag's bracket list is word-split unquoted (glob-fragile — an array avoids it); `persist-credentials: false` on the private checkout would keep the PAT out of the docker build context.

Mihaiii:
New run: https://github.com/UiPath/skills/actions/runs/32239477951

PR #2649, opened 2026-08-18
fix(uipath-agents,uipath-maestro-flow): fix six failing nightly agent tasks — skill-doc gaps, prompt redirect traps, checker/cleanup defects

andreibalas-uipath:
Fixes six failing nightly `uipath-agents` coder-eval tasks. Each fix is grounded in transcript analysis of the failing runs and verified with a local passing run on `claude-sonnet-5` (`experiments/default.yaml`, same model as nightly).

## 1–2. Coded-in-flow tasks: skill-doc gaps (`245cf5aef`)

`skill-agent-coded-in-flow-e2e` and `skill-agent-coded-in-flow-published` both ended `MAX_TURNS_EXHAUSTED` (80 turns, scores 0.690 / 0.483) with the same root cause: the agent completed the coded-agent stage, then spent its final 30+ tool calls grepping both skills for the agent node's `inputs` contract — documented only as a one-sentence fragment — and never edited the `.flow` file. A secondary defect cost the e2e run ~20 turns: `quickstart.md` Scenario 2 told the agent to run `uip codedagent new` from the solution root and claimed it creates a subfolder; it scaffolds into CWD.

- **`uipath-maestro-flow` `plugins/agent/impl.md`** — new **§ Wiring Inputs**: one `inputs.<field>` entry per `inputDefinition.properties` property; both value shapes (`jsExpression` bare `$vars` expression, `literal` `{{ }}` text template — both product-canonical); `=js:`-string anti-pattern; `<FIELD_TYPE>` read from the schema, never invented; generic trigger→agent→end worked example. Populated the node-instance example's `inputs`; corrected the "`inputDefinition` typically empty" claim; outputs rule now allows `output` declared with exactly `source: "=this"` + the agent's output schema.
- **`uipath-agents` `embedding-in-flows.md`** — § Wiring the Agent's Inputs expanded from one sentence to the same contract.
- **`uipath-agents` `quickstart.md`** — Scenario 2 step 3: `mkdir`/`cd` before scaffolding, `codedagent new` scaffolds-into-CWD note, `uv add` moved after `new` (requires the generated `pyproject.toml`), redundant step-4 `cd` dropped.

Grounded in two Studio Web solution exports of the same flow, identical except the input value shape — proving both shapes canonical. Examples use generic `<PLACEHOLDER>` values only.

**Passing runs:**

| Task | Nightly | This branch |
|---|---|---|
| coded-in-flow e2e | MAX_TURNS_EXHAUSTED, 0.690 | **1.000, 12/12, 60/80 turns** |
| coded-in-flow published | MAX_TURNS_EXHAUSTED, 0.483 | **1.000, 7/7, 57/80 turns** |

Published run's doc-consultation phase dropped from ~35 tool calls to 8; both runs wired the node in the documented `jsExpression` shape and passed `uip maestro flow validate`. `npm run skills:validate` passes (default + studioweb).

## 3. `deploy_folder_and_rebump`: checker + tenant-state defects (`fed0156f7`)

Checker demanded exactly version `0.0.2`, but versions left on the tenant feed by earlier runs forced the agent past it (a run legitimately reached 0.0.8) — failing correct behavior. Now accepts any bump strictly greater than the pre-seeded `0.0.1`, and a best-effort idempotent `post_run` sweep restores the acme-echo feed to the `0.0.1` baseline via the Orchestrator OData API (the CLI has no package-delete verb). Verified: SUCCESS 1.0; feed restored after the run.

## 4. `eval_custom_evaluator`: prompt triggered the functions redirect (`ff8d107e5`)

The prompt's "pure deterministic Python, no LLM" is the exact trigger for the skill's Coded Function redirect, so agents correctly rerouted to `uipath-functions` (no eval lifecycle) and never ran the graded `uip codedagent` evaluator commands. Reworded as a non-deterministic greeter (an agent workload that needs evals); hermeticity preserved via the latency bar and no-sign-in constraint. Verified: SUCCESS 1.0, all 4 criteria.

## 5. `hitl_create_task_with_app`: prompt demanded an unobservable outcome (`ca37fdbb4`)

A plain CreateTask resume delivers only the task's data dict (the app's inOut fields), never `task.action` — verified via live suspend/resume. The old prompt's "carry on with the call they made" demanded the approve/reject outcome, pushing agents to CreateEscalation (graded wrong) or into SDK-source research spirals that blew the 1200s budget. Reframed as a sign-off gate carrying the reviewer's comment forward; bindings hint dropped so the criterion tests the skill, not the prompt. Verified: 1.000 twice (261s/64 turns, 397s/95 turns).

## 6. `deploy_my_workspace`: prompt triggered the functions redirect (`bb1d8d89b`)

"Minimal no-LLM coded agent that echoes the input" hit the same Coded Function redirect, so agents shipped via `uip function pack/publish` and never matched the graded `uip codedagent deploy --my-workspace` pattern. Reworded as an LLM-backed one-line summarizer with no plausible deterministic implementation; ungraded local-run step dropped. Verified: SUCCESS 1.0, all 3 criteria, no `uip function` commands.

## 7. `openai_agents_handoff`: checker rejected kebab-case agent names (`ade3b8c93`)

Grader bug: the checker normalized agent names with `removesuffix("agent")`, so kebab-case names (`billing-agent` — the naming style the prompt itself models with `triage-agent`) normalized to `billing-` and failed the triage/billing/technical check on an otherwise perfect solution (nightly scored 0.524 with the three command criteria green). Normalization is now separator-aware (`[\s_-]*agents?$`) and accepts single-quoted `name=` kwargs. Local reruns surfaced a second latent trap: agents that model the customer's message inside `CustomerInput` make `message` a required input field, so the checker's fixed re-run payload fails context validation and the runtime silently passes `context=None`, crashing instructions callables. The prompt now pins the shape (single `customer_id` field; the message travels as the standard `messages` input) and names the `billing-agent`/`technical-agent` specialists explicitly, making the graded names ground-truth anchors. Verified: SUCCESS 1.0, 4/4 criteria, 229s — all nine checker assertions including the live `uip codedagent run agent` gateway re-run; generated entry-point schema requires exactly `['messages', 'customer_id']`.

### DIFF
--- a/skill-flavors/studioweb/uipath-agents/references/coded/quickstart.md
+++ b/skill-flavors/studioweb/uipath-agents/references/coded/quickstart.md
@@ -14,7 +14,7 @@ Use when the coded agent is tightly coupled to one flow and lives as a sibling f
 <!--skill-flavor:agent-solution-registration:end-->
 
 <!--skill-flavor:agent-scaffold-solution-root:start-->
-3. **Scaffold the coded agent as a sibling folder.** From the solution root (`/solution`) — `uip codedagent new` scaffolds into the **current directory**, it does NOT create a subfolder, so create the agent folder first and run everything inside it:
+3. **Scaffold the coded agent as a sibling folder.** From the solution root (`/solution`):
 <!--skill-flavor:agent-scaffold-solution-root:end-->
 
 <!--skill-flavor:agent-scaffold-result-paths:start-->
--- a/skill-flavors/studioweb/uipath-agents/references/coded/quickstart.md
+++ b/skill-flavors/studioweb/uipath-agents/references/coded/quickstart.md
@@ -14,7 +14,7 @@ Use when the coded agent is tightly coupled to one flow and lives as a sibling f
 <!--skill-flavor:agent-solution-registration:end-->
 
 <!--skill-flavor:agent-scaffold-solution-root:start-->
-3. **Scaffold the coded agent as a sibling folder.** From the solution root (`/solution`):
+3. **Scaffold the coded agent as a sibling folder.** From the solution root (`/solution`) — `uip codedagent new` scaffolds into the **current directory**, it does NOT create a subfolder, so create the agent folder first and run everything inside it:
 <!--skill-flavor:agent-scaffold-solution-root:end-->
 
 <!--skill-flavor:agent-scaffold-result-paths:start-->
--- a/skills/uipath-agents/references/coded/embedding-in-flows.md
+++ b/skills/uipath-agents/references/coded/embedding-in-flows.md
@@ -104,6 +104,22 @@ Without `--local`, `registry list`/`get` query the tenant registry (Orchestrator
 
 ## Wiring the Agent's Inputs
 
-For inputs that reference flow variables, use `{ "type": "literal", "expression": "{{ $vars.X }}", "fieldType": "string" }` — NOT `=js:...` expressions. `=js:` ships as a literal string to the agent activity and fails at runtime with `Cannot find name '<identifier>'`.
+One `inputs.<field>` entry per property in the agent's input schema (`entry-points.json`, mirrored in the definition's `inputDefinition.properties`). Two valid value shapes, same binding:
+
+```json
+"inputs": {
+  "<INPUT_FIELD>": {
+    "type": "jsExpression",
+    "expression": "$vars.<UPSTREAM_NODE_ID>.output.<FIELD>",
+    "fieldType": "<FIELD_TYPE>"
+  }
+}
+```
+
+- `jsExpression` — bare `$vars...` expression, no `=js:` prefix. Upstream values read as `$vars.<nodeId>.output.<field>`; flow globals as `$vars.<global>`.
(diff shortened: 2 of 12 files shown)


al3xanndru (approved):

skill-flavors/studioweb/uipath-agents/references/coded/quickstart.md
```diff
@@ -14,7 +14,7 @@ Use when the coded agent is tightly coupled to one flow and lives as a sibling f
 <!--skill-flavor:agent-solution-registration:end-->
 
 <!--skill-flavor:agent-scaffold-solution-root:start-->
-3. **Scaffold the coded agent as a sibling folder.** From the solution root (`/solution`):
+3. **Scaffold the coded agent as a sibling folder.** From the solution root (`/solution`) — `uip codedagent new` scaffolds into the **current directory**, it does NOT create a subfolder, so create the agent folder first and run everything inside it:
```
astridstaicu: the command `uip codedagent new` works in studio web? We should check. If not, we can replace it in the flavor with the `CreateProjects` tool

astridstaicu (approved):

skill-flavors/studioweb/uipath-agents/references/coded/quickstart.md
andreibalas-uipath: Confirmed that `uip codedagent new` does not work out of the box in SW, since it also requires a separate `mkdir` at the solution root level. Reverting the changes and will follow-up in a proper SW flavor override.

PR #2653, opened 2026-08-18
test(uipath-api-workflow): raise coverage 63% → 81%; fix cloud-diagnose guidance

rares-baesu-uipath:
## Why

API Workflow is red on the Coding Agents Scorecard, driven by **Test Coverage vs Skills**. Two causes:

1. The entire post-publish surface documented in `references/operating-published-workflows.md` — publish, deploy, invoke, diagnose a *deployed* workflow — had **zero** tests. `package_solution` stopped at `pack`; `operate/run_execute` runs locally.
2. Large documented areas had no coverage at all: 8 of the troubleshooting fault catalog's categories, 4 of the 10 control-flow patterns, `httpRetryConfig`, and rule 21.

## What this adds

**10 tasks (23 → 33).** Measured with one fixed rule set applied to both the 08/03 commit and this branch:

| | 08/03 (`3b49079c4`) | This branch |
|---|---|---|
| Overall | 63.5% | **81.4%** |
| Components | 22/55 (40%) | **37/54 (69%)** |
| Workflow steps | 6/8 | **7/8** |
| Critical rules | 18/23 | **21/23** |
| build / operate / diagnose | 71 / 31 / 53 | **78 / 74 / 75** |
| Worst mode | **31%** | **74%** |

`operate` had **zero** covered components. The mode spread — build-solid, blind elsewhere — is the substantive change.

### Tenant-backed (2)

- **`operate/publish_deploy`** — pack → publish → deploy → `or jobs start` → job `Successful` → `deploy uninstall`. First task past the publish boundary.
- **`diagnose/cloud_job_fault`** — root-cause a faulted deployed job. Self-seeding: `pre_run` deploys a workflow that passes `validate` and throws at runtime, then drives a job to `Faulted`.

### Offline (8)

`export_chain`, `loop_iterator_name`, `response_shape`, `nested_aggregation` (diagnose, each a checked-in fixture verified to fail in exactly one way) · `consent_gate`, `http_retry_config`, `skip_and_continue`, `multiway_branch` (build).

## Skill fix included

`operating-published-workflows.md` pointed agents at two dead ends. Verified against a deliberately-faulting deployed workflow (alpha, uip 1.200.0):

| Command | Reality |
|---|---|
| `uip or jobs get` | `Data.State` = `Faulted`, `Data.Info` carries the error |
| `uip or jobs logs` | Lifecycle lines only. Reports **`Workflow completed` for a Faulted job** |
| `uip traces spans get --job-key` | Fails: `"Error retrieving trace ID for job"` |

The doc recommended the bottom two and specifically steered API-workflow diagnosis toward `traces spans get`. An agent following it got nothing, or concluded the run succeeded. `cloud_job_fault` fails any diagnosis claiming success — the exact conclusion `jobs logs` produces. The `studioweb` flavor replaces that block wholesale, so the canonical edit does not leak into it.

## Verification

Every task run live: **10/10 SUCCESS at score 1.000.** Both tenant tasks verified to leave the tenant clean (no packages, no folders).

⚠️ **Run against a local dev build of uip 1.200.0, not the published CLI.** Needs one `run-coder-eval.yml` pass before merge.

## Notes for reviewers

Two authoring traps, both caught by a first run rather than by review — documented in the commit messages so the next author does not repeat them:

1. A `>-` block scalar folds newlines into spaces, silently collapsing multi-line Python in a `command:` into one invalid line. `bash -n` passes it — the shell is valid, only the embedded Python is broken. Non-trivial checks belong in a `check_*.py`.
2. An advisory criterion with non-zero weight dilutes the score of a task that passes cleanly (observed 0.889 vs 1.000). Advisories now use `weight: 0` + `pass_threshold: 0`, matching `operate/run_execute`.

Also worth knowing: a task whose `pre_run` timeout exceeds coder-eval's 300s cap is **silently skipped** at load and reported as `0/0 succeeded`, not failed.

---

## Review round: eight findings from @liviubobocu, all fixed (`581d2a06b`)

Each was reproduced before fixing and re-verified after. Two invalidated claims this PR had asserted, so those claims are **removed**, not softened.

| Finding | Fix |
|---|---|
| `check_publish_deploy_job.py` passed a **faulted** job — the grep matched the envelope's `"Result": "Success"` | Reads `Data.State`, requires exactly `Successful`. Verified across Faulted/Stopped/Pending/Successful and a no-State payload |
| `cloud_job_fault` **leaked its grading sentinel** — the seed left the fixture in the agent's cwd | Scaffold deleted once deployed; task still 5/5, so the sentinel came from `Data.Info` |
| `skip_and_continue` **never triggered the skip** — all-numeric inputs, TryCatch-deleted variant scored 8.5/12.5 | Every graded input now carries a non-numeric element. Proved by splicing the TryCatch out of a passing solution: both inputs then fail `Not a number: oops` |
| `validate` does **not** enforce the exponential-multiplier rule | Confirmed — exponential-without-multiplier, constant-with-multiplier and `"strategy": "banana"` all return `Valid`. Description corrected; validate criterion relabelled syntax-only, weight 2.0 → 1.0 |
| deploy regex matched `deploy uninstall`, so teardown satisfied "Agent deployed" | Narrowed to `deploy run` |
| robot/machine setup is hard for an agent to guess | Correct, and my first note was **wrong**: this tenant has 3 serverless templates, which is why jobs "just worked". Now documented as a tenant prerequisite — symptom (`jobs start` fails / job stuck `Pending`), check (`uip or machines list`), owner (`uipath-platform`) |
| **`Operation` is not null** | Correct. Absent from a stale local dev build; the published CLI emits it. Docstrings fixed, `folder_gone()` deleted, `already_uninstalled()` reads `Operation` as intended. Verified live: `already uninstalled; nothing to undo, failed=0` |
| **activation-decay didn't reproduce** | Removed everywhere. `4007` means **already uninstalled**, not a closed window — all 8 deployments were cleanly removed and nothing was ever stranded. Graded teardown stays, justified on its own merit |

**Automated review:** `response_shape`'s duplicate criterion relabelled. Its `nested_aggregation` finding was checked and **does not hold** — the buggy fixture returns `perRow: 6` for `[[1],[2],[3]]`, not 3, so that criterion discriminates correctly and is unchanged.

**Two prompts tightened** after a rebuilt CLI exposed real ambiguity: `http_retry_config` now names `Workflow.json` and forbids hand-written retry code (an agent had produced a Python `urllib` backoff module and never touched the workflow); `response_shape` routes through `validate` and rules out the validates-but-returns-`"(no output)"` near-miss an agent actually shipped.

### Verification

`publish_deploy` **6/6** and `cloud_job_fault` **5/5** against the **published** CLI on a live tenant · `response_shape` **3/3** · other diagnose tasks green · `Operation` path unit-checked against a real published-CLI row.

### ⚠️ Known fragile — needs a CI verdict

`http_retry_config` sits at the agent context ceiling for this skill (263 KB of SKILL.md + references) and has failed three distinct ways across runs — prompt-too-long, max-turns, and wrong-artifact. **1/3 locally.** Please let CI in a clean container decide; if it can't go green, drop that one task and land the other nine.

Local full-suite runs are not a useful signal here: six tasks hit `MAX_TURNS_EXHAUSTED` including **five that predate this PR**, on a laptop whose `skills/` tree the `prepack` bug repeatedly stripped mid-run.

### DIFF
--- a/tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/cloud_job_fault.yaml
+++ b/tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/cloud_job_fault.yaml
@@ -1,0 +1,107 @@
+task_id: skill-api-workflow-diagnose-cloud-job-fault
+description: >
+  E2E (diagnose): root-cause a DEPLOYED API workflow job that faulted in cloud,
+  from the job surface alone. The agent gets a job id in incident.json and
+  nothing else; starting a fresh job is graded as a failure.
+
+  Fills the skill's largest diagnose gap. The two existing diagnose tasks are
+  local — validate_fix repairs a file that fails `validate`, runtime_fix repairs
+  one that fails `run --no-auth`. Neither touches a cloud fault, and the whole
+  post-publish half of references/operating-published-workflows.md had no
+  diagnostic coverage.
+
+  It also grades the trap that surface actually sets. Verified on alpha
+  2026-08-18 (uip 1.200.0) against a deliberately-faulting deployed workflow:
+  `uip or jobs logs` returns only "Workflow started" / "Workflow completed" — it
+  reports COMPLETED for a Faulted job and never carries the error — and
+  `uip traces spans get --job-key` fails outright with "Error retrieving trace ID
+  for job". Only `uip or jobs get` carries the fault, in Data.State + Data.Info.
+  An agent that diagnoses from logs concludes the run was fine; the check fails
+  any diagnosis claiming success.
+
+  Self-seeding: pre_run deploys fixtures/faulting-workflow.json (passes
+  `validate`, throws at runtime), starts a job, and polls until it is Faulted.
+  The sibling platform task waits on an externally-provisioned
+  E2E_FAULTED_JOB_KEY and has consequently never run — a fixture that needs a
+  human is a task that stays skipped.
+
+  The deployment is left standing for the agent: uninstalling destroys the job
+  records, after which `uip or jobs get` returns Result: Failure with an empty
+  State. post_run tears it down; see cleanup_fault_deploy.py for why that
+  sometimes leaves a cosmetic history row.
+
+  Requires a tenant — runs in CI (run-coder-eval.yml) or against a local
+  `uip login`, not the tempdir smoke gate.
+tags: [uipath-api-workflow, e2e, mode:diagnose, lifecycle:discover]
+
+sandbox:
+  python: {}
+
+pre_run:
+  # uuid8 + parent_folder_path, and ensures the parent folder exists on the tenant.
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-platform/seed.py"
+    timeout: 60
+  # Split in two because coder-eval hard-caps a single pre_run step at 300s and
+  # pack + publish + deploy alone costs ~120s. Each phase gets its own budget.
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/seed_faulted_job.py deploy"
+    timeout: 300
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/seed_faulted_job.py fault"
+    timeout: 300
+
+post_run:
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/cleanup_fault_deploy.py"
+    timeout: 240
+
+run_limits:
+  expected_turns: 16
+  turn_timeout: 600
+
+initial_prompt: |
+  An API workflow we run on a schedule failed overnight and nobody knows why.
+  The handover left the job details in `incident.json` — that is everything we
+  have. The job is long finished, and I do not want it re-run: whatever broke
+  may have side effects.
+
+  Work out what actually went wrong, from the job itself.
+
+  Leave two things behind:
+  - `job_evidence.json` — the raw command output your diagnosis rests on
+  - `diagnosis.txt` — what failed and why, in your own words
+
+  Do NOT ask for approval, confirmation, or feedback.
+
+success_criteria:
+  - type: command_not_executed
+    description: "Agent diagnosed the existing job instead of re-running the workflow"
+    tool_name: "Bash"
+    command_pattern: '(uip|\$UIP)\s+or\s+jobs\s+start'
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent queried the faulted job (the only surface carrying the fault)"
+    tool_name: "Bash"
+    command_pattern: '(uip|\$UIP)\s+or\s+jobs\s+get\b'
+    min_count: 1
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: file_exists
+    description: "Raw evidence the diagnosis rests on was captured"
+    path: "job_evidence.json"
+    weight: 1.0
+    pass_threshold: 1.0
+
+  - type: file_exists
+    description: "A root cause was recorded"
+    path: "diagnosis.txt"
+    weight: 1.0
+    pass_threshold: 1.0
+
+  - type: run_command
+    description: "Diagnosis names the real upstream failure and does not claim the run succeeded"
+    command: "python3 $TASK_DIR/check_cloud_fault_diagnosis.py"
+    timeout: 180
+    expected_exit_code: 0
+    weight: 6.0
+    pass_threshold: 1.0
--- a/tests/tasks/uipath-api-workflow/http-retry/http_retry_config/http_retry_config.yaml
+++ b/tests/tasks/uipath-api-workflow/http-retry/http_retry_config/http_retry_config.yaml
@@ -3,15 +3,23 @@ description: >
   Build: author a workflow-level `httpRetryConfig` with exponential backoff. The
   156-line references/http-retry-config.md had zero coverage.
 
-  The graded detail is the constraint the reference calls out and the validator
-  enforces: `backoff.multiplier` is REQUIRED when strategy is "exponential" and
-  must be ABSENT otherwise. An agent that copies the constant-backoff example and
-  swaps the strategy string produces an invalid workflow, so `validate` passing is
-  a real signal here rather than a formality.
+  The graded detail is the constraint the reference documents: `backoff.multiplier`
+  is REQUIRED when strategy is "exponential". **`validate` does NOT enforce it** —
+  measured on uip 1.199.0, exponential-without-multiplier, constant-WITH-multiplier
+  and even `"strategy": "banana"` all return Status: Valid. The `json`-shape check
+  below is therefore the only gate that grades the rule; the validate criterion is
+  kept at low weight as a syntax sanity check and proves nothing about the retry
+  config.
 
   Local-only: no HTTP call is made, only the retry declaration is authored. Retry
   behaviour itself is not observable offline (it needs a failing GET against a live
   endpoint), which is why this grades the declaration and not a retry count.
+
+  The prompt names `Workflow.json` and forbids hand-written retry code on purpose:
+  an earlier revision said only "give the workflow a retry policy", and an agent
+  read that as an instruction to implement retrying, producing a Python module with
+  urllib and a backoff loop while never touching the workflow file or running the
+  CLI.
 tags: [uipath-api-workflow, integration, mode:build, lifecycle:generate, feature:http]
 
 sandbox:
@@ -22,12 +30,14 @@ run_limits:
   max_turns: 30
 
 initial_prompt: |
-  Create an API workflow project `FlakyFetch` that returns { "ok": true }, and give
-  the workflow a retry policy for its HTTP calls: up to 4 retries, starting at 2
-  seconds, backing off exponentially, capped at 60 seconds, retrying on 429 and 503
-  and on network errors, and honouring Retry-After.
+  Create an API workflow project `FlakyFetch` returning { "ok": true }, and declare
+  the workflow's HTTP retry policy in `Workflow.json` itself — a top-level setting,
+  not retry code you write.
 
-  Validate it. Do not run it.
+  Policy: 4 retries, 2s initial delay, exponential backoff, 60s cap, retry on 429
+  and 503 and on network errors, honour Retry-After.
+
+  Validate. Do not run.
 
 success_criteria:
   - type: run_command
@@ -60,13 +70,13 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: run_command
-    description: "Workflow validates with the retry config in place"
+    description: "Workflow still validates (syntax only — validate accepts invalid retry configs)"
     command: >-
       WF=$(find . -name Workflow.json -not -path './Workflow.json' | head -1);
       uip api-workflow validate "$WF" --output json | grep -q '"Status": "Valid"'
     timeout: 60
     expected_exit_code: 0
-    weight: 2.0
+    weight: 1.0
     pass_threshold: 1.0
 
   - type: command_executed
--- a/tests/tasks/uipath-api-workflow/http-retry/http_retry_config/http_retry_config.yaml
+++ b/tests/tasks/uipath-api-workflow/http-retry/http_retry_config/http_retry_config.yaml
@@ -1,0 +1,78 @@
+task_id: skill-api-workflow-httpretry
+description: >
+  Build: author a workflow-level `httpRetryConfig` with exponential backoff. The
+  156-line references/http-retry-config.md had zero coverage.
+
+  The graded detail is the constraint the reference calls out and the validator
+  enforces: `backoff.multiplier` is REQUIRED when strategy is "exponential" and
+  must be ABSENT otherwise. An agent that copies the constant-backoff example and
+  swaps the strategy string produces an invalid workflow, so `validate` passing is
+  a real signal here rather than a formality.
+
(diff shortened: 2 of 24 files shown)


tests/tasks/uipath-api-workflow/operate/publish_deploy/check_publish_deploy_job.py
```diff
@@ -0,0 +1,124 @@
...
+    )
+try:
+    payload = json.loads(status_file.read_text())
+except (OSError, json.JSONDecodeError) as exc:
+    sys.exit(f"FAIL: ./job_status.json is not valid JSON ({exc}) — expected raw `uip or jobs get` output")
+
+blob = json.dumps(payload)
+if not GUID.search(blob):
+    sys.exit("FAIL: ./job_status.json carries no job identifier — does not look like real CLI output")
+if not re.search(r"success", blob, re.I):
```
liviubobocu: check_publish_deploy_job.py passes a faulted job. re.search(r"success", blob, re.I) matches the CLI envelope's "Result": "Success", which is present on every successful call regardless of job state. Demonstrated against a faulted payload. Should read Data.State, as its sibling checker does.

tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/cloud_job_fault.yaml
```diff
@@ -0,0 +1,107 @@
+task_id: skill-api-workflow-diagnose-cloud-job-fault
```
liviubobocu, line 1: cloud_job_fault leaks its own grading sentinel. The seed leaves FaultSolution/OrderStatus/Workflow.json — containing ORDER_SERVICE_UNREACHABLE — in the agent's working directory, contradicting both the prompt ("that is everything we have") and the checker's stated guarantee. Delete the scaffold after deploy.

tests/tasks/uipath-api-workflow/skip-and-continue/skip_and_continue.yaml
```diff
@@ -0,0 +1,97 @@
+task_id: skill-api-workflow-skip-and-continue
```
liviubobocu, line 1: never triggers skip
 Both graded inputs are all-numeric; a variant with the TryCatch deleted passes 8.5 of 12.5 weight. Adding one junk element fixes it.

tests/tasks/uipath-api-workflow/http-retry/http_retry_config/http_retry_config.yaml
```diff
@@ -0,0 +1,78 @@
+task_id: skill-api-workflow-httpretry
```
liviubobocu, line 1: validate returns Valid for exponential-without-multiplier, constant-with-multiplier, and "strategy": "banana". The task's Python check does grade the rule correctly — only the description and the weight-2.0 "validates" criterion are unearned.

tests/tasks/uipath-api-workflow/operate/publish_deploy/publish_deploy.yaml
```diff
@@ -0,0 +1,131 @@
+task_id: skill-api-workflow-operate-publish-deploy
```
liviubobocu, line 1: deploy regex also matches deploy uninstall, so teardown satisfies "Agent deployed"

tests/tasks/uipath-api-workflow/operate/publish_deploy/publish_deploy.yaml
```diff
@@ -0,0 +1,131 @@
...
+
+  Because teardown removes the folder, job evidence must be captured DURING the
+  run into ./job_status.json — the grader cannot query jobs on a folder that no
+  longer exists. Same pattern as operate/run_execute's run_85.json.
+
+  The project is seeded (scaffolded via `uip solution init` + `uip api-workflow
+  init`, so it carries the real Studio Web shape) because authoring is not the
+  graded behavior here — the build-mode tasks own that.
+
+  Requires a tenant — runs in CI (run-coder-eval.yml) or against a local
```
liviubobocu, line 31: maybe document here another skill to create a robot, this is very difficult/time consuming for an agent to guess, at least from what I have tried so far

tests/tasks/uipath-api-workflow/operate/publish_deploy/cleanup_deploy.py
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env python3
```
liviubobocu, line 1: Operation is not null. Both rows returned Operation = Uninstall, OperationStatus = Successful. This invalidates docstrings in both cleanup_deploy.py and check_publish_deploy_job.py, and makes cleanup_deploy.py's folder_gone() workaround unnecessary complexity.

tests/tasks/uipath-api-workflow/operate/publish_deploy/cleanup_deploy.py
```diff
@@ -0,0 +1,159 @@
...
+        # alone would blow coder-eval's post_run cap. If the tenant is slower,
+        # the uninstall is still submitted server-side and converges unwatched.
+        code, env = uip("solution", "deploy", "uninstall", name, "--timeout", "100", "--yes")
+        if code == 0 and env.get("Result") == "Success":
+            logger.info("uninstalled deployment %s", name)
+            removed += 1
+        else:
+            failed += 1
+            logger.warning(
+                "could not uninstall %s: %s — likely the activation-decay window has "
```
liviubobocu: The activation-decay claim didn't reproduce. apiwffault-4b4e1119 uninstalled cleanly after standing far longer than "a few minutes" — no 4007. That claim is the stated justification for making teardown the graded agent's job and for tolerating leftover rows.

tests/tasks/uipath-api-workflow/operate/publish_deploy/check_publish_deploy_job.py
rares-baesu-uipath: Confirmed and fixed in 581d2a06b. The envelope carries `"Result": "Success"` on every successful call, so the grep could never fail. Now reads `Data.State` and requires exactly `Successful` — verified across Faulted / Stopped / Pending / Successful and a payload with no State field. (Fixing it surfaced a `NameError` on the new failure path, also fixed.)

tests/tasks/uipath-api-workflow/diagnose/cloud_job_fault/cloud_job_fault.yaml
rares-baesu-uipath: Right — the seed left the scaffold in the agent's cwd, so the sentinel was readable off disk and both the prompt and the checker's guarantee were false. Fixed in 581d2a06b: the scaffold is deleted once deployed. Task still passes 5/5, so the sentinel now genuinely comes from `Data.Info`.

tests/tasks/uipath-api-workflow/skip-and-continue/skip_and_continue.yaml
rares-baesu-uipath: Correct. Fixed in 581d2a06b — every graded input now carries a non-numeric element, and the prompt states a non-number must be raised on and caught. Verified by splicing the TryCatch out of a passing solution: both inputs then fail with `Not a number: oops`. Previously the TryCatch was decorative, so this component was being counted as covered by a test that never exercised it.

tests/tasks/uipath-api-workflow/http-retry/http_retry_config/http_retry_config.yaml
rares-baesu-uipath: Reproduced all three — exponential-without-multiplier, constant-with-multiplier and `"strategy": "banana"` all return `Valid`. Fixed in 581d2a06b: the description now states validate does NOT enforce it and names the json-shape check as the only real gate; the validate criterion is relabelled syntax-only and dropped from weight 2.0 to 1.0.

tests/tasks/uipath-api-workflow/operate/publish_deploy/publish_deploy.yaml
rares-baesu-uipath: Confirmed — `deploy\b(?!\s+(?:list|--help|-h))` matched `deploy uninstall`, so teardown alone satisfied it. Narrowed to `deploy run` in 581d2a06b; verified `uninstall` and `list` no longer match.

tests/tasks/uipath-api-workflow/operate/publish_deploy/publish_deploy.yaml
rares-baesu-uipath: You were right, and my first attempt at this note was wrong. I'd written that no machine template is needed — true only because this tenant already carries three serverless templates, which is why jobs 'just worked' for me. 581d2a06b documents it as a tenant prerequisite instead: serverless capacity must exist at tenant level (typically `Default Serverless`), the symptom when missing is `jobs start` failing or the job never leaving `Pending`, checked with `uip or machines list`, owned by uipath-platform.

tests/tasks/uipath-api-workflow/operate/publish_deploy/cleanup_deploy.py
rares-baesu-uipath: You're right. `Operation` is absent from a stale local dev build I was using, but the published CLI emits it — all eight deployments from this branch read `Operation=Uninstall, OperationStatus=Successful`. 581d2a06b corrects both docstrings, deletes the `folder_gone()` stand-in, and lets `already_uninstalled()` read `Operation` as originally intended. Verified live: `already uninstalled; nothing to undo, failed=0`.

tests/tasks/uipath-api-workflow/operate/publish_deploy/cleanup_deploy.py
rares-baesu-uipath: Conceded — the claim is removed everywhere in 581d2a06b. `4007` means ALREADY UNINSTALLED, not that a window closed; I was re-uninstalling completed deployments and invented a mechanism to explain the error. Nothing was ever stranded. Same root cause as your `Operation` finding: the stale dev build hid the field that would have told me. Graded teardown stays, but justified on its own merit as an operate skill rather than on the decay claim.

liviubobocu:
ran the skills via Claude, here are the found issues, most preblematic one being the tenant one:

1. Both tenant tasks are blocked today, and the new prerequisite note names the wrong check. uip or jobs start on the folder deploy run provisions fails HTTP 409: Couldn't find any user with unattended robot permissions in the current folder — bare, with --folder-path, and with --runtime-type Serverless. Every documented check passes: or machines list shows Default Serverless and [Default] Cloud Robots - Serverless, licenses info shows Unattended 1 allowed / 0 used / not expired, users list-in-folder shows Folder Administrator. All templates report 0 slots for every type. The same seed succeeded on this tenant yesterday (job 19fa9e54…, Type: Unattended, Successful), so it's undetectable tenant state. cloud_job_fault pre_run die()s; publish_deploy is unpassable. cloud_job_fault carries no prerequisite note at all despite the identical dependency.

2. seed_faulted_job.py throws away the only informative field. die(f"could not start a job …: {res.get('Message')!r}") prints 'Error starting job'; the reason lives in Instructions (HTTP 409: …). Same in its pack / publish / deploy dies, and the uip() helper discards returncode and stderr. I had to re-run by hand to learn why pre_run failed — which is exactly the "looks nothing like a test defect" scenario the new note warns about.

3. 503 is too weak a sentinel. SENTINELS is OR-ed, so diagnosis.txt reading "It failed with a 503 from somewhere" passes criterion 2 — a plausible guess, which the docstring says cannot pass. (A pasted job GUID containing 503 would too, ~0.7% of GUIDs.) Requiring order_service_unreachable, or both tokens, closes it.

4. response_shape's duplicate criterion was relabelled, not fixed. It's still the byte-identical command at weight 2.0, so one defect is scored twice and a pass gets 2.0 free points. Honest description now, but the redundant weight remains.


5. multiway_branch over-claims its TryCatch check. Description says "TryCatch present inside a branch"; the command is grep -q '"activityType": "TryCatch"' — presence anywhere. skip_and_continue got a real tree-walker (check_trycatch_in_loop.py) for exactly this reason.

Nits: cleanup_fault_deploy.py reads folder_path and never uses it; check_trycatch_in_loop.py would misflag a TryCatch that is the #Body value (it appends with the parent's in_body, not nested) — theoretical, since Studio Web always emits #Body as a container.

tests/tasks/uipath-api-workflow/operate/publish_deploy/publish_deploy.yaml
rares-baesu-uipath: Corrected in `f3fe87170` — and my first attempt at this was wrong twice over.

It was prose with no pointer (you asked for a reference), and it named the wrong cause: it told the reader to provision a `Default Serverless` machine template via `machines create --serverless`. Checking the tenant where this task passes: that template is `Description: "Automatically generated"` with every slot count at 0, and the host the job actually ran on does not appear in `machines list` at all. So neither the template nor `machines list` is the gate.

The gate is tenant runtime licensing — `RU` 300900 / `UNATT` 53 allocated on the working tenant, matching the platform skill's own line: *"A tenant runs unattended jobs against its runtime allocation (UNATT, RU, PLTU)"*. That lines up with the missing RU you hit.

The note now just hands the area to its owning skill rather than encoding a diagnosis:

- runtime license allocation → `uipath-platform/references/licensing/tenant-allocations.md`
- folder / machine / license wiring → `uipath-platform/references/orchestrator/setup-environment.md`
- starting jobs, runtime selection → `uipath-platform/references/orchestrator/run-jobs.md`

All three verified to exist. Left generic deliberately — the symptom is `jobs start` failing or a job stuck `Pending`, and which of the three is missing is `uipath-platform`'s call, not this task's.

liviubobocu (approved):

PR #2655, opened 2026-08-18
fix(uipath-maestro-bpmn): follow portable registry template contracts

DevMomo:
## Problem

Two authoring failures around registry templates that need no live tenant.

`Intsvc.TimerTrigger` is portable — `requiresDiscovery: false`, no binding, context, or input fields — but guidance lumped it in with connector activities, sending agents hunting for a connection and schema enrichment that don't exist for it.

Opaque output types got guessed away. A template output typed `custom` or `Actions.HITL` has no example to copy, so agents coerced it:

```xml
<!-- template says type="Actions.HITL" -->
<uipath:inputOutput id="Var_Task" name="task" type="jsonSchema" />   <!-- guessed -->
```

…or trawled sibling skills and fixtures for a schema the registry deliberately left opaque.

**Why it's in the stack:** it edits the same reference file as #3112, so the order is fixed.

## Fix

- Template output ↔ process variable is one contract: reproduce the **exact `type`**, including opaque and product-specific ones
- Unresolved portable nodes: escaped placeholders, keep the retrieved shape, label non-runnable
- `TimerTrigger` is portable; `EventTrigger` and connector waits still need CLI enrichment
- Every root `startEvent` carries one `uipath:entryPointId` as a direct child of its own `extensionElements`; subprocess starts don't

## Verification

`single_node/timer_start/` pins the pattern and asserts entry-point placement.

### DIFF
--- a/tests/tasks/uipath-maestro-bpmn/_shared/test_bpmn_assertions.py
+++ b/tests/tasks/uipath-maestro-bpmn/_shared/test_bpmn_assertions.py
@@ -112,10 +112,6 @@ class GeneratedProjectScaffoldTests(unittest.TestCase):
         with self.assertRaisesRegex(SystemExit, "current CLI root files map"):
             self.assert_scaffold()
 
-
-if __name__ == "__main__":
-    unittest.main()
-
     def test_rejects_a_second_entry_point_by_default(self) -> None:
         # refresh emits one entry per root manual start event; the default of 1
         # is the single-start subset, not the whole contract.
@@ -166,3 +162,7 @@ if __name__ == "__main__":
             expected_resource_count=0,
             expected_entry_points=2,
         )
+
+
+if __name__ == "__main__":
+    unittest.main()


skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -264,6 +264,13 @@ marks definitions that the skill keeps but does not author for new files.
 | `bpmn:EndEvent` | none, Message, Error, Terminate | Escalation, Compensate, Signal |
 | `bpmn:BoundaryEvent` | Message, Timer, Error | Escalation, Conditional, Signal, Compensate |
 
+Every root-level `bpmn:startEvent` is an entry point, whether it is manual,
+timer-triggered, or connector-triggered. Give it exactly one stable GUID in
+`<uipath:entryPointId value="..." />`. Subprocess start events do not carry an
+entry-point id. When replacing the initializer's manual start with a timer or
+connector start, move or create the entry-point id on the replacement and
+remove the manual start; do not retain a second start event.
```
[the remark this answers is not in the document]
DevMomo: Fixed in 4c249f1c6 — the rule now says the GUID is declared as a direct child of that start event's own <bpmn:extensionElements>, matching the rest of the skill.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/timer_start.yaml
```diff
@@ -57,6 +58,15 @@ success_criteria:
     weight: 1.0
     pass_threshold: 1.0
 
+  - type: command_executed
+    description: "Agent retrieved the Intsvc.TimerTrigger registry template"
+    tool_name: "Bash"
+    command_pattern: >-
+      uip\s+maestro\s+bpmn\s+registry\s+get\s+["']?Intsvc\.TimerTrigger["']?
+    min_count: 1
```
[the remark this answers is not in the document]
DevMomo: Good catch — fixed in 4c249f1c6. The pattern now requires a terminator after the extension type (closing quote, whitespace, or end of string), so `Intsvc.TimerTriggerX` no longer matches. Verified against three cases: bare and quoted `Intsvc.TimerTrigger` still match, `Intsvc.TimerTriggerX` does not.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -341,6 +341,14 @@ marks definitions that the skill keeps but does not author for new files.
 | `bpmn:EndEvent` | none, Message, Error, Terminate | Escalation, Compensate, Signal |
 | `bpmn:BoundaryEvent` | Message, Timer, Error | Escalation, Conditional, Signal, Compensate |
 
+Every root-level `bpmn:startEvent` is an entry point, whether it is manual,
+timer-triggered, or connector-triggered. Give it exactly one stable GUID in
+`<uipath:entryPointId value="..." />`, declared as a direct child of that start
+event's own `<bpmn:extensionElements>`. Subprocess start events do not carry an
+entry-point id. When replacing the initializer's manual start with a timer or
```
rockymadden: **Critical — "remove the manual start" produces a project that cannot be refreshed or packed.**

`deriveBpmnEntryPoints` (`packages/maestro-sdk/src/bpmn-packaging/entry-points.ts`) selects *only* manual start events:

```ts
element.$type === "bpmn:StartEvent" &&
(element.eventDefinitions?.length ?? 0) === 0 &&
!extensionValues(element).some((node) => node.$type === "uipath:event")
```

and then:

```ts
if (entryPoints.length === 0) throw new Error(
  "BPMN file must contain a root manual start event with a uipath:entryPointId.");
```

A timer start carries a `bpmn:timerEventDefinition`, so `eventDefinitions.length > 0` excludes it. A connector start carries `uipath:event`, so the third clause excludes it. Follow this sentence on a timer-started process and you get zero entry points, and `uip maestro bpmn refresh` throws before it writes anything.

That matters specifically because of the PR directly below this one in the stack: #2387 promotes `refresh` to a mandatory step 5 for any package-ready deliverable, and tells the agent a nonzero result is a source failure to fix in the BPMN. Here the BPMN is exactly what the skill instructed. `uip maestro bpmn validate` stays clean throughout, since derivation happens in `refresh`.

`single_node/timer_start/timer_start.yaml` makes this concrete: it requires "There must be NO manual start event." So the skill now ships an eval mandating a shape the packaging path rejects. It passes only because the task never calls `refresh` or `pack`.

Separately, the opening claim — "Every root-level `bpmn:startEvent` is an entry point, whether it is manual, timer-triggered, or connector-triggered" — isn't true of the generated artifact. A timer or connector start never becomes an `entry-points.json` entry. Putting a GUID there is harmless (`validateRootEntryPointIds` only checks GUID-ness when the element is present, it never requires one) but it doesn't make the node an entry point, and stating it this way invites the wrong mental model.

Suggested fix, whichever matches the intended product behavior:

- If a timer-only Maestro process is supposed to be packageable, this is a CLI gap and the doc should say so explicitly — "a timer or connector start is not an entry point; `refresh` currently requires at least one manual root start, so retain one" — rather than instructing removal.
- If removing the manual start is genuinely correct, then this shape is source-only by definition and the sentence should say that packaging is unavailable for it, so step 5 doesn't send the agent into a loop.

Either way, the two halves as written ("every start event is an entry point" + "remove the manual start") can't both hold alongside #2387.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -341,6 +341,14 @@ marks definitions that the skill keeps but does not author for new files.
 | `bpmn:EndEvent` | none, Message, Error, Terminate | Escalation, Compensate, Signal |
 | `bpmn:BoundaryEvent` | Message, Timer, Error | Escalation, Conditional, Signal, Compensate |
 
+Every root-level `bpmn:startEvent` is an entry point, whether it is manual,
+timer-triggered, or connector-triggered. Give it exactly one stable GUID in
+`<uipath:entryPointId value="..." />`, declared as a direct child of that start
```
rockymadden: **Minor — the direct-child rule is right, and it's worth one clause on why.**

The two code paths disagree, deliberately or not:

- `validateRootEntryPointIds` (`project-validator.ts:1075`) uses `findDescendants(startEvent, node => node.name === "uipath:entryPointId")` — it accepts the element nested anywhere under the start event.
- Entry-point derivation uses `extensionValues(startEvent)`, i.e. `extensionElements.values`, direct children only.

So an entryPointId nested inside `uipath:activity` passes `validate` and is silently invisible to `refresh`. That's the actionable half of this rule: "direct child" isn't a style preference, it's the difference between a caught error and a silent one, and `validate` will not tell you.

The new check in `check_timer_start.py` already encodes exactly this (and its comment says why `.//` was avoided) — the reference doc just doesn't pass the reason along to the reader.

rockymadden (approved):
## Approve

Tightest PR in the stack. Every one of the four verifiable claims checks out against source:

|---|---|---|
| `Intsvc.TimerTrigger` is portable — `requiresDiscovery: false`, no binding/context/input fields | bundled `validator/bpmn-spec.json`: `requiresDiscovery: false`, `bindingPattern: "none"`, `inputPattern: "none"`, `contextFields: []` | correct |
| `EventTrigger` / `WaitForEvent` still need enrichment | both `requiresDiscovery: true`, `bindingPattern: "connection"` | correct |
| `requiresDiscovery: true` in JSON renders `RequiresDiscovery: Yes` in the `list` table | `packages/maestro-tool/src/commands/registry.ts:132` — `et.requiresDiscovery ? "Yes" : "No"` | correct |
| `elementId` is additive; a scoped variable still resolves globally | `VariableUtil.ts:126-155` — `inputOutputVariables[id]` is set unconditionally, `activityToVariables[elementId]` only when `elementId` is present | correct |
| entryPointId must be a direct child of the start event's own `extensionElements` | derivation reads `extensionValues` = `extensionElements.values`, direct children only — a nested one is invisible to it | correct |
| subprocess starts carry no entry-point id | `deriveEntryPointsForProcess` reads `process.flowElements` with no recursion | correct |

The `VariableUtil.ts:126-155` citation is exact down to the line range, which made verification cheap. Worth keeping that habit.

One critical item inline — a single sentence in `structural-bpmn.md` that, combined with #2387's new mandatory refresh step, yields a project that cannot be packaged. Cheap to fix, and not a reason to hold the rest.

No AI slop; this is the leanest diff in the stack and every added line changes what the agent does. Two things I'd single out: the test comment explaining that `.//` was deliberately avoided *because* a descendant search would accept the exact placement the task exists to reject — that comment will still be useful in a year. And the `6.0 → 3.0` weight change lands on the on-taxonomy value for primary validation, correcting the same off-taxonomy weight I flagged on #2366.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Confirmed both halves against `entry-points.ts:60-72` — the filter excludes any start with an `eventDefinition` or `uipath:event`, then throws on an empty list. Fixed both: the opening claim now says only a manual root start derives an entry, and an `entryPointId` on a timer or connector start is accepted but inert; and the instruction is now to add the timer alongside the manual start, not to replace it, because `refresh` needs one. `timer_start` keeps its no-manual-start requirement but says in the prompt and the checker that this is a source-only shape with `refresh`/`pack` deliberately out of scope. Whether a scheduled process ought to be packageable at all is a real question — logged as a possible CLI gap rather than papered over here.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Added — the rule now carries the mechanism: `validate` finds the element at any depth, derivation reads direct children only, so a nested id passes one and is invisible to the other.

DevMomo:
Superseded by #3155 — same branch, same content. GitHub auto-closed this when a restack briefly left the head branch with no commits relative to its base; the commits were recovered from reflog and the two review threads here are already addressed in #3155.

PR #2659, opened 2026-08-18
fix(test): outcome-grade admin group-membership e2e instead of counting `members list`

t-hsia:
fix(test): outcome-grade admin group-membership e2e instead of counting `members list`

skill-admin-identity-group-membership failed at 0.89 because its final
criterion required `uip admin groups members list` to appear twice in the
transcript (min_count: 2). The agent produced the correct end state — created
the group, added two users, revoked the second — but listed once, so it scored
0.5 and the task FAILED. That criterion graded command typing, not the
membership outcome.

Replace it with the same outcome-grading pattern already used by the sibling
smoke: a run_command gate that reads the group's membership back from the live
tenant and asserts exactly one member remains (the task's stated goal). Keep a
members-list check as advisory (pass_threshold 0) so self-verification is still
reported without gating on how many times it was typed.

verify_group_member.py is parameterized via GROUP_NAME / EXPECT_MEMBERS env
vars; defaults preserve the smoke's existing behavior (ce-identity-smoke-group,
>=1 member) unchanged.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/tests/tasks/uipath-admin/group_membership_management_e2e.yaml
+++ b/tests/tasks/uipath-admin/group_membership_management_e2e.yaml
@@ -54,13 +54,29 @@ success_criteria:
     weight: 1.5
     pass_threshold: 1.0
 
+  # Outcome gate: read the group's final membership back from the tenant and
+  # require exactly one member — the task's stated goal ("confirm only one
+  # remains") after add-two / revoke-one. Replaces the old `members list` >=2
+  # command-count check, which graded how many times the agent TYPED the verify
+  # command rather than whether the end state was correct.
+  - type: run_command
+    description: "Group ends with exactly one member (first user retained, second revoked), read back from the tenant"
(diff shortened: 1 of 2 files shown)


t-hsia:
## coder-eval validation — all green

The failing criterion was converted from a brittle `command_executed` count (`members list` `min_count: 2` — graded *typing*, scored 0.5 → FAILED) to a `run_command` outcome gate that reads the group's membership back from the live tenant and asserts exactly one member remains (the task's stated goal).

| Round | Claude | codex |
|-------|--------|-------|
| 1 | ✅ score=1.000 (73s) | ✅ score=1.000 (80s) |
| 2 | ✅ score=1.000 (95s) | — |
| 3 | ✅ score=1.000 (62s) | — |

Claude 3/3 consecutive on unchanged HEAD; codex 1/1. The new outcome-gate criterion scored 1.00 on every run. Sibling smoke behavior is unchanged (`verify_group_member.py` defaults preserved).

Runs: [c-r1](https://github.com/UiPath/skills/actions/runs/32159285313) · [x-r1](https://github.com/UiPath/skills/actions/runs/32159290277) · [c-r2](https://github.com/UiPath/skills/actions/runs/32164866040) · [c-r3](https://github.com/UiPath/skills/actions/runs/32168471997)

yadvender-uipath (approved):

PR #2662, opened 2026-08-18
fix(uipath-maestro-bpmn): validate only after BPMN DI

DevMomo:
## Problem

The Critical Rule said only that the diagram is mandatory — "every node needs a `BPMNShape`, every flow a `BPMNEdge`" — which reads as an instruction to write them. So agents hand-emitted coordinates:

```xml
<bpmndi:BPMNShape id="Shape_Task_1" bpmnElement="Task_1">
  <dc:Bounds x="340" y="120" width="120" height="80" />
</bpmndi:BPMNShape>
```

`uip maestro bpmn format <file.bpmn>` generates that whole section. In one e2e authoring run the agent hand-wrote it and never invoked `format` — roughly 14 turns of edit/format/validate churn out of 52.

## Fix

- Critical Rule now states the mechanism: author with no `bpmndi` section, then run `format`. Hand-written DI is the `unknown command` fallback only.
- Complete DI **before** the first `validate`
- Removed the smoke prompt's `bpmndi:*` element listing — itself a nudge to hand-write
- Added an advisory `format` check (`pass_threshold: 0`), not gating: the blocking criteria already grade the resulting geometry

## Verification

Re-run: `skill-bpmn-author-validate` SUCCESS 1.00, with the agent calling `format`. Six negative controls still fail — 0×0 bounds, single waypoint, absent `BPMNDiagram`, `NaN` bounds, missing `dc:Bounds`, unconditioned flow.

Worth knowing: absent `dc:Bounds` makes `validate` fail, while 0×0 bounds pass silently.


skills/uipath-maestro-bpmn/references/cli-conventions.md
```diff
@@ -24,10 +24,19 @@ source (`packages/maestro-tool/src/commands/registry.ts`). Do not invent flags.
...
+## Validation order
+
+For file-based authoring, complete coherent BPMN DI before the first final
+`validate` call — see [Validation](structural-bpmn.md#validation). Then run:
+
+```bash
+uip maestro bpmn validate <file.bpmn> --output json
+```
+
+The `validate` command runs the canvas rules offline (it was added to the CLI in
```
nikhil-maryala: dont' give PR links in the instructions for skills. Just calling out using validate command is enough.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -517,21 +517,28 @@ Re-check after any edit:
 
 ## Validation
 
-Validate with the CLI — it runs the full PO.Frontend canvas rule set offline
-(the same Node/Edge/CanvasState reconstruction and every canvas rule), plus the
-deploy-readiness checks:
+Run final file validation only after coherent diagram interchange exists:
+complete `BPMNDiagram` coverage, a finite-bounds `BPMNShape` with positive
```
nikhil-maryala, line 542: let's not ask it to create diagrams on it's own. use the command for that
```uip maestro bpmn format <file.bpmn>

File: skills/uipath-maestro-bpmn/SKILL.md [hunk 3]
```diff
@@ -172,12 +172,17 @@ For registry-evidence-only tasks, be command-first and time-boxed:
...
-   when the CLI is unavailable. Every root start event needs a
+   when the CLI is unavailable. Do not copy CLI scaffold metadata shapes into a
+   synthetic local project. Every root start event needs a
    `<uipath:entryPointId value="<uuid>" />` child in its `extensionElements` or
    the project generates zero entry points.
-4. **Validate.** Run the CLI validator — it runs the full PO.Frontend canvas
-   rule set (structural rules plus variable, method-call, input-type, and
-   event-object checks) offline, plus deploy-readiness checks:
+4. **Complete layout, then validate.** For file-based authoring, complete
+   coherent BPMN DI before the first final CLI validation; see
```
nikhil-maryala: use CLi command to create Diagram

nikhil-maryala (changes requested):
use CLi instead of skill for generting diagrams

skills/uipath-maestro-bpmn/references/cli-conventions.md
DevMomo: Done in 14a90bae8 — dropped the UiPath/cli#3135 reference. The paragraph now just says validate runs the canvas rules offline, and to update the CLI if it reports an unknown command.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Done in 14a90bae8 — the Validation section now opens with 'Generate the diagram with uip maestro bpmn format <file.bpmn>, then validate. Do not hand-author the diagram when the command is available.' The DI coverage details stay as the definition of what complete looks like.

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Done in 14a90bae8 — step 4 now names the command directly: 'Generate the diagram with uip maestro bpmn format <file.bpmn> — do not hand-author it when that command is available.'

DevMomo:
@nikhil-maryala all three points are addressed in `14a90bae8` — re-requesting review.

- **Use the CLI for diagrams, don't ask the skill to draw them.** `SKILL.md` step 4 and the `structural-bpmn.md` Validation section now both lead with `uip maestro bpmn format <file.bpmn>` and say explicitly not to hand-author the diagram when that command is available. The DI coverage details stay only as the definition of what "complete" means, which is what the validate-ordering rule keys off.
- **No PR links in skill instructions.** Dropped the `UiPath/cli#3135` reference from `cli-conventions.md`; it now just states that `validate` runs the canvas rules offline and to update the CLI if it reports an unknown command.

Worth flagging while you're here: your point applies beyond this PR. #2867 (DI shape dimensions, further up the stack) currently claims `uip maestro bpmn format` emits shapes for `dataObjectReference`, `dataStoreReference` and `textAnnotation` — I tested it and `format` doesn't lay those out at all, and deletes pre-existing shapes for them. I'll correct that PR in the same spirit rather than leave the skill promising something the command doesn't do.

skills/uipath-maestro-bpmn/references/cli-conventions.md
DevMomo: Follow-up on this file in `04f2582d5` (branch also rebased onto `main`, now 0 behind): the PR reference stays gone, and the `Validation order` section now leads with the command instead of only talking about DI — "generate the diagram with `uip maestro bpmn format <file.bpmn>` — do not hand-author it", then complete DI, then `validate`. That closes the gap left by the first pass: all three files (`SKILL.md` step 4, `structural-bpmn.md` § Validation, and this one) now say the same thing.

Validation: `npm run skills:validate` → OK (default 27 skills / 1761 files; studioweb 27 / 1761 / 182 replacements). `check-cli-verbs.py` → 0 High, 0 Medium. All relative links resolve.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Confirming the final state of this after a rebase onto `main` (0 behind) and one more commit, `04f2582d5`.

This section still opens with "Generate the diagram with `uip maestro bpmn format <file.bpmn>`, then validate. Do not hand-author the diagram when the command is available." The DI coverage paragraph that follows is deliberately kept, but only as the *definition* of what "complete" means — it is what the validate-ordering rule keys off, and it carries the caveat that `validate` still reports `Valid` for degenerate geometry (0x0 bounds, single-waypoint edges), so a clean validate is not proof the canvas renders. It is not instructions to draw the diagram by hand.

I also carried your point into the smoke task, since docs alone would not have caught the failure mode:
- The prompt no longer spells out `bpmndi:BPMNDiagram` / `BPMNShape` / `BPMNEdge` — that was itself a nudge to hand-write them. It now states the outcome ("a shape for every node and an edge for every sequence flow") and defers the how to the skill.
- Added an advisory `command_executed` check for `maestro bpmn format` at `pass_threshold: 0`, so a hand-authored diagram shows up in the run record without docking an agent that reaches complete DI another way (per `.claude/rules/test-writing.md` on outcome-invisible conventions).

Worth recording why this matters: an independent e2e authoring run in this repo reproduced exactly the behavior you flagged — the agent hand-wrote the entire `bpmndi:BPMNDiagram` and never invoked `format`, costing roughly 14 turns of edit/format/validate churn.

Validation: `npm run skills:validate` → OK. Task YAML parses (6 criteria). `check_author_validate.py` compiles. `check-cli-verbs.py` → 0 High, 0 Medium; `maestro bpmn format` is present in the catalog snapshot (uip 1.202.0-dev.8552).

skills/uipath-maestro-bpmn/SKILL.md
DevMomo: Confirming the final state after a rebase onto `main` (0 behind) and `04f2582d5`.

Step 4 is now "**Complete layout, then validate.** Generate the diagram with `uip maestro bpmn format <file.bpmn>` — do not hand-author it when that command is available." It then links to `references/structural-bpmn.md#validation` for the required node/flow coverage rather than restating the DI geometry spec inline, so the authoritative copy lives in one place and cannot drift.

In this last commit I also made `references/cli-conventions.md` name the same command, so all three files agree instead of two of them naming `format` and the third only describing DI.

Validation: `npm run skills:validate` → OK, `hooks/validate-skill-descriptions.sh` → exit 0, all relative links in the skill resolve.

DevMomo:
@nikhil-maryala all three threads are addressed and resolved. Rebased onto `main` (was 44 behind, now 0, no conflicts) and pushed one more commit, `04f2582d5`, closing the last gap in your "use the CLI, not the skill, to generate diagrams" point.

**What changed since your review**

| Your comment | Where it landed |
|---|---|
| "use CLI command to create Diagram" (`SKILL.md`) | Step 4 is now **Complete layout, then validate** — leads with `uip maestro bpmn format <file.bpmn>`, "do not hand-author it when that command is available", then links to `structural-bpmn.md#validation` instead of restating the DI spec inline |
| "let's not ask it to create diagrams on it's own. use the command for that" (`structural-bpmn.md`) | § Validation opens with "Generate the diagram with `uip maestro bpmn format <file.bpmn>`, then validate. Do not hand-author the diagram when the command is available." DI coverage stays only as the definition of "complete" — it is what the ordering rule keys off |
| "don't give PR links in the instructions for skills" (`cli-conventions.md`) | `UiPath/cli#3135` dropped. The paragraph now just says `validate` runs the canvas rules offline, and to update the CLI if it reports an unknown command |

`04f2582d5` on top of that:

- **`cli-conventions.md` now names `format` too.** It previously only described DI ordering, so two of three files named the command and the third did not. All three now agree.
- **Smoke task prompt no longer enumerates the DI element names.** Listing `bpmndi:BPMNDiagram` / `BPMNShape` / `BPMNEdge` in the prompt was itself a nudge to hand-write them. It now states the outcome and defers the how to the skill.
- **Advisory `format` check added** at `pass_threshold: 0`, so a hand-authored diagram is visible in the run record without docking an agent that reaches complete DI another way — the pattern `.claude/rules/test-writing.md` prescribes for outcome-invisible conventions.

**Why the grading is on the DI outcome rather than on the `format` invocation** — flagging this since it is the one place I did not do the literal thing. The blocking checks grade the resulting geometry (finite bounds, positive width/height, at least two waypoints per edge), and `format` is graded advisorily. Per the repo's test-writing rules, a gating `command_executed` check on a command whose effect is already observable in the artifact docks agents that reach the same correct result differently. The advisory check still surfaces non-compliance in the report. Happy to make it blocking if you would rather the test hard-fail a hand-authored diagram.

**Corroboration for the premise:** an independent e2e authoring run in this repo reproduced exactly what you flagged — the agent hand-wrote the entire `bpmndi:BPMNDiagram`, never invoked `format`, and burned roughly 14 turns on edit/format/validate churn.

**Validation:** `npm run skills:validate` → `OK - default: 27 skills, 1761 files, 0 replacements; studioweb: 27 skills, 1761 files, 182 replacements.` · task YAML parses (6 criteria) · `check_author_validate.py` compiles · `check-cli-verbs.py` → 0 High, 0 Medium (`maestro bpmn format` confirmed present in the catalog snapshot) · `check-skill-status.py` and `check-skills-sh.py` → OK · `validate-skill-descriptions.sh` → exit 0 · all relative links resolve.

DevMomo:
Correcting my previous comment: the smoke suite has since finished and **it caught a real bug in this PR's own checker**, which my local validation missed. Fixed in `ff0434da0`.

**What broke**

`skill-bpmn-author-validate` failed, taking the suite to 92.3% (12/13) and under the 95% gate:

```
File "check_author_validate.py", line 132, in main
NameError: name 'path' is not defined. Did you mean: '_path'?
```

An earlier commit in this stack wrapped `parse_bpmn("InvoiceApproval")` in a `load_bpmn()` helper and rebound the result to `_path`, but the closing `print(f"OK: {path} ...")` still referenced `path`. The crash landed *after* every assertion had already passed, so the 4.0-weight criterion scored 0.00 and the task dropped to 0.619.

Worth noting what the run confirms about the actual subject of this PR: the advisory `maestro bpmn format` check **scored 1.00** — the agent did call `format` rather than hand-authoring the diagram — and the final `uip maestro bpmn validate` exited 0. So the guidance change works; only the grader crashed.

**The fix**

`load_bpmn()` was pure indirection with a single caller, so I removed it rather than patching the name, restoring the direct call and the `path` binding.

**How I verified it this time**

My earlier check was `python3 -m py_compile`, which only validates syntax and cannot see an unresolved name in a branch it never executes — that is exactly how this slipped through. I now actually *run* the checker against fixtures:

- Positive control (complete DI): exits 0 and reaches the previously-crashing final line.
- Six negative controls, all still caught: 0x0 bounds, single-waypoint edge, absent `BPMNDiagram`, non-finite `NaN` bounds, shape with no `dc:Bounds`, and a non-default flow missing its condition.

So the DI-geometry logic was sound all along; the `NameError` was the only defect. `npm run skills:validate` still prints OK and the branch is still 0 behind `main`.

The review threads remain addressed and resolved — this commit is a fix to my own regression, not a change in the reviewed guidance.

DevMomo:
Smoke re-run finished. **The fix in `ff0434da0` is confirmed:**

```
skill-bpmn-author-validate    status=SUCCESS   score=1.00
```

That is the task that was crashing with the `NameError` — now full score, including the 4.0-weight DI-geometry checker that previously scored 0.00.

**The suite gate is still red, but on a different task that this PR does not touch:**

```
skill-bpmn-operate-diagnose-minimal-fault-triage  status=FAILURE  score=0.84
  file_contains  FAILED  3/4 includes found (0.75)
  llm_judge      FAILED  0.58 < 0.70
Pass rate: 92.3% (12/13)
```

Evidence that this is not caused by this PR:

1. **The delta between the two runs is a single file** — comparing `04f2582d5` to `ff0434da0` shows only `tests/tasks/uipath-maestro-bpmn/smoke/check_author_validate.py`, which is used exclusively by the author-validate task. Both runs carried identical `SKILL.md` / `cli-conventions.md` / `structural-bpmn.md` changes.
2. **This task passed in the previous run** (`04f2582d5`, where author-validate was the only failure) and failed in this one — with no change between them that could reach it. `llm_judge 0.58` and `file_contains 3/4` are quality/nondeterminism failures, not a crash.
3. **It is already being fixed elsewhere.** #2743 — "test(maestro-bpmn): tolerate markdown line wrapping in fault-triage ownership tokens" — is open and unmerged, and targets exactly this `file_contains` brittleness on exactly this task.

So the 95% gate will stay red on this branch until #2743 lands, independent of anything here. I have deliberately not touched that task: it is outside this PR's scope and its fix already has an owner.

Everything in this PR's own scope is green: 25 other checks pass, 0 fail, `npm run skills:validate` prints OK, and the branch is 0 behind `main`. All three review threads remain addressed and resolved.

PR #2670, opened 2026-08-18
test(maestro-flow): scope file-attachment output match to declared out vars

rockymadden:
## Problem

`skill-flow-file-attachment-debug` fails at `weighted_score: 0.500`:

```
FAIL: No node matches type hint 'core.action.script'.
Node types seen: ['core.control.end', 'core.trigger.manual']
```

Chasing that surfaced **two** bugs in `check_file_attachment_flow.py`, the first masking the second.

### 1. An off-target node-shape gate

The checker opened with `assert_flow_has_node_type(["core.action.script"])` — but nothing asks the agent for a Script node. `a62c33fd8` ("relax file-attachment prompt to goal-level") deleted the prompt bullet that required one:

```
- A Script node reads the attachment from the trigger output and returns the
  file name, e.g. `return { fileName: $vars.<triggerId>.output.inputDoc.FullName };`
```

Its message says *"Checker is unaffected (discovers the file-typed var dynamically, matches the random filename anywhere in output)"* — accounting for the two dynamic assertions but missing this static gate above them. The task passed on luck for three months until an agent took the shorter route.

The rejected flow is correct per the skill. Its End node carries:

```json
"outputs": { "fileName": { "source": "=js:$vars.start.output.inputDoc.FullName" } }
```

verbatim the pattern `references/shared/node-output-wiring.md` prescribes — **End nodes (`core.control.end`)** → `outputs.<varId>.source`, `=js:` **YES** — with the `.FullName` accessor from `variables-and-expressions.md`. That run reported `finalStatus: "Completed"` with `fileName = "sample-report.txt"`. `flow validate` passed and the `--attachment` criterion matched; only this gate failed.

### 2. A vacuous assertion the gate was hiding

`assert_outputs_contain` flattens **every** global — and the `in` file variable's runtime value *is* the attachment object carrying `FullName`. Matching the random basename against that set is a tautology: it passes whenever a file uploads at all, regardless of what the flow computes.

Confirmed against a live tenant. A flow whose End node maps `fileName` to the literal `"sample-report.txt"` **still passed**, because the runtime globals are:

```json
{"start.output.inputDoc": {"ID": "...", "FullName": "evidence-<rand>.txt",
                           "MimeType": "application/octet-stream", "Metadata": {"size": "8"}},
 "fileName": "sample-report.txt"}
```

So deleting the gate alone would have left a criterion that **cannot fail**. `flow_check.py` already ships the right helper — `assert_named_output_contains`, whose docstring calls out precisely this trigger-input-echo trap ("an `invoiceNumber` input global makes the invoice string 'present' even when the agent never drafted it"). `file_attachment` never migrated to it.

## Changes

- Drop the node-shape gate and its now-unused import.
- Scope the basename match to the flow's declared `direction:"out"` globals, discovered from source so the check stays goal-level rather than pinning the name `fileName`. Shape-agnostic, but no longer vacuous.
- Module docstring records the tautology with the observed payload, so the scoping doesn't get "simplified" back to a whole-payload match.
- `file_attachment.yaml` `description:` — name both sanctioned shapes rather than mandating a Script node. `initial_prompt` unchanged; it was already correct.

Siblings calling `assert_flow_has_node_type` (`decision`, `switch`, `rpa`, `api_workflow`, `subflow`, `terminate`, `coded_agent`, `lowcode_agent`) each pin a node their prompt makes inevitable, and are untouched. No skill-doc change: the skill steered the agent to a valid, working flow.

## Validation

Reproduced the rejected flow in a fresh org (`solution init` + `flow init` + the same `trigger -> End` binding) and ran the patched checker end to end against a live tenant:

| case | expected | result |
| --- | --- | --- |
| End node reads `.FullName` via `=js:` (the flow that was rejected) | pass | **exit 0** — `output variable surfaced attachment FullName 'evidence-0d80c20d9332.txt'` |
| End node maps `fileName` to a hardcoded literal | fail | **exit 1** — `Output 'fileName' missing all of ['evidence-38c9a6dea640.txt']; fileName=["sample-report.txt"]` |

The second row is the case the pre-patch checker passed. `flow validate` → `Status: "Valid"`; `py_compile` and `yaml.safe_load` clean with 3 criteria intact. Test solution deleted afterward via `cleanup_solutions.py`.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/single_node/file_attachment/check_file_attachment_flow.py
+++ b/tests/tasks/uipath-maestro-flow/single_node/file_attachment/check_file_attachment_flow.py
@@ -4,13 +4,37 @@
 uploaded it and the flow surfaced the file's name as output.
 
 The attachment filename is generated here at check time with a random token the
-agent never saw, so the only way it can appear in the flow output is if the
-attachment was actually uploaded, resolved to a Flow Attachment object, read by
-the Script node, and mapped to an output variable — i.e. the full
-``--attachment`` path works end to end. An agent that hardcodes a filename
-literal cannot pass.
+agent never saw, so it can only reach a declared OUTPUT variable if the
+attachment was actually uploaded, resolved to a Flow Attachment object, read via
+``.FullName``, and mapped out — i.e. the full ``--attachment`` path works end to
+end. An agent that hardcodes a filename literal cannot pass.
+
+Node-shape agnostic on purpose. The feature under test is the ``--attachment``
+binding and ``file``-variable hydration, not any particular node: the skill
+documents reading a hydrated ``file`` variable BOTH in a Script node body and
+directly in an End node's ``outputs.<varId>.source`` via ``=js:`` (see
+``references/shared/node-output-wiring.md`` — End nodes). Both surface
+``.FullName``, so pinning ``core.action.script`` rejected a correct two-node
+``trigger -> End`` flow that the skill itself steers agents toward.
+
+Scoping the match to the flow's ``out`` globals is what keeps the check honest,
+and it is NOT interchangeable with :func:`assert_outputs_contain`. That helper
+flattens every global — including the ``in`` file variable, whose runtime value
+IS the attachment object carrying ``FullName``. Matching the basename against
+that set is a tautology: it passes whenever a file uploads at all, even for a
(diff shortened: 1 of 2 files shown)


jiyangzh (approved):
Assume locally run passed

PR #2676, opened 2026-08-18
fix(uipath-maestro-case): prune solution resources orphaned by a binding repoint or task deletion

song-zhao-25:
## Problem

`uip solution resources refresh` is **additive-only** — it never removes a resource whose binding was repointed away or deleted. Every repoint/removal procedure in the skill ended at *"regenerate `bindings_v2.json` + `resources refresh`"*, so the superseded resource survived in the solution registry.

Studio Web's `prepareForCustomDebug` then fails provisioning with `Sequence contains no matching element` (`errorCode 2106`) while `uip maestro case validate` stays `Valid` — validate never reads the solution registry. Hit twice in one build (api-workflow rename; shared connection split in two); `uip solution resources remove <key>` cleared it both times.

Reproduced independently in a fixture: delete a bound task, empty `caseplan.bindings` and `bindings_v2.resources`, run `refresh` → it reports `Synced 0 resources`, `Warnings: []`, and both orphan declarations remain on disk.

## Fix

New **§ Prune orphaned solution resources** in `bindings-v2-sync.md`, with pointers from every call site that previously ended at `refresh`: both repoint procedures, § Re-sync a task, § Rename or delete a global variable, § Replace a trigger with a different type, and `brownfield.md` after-edits rule 2.

Matching rules, all verified against live CLI output:

- **`Key` doesn't match across the artefacts.** `resources list` keys processes/apps on the Orchestrator entity UUID, `bindings_v2.json` on `<folderPath>.<name>` — match those by `Name`. Only connections match by `Key`. The removal itself takes the registry GUID, never the `bindings_v2.json` key.
- **Never prune the case project's own `process` + `package` pair** — `solution projects add` creates it, so it is absent from `bindings_v2.json` by design. A "not in `bindings_v2` → orphan" rule deletes the case project itself.
- **Never target a `package` directly** — removing the owning process/app cascades to that resource's own package and leaves siblings untouched.

Two adjacent defects found while wiring this up:

- `§ Delete a task`'s binding cascade covered **connector tasks only** and never routed to `§ Cleanup`, so deleting a process / agent / rpa / action / api-workflow / case-management task left both its binding pair and its solution resource behind.
- `§ Replace a trigger with a different type` handled bindings not at all, though manual/timer ↔ event adds or drops a Connection binding.
- `registry-discovery.md` told the agent to hand-delete the `package/` and `process/` declaration files — contradicting the new procedure and the test below. Replaced with `resources remove`.

## Evidence the guidance is load-bearing

New task `skill-case-edit-bound-task-add-remove` (local-only, no debug). Its prompt asks only for a task removal and a task addition — it never mentions the resource inventory, so the cleanup has to come from the skill.

Same test, same agent and model, one variable — the docs:

| | With this PR | Control: test on `main` docs |
|---|---|---|
| `weighted_score` | **1.0** | **0.4** |
| `case validate` | pass | pass |
| orphan pruned | pass | **FAIL** — both declarations left behind |
| ran `resources remove` | pass | **FAIL** — `Matched 0/1` |

[with fix](https://github.com/UiPath/skills/actions/runs/32206444174) · [control](https://github.com/UiPath/skills/actions/runs/32207313174)

Without the guidance the agent makes the edit, validates green, and stops — leaving exactly the orphan that breaks provisioning.

## Regression

[`cm_golden_expense`](https://github.com/UiPath/skills/actions/runs/32204790461) — 7/7, `weighted_score: 1.0`. The bindings_v2 **add** path (8-stage golden rebuild with connector bindings) is unaffected.

## Review response

An independent skill review scored the skill 3.1/5 and found four defects here, since fixed: the removal step named no key source; step 4 hard-coded `solution upload --force` (destroys Studio Web version history) with no consent gate; `§ Replace a trigger` was a missed call site; and `solution resources list` was absent from the sanctioned command inventory.

Per that review the prune pointer was also moved **out** of Critical Rule 14 — that rule gates the greenfield Phase-3 exit, where nothing has been repointed — into a Reference Navigation row. Rule 14 was the 4th-longest of 25 Critical Rules carrying six unrelated obligations.

Also appends the missing `<!-- END: brownfield.md -->` marker; Rule 24 makes observing it a hard stop before authoring from a reference, and `brownfield.md` was the only one lacking it.


tests/tasks/uipath-maestro-case/edit/bound_task_add_remove/fixtures/LinearThreeStages/resources/solution_folder/package/NameToAgeFixed.api.NameToAgeFixed.json
```diff
@@ -0,0 +1,32 @@
+{
```
charlesliu9, line 1: Is it a static test or may run debug? Static test is fine, but if you need to run it, avoid this API workflow as it has a high possibility of getting 429 error due to rate limit.

song-zhao-25: Static — no `debug`, no upload, no tenant. Criteria are `case validate` on the local file, a filesystem check, and transcript inspection.

Deliberately keeping it that way: the test's job is to prove the agent performs the documented cleanup (1.0 with the guidance vs 0.4 without on the same test), and a runnable version would instead be asserting Studio Web's `2106` provisioning behavior — platform, not skill. If the platform ever tolerates orphans, a runnable test would go green while the guidance silently rots.

Noted on the 429 risk if anyone makes it runnable later. FWIW this task *removes* the api-workflow task, so `NameToAgeFixed` is only ever a binding target, never invoked.

charlesliu9 (approved):

PR #2678, opened 2026-08-19
test(maestro-flow): stop input echoes from satisfying assert_outputs_contain

rockymadden:
Follow-up to #2670, which fixed one symptom of this. Independent branch off `main` — disjoint files, so it can land in either order.

## Problem

`variables.globals` carries **every** global — `in` as well as `out` — and `_leaves` flattens nested objects. So `assert_outputs_contain(payload, X)` never meant *"the flow output X"*. It meant *"X appears somewhere among all global values, inputs included."* Any needle that is also an input is matched by the input's own echo, and the assertion becomes a tautology.

Confirmed against a live tenant while fixing #2670. Binding `--attachment inputDoc=<random>.txt` puts the whole attachment object in globals, so searching for that random name **passed a flow whose End node mapped its output to a hardcoded literal**:

```json
{"start.output.inputDoc": {"ID": "...", "FullName": "evidence-<rand>.txt",
                           "MimeType": "application/octet-stream", "Metadata": {"size": "8"}},
 "fileName": "sample-report.txt"}
```

It's latent in **`multi_node/billing_dispute_resolution`** too — the heaviest flow task (840s). `INVOICE` lives inside the `WEBHOOK_PAYLOAD` the checker feeds in as the flow's object input, and 6 of the 11 follow-up verdict tokens (`credit`, `approv`, `duplicate`, `contract`, `290`, `300`) are in that payload as well. Both content assertions could pass without the flow reasoning about the dispute at all.

And in **`multi_node/multi_city_weather`**, which asserts `["Seattle", "Phoenix", "New York"]` to prove the loop iterated 3 times. If an agent declares those cities as an `in` array global — a normal reading of *"Loop over Seattle, Phoenix, and New York"* — the declaration alone satisfied the match, no loop required.

`flow_check` already ships the right per-variable helper — `assert_named_output_contains`, whose docstring names this exact trap ("matches trigger-input echoes… even when the agent never drafted it"). But relying on each author to remember it leaves the default matcher quietly wrong, so the guard belongs in the matcher.

## Fix

`assert_outputs_contain` subtracts the flow's inputs before matching, identifying them by three signals (`_declared_input_global_keys`), all active by default:

1. **Key shape** — a trigger-scoped input is keyed `<triggerId>.output.<varId>` at runtime; outputs are keyed by bare id.
2. **What the checker bound** — `run_debug` stashes the ids it passed via `--inputs` / `--attachment`. Exact, and covers plain inputs that signal 1 can't see.
3. **Source-declared `direction:"in"` ids** — catches an `in` global with a `defaultValue`, which echoes even though nothing bound it: invisible to signal 1 (bare id) and to signal 2 (never bound). Reads the project `run_debug` already resolved for its command line, so no call site has to opt in; an explicit `project_dir` overrides for a caller holding a payload from elsewhere.

**None of these globs the filesystem from CWD.** A first cut globbed `**/*.flow`; at a repo root that matches 428 flows through a `plugins/` symlink loop and reads whichever sorts first. It also tripled suite runtime — 20.55s to 2.90s once removed.

Failure reporting matters as much as the guard: when a needle is absent from the outputs but *would* have matched an input, the error says `INPUT ECHO`, names the culprit globals, and points at `assert_named_output_contains` — instead of a bare "missing". That case means the check was passing vacuously, so it needs a real output assertion, not a shrug. `allow_input_echo=True` restores the old behavior for a check that legitimately grades a round-trip. `collect_outputs` is unchanged, so callers using it for "did the run produce anything at all" stay permissive.

## No task migration needed

I simulated all 19 `assert_outputs_contain` call-site shapes against the guard rather than assuming:

| call site | result |
| --- | --- |
| `decision` (in `90` → needle `"warm"`) | passes |
| `switch` (in `2` → needle `"summer"`) | passes |
| `subflow` (in `"hello"` → needle `"olleh"`) | passes |
| no-input tasks (weather / slack / reading_list / jira / rpa / all 5 `edit/*`) | passes — nothing to echo |
| `multi_city_weather`, loop **emits** the cities | passes |
| `multi_city_weather`, loop emits **nothing** | **fails** `INPUT ECHO` (passed vacuously before) |
| `billing_dispute_resolution`, resolution **cites** the invoice | passes |
| `billing_dispute_resolution`, resolution **never** cites it | **fails** `INPUT ECHO` (ditto, incl. the verdict OR-set) |

The first three grade transformations absent from their inputs, so subtracting inputs cannot touch them. No checker file needs editing: the rows that flip to failing are flows that genuinely didn't produce the value, which is the grading these assertions were always meant to do.

Caveat: that's verified by unit tests and simulation, not by executing the e2e suite (needs a live tenant; `billing_dispute_resolution` alone is 840s). If a real run of either task fails after this, read it as a finding about the flow rather than a regression here.

## Tests

`197 passed` in `_shared/test_flow_check.py`, plus the `tests/scripts/test_runtime_payload_key_casing.py` guard. CI already runs this via `test-helpers.yml` → `pytest tests/tasks/uipath-maestro-flow/ -v`, so it's enforced on every PR — unlike the e2e criteria.

New cases cover: the live trigger-scoped attachment shape, a bound plain input (the billing shape), a `defaultValue` input via both an explicit `project_dir` and the auto-resolved one, a real output that *also* appears in an input (must still pass — no false failures), element outputs never being subtracted, `allow_input_echo` restoring old behavior, `collect_outputs` staying permissive, and an ordinary miss *not* emitting the `INPUT ECHO` hint so it can't misdirect.

That `allow_input_echo` test is the self-documenting proof of the delta: same payload, passes under the old behavior, raises under the new.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
@@ -61,6 +61,12 @@ _LAST_DEBUG_RAW: str | None = None
 # :func:`assert_outputs_contain`.
 _LAST_DEBUG_INPUT_IDS: set[str] = set()
 
+# Project dir :func:`run_debug` resolved for the most recent run. Stashed so the
+# source-declared `direction:"in"` signal works automatically instead of needing
+# every call site to opt in — and so the lookup is scoped to the project that
+# actually ran, never a bare glob from an arbitrary CWD.
+_LAST_DEBUG_PROJECT_DIR: str | None = None
+
 
 # A `uip maestro flow debug` run can die on a transient server-side error — a
 # gateway timeout / 5xx while polling the debug instance, which the CLI reports
@@ -121,11 +127,13 @@ def run_debug(
         cmd.extend(["--inputs", json.dumps(inputs)])
     for var_id, local_path in (attachments or {}).items():
         cmd.extend(["--attachment", f"{var_id}={local_path}"])
-    global _LAST_DEBUG_RAW, _LAST_DEBUG_INPUT_IDS
-    # Record what we bound as input so output assertions can discount echoes.
+    global _LAST_DEBUG_RAW, _LAST_DEBUG_INPUT_IDS, _LAST_DEBUG_PROJECT_DIR
+    # Record what we bound as input, and where the project lives, so output
+    # assertions can discount echoes of our own inputs.
     _LAST_DEBUG_INPUT_IDS = {str(k) for k in (inputs or {})} | {
         str(k) for k in (attachments or {})
     }
+    _LAST_DEBUG_PROJECT_DIR = project_dir
     for attempt in range(retries):
         r = subprocess.run(cmd, capture_output=True, text=True, timeout=timeout)
         _LAST_DEBUG_RAW = r.stdout
@@ -456,12 +464,13 @@ def _declared_input_global_keys(
        ids it passed via ``--inputs`` / ``--attachment``. Those are inputs by
        construction, and this is exact — no parsing, no guessing. It covers the
        plain (non-trigger) inputs that signal 1 cannot see.
-    3. **Source declaration** (only when ``project_dir`` is given). Reads that
-       project's first ``.flow`` and collects ``variables.globals[].id`` where
-       ``direction == "in"``. Catches inputs the checker never passed — an ``in``
-       global with a ``defaultValue`` still gets a runtime value and still
-       echoes. Opt-in because it needs a known project; parse failures degrade
-       to signals 1-2 rather than erroring.
+    3. **Source declaration.** Reads the project's first ``.flow`` and collects
+       ``variables.globals[].id`` where ``direction == "in"``. Catches inputs the
+       checker never passed — an ``in`` global with a ``defaultValue`` still gets
+       a runtime value and still echoes. The project defaults to the one
+       :func:`run_debug` resolved, so this needs no per-call-site opt-in;
+       ``project_dir`` overrides it for a caller holding a payload from
+       elsewhere. Parse failures degrade to signals 1-2 rather than erroring.
     """
     variables = _get_ci(payload, "variables", "Variables") or {}
     keys = list((_get_ci(variables, "globals", "Globals") or {}).keys())
@@ -476,6 +485,7 @@ def _declared_input_global_keys(
     # Signals 2 and 3 — ids known to be inputs, matched to keys by exact name or
     # by the leaf of a `<trigger>.output.<id>` key.
     declared = {str(i).lower() for i in _LAST_DEBUG_INPUT_IDS}
+    project_dir = project_dir or _LAST_DEBUG_PROJECT_DIR
     if project_dir:
         try:
             flows = sorted(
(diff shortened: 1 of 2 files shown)


tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -418,24 +524,72 @@ def _leaves(v: Any):
...
+                f"\nINPUT ECHO: {echo_only} appear ONLY in the flow's input globals "
+                f"{sorted(input_keys)}, not in its outputs. Before this guard existed "
+                f"the match was satisfied by the input itself, so this check passed "
+                f"vacuously. Grade the real output instead — "
+                f"assert_named_output_contains(payload, '<outVarId>', ...)."
+            )
         _fail_with_capture(
             f"Outputs missing {mode} {list(needles)}; present={present}; "
-            f"missing={missing}\nOutputs: {haystack[:1000]}"
+            f"missing={missing}\nOutputs (inputs excluded): {haystack[:1000]}{detail}"
```
rockymadden: Is this really needed? Outputs were not assumed to include inputs

nikhil-maryala (approved):

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Agreed — removed in 96de6f353. Nobody held that assumption, so annotating its absence explained an implementation detail instead of the failure: on the ordinary miss (the common case) it was pure noise, and on the vacuous-match case the `INPUT ECHO` block already spells the exclusion out. Plain `Outputs:` now.

rockymadden:
Updated: rebased onto current `main` (now containing #2670), addressed the inline note, and closed the two verification gaps the automated review flagged.

### Since the bot review

That review predates two commits, so a few of its details are stale: `project_dir` is no longer opt-in (signal 3 now reads the project `run_debug` already resolved, so it applies with no call-site change), and the count is `197 passed`, not 193.

- **"Unable to run pytest"** — CI confirms it. `maestro-flow checker unit tests` passes on this branch (27s), along with every other check.
- **Regex assumes trigger ids contain no dots** — leaving as-is. A deeper key just falls out of signal 1, which is the safe direction (less coverage, never a false subtraction), and signals 2-3 still catch it.

### Live verification

`run_debug` is modified by this PR, so a bug there would break every e2e checker — that path only gets exercised for real. Against a live tenant:

- The **merged #2670 checker** runs green through the modified `run_debug` — exit 0, `output variable surfaced attachment FullName 'evidence-5146571d4427.txt'`.
- The **guard fires on a real payload**, not just a synthetic one. Same live payload, a flow whose End node maps `fileName` to a hardcoded literal, needle `evidence-abed699d04d4.txt`:

```
OLD (allow_input_echo=True): PASSED  <- the tautology, on a live payload
GUARD: FAILED as intended -> INPUT ECHO: ['evidence-abed699d04d4.txt'] appear ONLY in
       the flow's input globals ['start.output.inputDoc'], not in its outputs.
```

from these runtime globals:

```json
"globals": {"start.output.inputDoc": {"ID": "...", "FullName": "evidence-abed699d04d4.txt",
                                      "MimeType": "application/octet-stream", "Metadata": {"size": "8"}},
            "fileName": "sample-report.txt"}
```

Test solutions deleted afterward via `cleanup_solutions.py`.

### One thing that run surfaced

The payload does carry a `variables.globalDefinitions` map, which raises the fair question of why the guard reads the `.flow` source at all:

```json
"globalDefinitions": {"start.output.inputDoc": {"name": "inputDoc", "type": "file"},
                      "fileName": {"name": "fileName", "type": "string"}}
```

It holds only `name` and `type` — never direction — so it cannot separate an input from an output. Documented on `_declared_input_global_keys` with the observed shape, so the next reader doesn't have to spend a tenant run finding that out.

The e2e suite itself still hasn't been run against this; `billing_dispute_resolution` and `multi_city_weather` remain worth watching on their next runs, per the simulation table in the description.

jiyangzh (approved):

PR #2680, opened 2026-08-19
fix(planner): state the non-adhoc selector rule for stage exit

abhiram-vad:
## Problem

`uipath-planner` documents the `selected-tasks-completed` non-adhoc restriction **only in task-entry framing**:

- `references/case/model.md:152` — bullet under §Sequencing & activation (task-entry mode table)
- `references/case/render-stages-tasks.md:78` — inside the table headed *"Which WHEN to write (task entry)"*

Nothing on the **stage-exit** path says it: the Lifecycle gates table lists `selected-tasks-completed` for Stage exit with no caveat, the stage condition tables defer to that table, the §Logical integrity walk had no selector check, and the SDD template's stage-exit row is bare vocabulary. A designer authoring a stage exit never encounters the rule, so a stage exit can be gated on an optional user-launched task. That gate never fires — the Case App omits `adhoc` tasks from the selected-task picker — and `uip maestro case validate` does not enforce the restriction, so the design reaches build looking clean.

Second gap: `uipath-maestro-case`'s handoff contract expects this lane to run a **"pre-approval adhoc-dependency audit"** that blocks the Case Review on an adhoc selector. No such step existed in the lane guide. The closest text was buildability must #7 (required human work → `action`, optional → `adhoc`), which never mentions selectors or gates.

## Change

| File | Change |
|---|---|
| `references/case/model.md` | Canonical, gate-agnostic **note 6** under §Lifecycle gates — the restriction holds at every gate that accepts the rule, plus the "validate does not enforce it" fact. Stage exit and Task entry rows now point at note 6. |
| `references/case/review.md` | New §Logical integrity **check 9** (selector integrity at both gates, with the repair: remodel the human work as an `action` task, never delete the guard). Also corrects Finalization check 12's stale count — the walk already had 8 checks while the text said "all seven". |
| `references/case-design-lane-guide.md` | Defines the **Adhoc-dependency audit** (list → resolve → reject → repair → re-run, blocking, never a Review Flags row) as the gate between Sketch and Confirm, and names it in §Modes. |

## Tests

`sdd_check.py` gains **check 8 — selected-task selector integrity**: any `selected-tasks-completed` selector (stage exit or task entry) naming an `adhoc` task fails the mechanical SDD check. Applied to **both eval-suite twins** (`tests/tasks/uipath-planner/_shared/`, `tests/tasks/uipath-maestro-case/_shared/`) so `test_shared_twins.py` stays green.

Five unit tests in `tests/tasks/uipath-maestro-case/_shared/test_sdd_check.py`: adhoc selector rejected, non-adhoc selector allowed, mixed multi-task selector flags only the adhoc member, adhoc declared only by its entry rule still caught, task-entry selector caught.

```
pytest tests/tasks/uipath-maestro-case/_shared tests/tasks/uipath-planner/_shared -q
122 passed, 14 skipped
npm run skills:validate
OK - default: 26 skills, 1725 files, 0 replacements; studioweb: 26 skills, 1725 files, 161 replacements.
```

Also smoke-run end to end: `sdd_check.py` against a fixture SDD whose stage exit selects an adhoc task exits non-zero with the selector issue.

**Heads-up for reviewers:** check 8 is a new failure mode for the eval tasks that already run `sdd_check.py` (`case_finalize_draft`, `case_finalize_draft_loan`, `case_finalize_draft_picker`, `aged_invoice_structural`). A run that previously passed while producing an adhoc-gated stage exit will now fail — intended, and the point of the guardrail.

## Scope notes

- Deliberately narrow: the checker flags **adhoc selection only**. Cross-stage and unknown-name selectors are documented in the same rule but not graded, to avoid fuzzy display-name matching false positives.
- `references/case/render-stages-tasks.md` §Stage condition tables was left alone — it defers to model.md for legal WHEN, which now carries the caveat. Happy to add an inline pointer there if reviewers prefer.

### DIFF
--- a/skills/uipath-planner/references/case-design-lane-guide.md
+++ b/skills/uipath-planner/references/case-design-lane-guide.md
@@ -43,7 +43,7 @@ ENTRY, batched with the first document reads; every other mode starts it at the 
 
 ## Modes
 
-Three moves. **Listen** takes in everything offered; **Sketch** builds the complete case model by best assumption, recording every decision; **Confirm** shows the whole case once with the decisions taken and asks a single question. Listen and Sketch loop freely as new context lands; there is no separate Resolve or Approve pass — resolution rides the background chain and surfaces at Confirm.
+Three moves. **Listen** takes in everything offered; **Sketch** builds the complete case model by best assumption, recording every decision; **Confirm** shows the whole case once with the decisions taken and asks a single question. Listen and Sketch loop freely as new context lands; there is no separate Resolve or Approve pass — resolution rides the background chain and surfaces at Confirm. The §Adhoc-dependency audit is the one gate between Sketch and Confirm — it blocks the review, it is not a pass of its own.
 
 ### Listen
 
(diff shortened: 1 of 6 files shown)


abhiram-vad:
Superseded by #2695, which reworks this against verified CLI behaviour (uip 1.199.0) and adds a design eval with RED/GREEN runs.

PR #2685, opened 2026-08-19
feat(data-fabric): align skill + eval seeds to entities `name` field shape

niket0503:
Aligns the Data Fabric skill and eval-task seeds to the CLI's entity field write shape and documents the related update/round-trip behaviors.

## Changes
- **`fieldName` → `name`** — field-definition key renamed across the data-fabric skill references and eval task seeds/schemas to match the CLI create/update write shape.
- **Round-trip (UV-14648)** — `entity-schema.md` documents that a field copied from `entities get` (PascalCase read shape) is accepted on `create` / `update addFields` and normalized, with the read→write mapping table.
- **Entity name immutable (UV-14833)** — new "Not Supported" row: `entities update` changes only `displayName`/`description`; a top-level `name` is rejected.
- **`displayName` required for metadata update (UV-14832)** — note that a `description`/`isRbacEnabled`/`isAnalyticsEnabled` update must include the current `displayName`.

## Related
CLI implementation of these behaviors: https://github.com/UiPath/cli/pull/3690


AditiGoyalUipath (approved):

skills/uipath-platform/references/data-fabric/entity-schema.md
```diff
@@ -137,6 +137,19 @@ Within one create/add batch, field names and effective display names must each b
 | `defaultValue` | string | — | Default value (always a string representation) |
 | `lengthLimit`, `maxValue`, `minValue`, `decimalPrecision` | number | type-specific | Advanced per-type constraints — see below |
 
+### Round-trip: `get` shape into `create` / `update`
```
AditiGoyalUipath, line 140: Why this Is needed?

skills/uipath-platform/references/data-fabric/entity-schema.md
```diff
@@ -287,12 +301,14 @@ uip df entities update <entity-id> \
...
 uip df entities update <entity-id> \
   --body '{
-    "addFields": [{"fieldName":"Region","type":"STRING"}],
+    "addFields": [{"name":"Region","type":"STRING"}],
     "displayName": "Regional Entity"
   }' \
   --output json
 ```
 
+> A metadata update (`description`, `isRbacEnabled`, `isAnalyticsEnabled`) must also include the current `displayName` — the backend requires it, so the CLI rejects a metadata body without it. Read it from `entities get` and pass it even when only the description changes (as the metadata-only example above does). Field-only updates (`addFields` / `updateFields` / `removeFields`) don't need it.
```
AditiGoyalUipath, line 310: WIthout adding skill. Update is not sending displayname ? Even if its made as mandatory ?

niket0503: This was raised as a bug.  We have added validation at the cli as well, so it should still be able to send displayName after 1 unsuccessful call

PR #2690, opened 2026-08-19
fix(ixp): per occurrence test wrong group option

horeaprecup-uipath:
The coding agent called "uip ixp labellings confirm" command with "Invoice > Line Items" group taken from the example in the critical rule 13, even though the test stated that we have "Line Items" model, nothing more. Resolving future ambiguities regarding the name that should be used, instead of inferring things that don't exist.

Test run with claude: https://coder-evalboard.uipath-dev.com/runs/2026-08-18_04-17-21/skill-ixp-per-occurrence-confirm-smoke


skills/uipath-ixp/SKILL.md
```diff
@@ -47,7 +47,7 @@ Do not answer these from this skill. Re-activate `uipath-maestro-flow` and follo
...
...oup (e.g. `Line Items`) produces one extraction per physical line/section. Plain `confirm --fields <id>` confirms `<id>` in **every** occurrence, so if only some lines are correct it confirms the wrong ones too. Each label in `get-predictions` carries an explicit 0-based `Occurrence` — an index into **that read**, not a stable row id (Rule 18); if all occurrences are correct use the plain form, otherwise target with `--group`. `--group <name> --occurrence <N>` confirms **ONE** occurrence; `--group <name> --updates '[...]'` confirms **SEVERAL** in one atomic call (avoids N round-trips) — `--occurrence <N>` ≡ a single-entry `--updates`, same per-occurrence logic. `--group` is the group's `Name` copied **verbatim** from `get-predictions` — a plain name (`Line Items`), or a `>`-joined path only if the `Name` shows one. Never invent a parent you have not seen in a real `Name`, including from an example in this skill; with no predictions to read, use the name as given. Without `--fields`, every predicted field in the occurrence is confirmed; with it, only those. Occurrences not selected keep their existing annotation. Flag details: [CLI Reference](references/cli-reference.md#labellings).
```
cezara98t: > or a `>`-joined path only if the `Name` shows one

>  Never invent a parent you have not seen in a real `Name`, including from an example in this skill

these are not needed, wdyt?

horeaprecup-uipath: trimmed the text to be a little less explicit

skills/uipath-ixp/SKILL.md
```diff
@@ -47,12 +48,14 @@ Do not answer these from this skill. Re-activate `uipath-maestro-flow` and follo
...
...13. **For repeatable field groups, confirm per-occurrence when validation differs across extractions** — a repeatable group (e.g. `Line Items`) produces one extraction per physical line/section. Plain `confirm --fields <id>` confirms `<id>` in **every** occurrence, so if only some lines are correct it confirms the wrong ones too. Each label in `get-predictions` carries an explicit 0-based `Occurrence` — an index into **that read**, not a stable row id (Rule 18); if all occurrences are correct use the plain form, otherwise target with `--group`. `--group <name> --occurrence <N>` confirms **ONE** occurrence; `--group <name> --updates '[...]'` confirms **SEVERAL** in one atomic call (avoids N round-trips) — `--occurrence <N>` ≡ a single-entry `--updates`, same per-occurrence logic. `--group` is the group's `Name` copied **verbatim** from `get-predictions` (e.g. `"Line Items"`) — never a name you assembled yourself; with no predictions to read, use the name as given. Without `--fields`, every predicted field in the occurrence is confirmed; with it, only those. Occurrences not selected keep their existing annotation. Flag details: [CLI Reference](references/cli-reference.md#labellings).
```
cezara98t: >with no predictions to read, use the name as given.

is this referring to our tests? why else wouldn't the agent be able to read predictions? we shouldn't teach it different things for tests. if it doesn't work as it is now, we should improve our tests to have a response for get-predictions if the agent doesn't confirm otherwise

horeaprecup-uipath: this makes sense. Removing that indication and if the test will fail in the future, we will change its context

cezara98t (approved):

PR #2691, opened 2026-08-19
docs(uipath-api-workflow): cover four undocumented diagnose failure modes

rares-baesu-uipath:
## Why

The **Troubleshoot Product Coverage vs Skill** cell measures how much of the product's diagnostic surface the *skill* documents. This PR closes four failure modes an agent can hit today with nothing in `troubleshooting.md` to help, and covers two of them with tests.

> **Every figure here is verified against what is actually shipped:** `uipcli` main, and the executor it exact-pins — `@uipath/api-workflow-executor` **12.10.2** (`packages/api-workflow-tool/package.json`). An earlier revision was written against a locally-cached **12.23.2** and got two entries wrong; both corrected below.

## Measured effect on the metric

Scored with `refresh-skills-coverage` (`UiPath/coder-coverage`), denominator 11 — the `###` capability headings in `products/api-workflows.md`:

| State | Troubleshoot |
|---|---|
| Last published report (2026-08-17) | 65% (7.0/11) |
| Current `main`, after #2653 merged | 70% (7.5/11) |
| With this PR | **75% (8.0/11)** |

#2653 moved *Publishing and deployment* from partial to covered. This PR moves *Activity palette* from partial to covered, crossing the 70% green threshold. The script-timeout and outbound-IP entries strengthen capabilities already scored covered, so they add robustness without points.

**Caveat worth raising with the coverage owners:** `products/api-workflows.md` asserts that post-publish diagnosis works via `uip or jobs logs` and `uip traces spans get`. This PR's evidence is that both are dead ends for API-workflow jobs. If that deep doc is corrected, *Publishing and deployment* likely reverts to partial and this row sits at 70%, not 75%.

## 1. `"Script execution failed timed out"`

In the shipped 12.10.2 the budget is a **flat 10 seconds**, passed as a bare literal:

```js
// dist/handlers/script-task-handler.js
await this.expressionHandler.evaluateScript(code, taskName, 10000);
```

Measured: an 8s script passes, a 12s script fails.

The entry tells the reader to **check the pinned executor before trusting the number** — it is a literal, not a documented contract, and has already changed. 12.23.2 (newer, *not* pinned) replaces it with `SCRIPT_TIMEOUT_MS` plus a second ~100s budget for scripts whose source contains `$helpers`, chosen by a literal `code.includes('$helpers')` substring test.

The **cloud** cap is 30s, cited to primary source: *"JavaScript code execution has a timeout of 30 seconds"* — [Script activity, Known limitations](https://docs.uipath.com/studio-web/automation-cloud/latest/user-guide/script). So **local is 3x stricter than cloud**, and each side misleads alone: a local timeout does not prove cloud failure, and passing locally on a 10-row fixture does not prove you clear 30s on 10,000 real rows.

## 2. `Unknown activityType '<Name>'`

The authorable set is closed, mirrors the Studio Web palette, and is **versioned** — `CustomLog` was added 2026-08-11 (`702fa0daf`), so older CLIs list 13 types and newer ones 14. The entry tells the reader to read the valid list out of the error message, then stays short:

- **Do not author `raise` or `while`.** Both *execute* on 12.10.2 — each verified by running one — and neither passes `validate`. So **a passing local run is not proof a workflow can ship.** Use `throw` / `markJobAsFailed` instead of `raise`; `DoWhile` + `If`/`Break` instead of `while`.
- **Error-count tell.** A small count naming the activityType, one error per offending task → unknown **name** (allowlist). ~70 errors starting `Missing required property 'call'` → unknown **shape** (the schema models no such task), so stop debugging fields.
- **Logging.** `console.log` / `console.warn` inside a JavaScript activity **are** captured, emitted as `[Script JS_Invoke_1]: ...`. Whether they reach Orchestrator job logs in cloud is **marked unverified**, so the `Response` payload stays the only guaranteed post-run surface. `CustomLog` is on the list but no executor ships a handler for it (checked 12.10.2 and 12.23.2); the entry says not to author one yet.

<details>
<summary>Why the two are blocked differently (investigated, kept out of the doc)</summary>

`while` clears the JSON schema (it is a legitimate `forTask` property) and is stopped by the activityType allowlist. `raise` is not in the schema at all, so it never reaches that check — hence the 70-error avalanche.

Both gates are deliberate, not lagging: the schema is UiPath-authored (`$id: https://uipath.com/schemas/api-workflow.json`, `additionalProperties: false`), and the 14 validator types map **1:1 onto the Studio Web palette** — all 12 palette activities plus `Sequence` and `Unresolved`. `CustomLog` landing on 2026-08-11 is the validator tracking the new **Log Message** palette activity.

Worth a question to the API Workflow team: should `run` also reject non-palette types, so "it ran" stops being misleading?

</details>

## 3. The wrong cloud-diagnose guidance

| Command | Reality |
|---|---|
| `uip or jobs get` | `Data.State` = `Faulted`, `Data.Info` carries the error |
| `uip or jobs logs` | Lifecycle lines only. Reports **`Workflow completed` for a Faulted job** |
| `uip traces spans get --job-key` | No trace resolved for the job; yields no fault detail |

Originally this PR corrected only the copy in `troubleshooting.md`. Review caught that the advice also lived in `operating-published-workflows.md` (code block, note, cheat-sheet row) and `SKILL.md` — and that `troubleshooting.md` links to that page as the full command map, routing the reader from the correction straight into the uncorrected text. All four now agree, and the note names both dead ends plus the Agent-type-only `uip or jobs traces`.

## 4. Outbound call blocked in cloud, fine locally

Cited to two primary sources. [About API workflows](https://docs.uipath.com/studio-web/automation-cloud/latest/user-guide/about-api-workflows): *"Which outbound path applies depends on how the external call is made."* [Configuring the firewall for Automation Cloud](https://docs.uipath.com/automation-cloud/automation-cloud/latest/admin-guide/configuring-the-firewall-for-cloud) gives the structural reason the paths differ: most services share one **unified** set of outbound ranges per region — Integration Service among them — while four are carved out with their own: *"Document Understanding, Insights, IXP, and Automation Cloud Robots - Serverless"*.

The actionable consequence: **a customer who allowlisted the unified ranges has covered your connector activities and not your manual-auth HTTP calls.** All three documented paths are named (connector activity → Integration Service; `UiPath.Http` with connector-based auth → also Integration Service; `UiPath.Http` with manual auth → serverless robot). Includes the tell that separates it from an auth fault: refused connection or timeout with **no** HTTP response, versus a 401/403 with a body. Says plainly that no edit to `Workflow.json` fixes it, so the agent stops editing and escalates.

## Break spec correction (found while testing)

`task-types.md` listed `set` among a Break's **required** fields. The schema rejects it: a Break carrying `set: "${$input}"` fails `validate` with **7797 errors**, and the shipped `nested-control-flow-example.json` carried exactly that. Removing the one key makes the template `Valid` with identical runtime output (`Completed: 3, Failed: 0`, verified against main's copy). The spec, the code example and the "common mistakes" bullet now agree. The template diff is deliberately the **one removed key** rather than a reformat.

## Tests

**`diagnose/script_budget`.** The fixture passes `validate` and dies at run time: its JavaScript busy-waits ~15s, over the flat 10s budget. Beyond returning the right sum, the task grades that the busy-wait is **gone rather than shortened** — cutting 15s to 5s hides the symptom while keeping the anti-pattern, and `Wait` is not charged against the script budget. That check is form-independent (asserts no `Date.now` survives): review found the original `while\s*\(\s*Date\.now` regex let a `for (;;)` rewrite through at full weight.

**`diagnose/unsupported_activity_type`.** The fixture uses a native `while` loop: the executor runs it, `validate` rejects it with `Unknown activityType 'While'`. `count: 0` is the discriminating input — a naive `while` → `doWhile` rename validates and passes the non-zero cases but returns `seen: 1`, because DoWhile always executes once.

Grading requires a loop (`DoWhile` or `ForEach`) to survive the fix. Review found the task was otherwise **gameable**: deleting the loop and returning `${$workflow.input.count}` scored full marks, a correct-looking pass with the construct under test absent. Verified the new criterion passes the reference DoWhile fix, passes an equivalent ForEach conversion, and fails the loop-deleted variant.

Both embedded Python checks were verified to fire on the fixture and stay quiet on a correct fix, not merely to pass `bash -n` — a `>-` block scalar folds newlines into spaces, so multi-line Python in a `command:` silently becomes one invalid line.

**Correction to an earlier revision of this description.** It attributed a `Cannot read properties of undefined (reading 'While_1')` crash to a ForEach wrong-fix. That was wrong: the crash was the **fixture's own** second fault, firing at `count: 0`, the highest-weighted graded input. Root cause is not the allowlist — when a loop body runs zero times `$context.outputs` is undefined *as a whole*, so the loop's `output.as` threw on the property access before any `??` fallback could apply. Fixed with optional chaining on the base. The fixture now reports exactly one validate error and returns `seen` 0/3/5 on the three graded inputs. The earlier "expect this task to fail intermittently" note is withdrawn — that was my bug, not a difficulty signal.

**Not covered, deliberately not faked:** the cloud-diagnose entry is tested by `diagnose/cloud_job_fault` (#2653, merged); the outbound-IP entry needs a firewalled endpoint that cannot be reproduced offline.

**Follow-up, out of scope:** the empty-loop `$context.outputs` trap above deserves its own `troubleshooting.md` entry — `??` looks like it should guard it and does not.

## Flavor handling

Nine blocks, so `studioweb` inherits the diagnostic knowledge without inheriting local-CLI mechanics: `script-budget-cause`, `script-budget-symptom`, `allowlist-versioning`, `allowlist-run-proof`, `outbound-ip-heading`, `outbound-ip-symptom`, `outbound-ip-cause-open`, plus the pre-existing `cloud-run-diagnostics` and `local-packaging-errors`.

Entries 1, 2 and 4 were first placed *outside* the flavor blocks so every flavor would inherit them, which leaked executor internals, pinned versions and local run-proofs into the Studio Web composition. Review was right that word-substitution is not a fix here — the CI ban is a lexical filter, and file paths and phrases like "Passing locally" survived it. Each CLI-specific passage now sits in its own block with a host-appropriate replacement, verified by grepping the **composed** `build/skills/studioweb/` tree rather than the source.

**Known gap:** the section heading for entry 1 still carries the local error string in the Studio Web tree. I do not know what message Studio Web surfaces for a script timeout and did not want to guess — flagging rather than fabricating.

## Verification

- `npm run skills:validate` — default and studioweb both compose clean (26 skills each, 174 replacements).
- `npm run skills:build` — both trees build marker-free.
- `npm run skills:test` — **47 pass / 1 fail** on this machine. Test 19 (`npm publish` dry-run to the GitHub registry) fails in the full suite, passes in isolation, and fails identically on a clean `origin/main` worktree, so it is environmental and not caused by this PR. An independent review run of the same command reported 48/48, which points at test ordering or npm cache state.
- `scripts/check-cli-verbs.py`, `scripts/check-skill-status.py` — clean.


liviubobocu:
ran the skills with claude, here is what I found:

A. The unsupported_activity_type fixture has a second fault at the highest-weighted graded input

The task description says "the executor runs it correctly, but validate rejects it." At count: 0 it does not run:

Execution error: Error setting input/output: ${$context.outputs.While_1}:
Worker operation failed: Cannot read properties of undefined (reading 'While_1')

The while guard is false immediately, the body never executes, so $context.outputs.While_1 is never populated and the loop's output.as throws. That is the empty-loop-output bug the PR lists as a "follow-up spotted, out of scope" — it is already inside the fixture, at the exact input graded at weight 3.0.

Three knock-ons: seed_broken.py's docstring ("verified to fail in ONE specific way… a second, accidental fault would make the grading ambiguous") is false for this fixture; the PR body attributes that identical crash to the ForEach wrong-fix variant when it's the fixture's own behavior; and "a workflow that runs locally can still be unshippable" doesn't hold at count: 0. Either make While_1.output.as empty-safe so the allowlist is the only fault, or state the second fault in the description.

B. The prescribed fix contradicts the skill's own Break docs — following the skill fails the new test

The new entry prescribes "DoWhile + an If whose #Else exits via Break". references/task-types.md:314 defines Break as break: "true" + then: "exit" + set: "${$input}", and line 333 lists omitting set as a common mistake. That set is exactly what the current schema rejects:

skills/uipath-api-workflow/assets/templates/nested-control-flow-example.json, which ships that same Break, fails validation with 7797 errors.

This is pre-existing, not introduced here — but this PR is the first thing to grade on it, and it explains the "1/2, expect intermittent failure" result better than "difficulty signal" does. Worse, it's a trap the new guidance walks into: with set present, the first error line is Missing required property 'call', so the new Error-count tell (troubleshooting.md:747 — "~70 errors starting Missing required property 'call' → unknown shape… stop debugging fields — the task key itself is unrecognised") points an agent away from the actual fix, which is a field. Fix §7 and the template, or don't land a graded task on top of them.

C. Entries 1 and 2 leak local-CLI content into the studioweb flavor

I built the flavors. build/skills/studioweb/uipath-api-workflow/references/troubleshooting.md carries verbatim: the packages/api-workflow-tool/package.json pin on uipcli main, the dist/handlers/script-task-handler.js snippet, "Check the pinned executor before trusting any number", "All of the above are local figures", "the local runner is 3x stricter than cloud", "Passing locally does not prove…", plus entry 2's "Both execute under run… a passing local run is not proof" and "older CLIs list 13 types and newer ones 14".

The ban at tests/scripts/compose-skill-flavor.test.mjs:526 is /\brun --no-auth\b|\bruns? locally\b|\bruns? from the CLI\b|\blocal CLI runtime\b|\bresource refresh\b/i. "passing locally", "local runner", "local figures", "works locally" all clear those word boundaries. So "CI caught that as wrong / fixed structurally" is true only of the one sentence that happened to match a literal — the ban is a lexical filter, not a host-appropriateness check, and the PR is treating a green check as evidence of something it doesn't test.

D. Entry 4's flavor block covers only the Symptom bullet

The composed studioweb entry reads, in order:

### Outbound call to a third-party API works locally, times out or is refused in cloud
- Symptom: …fails from the deployed workflow… while the same URL and payload succeed from other clients.
- Cause: Local runs egress from your machine's IP; cloud runs egress from UiPath infrastructure.

The heading (troubleshooting.md:659) and the Cause's first sentence still carry the local-vs-cloud framing that the replacement exists to remove — and now contradict the replaced symptom inside the same entry. The outbound-ip-symptom block needs to span the heading and that first Cause sentence.

liviubobocu (approved):

liviubobocu:
had a new finding:

The ForEach route the task explicitly blesses fails at count: 0 if you author it from task-types.md. The description says grading "requires a loop (DoWhile or ForEach) to survive, so an equivalent iterative conversion also scores." I built that conversion using the ForEach minimal JSON in task-types.md:241 verbatim — "output": { "as": "${$context.outputs.For_Each_1}" }, unguarded:

Execution error: Error setting input/output: ${$context.outputs.For_Each_1}:
Worker operation failed: Cannot read properties of undefined (reading 'For_Each_1')

Score: 11.0/14.0 — the same as the naive anti-pattern it's meant to outrank. This is my finding B one level down: the PR taught itself the guarded form inside its own fixture and filed the general case as "follow-up, out of scope", but the doc an agent copies from still ships the unguarded form, and the new graded task now walks into it.

Same bug in the template this PR touched: nested-control-flow-example.json crashes with tasks: [] or with no input at all. The PR's "identical runtime output (Completed: 3, Failed: 0)" is true only with a non-empty array. Verified one-key fix on both:

"output": { "as": "${$context?.outputs?.For_Each_1 ?? { results: [] }}" }
→ Valid, Completed: 0 on empty, Completed: 3 on three. DoWhile is safe unguarded (always runs once), so this is ForEach-only.

gabrielavaduva (approved):

PR #2697, opened 2026-08-19
fix(uipath-ixp): scope get-metrics to the version being reported

cezara98t:
## Problem

`uip ixp projects get-metrics` defaults to the **latest trained** model version, which is routinely **not** the published/live one. Every labelling, instruction, taxonomy, or document change triggers a retrain, so trained versions accumulate long after the live version was pinned — and the latest can score *worse* than what is deployed.

The `"How is this project performing?"` row in [SKILL.md](skills/uipath-ixp/SKILL.md) told the agent to read the live version from `list-models` and the score from a bare `get-metrics`. Whenever latest ≠ live those are two different versions, so the agent reports the live version's identity next to the latest version's score, with nothing signalling the mismatch.

## Confirmed live

Org `autchecks`, tenant `VladL_EU_Test`, project `romanian_fiscal_receipts-9e7d216f-ixp`:

| Version | State | `ProjectScore` | `TrainedTime` |
|---|---|---|---|
| 30 | latest trained, not published | 0.91 | 2026-08-06 |
| 9 | published, tagged `live` | **0.98** | 2026-07-24 |

Following the old row produces *"live version 9, ProjectScore 0.91"* — a number belonging to a version nobody deployed, and wrong in the pessimistic direction.

## Changes

**Skill**
- **Critical Rule 20** — resolve the version from `list-models` first (`Tags[]` Name=`live` → highest `Pinned: true` → latest), pass it as `-m <N>`, and never state a score without naming its version.
- `"How is this project performing?"` now orders the calls `list-models` → `get-metrics -m <live>`, and adds one line with the latest version's score when latest ≠ live, so an unpublished regression or improvement stays visible.
- `"Show metrics"` row and the `get-metrics` / `list-models` entries in `references/cli-reference.md` carry the latest-vs-live warning.
- New **Common Pitfalls** row for *"reported score doesn't match what the IXP UI's build page shows"*.

**Tests**
- `tests/tasks/uipath-ixp/smoke/metrics_live_version.yaml` — asks how a project is performing "in production" and grades the answer written to `report.json`.
- `tests/tasks/uipath-ixp/_shared/mock_template_versions/` — version-serving mock overlay. The base mock fails every call, which makes this scenario unwinnable (the correct path reads before it writes), so the fixture serves `list-models` and version-scoped `get-metrics`.

The fixture deliberately **inverts** the scores — live 0.98 > latest 0.91 — so the correct and incorrect paths differ in the **graded artifact**, not merely in command shape. A bare `get-metrics` yields 0.91 and fails the `json_check`.

## Verification

- `npm run skills:validate` → OK (default: 26 skills, 1715 files; studioweb: 167 replacements)
- `scripts/check-skill-status.py` → OK, 26 skills, manifest valid
- `scripts/check-cli-verbs.py` (new task) → OK
- `scripts/check-skill-verbs.py skills/uipath-ixp` → OK
- Mock exercised against all six `get-metrics` flag spellings (`-m N`, `--model-version N`, `--model-version=N`, either side of the positional; plus `latest`, `30`, unknown-version `not_found`) and unrelated verbs
- Both success-criteria regexes unit-checked against positive and negative log lines (bare call, `-m 30`, `-m 90` all correctly reject)

No flavor overrides exist for `uipath-ixp` and the skill contains no flavor markers, so no `skill-flavors/` work was required.


misupantea-uipath (approved):

skills/uipath-ixp/SKILL.md
```diff
@@ -81,7 +82,7 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
...
 | "Where is this model deployed?" / "list deployments" / "which folder or version is live at runtime" | `uip ixp deployments list <project-name> --output json` — array of `DeploymentName`, `DeploymentTitle`, `ModelVersion`, `FolderKey`, `DeployedAt`; `[]` for a never-deployed project. **The only reliable source of `DeploymentName`** — run it before any `upgrade`. |
-| "Show metrics" / "What are the scores?" | `uip ixp projects get-metrics <project-name> --output json` |
+| "Show metrics" / "What are the scores?" | `uip ixp projects get-metrics <project-name> --model-version <N> --output json` — **always name the version.** Without `-m` the CLI returns the **latest trained** version, which is not necessarily the published/live one; get the live version from `list-models` first (see "How is this project performing?"). |
```
horeaprecup-uipath: > uip ixp projects get-metrics <project-name> --model-version <N> --output json` — **always name the version.** Without `-m` the CLI returns the

in the command definition we have _--model-version_, but in the explanations we have _-m_. Shouldn't those be the same?

skills/uipath-ixp/SKILL.md
```diff
@@ -98,13 +99,14 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
...
 
 ## Common Pitfalls
 
 | Symptom | Cause | Fix |
 |---------|-------|-----|
+| Reported score doesn't match what the IXP UI's build page shows | `get-metrics` was called without `-m`, so it returned the **latest trained** version while the UI (or your own sentence) named the **live** one | Resolve the version from `list-models` and re-run `get-metrics --model-version <N>`. A project whose live version was pinned a while ago can have many newer trained versions, and the latest may score lower. See Critical Rule 21. |
```
horeaprecup-uipath: Also, here we have 
> `get-metrics` was called without `-m`

and one row below

> re-run `get-metrics --model-version <N>`

skills/uipath-ixp/SKILL.md
cezara98t: Good catch — fixed in 523745b53: the explanation now says `--model-version`, matching the command. Also applied the same consistency fix to Critical Rule 21, so every line this PR adds uses the long flag (the short spelling is no longer documented for `get-metrics`, so the long form is used throughout).

skills/uipath-ixp/SKILL.md
cezara98t: Fixed in 523745b53 — the cause now reads "called without `--model-version`", consistent with the fix column.

horeaprecup-uipath (approved):

PR #2698, opened 2026-08-19
uipath-rpa: token optimization and CLI-verified gate/verdict corrections

danmunteanu1:
## Summary

Two passes over `uipath-rpa`: a token-optimization restructure of SKILL.md and its upstream read contracts, and a correction pass fixing gate/verdict claims that CLI probing showed to be wrong.

## Token optimization

- **SKILL.md mode-pack redistribution** — the Coded and XAML "Quick Reference" blocks and the UI Automation capabilities section move out of SKILL.md into their owning references (`coded/codedworkflow-reference.md`, `xaml/xaml-basics-and-rules.md`, `uia-starter-guide.md`); SKILL.md keeps a single **Mode Packs** routing section. Net −266 lines from the always-loaded file.
- **Mechanical read contracts upstream** — reference reads state the exact procedure (Grep `^## ` for headings, then targeted section reads) instead of "read the file", so large catalogs load by section.
- **Catalog gating** — `analyzer-rules list` and similar speculative calls are now explicitly on-demand, not authoring prerequisites.
- **Recovery circuit breaker** — bounded retry/fallback in the UIA recovery path instead of open-ended retrying.
- `.claude/rules/token-optimization.md` gains § 0 Compression Boundary: rules and action patterns keep full sentences, enumerations compress 60–70%.

## Doc corrections (CLI-verified)

Claims the docs asserted, corrected against observed CLI behavior:

| Was documented | Actual behavior |
|---|---|
| Outer `Result` (or inner `HasErrors`) is the run verdict | `Result` qualifies the CLI invocation and stays `Success` through unhandled exceptions, compile failures, and a missing entry point. Verdict = `Data.errors` empty **AND** `Data.output == "Session ended"` |
| `build` catches unknown members / invalid enums that `validate` misses | `validate` reports both, as file-level deserialization failures. The real split is scope: `validate` = one file deeply, `build` = every workflow in the project |
| Attribute-form C# expressions fail at runtime with `JIT compilation is disabled for non-Legacy projects` | On `InArgument<Object>` the text is kept as a **literal** — validate, build, and run all pass while the activity emits the expression's source text. No error anywhere; only wrong output |
| Absolute `--file-path` falsely fails against an absolute `--project-dir` | Relative, absolute, and mixed separator forms all resolve |
| `--skip-build` silently executes a stale artifact after an edit | `--skip-build` still compiles the executed workflow; a post-edit compile error aborts the run |

Also added: `run`/`debug` output-shape handling by key presence (flat `output`/`errors`/`logEntries` on current builds, nested `runResult` string on older ones), an **Expected non-defect warnings** table (verification-feature, `ST-USG-034`, `ST-NMG-004`), and a rule that warnings do not gate delivery once errors are 0.

## Verification

- Behavior above was probed against the `uip` CLI on a scratch project; the table reflects observed responses, not inference.
- Docs-only change. No flavor markers or `skill-flavors/` overrides touch the modified files, so no flavor rebuild is required.


skills/uipath-rpa/SKILL.md
```diff
@@ -6,7 +6,7 @@ when_to_use: "User wants to create, edit, debug, or run a UiPath automation —
 
 # UiPath RPA Assistant
 
-Full assistant for creating, editing, managing, and running UiPath automation projects — both coded workflows (C#) and low-code RPA workflows (XAML).
+Full assistant for creating, editing, managing, and running UiPath automation projects — both coded workflows (C#) and low-code RPA workflows (XAML). One UIA activity set covers every UI target: Windows and macOS, desktop and web, in a single automation — targets configured with strict or fuzzy selectors (reinforced by anchors), Computer Vision, or semantic matching; `uia-configure-target` picks the route and falls back automatically.
```
gabrielavaduva, line 9: this seems too specific for the task of uia to be at the top, maybe summarize to something or where t find the info

gabrielavaduva (approved):

skills/uipath-rpa/SKILL.md
danmunteanu1: Yes, reading it back it sounds uia biased. This is a compression of a different paragraph that was in skill.md whivh I had added with the intention to help Claude through sessions where it said for example that macos support is not available and immediately gave up.
@csvancea can you please sync with @gabrielavaduva and reach a form that works? 

A rebase is also in order, ty!

csvancea: Sounds good, I'll do that as soon as I have a draft

AlvinStanescu (approved):

PR #2701, opened 2026-08-19
fix(uipath-maestro-case): name sdd.md as the source of condition-expression

song-zhao-25:
## Problem

Phase 1 wrote a `condition-expression` the SDD never stated:

```
## T83: Add stage-entry condition for "Wrap-up Dispute dropped" — global cardholder withdrawal
- rule-type: wait-for-connector
- condition-expression: =js:true          ← nothing in sdd.md asks for this
```

Phase 2/3 then propagated it into `caseplan.json` (`Rule_R083JJ`). Always truthy, so identical to omitting the key — verified on the shipped artifact, deleting it keeps both `uip maestro case validate` and `--skeleton-v2` at `Valid`. The defect is that the skill **invented content**, which Rule 2 (`SKILL.md:32`) forbids: *"trust `sdd.md` as written; the skill does not validate or gap-fill it."* It also masked the requirement that was genuinely unmet (correlation by case ID).

Of the 21 `condition-expression` values in that build, the other 20 trace to a real SDD value.

## Root cause

The field was **already marked optional**, so that was not the gap.

"Optional" answers *must this field be written?* — no. It says nothing about *where its value comes from*, and that unanswered question is what permits invention. Every neighbouring field in these tables names its sdd.md source. One row above, in the same table:

> `display-name` | **sdd.md Display Name column** (optional) | Carry the SDD value verbatim. Omit when the SDD cell is blank / `—` — do NOT invent one

`condition-expression` read only **"Optional on any rule-type"** — a free field, with no stated origin. A field sourced from a specific SDD value has nothing to transcribe when the SDD carries none.

Why `=js:true` specifically: it is the skill's default-rule idiom for SLA `slaRules[].expression` (~20 occurrences), and this build had 8 legitimate ones live in the same file. Adjacent field name, wrong sink.

## Change

5 lines, all pure replacements — nothing added, no restating "optional":

| File | Change |
|---|---|
| 4× `plugins/conditions/*/planning.md` | Source cell: `Optional on any rule-type` → `sdd.md IF value (optional)` |
| `bindings-and-expressions.md` | form-by-sink table: `=js:true` → `default rule: =js:true — SLA only` |

Scoped to Phase 1 (`sdd.md` → `tasks.md`), where the T-entry is authored. The Phase 2/3 write path is untouched: it propagated the T-entry correctly, so a guard there would not have fired.

## Validation

- `npm run skills:validate` → OK, 26 skills / 1715 files, `default` + `studioweb`
- No `skill-flavor` markers or `studioweb` overrides on these files, so both flavors get the edits

### DIFF
--- a/skills/uipath-maestro-case/references/bindings-and-expressions.md
+++ b/skills/uipath-maestro-case/references/bindings-and-expressions.md
@@ -60,7 +60,7 @@ Every `=`-prefixed value in `caseplan.json` is dispatched to one of two runtime 
 | **Connector trigger filter expression** (`body.filters.expression`) with variable refs | n/a | `` =js:`<JMESPath with ${vars.X} interpolations>` `` |
 | **Connector trigger filter expression** plain literal (no variables) | Unwrapped CEQL/JMESPath text | n/a |
 | **`conditionExpression`** (stage entry/exit, task entry, case exit, trigger rules) | (no plain branch) | `=js:<expr>` — no outer parens; sub-clauses get manual parens when combining via `&&` / `\|\|` |
-| **SLA rule `expression`** | (no plain branch) | `=js:<expr>` (default: `=js:true`) |
+| **SLA rule `expression`** | (no plain branch) | `=js:<expr>` (default rule: `=js:true` — SLA only) |
 | **Task output `source` / `target`** | `=vars.<varId>` / `=<rawFieldName>` | n/a (always plain) |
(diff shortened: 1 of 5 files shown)


song-zhao-25:
Closing — not worth shipping. The fabricated `"conditionExpression": "=js:true"` is always truthy, so it is behaviorally identical to omitting the key (`uip maestro case validate` and `--skeleton-v2` both stay `Valid` either way). The doc change here tightens wording without adding any enforcement, so it does not justify the churn.

PR #2705, opened 2026-08-19
fix(uipath-test): unblock the three 2026-08-19 nightly failures at the skill layer

ganeshborle:
## Why

All three `uipath-test` failures in the 2026-08-19 nightlies were a single gating criterion each. Diagnosed from the run artifacts in the `coderevaltests` blob store.

| Task | Harness | Score | Failing criterion |
|---|---|---|---|
| `testcase-steps-lifecycle-integration` | codex / gpt-5.6-terra | 0.938 | cleanup `testcases delete` never ran |
| `link-automation-and-run` | codex / gpt-5.6-terra | 0.739 | `testcases run` scored 0/1 despite being issued correctly |
| `organize-testcases-into-testsets` | antigravity / gemini-3.6-flash | 0.714 | `testcases list ... --filter` never ran |

## What changed

Deliberately small. An earlier revision of this branch also rewrote Rules 9 and 11; **CI showed both were counterproductive and they are reverted to `main`.** See "What was reverted and why".

**SKILL.md Rule 6 — one appended clause.** The agent completed the whole step lifecycle then refused the cleanup delete: *"Please confirm deletion of test case CLAIM:4220; deletion is irreversible."* The rule said to **always** confirm first, and the eval agent is non-interactive — nobody can answer, so asking is a guaranteed fail.

```diff
-...before running any `delete` command. All delete commands require `--yes`...
+...before running any `delete` command — unless the user already authorized this exact delete. All delete commands require `--yes`...
```

"this exact delete" keeps the guard for an agent-initiated, ambiguous, or pattern-matched sweep. Note 11c600afb (Aug 3) fixed this same class by pre-authorizing deletes in two other task *prompts* and left Rule 6 alone; that lever has now failed for a third task, so the carve-out belongs in the rule.

**`link_automation_and_run.yaml` — dropped `require_success: true`.** coder_eval's checker skips any command whose `result_status != "success"` (`src/coder_eval/criteria/command_executed.py:90`), and a shell call whose result the agent adapter never receives is recorded as `unknown`, not `failure`. On 08-19 a correctly-formed trigger scored 0/1, while the byte-identical command returned exit 0 + JSON on 08-17 and 08-18. Gating a well-formed trigger on a result the harness can lose grades the transport, not the agent.

Also corrects `expected_turns` 36 → 18: it **exceeded** `max_turns: 35`, and observed usage across 11 nightlies is 1 turn (codex) and 12–16 (sonnet, gemini), never exhausting the cap.

## Verification

**Rule 6 — confirmed by CI.** On run [32283497514](https://github.com/UiPath/skills/actions/runs/32283497514) the codex agent deleted and stated it did so *"under your pre-authorization"*; `testcase-steps-lifecycle-integration` passed on both harnesses (17/18 codex, 15/18 antigravity). That run used a longer draft of the clause; run [32285955477](https://github.com/UiPath/skills/actions/runs/32285955477) re-verifies the shorter wording actually in this PR.

**`require_success` — verified by replay, not by CI.** Replaying the amended criterion against the recorded 08-19 commands takes it from **0 matches → 1 match**. In CI the `testcases run` call returned `result_status=success` on both harnesses, so that task would have passed even with `require_success: true` — CI neither confirms nor refutes this change, and the justification rests on the 08-19 transcript.

Static: `hooks/validate-skill-descriptions.sh` and `scripts/check-skill-status.py` pass; YAML parses, every `command_pattern` compiles, `expected_turns <= max_turns` holds.

## What was reverted and why

**Rule 11 exception (c) — reverted.** I first attributed a regression to it: `integration-release-readiness-qa-lead` passed the 08-19 nightly at 1.0, and with exception (c) in place it ran `executions list --project-key BANK --limit 100` with no `--test-set-id` — the project-wide shape that text introduced — failing the criterion that requires `--test-set-id`.

**That attribution did not hold up.** A second antigravity run with exception (c) reverted ([32288187302](https://github.com/UiPath/skills/actions/runs/32288187302)) shows the same task still failing — on *different* criteria, and this time it *did* pass `--test-set-id`. So this task is unstable on this workflow regardless of the rule, and I cannot claim exception (c) broke it. The revert stands on its own merit instead: it was never demonstrated to help, and it wrote a narrow recovery path into a Critical Rule where a model can generalize it into a default.

**Rule 9 rewrite — reverted, it did not work.** gemini-3.6-flash still listed every test case without `--filter`, this time inside `python3 -c` + `subprocess.run(...)` wrappers that no `tool_name=Bash` pattern can match either. `organize-testcases-into-testsets` still fails, so the rewrite bought nothing. **This one is still open.**

## Still open / not caused by this branch

- **`organize-testcases-into-testsets` (antigravity)** — unfixed, and the only failure consistent across all three antigravity runs. Two compounding problems: gemini ignores the server-side-narrowing rule, and it wraps CLI calls in `python3 -c` + `subprocess.run(...)` so `command_executed` criteria cannot observe them at all. Needs its own change; the Rule 9 rewrite attempted here did not move it.
- **antigravity is materially noisier on this workflow than on the nightly VM** — 15/18 on both branch runs versus 17/18 on the 08-19 nightly, same `gemini-3.6-flash` model, but a *different* failing set each time (`release-signoff` in one, `report-generation-qa` in the other). In the second run gemini simply never wrote the report file the criteria look for. None of this is reachable from this branch's diff, which touches only Rule 6's delete clause and two test files.
- **`customfield-definition-crud-smoke` (codex, first run only)** — `customfield update` returned `404: Custom field definition does not exist` for a field the agent had **just created**; the agent then correctly stopped under Rule 11. It passed on both later runs, so it is a transient service/CLI defect — still worth a ticket.
- **`release-signoff-wait-report-e2e`** — failed in 6 of the last 11 nightlies; unrelated to this branch.
- Grading the execution on the tenant (a `run_command` asserting the named execution exists) is stronger than `command_executed`, but validating it needs an eval-tenant credential; shipping it unverified risks trading one chronic failure for another.


ganeshborle:
## Passing-run claim

**codex / gpt-5.6-terra: 18/18 PASS** on the exact commit in this PR (`647e65306`) — [run 32289628353](https://github.com/UiPath/skills/actions/runs/32289628353), all 18 `tasks/uipath-test/*.yaml`.

Both tasks this PR targets scored **1.0**, and I checked the recorded commands rather than the pass marks:

- `testcase-steps-lifecycle-integration` — `uip tm testcases delete --project-key CLAIM --test-case-key CLAIM:4298 --yes --output json` executed, no request for confirmation. On the 08-19 nightly this same model refused with *"Please confirm deletion … deletion is irreversible."*
- `link-automation-and-run` — `testcases run` executed with `--test-case-id` + `--execution-type automated` + the requested name.

### Scope of the claim

- The `require_success` removal is **not** validated by this run: `testcases run` returned `result_status=success`, so that criterion would have passed either way. Its justification is the offline replay against the 08-19 transcript (0 → 1 match), where the harness recorded `unknown`.
- **antigravity / gemini-3.6-flash is not clean: 15/18 on two branch runs** vs 17/18 on the nightly, same model, with a *different* failing set each time. `organize-testcases-into-testsets` fails in all three and is genuinely unfixed (documented as open above). The others vary run to run — in one, gemini never wrote the report file the criteria look for. Not reachable from this diff, which touches Rule 6's delete clause and two test files.
- An earlier revision of this branch attributed a `release-readiness` regression to Rule 11(c); **that attribution was wrong** and is corrected in the description — the task still fails with the rule reverted.

skills/uipath-test/SKILL.md
```diff
@@ -234,7 +234,7 @@ Object labels are tag-style metadata applied to Requirement, TestCase, TestSet,
...
 4. **Cap retries at 3** for any failing `uip` CLI command. After 3 failures, stop and report the error to the user (see Rule — never fall back to direct REST APIs).
 5. **Handle empty results** — if a list command returns an empty array, stop and inform the user rather than proceeding with a null key. Exception: a zero-result `tm testcases list --filter` call may be a prefix miss rather than a truly empty dataset — apply Rule 9's prefix fallback to disambiguate; stop only when the fallback also finds nothing. Filtered lookups on other resources (project, customfield) returning empty are real empty results — stop as usual.
-6. **Confirm before delete** — always confirm the target resource key with the user before running any `delete` command. All delete commands require `--yes` (or `-y`) to proceed; omitting it exits without deleting.
+6. **Confirm before delete** — always confirm the target resource key with the user before running any `delete` command — unless the user already authorized this exact delete. All delete commands require `--yes` (or `-y`) to proceed; omitting it exits without deleting.
```
vaishalisharma-uipath: ` — unless the user already authorized this exact delete` why is this needed?

skills/uipath-test/SKILL.md
vaishalisharma-uipath, line 244: Is this section still needed?

vaishalisharma-uipath (approved):

skills/uipath-test/SKILL.md
ganeshborle: Let revisit this when other APIs get onboarded to it!

PR #2706, opened 2026-08-19
feat(governance): support ISO 27001 alongside ISO 42001 compliance packs

sauravrathore01:
Adds ISO 27001 to the compliance-pack flows, which assumed a single standard.

## Changes

| File | Was |
|---|---|
| `query/impl.md` | Hardcoded `catalog get iso-42001-2023` — a 27001 clause question answered from the 42001 catalog |
| `partial-apply/impl.md` | Policies named `iso-42001-2023-…` — two standards collide in one namespace |
| `catalog/impl.md` | Claimed 42001 was the only standard; ignored `Available`, so GDPR/HIPAA/SOC 2/EU AI Act resolved as usable with no bundle behind them |

Fix: every pack-scoped command is `<packId>`-driven, every template renders `<packName>`, both resolved from `catalog list` at runtime. A new standard needs no doc change.

Also fixed while auditing: the `state list` template rendered `<packName>`, but that response carries no display name — now joins to `catalog list`.

## Testing

Checked against the live alpha catalog: field casing, `Available` semantics, whether two standards can be active on one tenant at once (they can), and 27001's six products.

Partial apply derives a policy name from each clause id (`A.6.2.8` → `a628`). Two clauses producing the same name would mean applying one silently overwrites the other, so names must be unique per pack.

Both confirmed against live catalog data: ISO 27001 16/16 unique, ISO 42001 15/15 unique.

(Checked initially against `fixtures/catalog-iso-42001.json`, which turned out to be a stale v3.0.0 capture missing three clauses — `A.6.2.5`, `A.6.2.6`, `A.8.3`. Re-run against live v4.0.0; those three yield `a625`, `a626`, `a83` and collide with nothing. The fixture is referenced by no task and is worth deleting or refreshing separately.)

`coder-eval`, 17 tasks, live tenant: **15 SUCCESS / 2 FAILURE**. Both failures are `llm_judge` returning `judge transport unconfigured`; every deterministic criterion passes. Details in the comment below.

All static gates green: `skills:validate` · `skills:build` · `skills:test` (48/48) · `check-skill-status` · `validate-skill-descriptions` · `check-cli-verbs` · `check-skill-verbs` · `check-task-driver`.

## Draft status

This PR edits the `description` in `SKILL.md`. That field is what decides whether the governance skill gets loaded for a given user request, so changing it can change which requests reach this skill at all.

CI checks that with `activation-gate.yml`: 65 sample prompts that should load this skill. All 65 currently do. The job fails if fewer than 59 do.

I added five standard names to the description and removed none, so no existing trigger phrase was taken away — but extra words can still weaken the match for prompts that already worked. At 100% the number can only stay flat or fall.

That job is skipped on drafts, and can't be run locally (it needs Bedrock credentials and a specific pinned model). **Marking this ready for review is what runs it for the first time.**

If it comes back under 59, the two fixes are: trim the description (the five added standard names are the only new tokens), or re-baseline `BASELINES_PCT` in `tests/scripts/activation_gate.py` after a fresh full activation run, which is the path that file documents.

Generated with Claude Code


sauravrathore01:
## Regression — 17 tasks, live tenant

**15 SUCCESS / 2 FAILURE.** No behavioral regressions.

Both failures are the same criterion, `llm_judge` → `(judge transport unconfigured)`:
- `unnamed-standard` (new, 0.769) — all 4 deterministic criteria pass
- `disambiguation` (pre-existing, 0.727) — all 3 deterministic criteria pass

Tasks covering what changed all scored 1.000: `partial-apply` (naming + copy), `state-list` (`<packName>` join), `query-clause` (un-hardcoded `catalog get`), `disable`, `org-scope`.

Two low-but-passing scores are advisory criteria (`pass_threshold: 0`), not regressions: `full-apply-e2e-live` 0.787, `interactive-values` 0.444.

`cleanup_compliance_pack.py`'s rewrite verified live — enumerated all six packs from `catalog list`, tolerated `404` on the unavailable ones, deleted a stray policy.

## Two follow-ups this run surfaced

**1. Topic-scope policy names — fixed here (cd33693b).** The live tenant had a policy named `iso-42001-2023-pii-ai-trust-layer`. `pii` was in no documented token set, because `planning.md` supports scoping by category keyword (traceability, guardrails/PII, model governance, robot allowlists) while `impl.md` only defined tokens for product, clause, and impact scopes. The agent followed planning.md, found no token, and coined one. Since the policy name is what makes a re-apply update the existing policy rather than create a second, a coined token means duplicate policies across runs. Now pins one token per planning.md row.

I first filed this as out-of-scope. It wasn't — this PR rewrote that naming note and added a fallback for unknown *products* while leaving topic scopes uncovered.

**2. `fixtures/catalog-iso-42001.json` — still open.** A stale v3.0.0 capture, referenced by no task, missing three clauses that v4.0.0 added. It caused a wrong claim in this PR: the clause-token uniqueness check initially reported 12/12 for 42001 from this file, when live has 15 clauses. Re-verified against live (15/15 unique), so the PR's claim is now sound.

The hazard is that it's a real captured API response, so trusting it is the natural move — which is exactly what happened. Suggest deleting rather than refreshing, since nothing reads it and a stale capture that looks authoritative is worse than none. Left out of this PR as a maintainer call.

Generated with Claude Code

abhinavmahanti (approved):

jianjunwang2 (approved):

PR #2708, opened 2026-08-19
fix(uipath-automationhub): byte-upload docs + hallucination-proof the flows for uncontexted users

parth-patil-uipath:
## Why

Alpha users **without any AH context** hit dead ends publishing a business process and uploading/retrieving documents. Two causes:

1. The skill on main still says documents are **link-only** ("byte upload not usable yet; host the bytes first") — so agents stall or invent URLs. @abhiram-vad fixed this on `fix/ah-process` but it was never PR'd; his commit is **included here verbatim** (first commit, his authorship).
2. Everywhere the skill *assumes* context a fresh user doesn't have, the agent is left to guess. This PR closes each of those holes with recipes **verified against `open-api-service`/`tenant-service` source and the live-tested `uip ah` CLI**.

## What changed

| Gap (what an uncontexted agent did) | Fix |
|---|---|
| Documents byte-upload "not usable" → invented hosting/URLs | **Abhiram's fix**: `file` object (`file_name`/`mimetype`/`file_content`/`file_encoding:"base64"`), XOR with `embed_link`, 300mb cap |
| "Follow the download reference" named no endpoint → fabricated routes like `/documents/{id}/download` | Real route documented: **`GET /download/file/{file_id}`** (takes `file_id`, **not** `document_id`; raw bytes, save with `-o`); file-backed vs link-backed entries distinguished |
| `document_type_id: <int>` with no source → guessed ids | Fixed platform table shipped (PDD=1, SDD=2, … MISC=9, from `tenant-service` `file.constants.js`) |
| Category recipe read ids off an existing process → impossible on a fresh tenant | Resolve via **`GET /hierarchy`** (works with zero processes) |
| "Owner/submitter must be real AH users" with no discovery → guessed emails, 400 `Cannot identify owner by email` | Resolve via **`GET /users`**, default to the signed-in user, confirm via `AskUserQuestion` |
| Missing inputs discovered one 400 at a time | New intake step: collect all six required inputs **before** the first POST |
| Success reported without checking | Verify-then-report: read the documents list back; never claim an attach that isn't in it |

Also: the skill description now triggers on download/fetch-document intents.

## Verification

- Endpoint routes confirmed in `open-api-service/src/api/v1/routes/open-api.routes.ts` (`/hierarchy`, `/users`, `/download/file/:id`, documents XOR handler).
- Document-type ids from `tenant-service/app/constants/file.constants.js` (`FILE_TYPES_ID_MAPPING`).
- Field shapes match the live-verified `uip ah documents list/download` (merged CLI).
- `npm run skills:validate`, `check-skills-sh.py`, `check-skill-status.py`, `check-skill-verbs.py` all green.

https://github.com/user-attachments/assets/a368f34c-6be9-4f18-827e-990fdbd1ce1d

Screenshot 2026-08-19 at 3 28 38 PM
Screenshot 2026-08-19 at 3 28 17 PM

Next step (tracked in RPANAV-18952): pressure-test through the Delegate in alpha with a no-context prompt.


parth-patil-uipath:
Review addressed in cb9e7636 — every point verified independently before changing anything:

**Must-fix 1 (201 nesting) — you're right, fixed.** Traced it in source: the controller unwraps the backend payload and rewraps via `PublicInfo.infoCreated("Resource Created", 201, data)` → the created object is under `data`. Both `api-endpoints.md` and `publish-process.md` Step 5 now say read **`data.process_id`** (and `data.document_id` for documents), plus a new guardrail: *a 201 means it WAS created — never re-POST on an undefined field read* (that's the duplicate-409 path you predicted). Ironic indeed — that line was main's own hallucination and the PR inherited it.

**Must-fix 2 (DSD label) — fixed.** Table now matches `file_types.sql` verbatim: *Development Specification Document*, singular *Input File*/*Output File*, *"Misc."*.

**TCD — annotated** rather than dropped: marked special (`zip`/`ssp` uploads only, **no embed_link**), so an agent picking 10 for a .docx gets warned by the doc instead of by a 400.

**/users and /hierarchy — spot-checked live on alpha just now**, exact envelopes documented: `data.users[]` with `user_email`/`user_is_active`, and `data.levels` + `data.categories[]`. The live capture also surfaced one more trap worth having: categories carry **`category_is_active`** and archived (0) nodes exist on real tenants — the category recipe now says pick active nodes only.

Thanks for the thorough pass — the Aug 13 captures were exactly the right bar to hold this to.

magickaichen (approved):
LGTM, make sure to attach the screen recording to the PR before this merges.

And looks like the demoed session is still reading context from user-level memory. We also need to cover cases where users have fresh sessions without any stored memory.

Try either turning off user memories in the session/app, or explicitly telling agents not to use memories in chat (whatever is feasible).

abhiram-vad (approved):

PR #2640, opened 2026-08-17
feat(planner): rewire case consumers to references/case/, retire the rules guide

abhiram-vad:
## What

PR 2/2 (stacked on #2639). The atomic flip: every consumer reads the new `references/case/` tree, and the monolith dies.

- **`case-design-lane-guide.md`** (314 → 284 lines, flow-only): §Tenant grounding, §Confirm's eight-section spec, and §Resolution ledger become pointers to `case/grounding.md` and `case/review.md`; the read budget names which file loads at which activity; the write-cadence sentence now states both halves explicitly (in-memory before the accept answer, write-early after).
- **`case-sdd-template.md`** — pure skeleton: the embedded authoring-rules block becomes links (two template-local rules stay); `min` restored in both SLA unit cells; the Tasks-file row names the build's `tasks/tasks.md`.
- **`case-sdd-examples.md`** — pure examples: notation/rule blocks become links to `variables.md`; Use Case 5 now frames when a Case Variables row is justified vs referencing the output directly; the "task-local variables" section teaches direct references as the default; the trailing checklist points at `review.md`.
- **`SKILL.md`** — the case navigation row lists all nine files by concern; Rule 13 names the new content-contract home.
- **`audit_sdd.py`** — now parses `model.md`'s three canonical tables and adds deterministic pre-build checks: task-type enum, WHEN ↔ Marks-Complete pairing legality, display-name charset. Smoke-tested: all three fire on seeded defects; graceful skip when model.md is absent. Docs and checker read the same source — they cannot diverge.
- **Deleted:** `case-authoring-rules-guide.md` (1,040 lines, 7 interleaved concern types, internally self-contradictory in two places). Zero dangling references (two test docstrings updated — comments only, no assertion changes).

## Verification

- All links + anchors across the planner case surfaces resolve (programmatic sweep, GitHub slug rules)
- `npm run skills:validate` passes both flavors
- Planner case eval suites dispatched on this branch — results will be posted here before merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2639, opened 2026-08-17
feat(planner): case knowledge as nine single-purpose reference files

abhiram-vad:
## What

PR 1/2 of the planner case-knowledge restructure. Adds `skills/uipath-planner/references/case/` — nine files, ~1,220 lines, replacing the 1,040-line `case-authoring-rules-guide.md` monolith (retired in PR 2/2) whose content was interleaved across 7 concern types and restated key rules up to 7× internally.

| File | Lines | Owns |
|---|---|---|
| model.md | 197 | Node kinds, the closed 9-value task-type enum, lifecycle gates + Marks-Complete pairing, exit types, secondary-stage semantics, sequencing grammar, edges-retired, naming rules |
| variables.md | 104 | Direct output references (the default wiring), the Case Variables declaration test, categories, outputs grammar, lineage closure, expressions |
| slas.md | 76 | SLA surfaces, entry requirements, breached/at-risk selection, the five-response model, silence defaults, validate-verified behavior matrix |
| principles.md | 113 | Authority hierarchy, render policy (`—`/`<UNRESOLVED>`/banned), domain fidelity, provenance, review items |
| authoring.md | 158 | Process→model method, other-path sweep, execution patterns, task typing + compliance overrides, XOR terminal patterns |
| render-case-definition.md | 127 | SDD §1 cell contracts (metadata, SLA Response Map, triggers, exits, variables table) |
| render-stages-tasks.md | 144 | SDD §2 cell contracts per task type, binding-cell grammar, I/O completeness |
| grounding.md | 87 | Design-time tenant resolution, the ONE batched gate, the resolution record |
| review.md | 217 | The eight-section Case Review, the 7-check stage-graph walk, the 20-check finalization list, the architect's lens |

## Design

- **One fact, one home.** Every rule is stated once; other files link (all links + anchors verified programmatically with GitHub slug rules).
- **Flow-aligned loading.** A session reads `model.md` once at lane entry; every other file loads when its activity starts — sketching, wiring data, rendering a section, confirming.
- **Deterministic claims.** Behavior-sensitive rules carry `(verified on uip 1.198.0-preview.102)` anchors with the exact validate messages quoted (e.g. the `wait-for-user` ↔ `user-selected-stage` pairing errors, the required-flag vacuity errors).
- **Conflicts resolved at the source.** The old guide's internal contradictions (e.g. `wait-for-user` legality on completion rows; the `jsonSchema` type both banned and mandated) are ruled once here, per live validation behavior.
[...]

PR #2709, opened 2026-08-19
fix(planner): reconcile the case render contracts with the template, and gate the checks that were missing

roberts-cliff:
Four commits from the 08/18 bug bash, one per root cause, each with a mechanical guard so the fix cannot silently regress. Fixes **D1, D2, D5, D6, D9** of the twelve findings, plus three defects found in the verification code while fixing them.

The through-line: **every one of the twelve bug-bash defects passed every existing gate.** The template conformance gate passed. The Finalization checklist passed. Structural counts came out clean. All twelve were found by a human noticing that two documents contradicted each other. That is the thing these commits try to make impossible rather than merely detectable.

---

## `706c4bbde` — connection lookup needs `--all-folders` (D5)

`uip is connections list --output json` searches only the default Integration Service scope. On a tenant that holds connections it returns:

```json
"Data": [],
"Instructions": "No connections found for any connector. Scope: default
                 Integration Service scope. Use --all-folders ..."
```

Measured on `popoc/DefaultTenant`: **0 with the default scope, 251 with `--all-folders`.** Outlook 365 alone has 59.

`grounding.md` said "check enabled connections" without naming the flag, so an agent following it sees zero and buckets every connector **Empty** instead of **Ambiguous**. Not a cosmetic mislabel — the two buckets offer the user *different options* at the resolution gate (`Empty` → "create during build"; `Ambiguous` → "pick a match"), so the missing flag silently changes the question the user is asked, and can offer to create a connector that already exists 59 times.

The scopes are also **not schema-compatible**: `--all-folders` returns PascalCase (`Id`, `Name`, `ConnectorKey`, `State`). Code parsing camelCase reads every field as absent. Both facts now in the reference.

> **The postmortem asked whether D5 was case-specific. It is not.** The guard caught the same un-scoped invocation in `uipath-rpa`, `uipath-api-workflow`, `uipath-platform/agent-workflow.md`, and `uipath-troubleshoot`. Those have different CODEOWNERS and are **deliberately not fixed here** — widening this commit into their references would make it unreviewable. The guard names them as follow-up.

## `41d3d2cd7` — the template is the shape, the references are the semantics (D1, D2, D6, D9)

**This is the systemic one.** #2639/#2640 split case knowledge into nine `references/case/*.md` files. Three of them describe the rendered SDD's shape — but so does `case-sdd-template.md`, and so does the conformance gate. Nobody said which wins, and `review.md` check 2 cites *both*, which forwards the conflict instead of resolving it.

An agent following the docs faithfully then **must** violate one of them. Five collisions, measured:

| where | claim | reality |
|---|---|---|
| `render-stages-tasks.md:9,10` | `` ### Stage {N}: {Name} (`{stage_id}`) `` — "MUST appear" | template and gate have no id |
| `render-stages-tasks.md:49` | `# \| Task ID \| Task \| Type \| Owner` | template has 9 columns |
| `render-case-definition.md:85` | `Field \| Operator \| Value \| Literal?` | "Literal" appears **nowhere** in the template |
| `render-case-definition.md:93` | `... \| Exit Type \| ...` | template has `THEN`, no Exit Type |

**Resolution — one authority per concern.** The template owns shape; the references own semantics. Shape assertions removed from both render contracts, replaced with pointers.

Two shapes existed only in a reference, so the **template gains them** rather than losing the capability:

- **`Exit Type` on Case Exit Conditions.** The template was the wrong one here, with independent evidence: `_shared/sdd_check.py` already parses that header for an `exit type` column. The template never offered it, so exit semantics (`exit-only` / `wait-for-user`) had nowhere to be authored.
- **`### Trigger Filter`.** Shape lived only in the reference; the template had no such section, so a filtered trigger could not be rendered conformantly.

**`stage_id` suffix: dropped.** Template and gate already agreed without it, so dropping is a one-file change where keeping it means editing the template, the gate, and every stage-naming cell. Stage display names are unique per case so cross-references still resolve; they are no longer greppable *by id*, which is the cost. **Reversible in `render-stages-tasks.md` alone — say the word if that trade is wrong.**

**Guard:** `scripts/check-case-render-authority.py`, wired as a pytest so it gates PRs.

## `3ddc9307a` — grade generated SDDs on shape, and fix the grader's own two bugs

The design-lane evals graded a generated SDD with substring checks: does a file exist, do these words appear, are there ≥6 stage headings. That is presence, not correctness — a 2,220-line SDD that invented its own Section 4 table format passes all four. `sdd_check.py` existed but was wired only into the three `case_finalize_*` tasks, never the two that exercise the design lane.

Wiring it in required fixing two grader bugs first, **both of which made it fail on correct input** — worse than not running it, because the first false positive is the one that ends a grader's credibility:

1. **Rows were split on escaped pipes.** A JS `||` must be written `\|\|` in a markdown table, so splitting on every `|` invented phantom cells and shifted every column right. On a real SDD it produced two confident, wrong findings (`invalid exit-type '\'`, and a legality error from reading the wrong column) on a **legal** table. Any SDD using a logical OR in an IF guard hit this — common in the mutually-exclusive completion/divert pairs the model itself prescribes.
2. **The finder globbed only `sdd.md`.** Direct design writes `<case-name>-sdd.md`; a draft writes `sdd.draft.md`. So it exited "no sdd.md found" before reading anything, and wiring it in unfixed would have produced a criterion that could never pass.

With both fixed it grades the bug-bash SDD clean: 39 variables, 33 `=vars` references resolve, lineage closes, 8 stages with legal conditions.

**First contact found real defects: 4 of 6 generated SDDs are mechanically unsound** — invisible to the substring checks.

## `c2b2c9b13` — gate a run as evidence, and make the authority check gate PRs

**`scripts/check-eval-run.py`.** A coder-eval run can finish with plausible scores and mean nothing. Three shapes observed in one day, each read as a regression:

- the skill was never installed (silent fallback to the built-in experiment, no `plugins:` block)
- the model never executed (expired credit ends the dialog; a finished `task.json` is still written)
- the judge had no transport (`llm_judge` scores 0.0, so every task is FAILURE regardless of merit)

Execution is measured by **output tokens, not turn count**. `codex` reports `total_assistant_turns == 1` on fully successful runs and `2` on runs that died on a provider error, so a turn threshold is near-inverted for it — a first draft used `turns >= 3` and rejected a known-good N=10 terra run 10/10. Verified against **six real archived runs** (3 good, 3 bad): correct 6/6, each failing for the right reason. `--regrade` relaxes the agent assertions for `coder-eval evaluate` (which runs no agent) while keeping the judge check.

**Authority check now gates PRs.** It shipped in `41d3d2cd7` as a script with no runner — the claim that it "would have failed the #2640 merge" was an overclaim, since nothing invoked it. A pytest wrapper gets it collected by the existing `test-helpers.yml` job, no workflow change, with a mutation guard that injects a bogus column list and asserts the checker names it.

---

## Guards are mutation-tested, not decorative

Every guard here was verified to **fail without its fix**:

| guard | reverted-fix result |
|---|---|
| `test_connection_scope_documented` | fails |
| `test_sdd_check_row_parsing` (4) | 3 of 4 fail; control still passes |
| `test_case_render_authority` | injected conflict detected and named |
| `test_check_eval_run` (9) | verified against 6 real runs, both directions |

One of those assertions **started out vacuous** — it checked for a message string the broken code never emitted, so it passed against the bug. Now requires evidence the file was actually read.

## Checks

```
tests/tasks/uipath-maestro-case/   109 passed
tests/tasks/uipath-planner/         74 passed
tests/tasks/uipath-maestro-flow/   181 passed
tests/tasks/uipath-agents/          17 passed
tests/scripts/ (4 guard jobs)       89 passed
                                   470 passed
skills:validate · check-skill-status · check-skills-sh · check-case-render-authority — all OK
```

Pre-existing and **not from this branch**: `pytest tests/` run wholesale hits 2 collection errors (`ModuleNotFoundError: _shared.case_check` in `e2e_expense_runnable` and `sla_with_escalation`). `git diff origin/main..HEAD` touches none of those files, and CI runs each suite separately, which is why the per-suite runs are clean. Worth a separate fix.

## Not in this PR

- **Cause 2** (D3, D7, D8, D11) — platform gaps the docs neither solve nor flag: an `adhoc` task cannot be waited on; one SLA status maps to one Response Map row; a condition-based SLA duplicates its responses per override; SLA units are calendar, not business days. These need product decisions, not a doc pass. The cheap half — an explicit "cannot be modelled" section — is worth doing next.
- **Cause 3** (D4, D10, D12) — under-specified references: the `=js:` evaluation sandbox, resolution-gate overflow past 4 questions, mid-run mode change.
- **`audit_sdd.py:280`** — a live guardrail bug found while testing: it picks Marks-Complete as *"the first cell containing Yes or No"*, which on a Stage Entry table is the `Interrupting` cell. It reported three illegal rule pairings in a correct SDD. It is also wired into only 1 of 4 design entry modes. A gate nobody runs, wrong when run. Wants its own commit.

## Reviewers

@abhiram-vad @RaduAna-Maria — `skills/uipath-planner/` is yours per CODEOWNERS:77. @abhiram-vad, this reconciles the render contracts from #2639/#2640; the `stage_id` question is the one real design call and I would value your view, since the greppable-cross-reference benefit was presumably deliberate.

### DIFF
--- a/scripts/check-case-render-authority.py
+++ b/scripts/check-case-render-authority.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python3
+"""Fail when the case render contracts and the SDD template disagree about SHAPE.
+
+Authority split (declared in `references/case/principles.md`):
+
+    the TEMPLATE is the shape      — headings, column lists, section numbering
+    the references are SEMANTICS   — what belongs in a cell, and why
+
+Both currently describe shape, with no tiebreaker, so an agent that satisfies one
+must violate the other. The existing gates cannot see this: the template conformance
+gate checks that headings EXIST, never that two documents AGREE, so contradictions
+pass every check and ship.
+
+What this catches:
+
+  1. A stage/task heading format asserted in a reference that the template does not use.
+  2. A pipe-delimited column list asserted in a reference with no matching table
+     header in the template.
+
+Read-only. Exit 0 clean, 1 on findings.
+
+    python3 scripts/check-case-render-authority.py [--json]
+"""
+from __future__ import annotations
+
+import argparse
+import json
+import pathlib
+import re
+import sys
+
+REPO = pathlib.Path(__file__).resolve().parent.parent
+TEMPLATE = REPO / "skills/uipath-planner/assets/templates/case-sdd-template.md"
+RENDER_DOCS = [
+    REPO / "skills/uipath-planner/references/case/render-stages-tasks.md",
+    REPO / "skills/uipath-planner/references/case/render-case-definition.md",
+]
+
+# A heading shape claim: a backticked `### Stage {N}: {Stage Name}`-ish string.
+HEADING_CLAIM = re.compile(r"`(#{3,5}\s+[^`]+)`")
+# A column-list claim: a backticked run containing at least two pipes.
+COLUMN_CLAIM = re.compile(r"`([^`\n]*\|[^`\n]*\|[^`\n]*)`")
+# A rendered markdown table header row in the template.
+TABLE_HEADER = re.compile(r"^\|(.+)\|\s*$", re.M)
+
+# Placeholder syntaxes differ between the docs ({N} / <N>) without being a conflict.
+PLACEHOLDER = re.compile(r"[<{][^<>{}]*[>}]")
+
+
+def normalize_heading(s: str) -> str:
+    s = s.strip().rstrip(".")
+    s = PLACEHOLDER.sub("*", s)
+    s = re.sub(r"\\`|`", "", s)
+    return re.sub(r"\s+", " ", s).strip()
+
+
+OPTIONAL_COL = re.compile(r"\(\s*\|[^)]*\)")
+
+
+def normalize_columns(s: str) -> tuple[str, ...]:
+    # `Field | Type | Binding (| Required)` documents an OPTIONAL column; the
+    # parenthetical is notation, not a cell. Treating it as one produced a false
+    # conflict against a template that defines both variants.
+    s = OPTIONAL_COL.sub("", s)
+    cells = [c.strip().strip("`").strip() for c in s.strip().strip("|").split("|")]
+    cells = [PLACEHOLDER.sub("*", c) for c in cells if c]
+    return tuple(c.lower() for c in cells)
+
+
+def template_facts(text: str) -> tuple[set[str], set[tuple[str, ...]]]:
+    headings = {
+        normalize_heading(line)
+        for line in text.splitlines()
+        if re.match(r"^#{3,5}\s+\S", line)
+    }
+    # The template also pins some heading shapes in HTML comments / inline backticks
+    # (secondary-stage task numbering, for one). Those are definitions too — not
+    # counting them made the checker report a conflict that does not exist.
+    headings |= {normalize_heading(m) for m in HEADING_CLAIM.findall(text)}
+    columns: set[tuple[str, ...]] = set()
+    for m in TABLE_HEADER.finditer(text):
+        row = m.group(1)
+        if set(row.replace("|", "").strip()) <= set("-: "):
+            continue  # separator row
+        cols = normalize_columns(row)
+        if len(cols) >= 2:
+            columns.add(cols)
+    return headings, columns
+
+
+def scan(doc: pathlib.Path, headings: set[str], columns: set[tuple[str, ...]]) -> list[dict]:
+    findings: list[dict] = []
+    if not doc.is_file():
+        return [{"file": str(doc), "line": 0, "kind": "missing", "claim": "", "detail": "file not found"}]
+    text = doc.read_text(encoding="utf-8")
+
+    for m in HEADING_CLAIM.finditer(text):
+        claim = normalize_heading(m.group(1))
+        if not claim.startswith("#"):
+            continue
+        if claim in headings:
+            continue
+        # Only flag stage/task headings — those are the shape the template owns.
+        if not re.search(r"\b(Stage|Task)\b", claim):
+            continue
+        findings.append({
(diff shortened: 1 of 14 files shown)


roberts-cliff:
Parking this as draft — superseded in part, and blocked on a decision that is not mine.

**Why it cannot merge as-is.** #2689 reverted #2639, which deleted `references/case/*`. Three of this PR's four commits edit files that no longer exist:

| commit | status |
|---|---|
| `706c4bbde` connection scope (D5) | edits `case/grounding.md` — deleted |
| `41d3d2cd7` render authority (D1/D2/D6/D9) | edits both `render-*.md` — deleted |
| `3ddc9307a` SDD grading + grader fixes | survives |
| `c2b2c9b13` run-is-evidence gate | survives |

**The surviving half is now #2710**, rebuilt fresh on `01f86d40a` and scoped to the eval harness only — no skill content, so it is independent of however the case-knowledge layout settles.

**What stays parked here:** the `--all-folders` connection fix and the render-contract/template authority reconciliation. Both are still real — the five shape collisions were measured, and the `--all-folders` behaviour was measured at 0 vs 251 connections on a live tenant — but they target a file layout that is currently in flux, and reconciling against a layout that may change again is wasted work.

**Separately, and more urgent than this PR:** `origin/main` currently carries **50 dangling relative links across 5 files**. #2689 reverted #2639 (which added `references/case/*`) but not #2640 (which rewired every consumer to point at them), so `SKILL.md`, `case-design-lane-guide.md`, `case-sdd-template.md`, `case-sdd-examples.md`, and `sdd-generation-guide.md` all link into files that no longer exist. Anyone running a case design against main today follows those links into nothing. That needs whoever owns the layout decision, not me guessing at it.

PR #2710, opened 2026-08-19
test(planner): grade generated SDDs on shape, fix the grader, gate runs as evidence

roberts-cliff:
Grades generated SDDs on shape instead of substring presence, fixes two bugs in the grader that made it fail on correct input, and adds a gate that verifies a coder-eval run actually measured something.

Scoped to the eval harness — no skill content.

---

## Generated SDDs were graded by substring

The two design-lane evals checked a generated SDD like this:

```
ls *sdd*.md | grep -q .                    a file exists
grep -Eiq 'withdrawn' 'rejected' ...       these words appear
grep -cE '^#+ +Stage [0-9]' >= 6           there are >= 6 stage headings
```

Presence, not correctness. A 2,220-line SDD that invents its own table format passes all four.

`sdd_check.py` — lineage closure, per-gate rule legality, task-type enum, entry/exit legality — already existed, but was wired only into the three `case_finalize_*` tasks and never into the two that exercise the design lane. Now wired into both.

**First contact found real defects: 4 of 6 generated SDDs are mechanically unsound.** None of it visible to the substring checks.

## The grader had two bugs, both failing on CORRECT input

These had to be fixed before wiring it in. A grader that cries wolf is worse than no grader — the first false positive is the one that ends its credibility.

**1. Rows were split on escaped pipes.** A JS `||` must be written `\|\|` inside a markdown table, so splitting on every `|` invented phantom cells and shifted every column right. Against a real SDD it produced two confident, wrong findings on a **legal** table:

```
invalid exit-type '\'                             <- landed on a lone backslash
'required-tasks-completed' is not legal for
  stage-exit exit (Marks=No)                      <- graded the wrong column
```

Any SDD using a logical OR in an `IF` guard hit this — common in the mutually-exclusive completion/divert pairs the case model prescribes.

**2. The finder globbed only `sdd.md`.** Build handoff writes `sdd.md`, but direct design writes `<case-name-kebab>-sdd.md` and a draft writes `sdd.draft.md`. So it exited *"no sdd.md found"* before reading anything — wiring it into a design task unfixed would have produced a criterion that **could never pass**.

With both fixed it grades a real SDD clean: 39 variables, 33 `=vars` references resolve, lineage closes, 8 stages with legal conditions.

`sdd_check.py` is a deliberate per-suite twin, and `test_shared_twins.py` caught that only one copy had been edited. Both updated.

## `scripts/check-eval-run.py` — is this run evidence?

A coder-eval run can complete, report per-task scores, and mean nothing. Three shapes observed in a single day, each initially read as a regression:

| failure | what it looks like |
|---|---|
| skill never installed | silent fallback to the built-in default experiment (no `plugins:` block) — the agent ran with no skill |
| model never executed | expired provider credit ends the dialog; a finished `task.json` with a score is still written |
| judge unreachable | `llm_judge` scores 0.0 with `(judge transport unconfigured)` — every task FAILURE regardless of merit |

The gate names each and exits 1, so a run is validated before anyone quotes it. It says nothing about whether the skill is correct — only whether the run is **evidence**.

**Execution is measured by output tokens, not turn count.** This matters: `codex` reports `total_assistant_turns == 1` on fully successful runs and `2` on runs that died on a provider error, so a turn threshold is near-inverted for it. A first draft used `turns >= 3` and **rejected a known-good N=10 terra run 10/10** — caught by running the gate against real archived runs rather than only synthetic fixtures.

`--regrade` drops the agent assertions for `coder-eval evaluate` runs, which execute no agent; asserting plugins/tokens/experiment there would fail every re-grade, and re-grading saved artifacts is the cheapest useful path. It keeps the judge check, so the cheap path is not where a transport gap hides.

## Guards are mutation-tested, not decorative

| guard | reverted-fix result |
|---|---|
| `test_sdd_check_row_parsing` (4) | 3 of 4 fail; the control still passes |
| `test_check_eval_run` (9) | verified against 6 real archived runs (3 good, 3 bad) — correct 6/6, each failing for the right reason |

One of those assertions **started out vacuous** — it checked for a message string the broken code never emitted, so it passed against the bug. It now requires evidence the file was actually read.

## Checks

```
tests/tasks/uipath-planner/       68 passed
tests/tasks/uipath-maestro-case/ 109 passed
skills:validate                   OK (26 skills, default + studioweb)
check-skill-status                OK
```

### DIFF
--- a/scripts/check-eval-run.py
+++ b/scripts/check-eval-run.py
@@ -22,7 +22,7 @@ number from it.
 Usage
 -----
     python3 scripts/check-eval-run.py RUN_DIR [--expect-experiment NAME]
-                                              [--min-turns N] [--json]
+                                              [--min-output-tokens N] [--json]
 
     # typical: the repo's own experiment must have been used
     python3 scripts/check-eval-run.py tests/runs/my-run --expect-experiment skill-tests-default
@@ -40,7 +40,7 @@ import sys
 JUDGE_UNCONFIGURED = "judge transport unconfigured"
 
 
-def replicate_findings(task_json: pathlib.Path, min_turns: int, regrade: bool = False) -> list[str]:
+def replicate_findings(task_json: pathlib.Path, min_output_tokens: int, regrade: bool = False) -> list[str]:
     """Reasons this replicate is not evidence. Empty list == trustworthy."""
     out: list[str] = []
     try:
@@ -84,7 +84,7 @@ def replicate_findings(task_json: pathlib.Path, min_turns: int, regrade: bool = 
             f"(turns={turns}, agent={t.get('agent_type')}). Usually a provider error "
             "(expired credit, auth); check conversation.log"
         )
-    elif min_turns > 0 and produced < min_turns:
+    elif min_output_tokens > 0 and produced < min_output_tokens:
         out.append(f"only {produced} output token(s) — suspiciously little work")
 
     for crit in t.get("success_criteria_results") or []:
@@ -104,7 +104,7 @@ def main() -> int:
     ap.add_argument("run_dir", type=pathlib.Path)
     ap.add_argument("--expect-experiment", default=None,
                     help="fail unless the run's experiment report names this experiment")
-    ap.add_argument("--min-output-tokens", dest="min_turns", type=int, default=1,
+    ap.add_argument("--min-output-tokens", type=int, default=1,
                     help="minimum model output tokens for a replicate to count (default 1)")
     ap.add_argument("--regrade", action="store_true",
                     help="the run came from `coder-eval evaluate` (no agent executed): "
@@ -123,7 +123,8 @@ def main() -> int:
     if args.regrade:
         report["experiment"] = "(re-grade — no experiment)"
     elif exp_md.is_file():
-        first = exp_md.read_text(encoding="utf-8").splitlines()[0] if exp_md.read_text(encoding="utf-8") else ""
+        content = exp_md.read_text(encoding="utf-8")
+        first = content.splitlines()[0] if content else ""
         name = first.split(":", 1)[-1].strip() if ":" in first else first.strip()
         report["experiment"] = name
         if args.expect_experiment and name != args.expect_experiment:
@@ -143,7 +144,7 @@ def main() -> int:
 
     bad = 0
     for tj in task_jsons:
-        findings = replicate_findings(tj, args.min_turns, regrade=args.regrade)
+        findings = replicate_findings(tj, args.min_output_tokens, regrade=args.regrade)
         rel = tj.relative_to(run_dir).parent
         report["replicates"].append({"replicate": str(rel), "findings": findings})
         bad += bool(findings)
--- a/scripts/check-eval-run.py
+++ b/scripts/check-eval-run.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python3
+"""Assert a coder-eval run measured what it claims, BEFORE anyone reads its scores.
+
+Why this exists
+---------------
+A coder-eval run can complete, report per-task scores, and mean nothing. Observed
+failures, all of which produced plausible numbers:
+
+  * the skill under test was never installed — the run silently fell back to
+    coder-eval's built-in default experiment, which has no `plugins:` block, so the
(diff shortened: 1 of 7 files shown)


roberts-cliff:
Both review items fixed in `1f5e7420a`, and I ran the coder-evals the lint bot asked for. One correction to my own PR body falls out of it.

## Review nits — fixed

- **`dest="min_turns"`** — vestige of the turn-count approach this PR replaces. The comparison was correct but the name lied about what it measured. Renamed throughout; `--min-output-tokens` now stores as `min_output_tokens`.
- **Double read of `experiment.md`** — collapsed to one read into a local.

Gate still correct on the same 6 real archived runs; `test_check_eval_run` 9 passed.

## Passing-run evidence

Ran both modified tasks, N=3, `claude-code` / `claude-sonnet-5`. Judge criteria skipped locally (no `ANTHROPIC_API_KEY` / Bedrock on this machine), so these numbers reflect the deterministic criteria only. **Instrument verified before reading any score** — `skill-tests-default`, plugin loaded, 6/6 replicates trustworthy.

| task | `sdd_check` | replicate findings |
|---|---|---|
| `loan_origination` | **1/3** | `00` `required-tasks-completed` + Marks=No ×2 · `01` `selected-stage-exited` + Marks Case Complete=Yes ×2 · `02` clean |
| `candidate_interview` | **1/3** | `00` seven WHEN↔Marks pairing errors across 6 stages · `01` 2 variables consumed but never produced, plus pairing errors · `02` clean |

## Correction: my "4 of 6 unsound" figure was wrong

It was measured badly. My triage passed an explicit path to `sdd_check`, and its glob picked the task's **input** `sdd.draft.md` — deliberately non-conformant — instead of the finalized output. Re-graded the way the harness actually does it (cwd = sandbox, no path argument, so the checker uses its own discovery order):

```
finalize_draft        3/3 PASS   <- I had reported these as FAILING
candidate_interview   1/3 PASS
loan_origination      1/3 PASS
```

So the design-lane pass rate is **~33%, not the ~67% I implied** — worse, not better. And `finalize_draft` is already at the bar; only the design lane is not.

## Are the failures genuine, or grader over-strictness?

The review asked this directly, and it decides whether a weight-3.0 criterion should gate CI. Answer: **genuine.** Every finding above is a WHEN ↔ Marks-Complete pairing violation or open lineage — both explicitly forbidden by the authoring rules (`Marks Complete: Yes` pairs only with `required-*`; `selected-*` is for `No`). I hand-verified the clearest case:

```
line 811   | envelopeId | -> docusignEnvelopeId |              extracts into it
line 846   | envelopeId | string | =vars.docusignEnvelopeId |  consumes it
Case Variables: 0 declarations
```

Outputs rows wire existing variables; they do not declare them. That is open lineage, and it is not the task-output exemption (which covers referencing a task's own auto-emitted field, not extracting into a named case var).

**One candidate for over-strictness**, flagged rather than defended: `invalid exit-type 'exit-only (→ Rejected)'`. The SDD annotated the exit-type cell with its destination. The destination is carried by `exitToStageId`, so the annotation is informal rather than wrong, and the grader rejects it as a non-enum value. Worth a maintainer's call.

## So: should this gate CI at weight 3.0?

Stating the trade rather than deciding it. The criterion finds real defects, **and** turning it on makes both design tasks red on current `main` (1/3 each). Three options:

1. **Land as-is** — the tasks go red until the design lane improves. Honest, and the failures are real.
2. **Advisory first** (`pass_threshold: 0`) — visible in reports, does not block, flip to gating once the rate comes up.
3. **Fix the defects first**, then land the criterion green.

I lean **2**, since it captures the signal immediately without a red baseline nobody scheduled work for — but it is the planner owners' call, not mine.

## The gate caught my own run

Worth recording, since it was the first real use. My first attempt at these numbers was **refused**:

```
default/skill-case-phase-0-loan-origination/00..02:
  - no plugin loaded — agent_config.plugins is empty
experiment: default          (expected skill-tests-default)
FAIL: this run is not evidence for 3/3 replicate(s)
```

My local wrapper was not passing `-e`, so coder-eval fell back to its built-in experiment, which has no `plugins:` block — the run measured a bare agent with no skill installed and still produced scores. Without the gate I would have posted that as the passing-run claim. Exactly the failure mode `check-eval-run.py` exists for, caught on its author.

PR #2711, opened 2026-08-19
chore(tests): drop expected_turns from every task YAML

bai-uipath:
The consumer-side cleanup for [Proposal: optimizing for wall clock time instead of expected turns](https://uipath.atlassian.net/wiki/spaces/CA/pages/91019149594/Proposal+optimizing+for+wall+clock+time+instead+of+expected+turns). The producer is [coder_eval_uipath#86](https://github.com/UiPath/coder_eval_uipath/pull/86), which derives each task's expected wall clock from run history and stamps it into `run.json`; the evalboard half is [coder_eval#125](https://github.com/UiPath/coder_eval/pull/125).

`expected_turns` is removed from all 935 task YAMLs that declared it. Nothing reads it any more: the expected line is now derived per task and per harness from the runs already in blob storage, so there is no number to type and none to go stale.

## Why the field is not worth keeping

| Problem | Evidence |
|---|---|
| It was never maintained | 147 commits in this repo touched `expected_turns`; six revised a value |
| The seed was a best-case extreme | Bulk-seeded from the luckiest run ever observed, so 44.7% of passing budgeted tasks exceed budget on a healthy night |
| Post-seed values are guesses | 49% are multiples of 5, and 204 tasks share a copy-pasted `300` against a p90 actual of 52, so they are permanently green |
| Coverage holes sat on the slow tasks | 249 of 1002 tasks had no budget at all, and they held 34.7% of task-seconds |
| A turn is not a unit of time | A `Read` and a 20-minute deploy both count as one |
| Not comparable across harnesses | On identical tasks Codex uses 0.50x Claude's turns |

## Scope

Only `expected_turns` goes. `max_turns`, `turn_timeout` and `task_timeout` are untouched, so every real cap and timeout in the suite is unchanged. The 239 `run_limits` blocks whose only key was `expected_turns` are removed outright rather than left empty. Eight comments that justified a turn budget were rewritten or dropped so no prose is left pointing at a field that no longer exists.

Safe to merge in any order relative to the two PRs above: `expected_turns` stays accepted and ignored on `RunLimits`, and a task without it reads as unscored rather than erroring.

## Testing done

- Every one of the 1191 task YAMLs parses and validates against `TaskDefinition`, and the same 12 tasks fail validation before and after this change, all for reasons unrelated to `run_limits` (the local venv's `coder_eval` predates `skill_triggered`, advisory zero weights, and `sandbox.record_cli`).
- `coder-eval plan` over the full suite: 1191 tasks planned, the same 12 pre-existing errors, no `run_limits` or `expected_turns` diagnostics.
- Machine-checked that the diff is semantics-preserving: each of the 935 files was loaded before and after and compared as parsed YAML, with the only permitted difference being the dropped key and the dropped empty block. Zero mismatches.
- No file gained a blank-line run or lost its trailing newline.


joe-prosser (approved):

bai-uipath:
Closing this PR as we'll keep this field for now.

PR #2718, opened 2026-08-20
refactor(planner): case knowledge as three files — interview, layers, template-as-contract — with a mechanized SDD validator

abhiram-vad:
## Why

#2689 reverted the nine case reference files (#2639) but not their consumer rewiring (#2640). Main before this PR:

- **30 dangling links** to `references/case/` from SKILL.md, the lane guide, the template, examples, and sdd-generation-guide
- **`audit_sdd.py` silently degraded** — `load_model_facts()` returns `{}` when `model.md` is absent, so the task-type enum, WHEN-pairing, and naming checks no-op with no error
- No case content contract at all: the nine files gone, the 1,040-line rules guide they replaced retired by #2640

## What — commit by commit

**1. `2fb0e34` — restore + restructure (four files).** The nine files' knowledge reorganized by consumption: interview lane / design layers / template / spec+validator. Zero broken links; validator parity with the nine-file baseline.

**2. `fd811b1` — mechanize.** Contract enforcement moves from prose into `audit_sdd.py`: gate-slot WHEN legality (stage-entry and task-entry tables were previously unchecked), Exit Type × Marks-Complete, `sla-status-change` title closure, name uniqueness, typed recipients, dead button routes, scaffold markers. Kills a **pre-existing false positive** (the WHEN×Marks check read entry tables' `Interrupting` column as Marks-Complete) and aligns the Buttons rule with shipped reality (the published golden maps buttons to the action task's own output).

**3. `f0de100` — dissolve the spec; three files.** The spec was structural redundancy — its render half re-described the template, its checking half re-described the layers guide and the script (the skeleton existed in 3 copies; "a task entry interrupts nothing" verbatim in 3 files). Representation now follows consumption:

| File | Role |
|---|---|
| `case-design-lane-guide.md` | Interview flow + tenant grounding + authoring policy + review items + resolution record + the Case Review |
| `case-design-layers-guide.md` | Design knowledge in four layers (skeleton → gates → data → time) + naming rules (validator-parsed) + ONE closure checklist (musts ∪ walk ∪ finalization ∪ advisory lens, deduplicated) |
| `case-sdd-template.md` | THE render contract — cell rules inline at the governing cell + Validation footer (markers, gate, enforcement list). Gains two sections it was missing: **SLA Response Map** (graded by CI checkers) and **Trigger Filter** |
| `scripts/audit_sdd.py` | Every deterministic check — zero context cost |

Found in the merge: the spec had already drifted from the template inside this one PR (Task-ID column, case-exit Exit-Type column — in neither the template nor any published golden).

**4. `d57221e` — PO.Frontend parity + eval move.** Mapped all 45 FE validation modules (`src/services/validation/case-mgmt`) to design-time homes. New deterministic checks: `case-entered` presence, empty condition tables, ≥1 trigger, **evaluation-precedence rules** (stage entry identical to a case-exit row = unreachable stage; unguarded exit sharing a guarded completion's WHEN = stage never completes — previously in neither validator nor docs), self-referencing entries, duplicate case-exit rows, vacuous `required-*`, SLA bounds. `skill-case-sla-sdd-response-map` moves to the planner suite with its checker + unit tests (planner is the sole case-SDD author; the task grades `sdd.md`, no build).

## FE rule coverage

| FE rule family | Design-time home |
|---|---|
| Structure (stages/triggers/tasks/conditions present, case-entered, names, duplicates) | `audit_sdd.py` |
| Rule legality (gate slots, pairing, exit types, selectors, SLA references) | `audit_sdd.py` (canonical tables in the layers guide) |
| Precedence & liveness (entry-vs-case-exit overlap, exit-overrides-completion, self-reference, vacuous required-*, SLA bounds) | `audit_sdd.py` + layers § Lifecycle gates rule 6 |
| Recipients/escalation completeness, expression discipline | layers/lane prose + architect's-lens review items (judgment or AST-level) |
| Build mechanics (ids, adhoc-case kind, connector event wiring) | `uipath-maestro-case` — intentionally not SDD-level |

## Size & verification

| State | Case doc set |
|---|---|
| Monolith era | 253 KB (~63k tok) |
| Nine-file (reverted) | 185 KB (~46k) |
| Main before PR (broken) | 100 KB (~25k) |
| **This PR** | **~159 KB markdown (−14% vs reverted), and the separate render-time spec read is gone** |

- Both published goldens `AUDIT OK`; a real defect surfaced in the 0817 build (no completing case-exit row); seeded-defect SDDs yield exactly the expected findings (12/12 + FE-parity seeds)
- 0 broken links/anchors; 0 stale references repo-wide; `skills:validate` + `skills:build` pass both flavors; moved task's 16/16 unit tests pass; twins guard intact

## Eval run — final: 11/12 green

12-task SDD-generation set (planner `case_design_*`, `case_finalize_draft*`, `case_sla_response_map`, `pdd_to_sdd`, `e2e_rpa_sdd`, `asdd_crosswalk`, maestro-case `design_delegation_*`), claude-sonnet-5 on the GH runner:

- **9/12 first run** ([32414845933](https://github.com/UiPath/skills/actions/runs/32414845933))
- **`case_finalize_draft`: SUCCESS 1.0 in 868s** after fixing the naming trap the first run exposed (the validator applied the minting charset to draft-carried em-dash names the naming contract forbids renaming — unwinnable repair loop; `--draft` now threads carried names through as exempt)
- **`case_sla_response_map`: SUCCESS 1.0 in 412s** — the first sonnet pass this task has ever had. A/B against main showed pre-PR runs were fast-but-wrong (513s, score 0.25, no SLA Response Map section — it existed only in the spec file the agent never read; this PR moved it into the template). Turn budget raised to 3600 for observed thinking-turn variance ({412s, >1800, >1800, >2400} draws)
- **`e2e_rpa_sdd`: pre-existing flake, not this PR** — agent left `<SOLUTION_DIR> <OUTPUT_DIR>` literals from the RPA template's Next Steps; RPA template untouched here. Tracked separately.

## Follow-up (separate PR)

Cross-skill drift guard: twin contract block + CI parity check vs `uipath-maestro-case`.


charlesliu9:
- P1 skills/uipath-planner/scripts/audit_sdd.py:345 — table_rows() does not recognize the Case Triggers header (| T# | Trigger Type | ... |) as a header. Consequently, skills/uipath-planner/scripts/audit_sdd.py:510 sees that header as a data row and passes an SDD with no actual trigger. Add T# to the header exclusions or make header detection structural.

  - P2 tests/tasks/uipath-planner/_shared/sla_response_map_check.py:178 — SLA-response-map closure matches sla-status-change entries by SLA title only, ignoring the call’s target and the map’s Scope. A case-level and stage-level SLA may legally share a title, so a map row for the wrong target can satisfy the new evaluator. Match root to case and a stage target to stage: <name> as well as matching title.

roberts-cliff (approved):
:shipit:

charlesliu9 (approved):

skills/uipath-planner/assets/templates/case-sdd-examples.md
```diff
@@ -514,5 +514,18 @@ Full marker semantics: the build skill's `bindings-and-expressions.md` § In-exp
...
+AP Review exits to an SLA Escalation lane on `requiresEscalation`, else completes ([layers § Secondary-lane entry shapes](../../references/case-design-layers-guide.md#secondary-lane-entry-shapes)):
+
+| Stage | Condition | WHEN | IF | Exit Type | Marks Complete |
+|---|---|---|---|---|---|
+| AP Review | exit (complete) | `required-tasks-completed` | `=js:(vars.requiresEscalation !== true)` | `exit-only` | Yes |
+| AP Review | exit (divert) | `selected-tasks-completed("AP ownership review")` | `=js:(vars.requiresEscalation === true)` | `exit-only` (exit target → SLA Escalation) | No |
+| SLA Escalation | entry (`Interrupting: Yes`) | `selected-stage-exited("AP Review")` | `=js:(vars.requiresEscalation === true)` | — | — |
+| SLA Escalation | exit | `required-tasks-completed` | — | `return-to-origin` | Yes |
+
+On escalate the divert fires (completion's inverse `IF` is false), the lane runs, `return-to-origin` re-activates AP Review; on non-escalate the completion fires and the next stage enters via its own `selected-stage-completed("AP Review")`. The decision is read directly from the producing action's output — never relayed through a §1.5 variable.
```
RaduAna-Maria: Is this really necessary?

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -118,6 +124,26 @@ Skip entirely when no review items are open. Blocking = yes keeps Planner Handof
 | At-Risk | <percentage>% of SLA duration | Notify: <recipient or group> | <non-empty root-unique escalation title, no `:`> |
 | Breached | 100% of SLA duration | Notify: <recipient or group> | <non-empty root-unique escalation title, no `:`> |
```
RaduAna-Maria: Put the necessary information directly in the table rather than having it in comments.

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -118,6 +124,26 @@ Skip entirely when no review items are open. Blocking = yes keeps Planner Handof
...
+start-task, a stage-entry row for enter-stage, a stage-exit / case-exit row), and every sla-status-change
+row in the SDD has a row here. Interrupting: `—` for notify-only AND every start-task (a task entry
+interrupts nothing — never Yes/No); otherwise Yes/No matching the produced stage-entry row. Breach cells
+that enter a lane read `enter-stage: <Secondary Stage Name>`, never `Notify: <role>`. Default when the
+source states no response: both statuses notify-only, Target and Interrupting `—` — never invent a stage,
+task, or routing change for a notification. -->
+
+| Scope | SLA | Status | Response | Target | Interrupting | Rationale |
+|-------|-----|--------|----------|--------|--------------|-----------|
+| <case \| stage: <StageName> \| task: <TaskName>> | <that target's SLA Title> | <At-Risk \| Breached> | <notify-only \| start-task \| enter-stage \| exit-stage \| exit-case> | <— for notify-only; task name; stage name; produced exit row> | <— \| Yes \| No> | <why this response fits the source> |
```
RaduAna-Maria: Same as above

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -130,15 +156,33 @@ Skip entirely when no review items are open. Blocking = yes keeps Planner Handof
...
+<!-- Only when ≥1 trigger declares a filter; omit otherwise. AND/OR tree; nested {op, clauses} groups
+flatten into rows. Operators PascalCase, case-sensitive: Equals, NotEquals, Contains, NotContains,
+StartsWith, EndsWith, GreaterThan, GreaterThanOrEqual, LessThan, LessThanOrEqual, In, NotIn, IsNull,
+IsNotNull. Avoid `Literal: No` for unverified runtime expressions — lossy fallback; prefer literal values
+or a review item. -->
+
+| Field | Operator | Value | Literal? |
+|-------|----------|-------|----------|
+| <payload field> | <operator> | <value> | <Yes \| No> |
+
```
RaduAna-Maria, line 173: Same

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -496,4 +572,44 @@ Implementation tasks **do not live in this SDD** — they live in the planner's
 
 ---
 
+<!-- ============================ VALIDATION (do not render) ============================
```
RaduAna-Maria, line 587: I am not sure about this. Rather than having a validation here and spending time and tokens editing the plan file due to case /name mismatches, leave the downstream skill handle and fill in the minor gaps

RaduAna-Maria (changes requested):
Nice work overall! There are still parts that are not meant to be in the planning stage, like [this](https://github.com/UiPath/skills/pull/2718/changes#diff-8a1523a6306cafcea1ef8eb97a16918b219ad1ca320ebe620158071ef6d6fd7dR275) for example. I will also post an analysis from Claude in order to pin point the exact improvements.

skills/uipath-planner/scripts/audit_sdd.py
```diff
@@ -265,37 +279,429 @@ def model_findings(text: str) -> list[str]:
...
+    if row and row.group(1).strip() not in ("—", ""):
+        titles["root"] = {row.group(1).strip()}
+    for _, stage_name, block in stage_blocks(text):
+        found = {m.group(1).strip() for m in re.finditer(r"^\*\*SLA Title:\*\*\s*([^\n<]+)", block, re.M)}
+        if found:
+            titles[stage_name.casefold()] = found
+    return titles
+
+
+def contract_findings(text: str, facts: dict) -> list[str]:
```
RaduAna-Maria, line 417: I have doubts about the cost/benefit of this machinery. It's 850 lines of regex over markdown, it parses a prose guide at runtime, and it drives a mandatory generate-then-repair loop in every mode — and the PR's own eval history already shows the failure mode I'm worried about: the em-dash naming check produced an unwinnable repair loop that burned an entire eval run before `--draft` exempted it. Every false positive costs full repair turns, which is exactly the time and token waste the mechanization was supposed to remove. Please share the eval data on repair rounds actually needed and trim any check that risks false positives more than it catches real defects; correct-by-construction prose plus a lean audit beats a fat audit that agents iterate against.

skills/uipath-planner/scripts/audit_sdd.py
```diff
@@ -220,18 +222,19 @@ def lineage_findings(text: str) -> list[str]:
...
-    table (legal WHEN rules per Marks-complete value), and the `### Naming rules`
-    fenced regex. Returns {} when model.md is absent so the shape audit still runs.
+    Reads the `### Task types` table (column 1 literals), the `### Lifecycle
+    gates` table (legal WHEN rules per gate slot), and the `### Naming rules`
+    fenced regex — all from case-design-layers-guide.md. Returns {} when the
+    layers guide is absent so the shape audit still runs.
     """
     try:
-        text = MODEL_MD.read_text(encoding="utf-8")
+        text = LAYERS_MD.read_text(encoding="utf-8")
```
RaduAna-Maria, line 242: The PR description criticizes main because `load_model_facts()` returned `{}` silently and the checks no-op'd with no error — but this version keeps that exact failure mode: a missing file returns `{}` at L238–239 and an empty parse is downgraded to `{}` at L264–266. Restoring the guide fixed the trigger, not the mechanism; the next rename or reformat silently disables the model checks again. Emit a finding when the facts come back empty.

skills/uipath-planner/references/case-design-layers-guide.md
```diff
@@ -0,0 +1,422 @@
...
+|---|---|
+| External system, portal, form, inbound event, or record-created start | `Intsvc.EventTrigger` — a tenant object start stays an event trigger even when provisioning is missing (unresolved detail becomes a placeholder later; never downgrade to Manual) |
+| Schedule / recurring | `Intsvc.TimerTrigger` |
+| Otherwise (user or API call) | `Manual` |
+
+### Task types
+
+The enum is closed — exactly these nine literals, used verbatim as the SDD `Type:` value. The `type` says **how the work gets done**, not what it is about — read the verb + the actor:
+
+| Type | Pick when the work is… |
```
RaduAna-Maria, line 48: This table, the `### Lifecycle gates` table, and the Naming-rules regex are parsed at runtime by `audit_sdd.py`, which makes their markdown formatting load-bearing: renaming a heading or reshaping a column silently disables validation (see the `load_model_facts` comment). Add a guard comment on all three: `<!-- parsed by scripts/audit_sdd.py — do not rename this heading or reformat this table -->`.

skills/uipath-planner/references/case-design-layers-guide.md
```diff
@@ -0,0 +1,422 @@
...
+| `=jsonString:<json>` | JSON-as-string — connector `Operation Configuration` carry-through only |
+| `=datafabric.<path>` | Data Fabric reference |
+| `=orchestrator.JobAttachments` | File slot |
+| `=response` / `=result` / `=Error` | Conventional response handles |
+
+Bare field-name lists (`**Inputs:** a, b, c`) are forbidden — use the table with one form per cell.
+
+---
+
+## Layer 4 — Time: SLAs & escalations
```
RaduAna-Maria, line 334: This layer restates maestro-case's `sla-response-shapes.md` almost in full — the response table, the breach-vs-at-risk escalation semantics, the never-start-task-as-stage-entry rule, and the Verified table. That knowledge has to stay in maestro (brownfield needs it), so this copy is permanent double maintenance. Reduce this layer to the decision surface the design conversation actually needs and source the rest from the maestro-case skill; `sla-response-shapes.md:3` also still claims the planner links to it instead of restating it, which stopped being true — fix that in this PR too.

skills/uipath-planner/references/case-design-layers-guide.md
```diff
@@ -0,0 +1,422 @@
...
+
+Never author `start-task` as a stage-entry row on the breached stage: it validates, but stage re-entry re-runs every task whose `Run Only Once` is `No` — a breach meant to add one manager check silently re-runs the whole stage.
+
+### Defaults when the source is silent
+
+- No stated response → both statuses `notify-only`. Never invent a stage, task, or routing change.
+- At-risk threshold: SLA ≤ 3 days → 75%; 3–10 days → 70%; > 10 days → 80%.
+- Recipients: at-risk → the owner persona's user group; breached → the leadership tier (Compliance for regulation-driven cases). Record substituted defaults with provenance.
+
+### Verified behavior (uip 1.198.0-preview.102)
```
RaduAna-Maria: "verified on uip 1.198.0-preview.102" appears five times in this file and the whole table duplicates maestro's § Verified. Version-pinned probe evidence goes stale and is for skill authors, not something an agent needs on every read — keep the rules, move the evidence into the tests or the PR description.

skills/uipath-planner/references/case-design-layers-guide.md
```diff
@@ -0,0 +1,422 @@
...
+
+### Expressions
+
+- `=js:`-prefixed JavaScript. Namespaces available to `=js:` evaluation: `vars`, `response`, `bindings`, `iterator`, `metadata`. Assignment operators are forbidden in every case expression.
+- A rule's condition expression gates CASE STATE only (`vars.*`, `metadata.*`) — there is no `event` namespace. In-rule extract-then-gate does not work at runtime (the gate evaluates before the extract writes): extract `response.field -> caseVar` on the connector rule and gate a DOWNSTREAM condition instead.
+- Use strict equality (`===` / `!==`); write mutually exclusive branch guards as exact inverses so completion and divert rows cannot dual-fire.
+- Thresholded policy ("Credit Analyst only over $5M") lands in an executable cell — owner/recipient, WHEN/IF, or a task input — with the numeral written out (`5000000`), actor and attribute on one line. Prose or a persona-table mention alone is a render failure.
```
RaduAna-Maria, line 303: This rule is currently stated in five prose locations (here, closure item 6, lane guide L144 and L146, §Resumption step 9, template footer item 9) on top of the `unencoded_thresholds()` check. Make this bullet plus the audit check the single home and reduce the other mentions to pointers.

skills/uipath-planner/references/case-design-layers-guide.md
```diff
@@ -0,0 +1,422 @@
...
+Never the designer's `any` escalation sentinel. Borrowed and dangling references fail validate: `The escalation referenced by rule … no longer exists` / `The SLA referenced by rule … no longer exists`.
+
+### Choosing the response
+
+Pick from the source's words — WHERE the work lives, never whether it interrupts. A named task never justifies a new stage.
+
+| Response | Source says | What you author | Interrupting cell |
+|---|---|---|---|
+| `notify-only` | notify / alert / page someone, nothing more | An escalation on the target's SLA rules — no stage, task, or condition | `n/a` |
+| `start-task` | Follow-up work inside the SAME breached stage ("as part of the review", a named task for a manager or peer) | One task in the breached stage carrying `sla-status-change` as its OWN task-entry row, against that stage's (or the case's) SLA | `—` — a task entry interrupts nothing; never `Yes`/`No` |
```
RaduAna-Maria, line 373: This exact phrase — the one the PR description cites as the three-copy defect the spec dissolution removed — still has three verbatim copies in the planner: here, the lane guide's Case Review section 4, and the template's Response Map comment. Keep it here and in the template's render comment; the lane guide copy should point instead.

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -118,6 +124,26 @@ Skip entirely when no review items are open. Blocking = yes keeps Planner Handof
 | At-Risk | <percentage>% of SLA duration | Notify: <recipient or group> | <non-empty root-unique escalation title, no `:`> |
 | Breached | 100% of SLA duration | Notify: <recipient or group> | <non-empty root-unique escalation title, no `:`> |
 
+<!-- Required when Case SLA is set; both rows concrete, no `—`. Breached recipient = one tier up
+(leadership; Compliance for regulation-driven cases). Display Name defaults to `Escalation Rule {N}` only
+when no sla-status-change row references it; defaulted recipients get provenance `default applied`. -->
+
+### SLA Response Map
```
RaduAna-Maria, line 130: The eval A/B justifies adding this section, but as written the same SLA-response facts now render three times per SDD — the per-stage SLA tables, this map, and the Case Review — with a both-ways closure check policing the self-duplication. Make the map the single authored source of responses and let the stage SLA cells carry only titles and durations.

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -332,8 +400,16 @@ Template-local rules:
 **Resolved Resource:** <concrete intended resource name; never <UNRESOLVED>>
 **Folder Path:** <resolved exact folder path or <UNRESOLVED>>
 **Resource Identity:** <apiWorkflowId / agentId / processOrchestrationId (+version) or <UNRESOLVED>>
-**Binding Sub-Type:** <Api \| Agent \| ProcessOrchestration \| —>
-**Dispatch / Operation:** <selector and value for shared facades, or —>
+**Binding Sub-Type:** <Api (api-workflow) \| Agent (agent) \| ProcessOrchestration (process) \| — (rpa) — omitting it makes Studio Web report the resource as not found>
```
RaduAna-Maria: The sub-type is derivable 1:1 from the task type, and "omitting it makes Studio Web report the resource as not found" is build/IDE mechanics — exactly what the lane guide says stays build-side. Drop the cell and make its presence a Phase 3 emission check in the build skill.

skills/uipath-planner/assets/templates/case-sdd-template.md
```diff
@@ -496,4 +572,44 @@ Implementation tasks **do not live in this SDD** — they live in the planner's
...
+
+Gate: run  python3 "<skill folder>/scripts/audit_sdd.py" <sdd path> [--draft <draft path>]  on the
+on-disk file BEFORE the Status: ready flip — in every mode. RUN it, never open the script source —
+its findings are the interface. Minting charset applies to names THIS run minted; names carried from
+a draft are preserved verbatim and exempt (pass --draft so the validator knows), ':' banned always. Repair findings with Edit, re-run to AUDIT OK
+(max 3 rounds, then stop and present findings). Never ship a summary SDD (top-level headings like
+## Source / ## Case Objective / ## Stages / ## Task Plan, or build-mode/path narration) even if a later
+caseplan.json would validate — rewrite from the model and this template.
+
+The validator enforces (manual fallback when python3 is unavailable — every item must hold):
```
RaduAna-Maria: Items 1–9 are a prose copy of the validator, paid on every template read, to cover the rare case where python3 is absent. Shrink them to one-line check names pointing at the script, and have the invocation try `python`/`py` before falling back to manual verification — Windows machines usually don't have a `python3` alias.

skills/uipath-planner/references/case-design-lane-guide.md
```diff
@@ -2,13 +2,13 @@
...
+**Draft finalization budget (hard):** read this file's finalize sections (§Resumption, §Terminal step) plus `sdd.draft.md` and the template (its validation footer is the gate) — once each; work from those reads. Scripts are RUN, never read: invoke `audit_sdd.py` directly and act on its findings — opening the script source is a budget violation. The normalization contract (numeric `Task S{K}.{M}` headings, `**Task envelope**` markers, plain `<UNRESOLVED>` markers) lives in those sections — skipping them ships the draft's defects into the final SDD. Do NOT open [case-design-layers-guide.md](case-design-layers-guide.md) or `case-sdd-examples.md`: finalization normalizes structure, it does not redesign. No subagents, no background tasks, no tenant discovery unless identities are needed and a session exists. Write via the §Terminal step write-early cadence.
```
RaduAna-Maria: The read-ban on the layers guide during finalization is what forces finalize rules to be copied into this file — the contract is now spread across §Read budget, the Entry-modes row, §Terminal step, the 14-step §Resumption, and the template footer. Since finalization already has to read the template, make the template footer the complete finalize contract and shrink §Resumption to what is genuinely specific to resuming from a draft.

skills/uipath-planner/references/case-design-lane-guide.md
```diff
@@ -77,70 +122,108 @@ When the user mentions `file`, `attachment`, `PDF`, `upload`, `evidence`, `recei
...
...ion footer) plus a reachability spot-check. In PROSE, reference the rule as bare `sla-status-change` — never a partial call form with placeholder args; build-side checkers reject wrong arity anywhere. A source rule that depends on a business attribute or threshold (for example, department- or seniority-gated eligibility) must be represented in an executable condition, output mapping, or guarded recipient/assignment expression, not only in a Design Rationale. When the prompt says every primary phase/stage has an SLA target, render each per the template's Stage SLA block (deterministic default titles are in its SLA Title cell) with concrete titles and at-risk/breach display names; every `sla-status-change` reference uses those exact titles, and per-stage SLA tables stay consistent with the SLA Response Map (breach cell reads `enter-stage: <Secondary Stage Name>`, never `Notify: <role>`). Do not iterate for polish; the later build run re-validates everything. This bound applies on top of the seed-first cadence (§Terminal step): AFTER the accept answer, a fresh design writes early and per-section. Before the accept answer the model stays in memory (no draft file, no checkpoint writes).
```
RaduAna-Maria: This is ~350 words in a single paragraph, and it fuses the durable contract (exact shape, concise content) with clauses that read like patches for specific eval prompts ("when the prompt says every primary phase/stage has an SLA target…"). Break it into numbered rules per the repo's content-quality standard and separate the eval-specific clauses so this kind of accretion stays visible — this is the same pattern that produced the 1,040-line guide this work retired.

skills/uipath-planner/assets/templates/case-sdd-examples.md
```diff
@@ -514,5 +514,18 @@ Full marker semantics: the build skill's `bindings-and-expressions.md` § In-exp
 ## Checks
 
 No checklist lives here. Variable and output row consistency is normative in
-[`variables.md`](../../references/case/variables.md); the pre-confirmation gate is
-[`review.md § Finalization`](../../references/case/review.md).
+[layers § Layer 3](../../references/case-design-layers-guide.md#layer-3--data-variables--expressions); the pre-confirmation gate is
+[layers § Layer closure](../../references/case-design-layers-guide.md#layer-closure--the-design-checklist).
+
+## Decision-routed return lane (divert + inverse guards)
```
RaduAna-Maria: The guards here read `vars.requiresEscalation` while the WHEN names the producing task, which is the stale-read anti-pattern if that's a declared §1.5 variable — it's only correct because it's the task's own output with no variable row, and since both cases are spelled `vars.X`, the table can't show which one applies. Add one line making it explicit: "`requiresEscalation` is the action's own output — no Case Variables row exists."

RaduAna-Maria (changes requested):
This PR fixes real breakage on main and the direction is right — the deterministic audit, the FE-parity checks, and the consumption-ordered restructure are all improvements, and the eval evidence backs them. But it also cements some structural problems that will only get more expensive to undo later, so I'd like the points below addressed in this PR rather than deferred.

**1. Duplication — the PR's own goal isn't met yet.** Commit `f0de100` promises "each rule stated once, at its consumption point," but several rules still appear in three to five places: the threshold-encoding rule has five prose homes plus the audit check, "a task entry interrupts nothing" has three verbatim copies, never-downgrade-to-Manual has three, and "scripts are RUN, never read" has four. Together this is roughly 15–20% of the ~130KB design read set. Please collapse each rule to one home and turn the other mentions into pointers — the inline comments identify them.

**2. Product semantics must have one home, and that home is `uipath-maestro-case`.** The layers guide's Layer 4 restates maestro's `sla-response-shapes.md` almost in full, the naming rules mirror maestro's Rule 25, and the "Verified" tables exist in both skills. Maestro cannot give this knowledge up — brownfield edits design deltas with no SDD and no planner — so keeping a second copy in the planner means permanent double maintenance. Restructure so the planner carries only the minimal decision surface the design conversation truly needs and obtains the rest through the same in-conversation delegation this PR already uses in the other direction. This also sets the precedent before BPMN/Flow lanes copy the pattern: the case pack is already 206KB — 29% of the planner — and repeating that per product turns the planner into a monolith.

**3. The interface between the two skills is described twice, in full, on both sides.** Maestro's Rule 1 restates the lane's entire eight-section Case Review ("Case Review" appears 11 times in that SKILL.md), so every lane change needs manual syncing in the sibling skill. Maestro only needs its own obligations — when to hand off, the receipt spot-check, and where it resumes. Please trim its side in this PR as well.

**4. Build mechanics don't belong in the design surface.** Some of these pre-date this PR, but this PR owns the case contract now, so clean them up here: the on-disk `serviceType` enum note (`case-sdd-template.md:158` — the lane guide itself says that mapping is build-side), the BPMN-serialization rationale in the Default note (`:206` — keep the string-encode rule, drop the mechanism), the `Service Type` cell (`:350`) and `Binding Sub-Type` cell (`:403`) which are both 1:1 derivable from the task type, and Section 4's typed connector Operations and field lists, which ask design time to author schema facts it is forbidden to discover and that Phase 3 re-derives anyway — while Rule 2's "trust as written" locks in whatever was invented.

skills/uipath-planner/assets/templates/case-sdd-examples.md
abhiram-vad: Agreed — the whole file is deleted; its patterns are now a 13-row grammar table at the head of Layer 3.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Done — the escalation recipient rules moved into the table cells; the comment keeps only what a cell can't carry.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Done — the Response Map's Target and Interrupting rules are in the cells now.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Done — the full PascalCase operator list is in the Operator cell.

skills/uipath-planner/scripts/audit_sdd.py
abhiram-vad: Fixed — `load_model_facts()` returns a reason and the audit emits a "model checks disarmed" finding instead of silently no-op'ing.

skills/uipath-planner/references/case-design-layers-guide.md
abhiram-vad: Added the guard comment on all three parsed blocks (Task types, Lifecycle gates, Naming rules fence).

skills/uipath-planner/references/case-design-layers-guide.md
abhiram-vad: Removed — zero version pins left in the layers guide; the probe evidence lives in the tests and the PR.

skills/uipath-planner/references/case-design-layers-guide.md
abhiram-vad: Down to one canonical bullet plus two pointers (the closure item and the finalize scan).

skills/uipath-planner/references/case-design-layers-guide.md
abhiram-vad: The lane guide copy is a pointer now; layers and the template keep it, as you suggested.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Done — the map is the single authored source and the per-stage SLA blocks carry only title and duration.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Dropped the cell.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Shrunk to ten one-line check names, and the gate now tries `python3` → `python` → `py` before falling back to manual verification.

skills/uipath-planner/references/case-design-lane-guide.md
abhiram-vad: Broken into four numbered rules, and the eval-shaped clauses are deleted rather than kept alongside.

skills/uipath-planner/assets/templates/case-sdd-examples.md
abhiram-vad: Moot — the examples file is deleted.

skills/uipath-planner/references/case-design-lane-guide.md
abhiram-vad: Moved the finalize shape contract into the template's Validation footer (the file finalization already reads), so §Resumption now keeps only draft-specific repairs.

skills/uipath-planner/scripts/audit_sdd.py
abhiram-vad: Trimmed the false-positive class you flagged: the naming charset no longer gates (advisory only), leaving `:` as the sole blocking name rule — 40 of 1787 corpus findings were user-supplied business names like `Credit & Document Verification`.

skills/uipath-planner/assets/templates/case-sdd-template.md
abhiram-vad: Keeping this one: a built case failed at runtime with "RequestBody not found, did you mean request_body", and the check only fires when a Section 2 read path contradicts the same SDD's own Section 4 list.

skills/uipath-planner/references/case-design-layers-guide.md
abhiram-vad: The repo rule (skills must work with siblings absent) makes this duplication mandatory, so it's now a declared canonical+twin pinned by `test_case_twin_parity.py`, and the stale claim in `sla-response-shapes.md:3` is fixed.

abhiram-vad:
P1 and P2 are both fixed on the branch: header detection is structural now (`T#` included), and the response-map closure matches scope to target instead of SLA title alone.

skills/uipath-planner/SKILL.md
```diff
@@ -46,24 +44,22 @@ The skill has three paths, decided by the **Entry Guard**:
...
-   references/case-design-lane-guide.md). Route to Phase D — Design through
-   the product's lane when any of these fire (the path taxonomy stays the three
-   paths above; a lane is Phase D's product-specific entrance):
-   - a build request handed off by that product's build skill because its
-     design input is missing (same conversation) → the lane's Build
-     handoff mode,
-   - a design request for that product with no PDD → the lane, direct,
-   - the product's design draft exists (or is named) and the user asks to
-     finalize it → the lane, draft finalization.
+0. Case Management design signals — checked first. Any of these routes to
```
RaduAna-Maria: Revert this. Since this is a general planning skill it cannot check for Case Management as step 0.

skills/uipath-planner/assets/templates/case-sdd-template.md
RaduAna-Maria: If it fails at build/runtime then it is a `uipath-maestro-case` problem to solve, not something the planner should account for. In the planning step you need to decide what to build and how the structure would look like, the how mechanics are out of scope.

skills/uipath-maestro-case/references/sla-response-shapes.md
```diff
@@ -1,6 +1,8 @@
...
+Rules for turning an SLA at-risk / breach event into case behavior, on the BUILD side — SKILL.md Rule 21, [brownfield.md § SLA responses](brownfield.md#sla-responses-in-a-brownfield-edit), and the SLA / condition / task plugins all link here instead of restating it.
+
+> **Twin of a canonical page in `uipath-planner`.** The case SLA response model is owned by `uipath-planner`'s case design layers guide (Layer 4) — that is where the product's case knowledge and guidance live. This page exists because skills in this repo must function with siblings absent and may not read another skill's files, and because a brownfield edit designs deltas with no SDD and no planner in the loop. **It is a twin, not a second source of truth: a change to the response model lands in the planner first, then here, in the same PR.** The invariants both sides must agree on are pinned by `tests/tasks/uipath-planner/_shared/test_case_twin_parity.py`.
```
RaduAna-Maria: Please keep this file as is and keep it as the sole owner of those information. No need to have this detailed guidelines in the planner as well.

skills/uipath-planner/references/case-design-lane-guide.md
```diff
@@ -77,98 +106,141 @@ When the user mentions `file`, `attachment`, `PDF`, `upload`, `evidence`, `recei
...
 
-The exact rendered SDD text must pass this gate before it leaves the lane — in every mode against the **on-disk file assembled by the write-early cadence, before the `Status: ready` flip** (one structural Read is the check). This is a render check, not a second design review. Do not use the read to redesign the case.
+Start with `## Case Review: <Case name>`, then exactly this order:
 
-Required shape:
+1. **Case Snapshot** — `Item | Proposed design`. Rows: `Objective`, `Starts when`, `Primary personas`, `Successful completion`, `Other terminal outcomes`, `SLA coverage`. Assumed values marked `(assumed)`. No case ID prefix unless it affects a decision.
```
RaduAna-Maria, line 138: keep this as well ` `## Document History`, then the `## Planner Handoff` header + `` marker, then `## Table of Contents` — the universal planner scaffold (Rule 5); the case body follows.`

skills/uipath-planner/references/case-design-lane-guide.md
```diff
@@ -77,98 +106,141 @@ When the user mentions `file`, `attachment`, `PDF`, `upload`, `evidence`, `recei
...
-5. Report the path in one line, then fork by mode. **Build handoff:** do NOT stop — the Build answer already carried consent, so the build skill's phases start immediately in this conversation (`uip solution init` + its Phase 1, which verifies the resolved identities instead of re-discovering them). **Direct design / draft / finalization:** STOP — the SDD write is a turn boundary; task derivation (Lane A) or the build (`uipath-maestro-case`) continues on a later turn, opt-in, and `## Next Steps` in the written SDD points there.
+Mechanized by `audit_sdd.py` — the template's § Validation footer is the contract (document skeleton, per-block markers, forbidden summary-only sections). Run it against the **written file, before the `Status: ready` flip** — every mode; one structural Read is allowed to repair findings. This is a render check, not a second design review; on failure, rewrite from the model and template — never a summary SDD, even if a later `caseplan.json` would validate.
+
+### Terminal step — write the SDD
+
+On the accept answer: compose the complete document and write it in ONE Write, gate it, flip it. The mode decides only the filename and whether the turn ends.
```
RaduAna-Maria: Please change this to write in batches. This is the way the SKILL.md defines how the SDD should be written on disk in order to not lose anything in case of a mid turn failure. Having contradictory guide lines here might confuse the agent.

RaduAna-Maria (changes requested):

roberts-cliff (approved):

RaduAna-Maria (approved):

PR #2719, opened 2026-08-20
fix(uipath-maestro-flow): correct variableUpdates schema, loop ports, and loop output shape

rockymadden:
## Why

`skill-flow-loop-multiply` fails with `result: "NaN"` instead of 3315. Root-causing that run turned up four stale facts in the skill, one of which is a hard blocker.

The eval agent did the right thing: it read `loop/impl.md` and implemented the **"State Accumulation with variableUpdates"** section — the section that exists for exactly this task. The schema documented there is stale.

`variableUpdates[].expression` is documented as a `=js:` string. That form was dropped from the file format in schema version **1.3**; it is now an object `{ type, expression, fieldType }`. It is also the **only** strictly-typed expression site in the schema — `nodes[].inputs` and `nodes[].outputs[].source` are `Record<string, unknown>` and `edges[].data` is loose, so all of them still accept legacy `=js:` strings. That asymmetry is exactly why the stale docs read as correct. `flow init` scaffolds version 1.9, so the documented shape hard-fails:

```
[MIGRATION] Workflow migration failed at 1.9→1.10:
Offending field(s): variables.variableUpdates.<nodeId>.0.expression
"Retry": "RetryWillNotFix"
```

With `RetryWillNotFix` and no guidance, the agent burned ~8 turns, abandoned the accumulator pattern, and fell back to reducing `$vars.<loopId>.output` — whose element shape the skill never documented. Entries are keyed by body node id, not the body's bare return value, so `item.number` was `undefined` and `1 * undefined` → `NaN`. That variant **validates clean**, so nothing caught it.

This object form is not new to the skill — `plugins/agent/impl.md` already documents `{ type, expression, fieldType }` with a bare, unprefixed body for agent node inputs. The `variableUpdates` docs simply never got migrated.

## What changed

| Defect | Fix |
| --- | --- |
| `variableUpdates[].expression` documented as a `=js:` string (5 sites) | Object form everywhere, with the `fieldType` derivation rule and the "no `=js:` prefix *inside* the object" gotcha |
| Loop aggregate output element shape undocumented | New **Aggregated loop output** section — correct vs. incorrect reducer side by side, and a note that `flow validate` cannot catch the wrong accessor |
| Loop ports documented as `output` / `loopBack` | Registry exposes outer `input`/`success`/`error` and inner `start`/`continue`/`break`. There is no `output` port — `output` is an output *variable* |
| `currentIndex` (0-based) documented; doesn't exist | `currentIteration` (1-based), corrected in 5 places |

Also noted: `uip maestro flow variable-update add` still emits the legacy string form (fixed separately in UiPath/cli#3721), and unmapped `inout` globals raise `MISSING_OUTPUT_MAPPING`.

### The corrected patterns were executed, not just validated

Two flows scaffolded from `uip solution init` + `uip maestro flow init`, authored strictly per the corrected docs, then run with `uip maestro flow debug`:

**1. The documented accumulator pattern** (`inout` global + object-form `variableUpdates`) — `finalStatus: Completed`:

```json
{ "result": 3315,
  "loop1.output": [ { "script1": { "output": 13 } },
                    { "script1": { "output": 195 } },
                    { "script1": { "output": 3315 } } ] }
```

**2. Both reducers from the new docs section, in a single run** — body node returns `{ number: n }`, mirroring the failing artifact:

```json
"loop1.output": [ { "script1": { "output": { "number": 13 } } },
                  { "script1": { "output": { "number": 15 } } },
                  { "script1": { "output": { "number": 17 } } } ]
```
| accessor | result |
| --- | --- |
| `item.script1.output.number` (documented CORRECT) | **3315** |
| `item.number` (documented WRONG) | **`NaN`** |

Both runs confirm the aggregate shape independently of the original failing run. Both test solutions were deleted afterwards (`uip solution delete`).

### The schema boundary, per version

`variableUpdates[].expression` parsed against every versioned file schema, varying nothing else:

| version | `=js:` string | object form |
|---|---|---|
| 1.0–1.2 | ACCEPTED | ACCEPTED |
| **1.3–1.10** | **REJECTED** | ACCEPTED |

`CURRENT_WORKFLOW_VERSION = 1.10`, `MINIMUM_SUPPORTED = 1.9`.

### What `flow validate` does and does not enforce

| Case | Result |
| --- | --- |
| Legacy string form | `Failure` — the `RetryWillNotFix` migration error above |
| Object form | `Valid` |
| Old documented ports `output` / `loopBack` | `Failure` — `Edge references undeclared source handle "output" … rewire to one of: success, error, start` |
| Expression object missing a key / extra key / bad `type` | `Failure` — `[…expression] Invalid input` (strict) |
| `fieldType` disagreeing with the target variable | `Valid` — **not** enforced |
| `=js:` left inside the object `expression` | `Valid` — **not** enforced |
| `variables.nodes` bound to `outputId: "currentIndex"` | `Valid` — bindings aren't checked against declared outputs |

The last three are why there is a second commit. My first pass asserted that `fieldType` "must match" the target variable, that a stray `=js:` would produce a specific runtime error, and that `currentIndex` "resolves to `undefined`" — none of which validation enforces, and the runtime claim I had not observed. Those now state what is checkable, with the enforcement boundary called out, since writing unverified assertions into a reference doc is the failure this PR exists to fix.

Loop handles, `currentIteration`, and the aggregate-output description were read from `uip maestro flow registry get core.logic.loop`. `.maintenance/check-all.sh`, `npm run skills:validate`, and `npm run skills:build` pass; the 4 reported bad anchors are pre-existing on `main` (verified by stashing).

Docs-only — no test or skill-structure changes.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/references/editing-operations-json.md
+++ b/skills/uipath-maestro-flow/references/author/references/editing-operations-json.md
@@ -326,7 +326,7 @@ Use `Edit` to add an entry to `variables.variableUpdates.<NODE_ID>`:
 
 Only `inout` variables can be updated. `in` variables are read-only.
 
-> **`expression` is an object here, not a `=js:` string** — unlike `inputs` and End-node `outputs[].source`, which still accept the string form. `fieldType` must match the target variable's declared `type` (`integer` → `number`). The string form fails `flow validate` with `[MIGRATION] Workflow migration failed at 1.9→1.10 … Offending field(s): variables.variableUpdates.<NODE_ID>.0.expression`.
+> **`expression` is an object here, not a `=js:` string** — unlike `inputs` and End-node `outputs[].source`, which still accept the string form. Set `fieldType` to the target variable's declared `type` (`integer` → `number`); `flow validate` does not cross-check it against the variable, so a mismatch passes silently. The string form does not: it fails with `[MIGRATION] Workflow migration failed at 1.9→1.10 … Offending field(s): variables.variableUpdates.<NODE_ID>.0.expression`.
 
 > **Do not use `uip maestro flow variable-update add`** — it still writes the legacy string form and produces a file `flow validate` rejects. Use `Edit`.
 
--- a/skills/uipath-maestro-flow/references/author/references/editing-operations-json.md
+++ b/skills/uipath-maestro-flow/references/author/references/editing-operations-json.md
@@ -312,7 +312,11 @@ Use `Edit` to add an entry to `variables.variableUpdates.<NODE_ID>`:
       "<NODE_ID>": [
         {
           "variableId": "<INOUT_VARIABLE_ID>",
-          "expression": "=js:<EXPRESSION>"
+          "expression": {
+            "type": "jsExpression",
+            "expression": "<BARE_JS_NO_PREFIX>",
+            "fieldType": "<TARGET_VARIABLE_TYPE>"
+          }
         }
       ]
     }
@@ -321,6 +325,10 @@ Use `Edit` to add an entry to `variables.variableUpdates.<NODE_ID>`:
 ```
 
 Only `inout` variables can be updated. `in` variables are read-only.
+
+> **`expression` is an object here, not a `=js:` string** — unlike `inputs` and End-node `outputs[].source`, which still accept the string form. `fieldType` must match the target variable's declared `type` (`integer` → `number`). The string form fails `flow validate` with `[MIGRATION] Workflow migration failed at 1.9→1.10 … Offending field(s): variables.variableUpdates.<NODE_ID>.0.expression`.
+
+> **Do not use `uip maestro flow variable-update add`** — it still writes the legacy string form and produces a file `flow validate` rejects. Use `Edit`.
 
 ---
 
--- a/skills/uipath-maestro-flow/references/author/references/editing-operations.md
+++ b/skills/uipath-maestro-flow/references/author/references/editing-operations.md
@@ -84,7 +84,7 @@ These apply regardless of which strategy you use.
 
 - `targetPort` is required on every edge — validate rejects edges without it
 - See [file-format.md — Standard ports](../../shared/file-format.md) for port names by node type
-- Dynamic ports: decision (`true`/`false`), switch (`case-{id}`/`default`), HTTP (`branch-{id}`/`default`), loop (`output`/`success`/`loopBack`)
+- Dynamic ports: decision (`true`/`false`), switch (`case-{id}`/`default`), HTTP (`branch-{id}`/`default`), loop (`start`/`continue`/`break` inner, `success`/`error` outer)
 
 ### Validation
 
--- a/skills/uipath-maestro-flow/references/author/references/planning-arch.md
+++ b/skills/uipath-maestro-flow/references/author/references/planning-arch.md
@@ -229,7 +229,7 @@ Use this when defining edges. Every edge requires a `sourcePort` and `targetPort
 | `core.logic.delay` | `input` | `output` |
 | `core.logic.decision` | `input` | `true`, `false` |
 | `core.logic.switch` | `input` | `case-{id}` (dynamic per case), `default` |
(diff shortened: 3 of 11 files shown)


rockymadden:
Thanks — all three findings were real. Fixed in acf39ea and 0d3ac2c, each with a proof rather than a claim.

## 1 & 2. The two `=js:` remnants — confirmed, fixed

Both real, and my grep is why I missed them: I searched `"expression": "=js:` (JSON-key shaped), which never matches a fenced example or a prose mention.

- **Inside-loops note** — I fixed the twin sentence in `loop/impl.md` and missed this one. Now `$vars.bodyNode.output` as the object's `expression`.
- **Variable Update Expressions section** — now shows the object form and links back to the schema.

The `.concat()` example there was a *new* claim (`fieldType: "array"`), so I ran it rather than assuming. Array-typed `inout`, accumulated via the exact idiom in the doc, `finalStatus: Completed`:

```
collected = [26, 30, 34]     // numbers [13,15,17], body: $vars.items.concat([$vars.loop1.currentItem * 2])
```

## 3. The `loopBack` filter — real bug, but the causality is inverted

> Now that the skill teaches `continue` instead of `loopBack`, agent-generated flows will use `continue`, and these filters will fail

This is **pre-existing, not fallout from the doc change**. `loopBack` has never been a real handle — `flow validate` rejects it outright:

```
Edge references undeclared target handle "loopBack" on node "loop1" …
rewire to one of: input, continue, break
```

So every flow that ever reached these utilities already used `continue`. The filter has always been dead, the back edge always stayed in the graph, and reachability was always over-approximated:

```
producer 'body' inside a loop, asked for reachable End nodes
  old filter {loopBack}            -> {'end1'}   # walked body -> loop -> after -> end1
  new filter {continue,break,...}  -> set()      # correct: body only loops back
```

`successful_end_ids` would report an End reachable from a loop body, and the billing detector's mutual-unreachability check could call two branches connected *through* the loop. Fixed in both files via a shared `LOOP_BACK_PORTS = {"continue", "break", "loopBack"}` — keeping `loopBack` so any legacy fixture still filters.

### The 102 passing tests were not evidence

Worth flagging: none of the existing shared-advisory tests constructs a loop, so the suite could never have caught this and its green was meaningless here. Added two cases over `successful_end_ids` with real loop wiring. They **fail on the old filter and pass on the new**:

```
old filter -> AssertionError: assert {'end1'} == set()   (both new tests FAILED)
new filter -> 13 passed
```

## Scope correction

The PR is **no longer strictly docs-only** — it now touches two test utilities and adds test coverage. Called out in the commit message and the PR body. Happy to split the utility fix into its own PR if you'd rather keep this one to documentation; it's self-contained. Flagging explicitly that the fix changes advisory scoring behavior, since the filter goes from dead to live.

## On the CLI follow-up

You noted the "do not use `variable-update add`" warnings will need updating if the CLI is fixed. That fix is up as UiPath/cli#3721 — `add` now emits the object form, `list` reads both shapes. I'll drop those two warnings here once it merges.

`.maintenance/check-all.sh`, `skills:validate`, and `skills:build` still pass.

jiyangzh:
## Review — two optional nits (both pre-existing)

All four stale facts corrected, verified against the live registry. Cross-file consistency across all 8 doc files is clean. Grep sweep found no remaining stale `currentIndex`, `loopBack`, or legacy `=js:` string references. Advisory filter fix + tests are correct.

### 1. `variables-and-expressions.md:383` — stale summary sentence (low)

This line is outside the diff but in the same file:

```
Used for conditions, input values, variable updates, and output mappings.
```

After this PR, variable updates use the object form, not `=js:` strings. The detailed Variable Updates section (line 262) is correct — this is a stale summary. Suggest dropping "variable updates" from that list while you are in the file, since it is the same class of stale-doc issue the PR addresses.

### 2. `loop/impl.md:101` — code comment duplicates prose (low, pre-existing)

```js
// In a Script node inside the loop body (parentId must be set to the loop node)
```

The surrounding prose at line 99 and the table at line 109 already cover this. Token-optimization rule §2 says strip when prose already explains. Not introduced by this PR — noting for awareness only.

jiyangzh (approved):

PR #2725, opened 2026-08-20
feat(uipath-automationhub): CLI-first flows via uip ah, with transport preflight

parth-patil-uipath:
## Why

RPANAV-18939 / Kurt's "update the skill to use the CLI": now that the `uip ah` surface is merged (UiPath/cli#3602 stack), the skill's flows should run through it — auth handled entirely by the CLI (Delegate env-auth or `uip login`), uniform `Result`/`Code`/`Data` envelopes, no token handling, no curl.

## Design: additive + preflight-routed (safe to merge today)

**No released `uip` ships `ah` yet** (latest = 1.199.0, three days before the merge). So Step 0 runs `uip ah --help` once per session:
- succeeds → new **CLI flows** (`references/*-cli.md`, catalog in `cli-commands.md`)
- `unknown command 'ah'` → the existing **raw-API flows**, unchanged
- never mixed in one run

The skill therefore works on every install today and flips to the CLI automatically as `uip` updates. Deleting the API flows later is a one-commit cleanup.

All domain hardening from #2708 (schema-derived required set, tenant-required app questions, doc-type table, wrapping rules, verify-then-report, mandatory deep links with slug recovery) is carried into the CLI flows verbatim — only the transport changed.

## Validated live through the Delegate (both routing arms)

- **Fallback arm**: session on `uip` 1.199 — preflight detected the missing `ah`, announced it, completed the publish on the API path.
- **CLI arm**: session on a dev 1.201 build — full publish ran on `uip ah` only (~90 invocations, 13 distinct commands; zero curl executed, zero tokens anywhere in the transcript): `idea-flows list` → `schema get --idea-flow-id` → `categories get`/`users list`/`applications list` discovery → `automations create --from-schema --idea-flow-id` → 3× `documents create --file` (byte upload) → `documents list` verification → slug-based deep links. Result: process 18 on the alpha test tenant, 3 verified attachments, working links.

## Dependency

The CLI arm's `documents create --file` and `automations create --idea-flow-id` land in **UiPath/cli#3720** — needed in a released `uip` before the CLI arm activates for real users (the preflight keeps everyone safe meanwhile; the version note in `cli-commands.md` covers a CLI that has `ah` but predates #3720).


parth-patil-uipath:
## Review — verified live against `uip` 1.201.0 + AH source

Ran every documented command/flag against the released CLI (my install upgraded to 1.201.0, which ships the full `ah` surface) and checked the AH-side claims against `ah-co`/web-client source. Result: **the catalog is remarkably accurate — 13 of 14 commands verified exactly — but one command in Step 6b does not exist.**

### ❌ Must fix: `uip solution list` is not a command

```
$ uip solution list --help
{ "Result": "ValidationError", "Message": "error: unknown command 'list'", … }
```

The `solution` surface has `cleanup/delete/deploy/download/init/pack/packages/projects/…` — no `list`, and nothing that returns solutions with `Id`/`Name`/`Projects[]` by name-pattern. Nearest real commands:
- `uip solution packages list --name <pattern>` — server-side name filter, but returns **packages** (name/version/publish date), not solutions with projects
- `uip solution download <solution-id>` — needs the UUID already in hand (help says it's emitted by `solution upload` / `flow debug`)

So the Step 6b discovery recipe (`uip solution list --name → Id, Name, Projects[]`) will hit unknown-command and strand the agent — exactly the hallucination class this PR exists to kill, which is why I'd block on it. Options: rewrite discovery around what exists today (ask the user for the designer URL / solution id, which they have from Studio Web), or if a `solution list` ships in an upcoming CLI PR, gate the step with a version note naming it — same pattern as the #3720 note in `cli-commands.md`.

### ✅ Everything else verified

- **Commands/flags, all exact**: `automations create --from-schema --idea-flow-id --file` · `schema get --destination` · `documents create --title/--description/--document-type-id/--file⊕--embed-link` (the XOR is even in the CLI's own help text) · `documents download <file-id> --destination` (help confirms file-id, not document-id) · `categories get` · `users list --limit` · `applications list` · `automations get --all-fields` / `update --file` / `list --search` · `idea-flows list` · `components list --automation-id`
- **Envelope verified live**: `{Result, Code, Data}` with `Code: AhAuthInfoGet` matching the `Ah<Group><Verb>` pattern; `Data.Tenant.Url` + `Data.User` exactly as documented
- **Studio Web link shape confirmed in source** (`ah-co/src/helpers/co-calculated-value-type.helper.ts`): JSON-string value parsed to `{url, name, hasProcessMap}`, `name` falls back to `url`, `hasProcessMap` coerced boolean. (FYI the interface also has optional `groupId`/`groupName` — fine to omit.)
- **Deep link route confirmed**: `/automation-profile/:slug/documentation` exists in web-client routes
- The preflight design + the #3720 version-note pattern for flag-level drift is the right call, and the "one home per fact" extending-guidance is a genuinely good structure

### 🟡 Minor

1. **PR body is stale**: "No released `uip` ships `ah` yet (latest = 1.199.0)" — 1.201.0 is released **with** the full `ah` surface (all flags from the #3720 note included). Worth reframing the body and tightening the `cli-commands.md` version note to "requires `uip` ≥ 1.201.0".
2. `publish-process-cli.md` Step 6 restates `PDD → 1, SDD → 2, else 9` — a small fork of the fact the PR's own rule says lives only in `api-endpoints.md`. Pointer-only would be purer (the same step already links there).

parth-patil-uipath:
All review findings addressed in 08c4e704 — with one correction to the review itself:

**Blocker (`uip solution list` fabricated) — confirmed and fixed.** You're right, and the irony is noted: I sourced it from the cli repo's *scenario doc* without executing it — the exact failure mode this PR polices. Verified: `unknown command 'list'` even on a dev 1.201.0 build (the solution tool's listing surface is unshipped). Step 6b discovery is now **ask-the-user**: the caller supplies the designer URL (copied from Studio Web with the solution open), or a solution id + project id to build it from — never invented, never searched. The catalog row is gone.

**Correction on the version minor:** 1.201.0 is **not** a public release — the latest GitHub release is still **1.199.0** (Aug 11). The 1.201.0 on this machine is a local dev build linked into the Homebrew path for Delegate testing (that's also what your live runs executed, which is why `documents create --file` worked — it carries unmerged cli#3720). The version note now states this precisely: `ah` first appears in 1.201.0, no public release ships it yet, preflight covers older installs, and the two #3720 flags get their own unknown-flag guidance.

**Doc-type restatement — fixed** per the one-home rule: Step 6 now points at the `api-endpoints.md` table without repeating ids.

Also from the automated review: the Authentication section is retitled **"raw-API flows only"** with an explicit skip-note for the CLI path, and `components list --automation-id` (referenced by the get flow) now has its catalog row. The CLI-arm e2e coverage gap stands — that needs an `ah`-capable released `uip` in the test harness, same constraint as activating the arm itself; tracked with RPANAV-18939.

Live verification appreciated — it caught what doc-reading couldn't.

alena-apanovich-1812:
Reviewed the CLI-first conversion — solid, additive design that's safe to merge today (preflight keeps every current install on the working path), and nice catch removing the hallucinated `uip solution list` earlier. A few doc fixes worth making first:

**Suggestions**
- **SKILL.md** — the intro line "All flows share the Authentication section above and the endpoint catalog in `api-endpoints.md`" is now stale; CLI flows skip both. Reword to scope it to the raw-API flows.
- **Deep-link base URL (both `*-cli.md` flows)** — template is `{baseUrl}/{org}/{tenant}/…` but the only source given is `auth-info get → Data.Tenant.Url`. Unclear whether `Tenant.Url` already contains `{org}/{tenant}` (→ double-append) and where org/tenant come from on the CLI path. Please pin the exact `Tenant.Url` shape or spell out org/tenant resolution.
- **"`ah` present but pre-#3720" middle state** — preflight passes, then `automations create --idea-flow-id` / `documents create --file` fail mid-run. Docs say "stop (or use raw-API path)" which conflicts with "never mix transports." Not a blocker today, but worth resolving before 1.201.0 ships. (Get flow is unaffected.)

**Nits** — Studio Web format is restated in Step 6b *and* `api-endpoints.md` (minor tension with your "one home per fact" rule); the API fallback flow has no SW-link step.

Also confirm CI's `uip-check` tolerates the #3720-dependent commands no released `uip` exposes yet.

**Verdict: ready to merge after the two suggestion fixes.** 👍

magickaichen (changes requested):
I found two blockers at `e600b9c0d`; this is my complete known blocker set for the current head.

1. **The CLI deep-link recipe double-appends the tenant path.** Both CLI flows tell the agent to build `{baseUrl}/{org}/{tenant}/automationhub_/automation-profile/{process_slug}`, using `uip ah auth-info get` → `Data.Tenant.Url` as the base ([get flow](https://github.com/UiPath/skills/blob/e600b9c0deb69d5b1f49c4960322caf1e41ba0a8/skills/uipath-automationhub/references/get-process-cli.md#L43-L56), [publish flow](https://github.com/UiPath/skills/blob/e600b9c0deb69d5b1f49c4960322caf1e41ba0a8/skills/uipath-automationhub/references/publish-process-cli.md#L96-L110)). But `Data.Tenant.Url` maps to `tenant_company_url`, which Automation Hub already sets to `.../{org}/{tenant}/automationhub_` ([definition](https://github.com/UiPath/Automation-Hub/blob/edd66c55506a20818f729f83e1dd4edd2553f108/frontdoor-service/app/middleware/fps.ts#L189-L192)); AH itself builds links as `${tenant_company_url}/automation-profile/${processSlug}` ([usage](https://github.com/UiPath/Automation-Hub/blob/edd66c55506a20818f729f83e1dd4edd2553f108/tenant-service/app/controllers/external-connections/jira-connection.controller.ts#L680-L685)). Please use `${Data.Tenant.Url}/automation-profile/${process_slug}` and append only `/documentation` where needed.

2. **The CLI auth boundary is still contradictory.** Step 0 says the CLI path must never touch a token, but the routing section says “All flows share the Authentication section above,” sending the agent back into raw-token/header handling ([SKILL.md](https://github.com/UiPath/skills/blob/e600b9c0deb69d5b1f49c4960322caf1e41ba0a8/skills/uipath-automationhub/SKILL.md#L15-L23), [routing sentence](https://github.com/UiPath/skills/blob/e600b9c0deb69d5b1f49c4960322caf1e41ba0a8/skills/uipath-automationhub/SKILL.md#L48-L56)). Please scope that sentence to raw-API flows only.

Non-blocking follow-ups:

- When an `ah`-capable CLI lacks the #3720 flags, stop and require an upgrade instead of suggesting a raw-API switch after the transport has already been selected.
- Add explicit coverage for `uip ah --help` routing and the raw-API fallback arm.
- Bind smoke operation criteria to actual `uip ah ...` commands; the current bare verb patterns can pass without proving the intended transport.
- Keep transport-independent domain facts in `api-endpoints.md` as the new one-home rule requires, and use the repository's `-guide.md` naming convention for the new flow guides.

parth-patil-uipath:
Both blockers fixed in e1c7eb30d (plus main merged down):

**1. Deep-link double-append — fixed.** Confirmed against AH source: `Data.Tenant.Url` maps to `tenant_company_url`, which frontdoor already builds as `{base}/{org}/{tenant}/automationhub_` (fps.ts). Both CLI flows now build `{Data.Tenant.Url}/automation-profile/{process_slug}` and each carries an explicit never-re-append-org/tenant warning.

**2. CLI auth boundary contradiction — fixed.** The routing sentence now scopes the Authentication section and `api-endpoints.md` catalog to the raw-API flows only; the CLI flows are explicitly told they never touch either.

Non-blocking follow-ups, also taken:
- **Upgrade-not-fallback**: when an `ah`-capable `uip` rejects the #3720 flags, `cli-commands.md` now says upgrade and stop — never switch to the raw-API path after the transport was selected.
- **Smoke criteria bound to the transport**: every command_executed pattern is now `uip ah <group> <verb>`-anchored, and both smokes gained an explicit `uip ah --help` preflight-routing criterion.
- **`-guide.md` naming**: the two new flow files are renamed `get-process-cli-guide.md` / `publish-process-cli-guide.md`; all links updated.

Deferred (happy to do in a follow-up): an eval for the raw-API fallback arm needs an image whose `uip` predates the `ah` surface — not expressible with the current smoke fixtures. `check-skill-verbs`/`check-cli-verbs` run clean (0 High, 0 Medium) on the new head.

magickaichen (approved):
Re-review at `6d93e8ab9`, against my previous review at `e600b9c0d`. Discovery diff is `e1c7eb30d` (the only PR-authored commit since); `6d93e8ab9` is a clean `origin/main` merge that touches no PR-owned file.

## Prior blocker ledger — both fixed ✅

1. **Deep-link double-append — fixed.** Both CLI guides now build off `Data.Tenant.Url` and carry an explicit "never re-append org/tenant segments" rule ([get-process-cli-guide.md#L50-L56](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/references/get-process-cli-guide.md#L50-L56), [publish-process-cli-guide.md#L106-L110](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/references/publish-process-cli-guide.md#L106-L110)).
2. **CLI auth boundary — fixed.** The routing sentence is now scoped to the raw-API flows ([SKILL.md#L50](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L50)), and the Authentication heading itself says "raw-API flows only".

Follow-ups (a) upgrade-and-stop, (c) `uip ah`-bound smoke criteria, and (d) the `-guide.md` renames all landed as well. No new P1 was introduced, so this is an approve. Everything below is non-blocking.

## Standards

**1. The scoping fix reads slightly too wide, and the CLI guides aren't self-contained yet (P2, introduced in `e1c7eb30d`).** [SKILL.md#L50](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L50) now says the CLI flows "never touch either" — read plainly, that includes the endpoint catalog in `api-endpoints.md`. But [publish-process-cli-guide.md#L68](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/references/publish-process-cli-guide.md#L68) sends the agent straight there for the document-type ids ("read it there; do not guess ids"), and that table lives only in [api-endpoints.md#L89-L100](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/references/api-endpoints.md#L89-L100). It also sits against [#L19](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L19) ("the domain contract … lives in `api-endpoints.md`") and the one-home rule at [#L61](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L61). Would it work to scope L50 to auth and HTTP mechanics specifically, and keep the domain contract explicitly shared? The same reading makes "the CLI files are self-contained by design" at [#L64](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L64) not quite true today.

**2. Two extension instructions for one file (P3).** The future-operation row at [SKILL.md#L57](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L57) still says `references/<operation>.md` while [#L62](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/SKILL.md#L62) now says `references/<operation>-cli-guide.md`. Worth aligning the row with the rename.

**3. Smoke `command_pattern` style (P3).** The new criteria use literal single spaces (`uip ah automations get`), while the sibling criterion in the same file uses `'(?i)uip ah '` and the wider `tests/tasks` convention is `\s+` separators with a `(uip|\$UIP)` alternation. Cosmetic, but it drifts within one file.

The `-guide.md` renames themselves are correct and complete — `CONTRIBUTING.md` documents "Guides end with `-guide.md`", and no dangling `*-cli.md` links remain.

**Standards summary:** 3 findings, none blocking; worst is the over-wide "never touch either" scoping at `SKILL.md#L50`.

## Spec

Spec source: [RPANAV-18939](https://uipath.atlassian.net/browse/RPANAV-18939) — "skill invokes `uip ah` for publish + get, and both flows pass against a real tenant" — plus the PR description's stated design.

**1. The raw-API fallback arm has no coverage (P2, pre-existing — I missed it at `e600b9c0d`).** Both smoke files now assert an `ah`-capable image and match only `uip ah …` ([smoke_get_process.yaml#L6-L8](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/tests/tasks/uipath-automationhub/smoke_get_process.yaml#L6-L8)). At base they exercised the raw-API path, so the `unknown command 'ah'` arm the PR keeps alive is now untested by anything in-repo. Since the fallback is the arm every current install actually takes, a third task pinned to a pre-`ah` CLI would be worth adding before the fallback retires.

**2. The 200 MB file cap looks off (P2, pre-existing).** [publish-process-cli-guide.md#L69](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/references/publish-process-cli-guide.md#L69) says "200 MB cap", while the transport-independent contract says 300mb ([api-endpoints.md#L102](https://github.com/UiPath/skills/blob/6d93e8ab9f003b1e09a8319e2d6d77d5b604c9f7/skills/uipath-automationhub/references/api-endpoints.md#L102)) and so does the API twin. I couldn't find a 200 MB guard in UiPath/cli#3720's `documents.ts` either — if the CLI does enforce its own limit, could you cite it? Otherwise this reads as a forked fact against the L61 one-home rule.

**3. Slug URL-encoding dropped on the CLI path (P3, pre-existing).** `publish-process.md#L153` requires URL-encoding the slug; the CLI publish guide's link section has no equivalent.

**4. Studio Web link (Step 6b) is beyond the ticket's "swap the transport" scope (P3, pre-existing).** Well-guarded with a never-invent rule and it has a documented home, so this is just a note.

On acceptance: the "passes against a real tenant" evidence lives entirely in the PR description's two transcripts — there's nothing in-repo confirming it. Not disputed, just flagging where it rests. Parity with #2708's hardening otherwise checks out (schema-derived required set, tenant-required app questions, doc-type table by reference, wrapping rules, verify-then-report, slug recovery).

**Spec summary:** 4 findings, none blocking; worst is the untested raw-API fallback arm.

parth-patil-uipath:
@alena-apanovich-1812 following up on your review — status of each point:

**Already fixed in e1c7eb30d** (with Kaichen's blockers):
- **Stale intro line**: the routing sentence now scopes the Authentication section + `api-endpoints.md` catalog to the raw-API flows only.
- **Deep-link base URL**: pinned — `Data.Tenant.Url` maps to AH's `tenant_company_url`, which frontdoor already builds as `{base}/{org}/{tenant}/automationhub_` (fps.ts:191). Both CLI flows now build `{Data.Tenant.Url}/automation-profile/{process_slug}` with an explicit never-re-append warning.
- **Pre-#3720 middle state**: resolved to upgrade-and-stop — an `ah`-capable `uip` that rejects the #3720 flags means upgrade; never switch to the raw-API path after the transport was selected at preflight. (And #3720 itself is approved + in sync with main, so the window should be short.)

**Nits — now fixed in 2e45b2608**:
- Studio Web value format lives only in `api-endpoints.md` now; the CLI flow's Step 6b references it instead of restating (keeping only the CLI-specific discovery recipe).
- The raw-API publish flow gained its missing Step 6b (SW link via `PATCH /automations/{id}` with a `user_inputs` body), referencing the same one-home fact.

**uip-check on #3720-dependent commands**: confirmed tolerated — `check-cli-verbs` runs against the 1.201.0-dev catalog which already carries the `ah` surface (0 High / 0 Medium locally), and the PR's own `uipath-automationhub` CI checks are green, smoke tests included (the earlier smoke failure was a transient Bedrock IAM 403 in the eval runner; the re-run passed).

abhiram-vad (approved):

PR #451, opened 2026-04-28
test(uipath-maestro-case): add single-node tests for non-connector ta…

charlesliu9:
Adds e2e tests covering agent, api_workflow, case_management, process, and rpa task types, plus a shared case_check helper that locates the generated caseplan.json, runs `uip maestro case validate`, and asserts the expected task type is present.

PR #2726, opened 2026-08-20
Add Maestro SDK preview eval staging

tmatup:
## Summary

- stage the complete skills plugin, including `commands/` and `hooks/`, with the three `skills/uipath-maestro-*` directories replaced byte-for-byte by their synced `preview/` snapshots
- provide the `maestro-flow-sdk` experiment and Docker overlay used by both ad-hoc and scheduled evaluation
- keep checker resolution on the real skills checkout while the agent loads the staged plugin
- resolve the published Flow SDK to an exact version per build and record it with the skills and flow-builder-sdk revisions

The canonical schedule and exact-run comparison now live in companion PR UiPath/coder_eval_uipath#93, following reviewer feedback. This PR no longer adds an independent scheduler or searches blob history.

## Verification

- corrected post-#451 nine-task sweep: [run 32452595430](https://github.com/UiPath/skills/actions/runs/32452595430), **9/9 SUCCESS**, pass rate 1.0, mean weighted score 0.9899
- exact sweep provenance: skills `943641d7e838b20b856a6d56e03789797e231575`, flow-builder-sdk `32331e449e8a3c8126acf9564f45bb86b47c469f`, `@uipath/flow-sdk@3.7.0`, Codex `gpt-5.6-terra`
- artifact audit: every task linked all three staged preview skills; all 9 pre-runs and all 24 post-runs exited 0; all nine SDK-provenance files agree; no unresolved-plugin, zero-skills, checker-import, or auth errors
- one-task plugin smoke: [run 32424560311](https://github.com/UiPath/skills/actions/runs/32424560311), SUCCESS at 0.909; all three live malformed-input debug cases passed
- staging/sync focused tests: 6/6 pass; actual plugin-tree staging verified with all three preview replacements plus commands and hooks
- full `npm run skills:test`: 55/57 pass; the two failures are pre-existing npm registry-state dry-run failures (`--tag` / published latest higher than fixture 1.2.3)
- Windows coder-eval compatibility fix is isolated in UiPath/skills#2733; its full Windows smoke run [32450974211](https://github.com/UiPath/skills/actions/runs/32450974211) passed

🤖 Generated with Claude Code


tmatup:
Addressed the review’s provenance-note suggestion in `819c20174`: the experiment comment now describes the runtime provenance mechanism without embedding SHAs that will drift after auto-sync. The exact skills and flow-builder-sdk revisions remain recorded in each run’s `preview-plugin-source.json` and per-task `preview-sdk-provenance.json`.

The two manual-review notes are also covered: current VM artifacts record task-level `model_used` (verified against a real VM `run.json`), and the Windows `httpx` pin is scoped to the coder-eval 0.10.2 compatibility gap.

dmetzgar (approved):

bai-uipath (commented):
## Wrong repo

The staging content belongs here. The schedule and the comparison don't.

`skills` has no scheduled eval workflow today, and the 04:00 nightly this joins against lives in coder_eval_uipath (`daily.sh` + the ADO pipeline). This stands up a second, independent eval scheduler in a different repo, on a different runner pool, against the same staging tenant, coupled to the first only by a hand-picked cron offset and a blob container.

Please move the schedule and comparison there as a fourth RunConfig. That pipeline already carries several canonical runs a day. The experiment YAML, task manifest, Dockerfile and staging scripts are correctly placed here.

## What that costs

- **The baseline join is an unbounded reach-back.** No age bound, and ADO runs claude-code Mon and Thu while this cron is Mon-Fri, so most runs compare against a different day silently. A nightly that dies uploads nothing, which widens the gap further.
- **The join already exists upstream.** RunConfig identity, the run-history fetch with its partial/adhoc filters, and the upload path all do this already. Roughly 450 lines reimplemented in bash and Node.
- **Baseline discovery is capped at 5000 blobs.** `az storage blob list` truncates silently on a container holding millions, so the candidate set need not contain a recent run.
- **A codex dispatch inverts the comparison.** Maturity-imputed rows score a synthetic 1.0 at zero cost and the row projection doesn't exclude them. Claude is unaffected.

Minor: the inherited `/.uipath` mount is inert under the forwarded host HOME.

uipreliga:
## Code review — preview nightly eval arm

Multi-model review (Gemini 3 + Opus sub-agent), with the load-bearing claims verified against the repo rather than taken at face value. Two high-severity findings needed recalibration and one is a confirmed false positive — both noted below.

### High

**1. Azure baseline listing silently truncates to the *oldest* runs** — `.github/workflows/preview-nightly.yml:96-108`

`az storage blob list` defaults to `--num-results 5000`, and `--query` is a client-side JMESPath filter applied *after* retrieval. Per `CLAUDE.md` this container stores `<run-id>/<variant>/<task-id>/<replicate>/task.json` for every task of every run, so one nightly contributes hundreds of blobs. Blobs return in lexicographic name order, so once the container exceeds 5000 blobs the window covers only the earliest run-ids — recent nightlies are never listed. Result is either `No regular-nightly baseline matches …` every night, or a silent comparison against a months-old baseline whose model happens to match.

Fix: list run-id prefixes with `--delimiter '/'` (or `--prefix "$(date -u +%Y-%m)"`), then download `<run-id>/run.json` per candidate, tolerating 404s.

**2. Staged preview plugin drops `hooks/` and `commands/`** — `scripts/stage-preview-plugin.mjs:83-96`

The baseline arm points `plugins:` at the whole repo (`nightly.yaml:39-41`), so it loads `hooks/hooks.json` (SessionStart / PostToolUse / SessionEnd / Stop / StopFailure) and `commands/`. The staged tree copies only `.claude-plugin/` and `skills/`.

Calibration note: the hooks are `suggest-permissions`, `set-session-env`, and `send-telemetry` — telemetry and advisory, not capability-changing — so scoring impact is likely small. The real cost is that the two arms aren't environment-identical, preview runs emit no telemetry and no `UIPATH_SESSION_ID` correlation, and nothing in the script, its comments, or its tests records the omission as deliberate. Either copy them, or assert the exclusion in a test plus a comment.

### Medium

**3. Baseline timestamp fallback is inverted** — `compare-preview-nightly.mjs:78-81` + the download loop at `preview-nightly.yml:113-122`

`runTimestamp` falls back to file mtime; the loop writes candidates newest-first, so the newest run gets the *oldest* mtime. If `run.json` ever names its timestamp anything but `end_time`/`start_time`, `selectBaseline` returns the oldest of 60 downloads, silently. Every test supplies `end_time`, so CI can't see it. Derive the timestamp from the run-id already embedded in the filename — it's already the sort key upstream.

**4. `oneTask` aborts the whole report** — `compare-preview-nightly.mjs:46-50`

Two triggers: a preview task that produced no row, and a baseline with more than one replicate — `baselineMatches` only checks `.some()`, so a multi-replicate baseline passes selection and then crashes the join with `must contain exactly one <id> row; found 2`. One bad row loses the other eight comparisons.

**5. Unpinned SDK, and its version never reaches the artifact** — `tests/docker/preview-nightly.Dockerfile:5`

`FLOW_SDK_VERSION=latest` is never overridden, while everything else in the arm is commit-pinned. The resolved version lands in `preview-sdk-provenance.json` *inside each sandbox*, where nothing reads it — so a day-over-day delta has no record of which SDK build produced it. Resolve the exact version in the build step, pass it as `--build-arg`, and surface it in the comparison's `config` block.

**6. Hardcoded container mount path** — `tests/experiments/preview-nightly.yaml:38`

`${SKILLS_REPO_HOST_PATH}:/home/runner/work/skills/skills:ro` works only because the frozen checkers walk up from `$TASK_DIR` looking for `_shared` (confirmed at `check_billing_invoice_lookup.py:19-21`). Any runner whose `GITHUB_WORKSPACE` differs and every checker fails at once, presenting as a total preview regression. `${SKILLS_REPO_HOST_PATH}:${SKILLS_REPO_HOST_PATH}:ro` is self-describing and can't drift.

**7. Upstream SDK checkout lands inside the mounted workspace** — `.github/workflows/run-coder-eval.yml:396-402`

`preview-flow-builder-sdk/` is checked out into `github.workspace`, which `preview-nightly.yaml` mounts read-only into every eval container at two paths, and the system prompt permits reading everywhere but `runs/*`. The agent can therefore read `typescript/sdk/skill/` — the very docs the preview snapshot was generated from — during the eval. Checking out to `${{ runner.temp }}` and pointing `--build-context` there costs one line.

**8. `httpx` pin covers 1 of 5 install sites** — `.github/workflows/smoke-rpa-skills.yml:133`

`smoke-skills.yml:351`, `:695`, and `run-coder-eval.yml:324`, `:752-758` install the same pinned coder-eval 0.10.2 the same way, unpinned. Either the breakage is Windows-tempdir-specific (in which case the comment should say so) or four other sites are equally broken. Unrelated to this arm — probably belongs in its own PR.

**9. `compare` runs when `eval` was *skipped*** — `.github/workflows/preview-nightly.yml:75`

Running on `failure` is deliberate and correct here — the arm is expected to have failing tasks and the delta table is still the deliverable. But running on `skipped` (when the `tasks` job fails) burns an Azure login and 60 blob downloads before failing on the `[ -f "$preview_json" ]` guard. `needs.eval.result == 'success' || needs.eval.result == 'failure'`.

### Low

- **Duplicate provenance-pin logic.** `scripts/sync-maestro-sdk-preview.mjs:146` already implements this — with a *tighter* per-file anchored regex and the same cross-file disagreement check. `stage-preview-plugin.mjs:53-59` reimplements it with a looser `typescript/sdk/skill/[^`]+` pattern not anchored to the expected source file for that skill. Export the existing one.
- **Dead guard.** `environment_info.preview_nightly` (`compare-preview-nightly.mjs:59`) is written nowhere in the repo, and preview archives are `NF=3` so the `awk -F/ 'NF == 2'` filter already excludes them. Unreachable.
- **Tautological fallback.** The no-`run_config` branch (`:73-75`) re-hardcodes `claude-code`/`alpha`, which the workflow always passes verbatim. Reachable but untested.
- **Model accessor disagreement.** `singleModel` reads `model_used ?? agent_config?.model`; `baselineMatches` reads only `model_used`.
- `aggregate` sums nulls as 0 (`:126-138`); triplicated `9`; jq diagnostics filter `.status` while everything else prefers `final_status` (`run-coder-eval.yml:606`); `mkdir -p /tmp/runs` in an `always()` step defeats the summary's "no run directory produced" guard; the summary now walks all of `/tmp/runs`, which would fold in stale `task.json` on a non-ephemeral runner; `mkdirSync(output)` runs before `assertCleanPluginInputs`, so a failed clean check leaves an empty dir that passes the emptiness precondition on retry; the manifest is parsed by `sed` in YAML and `.trim()` in JS; `[ ! -e node_modules ]` hard-fails a pre_run re-run; `--build-context` is buildx-only (the neighbouring steps only need classic-BuildKit `--secret`); `secrets: inherit` is broader than needed; `PREVIEW_SKILLS_PLUGIN_PATH` is passed into the container as a host path nothing reads.

### False positives / overstated

- **`set -euo pipefail` under dash** — false positive. `tests/docker/Dockerfile:62` already does exactly this on the same base image and builds today.
- **"Compare job fails deterministically on the artifact root"** — overstated. The documented layout (`<run-id>/run.json`) matches the new root-level paths, so the least-common-ancestor calculation does put `run.json` at the artifact root. Still worth hardening with `find preview-result -name run.json -print -quit` rather than depending on it implicitly.
- **Summary-step rewrite as a general regression** — it's a *fix* for the current layout, not a gratuitous change. Residual risk is the stale-runner case only.

### Positive observations

Image-overlay ordering is right — `skills-image:preview-nightly` builds `FROM skills-image:latest` *after* the delegate overlay retags it, so `experiment=preview-nightly` composes with `agent=delegate-sdk` instead of silently dropping that layer. Secret hygiene is sound: the token-bearing `.npmrc` is written and removed inside one `RUN` behind `--mount=type=secret`, and `persist-credentials: false` is correct given the checkout doubles as a build context. Every third-party action is SHA-pinned. `assertCleanPluginInputs` plus the short-SHA→full-SHA `gh api` resolution with a prefix check are real provenance guards, not ceremony. `errorOnExist` + `COPYFILE_EXCL` + the empty-output precondition make staging fail loudly, and the "refuses a non-empty destination" test asserts the file survived, not just the exit code. Windows exclusion is enforced at partition time rather than discovered mid-run. Input parity between the new `workflow_call` block and `workflow_dispatch` is complete — all ten checked.

### Test status

The 8 new tests pass. Untested and load-bearing: the mtime fallback (3), duplicate baseline rows (4), the no-`run_config` branch, `metrics`/`aggregate` null handling, and — the gap that let (2) through — no assertion about which top-level plugin directories reach the staged tree. `npm run skills:test` also has a pre-existing failure in `compose-skill-flavor.test.mjs` (npm publish prerelease tag), unrelated to this PR.

### Overall

Architecturally sound with unusually careful workflow plumbing; the defects are concentrated in the comparison's silent-degradation paths rather than its structure. Suggest fixing 1–4 before merge, splitting 8 into its own PR, and treating the rest as follow-ups.

uipreliga (changes requested):
I think most of this PR (what @bai-uipath said) should be in a different repo. 

why do we need different experiment setup or docker file? let's rename them so that they reflect the flow sdk experiments.

tmatup:
@bai-uipath @uipreliga Thank you — the review changed the architecture. The schedule and comparison have been removed from this repo and now live in the canonical ADO nightly as UiPath/coder_eval_uipath#93. The paired arm runs sequentially after the same build's Codex live-v1 result, compares those exact two run artifacts by task id and RunConfig, rejects maturity-imputed/missing/duplicate rows, and uploads only after the canonical result and Slack notification. There is no blob-history search, second scheduler, or 5,000-blob cap.

The remaining skills-side findings are addressed in `a5e297d89`:

- staging now copies `commands/` and `hooks/`, with assertions
- experiment/Docker/manifest names now describe `maestro-flow-sdk` rather than a scheduler
- the SDK build resolves `@latest` once, passes the exact version as a build arg, and records that version plus both repository revisions; #93 also embeds them in the comparison config
- the checker mount is `/skills-repo`, and the ad-hoc upstream source is materialized under `RUNNER_TEMP`, outside that mount
- the unrelated `httpx` pin is reverted
- failed dirty-input staging no longer creates its destination, and the unused container passthrough was removed

Focused staging/sync tests pass 5/5, actual full-tree staging includes the support directories and all three byte-identical preview replacements, and coder_eval_uipath `make verify` passes 399 tests at 93.30% coverage. A post-#451 nine-task run is in progress: https://github.com/UiPath/skills/actions/runs/32450668303

🤖 Generated with Claude Code

tmatup:
The independent Windows `httpx` workaround is now split into UiPath/skills#2733. Fresh #2726 CI provided direct evidence for its scope: all three RPA tasks failed before execution because coder-eval 0.10.2 could not import `llm_judge` without `httpx`; the prior run with the pin passed 3/3. #2726 itself remains free of that unrelated change.

tmatup:
The portable corrected sweep is complete: [run 32452595430](https://github.com/UiPath/skills/actions/runs/32452595430) finished **9/9 SUCCESS** on the exact frozen task list with Codex `gpt-5.6-terra`.

Artifact verification (not just the workflow summary):

- skills `943641d7e838b20b856a6d56e03789797e231575`
- flow-builder-sdk `32331e449e8a3c8126acf9564f45bb86b47c469f`
- `@uipath/flow-sdk@3.7.0`
- all nine task logs link `uipath-maestro-flow`, `uipath-maestro-case`, and `uipath-maestro-bpmn` from the staged plugin
- all 9 pre-run commands and all 24 post-run commands exited 0
- all nine per-task provenance files agree
- no unresolved-plugin, zero-skills, `_shared` import, or auth-context errors

The run's pass rate is 1.0 and mean weighted score is 0.9899; billing invoice lookup is the only non-perfect weighted score (0.909), from a non-blocking advisory criterion. No content gap remains from this sweep.

This closes the M2 ad-hoc proof. The schedule and paired same-build comparison remain in UiPath/coder_eval_uipath#93 as requested.

🤖 Generated with Claude Code

PR #2731, opened 2026-08-21
fix(ci): stop compliance-pack tasks racing on the shared tenant

sauravrathore01:
The 15 compliance-pack tasks share ONE ISO 42001 pack state on the test tenant and mutate it: disable_smoke turns the pack off while restore_smoke and drift_restore_smoke need it on for their whole turn. Run in parallel, one task's mutation lands mid-turn in another's — the agent correctly refuses to restore an inactive pack, and the test scores it a failure (CI 30455293977; run 2026-08-20_19-01-47, score 0.727).

The pre_run enable added earlier can't close the window: it runs before the agent starts, and its "already active -> skip" branch means the task never owned the enable in the first place.

Two changes:

- Ownership marker. setup_enable_compliance_pack.py drops a marker in the task sandbox only when it actually enabled the pack; cleanup_compliance_pack.py disables only when that marker is present, so the 12 tasks that never enable anything stop mutating pack state. Sandbox scoped via cwd, not /tmp — pre_run/post_run share one orchestrator container across parallel tasks, so a fixed /tmp path would be read and deleted by every concurrent task.

- Serial pass. smoke-skills.yml splits the selection: compliance-pack tasks run at -j 1 after the parallel pass, everything else keeps TASK_PARALLELISM. Needed because disable_smoke's agent disables the pack as its whole purpose — no cleanup gate can prevent that. Both passes write into the same run root, so the summarize step and the LLM reviewer gate now glob all run dirs instead of only the newest (the pass-rate gate already did).

Adds test_marker_gate.py: runs cleanup twice in a temp sandbox and asserts disable_pack() skips without a marker and proceeds with one. Offline, no auth.

Generated with Claude Code


jianjunwang2 (approved):

t-hsia (approved):

tests/tasks/uipath-governance/compliance-pack/cleanup_compliance_pack.py
```diff
@@ -92,6 +103,11 @@ def get_tenant_id():
 
 
 def disable_pack():
+    if not MARKER.exists():
```
abhinavmahanti: the marker gate keys on "did my setup enable this?", but six tasks call cleanup_compliance_pack.py from pre_run to force the pack off before the agent starts, and in a fresh sandbox no marker can exist so disable_pack() returns at line 1 and that reset becomes dead code. apply_commands_smoke:46-48 then fails with a correct agent, and full_apply_e2e_live's own header ("pre_run cleanup guarantees the pack starts off") is no longer true.

sauravrathore01: reverted the change.

abhinavmahanti (approved):

uipreliga:
## Code review — tenant-race serialization

Reviewed `ee0b3aab5` + `9fd323d24`, cross-checked with a second-opinion pass from Gemini 3 and GPT-5.

The core fix is correct, and the reader changes are the genuinely important part: `rglob` over the whole `/tmp/runs` tree in the summary, reviewer gate, and verdict now matches the pass-rate gate that already did it. Gating on "newest run dir" after a split would have silently graded half the suite — the LLM-reviewer gate catch in the second commit was the subtle one and it's right.

Findings below, most severe first.

### High

**1. Both `coder-eval` calls share one `--run-dir /tmp/runs` — worth confirming this actually yields two sibling run dirs**
`smoke-skills.yml:487,495` · `run-coder-eval.yml:508,516`

The design rests on "two passes → two run dirs under `/tmp/runs`, so readers glob all of them." Testing locally with `coder-eval 0.9.6`, `--run-dir <path>` makes `<path>` **the run directory itself** — `experiment.json`, `run.json`, `experiment.html`, `resume_fingerprint.json` land directly in it, with `<variant>/<task>/<NN>/task.json` beneath. Under those semantics the two passes write into the *same* run and the second overwrites the first's aggregate `experiment.*` / `run.json`. Per-task `task.json` files live at distinct subpaths and survive, so the gates fixed here still see everything — but the uploaded HTML report would describe only the serial pass, and a second invocation into a non-empty run dir may trip resume-fingerprint handling.

CI pins `0.10.2`, so this may well have changed. (Gemini asserted the timestamped-subdir layout, but inferred it from the workflow's own `/tmp/runs/*/` globs — circular, so not evidence.) If `--run-dir` is still the run dir itself on 0.10.2, giving each pass its own (`--run-dir /tmp/runs/parallel` / `/tmp/runs/serial`) makes the design true by construction rather than by luck.

### Medium

**2. `Summarize results` can't reach its own no-results branch** — `smoke-skills.yml:515`

```bash
run_dirs=$(ls -td /tmp/runs/*/ 2>/dev/null)
```

GH's default shell is `bash -e -o pipefail`. With no match the glob stays literal, `ls` exits 2, and the assignment aborts the step — so `::warning::No run directory found` + `exit 0` is dead code, and a catastrophic-install run dies here with a bare error instead. The twin at `run-coder-eval.yml:565` got exactly the right guard (`|| true` inside the substitution) in this same PR; this one didn't. The latent bug predates the PR (pipefail made `| head -n 1` fail too), but the asymmetry between the twins is new.

**3. Serializing within a run doesn't close the race** — `smoke-skills.yml:3-5` · `run-coder-eval.yml:94`

`smoke-skills`' concurrency group is per-`head_ref`, so two open PRs run compliance-pack against the same tenant simultaneously; `run-coder-eval` deliberately has no group at all, so an ad-hoc run collides with any PR. The mechanism narrows the window but the failure mode in the commit message stays reachable. `tests/README.md:103` ("Serialization is the whole mechanism") reads as if it's fully solved — worth an explicit known-gap line. If flakes persist, the real fix is a repo-wide concurrency group (or a tenant lock) for the exclusive suite.

**4. README claim is off by one task** — `tests/README.md:100`

"every task calls it from `post_run` to undo" — 15 of 16. `nonregression_branch_a_smoke.yaml` calls `cleanup_policy.py`, not `cleanup_compliance_pack.py`. The "six tasks call it from `pre_run`" count is exactly right.

**5. Path prefix over-selects** — `smoke-skills.yml:467` · `run-coder-eval.yml:497`

That same `nonregression_branch_a_smoke.yaml` never touches the ISO 42001 pack (it creates an AITL policy) yet lands in the `-j 1` pass purely for living in the directory. A `tenant-exclusive` tag on the task YAMLs would be precise, would drop the "extend the case branch in both workflows" maintenance instruction at `README:105-106`, and is the repo's established convention for exactly this — `windows`, `smoke`, and the benchmark dirs are all tag-driven. The README's counter-argument ("it inherits the serial pass, no tagging needed") is fair on KISS grounds, but it buys zero-config at the cost of duplicating the selector across two workflows and mis-serializing tasks.

### Low

**6. Dead guard that can kill the step** — `smoke-skills.yml:470`

`[ -n "$f" ] && selected+=("$f")` is the loop body's last command; on an empty line it returns 1 and `set -e` exits. `sort -u` won't emit blanks here, so it's unreachable today — but it's a landmine, not a safety net. `[ -z "$f" ] || selected+=("$f")` inverts it safely.

**7. Unneeded `set -u` armor** — `smoke-skills.yml:477`

`"${selected[@]+"${selected[@]}"}"` only matters under `set -u`, which this step doesn't enable; plain `"${selected[@]}"` reads far better. Note `run-coder-eval.yml:499` iterates `$TASK_GLOBS` directly with no such ceremony — the two blocks the comments call mirrors aren't actually written the same way.

**8. Glob re-expansion is more machinery than the job needs** — `smoke-skills.yml:468-475`

`shopt -s nullglob globstar` is already set at the top of the step, so a plain nested `for` builds `selected` in three lines. The `while` / process-substitution / `sort -u` exists only for dedupe, which `detect` already prevents (it skips task files covered by a whole-skill glob). Not wrong, just heavier than the problem.

**9. Duplication across two workflows + README prose** — `run-coder-eval.yml:490-521` ≈ `smoke-skills.yml:456-502`

The comment blocks are near-verbatim copies with slightly different wording, and the README asks maintainers to keep both in sync by hand. A tag (finding 5) or a shared `.github/scripts/` helper collapses this. Also worth noting: `TASK_PARALLELISM` defaults to `1` in smoke (`smoke-skills.yml:443`), so on the default config the split does two sequential runs to achieve what one `-j 1` run already did.

### Checked, not issues

- **Empty tag selection is benign.** Ran `coder-eval run <e2e-only-file> --tags smoke` locally: `Running 0 task(s)`, exit 0. A serial pass containing only `full_apply_e2e_live.yaml` won't set `rc=1`.
- **`exit $rc` propagation is correct** in both workflows, and running the serial pass even after the parallel pass fails is the right call — you get all results.

### Suggested before merge

Confirm the `--run-dir` semantics under 0.10.2 (1), add the `|| true` at `smoke-skills.yml:515` so the twins match (2), and fix the README's "every task" claim (4). The path-prefix-vs-tag choice and the extra bash machinery are judgment calls worth revisiting, not blockers.

rockymadden (approved):

akshaylive (approved):

akshaylive (changes requested):
On second thoughts I think this won't work for the nightly tests. Can we fix this issue using a different method?

bai-uipath (changes requested):
this PR adds steps specific to these tests to the coder eval pipeline for everyone, let's not do that lightly

the usual pattern for isolation is to generate a random unique id per test in pre_run that is used throughout the test so the test can mutate it and clean it up without affecting other tests, can we do that here?

PR #2735, opened 2026-08-21
docs(agents): document %ASSETS/<ASSET_NAME>% in coded agent env vars

alessandromuresan:
## What

A coded agent's environment variable can carry `%ASSETS/<ASSET_NAME>%` as its whole value; the platform substitutes the Orchestrator asset's value before the agent process starts. This documents that format, and — because nothing existed to hang it off — documents coded-agent runtime environment variables at all.

Before this PR, `%ASSETS`, `projectConfiguration`, `dotenv` and `load_dotenv` had **zero hits** repo-wide. `.env` was described only as the carrier for `UIPATH_PROJECT_ID`.

## New reference

`skills/uipath-agents/references/coded/lifecycle/environment-variables.md`

| Section | Why it's there |
| --- | --- |
| Three stores (`.env` / AgentHub project configuration / Orchestrator process-job) | An agent's first instinct is to edit `.env` and expect the cloud to see it. `.env` is not pushed, so it won't. |
| Format rules | Whole value only; case-sensitive `%ASSETS/` prefix; name matched case-insensitively; one asset per variable |
| Lookup folder | Personal workspace when debugging, the **job's** folder when published — this is the mechanism behind "resolves when deployed, not when debugging" |
| Value types | Text/Bool/Integer/Secret resolve; KeyValueList and connection-string types never do; credentials differ across paths |
| Failures are silent | No job failure, no log, variable dropped from the process environment |
| Anti-patterns + troubleshooting | |

Two behaviours get anti-pattern entries because they cost real debugging time and nothing warns you:

1. **Substitution is server-side.** `uip codedagent run` reads `.env` verbatim, so `os.getenv` returns the literal `%ASSETS/Name%`.
2. **An unresolvable reference is silent.** It neither fails the job nor logs anything — the variable is simply unset.

## Routing

- `uipath-agents/SKILL.md` — Task Navigation row
- `coded/quickstart.md` — Lifecycle Stages row, which is the table that tells an agent to load a reference only when it reaches that stage

Both insertions are **outside** all `skill-flavor` blocks (the first starts at quickstart line 233), so `studioweb` inherits them rather than needing an override.

## Corrections to existing content

Each of these previously made a claim that this change contradicts or leaves half-told:

| File | Was |
| --- | --- |
| `coded/lifecycle/file-sync.md` | Listed `.env` as excluded from push without saying where env vars *do* live; "do not add them to `.env`" invited over-generalising from credentials |
| `coded/lifecycle/deployment.md` | Configuration Files table had no `.env` row; "no manual `.env` wiring is required" (true of credentials) read as true of all environment variables |

## Deliberately not touched

- **Low-code references.** Low-code agents have no environment-variable concept. An earlier revision annotated `environmentVariables` / `referencedAssets` in `lowcode/capabilities/process/solution-files.md`; that was reverted and the citation removed from the new reference. Worth knowing separately: I could not substantiate `referencedAssets` — it is absent from the spec of a real cloud-refreshed agent process resource even for a solution that *does* reference an asset from an env var (where the asset instead appears as its own `asset/secretAsset/<Name>.json` resource). `solution-files.md:53` still asserts the field; that looks stale but is pre-existing and out of scope here.
- **`uipath-platform` / `run-jobs.md`.** An earlier revision added the `%ASSETS/` form next to the `KEY=VALUE` wire format there. Dropped — the shape is already clear from the new reference, which covers the published path and its lookup folder, so duplicating it in the platform skill only adds a second place to keep in sync.
- **A test task.** `tests/tasks/` needs tier/tag/sandbox decisions I'd rather agree first. Happy to follow `/test-coverage` → `/generate-task` → `/lint-task` in a follow-up. Nothing existing regresses: the tests asserting on `.env` (`push_pull_roundtrip`) cover push/pull behaviour, which is unchanged.
- **MCP server env vars** use the same `%ASSETS/` syntax — it's what AgentHub's substitutor was originally built for — and `skills/uipath-mcp-servers` documents no env vars at all. Out of scope for "coded agent env vars", but it's the obvious remaining gap.

## Reviewer note

The new reference links once into `uipath-platform` (`run-jobs.md`), as a pointer to where a published agent's environment variables are set. There is precedent for cross-skill links in `uipath-agents/SKILL.md`, and the sentence stands alone without following it (the `uip or processes update --environment-variables` command is given inline), so the skill still functions in isolation per the self-containment rule. Flagging it since it's on the PR checklist.

## Verification

- `npm run skills:validate` — `default: 26 skills, 1717 files`; `studioweb: 26 skills, 1717 files, 167 replacements`
- All three relative links in the new reference resolve to existing files
- `scripts/check-skill-verbs.py` on the new reference — `OK, no stale uip verbs found`
- CODEOWNERS needs no change: `uipath-agents` is covered by a directory-prefix entry; the new-skill rule doesn't apply
- Prose follows `.claude/rules/token-optimization.md`; placeholders use the `<UPPER_SNAKE_CASE>` convention from `.claude/rules/content-quality.md`

## Provenance

The format and its edge cases were established while implementing and debugging it end-to-end: server-side resolution in AgentHubService#1639, the editor and asset-registration in UiPath/flow-workbench#3123, and a live repro on alpha for the silent-failure and lookup-folder behaviour.

### DIFF
--- a/skills/uipath-agents/references/coded/lifecycle/environment-variables.md
+++ b/skills/uipath-agents/references/coded/lifecycle/environment-variables.md
@@ -1,0 +1,90 @@
+# Coded Agent Environment Variables
+
+Where a coded agent's environment variables are stored, which store the runtime reads, and how to pull a value from an Orchestrator asset instead of hardcoding it.
+
+## Three stores, one variable
+
+Not the same store. Only one is what the cloud runtime reads.
+
+| Store | Read by | Authored with |
+|-------|---------|---------------|
+| `<PROJECT_DIR>/.env` | Local runs only (`uip codedagent run`, `uipath run`) | Edit file directly |
+| Project configuration (AgentHub) | Cloud runtime, every debug and Studio Web run | Studio Web → agent properties → **Coded agent environment variables**. VS Code extension mirrors `.env` here on debug |
+| Orchestrator process / job environment variables | Published agent started as an Orchestrator job | `uip or processes update --environment-variables`, or per job on `jobs start` — see [`uipath-platform` run-jobs](../../../../uipath-platform/references/orchestrator/run-jobs.md) |
+
+Consequences:
+
+1. **`.env` is not pushed.** `uip codedagent push` excludes it — see [file-sync](file-sync.md) § Files Involved. Editing `.env` alone changes nothing in the cloud.
+2. **Local and cloud runs can disagree.** Different stores. Diagnose a cloud-only failure by comparing both, not by trusting `.env`.
+3. **No credentials in `.env`.** `UIPATH_URL`, `UIPATH_ACCESS_TOKEN`, org/tenant identifiers come from the `uip login` session. `UIPATH_PROJECT_ID` is the one identity value that belongs there.
+
+## Referencing an Orchestrator asset
+
+Set a variable's whole value to `%ASSETS/<ASSET_NAME>%`. The platform substitutes the asset's value before the agent process starts; the agent reads an ordinary environment variable and never calls the Assets API.
+
+```env
+MY_API_KEY=%ASSETS/ServiceApiKey%
+DATABASE_HOST=%ASSETS/DbHost%
+```
+
+```python
+import os
+
+api_key = os.getenv("MY_API_KEY")
+```
+
+The asset name stays out of the code, so it can be re-pointed per environment without a code change.
+
+### Format rules
+
+1. **Whole value only.** `%ASSETS/Name%` resolves. `https://%ASSETS/Host%/api` does not — passed through as literal text, no error.
+2. **Uppercase prefix: `%ASSETS/`.** One parser in the chain compares the prefix case-sensitively, so `%assets/Name%` silently fails to resolve on the published path even though the design-time editor accepts it.
+3. **Asset name matched case-insensitively.** Name only between the slash and closing `%` — no folder path.
+4. **One asset per variable.** No concatenation, no default value.
+
+### Lookup folder
+
+The reference carries no folder, so the folder comes from the launch path. Most common reason a reference resolves in one place but not another:
+
+| Launch path | Lookup folder |
+|-------------|---------------|
+| Debug run (Studio Web or VS Code extension) | User's **personal workspace** |
+| Published agent started as an Orchestrator job | The **job's** folder — not the agent's, not where the package was published |
+
+An asset that exists only in a solution folder resolves for the deployed agent but not while debugging. Create it in the personal workspace too when both paths are needed.
+
+### Asset value types
+
+`Text`, `Bool`, `Integer`, `Secret` resolve. `KeyValueList` and connection-string types have no single-string form and never resolve. Credential types behave differently across the debug and published paths — do not reference them from an environment variable.
+
+### Failures are silent
+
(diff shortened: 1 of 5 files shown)


skills/uipath-agents/references/coded/lifecycle/environment-variables.md
```diff
@@ -0,0 +1,95 @@
...
+DATABASE_HOST=%ASSETS/DbHost%
+```
+
+```python
+import os
+
+api_key = os.getenv("MY_API_KEY")
+```
+
+Prefer this over `sdk.assets.retrieve(...)` for startup configuration: keeps the asset name out of code, re-pointable per environment without a code change.
```
radugheo: I think `re-pointable per environment without a code change` (for `api_key = os.getenv("MY_API_KEY")`) can be confused with `That call ... stays overridable per environment` from the 4th antipattern below, which is actually for `sdk.assets.retrieve(...)`.
https://github.com/UiPath/skills/pull/2735/changes#diff-c0ccfa96050f1415bc177c0693f8e07686afcc274e452b252c93de0c3317147dR83

skills/uipath-agents/references/coded/lifecycle/environment-variables.md
```diff
@@ -0,0 +1,95 @@
...
+| Launch path | Lookup folder |
+|-------------|---------------|
+| Debug run (Studio Web or VS Code extension) | User's **personal workspace** |
+| Published agent started as an Orchestrator job | The **job's** folder — not the agent's, not where the package was published |
+
+An asset that exists only in a solution folder resolves for the deployed agent but not while debugging. Create it in the personal workspace too when both paths are needed.
+
+### Asset value types
+
+`Text`, `Bool`, `Integer`, `Secret` resolve. `KeyValueList` and connection-string types have no single-string form and never resolve. Credential types behave differently across the debug and published paths — do not use them for an environment variable; read a credential with `sdk.assets.retrieve_credential(...)`.
```
radugheo: Should we also add the asset type `KeyValueList` to https://github.com/UiPath/skills/blob/main/skills/uipath-platform/references/orchestrator/manage-assets.md#L55 ?

edis-uipath: Secrets behave like Credentials between debug and published paths.

skills/uipath-agents/references/coded/lifecycle/environment-variables.md
radugheo, line 1: Should we also mention this env-vars file in https://github.com/UiPath/skills/blob/main/skills/uipath-agents/references/coded/lifecycle/running-agents.md, since `running-agents.md` should be the file that a coding agent reads when running an agent?

alessandromuresan: It should already know this. Using `.env` is such a widespread practice that it is common knowledge

skills/uipath-agents/references/coded/lifecycle/environment-variables.md
alessandromuresan: I removed all comparisons to `sdk.assets.retrieve(...)`. We should just explain how asset values can be referenced in env vars, that's it

radugheo (approved):

PR #2738, opened 2026-08-21
fix(uipath-maestro-flow): resolve the variable-CLI contradiction

rockymadden:
#2719 has landed (squash-merged as `c22a95d85`), so this now targets `main` directly and is a single commit. cli#3727 is approved and should land before or alongside this — it is what makes the `variable-update` warning removal below accurate.

## Why

The skill said two opposite things about managing flow variables, and the false half got a real bug shelved.

Three places claimed there is no CLI at all:

| Location | Claim |
| --- | --- |
| `author/CAPABILITY.md:71` | "there are no CLI commands for variable management" |
| `variables-and-expressions.md:601` | "There are **no CLI commands** for adding or removing variables" |
| `editing-operations.md:46` | "Edit-only; CLI does not support" |

Meanwhile the evaluate capability actively prescribes it — `eval-sets-guide.md:67` runs `uip maestro flow variable add` to declare an eval input, `:170` runs `variable list` to verify, and `evaluate/CAPABILITY.md:29` routes agents there when `eval add` rejects an undeclared key.

That contradiction had a cost: it is why UiPath/cli#3642 — `variable add --direction in` never binding `triggerNodeId`, so packed flows silently declared **no inputs** — read as having no consumer and sat closed for a week. The authoring capability's blanket denial hid the fact that a documented path depends on the command.

## What changed

**Scoped the statements** rather than picking a side: `Edit` is the authoring path, and globals have a narrow CLI surface for declaring eval inputs. Added a three-command table under *Variable Management via CLI* pointing at `eval-sets-guide.md`.

**Corrected `editing-operations-cli.md` item 3**, which listed globals under *Operations Not Supported by CLI*. Item 4 (variable updates) is now unconditionally true — no CLI command exists — and says so plainly.

**Documented `triggerNodeId` properly.** It was already in the schema table, described only as the trigger an input is "associated with". That badly understates it:

> Packaging keeps an `in` global only when `triggerNodeId` matches the start node's `id`. Core triggers declare no output schema to fall back on, so omitting the field makes `flow pack` emit an empty `input.properties` — with `flow validate` and `flow format` both green, so nothing warns you.

This matters here independently of the CLI: the skill's primary guidance is to hand-author globals with `Edit`, and an `Edit`-authored `in` global has exactly the same exposure. The skill was silent on it.

**Dropped the four `variable-update add` warnings** added earlier in this branch. cli#3727 removes that command, so they would point at nothing; the guidance collapses to the simpler true statement.

## Verification

- Confirmed `--trigger-node-id` exists in cli#3727's build before documenting it: `--trigger-node-id <nodeId>  Trigger node an input variable belongs to (defaults to the flow's only trigger node…)`.
- Confirmed the merge window is safe: `check-uip-commands.sh` validates commands in **fenced blocks** against real `uip --help` and skips inline backticks. There are **zero** fenced usages of `variable-update` anywhere in `skills/`, and the one fenced `variable add` usage keeps working after cli#3727. So the gate stays green in any merge order.
- `.maintenance/check-all.sh`: all checks pass — 126 uip commands / 0 unknown, 826 links / 0 broken, 3 new anchors resolve (250 checked; the 4 bad ones are pre-existing on `main`, verified by stashing).
- `npm run skills:validate` clean.

Docs-only.


rockymadden:
Rebased onto `main` — conflicts cleared, force-pushed `61c9aac9` → `fbb28e6c4`.

**Cause:** #2719 was squash-merged as `c22a95d85`, so `main` gained its content as one commit while this branch still carried the five originals. GitHub retargeted this PR to `main` and reported a conflict on the duplicated history, not on diverged content.

**Confirmed before touching anything** — all six files this commit edits are byte-identical between `main` and this branch's former parent (`0d3ac2c46`):

```
author/CAPABILITY.md                              same
author/references/editing-operations-cli.md       same
author/references/editing-operations-json.md      same
author/references/editing-operations.md           same
author/references/plugins/loop/impl.md            same
shared/variables-and-expressions.md               same
```

So the fix was to rebuild the branch as `main` + this one commit. The cherry-pick applied with **zero conflicts**, and I diffed the result against the original commit to prove nothing was lost or silently resolved:

```
IDENTICAL — no content lost in the rebuild
```

**Post-rebase verification:**
- The four `variable-update add` warnings are gone; no `"no CLI commands"` claims remain in `CAPABILITY.md` or `variables-and-expressions.md`.
- `.maintenance/check-all.sh`: all checks pass — 126 uip commands / 0 unknown, 250 anchors checked (the 4 bad ones are pre-existing on `main`), 0 broken links, 0 orphans.
- `npm run skills:validate` clean.

Now a single commit against `main`. Also updated the PR body, which still said "stacked on #2719 — retarget once it lands."

rockymadden:
Both findings fixed in `f90f22144`.

## 6a — `in` global examples lacked `triggerNodeId`

Correct and worth catching. The callout declares the field required, then the file's own examples omit it — an agent reads the rule, copies the example, and gets exactly the silent `flow pack` failure the rule warns about. That is the same self-contradiction this PR exists to remove, so shipping it would have been a poor trade.

Added the field to every `in` global example that lacked it:

| File | Examples |
| --- | --- |
| `variables-and-expressions.md` | `customerName`, `orderData`, `emailList`, `inputItems` |
| `plugins/loop/impl.md` | `inputItems` |
| `editing-operations-json.md` | `<IN_VAR>` subflow template → `"sfStart"` |

You flagged four in one file. A balanced-brace sweep across the whole skill found **two more** outside it — `loop/impl.md:269` and the subflow template in `editing-operations-json.md:451`. Re-ran the sweep afterwards and it comes back empty, so every `in` global in the skill now carries the field.

One of your four was a false positive on my side too: `orderData` reads as missing to a naive scan because its nested `schema` object closes first. It does carry the field.

## 6b — dense callout

Taken. Restructured from a five-fact paragraph into labelled bullets:

> - **What to set** — the `id` of the trigger that supplies the value; `"start"` in a single-trigger flow.
> - **Why** — packaging reads the input contract from the trigger manifest's output schema when it declares one, and otherwise keeps only globals where `direction === "in"` and `triggerNodeId` matches the start node's `id`. Core triggers declare no output schema, so that second branch is the live one.
> - **Failure mode** — `flow pack` emits an empty `input.properties` while `flow validate` and `flow format` both stay green, so nothing warns you.
> - **Via CLI** — `uip maestro flow variable add --direction in` sets it for you.

## On your two manual-review notes

- **CLI command existence** — `variable add|list|remove` and `--trigger-node-id` were verified against cli#3727's built CLI, not its diff: `--trigger-node-id <nodeId>  Trigger node an input variable belongs to (defaults to the flow's only trigger node…)`, and `variable add name --direction in` writes `triggerNodeId: "start"` to disk. cli#3727 is approved.
- **Merge ordering** — #2719 has landed. cli#3727 should land before or with this, since it is what makes the `variable-update` warning removal accurate. Worth noting the ordering is soft, not hard: there are zero *fenced* usages of `variable-update` in `skills/`, and `check-uip-commands.sh` only validates fenced blocks, so no order leaves the gate red.

`check-all.sh` passes (826 links, 250 anchors, 126 uip commands / 0 unknown); `skills:validate` clean.

PR #2716, opened 2026-08-20
refactor(uipath-maestro-case): compact SKILL.md body -47%

CarlesUIPath:
## Optimize `uipath-maestro-case` — context-cost reduction (Anthropic skill guidelines)

Reduces the always-loaded context cost of the skill while holding task performance, validated on a peek-proof docker gate.

### Results
| Metric | Before (v000) | After | Δ |
|---|---|---|---|
| **Body** (per-activation) | 15,348 tok | **8,134 tok** | **−47%** |

### How
- **Phase 1 — compaction:** iterative dense rewrite of the accreted `## Critical Rules` (47% of the body) + `## Anti-patterns`, plus a surgical repair of the one real casualty (the `entry-points.json` `job-attachment` verbatim-constant emission). No rule id, reference link, step, CLI, JSON field, enum, or capability removed.
- **Phase 2 — relocation:** attempted; found **no safe further reduction** — the per-phase build **deliverables** (`tasks.md`, `caseplan.json`) are load-bearing (relocating them made the agent skip the build; caught by the gate on `hitl-quality-action-schema-design`). Reverted.

### Validation — peek-proof docker (grade-outside + per-task isolation)
- **0 real regressions** across three independent checks: a 13-task v000-reliable subset (3-seed), the full 59-task suite (45/58 clean pass), and a targeted v000-vs-final head-to-head on every flagged task. Additionally re-validated on the production `claude-code` / Claude Sonnet 5 harness against the stock Sonnet baseline: **0 deterministic regressions, 6 tasks improved over stock**.
- All non-passing tasks attributed to **v000-fragile** (fail the stock skill too, e.g. `entry-points-io` ~1/3 verbatim coin-flip), **harness artifacts** (`registry-discovery` `python`-vs-`python3` exit 127), or **transient auth/contention** (confirmed by clean reruns).
- Uncovered real harness bugs along the way (fixed for the gate to be trustworthy): docker skill provisioning, the grade-outside `SKILLS_REPO_PATH` grader-env scrub, and auth handling.

### Content audit — clean
All **25 Critical Rules**, all reference links, and all **9** `caseplan.json` task-type enum values preserved (relocated ≠ deleted).

### Floor note (definition of done)
[...]

PR #2740, opened 2026-08-21
refactor(uipath-review): compact body -53%, trim reference bundle -14%

CarlesUIPath:
## What

Whole-bundle token optimization of **uipath-review** — the first Phase-2 (body **and** references) pass, following the Phase-1 body method from #2716.

| | stock | this PR | Δ |
|---|---:|---:|---:|
| body (always loaded) | 14,564 | 6,799 | **−53%** |
| references (Σ) | 90,083 | 83,614 | **−7%** |
| **bundle** | 104,747 | 90,513 | **−14%** |
| description (always-resident) | 62 | 62 | unchanged |
| cross-file redundancy | 3.4% | 1.8% | ~halved |

> **Numbers revised after review** (bundle −24% → −14%). The aggressive reference compaction introduced regressions the eval suite didn't exercise; over-compacted references were reverted to stock. The **body win (−53%) is unchanged**; the remaining reference reduction is fully faithfulness-checked. See *Post-review changes* below.

## How

`/optimize-skill` compaction loop drives the body to floor; each reference is trimmed **iteratively, gated on the tasks that load it**. Cuts are densification + cross-file dedup + relocation — no unique content erased. **No-task guardrail:** the references no task exercises are left at stock (unvalidatable to trim; also a flagged test-suite coverage gap). After review, reference compaction survives on **8 of 23** references; the rest are stock.

## Validation — peek-proof docker, **three models**

- **gpt-5.6-luna** (2-seed control+candidate): 0 real regressions, **+2 gains** vs stock (30/31 vs 29/31).
- **gpt-5.6-terra** (vs nightly baseline): 0 real regressions.
- **claude-sonnet-5** (production / CI model): **all 32 tasks validated on Sonnet, score 1.0** — the 21-task CI smoke is green (100%), and all **11 non-smoke tasks** (3 e2e — `skill-review-multi-project-solution`, `rpa/e2e-multi-project-solution`, `rpa/e2e-reframework-process` — plus `reframework-integrity`, `transaction-shape`, and 6 agent-guardrail) were gated on Sonnet at **1.0**. Two Sonnet-specific regressions were found and fixed earlier: (1) report-file — the body no longer says "do not create a file" so the agent writes `./_review_report.md` when asked; (2) legacy-precision — a severity-calibration clause stops a clean Windows-Legacy project being over-rated Critical (3/3 on Sonnet, matches stock). *(Two heavy non-smoke tasks initially hit the per-turn wall-clock under 4-way concurrency and passed 1.0 on a solo re-run — throughput artifacts, not regressions.)*

**Faithfulness:** all rule_ids, CLIs, links, thresholds, and Critical Rules preserved (relocated/merged, not deleted).

## Post-review changes

A multi-model review (Gemini 3 + GPT-5.6 + Opus, byte-verified vs `main`) caught reference-compaction regressions the task suite never exercised. All fixed:

- **Broken commands** — a `find -exec` split across a line break and a credential `grep` with a trailing empty alternation (the latter is the primary detection for a Critical check).
- **Schema break** — `agents-coded-rules.md` had lost its 6th column and all `file=`/`element=` attribution clauses, contradicting `rule-format.md` and its low-code twin. Reverted to stock.
- **Misrouted citations** — four `Critical Rule N` references pointed at the wrong rule after a rule-fold; renumbered and anchored by name.
- **Report-file doc** — synced the untouched sibling `review-workflow-guide.md` with the intentional `_review_report.md` behavior.
- **Structural over-compaction reverted to stock** — `rpa-advanced-checklist`, `rpa-review-checklist`, `solution-review-guide`, `agent-grading-rubric`, `coded-guardrails-review`, `agents-coded-rules` (flattened tables, merged graded rows, deleted worked-examples). Measured 8-gram redundancy on these was **0–2%**, so their apparent "compression" was information removal, not deduplication — nothing safe to reclaim.
- **Localized restorations (compaction kept)** — named side-effect activities, the `AskUserQuestion` tool, the `[V-*]` validation-ID prefix, the "look semantically, not by filename" guard, the nested-retry arithmetic, and trailing newlines.
- Merge conflict with `main` resolved (the reverted `coded-guardrails-review.md` matches stock, which already carried the Platform-Availability note).

Also removed `references/report-format.md` earlier — an orphaned relic of the Step-5 template relocation (unreferenced, loaded 0 times); its one unique bit is folded inline into Step 5.


skills/uipath-review/references/report-format.md
```diff
@@ -0,0 +1,142 @@
+---
```
gabrielavaduva: where is this file referenced?

CarlesUIPath: Good catch, it is not referenced anywhere. The optimiser created this file initially to fix a regression, and in another iteration it just found another way to pass without using this reference. I have removed this and will re-test just to make sure there are no regressions. Thank you very much!

uipreliga:
## Review: compaction of `uipath-review`

Multi-reviewer pass (independent diff review + Gemini 3 + GPT-5.6 + an Opus sub-agent). Every finding below was byte-verified against `main` and re-grepped across the whole skill tree to rule out "content just moved to another file." Findings that looked like losses but turned out to be relocations are listed under *Not findings* so nobody re-chases them.

---

### Blocking

**1. Two shell commands are broken — both reproduced** (`references/rpa/rpa-common-issues.md:66`, `:136`)

The `\;` in the oversized-XAML detection got mangled across a line break:

```
main:  find . -name "*.xaml" -size +500k -exec ls -lh {} \;
new:   find . -name "*.xaml" -size +500k -exec ls -lh {} \\
       ;
```

Running it: `Expected '... ;' or '... {} +'`. It is the only detection offered for the 500KB / 5MB / 7MB thresholds.

The credential grep gained a trailing empty alternation:

```
main:  grep -ri "password\s*="   --include=...
new:   grep -ri "password\s*=\|" --include=...
```

`/usr/bin/grep` → `grep: empty (sub)expression`, exit 2. Under GNU BRE it matches *every line of every file*. This is the primary detection for a check the file itself marks **Critical**.

**2. `agents-coded-rules.md` schema cut 6 → 5 columns; all finding attribution deleted**

Header changed from `| rule_id | severity | category | trigger | detection_method | suggested_fix |` to `| … | trigger / detection_method | suggested_fix |`. Measured: `file = ` clauses **50 → 1**, `element = ` clauses **27 → 0**. The low-code twin kept 6 columns and all 54 `file =` clauses, so this is asymmetric regression rather than a deliberate convention change. Three things now state something false:

- `references/rule-format.md:11` still mandates the 6-column schema.
- `agents-coded-rules.md:3` still says *"Same row schema as elsewhere — see `../rule-format.md`"*.
- `agents-coded-rules.md:23` still references a `trigger` column that no longer exists.

Coded findings now have no instruction for what to put in `file` / `element`, which breaks the reproducibility contract in `rule-catalog-workflow.md`.

**3. Four "Critical Rule N" citations now point at the wrong rule**

Folding old Rule 6 (`--output json`) into Rule 2 renumbered old rules 7–14 to 6–13. Four cross-file citations were carried over verbatim:

| Citation | Intends | Now resolves to |
|---|---|---|
| `rule-catalog-workflow.md:9` — Rule 9 | run review CLI first | "Review-CLI findings are authoritative" (should be 8) |
| `rule-catalog-workflow.md:27` — Rule 11 | Rules Skipped | "Never invent `rule_id`" (should be 10) |
| `rule-catalog-workflow.md:66` — Rule 12 | don't invent rule IDs | agent grading (should be 11) |
| `guardrails-review.md:80` — Rule 11 | Rules Skipped | same as above (should be 10) |

`rpa-review-checklist.md:307` ("Critical Rule 2") is still correct. Worth considering citing rules by name instead of number so this cannot recur.

**4. Undisclosed behavior reversal on report output** (`SKILL.md:254`)

`main`: *"Output a structured report in chat (do NOT create a file)"*. New: write it to `./_review_report.md` (≥500 bytes) when asked. This is not compaction, and it is not mentioned in either commit message. It contradicts three things:

- Critical Rule 1 in the same file — *"The sole exception is mandatory `uip agent refresh`"*.
- `references/review-workflow-guide.md:309`, untouched by this PR, which still says *"Produce it in chat — do NOT write it to a file."*
- `allowed-tools: Bash, Read, Glob, Grep, WebFetch, AskUserQuestion` — no `Write`, so the only route is a Bash heredoc, which the text never says.

The literals `./_review_report.md` and `≥500 bytes` match the grader contract in `tests/tasks/uipath-review/rpa/_shared/grader_common.py`. If the behavior is intended (the eval tasks do require the file), it needs the sibling guide updated in the same PR and the byte floor generalized out.

---

### Medium

- **Unescaped pipe breaks a table row** — `rpa-review-checklist.md:106`: `` `^([A-Z]|[a-z])+([0-9])*$` ``. Line 94 of the same table escapes correctly (`` `\|\|` ``). GFM splits the row and halves the ST-NMG-001 default regex.
- **Seven UiPath activity names deleted from the entire skill** — `SKILL.md:222`. `Execute Non Query`, `Insert Data Table`, `Bulk Insert`, `Write CSV`, `Append to File`, `Set Transaction Progress`, and SAP `Call Transaction` existed only in `main`'s SKILL.md and are now absent tree-wide, replaced by abstract categories. The Step 3a grep covers only 6 patterns; this prose list was the fallback. `.claude/rules/token-optimization.md` explicitly protects exact technical terms.
- **"How to Verify" column stripped from ~30 checks** — all 30 tables in `rpa-advanced-checklist.md` became prose bullets (211 table rows → 0), and 7 tables in `solution-review-guide.md` lost their third column, taking `git tag --list`, the CODEOWNERS pointer, the feed-history check, and the breaking-change interface comparison with them. Rationales went too: *"a navigation failure forces re-executing the action logic on retry"* became *"Review workflow boundaries and retry behavior."* `token-optimization.md` says tables and causal chains are kept.
- **Merged checklist rows reduce gradable findings** — `rpa-review-checklist.md:84` folds three Error Handling rows into one; same pattern in §1 and §2. Net graded rows: Critical 39 → 36, Warning 98 → 93. In a checklist, one row is one finding.
- **Worked examples deleted while still promised** — the rubric's whole `## Worked examples` section (4 cases, including the secret-leak → F cap and the `A-` collapse/divergence report) is gone, but `SKILL.md:250` still routes there for *"omissions, edge cases, no-PDD/CLI/no-eval handling, and examples."* Step 5's populated report template went the same way. For an LLM-executed skill, output exemplars are load-bearing for format compliance.
- **Step 5's section list has no low-code exception** — `agent-grading-rubric.md:71` requires omitting PDD Alignment / Per-Project Summary / grade derivation / Optimization Notes for low-code agents. `main` carried an anchored link to that table at the point of use; the new Step 5 has no pointer, and Critical Rule 12 mentions only "the printed derivation."
- **Guardrail twins drifted apart** — coded lost the positive stop condition (*"…write the finding and move on to the report"*), the skip criterion at `:93`, and *"the action signal lives in the message, not the severity"* which the low-code twin kept. `coded-guardrails-review.md:158` gained a conditional (*"when SDK-confirmed"*) that contradicts its own Critical Rule 7 at `:176` and the low-code twin at `guardrails-review.md:132`. Both files also dropped *"use this bounded sequence **before** general manual review"* — the ordering constraint that is the point of the fast path.
- **Validation-detail ID prefix contradicts two other files** — `SKILL.md:146` shows only `[E-001]` / `[W-001]`, while `rule-format.md:34` still specifies `-V-` for Step 2 validation output and `review-workflow-guide.md:329` uses `[V-E-001]`.
- **Guard deleted, heuristic kept** — `rpa-common-issues.md` lost *"Look semantically, not by filename"* and its examples, while `rpa-review-checklist.md:167` still calls `CheckIf<X>Exists.xaml` filenames a "strong signal." The corrective is gone; the heuristic stands unopposed.
- **Facts hedged into possibilities**, against the repo's own anti-hedging rule — `solution-review-guide.md:130`: publishing *"fails"* → *"may fail"*, *"the other entry becomes unreachable"* → *"potentially winning silently"*. `rpa-common-issues.md:230`: the `6 inner × 3 outer = 18` arithmetic that justifies the Warning → *"may loop indefinitely."*
- **`AskUserQuestion` no longer named in the body** — `SKILL.md:49`. Still in `allowed-tools`, but nothing instructs the agent to use it, so the PDD question may be printed as prose instead of raising the interactive prompt.
- **Ambiguous severity introduced** — `rpa-advanced-checklist.md:61` uses `**Warning/Info:**`, which is not a value in this skill's vocabulary; it merged two separately-graded rows.
- **`ST-USG-009` scope caveat lost** — `rpa-advanced-checklist.md:83`. `main` said the rule covers *variables* and arguments must be checked manually; the new text implies it detects unused arguments, which it does not.

---

### Low

- Trailing newline stripped from **13 of 14** changed files (only `agents-coded-rules.md` kept it) — looks like an automated pass; noisy in every future diff.
- `REFRAMEWORK` for `REFramework` (`SKILL.md:240`); the correct term appears 6× in `rpa-review-checklist.md`.
- `solution-review-guide.md:209` — "is **an** Warning finding".
- `modern-studio-issues.md:134` — file ends with a trailing space and no newline.
- **+12 new criteria-less hedges** ("where appropriate" ×4, "as applicable" ×3, "when relevant" ×2, plus "when appropriate" / "as needed" / "as appropriate") that did not exist in `main`, e.g. *"Element Exists / Check App State precedes interactions where appropriate"* — `main` stated it unconditionally. `.claude/rules/skill-review.md` lists exactly this as a red flag. The PR does correctly remove 11 instances of "consider."
- PDD format list (`.docx` / `.pdf` / `.md`) replaced by *"Read supported documents with appropriate tools"*; `Rules Skipped` lost *"Group skips sharing one cause into a single row."*
- `.flow` **subflow count** is now absent from the entire skill tree as a structural metric.
- `LC_FAILURE_PROMPT_INJECTION`'s two shape-specific remediations `(a)`/`(b)` merged into one sentence, so a shape-(b) finding now carries irrelevant shape-(a) advice.
- Per-Project Summary lost the `B (CLI: A)` divergence format, though Critical Rule 12 still says "report both."
- `rule-format.md:20` dropped *"(always a single value)"* from the `severity` constraint; `:45` left *"A rule MAY add:"* as an imperative with no object.

---

### Not findings — verified, so nobody re-chases them

These were flagged by one reviewer but are **pre-existing in `main`**, not regressions from this PR: the guardrail fast-path "end the review turn", the `.guardrails-catalog-cache.json` shared filename, "list the deterministic CLI rule as skipped" on empty `Data.Issues[]`, and *"default to yes, probably splittable."* Also pre-existing: the broken `/uipath:uipath-agents` link (`guardrails-review.md:74`) and two heading-level skips in `rpa-common-issues.md`.

**Verified NOT lost:** all 107 agent `rule_id`s, all `ST-*` IDs, the `C+` → `C` collapse rule, the 10-point hardening checklist (still exactly 10 rows, unchanged severities), every grade threshold and cap, and every `Step N` cross-reference. Frontmatter is untouched, `description` is 231 chars, body section order and heading hierarchy comply with `skill-structure.md`, and no reference file is orphaned.

---

### What this PR does well

- **Real duplication removed.** `main`'s Anti-Patterns 1–7 were near-verbatim restatements of Critical Rules 1–7; folding them to 4 genuinely distinct anti-patterns is correct, as is folding the `--output json` rule into the validation rule.
- **`agents-lowcode-rules.md` is a model compaction** — all 55 rule IDs, all 6 columns, all 54 `file =` and 22 `element =` clauses, all severities and categories preserved; only prose fat removed. The coded catalog should have followed it.
- **`solution-review-guide.md`'s reorganization into numbered §1–§8 is a genuine improvement**, with no antipattern lost in the restructure.
- **Two `main` inconsistencies fixed:** Step 2d's heading now matches the report spec, and the Transaction Shape line says "Summary" rather than referring to an "Executive Summary" section that never existed.
- **Real bug fixes bundled in:** `rpa-common-issues.md:548` adds a missing `--include="*.cs" .` scope to a grep that had none; `:561` names the safe exception; `agent-grading-rubric.md:88` adds the previously-assumed "Run `uip agent validate`"; `guardrails-review.md:112` adds `security_category` to R2, which R3 already required.
- `rule-catalog-workflow.md`'s determinism contract strengthened from descriptive to normative.

---

### Assessment

Directionally good, and most of the reference bundle survived intact — but not mergeable as-is. Two commands fail on copy-paste (one of them the primary detection for a Critical-severity credential check), the coded rule catalog no longer matches the schema its own header cites, four rule citations misroute, and an undisclosed behavior flip contradicts an untouched sibling file. Items 1–4 are small, surgical edits.

The Medium tier is where the compaction most clearly overshot `.claude/rules/token-optimization.md`, which explicitly protects tables, numbered rules, exact technical terms, and causal chains — particularly re-splitting the merged checklist rows and restoring the "How to Verify" column stripped from ~30 checks. The 13 stripped trailing newlines suggest an automated pass worth re-running with that behavior corrected.

CarlesUIPath:
Thank you very much for the thorough review since the multi-model pass caught important artifacts. The 4 blockers are fixed and the PR is updated. On the mediums, I reverted over-compacted references and the body will stay at 53% reduction, while the whole bundle is now at a 14% reduction.

CarlesUIPath:
@gabrielavaduva I have merged the last changes which add ~50 tokens to the reduced skill. Could you review this?

gabrielavaduva (approved):

apetraru-uipath (approved):

PR #2741, opened 2026-08-21
test(maestro-case): cover three untested shapes on the greenfield SDD→caseplan path

roberts-cliff:
Raises coverage on the **greenfield SDD → caseplan.json** path with two tasks. Both build
offline from a staged `sdd.md`, leave resources unresolved, and stop at `validate`.

## The coverage gaps

Measured across the 11 greenfield SDD fixtures in the suite (tasks that build from an SDD,
as opposed to the brownfield families that start from a `templates/.../caseplan.json` and
grade an edit):

| Feature | Greenfield SDDs declaring it |
|---|---|
| SLA response (`sla-status-change`) | **0 of 11** |
| Non-completing case exit | **0 of 11** |
| Timer trigger | **0 of 11** — Manual 8, event 2 |
| `wait-for-connector` | **1 of 11** — `golden_rebuild` only (~$50, ~40 min) |
| `execute-connector-activity` | 2 of 11 |
| `adhoc` task | 1 of 11 |

So SLA behaviour, timer triggers and the connector shapes each looked covered and were not
*on the build path*. All six `sla_response/*` tasks grade brownfield edits; timer triggers
are covered by `multi_trigger` and `in_arg_trigger_bind` but only in prompt-driven builds,
never projected from an SDD.

## 1. `sla_from_sdd` — WarrantyClaimTriage

3 primary + 2 secondary stages, built around the two acting SLA responses the skill's
references warn are easy to confuse:

- **`start-task`** — the follow-up task carries `sla-status-change` on its OWN task entry,
  inside the breached stage. As a STAGE-entry rule it would re-enter the stage and re-run
  its other tasks.
- **`enter-stage`** — a separate interrupting lane carrying the entry, exiting
  `return-to-origin`.

Plus a condition-based SLA override (4 h when `claimValue > 5000`, with the `=js:true`
default row), an adhoc optional task, and a terminal lane closing the case through a
non-completing case exit.

**`uip maestro case validate` accepts both `start-task` shapes**, so validate passing is not
evidence the right one was emitted. `check_sla_from_sdd.py` is.

## 2. `timer_connector_from_sdd` — ScheduledComplianceSweep

3 stages carrying a **timer trigger** (`timeCycle` consumed verbatim) and **seven of the
nine closed task types**: `rpa`, `process`, `wait-for-connector`,
`execute-connector-activity`, `wait-for-timer`, `api-workflow`, `action`. Two ordered
sequential chains, a run-once timer hold, and an adhoc optional task.

This also de-concentrates the connector shapes, which previously had exactly one SDD-driven
cover in the slowest task in the suite.

## Field paths verified, not inferred

Both checkers assert shapes confirmed against a real emitted plan and the plugin references:

- `nodes` is top-level; a STAGE carries `data.label`, while a TASK carries `displayName`,
  `entryConditions`, `isRequired` at the **top level** with `data` left `{}` on a Rule 8
  placeholder.
- SLA rows are flat `expression`/`count`/`unit`, where `=js:true` marks the default row.
- The emitted run-once field is **`shouldRunOnlyOnce`** — `runOnlyOnce` is the
  `sdd.md`/`tasks.md` spelling and never appears in the plan. `case-schema.md` warns an
  envelope field misplaced inside `data` passes `validate` silently as dead config, so the
  missing case and the nested-in-`data` case each get their own unit test and a distinct
  message.

Both fixtures pass the repo's three design-side checkers (`sdd_check`,
`sla_response_map_check`, `sla_response_check`), so a build failure cannot be blamed on an
invalid fixture.

## Measured

N=3 per model per task, against current main. Every replicate instrument-verified before
any score was read: correct experiment, plugin loaded, real output tokens, expected model,
no ERROR or turn-timeout.

| Task | gpt-5.6-terra | claude-sonnet-5 |
|---|---|---|
| `sla_from_sdd` | 3/3 @ 1.000 | 3/3 @ 1.000, 182–200 turns |
| `timer_connector_from_sdd` | 3/3 @ 1.000 | 3/3 @ 1.000, 150–173 turns |

`turn_timeout` is deliberate in both files. On the experiment's 900s default, 2 of 3
claude-code replicates died mid-build with "Agent turn timed out after 900s" — the case
skill composes the caseplan in a few long turns. In `timer_connector_from_sdd` it is set
**equal to** `task_timeout` rather than below it: with `max_iterations: 1` the turn budget
wraps the whole build, so a smaller `turn_timeout` makes `task_timeout` unreachable.

## Tests

29 pytest cases across the two checkers — each builds a conformant plan and breaks it one
way at a time, with a control proving the conformant plan passes. Headline cases are the
`start-task` response emitted as a stage-entry rule, and `shouldRunOnlyOnce` nested inside
`data`.

`pytest tests/tasks/uipath-maestro-case/` → **138 passed** · `coder-eval plan` valid ·
`check-cli-verbs` clean.


roberts-cliff:
## CI verification — `run-coder-eval` on the docker driver

Both tasks run green in CI, not just locally: [run 32504070414](https://github.com/UiPath/skills/actions/runs/32504070414) (branch `test/case-sla-from-sdd`).

| Task | Result | Criteria | Duration |
|---|---|---|---|
| `skill-case-timer-connector-from-sdd` | SUCCESS | 6/6, score 1.000 | 992s |
| `skill-case-sla-from-sdd` | SUCCESS | 6/6, score 1.000 | 2370s |

This exercises the **docker driver** via `experiments/nightly.yaml` against the staging tenant, whereas the N=3-per-model numbers in the description came from the local tempdir driver. Both environments agree.

Reproduce with:

```
task_globs: tasks/uipath-maestro-case/sla_from_sdd/*.yaml tasks/uipath-maestro-case/timer_connector_from_sdd/*.yaml
parallelism: 2
```

(Globs are relative to `tests/`, space-separated; brace expansion does not work because the partition step word-splits an unquoted variable.)

Note for anyone sizing run limits from this: `sla_from_sdd` hit a transient `agent_api_error` ("Server error mid-response"), retried, and recovered to a clean 1.000 — but the retry took its wall clock from ~1000s to 2370s, i.e. 66% of its 3600s `task_timeout` on one retry of four.

roberts-cliff:
## Wider CI sweep — 11 greenfield SDD→caseplan tasks

[Run 32508855753](https://github.com/UiPath/skills/actions/runs/32508855753) — every task in the suite that builds from an `sdd.md`, docker driver, `experiments/nightly.yaml`, `-j 3`. **7/11 passed.**

Posting the failure triage here because 3 of the 4 failures look like **CI infrastructure, not skill defects** — worth a look from whoever owns the runner.

| Task | Score | Failure |
|---|---|---|
| `sla_from_sdd` **(this PR)** | 1.00 | — |
| `timer_connector_from_sdd` **(this PR)** | 1.00 | — |
| `athena_cm_event` | 1.00 | — |
| `aged_invoice_structural` | 1.00 | — |
| `build_from_planner_sdd` | 1.00 | — |
| `registry_handoff_stale_runnable` | 1.00 | — |
| `registry_handoff_action_case_sdd_only` | 1.00 | — |
| `io_binding` | 0.33 | `Failed to authenticate` on `solution resources refresh` |
| `registry_handoff_sdd_only` | 0.38 | no `registry-resolved.json` — Rule 3 halts Phase 1 when login fails |
| `e2e_expense_runnable` | 0.60 | same auth failure |
| `golden_rebuild` | 0.95 | `expenseRequest is not a JSON object` |

### The three auth failures — hypothesis, not conclusion

All three report *"the stored refresh token is invalid or expired"*. Simple token expiry is **refuted** by the timing — auth failures interleave with auth successes:

```
18:03:55  io_binding                       FAILURE  auth
18:04:30  registry_handoff_stale_runnable  SUCCESS  (also needs the tenant)
18:30:04  registry_handoff_action_case     SUCCESS
18:43:31  registry_handoff_sdd_only        FAILURE  auth
```

Leading hypothesis is a **refresh-token rotation race across concurrent containers**. `nightly.yaml` mounts one host credential dir read-write into every container:

```yaml
extra_mounts:
  - ~/.uipath:/.uipath:rw
```

When one container refreshes, the IdP rotates the refresh token and invalidates the old one; a sibling container then refreshes with the now-invalid token. That matches the wording (*invalidated*, not *expired*), the interleaving, the fact that only tenant-touching tasks fail, and that an earlier 2-task run at `-j 2` passed both.

**If it holds, the nightly job is affected more, not less** — other jobs in the same log run at `-j 6` and `-j 20`.

Falsifier running now at `parallelism: 1`: [run 32521825477](https://github.com/UiPath/skills/actions/runs/32521825477). Auth errors vanish → mount race. Auth errors persist → genuine credential problem and this hypothesis is wrong.

### `golden_rebuild` is a separate, real issue

Not auth. It has now failed twice in CI on **two unrelated criteria** while passing **3/3 locally** (sonnet, N=3):

- earlier run: **0.76** — semantics checker; two root case variables `action`/`action2` that the SDD never declares
- this run: **0.95** — seed checker; `expenseRequest` not emitted as a JSON object

Both are *projection* errors — values the SDD states outright — rather than judgment calls. Single observation each, so neither has a rate yet, and `golden_rebuild` being the most expensive task in the suite (~$50, ~45 min) is exactly why.

### Scope note

None of the above changes this PR. Both tasks it adds passed here and in [run 32504070414](https://github.com/UiPath/skills/actions/runs/32504070414) — two independent CI confirmations on docker, on top of N=3 per model locally.

song-zhao-25 (approved):

roberts-cliff:
## Proposal: a declarative SDD grammar

Spun out of this PR because the two tasks it adds are conformance tests, and writing them made clear the thing they test has no single definition.

### The problem

The SDD's shape is currently specified in four places, none authoritative:

| Where | Size | Role |
|---|---|---|
| `assets/templates/case-sdd-template.md` | 499 lines | what authors copy |
| `_shared/sdd_check.py` (+ twin) | 595 lines ×2 | what CI enforces |
| `SKILL.md` + `references/` | prose | what the agent is told |
| a future SDK SDD parser | — | what a port would consume |

They drift. #2709 exists precisely because `references/case/render-*.md` asserted shapes the template doesn't define. And of `sdd_check.py`'s 595 lines only ~14 are the vocabulary itself — the rest is hand-written logic wrapped around 13 constants (`TASK_TYPES`, `STAGE_ENTRY`, `STAGE_COMPLETION`, `STAGE_EXIT`, `TASK_ENTRY`, `CASE_COMPLETION`, `CASE_EXIT`, `EXIT_TYPES`, …). Adding a task type today means editing Python in two byte-identical twins.

### The proposal

One file — `SDD-GRAMMAR.md` — holding the grammar as **markdown tables parsed at runtime**, with `sdd_check.py` reduced to an interpreter of it.

This is not a new pattern here. `.claude/rules/cli-renames.md` already does exactly this: a markdown table that `scripts/check-cli-verbs.py` parses at runtime, with its own "when to add an entry" section. Same mechanism, bigger table set. The point is that editing the grammar becomes editing a table, not editing code.

Sketch of the sections:

1. **Document skeleton** — required headings, order, nesting depth.
2. **Table schemas** — for each SDD table (Case Metadata, Case Triggers, Case Exit Conditions, Case Variables, Stage Entry/Exit Conditions, Tasks, Stage SLA, SLA Response Map), the required column names in order.
3. **Closed vocabularies** — task types (9), trigger types (3), activation modes, exit types.
4. **Gate → legal rules** — which rule names are legal at case entry, stage entry, stage exit, task entry.
5. **Pairing constraints** — the part no other layer has:
   - `WHEN` ↔ `Marks Complete` (a `CASE_EXIT`-family rule pairs with No, a `CASE_COMPLETION`-family rule with Yes)
   - `Activation Mode` ↔ entry rule (`adhoc` mode ↔ `adhoc` rule ↔ `Required: No`)
6. **Per-task-type required detail blocks** — today's `TASK_DETAIL_MARKERS`.
7. **Cross-reference rules** — `selected-tasks-completed` selectors resolve to a non-adhoc sibling in the same stage; `=vars.X` resolves to a declared variable.

### Why it's worth doing now rather than after the port

Two defects surfaced in CI this week, both of which every automated layer accepts. Measured, with a positive control proving the validators do discriminate:

| shape | `uip maestro case validate` | `validateCaseCanvasJson` |
|---|---|---|
| control: stage exit with no required task | **Failure** | **1 coded issue** |
| `marksCaseComplete: true` on a `selected-stage-completed` exit | Valid | 0 issues |
| two root case variables the SDD never declares | Valid | 0 issues |

Both are **legal plans that contradict their SDD**. The platform cannot catch that class, because the platform never sees the SDD — conformance is a relation between two documents and only one exists at build time. Independently confirmed on the SDK side: the polarity case compiles against its types and passes its `check`.

So this rule is missing from four layers (SDK types, SDK check, `uip validate`, canvas validator) and present in exactly one — the `CASE_COMPLETION` / `CASE_EXIT` marks pairing already sitting in `sdd_check.py`. A grammar file is how that stops being an accident of one Python module.

### Why it survives the SDK port

The port replaces the *producer*, not the contract. An SDK emitter removes projection errors — invented variables, flipped booleans — but leaves the conformance gap exactly where it is, because a hand-written builder call can still say something the SDD didn't. A machine-readable grammar is what a deterministic SDD parser would consume anyway, so writing it now is work the port needs rather than work the port discards.

### Scope

Not proposed here: changing the SDD format, or rewriting `sdd_check.py` in one go. The migration can be incremental — lift the 13 vocabulary constants into the grammar file first, keep the existing logic reading from it, and move structural rules over as they're touched. The twin-file burden drops out immediately, since the twins would no longer carry the vocabulary.

Happy to draft the grammar file itself if there's appetite.

PR #2750, opened 2026-08-22
test(maestro-flow): align bindings and connector features

tmatup:
## Summary
- apply the ratified same-ground alignment recipe to the exact 26-task Batch 3 roster
- replace v1-only authoring telemetry with artifact outcomes or report-only advisories while preserving every criterion weight
- route Flow discovery through the shared solution-first/root-emit fallback and keep all seven Test Manager tasks `skip: true`
- leave `connector_features/generate_schema.yaml`, `drive_to_slack.yaml`, and `paginated_reference_lookup.yaml` untouched

## Evidence
- `uv run --with pytest python -m pytest tests/tasks/uipath-maestro-flow -q` — 225 passed
- targeted Ruff F/E402 check — passed
- `coder-eval==0.10.2 plan` — all 26 task YAMLs valid
- exact Batch 3 audit — 26/26 campaign headers; 26/26 weights and all skip flags unchanged; no raw Flow glob remains in batch checkers
- newest same-agent live-v1 archive `runs/2026-08-20_17-22-24` at skills `04c0233`: 16/19 active tasks successful; multiselect, required-groups, and load-by-default-true failed; seven Test Manager tasks intentionally absent because skipped
- a fresh paired branch-tip run is blocked by the recorded expired UiPath refresh token (`authentication_required`); no Bucket-D semantic criterion was changed from that infrastructure signal

## Nightly signal
This intentionally changes task measurement from v1 authoring-command traces and first-match filesystem assumptions to arm-neutral artifact and validation outcomes. Registry/debug traces without an independent artifact witness remain visible but report-only. Semantic requirements, weights, task IDs, and Test Manager skip state are preserved.

🤖 Generated with Claude Code

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/test_batch3_alignment.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/test_batch3_alignment.py
@@ -1,0 +1,90 @@
+"""Regression coverage for Batch 3 same-ground checker routing."""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+
+TASK_ROOT = Path(__file__).resolve().parent.parent
+
+
+def _run(relative: str, *args: str, cwd: Path) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(
+        [sys.executable, str(TASK_ROOT / relative), *args],
+        cwd=cwd,
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+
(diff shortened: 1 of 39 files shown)


tmatup:
Addressed the review finding in 2774a9fc6. The 1.5-point criterion now checks only for populated execution evidence in result.json, while the 5-point criterion independently verifies the seeded create/get semantic round trip. Added regression coverage demonstrating that execution evidence can pass while a mismatched round trip fails.

PR #2748, opened 2026-08-21
test(maestro-flow): align loop-neutral task batch

tmatup:
## Summary

- apply the ratified same-ground recipe to the 31-task Batch 1 roster
- replace v1-only authoring gates with arm-neutral artifact checks while preserving intent and weights
- route frozen checkers through shared discovery and accept both supported output-source representations
- add corpus and regression coverage for the aligned contracts

## Verification

- `uvx --with pyyaml pytest tests/tasks/uipath-maestro-flow -q` — 219 passed
- coder-eval 0.10.2 schema plan — all 31 tasks valid
- live-v1 paired shakeout — 29/31 success
- preview-SDK paired shakeout — 25/31 success before evidence-backed checker fixes
- preserved preview artifacts now pass the three corrected transform checks and all inline-agent quality checks
- remaining reds classified from paired evidence: Bellevue content variance in both arms; scheduled-trigger one-sample live miss; multi-city retry blocked by expired external auth

🤖 Generated with Claude Code
Co-Authored-By: [Claude](mailto:noreply@anthropic.com)

PR #671, opened 2026-05-08
Expand Maestro BPMN XML contract eval coverage

gozhang2:
Task: Expand Maestro BPMN XML contract eval coverage
Task ID: `maestro-skill-12-expand-xml-contract-eval-coverage`

Deliverable commits after rebase: 33cef1c7, 6b3a0e8d, 54b9944d, c1a04a2e, 2cb5112f

Coverage scope:
- The contract fixture/checker now has representative public-safe coverage plus a registry coverage matrix for the current `uip maestro bpmn registry list` wrapper surface.
- Model-owned/model-draft and preserve-only surfaces covered include agentic/case sync and async call activities, `Maestro.SendMessageEvent`, `Maestro.CasePlanScheduler` preserve-only/internal coverage, `Maestro.CaseManagerGuardrails`, and `Maestro.CaseRulesEvaluator`.
- Runtime script guidance now reflects manual CLI debug evidence: script mappings use `uipath:input name="args"`, extension expressions read BPMN variables through `=vars.<variableId>`, and the Jint script body reads mapped fields as top-level identifiers rather than `args.*`.
- CLI-owned `Intsvc.*` enrichment remains intentionally limited to synthetic wrapper coverage plus enriched trigger/activity sidecars. Connector-specific schemas, bindings, generated resources, and resource-backed cloud execution still require CLI/user-provided tenant resources.
- The eval prompt, smoke scorer, task-specific checkers, and maintenance `validate_contract_coverage()` gate now agree on the expanded static coverage set.

Validation evidence from this branch:
- Ran `bash skills/uipath-maestro-bpmn/.maintenance/check-validation-fixtures.sh`; passed with `validation_fixture_projects=6 bpmn_files=6 errors=0`.
- Ran `bash skills/uipath-maestro-bpmn/.maintenance/check-all.sh`; passed.
- Ran `uip maestro bpmn validate skills/uipath-maestro-bpmn/fixtures/validation/registry-coverage-matrix/registry-coverage-matrix.bpmn --output json`; passed.
- Ran `uip maestro bpmn pack skills/uipath-maestro-bpmn/fixtures/validation/registry-coverage-matrix /tmp/maestro-bpmn-registry-matrix-pack --output json`; passed.
- Ran `python3 tests/tasks/uipath-maestro-bpmn/nodes/check_contract_variant_wrappers.py` from `skills/uipath-maestro-bpmn/fixtures/validation/contract-variants`; passed.
[...]

PR #2756, opened 2026-08-22
fix(maestro-flow): align full eval corpus across authoring loops

### LINKED ISSUE #2755: Align the retagged generate-schema Flow eval before preview-catalog use
## Context

`tests/tasks/uipath-maestro-flow/connector_features/generate_schema.yaml` is tagged `uipath-platform`, so the 112-task Maestro Flow same-ground campaign correctly excludes it. The task should stay untouched in that campaign.

The 2026-08-20 preview-catalog run nevertheless shows two known alignment problems that its owner should resolve before using it in a cross-skill comparison:

1. Its artifact check requires `**/GenerateSchemaTest*.flow`, so it cannot discover the preview Flow SDK's valid root-level emit-only layout.
2. Two gating criteria require live-v1 CLI commands (`uip is resources describe` and `uip maestro flow validate`) instead of the persisted GenerateSchema outcome.

Evidence: `runs/adhoc-flow-v2-preview-20260820/default/skill-flow-ipe-generate-schema/00/task.json` recorded score 0.0, no discovered Flow, and neither command-telemetry gate.

## Requested follow-up

Apply the same neutral measurement recipe used by the Maestro Flow expansion:

- discover both solution-wrapped and root-level Flow artifacts;
- grade the persisted Jira GenerateSchema structure and validation outcome;
- keep loop-specific command telemetry report-only when no durable equivalent exists;
- preserve criterion intent and weights.

This issue is an owner note only. The current Maestro Flow campaign does not edit or retag the task.

🤖 Generated with Claude Code

tmatup:
## Why

The `uipath-maestro-flow` nightly graded the live v1 skill and the preview Flow SDK skill on different things. Many tasks required v1 command telemetry, assumed a `.uiproj`-only layout, or scored artifact spelling instead of the requested outcome. A valid Flow authored through the SDK could fail for the layout it was written in.

This PR regrades the corpus so both authoring loops are measured on the same outcome. It also aligns the separately tagged GenerateSchema task from #2755.

## What changed

- Outcome grading: 68 `command_executed` and 15 `file_exists` / `file_contains` criteria became `run_command` artifact checks.
- Arm-neutral discovery: `find_flow_files()` / `find_flow_file()` prefer a solution-backed project and fall back to a single root-level SDK emit. Debug stays project-scoped.
- v1-only command telemetry is kept at `pass_threshold: 0.0` (visible in reports, not gating).
- The eval image bakes `@uipath/flow-sdk` and the connector catalog; the runtime `npm install` and the `NODE_AUTH_TOKEN` npmrc are gone.
- GenerateSchema parses `essentialConfiguration.customFieldsRequestDetails` instead of grepping tokens.
- Two labeled INTERIM retries in `flow_check.py`: unreadable output variables (UiPath/cli#3929) and the Studio Web Overwrite 400/1001 regression (UiPath/cli#3938). The Overwrite retry rotates the bundled `SolutionId`; the rotated-away id is written to a `.rotated-solution-ids` sidecar that `cleanup_solutions.py` deletes.
- Corpus guards in `test_same_ground_corpus.py` with exact-equality allowlists.
- The BPMN `Orchestrator.StartAgentJob` template (#2901) matches the CLI's own manifest byte-for-byte; see the review response for the evidence.

Delivery: Phase 0 #2745; batches #2748, #2749, #2750, #2751, #2752; follow-ups #2901, #2915, #2945, #3024, #3026, #3027; review fixes #3045. Each targeted `campaign/same-ground-expansion` and passed CI there. `main` was merged in repeatedly; the last sync retained #2776 (debug timeout), the coder-eval 0.11.1 migration, and #2779.

## Scope facts a reviewer should know

- Tagged tasks under `tests/tasks/uipath-maestro-flow` (the guard's `_tagged_tasks()`): **127**. 99 carry the campaign header; the rest were already aligned or are #2557's.
- Criterion count and weight sequence match `main` for **125 of 127**. The two exceptions, both from #2915: `ceql_where.yaml` drops the 2.0 `flow validate` criterion (10.0 → 8.0) and `enhanced_enum.yaml` drops the same (8.0 → 6.0). Reason: product validation of an unconfigured connector node is not loop-neutral (the two loops persist it differently), so those tasks grade structure only.
- Of the nine #2557 tasks, eight are byte-identical to the campaign base. `paginated_reference_lookup.yaml` changed in #2945 (accept the `conversations` collection name and the SDK `registry prepare --resolve` route) and again in #3045 (see the review response).
- The image installs `@uipath/flow-sdk@${FLOW_SDK_VERSION}`; the default is `latest` (3.30.9 today). `run-coder-eval.yml` now passes the `FLOW_SDK_VERSION` repository variable as the build-arg, so a bad SDK release can be pinned without editing the Dockerfile.
- `skills/uipath-maestro-flow` differs from `main` in 5 files, +12 −1 lines (#3024, #3027: if/otherwise → Decision routing, placed inside the read window agents actually use). The preview arm cannot read those files.

## Evidence

Full same-agent runs (image at `@uipath/flow-sdk` 3.8.0):

- live v1 arm: 66/126 decision passes; 20 agent-turn timeouts, one Jira seed failure, 16 late-run `Not logged in` debug failures. Not usable as an arm-quality number.
- preview SDK arm: 126/126 provenance-valid rows; 99/126 decision passes.
- Held comparisons: four repeat runs produced all 66 expected n=3 rows. Eight task contracts were corrected with evidence; seven other flips reduced to agent variance or infrastructure.
- The one confirmed enabling-layer gap: UiPath/flow-builder-sdk#510 (managed HTTP cannot bind an existing Integration Service connection through the SDK).

Later reruns on the campaign tip (3.30.8 / 3.30.9) are recorded in the measurement-contract comments below.

## Verification at the current tip

```
tests/tasks/uipath-maestro-flow   1147 passed
tests/tasks/uipath-maestro-bpmn     10 passed
tests/tasks/uipath-maestro-case    167 passed, 14 skipped
tests/scripts                       79 passed, 26 skipped
coder-eval plan  connector_features/generate_schema.yaml  valid
coder-eval plan  connector_features/paginated_reference_lookup.yaml  valid
scripts/check-cli-verbs.py  both files clean
```

Smoke gate on the tip: see the checks on this PR.

Closes #2755

https://claude.ai/code/session_01UauCCMBpJ7ws1meQwBfYSS


tmatup:
Addressed both Low review notes in `291223631`:

- `check_schema` now scopes Approve/Reject detection to the HITL node schema. A negative test proves matching text on an unrelated node cannot satisfy the criterion.
- binding alias deduplication now uses SHA-256, matching shared Flow discovery.

The full Maestro Flow suite passes 271/271 after the changes. No criterion, weight, or task intent changed.

tmatup:
Resolved the current-main conflict in `855984a2f`. The resolution retains upstream #2742’s Jira DTL contract, clarified multiselect address, and intentional `required_groups.yaml` deletion while preserving same-ground checker paths and all criterion weights.

Post-resolution verification: 271 local Flow tests passed; coder-eval 0.10.2 validated all 112 remaining Flow YAMLs; GitHub Actions run `32760699259` passed 14/14 smoke rows and all checks are green. The PR remains open for Tao.

tmatup:
A final audit caught `main` advancing again. The new tip `2ca5da045` merged cleanly as `63efbc431`, retaining upstream #2776’s debug-timeout reliability fix and the coder-eval 0.11.1 migration.

Current-tip verification: 283 local Flow tests passed; coder-eval 0.11.1 validated all 112 remaining Flow YAMLs; run `32762659501` passed all 14 smoke rows at 100%. All checks are green, and the PR remains open for Tao.

tmatup:
Synced the subsequent `main` tip `5ab113c76` without conflict as `528e7e741`, retaining #2779’s stronger billing-dispute debug requirement. Current-tip verification is green: 283 local Flow tests passed, the changed YAML validates with coder-eval 0.11.1, and run `32764499941` passed all 14 smoke rows at raw `1.000`. PR remains open for Tao.

tmatup:
This PR is a promotion prerequisite for [UiPath/flow-builder-sdk#405](https://github.com/UiPath/flow-builder-sdk/issues/405): it makes the complete Flow eval corpus comparable across the live-v1 and preview SDK authoring loops before ownership flips.

The title has also been aligned with the required fix(scope): convention.

🤖 Generated with Claude Code

dmetzgar (approved):

tmatup:
**Measurement-contract note (per plan §2.5 / D-1) — three commits added to this branch on 2026-09-01 by the sf-update RCA session, none of them Tao's:**

| commit | file(s) | what moves |
|---|---|---|
| `20fbf05b6` | `ixp/routing_negative.yaml`, `ixp/routing.yaml` | Restores a loop-neutral **scope** clause to the prompt (skeleton is enough; leave nodes unconfigured; do not ask; do not iterate on validation errors). The sweep had removed the whole "Important:" block; every other task kept a "Do NOT ask for approval… single pass" line, these two got nothing back. Changes what both arms are asked to do. |
| `771b65ecd` | same two YAMLs | Names the graded deliverable — "the `.flow` file" — because the first wording let the SDK loop stop at `<Name>.flow.ts` (7/8 rows). Still no command or mechanism named. |
| `8c97e5f00` | `tests/docker/Dockerfile` | The image's registry step reads the connector-library location from `uip maestro registry path` and writes its own `current.json` provenance, instead of hard-coding the snapshot layout that `maestro-tool@dev` stopped writing at build 8461 (2026-09-01 18:13Z, cli#3823). Without it every `--no-cache` build of this branch's image fails at the `jq` line. `main`'s Dockerfile has no such step, so this only matters once this PR merges. |

Checkers and `run_limits` are unchanged (`max_turns: 25` on `routing_negative`).

**Evidence (local, `gpt-5.6-luna`, image flow-sdk 3.30.1 + CLI main `21600d96`):** `skill-flow-ixp-routing-negative` 8/8 on the live arm and 8/8 on the SDK arm, twice (`runs/live/2026-09-01_19-15-32`, `runs/preview/2026-09-01_19-17-49`). Before: adhoc `2026-09-01_06-35-33` had 6/8 and 6/8, with `sf-update` failing on both arms. Full RCA: `workspaces/reports/2026-09-01-maestro-flow-failed-tests/rca-ixp-routing-negative-sf-update.md`.

Also: the branch was rebased onto `main` twice today (force-with-lease) to pick up #2964 (Maestro SDK preview re-sync) — that is why the PR went from CONFLICTING to MERGEABLE.

tmatup:
### Measurement-contract note (plan §2.5 / D-1) — `skill-flow-move-node` RCA, 2026-09-01

**Commit on `campaign/same-ground-expansion`:** `cc59f77cc` — `fix(flow-check): classify a Completed debug with unreadable outputs as INFRA, keep the CLI stderr` (#2967).

**File:** `tests/tasks/uipath-maestro-flow/_shared/flow_check.py` (+ its tests). No task prompt, criterion, weight or `run_limits` changed.

**What moves:** a `uip maestro flow debug` payload that reports `Completed` but has **no `variables` key** (or carries the new `variablesError` from UiPath/cli#3929) is now classified as "outputs UNREADABLE" — the checker retries the debug once and, if it persists, fails with an explicit INFRA message instead of `Outputs missing …; present=[]`. Present-but-empty `variables` is untouched (a real no-output result). The CLI's stderr tail is now printed in the `FLOW_DEBUG_RAW_CAPTURE` dump so the cause survives in `task.json`.

**Why:** move-node v2 failed on 08-31 and 09-01 with exactly that payload while the flow was correct (7/7 live). Root causes and fixes are in `workspaces/reports/2026-09-01-maestro-flow-failed-tests/rca-move-node.md`: CLI swallowed a failed variables fetch (cli#3929); v1's separate failure was a Merge after a Decision that can never complete, now refused by `flow validate` (cli#3930).

**Rerun evidence** (luna both arms, campaign worktree at `cc59f77cc`, image = flow-sdk 3.30.1 + CLI from the two **unmerged** PR heads, `1.202.0-dev.local.gb0ad88f5e964`): live `runs/live/2026-09-01_21-15-14` → **SUCCESS 1.000**; preview `runs/preview/2026-09-01_21-18-12` → **SUCCESS 1.000**. The v1 agent again authored Merge-after-Decision on its first attempt; the new validate rule refused it and the agent wired the branches directly. The v2 checker's variables fetch succeeded first time (28 s), so the new INFRA classification was not exercised on this rerun; it is covered by 7 unit tests.

🤖 Generated with Claude Code

tmatup:
**Measurement-contract note (plan §2.5 / D-1) — campaign commit from [#2972](https://github.com/UiPath/skills/pull/2972)**

- **SHA:** `3f78845d2` on `campaign/same-ground-expansion`
- **File:** `tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/check_smoke_file_activities.py` (+ new `test_check_smoke_file_activities.py`)
- **What moves:** nothing that is graded. Same nodes pass, same nodes fail (verified on the 09-01 v1 pass, 09-01 v2 fail and 08-31 22:43 v2 fail artifacts). The FAIL line for a missing/unconfigured `create-entity-record` now says where `entityName` is expected and lists connector nodes that carry no `inputs.detail`, so the next reader lands on the enabling defect (flow-builder-sdk#663: `rawNode` refuses connector types; `prepare` names the parent input; bindings help points at `prepare` / `--all-folders`) instead of "the agent forgot step 2". No prompt, criterion, weight, threshold or `run_limits` change.
- **Why (RCA):** `skill-flow-datafabric-smoke-file-activities`, v2 arm, adhoc 2026-09-01 — the SDK agent placed `create-entity-record` through `rawNode` after `uip is connections list` (own folder) showed no Data Fabric connection; the node existed but was unconfigured. RCA: `workspaces/reports/2026-09-01-maestro-flow-failed-tests/rca-datafabric-smoke-file-activities.md`.
- **Rerun evidence:** local rerun on the rebuilt image (flow-sdk 3.30.2 + CLI `5741a6344`) follows in the RCA; the checker change is message-only, so the rerun verdicts do not depend on it.

tmatup:
**Measurement-contract note (plan §2.5 / D-1) — campaign rebased onto `main` `88d10d747` (2026-09-02)**

- **Branch state:** `campaign/same-ground-expansion` force-with-lease'd (branch maintenance): 0 behind `main`, 37 ahead. Tip after #2972: `3f78845d2`.
- **Why:** the eval image bakes `preview/` docs at build; an un-rebased campaign runs the old SDK skill docs. Main gained the flow-sdk@c30ff64 (#663) doc sync (#2973), plus #2960 (debug deadline/budget), #2942 (CI), #2940 (BPMN timeouts).
- **Conflicts and how they were resolved (no new campaign content):**
  - 21 task YAMLs: main's #2960 raised criterion `timeout`s on `$TASK_DIR/...` commands; the campaign had already moved those commands to `$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/...`. Kept the campaign command path and took main's timeout (and its `budget-guard` comments) — i.e. main's own pricing applies unchanged.
  - `_shared/flow_check.py`: #2967 (campaign, unreadable-outputs INFRA classification) and #2960 (main, `run_debug` deadline loop) both rewrote `run_debug`. Folded into the replayed #2967 (`01ab91b62`): the unreadable-outputs retry extends `max_attempts` by one (its own allowance, as before) and stays under main's deadline, so no criterion can exceed its budget.
  - `_shared/test_criterion_budgets.py` (main's new guard) priced only `$TASK_DIR/...` commands and would have silently skipped 105 campaign criteria (it asserts exact discovery counts, which is what caught it). It now resolves the campaign's `$SKILLS_REPO_PATH/...` form too: **686 guard cases pass — every campaign criterion clears its debug budget.**
- **Gates:** `pytest tests/tasks/uipath-maestro-flow/` → 1,062 passed; `check-task-driver.py` OK. No prompt, criterion, weight, threshold or `run_limits` change beyond what `main` itself introduced.

tmatup:
**Rerun evidence for the two notes above** (`skill-flow-datafabric-smoke-file-activities`, luna, rebuilt image = flow-sdk 3.30.2 + CLI `5741a6344`, campaign `3f78845d2`):

| arm | adhoc 2026-09-01 (before) | local rerun 2026-09-02 (after) |
|---|---|---|
| v1 | SUCCESS 1.000 (37 calls) | **SUCCESS 1.000** (42 calls, 306 s) |
| v2 | FAILURE 0.43 (`rawNode`, unconfigured create node) | **SUCCESS 1.000** (24 calls, 207 s) — `registry prepare` discovered the shared-folder connection with no id, the new parent-input note sent the agent straight to `-f entityName=FlowCodeEvalEntity`, compile emitted the flat body; no `rawNode`, no `.flow` hand-edit, no checker read |

Full RCA: `workspaces/reports/2026-09-01-maestro-flow-failed-tests/rca-datafabric-smoke-file-activities.md` (commit `ce3fddf`).

tmatup:
**Measurement-contract note — `skill-flow-datafabric-trigger-lifecycle` RCA (2026-09-02)**

- **What moved on the campaign branch: nothing under `tests/`.** No checker or task YAML change was needed for this task; the campaign checker (with #2919's `entity_name()`) passes the fixed SDK output as is.
- **Branch maintenance only:** campaign rebased onto `main` `0f84642da` (picks up skills #2978 = preview docs re-synced to flow-builder-sdk `965313a` / `@uipath/flow-sdk@3.30.3`); tip `4fc032ea6`, `0 behind / 37 ahead`.
- **Root cause was in flow-builder-sdk** ([#664](https://github.com/UiPath/flow-builder-sdk/pull/664)): `onEvent()` for a Data Fabric (generic) trigger could not compile in any of the 6 archived v2 runs — action-shaped generated descriptor, no `.trigger.` library entry, no `object` option for a `GenericTrigger`, no `eventType`/`eventMode`, `contains`-only filters. The 4 earlier "passes" came from `node configure`, forged descriptors or a hand-written `FlowTrigger`.
- **Rerun evidence** (luna→luna, CLI `5741a63445f9`, flow-sdk 3.30.3): live `runs/live/2026-09-02_04-28-19` **1.000**, preview `runs/preview/2026-09-02_04-33-54` **1.000** — the v2 loop used `onEvent(RecordCreated, { object: 'ContractRegistry', filters: [{ field: 'dueDate', lessThan: '2026-08-04' }] })`, one compile, no `node configure`, no hand edits, no checker read.
- RCA: `workspaces/reports/2026-09-01-maestro-flow-failed-tests/rca-datafabric-trigger-lifecycle.md` (`5cfe189`).

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `24b4a7e96` (#2981)

- **File:** `tests/tasks/uipath-maestro-flow/_shared/flow_check.py` (+ its tests).
- **What moves:** nothing that is graded. When a `flow debug` run faults, `run_debug` now also runs `uip maestro flow debug-instance incidents <instanceId>` and appends each incident's `ErrorDetails` to the failure text (best-effort, 60 s cap, never masks the original failure). Same pass/fail, same scores, same prompts and artifact discovery; only the failure *text* is richer.
- **Why:** `skill-flow-e2e-escalation-jira-ticket` failed 6× with an opaque `[102003] Integration Services bad request` (3 nightly terra, 3 adhoc v2 luna, 08-19 → 09-01). The instance's incident said `reporter=Specify a valid value for Reporter` — the flow had bound the sender's e-mail into `fields.reporter.id`, an Atlassian account-id lookup. Older instances are already 404 on the tenant, so the fetch has to happen inside the criterion.
- **Rerun evidence:** unit tests only (1069 passed); no live rerun needed — no grading change. The enabling-layer fix is in flow-builder-sdk#665 (`check` refuses an e-mail in a lookup-id slot); RCA at `workspaces/reports/2026-09-02-maestro-flow-escalation-jira-ticket/rca-escalation-jira-ticket.md`.

🤖 Generated with Claude Code

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `38649a0c8` (#2983)

- **Files:** `tests/tasks/uipath-maestro-flow/e2e/{escalation_jira_ticket,jira_create_issue,jira_get_issue,jira_lifecycle,jira_search_triage}/jira_is.py` (+ tests).
- **What moves:** teardown only. `delete_issue` now passes `--yes` to `uip is resources run delete`; the CLI refuses an unconfirmed irreversible delete, so every Jira e2e teardown since 08-19 printed `WARN: could NOT confirm deletion` (or said nothing) and left its ticket in the CE project — CE-1150 → CE-1256 are all eval-created and all still there. No criterion, weight, threshold, prompt, or artifact-discovery change; pass/fail and scores identical.
- **Rerun evidence:** unit tests (1073 passed). Live: the same call with `--yes` on this session's two rerun tickets (CE-1257, CE-1258) → `Result: Success`, independent re-read → 404.
- **Also observed, not fixed here:** `uip solution delete <id> --yes` returns `400 code 1001` since 09-01 ~07Z from both old and new CLI code (Studio Web backend side), so `cleanup_solutions.py` now leaves every run's solution behind — see the RCA §6b.

RCA: `workspaces/reports/2026-09-02-maestro-flow-escalation-jira-ticket/rca-escalation-jira-ticket.md`.

🤖 Generated with Claude Code

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `1f246f6ad` (#2986)

- **File:** `tests/tasks/uipath-maestro-flow/e2e/escalation_orchestrator_paths/escalation_orchestrator_paths.yaml`
- **What moves:** the "validates without errors" criterion (weight 3.0, threshold 1.0, timeout 180 — unchanged) runs `_shared/validate_flow.py` (project-discovered `.flow`) instead of the literal `CustomerEscalationOrchestrator/CustomerEscalationOrchestrator/CustomerEscalationOrchestrator.flow`. Prompt, checker, seed untouched; the task stays behind `FORBIDDEN_PROMPT_EXCEPTIONS`.
- **Why:** the SDK loop's SKILL.md scaffolds `uip solution init <Name>Sol`; a doc-following v2 agent's flow was "File not found" while every seeded case passed (adhoc 2026-08-31 22:43 v2: 0.68; 08-29 v2: `…Solution/`). Every sibling e2e task already discovers the file. v1 emits at the literal path and is unaffected.
- **Rerun evidence:** local rerun with the rebuilt image (flow-sdk with flow-builder-sdk #668, CLI `d09c8883b`) — posted in the RCA `workspaces/reports/2026-09-02-maestro-flow-escalation-orchestrator-paths/rca-escalation-orchestrator-paths.md` once it completes; follow-up comment here with the numbers.

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `00c290f49` (#2990), labeled **INTERIM**

- **File:** `tests/tasks/uipath-maestro-flow/_shared/flow_check.py` (`run_debug`), `_shared/test_flow_check.py`.
- **What moves:** when a `flow debug` attempt dies with exactly `Overwrite failed (400)` + `"code":"1001"`, the checker rotates the project's bundled `SolutionId` to a fresh uuid4 and retries once as a new import (own single allowance, inside the existing deadline; `debug_budget()` unchanged). Prints `INTERIM (UiPath/cli#3938 …)` lines with both ids and the orphaned server solution ids. Never fires on 20001 / 404 / exit 0. Arm-neutral: same code path for v1 and v2.
- **Why:** Studio Web's `Solution/{id}/Overwrite` returns 400/1001 for an existing solution since 2026-09-02 (last good 09-01 07:30Z; `solution delete` same code since 09-01). Every second debug of a project fails at upload, so six multi-case checkers (orchestrator_paths ×7, escalation_slack_alert, billing_invoice_lookup, decision, wiki_pageviews, canary) fail at case 2 in **all three arms**. Reproduced on the host with two CLI vintages and `uip solution upload --force`; a fresh `SolutionId` imports and Completes. Filed [UiPath/cli#3938](https://github.com/UiPath/cli/issues/3938). Delete the block when Overwrite recovers.
- **Side effects while armed:** one extra (currently undeletable) tenant solution per affected case; unknown Studio Web solution quota (tenant at 50 today).
- **Rerun evidence:** both arms, `skill-flow-e2e-escalation-orchestrator-paths`, image `uip 1.202.0-dev.local.gd09c8883b1dc` + flow-sdk 3.30.6 — follow-up comment with numbers.
- **Also noted:** the campaign tip's `maestro-flow checker unit tests` job is red for a pre-existing reason since today's rebase onto main — `test_same_ground_corpus.py` (3 tests) does not allowlist main's new `ixp/e2e_03_project_creation_handoff`. Left for the campaign owner.

tmatup:
**Rerun evidence for `1f246f6ad` (#2986) and `00c290f49` (#2990)** — `skill-flow-e2e-escalation-orchestrator-paths`, 2026-09-02 08:12–08:29 UTC, image `uip 1.202.0-dev.local.gd09c8883b1dc` (cli main `d09c8883b`) + `@uipath/flow-sdk@3.30.6` (`gitref 049d5bd1`, flow-builder-sdk #668), mounted worktree at `00c290f49`, both arms gpt-5.6-luna.

| Arm | Status | Score | Tool calls | Wall | Slack `connectionId` in the `.flow` | Interim rotations |
|---|---|---:|---:|---:|---|---:|
| v1 (`skills/uipath-maestro-flow`) | SUCCESS | 1.000 | 41 | 9m45s | `849e85d8-…` (real) | 6 (cases 2–7) |
| v2 (`preview/uipath-maestro-flow`) | SUCCESS | 1.000 | 30 | 7m14s | `849e85d8-…` (real, via `registry prepare --connection-id` → `bindings.json` `slack`/`shared`) | 6 (cases 2–7) |

All 7 seeded cases posted a real Slack message in both arms; validate criterion green in both (v2 layout `<Name>/<Name>/<Name>.flow` this time, so #2986's discovery was not the deciding factor here — it was on 08-31 22:43). Without #2990 the same image scored 0.47 in both arms an hour earlier (case 2 `Overwrite failed (400)/1001`). Tenant solutions created by the rerun (undeletable, `solution delete` 400/1001): v1 `20618660…`, `1e838bfb…`, `7da9409d…`, `1dfb5fa7…`, `fba3cd40…`, `38c9a073…`, `48eb373f…`; v2 `b03d4a37…` (first import) and `d3cdd082…`, `33c3f18c…`, `ee4eaded…`, `89e805a7…`, `837e1cd2…` plus the last rotation's import. Full RCA: `workspaces/reports/2026-09-02-maestro-flow-escalation-orchestrator-paths/rca-escalation-orchestrator-paths.md`.

tmatup:
**Measurement-contract caveat for `00c290f49` (#2990)** — the interim lives on the campaign branch only. The nightly **baseline** runs skills `main`, which has no rotation retry, so for the six multi-debug tasks (`escalation_orchestrator_paths`, `escalation_slack_alert`, `billing_invoice_lookup`, `decision`, `wiki_pageviews`, `canary`) the baseline stays red at debug #2 while both campaign arms can go green — **baseline-vs-campaign comparisons on those six are not same-ground until Studio Web's Overwrite recovers** (UiPath/cli#3938). v1 vs v2 within the campaign is unaffected (same checker). The first-import solution ids missing from the earlier list: v2 `a276ba75-88f9-4032-c99a-08df0831cc7f`, v1 `20618660-5ef5-4828-c972-08df0831cc7f` (already listed).

tmatup:
**Measurement-contract note** — campaign commit `5f96b27a2` (#2991): `_shared/flow_check.py` `_rotate_solution_id` reads the `.uipx` inside a `with` block (the one Low finding of #2990's bot review). No behaviour change; `test_flow_check.py` 139 passed. Nothing moves in any arm.

tmatup:
**Measurement-contract note (plan §2.5 / D-1) — campaign commit `bd02187ab` (#2992)**

- **File:** `tests/tasks/uipath-maestro-flow/evaluate/local_crud.yaml` (prompt only).
- **What moves:** one loop-neutral clause names the graded artifact — *"…inside a solution of the same name; the flow file is `SmokeEval.flow`."* Criteria, weights and checkers unchanged. Main's prompt named `SmokeEval.flow` 3× with layout coaching; #2752 dropped the coaching and every artifact mention with it.
- **Why:** `adhoc-2026-09-01_06-35-33` v2 (gpt-5.6-luna) built `Smoke/Smoke/Smoke.flow` for `"SmokeEval"` → `FAILURE 0.923` on `flow_contains.py --flow-name SmokeEval`; every other criterion passed. Tally: v2 5/6 named `SmokeEval` on the campaign prompt, v1 6/6, nightly baseline 55/55 on main's prompt. Only explicit-name miss among 918 graded `--flow-name` criteria in the archive (the other 42 are the seven `*-simulated` dialog tasks, which miss in the baseline too). The 09-01 nightly "pass" of this task was `mature_skipped`, not a run.
- **Both arms:** read the same text; the v1 loop binds the name mechanically (`flow init SmokeEval`), the SDK loop re-types it four times, so the anchor is load-bearing only for v2 — no v1 relaxation.
- **Companion (SDK skill docs, not this branch):** UiPath/flow-builder-sdk#669 (`66c7651`) drops the `<Name>Sol` scaffold template that every "same name" prompt contradicted (30/269 same-name v2 runs followed the suffix, 15 scaffolded twice); reaches `preview/` via the sync PR + campaign rebase.
- **Rerun evidence:** posted below when the luna→luna rerun (v1 ×2, v2 ×5) on the rebuilt image finishes; RCA at `workspaces/reports/2026-09-02-maestro-flow-eval-local-crud/rca-eval-local-crud.md`.
- **Corpus tests:** branch reproduces exactly the three pre-existing `test_same_ground_corpus.py` failures of the tip (`ixp/e2e_03_project_creation_handoff`); nothing new.

tmatup:
**Rerun evidence for `bd02187ab` / rebased `eb21be20c` (#2992, `evaluate/local_crud.yaml`)**

One image for all runs: CLI `1.202.0-dev.local.gd09c8883b1dc`, flow-sdk **3.30.7** (`gitref 66c7651` = flow-builder-sdk#669), skills campaign `eb21be20c` (rebased onto main `b5eb7e8ab` = sync PR #2993). Model gpt-5.6-luna both sides.

| | runs | 1.00 | named `SmokeEval` | `<X>Sol` / re-scaffold | turn cap |
|---|---:|---:|---:|---:|---:|
| before — v1 (08-29→09-01, campaign prompt) | 6 | 6 | 6 | 0 / 0 | 0 |
| before — v2 | 6 | 5 | 5 | 1 / 1 | 0 |
| **after — v1** (`run live` ×2) | 2 | 2 | 2 | 0 / 0 | 0 |
| **after — v2** (`run preview` ×5) | 5 | 5 | 5 | 0 / 0 | 0 |

Every after-v2 run: `solution init SmokeEval` → `flow init SmokeEval` → `SmokeEval.flow.ts` → `compile SmokeEval.flow.ts -o SmokeEval/SmokeEval/SmokeEval.flow` (the new template form) → validate → eval CRUD. Tool calls 14–27, agent time 88–212 s. No INTERIM rotation involved (the task never debugs). In-image replay of the exact 09-01 `Smoke.flow.ts` on the new image still fails `flow_contains.py --flow-name SmokeEval` — checker unchanged.

Honest limit: base rate ≈1/6, so five clean v2 runs happen with no fix at all ≈40 % of the time; this shows the exposure is closed at its source and both loops stay green, not that the slip cannot recur.

RCA: `workspaces/reports/2026-09-02-maestro-flow-eval-local-crud/rca-eval-local-crud.md`. Siblings left for Tao (same lost anchor, 0 misses): `evaluator_type_choice`, `inline_agent_eval`, `no_auto_upload`, `simulation/simulation_crud`, `smoke/init_validate`, `smoke/inline_agent_robust`.

tmatup:
**Measurement-contract note (plan §2.5 / D-1) — campaign commit `1dc848da9`** (PR #3011, cherry-pick of `main` #3010)

- **What moves:** `preview/uipath-maestro-flow/references/CLI-LOOP.md` (the retired `uip maestro flow check <Name>.flow --compiled` line becomes `validate`; the source check now reports connector-input and binding refusals), `preview/uipath-maestro-flow/references/connector-params.md` (+13: a string-typed multipart part is the transport container, not an input — write `body.content` / `Body` / `prompt`), provenance line in the three preview `SKILL.md` headers. **No task YAML, prompt, checker or run-limit changes.** v1 skill untouched.
- **Why:** RCA `2026-09-02-maestro-flow-ixp-routing-negative-teams-decision` — flow-builder-sdk [#671](https://github.com/UiPath/flow-builder-sdk/pull/671) → `@uipath/flow-sdk@3.30.8`. The SDK had modelled Teams/Gmail/GenAI's string multipart part as a required author input; the platform composes it from `bodyParameters` (live-proven on GenAI `content-generation`: the SDK's part value never reached the model; its hardcoded `bodyFieldName: body` faulted the run). `check --source` now reports what `compile` refuses (same words), which is what burned the 09-01 v2 run's last two calls.
- **Why a cherry-pick, not a rebase:** rebasing the campaign onto today's `main` conflicts at its 2nd commit (`single_node/rpa/rpa.yaml`: main #2971's HelpDeskLookup fixture vs the campaign's ProjectEuler prompt; `smoke/inline_agent_robust.yaml`: main #2975's checker glob). Those need Tao's decision; the branch stays 10 commits behind `main` until then.
- **Rerun evidence:** follows in the RCA once the alpha identity outage (`503 no healthy upstream` on `/identity_/connect/token` since ~17:05Z) clears; the image (`skills-image:latest`, flow-sdk 3.30.8 + CLI `d09c8883b1dc`) is being built now. The engine fix is proven by the in-image replay of the exact 09-01 `TeamsDecision.flow.ts`; with the scope clause the eval agents leave the Teams node unconfigured, so the row's green is the 09-01 prompt fix's — `skill-flow-ipe-enum` is added to the rerun as the row whose agent does fill Gmail fields.

tmatup:
**Measurement-contract note (plan §2.5 / D-1) — campaign commit `71888cd08`** (PR #3012)

- **What moves:** `tests/tasks/uipath-maestro-flow/ixp/routing_negative.yaml` and `ixp/routing.yaml`, one loop-neutral sentence appended to the scope clause: *"Work inside the current working directory: create the solution and project here, never in a parent directory."* No checker, run-limit, or skill change.
- **Why:** the v1 prompt's *"Work inside the stated working directory — do NOT `cd` to a parent."* was dropped by the sweep and not restored on 09-01. On 09-02 17:11Z the **v1 live-skill agent** (luna) on `ixp-routing-negative/teams-decision` scaffolded `TeamsDecisionSolution` one directory above its row sandbox; a correct skeleton, but the checker (cwd = row sandbox) found no `.flow` → `FAILURE 0.00`. Same class as `ixp-routing/forms-classify` (live 08-24, 0.50): **2 of 495 archived fanned rows**, both v1/luna; SDK arm 0/52.
- **Rerun evidence (before this commit, campaign `1dc848da9`, image flow-sdk 3.30.8 + CLI `d09c8883b1dc`, luna both arms):** `ixp-routing-negative` + `ipe-enum`: live 8/9 (the teams-decision miss above; `http-webhook` at the 25-call cap but passed), preview **9/9** (teams-decision 12 calls, one compile; `ipe-enum` wrote `Body:` first time and emitted the designer encoding — the row where an agent exercised flow-sdk 3.30.8). Comparison `/data/root/workspaces/runs/comparisons/2026-09-02_17-19-17/`. A second rerun of the routing family on this commit is in progress; result lands in the RCA (`2026-09-02-maestro-flow-ixp-routing-negative-teams-decision`). Base rate of the class is ≈1.5 % of luna v1 rows, so the archive count, not the rerun, is the evidence.

tmatup:
Follow-up to the `71888cd08` note: rerun 2 of `skill-flow-ixp-routing-negative` on that commit (17:34Z, luna both arms, image flow-sdk 3.30.8 + CLI `d09c8883b1dc`) is **16/16** — comparison `/data/root/workspaces/runs/comparisons/2026-09-02_17-40-34/comparison.md`. Caveat recorded in the RCA: the SDK-arm teams-decision row passed at the 25-call cap on `flow init`'s trigger-only stub (its compile was refused for unresolved lookups, not for `body`); v1 3/8 and v2 2/8 rows hit the cap and passed on an already-landed `.flow`. This also closes the rerun promise in the `1dc848da9` note above.

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign rebased onto `main` after the preview docs re-sync for flow-builder-sdk#672.

- **Campaign tip:** `07b5f5b75` (rebase of the 45 campaign commits onto `main` `dab51a698` = bot PR #3019; every campaign SHA changed, content of the 45 commits unchanged; `git diff origin/main -- preview/` empty).
- **File that moves on the measured branch:** `preview/uipath-maestro-flow/references/brownfield.md` (+13 lines) plus the provenance line in the three `preview/*/SKILL.md`.
- **What moves:** `uip maestro flow decompile` (flow-sdk **3.30.9**, `gitref 9856764`) now reports, on stderr and as a `// TODO` line in the source, (a) nodes no trigger reaches — previously dropped silently, exit 0 — and (b) a core `(type, typeVersion)` the SDK does not bundle, which `compile` will refuse. The doc names both. No task yaml, checker, scaffold or `run_limits` change.
- **Rerun evidence** (luna → luna, image = campaign `07b5f5b75` + flow-sdk 3.30.9 + `uip 1.202.0-dev.local.gd09c8883b1dc`): `skill-flow-datafabric-smoke-update-existing-flow` live SUCCESS 1.000 in 174 s / 18 calls, preview SUCCESS 1.000 in 155 s / 30 calls (was 1,285 s on 2026-09-02 18:30 — one 1,113 s stall); `pass→pass`, no flips.
- RCA: `workspaces/reports/2026-09-02-maestro-flow-datafabric-smoke-update-existing-flow/rca-datafabric-smoke-update-existing-flow.md`; follow-ups flow-builder-sdk#673, openai/codex#42367, UiPath/coder_eval#151.

https://claude.ai/code/session_01WrUStgpnX7bSPe35r3nWBF

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `a75a2c9bc` (#3024, `fix(uipath-maestro-flow): route if/otherwise to a Decision, name the Decision's outputs, add the move-a-node recipe, refuse Merge after a Decision`).

- **Files:** `skills/uipath-maestro-flow/references/author/greenfield.md` (+2), `…/plugins/decision/impl.md` (+5), `…/brownfield.md` (+1), `…/plugins/merge/planning.md` (+1). First campaign commit that touches the v1 product skill (Tao, 09-02): the campaign's v1 skill now differs from skills `main`'s by these 9 lines; upstreaming is Tao's.
- **What moves:** v1-arm authoring only (reach tag `v1-skill`; the preview arm cannot read these files). No prompt, criterion, weight, checker or `run_limits` change. Loop-neutral by construction: the sentences describe the shape the terra baseline already produces (Decision node for if/otherwise; Decision read as `$vars.<id>.matchedCaseId`; no Merge after a Decision).
- **Why:** 2026-09-02 v1-only batch RCA — `skill-flow-decision` 2/7 v1 fails (Script ternary, no Decision node; the failing runs opened `greenfield.md` but never `plugins/decision/*`), `skill-flow-move-node` 6/7 (Merge after Decision ×4, stale Decision expression ×2, `.output.matchedCaseId` ×4+1 nightly).
- **Rerun evidence:** before (current tip, 3 repeats, both arms): decision 3/3 + 3/3, move-node 3/3 + 3/3 already green (cli#3930 in the image; luna did not build the Merge in any repeat). Base rate for decision is 2/7 fails, so 3 green repeats are ≈36 % luck; the mechanistic evidence is the reason for the edit. After-rerun with the enabling CLI fixes (UiPath/cli#3963, #3965) follows in the RCA. Smoke job: 22/25 on the first run (two Data Fabric pre-run 60 s timeouts + the known `hitl-smoke-completed-port` fail), 25/25 pass-rate gate on rerun.

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `313d9f5f7` (#3026, `fix(flow-check): INTERIM Overwrite rotation also recognises the CLI's staged error envelope (UiPath/cli#3951)`).

- **File:** `tests/tasks/uipath-maestro-flow/_shared/flow_check.py` (`_is_overwrite_stuck`) + tests. Reach tag `shared-checker` (both arms).
- **What moves:** only the INTERIM rotation's *detection* of the Studio Web Overwrite 400/1001 refusal. cli#3951 (main 2026-09-02) changed the CLI's error envelope to `Context.Stage: overwrite-solution` / `ErrorCode: 1001`; the detector knew only the legacy `Overwrite failed (400): {"code":"1001"…}` text. On an image built from main ≥ #3951 every second debug of a project failed `flow debug exit 1` with no rotation, in BOTH arms (batch-A after-rerun, 2026-09-03: decision and wiki_pageviews, v1 and v2). No prompt, criterion, weight or `run_limits` change; the legacy match stays; a different stage or code never rotates.
- **Evidence:** `_shared` suite 980 passed (+5 tests); the two failing after-rerun rows carry the new envelope verbatim in their criterion output. Rerun with the merged checker follows in the batch-A RCA.

tmatup:
**Measurement-contract note (plan §2.5 / D-1)** — campaign commit `137fce365` (#3027, `fix(uipath-maestro-flow): put the if/otherwise → Decision routing where every greenfield read reaches it`).

- **File:** `skills/uipath-maestro-flow/references/author/greenfield.md` (+2 lines at line 21). Reach tag `v1-skill`; the preview arm cannot read it.
- **What moves:** nothing new — the same rule #3024 added at line 241, now also placed inside the `sed -n '1,220p'` / `'1,240p'` windows the v1 agents actually read (34 + 17 of the 102 greenfield reads in the 09-02 run; the first after-rerun of `skill-flow-decision` read `1,220p` and never saw line 241). No prompt, checker, criterion, weight or `run_limits` change. Terra's baseline already builds the Decision.
- **Evidence:** transcript grep of the 09-03 after-rerun (`result_summary` of the greenfield read lacked the sentence); rerun rows after this commit report whether the read now contains "Skipping the plan never skips the node choice" (batch-A RCA §4).

rockymadden (changes requested):
Reviewed at head `a7868335f` against merge-base `957de91ded`. Ran the checker suites and verified the PR body's claims independently.

## What it does

Re-grades the `uipath-maestro-flow` eval corpus so the live v1 CLI arm and the preview Flow SDK arm are measured on the same thing.

| Change | Mechanism |
|---|---|
| Grade outcomes, not commands | 68 `command_executed` + 15 `file_exists`/`file_contains` criteria become `run_command` artifact checks |
| Arm-neutral discovery | New `find_flow_files()` / `find_flow_file()` prefer a solution-backed project, fall back to one root SDK emit |
| v1 telemetry demoted | Solution/init/node-add criteria kept at `pass_threshold: 0.0` for diagnosis |
| SDK baked into the image | `@uipath/flow-sdk` + connector catalog pre-installed; runtime `npm install` and the `NODE_AUTH_TOKEN` npmrc are gone |
| Stronger GenerateSchema | Parses `essentialConfiguration.customFieldsRequestDetails` instead of grepping raw tokens |
| Two INTERIM retries | Unreadable output variables, and the Studio Web `Overwrite` 400/1001 regression |
| Corpus guards | `test_same_ground_corpus.py` with exact-equality allowlists |

## What is genuinely good

- Removing `NODE_AUTH_TOKEN` from the sandbox and baking the SDK is a real security and reliability win.
- Negative controls were not dropped, they were **reimplemented** as artifact checks (`check_inline_agent_eval.py --check no-deterministic`). Better than the regex they replaced.
- The GenerateSchema rewrite is a strict tightening.
- Exact-equality allowlists (`assert offenders == ALLOWLIST`) stop stale entries hiding finished work.
- `_incident_details()` surfacing the backend's own fault message is exactly the diagnostic that was missing.

Suites pass locally:

```
flow  1145 passed   bpmn  10 passed   case  167 passed, 14 skipped
```

(`tests/scripts` had 1 failure, missing PyYAML in my env, not a PR issue.)

---

## MAJOR 1 - `paginated_reference_lookup.yaml` no longer tests pagination

`tests/tasks/uipath-maestro-flow/connector_features/paginated_reference_lookup.yaml:44`

```diff
-    description: "invoked `uip is resources run list ... curated_channels` more than once"
-    min_count: 2
+    description: "resolved the channel through the library - paged loop or `registry prepare --resolve`"
+    min_count: 1
```

Three compounding problems:

- The task exists **because** `simple` lives on a later page. `min_count: 1` means one page-1 call, which by construction does not contain `simple`, now scores 3.0.
- Criterion 2 accepts `--resolve channel:`, and `require_success` defaults to `False` in coder-eval (`tests/test_custom_lint.py:2497`: *"a criterion counts an invocation that failed"*). Typing the flag on a command that errors passes it.
- `uip maestro flow registry prepare` **is not a CLI verb.** `assets/uip-catalog-snapshot.json` (generated `2026-09-03T05:15:27Z`, CLI `1.202.0-dev.8509`) has `registry get|list|pull|search`, no `prepare`. `scripts/check-cli-verbs.py` skipped both patterns as "too dynamic to verify", so nothing caught it. `test_same_ground_corpus.py:236` then asserts the pattern matches that phantom command.

Net: 6.0 weight is earnable with no pagination, against 3.0 for the artifact criterion that checks the resolved id `C083AN4E61E`. The free criteria outweigh the real one 2:1.

**Proposed fix:** restore `min_count: 2` on the manual route, split the SDK route into its own criterion with `require_success: true`, and confirm `registry prepare` exists. Or, in keeping with the rest of this campaign, drop both to `pass_threshold: 0.0` and let the resolved-id checker carry the weight.

## MAJOR 2 - The BPMN `StartAgentJob` rewrite is out of scope and unevidenced

`skills/uipath-maestro-bpmn/validator/bpmn-spec.json:213` and `skills/uipath-maestro-bpmn/SKILL.md:289`

This is a shipped skill behavior change riding in an eval-alignment PR. The new template is an outlier on every axis (I audited all 29 templates in the spec):

- **1 of 29** with no `<bpmn:extensionElements>` wrapper.
- **1 of 29** using the `type` attribute instead of a nested `<uipath:type value=...>`.
- Drops `releaseKey` and `folderId` entirely; flips `name` and `folderPath` to `required: true` + `binding: true`.
- Invents `{nameBindingId}` / `{folderPathBindingId}`; the other 14 binding templates all use `{bindingId}`.
- Softens SKILL.md critical rule #6 from *"Node type is a child element, never an attribute"* to *"Both declarations are supported."*

`_shared/test_bpmn_check.py:66` asserts the new shape, but a unit test asserting the spec matches itself proves nothing about canvas import.

**Ask:** a `.bpmn` that actually imported with this shape, or split this into its own PR owned by the BPMN team. Also, SKILL.md prose says `bpmn:serviceTask` (lowercase) while the template emits `bpmn:ServiceTask`.

## MAJOR 3 - The Overwrite rotation leaks tenant solutions

`tests/tasks/uipath-maestro-flow/_shared/flow_check.py:207`

`_rotate_solution_id()` rewrites the bundled `SolutionId` in the `.uipx` and retries as a new import. The comment is honest about the cost:

> *one extra tenant solution per affected case, which `cleanup_solutions.py` cannot see (it reads the `.uipx`'s CURRENT id)*

`cleanup_solutions.py` is unchanged in this PR. Confirmed at line 44: it globs `**/*.uipx` and reads `data["SolutionId"]`, so the rotated-away id is unreachable. Across a nightly over 127 tasks with multi-debug checkers, that accumulates silently. Printing to stderr is not cleanup.

**Proposed fix:** append rotated ids to a sidecar (`.rotated-solution-ids`) next to the `.uipx`, and have `cleanup_solutions.py` read it alongside the current id. Roughly 10 lines, closes the leak, and the file disappears with the block when `Overwrite` is fixed.

Secondary concern: the grader mutates the artifact under test mid-run. Later criteria in the same task see a different `SolutionId` than the agent wrote.

## MAJOR 4 - The PR body's verification section does not describe this branch

The body is the reviewer's only map through 263 files, and it has drifted:

| Claim | Actual |
|---|---|
| "111 YAML files; SHA-256 `2dc7f0a...`" | **127** tagged tasks (using the guard's own `_tagged_tasks()` logic); 99 carry the campaign header |
| "criterion counts and weight sequences match current `main` for all 111" | **False for 2 files** (below) |
| "289 passed" | **1145 passed** |
| "`@uipath/flow-sdk` 3.8.0" | `latest` is now **3.30.9**, and `FLOW_SDK_VERSION` is never passed as a build-arg by any workflow |

The two weight reductions:

```
ceql_where.yaml      2 criteria -> 1   total 10.0 -> 8.0   (dropped `flow validate`, weight 2.0)
enhanced_enum.yaml   3 criteria -> 2   total  8.0 -> 6.0   (dropped `flow validate`, weight 2.0)
```

Both are defensible (product validation is not loop-neutral for an unconfigured connector) but they are the exact opposite of what the body promises, and they are the changes a reviewer would most want called out.

On the SDK version: `tests/docker/Dockerfile:22` defaults `FLOW_SDK_VERSION=latest` and no workflow overrides it, so tonight's image gets 3.30.9 while the evidence is from 3.8.0. That is a 22-minor-version drift on a preview package. It matches the existing `CLI_VERSION=dev` convention, so I would not block on it, but wiring `FLOW_SDK_VERSION` through as a build-arg from a repo variable gives you a pin without editing the Dockerfile when a bad SDK release lands.

---

## MINOR 5 - Two prompts now name an SDK-only verb

`ceql_where.yaml:35`, `enhanced_enum.yaml:22`: *"Validate the final flow file"* became *"Compile the final flow file."*

`maestro flow compile` is a real verb, but it appears **nowhere** in `skills/uipath-maestro-flow/`, and it is the SDK direction (`.flow.ts` -> `.flow`). For a v1 agent authoring JSON directly, "compile the flow file" is a no-op with no skill guidance. That biases the prompt toward the SDK arm, which is the asymmetry this campaign exists to remove, just pointed the other way.

**Fix:** keep the prompt on `validate`, or document `compile` in the flow skill.

## MINOR 6 - `init_validate.yaml` loses solution-first coverage outright

```diff
-  Tests whether the skill teaches the correct solution-first workflow and CLI usage.
+  Tests whether the skill produces a valid Flow artifact.
```

The prompt's entire 3-step solution-first block is deleted and the `.uipx` registration `json_check` drops to `pass_threshold: 0.0`. Separately, 34 prompts gain *"Keep the Flow inside a solution of the same name"*, which tells the agent the layout the eval used to test it for.

Both are deliberate trades for comparability, but if the skill regresses on teaching solution-first, nothing in the corpus catches it. **Question:** is that coverage preserved anywhere, for example a v1-only tagged task?

## MINOR 7 - 99 task files point at a document that is not in this repo

The campaign header added to 99 YAMLs references `workspaces/docs/flow-sdk/2026-08-21-maestro-flow-alignment-expansion-plan.md`. That path does not exist here (it lives in `tmatup/workspaces`). 99 dangling references no reviewer or future maintainer can follow. Inline the two or three load-bearing decisions, or drop the path.

## MINOR 8 - Unanchored `rglob` in the BPMN criteria

`wiki_pageviews.yaml:64` is now `Path('.').rglob('operate.json')`, and `debug_workflow_mocked.yaml:64` is `rglob('summary.md')`. Both sandboxes copy all of `skills/uipath-maestro-bpmn` in via `template_dir`.

I checked: no `operate.json`, `summary.md`, or `.bpmn` in that tree today, so **this is safe as merged**. But the day a reference doc ships an example `operate.json`, the criterion passes for free.

`simple_approval_bpmn.yaml:52` already shows the right pattern (anchor to the discovered `.bpmn`'s parent). Worth applying to `operate.json` for consistency.

## MINOR 9 - `glob.glob('**', recursive=True)` follows the new `node_modules` symlink

`stage-preview-sdk-workspace.sh:79` symlinks `/opt/preview-flow-sdk/node_modules` into the workspace. Verified behavior difference on Python 3.14:

```
glob.glob('**/*.flow', recursive=True)  ->  ['real/deep/x.flow', 'link/deep/x.flow']   # follows
Path('.').rglob('*.flow')               ->  ['real/deep/x.flow']                       # does not
```

So `assert_no_flow_files()` (`flow_check.py:1826`) and `_find_project()` walk the SDK tree. I installed `@uipath/flow-sdk@3.30.9` plus its 16 deps and found zero `.flow`, `project.uiproj`, `.uipx`, or `operate.json`, so **no current false negative**. Still, one SDK fixture away from breaking every read-only preview task, and it is a latency cost on every scan. `Path.rglob` or an explicit `node_modules` prune closes it.

## NITS

- `tests/scripts/stage_shared.sh:36` guard `tests/tasks | tests/tasks/*` also matches `tests/tasks/../../etc`. A stated guard that does not guard.
- `validate_flow.py:167` `if not flows:` is dead. `find_flow_files()` calls `_fail()` rather than returning empty.
- `tests/docker/Dockerfile:184` fallback branch sets `FLOW_SDK_CONNECTORS_DIR` to a freshly `mkdir -p`'d empty directory on the new CLI layout. Anything reading it silently gets nothing.
- `case_check.py:64` line is well over 100 chars.
- `CODEOWNERS:22` moves `/preview/uipath-maestro-case/` from `@dmetzgar @tmatup` to `@UiPath/maestro-cli-team`. Unrelated governance change in the last commit. Removing a named owner should carry that person's sign-off, not ride a 263-file eval PR.

---

## Two style pushbacks

**The PR body.** Long, dense, and confident, with vocabulary that reads as authority rather than fact: "ratified same-ground recipe", "born-neutral set", "provenance-valid rows", "every red is classified". Four of its verification claims are wrong or stale (MAJOR 4). On a 263-file PR the body *is* the review surface, so unverifiable confidence is worse than a short honest summary.

**Comment-to-code ratio in `flow_check.py`.** The `_OVERWRITE_STUCK` block is **24 lines of prose above 3 constants**, including archaeology ("last known good: 2026-09-01 07:30Z", "the only earlier archived Overwrite 400 was code 20001"), and `_variables_unreadable` has 11 lines above one constant. The *what and why* here is genuinely non-obvious and deserves a comment. The changelog does not. Cut each to the 3 or 4 lines a future maintainer needs to decide whether to delete the block, and put the archaeology in the linked issue.

---

The direction is right and much of the execution is careful. But merging as-is ships:

1. A grading regression that removes the point of a task (MAJOR 1).
2. An unrelated, unevidenced skill change that could break BPMN canvas import (MAJOR 2).
3. A known tenant resource leak with no cleanup path (MAJOR 3).
4. A body whose verification section does not match the branch (MAJOR 4).

MAJOR 1 and 3 are small, mechanical fixes. MAJOR 2 wants either import evidence or its own PR. MAJOR 4 is a rewrite of the body, which for a PR this size is not optional. Once those land I would approve without needing another full pass.

tmatup:
### Review Response

Code changes landed on the campaign branch via #3045 (merge commit `19322e32c`). The PR body is rewritten (MAJOR 4). Per item:

**MAJOR 1. `paginated_reference_lookup.yaml` no longer tests pagination**

**Fixed** in #3045. Two of the three points were right: `uip maestro flow registry prepare` is not a CLI verb (the SDK form is `npx flow-sdk registry prepare`; the corpus test's sample command was wrong, the yaml pattern matched the real form by accident), and `min_count: 1` on the manual list call was a regression. One correction to the framing: the task decision is a strict AND over gating criteria (coder-eval `orchestrator.py`, "all criteria passed their thresholds"), and the artifact criterion requires the page-2 id `C083AN4E61E`, so a page-1-only run could inflate the score but never pass. Now:
- criterion 1 is the v1 paged loop at `min_count: 2`, advisory (`pass_threshold: 0.0`) because the SDK arm never runs it; its 3.0 weight is only earned when the loop actually ran twice;
- criterion 2 gates on "went past page 1" for either route: a channel-resource `list` call carrying `nextPage=` in the same command, or `registry prepare … uipath-salesforce-slack … --resolve channel:`, with `require_success: true`;
- the corpus test asserts the negatives: page-1-only, `--resolve channel:` typed on `flow validate`, bare `echo nextPage=`.

Known trade on `require_success`: coder-eval derives success from the tool result's error flag, so a `uip … nextPage=… | jq` pipeline where jq fails after a good page-2 fetch is not counted. Criterion count and weights are unchanged. Not verified live yet; the next nightly on the branch is the proof.

**MAJOR 2. BPMN `StartAgentJob` rewrite is out of scope and unevidenced**

**Keeping the template; fixed the prose casing** (`bpmn:ServiceTask`). Evidence the review asked for, from the CLI repo rather than a canvas import:
- the skill's `bpmn-spec.json` entry for `Orchestrator.StartAgentJob` is byte-equal (canonical JSON) to the CLI's own registry manifest, `packages/maestro-sdk/src/manifest/bpmn-spec.json`;
- the CLI validator's tests (`packages/packager/packager-tool-bpmn/tests/bpmn-tool.spec.ts`) include "should validate a runtime-authored StartAgentJob task with agent bindings" for this exact shape, and "should fail StartAgentJob when old releaseKey context is used", which rejects the previous template with "must use binding-backed name and folderPath".
So the outlier shape is the CLI's contract, and the template this PR replaced was the one the CLI refuses. What does not exist is a canvas-import screenshot; the CLI validator is the import gate we have. It landed as #2901 into the campaign branch because the contract-variant task failed on a correct artifact. `skills/uipath-maestro-bpmn/` is owned by `@UiPath/maestro-cli-team`, the same team that owns the campaign. Whether it should still ship as its own PR to `main` is @tmatup's call.

**MAJOR 3. Overwrite rotation leaks tenant solutions**

**Fixed** in #3045, the way proposed: `_rotate_solution_id` appends the rotated-away id to `.rotated-solution-ids` beside the `.uipx`; the flow `cleanup_solutions.py` reads the sidecar and deletes those ids with the current one. Tests cover the write and the read (dedupe, blank lines). Caveat from UiPath/cli#3938: `uip solution delete` returns the same 400/1001 today, so the deletes will fail until Studio Web recovers; the ids are recorded either way and cleanup succeeds once delete works. On the secondary concern: no flow task criterion reads `SolutionId`, so the mid-run mutation changes nothing later criteria see.

**MAJOR 4. PR body does not describe the branch**

**Fixed**: body rewritten with the counts recomputed at the tip. 127 tagged tasks (99 with the campaign header); criterion count and weights match `main` for 125 of 127, the exceptions being exactly the two you found (#2915, both drop the 2.0 `flow validate` criterion), now called out with the reason; flow suite 1147; SDK `latest` = 3.30.9 and `run-coder-eval.yml` now passes `vars.FLOW_SDK_VERSION` as the build-arg (default `latest`). One more drift you did not list, now stated in the body: of the nine #2557 tasks, `paginated_reference_lookup.yaml` was changed by #2945 and again here; the other eight are byte-identical to the base.

**MINOR 5. Prompts name an SDK-only verb**

**Fixed**: both prompts say "Produce the final flow file." Neither `validate` nor `compile` is graded in those two tasks (the validate criterion was removed in #2915), so the prompt names neither.

**MINOR 6. `init_validate.yaml` loses solution-first coverage**

**Answer: no, that coverage is not preserved as a skill-teaching test.** The advisory criteria (`uip solution init`, `.uipx` registration) still report whether the v1 agent did it, so a regression is visible in the nightly report, but nothing gates on it, and 34 prompts now state the layout. A v1-only tagged task that keeps the old prompt is the way to restore it; that is a corpus-shape decision for @tmatup, not something this PR adds.

**MINOR 7. 99 files reference a document not in this repo**

**Fixed**: the `# See workspaces/docs/…` line is removed from all 99 headers. The three load-bearing decisions (loop-neutral v1-base prompt, outcome-graded criteria at preserved weights, arm-agnostic discovery) were already the header's other lines.

**MINOR 8. Unanchored `rglob` in BPMN criteria**

**Fixed** for `operate.json`: anchored to the discovered `WikiPageviews.bpmn`'s parent, same pattern as `simple_approval_bpmn.yaml:52`. `summary.md` left as is: the prompt asks for the file but does not say where, so anchoring would grade a location the agent was never given.

**MINOR 9. `glob.glob('**')` follows the `node_modules` symlink**

**Fixed**: the three cwd-rooted scans (`_find_project`, `find_flow_files`' project candidates, `assert_no_flow_files`) use `_rglob_pruned`, an `os.walk` that never descends into `node_modules`. Explicit prune rather than `Path.rglob` because the symlink behaviour of `rglob` differs across Python versions. Test: a symlinked SDK tree holding a `.flow` and a `project.uiproj` is invisible; a real `.flow` still trips the negative check.

**Nits**

- `stage_shared.sh`: both sides go through `realpath -m`; `tests/tasks/../../etc` is refused (checked by hand).
- `validate_flow.py`: dead branch removed; its test now asserts the real `SystemExit` path.
- Dockerfile:184: nothing reads `FLOW_SDK_CONNECTORS_DIR` (not the SDK, the stage script, or the workflows), so the variable, the symlink and the empty-dir fallback are removed rather than made loud.
- `case_check.py`: the line is wrapped. There is no ruff/flake8 line-length config in the repo, so this is convention only.
- CODEOWNERS: `a7868335f` is reverted on the branch and reopened as #3046 against `main` for @dmetzgar's sign-off.

**Style**

- Body: rewritten shorter, without the campaign vocabulary, with the numbers a reviewer can recompute.
- `flow_check.py`: the `_OVERWRITE_STUCK` block is 4 comment lines + 3 for the second envelope; `_VARIABLES_UNREADABLE` is 4. The evidence tables live in UiPath/cli#3938 and UiPath/cli#3929.

https://claude.ai/code/session_01UauCCMBpJ7ws1meQwBfYSS

tmatup:
Correction to MAJOR 1 above: the resolver verb does exist in the CLI, as `uip maestro registry prepare` (top-level `maestro registry`, UiPath/cli#3969, in the catalog snapshot). It is not under `maestro flow registry`, which is where the corpus test's sample command and the review looked. The yaml pattern matches both the `npx flow-sdk registry prepare …` and `uip maestro registry prepare …` forms; #3047 adds the uip form to the corpus test.

https://claude.ai/code/session_01UauCCMBpJ7ws1meQwBfYSS

.github/workflows/smoke-skills.yml
```diff
@@ -368,12 +368,13 @@ jobs:
...
-      # `-dev.<run>` prerelease under the `dev` dist-tag (cli PR #2650 renamed the
-      # old `alpha` tag → `dev` on 2026-06-26), so smoke tracks main — not stale
-      # `latest` from public npm (which lags by ~2 weeks).
+      # Build custom smoke image extending coder-eval-agent with @uipath/cli,
+      # every whitelisted tool plugin from GitHub Packages @dev, and the latest
+      # published @uipath/flow-sdk (see tests/docker/Dockerfile). Every push to
+      # `cli/main` publishes a `-dev.<run>` prerelease under the `dev` dist-tag
+      # (cli PR #2650 renamed the old `alpha` tag → `dev` on 2026-06-26), so
+      # smoke tracks main — not stale `latest` from public npm (which lags by ~2
+      # weeks).
```
rockymadden, line 377: **Minor: the `FLOW_SDK_VERSION` pin does not reach this image.**

`run-coder-eval.yml:336` now passes `--build-arg FLOW_SDK_VERSION="${{ vars.FLOW_SDK_VERSION || 'latest' }}"`, but the `docker build` just below (line 383) does not, so it always takes the Dockerfile default `latest`. Both steps tag `skills-image:latest`.

Effect: setting the repo variable to stop a bad SDK release pins nightly but not smoke, and the two images can silently carry different `@uipath/flow-sdk` versions. That is the reverse of what smoke is for.

One line, same as the other workflow:

```yaml
--build-arg FLOW_SDK_VERSION="${{ vars.FLOW_SDK_VERSION || 'latest' }}" \
```

skills/uipath-maestro-bpmn/validator/bpmn-spec.json
```diff
@@ -218,12 +205,12 @@
...
+      "xmlTemplate": "<bpmn:ServiceTask id=\"{id}\" name=\"{name}\">\n  <uipath:activity type=\"Orchestrator.StartAgentJob\" version=\"v1\">\n    <uipath:context>\n      <uipath:input name=\"name\" value=\"=bindings.{nameBindingId}\" />\n      <uipath:input name=\"folderPath\" value=\"=bindings.{folderPathBindingId}\" />\n    </uipath:context>\n    <uipath:input name=\"JobArguments\" type=\"json\" target=\"bodyField\"><![CDATA[{}]]></uipath:input>\n    <uipath:output name=\"Process response\" type=\"Orchestrator.RunJob\" var=\"{varId}\" />\n  </uipath:activity>\n  <bpmn:incoming>{incomingEdge}</bpmn:incoming>\n  <bpmn:outgoing>{outgoingEdge}</bpmn:outgoing>\n</bpmn:ServiceTask>",
```
rockymadden, line 213: This entry is correct: byte-equal to `cli@origin/main`'s `packages/maestro-sdk/src/manifest/bpmn-spec.json`. My MAJOR 2 was wrong and I withdraw it.

**Follow-up (not blocking this PR): the same drift is still live in three other templates.**

Of the 29 `extensionTypes` shared with the CLI manifest, 4 still differ. `BPMN.Variables` is notes-only, but the three `Intsvc` ones are substantive:

| Type | Field | CLI | Skill |
|---|---|---|---|
| `Intsvc.ActivityExecution` | `xmlTemplate` | `<bpmn:sendTask>` | `<bpmn:SendTask>` |
| | `outputName` | `response` | `result` |
| | `outputType` | `jsonSchema` | `custom` |
| | `outputSource` | `=response` | `.` |
| `Intsvc.EventTrigger` | `inputFields`, `xmlTemplate` | differ | |
| `Intsvc.WaitForEvent` | `inputFields`, `xmlTemplate` | differ | |

The casing is not cosmetic. A real Studio Web export in `PO.BpmnEngine` (`tests/BpmnEngine/UiPath.PO.Worker.Tests/Workflows/Bpmn/TestData/MultiInstanceEmailWorkflow.bpmn`) emits `<bpmn:sendTask>`, matching the CLI, not this file. And an agent following the skill's `outputName`/`outputSource` names and sources the activity output wrong.

This is exactly the failure #2901 fixed for `StartAgentJob`, still present for `Intsvc.*`. Worth an issue, and worth a CI guard that diffs this file against the CLI manifest so the drift cannot recur silently.

tests/tasks/uipath-maestro-flow/connector_features/paginated_reference_lookup.yaml
```diff
@@ -38,24 +38,28 @@ initial_prompt: |
...
+    pass_threshold: 0.0
 
   - type: command_executed
-    description: "Agent used `nextPage=<token>` to advance pagination at least once"
+    description: "Agent went past page 1 — a channel-resource list call carrying `nextPage=<token>`, or `registry prepare … --resolve channel:` which pages inside `prepare`"
     tool_name: "Bash"
-    command_pattern: 'nextPage='
+    command_pattern: '(uip\s+is\s+resources\s+run\s+list\s+\\?"?uipath-salesforce-slack\\?"?\s+\\?"?(curated_channels|conversations)[^\n]*nextPage=|registry\s+prepare\s+\\?"?uipath-salesforce-slack\\?"?[^\n]*--resolve\s+\\?"?channel:)'
     min_count: 1
+    require_success: true
```
rockymadden, line 72: Good fix, and the disclosed pipeline caveat is real. One case it does not cover:

`command_executed.py:173` is `cmd.result_status != "success"`, and `codex_agent.py:620` (`close_open_tools`) stamps `unknown` on any tool that started but never completed. So a command cut off by a turn timeout also fails to count here, not just a failing pipeline.

That matters for this arm specifically: the evidence run had 20 agent-turn timeouts on live v1. This criterion is gating at 3.0, so a truncated run that *did* page correctly scores 0 on it.

Not worth changing pre-merge, since the artifact criterion (`C083AN4E61E`) already gates the real outcome. But if the next nightly shows this criterion flaking on otherwise-passing rows, dropping it to `pass_threshold: 0.0` costs no coverage.

rockymadden (approved):
## Re-review at `957f5b72f`

Verified the fixes against the CLI repo and coder-eval source rather than the response alone.

```
tests/tasks/uipath-maestro-flow   1147 passed
tests/tasks/uipath-maestro-bpmn     10 passed
tests/tasks/uipath-maestro-case    167 passed, 14 skipped
```

Matches the rewritten body exactly.

| Item | Status |
|---|---|
| MAJOR 1 pagination gate | Fixed. `min_count: 2` restored, gating criterion anchored + `require_success: true`, corpus test now asserts the negatives |
| MAJOR 2 BPMN `StartAgentJob` | **My error, see inline** |
| MAJOR 3 solution leak | Fixed. Sidecar round-trip tested end to end |
| MAJOR 4 stale body | Fixed. 127 / 125-of-127 / 1147 / 3.30.9 all recompute correctly |
| MINOR 5 SDK-only verb | Fixed ("Produce") |
| MINOR 6 solution-first coverage | Answered, declined as a corpus-shape call. Fair |
| MINOR 7 dangling doc ref | Fixed across all 99 headers |
| MINOR 8 unanchored `rglob` | Fixed for `operate.json`; `summary.md` left open with a sound reason |
| MINOR 9 `node_modules` symlink | Fixed. Verified `_rglob_pruned` by hand |
| Nits | All 5; CODEOWNERS spun out as #3046, already merged |

`_rglob_pruned` against a symlinked SDK tree:

```
plain glob  : ['node_modules/deep/x.flow', 'proj/a.flow', 'real/deep/x.flow']
pruned      : ['proj/a.flow', 'real/deep/x.flow']
```

### Two corrections to my last review

**MAJOR 2 was wrong.** The skill's `Orchestrator.StartAgentJob` entry is byte-equal (canonical JSON) to `cli@origin/main`'s `packages/maestro-sdk/src/manifest/bpmn-spec.json`, and `packager-tool-bpmn/tests/bpmn-tool.spec.ts:813` asserts the *old* `releaseKey` shape must fail. The template this PR replaced is the one the CLI refuses. That edit was a correction, not a regression.

**MAJOR 1 was overstated on impact.** `orchestrator.py:1956` gates strict-AND over every gating criterion, and the artifact criterion requires the page-2 id. A page-1-only run inflated the weighted score but could never pass the task. The fix is still worth having; "false pass" was the wrong word.

### Approving

Every blocker is resolved: three fixed as proposed, one refuted with evidence I could independently verify. The response argues from source rather than assertion, and the corpus test gained negative assertions I did not ask for.

Three non-blocking items inline. The `bpmn-spec.json` one is the only one worth a follow-up issue: it generalizes #2901 and I found real-export evidence for it.

rockymadden (approved):

.github/workflows/smoke-skills.yml
tmatup: Fixed in 843cb7428. The smoke-image build now receives the same `FLOW_SDK_VERSION` repository variable as the nightly image, so both builds honor a rollback pin.

skills/uipath-maestro-bpmn/validator/bpmn-spec.json
tmatup: Fixed the current drift in 843cb7428. All shared `extensionTypes` now match the current `UiPath/cli` manifest after canonical JSON normalization, including the three `Intsvc.*` contracts and `BPMN.Variables`; focused tests pin the runtime-significant fields. Filed #3052 for the trusted scheduled cross-repository drift guard so no private-repo credential is exposed to pull-request code.

tests/tasks/uipath-maestro-flow/connector_features/paginated_reference_lookup.yaml
tmatup: Fixed in 843cb7428. The page-2 command criterion is now advisory because a turn timeout records an otherwise-correct command as `unknown`; `require_success: true` remains for useful telemetry, while the resolved `C083AN4E61E` artifact criterion remains the gating authority.

tmatup:
## `Smoke RPA Skill Tests (Windows)` red on this PR: root cause and fix

**Short version:** the failure is in the `uipath-rpa` skill on `main`, not in this PR. Fix PR: #3055.

**What failed.** `skill-rpa-execution-map-greenfield` scored 0.62 against the 0.7 gate (run 33831140156). This PR does not touch `skills/uipath-rpa`; the Windows smoke ran because `tests/experiments/*.yaml` changed, which the detect job treats as an infra change.

**Why.** The task caps the agent at 10 turns. Every CI run of it since it became a smoke gate (08-27) ended in `MAX_TURNS_EXHAUSTED`: 0.73, 0.77, 0.77, 0.77, and now 0.62. The gate scores weighted_score, so four of those counted as passes. In this run the agent spent all 10 turns on reads and `init` and wrote no XAML. #2698 (08-26) made Rule 22 a two-step read (Grep a marker, then Read with a limit) for `xaml-basics-and-rules.md` and the same for `common-pitfalls.md`; each pair is two sequential turns, while the journey map still budgeted one parallel Read.

**Fix (#3055).** Split the grep-gated catalog out of `xaml-basics-and-rules.md` so Rule 22 is a plain one-turn Read again; move the pitfalls heading Grep into T1 and its section Reads into one T2 turn. Journey maps re-derived (greenfield stays ≤5 turns). No change to `max_turns` or the gate threshold.

**Verified.** On #3055 the Windows smoke reports `skill-rpa-execution-map-greenfield status=SUCCESS score=1.00` (run 33842310232), the first time this task has finished in CI. All four Windows rows SUCCESS at 1.00.

**Plan.** Once #3055 merges, merge `main` into `campaign/same-ground-expansion` and let the required check re-run. Not re-running the failed job now: historical odds of clearing 0.7 are 4/5 but the task has never actually finished.

PR #2761, opened 2026-08-24
test(uipath-maestro-case): contract-execution golden-rebuild task (SLA + interruption surface, field-name regression)

song-zhao-25:
## What

Adds `tests/tasks/uipath-maestro-case/golden_rebuild/contract_execution/` — a live-tenant golden-rebuild task that rebuilds an 8-stage / 30-task contract review-and-execution case from an approved SDD whose resource references are real.

Complements the existing `golden_rebuild/cm_golden_expense` (which covers the feature matrix — all 9 task types). This one puts its weight on the **SLA and interruption surface**, which nothing covered before:

- 10-business-day case SLA + 7 stage SLAs, each with at-risk(70%)/breach escalations
- 7 `sla-status-change` task triggers, each scoped to its own stage's SLA
- an interrupting `return-to-origin` intervention lane
- 3 interrupting secondary stages + 2 terminal non-completing case exits
- a corrections loop that re-enters Stage 1 from Counsel review
- two **distinct** HTTP Webhook connections — one as a task, one as a stage-entry `wait-for-connector` rule

Stops at `validate` — not `debug` (the case suspends on HITL actions, SLA clocks, and two webhook events).

## Why the field-name grader exists

`check_contract_exec_fieldnames.py` pins a build-then-fail-at-runtime class of defect. Task-output sub-property names (`request_body`, `unusualClauses`, `authorityLevel`, `deviationFlags`) are case-sensitive at runtime but **invisible to `uip maestro case validate`**. A build that PascalCases them validates clean, publishes clean, then dies in Studio Web with *"RequestBody not found, did you mean request_body"* — taking every gate that reads it. The same expressions must also null-guard the object (`(vars.X || {}).Y`); an unguarded deref throws before the producing task has run.

The grader parses the declared names out of `fixtures/sdd.md` and asserts the caseplan uses that exact casing in guarded form, plus that the webhook event schema keys survive a PascalCasing `--output json` serializer.

## Design

Four deterministic graders, no `llm_judge`. **Every expectation naming SDD content is parsed from `fixtures/sdd.md` at grade time**, so re-sweeping the fixture after a tenant reinstall updates agent input and graders in one file.

| Grader | Weight | Covers |
|---|---|---|
| `check_contract_exec_case.py` | 4.0 | stages, secondary lanes, transitions, case exits, task/required/run-once sets, return-to-origin, case + 7 stage SLAs, `sla-status-change` scoping, runtime flags |
| `check_contract_exec_bindings.py` | 4.0 | live resource keys in caseplan + `bindings_v2`, both connections, activity type, the two distinct webhooks, no skeletons |
| `check_contract_exec_semantics.py` | 5.0 | dataflow, output-id uniqueness, extract/custom outputs, gate polarity, literal fidelity |
| `check_contract_exec_fieldnames.py` | 3.0 | property casing + null-guard form + event schema keys |

`test_contract_execution_checkers.py` — 18 offline tests covering the topology and field-name graders, including negative cases for every regression they exist to catch.

`run_limits` note: `turn_timeout` is set equal to `task_timeout` (18000). One harness "turn" is the entire agent query stream, so `turn_timeout` is the binding cap; a smaller value silently makes `task_timeout` dead config and kills a healthy long build with status ERROR / score 0.0 regardless of work completed.

## Tenant precondition

`ContractOperationsApi`, `ContractAnalysisAgent`, `ContractObligationTracking`, `ContractCaseActionApp` deployed at the folders pinned in `fixtures/sdd.md`, plus both HTTP Webhook connections. All verified present on the coder-eval tenant at authoring time. A reinstall re-mints every ID — re-sweep `fixtures/sdd.md`, the single source both the agent and the graders read.


roberts-cliff (approved):

roberts-cliff:
Kicked off coder-eval on this branch for both harnesses, on the new task:

| harness | run |
|---|---|
| claude | https://github.com/UiPath/skills/actions/runs/33421128302 |
| codex  | https://github.com/UiPath/skills/actions/runs/33421116587 |

Both against `test/case-golden-rebuild-contract-execution`, task glob `tasks/uipath-maestro-case/golden_rebuild/contract_execution/contract_execution.yaml`, `-j 1`.

I'll post the results here. Two things I'll separate out when reading them, since a red run on a brand-new task is ambiguous:

- **Known open defects in this family.** The sibling `golden_rebuild/cm_golden_expense` is currently failing on unresolved `$xref(` survivors (Step 11.5 / Check 4 going unexecuted) and on `bindings_v2` `resourceSubType` being null. If this task trips on either, that is a pre-existing skill defect rather than anything wrong with the task.
- **Platform.** Alpha is degraded right now — Orchestrator SQL is at roughly 38% failure today and `case debug` runs are returning `Faulted`. Anything that dies at the debug step is likely environmental.

I'll read the produced artifacts rather than just the checker verdict — these checkers fail fast, so an earlier assertion can mask whether the later ones would pass.

StephenHanzlik:
Claude coder eval failed.  Looking into it.

StephenHanzlik:
Re-ran the first one - https://github.com/UiPath/skills/actions/runs/33421128302

StephenHanzlik:
FYI the above table has the models inverted.  The Codex one failed twice now.  Claude is still working through the eval.

StephenHanzlik (approved):

StephenHanzlik:
Coder Eval Runs:

Codex: https://github.com/UiPath/skills/actions/runs/35770649458
Claude: https://github.com/UiPath/skills/actions/runs/35765710271

PR #2763, opened 2026-08-24
fix(hitl-skill): correct Maestro BPMN QuickForm node shape

dushyant-uipath:
## Summary

The Surface: Maestro section of uipath-human-in-the-loop's SKILL.md documented an incomplete and incorrect node shape for QuickForm HITL tasks on Maestro BPMN. Confirmed by directly reproducing a working QuickForm task against a real Studio Web BPMN process and inspecting a real exported .bpmn file.

- The registered `uipath.human-in-the-loop.quick-form`/`.coded-action-app` strings are registry shortcut (palette) identifiers. They are never written literally into the XML.
- The real element is `bpmn:task` with `uipath:type value="Actions.HITL" version="v2"`. `bpmn-spec.json`'s `bpmn:UserTask`/`v1` template belongs to the app-based/coded-action-app path.
- Studio Web's "Edit Schema" canvas fails to open without `<uipath:inputSchema>` and `<uipath:input name="HitlTaskArguments">`. Confirmed by reproduction.
- The `Action` output needs a matching process-level `uipath:inputOutput` variable declaration.
- `_schemaFileId` is a server-assigned file ID. It cannot be invented as a placeholder UUID: doing so causes a 404 on Studio Web's internal file-rename call, and no `uip` CLI command resolves it today. The corrected doc spells out the only known working procedure: upload once, look up the real ID via `FileOperations/Structure`, patch it in, then push back via a targeted single-file `PUT`. A second whole-project upload mints new file IDs for everything and undoes the patch.
- Missing `bpmndi:BPMNEdge` diagram-interchange entries leave sequence flows logically connected but visually disconnected in the canvas.

This mirrors and cross-links the equivalent Case Management fix in #2056 (hitl-casetask-action.md), which documents the identical .hitl.json sidecar shape and the same _schemaFileId reconciliation problem for the Case surface.

## Test plan

- [x] Reproduced a QuickForm HITL task from scratch (`uip maestro bpmn init`, hand-authored .bpmn + .hitl.json per this doc's corrected shape)
- [x] `uip maestro bpmn validate` passes clean
- [x] Uploaded to Studio Web, resolved the real _schemaFileId via FileOperations/Structure, patched, pushed via targeted PUT
- [x] Confirmed live in Studio Web: task renders, Edit Schema opens the canvas, connecting edges render between nodes

### DIFF
--- a/skills/uipath-human-in-the-loop/SKILL.md
+++ b/skills/uipath-human-in-the-loop/SKILL.md
@@ -298,7 +298,44 @@ response = interrupt(CreateTask(
 
 ### Surface: Maestro
 
-QuickForm and coded-action-app HITL nodes are both supported on Maestro BPMN processes — `uipath.human-in-the-loop.quick-form` and `uipath.human-in-the-loop.coded-action-app` are registered element types in the BPMN validator ([bpmn-spec.json](../uipath-maestro-bpmn/validator/bpmn-spec.json)), same node-type strings as the Flow surface. Write the node directly into the `.bpmn` XML as a `bpmn:UserTask` with a `uipath:activity` extension element (see the `Actions.HITL` extension type in the validator spec for the app-based/coded-action-app XML shape and context fields — `appId`, `appVersion`, `actions`, `key`, `taskTitle`).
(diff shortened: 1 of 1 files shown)


dushyant-uipath:
Folded into #2056 instead of a separate PR — same commit now lives there.

PR #2764, opened 2026-08-24
fix(data-fabric): clarify schema `name` vs query `fieldName`; accept `--file` in entity-scope create check

niket0503:
Follow-up to the field-`name`-shape work (merged), from two coder-eval failures.

## 1. Skill: `name` vs `fieldName` (root cause of an expense-tracker failure)
Data Fabric uses **two keys for "a field's name" depending on context**:
- **Entity schema** (`create` / `addFields`) → `name`
- **Query filters / `sortOptions`** (`records query`) → `fieldName`

When a task mixes both (create entities *and* query them), agents conflate the keys and send `fieldName` in a create body. The field then arrives with no name and CEP rejects the **whole** atomic create with `"Field name is not provided in request."`, collapsing the task. Observed live on gpt-5.6-terra (CLI `1.201.0-dev.8272`, which includes the `fieldName`→`name` rename).

`entity-schema.md` now calls this out explicitly at the field-definition rules.

## 2. Eval task: entity-scope create check should accept `--file`
`integration_entity_scope`'s "created CE_FolderScopeTest with a valid schema" `command_executed` check required the field schema **inline** (`(?=.*"name")`). An agent (Sonnet) that writes the schema to a file and runs `entities create … --file schema.json --folder-key …` — a valid approach — false-failed, since `"name"` is in the file, not the command. Loosened the pattern to accept `--file` as valid schema evidence.

No checks deleted; correctness stays gated by outcome checks.

### DIFF
--- a/tests/tasks/uipath-platform/data-fabric/integration_entity_scope.yaml
+++ b/tests/tasks/uipath-platform/data-fabric/integration_entity_scope.yaml
@@ -71,7 +71,7 @@ success_criteria:
   - type: command_executed
     description: "Agent created CE_FolderScopeTest in a folder with a valid schema body"
     tool_name: "Bash"
-    command_pattern: '(?s)uip\s+df\s+entities\s+create\s+"?CE_FolderScopeTest"?(?=.*--folder-key)(?=.*\\?\"name\\?\")'
+    command_pattern: '(?s)uip\s+df\s+entities\s+create\s+"?CE_FolderScopeTest"?(?=.*--folder-key)(?=.*(\\?\"name\\?\"|--file))'
     min_count: 1
     weight: 3.0
     pass_threshold: 1.0


tests/tasks/uipath-platform/data-fabric/integration_entity_scope.yaml
```diff
@@ -71,7 +71,7 @@ success_criteria:
   - type: command_executed
     description: "Agent created CE_FolderScopeTest in a folder with a valid schema body"
     tool_name: "Bash"
-    command_pattern: '(?s)uip\s+df\s+entities\s+create\s+"?CE_FolderScopeTest"?(?=.*--folder-key)(?=.*\\?\"name\\?\")'
+    command_pattern: '(?s)uip\s+df\s+entities\s+create\s+"?CE_FolderScopeTest"?(?=.*--folder-key)(?=.*(\\?\"name\\?\"|--file))'
```
AditiGoyalUipath, line 74: Why this is needed? This is tigheting the validation no

skills/uipath-platform/references/data-fabric/entity-schema.md
```diff
@@ -20,6 +20,7 @@ uip df entities create "MyEntity" \
 ```
 
 - `fields` array is **required**. Each entry must include `name`.
+- **The schema field key is `name`, NOT `fieldName`.** `fieldName` is the *query* key — it belongs only in `records query` filter bodies and `sortOptions` (see [records-query.md](./records-query.md)), never in `create`/`addFields`. Using `fieldName` here sends the field with no name and the server rejects the whole create: `"Field name is not provided in request."`
```
AditiGoyalUipath: Do we need to mention this specifically that it is not fieldName

tests/tasks/uipath-platform/data-fabric/integration_entity_scope.yaml
niket0503: Its not tightening, its allowing a file to be passed as argument to the commend

skills/uipath-platform/references/data-fabric/entity-schema.md
niket0503: There was one instance where agent still tried to use fieldName.
I am removing this for now, will monitor if it still uses fieldName

AditiGoyalUipath (approved):

PR #2766, opened 2026-08-24
fix(process-mining): published-stage querying works after a published-stage ingestion

Cellcote:
The skill tells agents that `uip pm query --stage published` is a dead end. It
isn't. Reported in review of #2547, which had encoded the same false claim as a
graded criterion (removed there).

## The claim

`references/lifecycle-and-rbac.md` stated it most strongly:

> **CLI caveat:** `uip pm query --stage published` is currently **not reachable** —
> `/query/{id}/published` needs a completed ingestion on that stage and no `uip pm`
> path produces one (`apps publish` answers `IngestionNeeded: true`, but a following
> `ingestions create --wait` still leaves published querying at
> `UserError_InvalidOrNoIngestion`; verified against a live tenant). So publishing
> promotes the app to the **dashboards**, but **CLI-driven `query` stays on `dev`**.

## Why it's wrong

`ingestions create` takes `--stage`:

```
$ uip pm ingestions create --help
  --stage <stage>    App stage to ingest into
```

So `uip pm ingestions create <app> --stage published --wait` **is** the "path that
produces one", and once it completes, `query --stage published` works. The
half-truth in the old text is the middle clause: a plain `ingestions create --wait`
without `--stage` ingests into `dev`, which of course leaves published querying
at `UserError_InvalidOrNoIngestion`. That was read as "unreachable" rather than
"you ingested into the wrong stage".

The skill was already self-contradictory about it — six lines below the caveat,
the promote loop ended with "→ analyse on `--stage published`", and
`references/uip-pm-cli.md` said "Develop on `dev` (a subset), `publish`, then read
`--stage published`". Only the caveat and the two places quoting it were wrong.

## Changes

| File | Change |
|---|---|
| `SKILL.md` (Rule 8) | published querying works after a published-stage ingestion, and is optional rather than a dead end |
| `references/querying.md` | `--stage published` needs a completed ingestion on that stage; it 400s `UserError_InvalidOrNoIngestion` only until then |
| `references/lifecycle-and-rbac.md` | caveat rewritten as guidance; `--stage published` named on the ingestion in the promote loop |

Net −12 lines of wrong guidance, +15 of correct.

## Deliberately unchanged

Everything about **writes**, which remains true — `published` is read-only:

- Rule 5: "Only `dev` is writable (`published` is read-only)"
- `references/pre-flight.md`: "The backend allows `PUT` on the dev stage; `published` is …"
- `references/uip-pm-cli.md` § Stages: writes are dev-only

Correspondingly, the `data-mapping update --stage published` negative in
`tests/tasks/uipath-process-mining/mapping_fix_in_place.yaml` stays — it asserts
the write rule, not the query rule.

## Checks

- No `skill-flavor` markers in any of the three files, and no
  `skill-flavors/*/uipath-process-mining` overrides exist, so there is nothing to
  mirror.
- Frontmatter untouched (body-only edits), so this does not trip the
  `activation-gate.yml` recall-eval.
- `scripts/check-skill-verbs.py` on the skill: **0 new findings**. The one
  pre-existing Stale hit (`pm apps create My Process`, from a quoted app name in
  an example in `uip-pm-cli.md`) is untouched by this PR and is a tokenizer
  artifact, not a real verb.
- `uip pm ingestions create --stage` and `uip pm query --stage` both verified
  against the CLI directly.

## Note for reviewers

I have **not** re-verified the end-to-end behaviour against a live tenant — I do
not have one wired up. The correction rests on the CLI surface (`--stage` exists
on `ingestions create`) plus the report from @Cellcote that querying published
works after a published-stage ingestion. If anyone can confirm on a real app,
worth adding to the PR before merge; the original claim carried a "verified
against a live tenant" note and was still wrong, so a second pair of eyes is
warranted.


Cellcote:
Folded into #2547 rather than shipping separately.

The eval criterion removed in #2547 encoded exactly what the skill teaches, so the two halves are one change: dropping the criterion without correcting the skill would leave agents following the false guidance. Same commit (`fix(process-mining): published-stage querying works after a published-stage ingestion`), cherry-picked onto `feat/pm-evalset-coverage`.

Nothing lost — the full rationale, the `ingestions create --stage` evidence, and the "not verified against a live tenant" caveat are all carried over into #2547's description.

PR #2776, opened 2026-08-24
fix(uipath-maestro-flow): stop grading slow debug runs as flow failures

rockymadden:
## Motivation

`skill-flow-wiki-pageviews` failed a **gating** criterion with `weighted_score: 0.615`. The recorded cause in `review.json` was "the error output port was not connected."

It was connected. The artifact is correct, and it passes the unmodified checker — I re-ran it **9/9, 18–26 s per run**, on both CLI `1.202.0` and the eval image's own `1.201.0-dev.8319`. The element trace shows the error path executing exactly as designed:

```
_Implicit_HttpTask_fetchWikimediaPageviews1      SendTask          Failed
_Implicit_BoundaryError_fetchWikimediaPageviews1 BoundaryEvent     Completed
_Implicit_EndEvent_Failure_...                   EndEvent          Completed
_Implicit_PostSubprocessGateway_...              ExclusiveGateway  Completed
doneArticleNotFound                              EndEvent          Completed
```

The real failure was a `subprocess.TimeoutExpired` at `flow_check.py:138`. `run_debug` caps the subprocess **below** the CLI's own poll budget (600 s = `maxPolls 300 × pollInterval 2000ms`) at **41 of its 53 call sites**, and never passes `--timeout`. So a run that merely overran was SIGKILLed mid-poll, the exception escaped uncaught, and the criterion scored 0 with no payload, no `instanceId`, no incidents, and no retry. That erasure is also why the post-hoc review had to guess a cause — and guessed wrong.

This is suite-wide, not task-specific: 2 of 3 criteria in this task are live cloud execution, so infra variance is currently graded as agent error.

## Changes

**`tests/tasks/uipath-maestro-flow/_shared/flow_check.py`**

- **Pass `--timeout` to the CLI**, derived from the subprocess cap minus 60 s of headroom for the phases `--timeout` does *not* bound (solution upload, Studio Web debug provisioning, begin-session, create-instance — all unbounded `fetch` calls with no `AbortSignal`). The CLI now self-terminates with a parseable envelope *before* our cap fires. Fixes all 53 call sites without touching any of them.
- **Set `UIP_LOG_LEVEL=info`** so stderr carries `jobKey` / `instanceId` / Studio Web URL. The polling-timeout envelope carries none of them, so this is the only way to find the instance afterwards.
- **Treat the CLI's poll-budget expiry as transient** and retry it — matching on the message, because the CLI labels that path `Retry: "RetryWillNotFix"`, which is wrong for a poll timeout. Capped at 2 attempts: each burns the full budget, and 3 × a 300 s cap overruns the 600 s criterion budget in the task YAMLs.
- **Catch `subprocess.TimeoutExpired`** and fail through `_fail_with_capture` with partial stdout/stderr retained. `TimeoutExpired` returns **bytes** even under `text=True`, hence the `_as_text` helper.

**Docs — `UIPCLI_LOG_LEVEL` is a no-op.** The CLI reads `UIP_LOG_LEVEL` (`packages/common/src/logger.ts`); `UIPCLI_LOG_LEVEL` appears nowhere in the shipped bundle, including 1.201. Every documented `flow debug` invocation was prefixed with a variable that does nothing — the same blind spot that left this eval with no evidence. Renamed across the flow references and `preview/`, and corrected the claim that the level applies only to `flow debug` (it is global).

## Verification

- `207` flow-suite unit tests pass, `12` new — covering the derived `--timeout`, its floor, the env var (set + operator override), poll-timeout retry and its 2-attempt cap, the `TimeoutExpired` path retaining the `instanceId`, and `_as_text` decoding.
- The **unmodified** `check_wiki_pageviews_flow.py invalid_error` passes end-to-end against a live tenant with the patched `run_debug` — confirming the new `--timeout` flag does not break a real invocation. All solutions created during verification were deleted.
- The 2 collection errors in `tests/tasks/uipath-maestro-case/` are pre-existing on a clean `main` (verified via stash) and unrelated.

## Notes for reviewers

- **`preview/` is included** (2 lines, same env-var bug), which pulls in @dmetzgar / @tmatup. Happy to drop it into its own PR if you'd rather keep the reviewer set tight.
- **Not fixed here, deliberately:** `tests/tasks/uipath-maestro-case/_shared/case_check.py` has the identical defect with a 540 s cap under the same 600 s CLI budget — different skill and CODEOWNERS, so it wants its own PR. `tests/tasks/uipath-maestro-flow/canary/canary.py` uses 780 s, which is *above* the CLI budget, so it is not exposed to the SIGKILL path (it still lacks the log level and the `TimeoutExpired` guard).
- **Follow-up on the CLI side:** include `instanceId` / `jobKey` in the polling-timeout error envelope, fix the `RetryWillNotFix` label, and bound the pre-poll phases with an `AbortSignal`. This PR does not depend on any of it.
- **`skill-flow-wiki-pageviews` should be rescored** — its `FAILURE` is a false negative, and `review.json`'s stated cause retracted.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
@@ -68,40 +68,26 @@ _LAST_DEBUG_INPUT_IDS: set[str] = set()
 _LAST_DEBUG_PROJECT_DIR: str | None = None
 
 
-# The CLI polls the debug instance for `--timeout` seconds and only then gives
-# up with `Debug polling timed out after Ns`. Its default budget is 600s
-# (maxPolls 300 × pollInterval 2000ms), while every call site here caps the
-# subprocess well below that — 180/240/300s at most of them. Without an explicit
-# `--timeout` the checker therefore SIGKILLs the CLI mid-poll and keeps NOTHING:
-# no payload, no instanceId, no incidents, just a `subprocess.TimeoutExpired`
-# traceback. A merely slow cloud round-trip then reads as "the agent built the
-# flow wrong" (skill-flow-wiki-pageviews scored 0 on a gating criterion whose
-# artifact passes this same checker on re-run).
-#
-# So derive the CLI's budget from the subprocess cap and reserve this much for
-# the phases `--timeout` does NOT bound — solution upload, Studio Web debug
-# provisioning, begin-session, create-instance, all unbounded `fetch` calls in
-# the CLI. The CLI then self-terminates with a parseable envelope *before* the
-# subprocess cap fires, and the failure is diagnosable.
+# The CLI polls for `--timeout` seconds (default 600) before giving up, while
+# call sites here cap the subprocess at 180-300s. Without an explicit
+# `--timeout` we SIGKILL the CLI mid-poll and keep nothing — no payload, no
+# instanceId, no incidents. The headroom covers the phases `--timeout` does not
+# bound (upload, provisioning, begin-session, create-instance), so the CLI
+# always self-terminates first, with a parseable envelope.
 _CLI_TIMEOUT_HEADROOM_SECONDS = 60
 _MIN_CLI_TIMEOUT_SECONDS = 30
 
-# `UIP_LOG_LEVEL` — NOT `UIPCLI_LOG_LEVEL`, which the CLI never reads (see
-# packages/common/src/logger.ts; the string is absent from the shipped bundle).
-# At `info` the run narrates jobKey, instanceId and the Studio Web URL to
-# stderr. That is the ONLY way to find the instance after a timeout: the
-# polling-timeout error envelope carries none of them.
+# `UIP_LOG_LEVEL`, not `UIPCLI_LOG_LEVEL` — the CLI never reads the latter. At
+# `info` it narrates jobKey / instanceId / Studio Web URL to stderr, the only
+# way to find the instance after a timeout.
 _DEBUG_LOG_LEVEL = "info"
 
-# The CLI's own poll-budget expiry, matched on the message because the CLI
-# labels this path `Retry: "RetryWillNotFix"` — wrong for a poll timeout, which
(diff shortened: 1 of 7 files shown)


tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -68,6 +68,42 @@
...
+# The CLI's own poll-budget expiry, matched on the message because the CLI
+# labels this path `Retry: "RetryWillNotFix"` — wrong for a poll timeout, which
+# is precisely the case worth one more attempt.
+_DEBUG_POLL_TIMEOUT_MARKER = "debug polling timed out"
+
+# A poll timeout burns the whole subprocess budget on every attempt, so cap it
+# below `retries`: 3 × a 300s cap is 900s against task YAMLs that allow 600s for
+# the criterion, i.e. the third attempt could never finish. Other transient
+# markers fail fast and keep the full `retries` budget.
+_POLL_TIMEOUT_ATTEMPTS = 2
```
jiyangzh: nit: verbose

jiyangzh (approved):

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Agreed.

nikhil-maryala (approved):

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Trimmed — the block is down from ~35 comment lines to 12, keeping only what is not derivable from the code (the CLI 600s default, the `UIP_LOG_LEVEL` spelling, why the marker matches on Message, why the poll-timeout retry budget is below `retries`). Also folded the duplicated stdout+stderr lowercasing into `_output_blob`. 2110c155f

PR #2782, opened 2026-08-24
fix(uipath-maestro-case): make the bindings sidecar check executable and reachable

song-zhao-25:
## Problem

Step 12 Check 7 is prose: *"Compare `bindings_v2.json.resources[]` with the complete projection of top-level `caseplan.json.bindings[]`."* Across three live builds of the same SDD, the agent transcribed that check into its execution plan and shipped violations anyway — because `uip maestro case validate` returns `Valid` for every one of them.

The sidecar came out wrong **3/3**:

| Run | `bindings_v2` entry keys produced | What it is |
|---|---|---|
| 1 | `default, id, name, propertyAttribute, resource, resourceKey, type` | caseplan `bindings[]` dumped verbatim |
| 2 | `externalId, id, kind, name, properties` | invented from scratch |
| 3 | `default, id, name, propertyAttribute, resource, resourceKey, type` | dumped verbatim again |

Required is `{resource, key, value:{name:{defaultValue}, folderPath:{defaultValue}}, metadata:{subType}}`. Two of three runs made exactly the mistake `bindings-v2-sync.md` opens by warning against — *"This file uses a **different format**"*.

The consequence isn't cosmetic: `uip solution resources refresh` is the only consumer, so a wrong shape means Studio Web cannot resolve resources at deploy or debug, long after a clean `validate`.

## Why a script rather than more documentation

Four enforcement mechanisms already exist for this, all of them in the agent's context, all bypassed:

| Mechanism | Outcome |
|---|---|
| Links from the authoring site to `bindings-v2-sync.md` | 128 filename mentions in content it read; file never opened |
| Rule 14 / Step 12 Check 7 (self-attested) | transcribed into plan, not executed |
| Rule 24 (mandatory read-to-EOF, END marker) | honored for 6 references, skipped for this one |
| `scripts/audit_plan.py` (mandatory executable gate for `tasks.md`) | pattern works and is established — this PR reuses it |

Adding a fifth instruction is the same bet. What the failing cases have in common is that nothing produced an exit code.

## Placement matters

`package.json` `files` ships only `scripts/npm-package-lifecycle.mjs` from repo-root `scripts/`. So the existing `scripts/check_connection_parity.py` **never reaches consumers** — it is a CI-only gate. A build-time check has to live inside the packaged skill tree, next to `audit_plan.py`.

## Changes

**`skills/uipath-maestro-case/scripts/audit_bindings.py`** (new) — read-only, stdlib only, exit 0/1. Covers Checks 7 and 11 mechanically: sidecar entry shape, grouped parity of every distinct `resourceKey`, connection A/B/C parity, connector-node reachability, `resourceKey` self-consistency.

**Detection no longer trusts field values**, which is how a real build evaded the old check and reported `OK`:

- connections identified by `propertyAttribute`, not `resource == "Connection"` — a build writing `resource: "<connectorKey>"` counted zero connections
- connector usage derived from node **type** (`wait-for-connector` / `execute-connector-activity` tasks, `wait-for-connector` rules) instead of a text scan for `connectorKey` — a degraded node names no key at all, so the scan reported zero connector usage for a case that plainly calls one

**`references/implementation.md`** — Check 7 now invokes the audit as a mandatory `AUDIT OK` gate, with a `python3`-absent manual fallback and a cross-reference from Check 11.

**`scripts/check_connection_parity.py`** — **deleted.** The new audit is a strict superset (identical verdicts on four real builds, plus it catches the one the old check passed), and that file never shipped to consumers anyway. Its only repo consumer was the `connection_parity` e2e task, now repointed.

**`connection_parity.yaml`** — invokes the audit directly and asserts the **whole** bindings chain, not just the connection subset. For a connector-only case that adds sidecar-shape coverage: A/B/C can agree while the sidecar is a verbatim dump of caseplan `bindings[]`. Its SDD is unchanged — the non-connector paths (grouped parity, `resourceKey` composition) are no-ops on a single connector task, and are covered by the offline unit tests and by #2761's case (4 non-connector resources + 2 connections).

**`placeholder` semantics, decided explicitly:** a `placeholder` connectorKey remains the declared-unresolved stub contract and no longer trips the undeclared-connector check, while a node carrying *no* key is reported as a degraded shape. Flag if you'd rather a surviving placeholder hard-fail at Phase 3 exit.

## Verification

| Build | old script | this PR |
|---|---|---|
| reference (correct) | OK | **exit 0** |
| run 1 | BLOCKED | exit 1 |
| **run 2** | **OK — false negative** | **exit 1** |
| run 3 | BLOCKED | exit 1 |

The false negative is closed. Run 2 now reports what it was hiding: zero Connection sidecar entries, and `task 'Wait for Signature Result' carries no connectorKey in its context`.

`tests/scripts/test_audit_bindings.py` — 12 offline cases: all three observed sidecar shapes, tenant-UUID `resourceKey`, the inline-sibling `solution_folder` exception, both evasion paths, and the connector-only shape case the repointed e2e now covers. All pass.

**Risk worth naming:** `connection_parity` now also fails on a wrong-shaped sidecar. That is the intent, but it means a previously-green e2e can go red on a real defect — the big case got shape wrong 3/3, though this one is a single task and much simpler.

## Scope

Bindings and resources only. The other repeat failure from the same runs — unguarded `=js:` task-output dereferences, where `(vars.X || {}).Y` appeared **zero** times across all three builds — lives in caseplan expressions and is deliberately **not** covered here.

## Draft because

Requested review before merge. Independent of #2761 (contract-execution eval task) — no shared files, either merge order works.


song-zhao-25:
Closing in favour of a different approach.

The premise here was detection: an executable gate that catches a skipped `bindings_v2.json` sync. Measurement across 219 real solution trees on this machine showed the underlying problem is real and frequent — **25% of substantial builds (17/69) have a caseplan/sidecar mismatch, 11 of them with an empty sidecar** — but also showed this fix is the wrong shape for it.

Two reasons:

1. **The hole in the approach.** The failure is an agent not following a documented step. This PR adds another documented step that checks whether it followed the first one. An agent that skips the sync can equally skip the audit — both are prose in the same skill.

2. **The sidecar carries no original information.** It is a pure projection of caseplan `bindings[]`. A generator reproduces 140 of 145 correct sidecars exactly from the caseplan alone, and all 5 mismatches are cases where the generator is right and the file on disk is wrong (one uppercase `ConnectionId`, four inline-sibling `folderPath` values). Generating it makes three of this PR's four checks impossible by construction rather than detectable after the fact.

Replacing this with a generator that performs the `bindings-v2-sync.md § Regenerate` transformation directly. Follow-up worth filing separately: `uip solution resources refresh` already knows the sidecar is empty — it reports `Synced 0 resources` with `Result: Success` and `Warnings: []`. A warning there would surface this class of failure with no skill changes at all.

PR #2784, opened 2026-08-24
chore: refresh uip CLI catalog (1.202.0-dev.8372)

uipreliga:
Automated refresh against `@uipath/cli@1.202.0-dev.8372`. The catalog has 1544 reachable verbs after installing all available plugins.

Merges automatically once all checks pass (handled by `automerge-catalog-refresh.yml`). If a check fails, the failure is real drift (a CLI release removed/renamed a verb still referenced in skill docs or task YAMLs); investigate the gate annotations and either sweep the docs or, if the verb is intentionally retired, add it to `.claude/rules/cli-renames.md`. `/lint-task` and the CLI Verb Gate will surface the affected references.


celestinryf:
Closing this one: I dispatched the refresh workflow manually this evening, and this PR then hit the new required-review rule and would make tonight's scheduled 04:00 UTC refresh skip itself while it stays open. Closing restores the normal nightly flow; tonight's run rebuilds the same catalog (or newer) and opens a fresh PR.

PR #2796, opened 2026-08-25
feat(uipath-agents): document client-side tools

StefanPopaUi:
## Why

Client-side tools are [documented](https://docs.uipath.com/agents/automation-cloud/latest/user-guide/tools-client-side) and configurable in the agent designer, but the `uipath-agents` skill had no coverage. A coding agent editing project files had no way to discover the type or its required fields.

## What's here

**New:** `references/lowcode/capabilities/client-side-tools/client-side-tools.md`

The product documentation is the source of truth. When to use, the two-schema contract, the runtime sequence, tool-call confirmations, and limitations all follow docs.uipath.com and are attributed there.

Three additions go beyond the docs, each checked against source and labelled in the text:

| Addition | Verified against |
|---|---|
| `resource.json` field requirements — `type`, `referenceKey`, `location`, `properties.folderPath`, and which fields are non-optional | `clientSideToolResourceSchema` and `uip agent validate` |
| Schema keywords the pre-built form renders (`oneOf` with `const` + `title`, `format`, `minimum`/`maximum`, nested `properties`, `items`) | `AgentSchemaForm` in `@uipath/ui-widgets-conversational-agent-chat` |
| The generated form exposes no styling hooks | The widget's `ClientSideTool` component |

The casing note earns its place: `"ClientSide"` is what the Python models use and they tolerate either via a case-insensitive enum, but the zod schemas don't — the packager fails with an `invalid_union` / `No matching discriminator` error naming neither the bad value nor the legal set. There's a related trap worth knowing: an unrecognised discriminator makes validation skip the client-side checks, so a badly-typed resource can look *cleaner* than a nearly-correct one.

**Wired into:** the Capability Registry in `lowcode.md`, Task Navigation in `SKILL.md`, and the tool-kind router at the top of `capabilities/process/process.md`.

## On agent-type scope

An earlier revision said "conversational agents only". That's been removed, because the sources disagree:

- `tools-client-side.md` states it.
- `agent-tools.md` lists client-side tools among the general tool types with no agent-type qualifier.
- `clientSideToolResourceSchema` sits in the **general** tool union, not a conversational-scoped one.
- The designer permits adding them to non-conversational agents.

Rather than pick a side, the doc now states the constraint the documentation actually describes — the **surface** must support client-side tool handling and register a handler per tool, or the call returns an error. If the conversational-only restriction is real and enforced somewhere, that's worth correcting in `tools-client-side.md` too, since the general Tools page currently implies otherwise.

## Verification

`npm run skills:validate` passes (`default: 26 skills, 1722 files`; `studioweb: 174 replacements`). All four relative links in the new doc resolve.

The field requirements were exercised end-to-end on staging — two tools (`askApproval`, `chooseQueryVersion`) built this way pack, publish, deploy, and render.

I couldn't run `scripts/check-cli-verbs.py` locally; it needs PyYAML and I didn't want to install into the system Python. The doc references only `uip agent validate`.

[FOR-1419]: https://uipath.atlassian.net/browse/FOR-1419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-agents/references/lowcode/capabilities/client-side-tools/client-side-tools.md
```diff
@@ -0,0 +1,153 @@
+# Client-Side Tool Capability
```
maxduu, line 1: I'll let @norman-le review- one immediate thing is that we should make it clear that this is only applicable for conversational agents. Might want to reflect that in the lowcode agent/conversational-agent md files and critical-rules as well.

StefanPopaUi: HI @norman-le could you help us review this PR?

norman-le: I agree with Max here - we should explicitly mention this is for conversational agents. These tools cant be added in Agent builder for autonomous agents, and will also not work with them.

skills/uipath-agents/SKILL.md
```diff
@@ -50,6 +50,7 @@ Determine the agent mode before proceeding:
 | Create or scaffold a new low-code agent project | Low-code | [lowcode/lowcode.md](references/lowcode/lowcode.md) § Quick Start | `lowcode/project-lifecycle.md`, `lowcode/agent-definition.md` |
 | Edit `agent.json` (prompts, model, schemas, contentTokens, entry-points.json) | Low-code | [lowcode/lowcode.md](references/lowcode/lowcode.md) § Capability Registry | `lowcode/agent-definition.md` |
 | Add a low-code tool (Orchestrator process — RPA / agent / API / agentic — or Integration Service) | Low-code | [lowcode/lowcode.md](references/lowcode/lowcode.md) § Capability Registry | `lowcode/capabilities/process/*`, `lowcode/capabilities/integration-service/*` |
+| Add a client-side tool (executes on the client surface, not the server) to a low-code agent | Low-code | [lowcode/lowcode.md](references/lowcode/lowcode.md) § Capability Registry | `lowcode/capabilities/client-side-tools/client-side-tools.md` |
```
norman-le, line 53: Not sure if this needs the same wording as mentioned above to be explicit capabilities for conversational agents?

tests/tasks/uipath-agents/lowcode/client_side_tool/client_side_tool.yaml
```diff
@@ -0,0 +1,83 @@
+task_id: skill-agent-client-side-tool
+description: >
+  Client-side tools. End-to-end test: scaffold a standalone low-code
+  agent whose goal requires reading data that only exists on the client
```
norman-le, line 4: Dont need this or the test file.

Since client-side tools are conversational-only, a passing  test produces an agent that would fail at runtime. It's validating a configuration the product doesn't support.

PR #2800, opened 2026-08-25
Close the licensing Troubleshoot coverage gap

biftime:
## Why

The Coding Agents Scorecard reports **Licensing (uip platform)** at **50% Troubleshoot Product Coverage vs. Skill**. This PR closes that gap. Scope is deliberately confined to licensing.

I checked whether the 50% was simply untagged tests. It wasn't — licensing diagnosis existed nowhere:

- `uipath-platform` had no diagnose reference for licensing. The workflow guides carry `Error Conditions` tables, but those are keyed by **CLI error string** — a reverse lookup for an error you already hold. Nothing routed from a reported *symptom* to a layer, a command, and an interpretation.
- `uipath-troubleshoot` has no licensing domain under `references/products/`; licensing appears there only incidentally, as a cause inside orchestrator and UIA playbooks.
- All 12 licensing tasks are read/set operations. **Zero** were diagnose-shaped.

### 1. Licensing diagnose capability (`6c0506a`)

New `skills/uipath-platform/references/licensing/diagnose/`, following the `references/diagnose/` pattern already used by `uipath-admin` and `uipath-governance`:

| File | Contents |
|---|---|
| `CAPABILITY.md` | Capability index, 7 critical rules, anti-patterns |
| `references/troubleshooting-guide.md` | Five-step layer-first ladder |
| `references/failure-modes.md` | Seven symptom-keyed patterns |

The organising idea: licensing is **four independent layers** (account pool → tenant allocation → user bundle → Orchestrator slot), and tenant allocations vs. user bundles are *separate license types*, not a hierarchy. A user holding every bundle still cannot make an unattended job run. Answering at the wrong layer produces a confident wrong answer, so Step 1 routes to the layer before any command is issued.

Every claim is grounded in behaviour already documented in the workflow guides. Where reconciliation would need an undocumented rule, the guide says to escalate rather than derive it. The failure-mode patterns deliberately do **not** restate the existing error tables — they link to them.

**Scope boundary:** this diagnoses licensing *state*. `uipath-platform`'s description routes root-cause to `uipath-troubleshoot`, so Step 5 settles the entitlement question and hands the runtime causal chain onward with its evidence.

Plus 3 `mode:diagnose` tasks — the first integration-tier licensing tasks. They grade diagnostic behaviour: that the agent reads the group **rule** rather than stopping at the user record, and that it never issues a `set` while diagnosing (a `set` is an overlay that destroys the evidence of the prior state).

### 2. Tag the licensing tasks (`f88d546`)

`tests/README.md` and `.claude/rules/test-writing.md` require `skill` + `tier` + `mode:*` + `lifecycle:*` on every task. None of the 12 licensing tasks carried `mode:*` or `lifecycle:*`.

Not cosmetic: the scorecard splits coverage by mode, so a task with no `mode:*` tag is invisible to its product's Build / Operate / Troubleshoot columns — it passes in the weekly run and still contributes 0% to the number it should be moving.

All 12 are read/set operations against live tenant state, so they are `mode:operate`; lifecycle is `discover` for reads and `setup` for the three `set` tasks.

### 3. Ground the criteria in observed CLI behaviour (`c6ab3a0`) — addresses the lint advisory

**Three tasks could pass without the CLI ever running.** `groups_rules_details` and `groups_rules_list` graded only `file_exists` plus `file_contains: ['"Result"']` — writing `{"Result":"x"}` to the expected path scored full marks having invoked nothing. `licensing_read_e2e` was better off (`licensing_check.py` validates envelope shape and `Result == "Success"`) but a hand-authored envelope satisfies a shape check too. All three now carry the `_shared/uip-shim` sandbox config and grep `.uip-cmdlog` for the verbs they claim to exercise, matching the 12 siblings that already did. `licensing_read_e2e` also gains the read-only guard its prompt always implied but nothing checked.

> The lint suggested grepping for `uip platform groups rules list`. **That verb does not exist** — `assets/uip-catalog-snapshot.json` has `groups rules get`, `details`, `set`. Used `get`.

**Two diagnose tasks were self-report-heavy**, which the repo's own rule against grading self-reports warns about:

| Task | Before | After |
|---|---|---|
| `diagnose_allocation_no_effect` | 5.0 behavioural vs 7.5 self-written | **6.0 vs 5.0** |
| `diagnose_consumables_scope` | 3.5 vs 8.5 | **7.0 vs 6.0** |
| `diagnose_user_missing_bundle` | 8.0 vs 3.5 | unchanged |

Rebalanced onto the documented weight scale: reads and the no-mutation guard are primary validation (3.0), the written analysis is artifact content (2.0), its existence is supporting (1.0). `diagnose_consumables_scope` also gains a genuinely behavioural criterion — the agent must run the summary **both** scoped and unscoped, which is the comparison the failure-modes guide prescribes for telling suppression apart from missing consumption. That grades the diagnostic move rather than an assertion about it.

Every new guard was verified in both directions, including that an absent `.uip-cmdlog` fails rather than passing vacuously. The unscoped check is written `... | grep -v -- '--tenant' | grep -q .` rather than `grep -vq`, because `-vq` is not consistent across grep implementations.

## What this does and does not move

Tagging the 12 existing tasks **does not raise the Troubleshoot number** — they are correctly `mode:operate`. It makes the number *truthful*, not higher. Licensing's genuine `mode:diagnose` count goes 0 → 3, all from commit 1.

## Reviewer notes

- **The `uipath-troubleshoot` handoff has no receiver.** Step 5 hands runtime causality to a licensing domain that doesn't exist yet. Natural follow-up.
- The scorecard reports Licensing at **7/7** tests; the repo had **12** before this PR (15 after). Worth reconciling before treating the 64% Test Coverage figure as precise.
- **Domain accuracy** of the seven failure-mode patterns (absolute vs. delta quantity, overlay behaviour, bundle windows, scope suppression) is drawn from the existing workflow guides rather than verified against a live tenant. A domain expert should confirm.

## Two findings for the scorecard owners, deliberately not fixed here

Both are out of scope for a licensing PR, but they affect how the whole Troubleshoot column reads:

1. **144 tasks repo-wide carry no `mode:*` tag** (479 have some tag problem), and nothing enforces the rule. Every one contributes 0% to its product's mode coverage while still passing. I had a validator + changed-files CI gate in this PR and removed it to keep the scope clean — happy to raise it separately.

2. **`uipath-troubleshoot` holds 264 diagnose tasks**, all tagged `uipath-troubleshoot`, so the per-product Troubleshoot columns cannot see any of them. The repo centralises diagnosis in one skill while the scorecard measures it per product row. Products with their own diagnose tasks score well (admin 30, governance 17); products relying on the central skill score badly. Licensing at 50% is one instance of a column-wide measurement mismatch.

Also worth a separate fix: 3 llmgateway tasks are tagged `mode:troubleshoot`, which is not a valid value (the column is fed by `mode:diagnose`), so they silently drop out of every mode slice. Reverted here to keep this PR licensing-only.

## On the lint's Theme 1

The advisory flags all three diagnose tasks for the self-report axis and suggests *"consider whether a carve-out for diagnose-mode tasks is warranted."* That is a rubric question for whoever owns `.claude/commands/lint-task.md`, not something to settle in this PR. The rebalance above narrows the exposure; the residual tension is real — for a diagnose task the written analysis genuinely **is** the deliverable, not a meta-report about a separate action.

## Verification

- All relative links and anchors in the new references resolve
- All 15 licensing tasks carry valid `mode:*` + `lifecycle:*`; every task now has at least one behavioural criterion
- `check-skill-status.py` and `check-task-driver.py` pass
- SKILL.md frontmatter valid, description 981 chars (cap 1024)

### DIFF
--- a/skills/uipath-platform/SKILL.md
+++ b/skills/uipath-platform/SKILL.md
@@ -181,6 +181,7 @@ Choose the appropriate operation from the Task Navigation table below. For `uip 
 | **Assign user/group license bundles** | [references/licensing/user-licenses-allocations.md](references/licensing/user-licenses-allocations.md) |
 | **Report on license consumption** | [references/licensing/consumables-report.md](references/licensing/consumables-report.md) |
 | **Understand licensing concepts** | [references/licensing/licensing.md](references/licensing/licensing.md) |
+| **Diagnose a licensing symptom** | [references/licensing/diagnose/CAPABILITY.md](references/licensing/diagnose/CAPABILITY.md) |
 | **Full CLI command reference** | [references/uip-commands.md](references/uip-commands.md) |
 | **Build/run/validate coded workflows** | [/uipath:uipath-rpa](/uipath:uipath-rpa) |
 
@@ -313,7 +314,7 @@ Every `uip` command accepts:
 - **[Data Fabric](references/data-fabric/data-fabric.md)** — Entity schemas, records CRUD, query filters and aggregates, choice sets, file attachments, CSV bulk import, folder scoping
 - **[LLM Gateway — BYO Connections](references/llmgateway/byo-connections.md)** — Register tenant-owned LLM keys against UiPath products
 - **[Guardrails — BYOG Configurations](references/guardrails/byo-configurations.md)** — Manage tenant-registered bring-your-own guardrail (BYOG) configurations and diagnose their underlying Integration Service connections
-- **[Licensing](references/licensing/licensing.md)** — Tenant allocations, user/group bundles, consumables reporting
+- **[Licensing](references/licensing/licensing.md)** — Tenant allocations, user/group bundles, consumables reporting, [diagnose](references/licensing/diagnose/CAPABILITY.md)
 - **[Coded Workflows](/uipath:uipath-rpa)** — Building coded automation projects
 
 > **Trouble?** If something didn't work as expected, use `/uipath-feedback` to send a report.
--- a/skills/uipath-platform/references/licensing/diagnose/CAPABILITY.md
+++ b/skills/uipath-platform/references/licensing/diagnose/CAPABILITY.md
@@ -1,0 +1,79 @@
+# Diagnose — Investigate Licensing State Failures
+
+Capability index for diagnosing licensing symptoms via `uip platform`: entitlement gaps, allocation changes that did not take effect, and consumption reports that read wrong.
+
+> **Where you came from / where to go next.** Diagnose is downstream of Operate (a user lost access, an allocation did
+> not apply, a report read wrong). Fixes — `tenants licenses set`, `users licenses set`, `groups rules set` — are Operate
+> actions requiring explicit user consent after root cause analysis.
+>
+> **Scope boundary.** This capability diagnoses **licensing state**: which layer holds (or fails to hold) an entitlement,
+> and why a licensing command's result differs from what the user expected. It does NOT diagnose runtime faults that
+> merely *mention* a license (a job stuck Pending, Studio failing to start, a robot not connecting). For those the
+> licensing layer is one hypothesis among many — establish the entitlement facts here, then hand the runtime causal chain
+> to `uipath-troubleshoot`.
+>
+> **Inherits universal rules from [SKILL.md](../../../SKILL.md).** Use `uip platform ... --output json` for all diagnostic reads.
+
+## When to use this capability
+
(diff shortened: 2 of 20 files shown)


biftime:
@claude review

vstoleru-uipath (approved):

denispetre (approved):

busesorin94 (approved):

PR #2801, opened 2026-08-25
feat(uipath-aops): add StudioAdmin AOps skill for SourceControl + CICD pipelines

AndreiRR24:
## What

Adds `skills/uipath-aops` — a skill that teaches agents to compose a `PipelineDto` binding an SC-synced Studio Solution or standalone project to an already-published Orchestrator process, then create, run, and follow the pipeline.

Five test tasks under `tests/tasks/uipath-aops/`, plus CODEOWNERS, `assets/skill-status.json`, and `skills.sh.json` registration.

## Why now

The skill was drafted against `uip sc` + `uip cicd`. Those two tools were unified into `@uipath/aops-tool` (prefix `uip aops`) by [cli#2321](https://github.com/UiPath/cli/pull/2321), and `@uipath/cicd-tool` no longer resolves on npm. A skill written against the old prefixes emits commands that cannot run at all, so this lands the skill against the current surface rather than shipping a broken one. All 44 `aops` verbs it documents are present in `assets/uip-catalog-snapshot.json`.

## Behavior deltas the skill has to teach

The verb trees merged without renaming a subcommand, but the surrounding behavior moved. Each of these silently breaks a workflow:

| Change | Why it matters |
|---|---|
| stdout `Data` keys are PascalCased; `--file` input stays camelCase | Capturing output and feeding it back is now a trap — `pipeline update` rejects such a file outright |
| `pipeline update --file` rejects unrecognized / miscased keys (exit 3) | Previously a full-replace PUT would silently wipe the fields it failed to carry |
| `pipeline delete` requires `-y` / `--yes` | The CLI never prompts; without the flag nothing happens |
| `repo project-files <repo-id>` returns the project path | Replaces the "CLI gap → probe or ask the user" workaround for standalone projects |
| `--output-filter` refused when `--limit` came from its default | Filter would otherwise run over a silently capped page |
| `ConnectionType` is `Azure`, not `AzureDevOps` | Provider branching would never match |
| `pipeline executions` has no `--take`; `--offset` must be a multiple of `--limit` | Old pagination examples fail client-side |
| `execution logs --follow` defaults to a 30 min timeout and streams to stderr | Documented as 10 min and stdout |

`references/migration-guide.md` carries the full verb map plus these deltas, so an agent that hits a stale command can translate without re-deriving it. `.claude/rules/cli-renames.md` gains `sc` → `aops` and `cicd` → `aops` so `/lint-task` can suggest the replacement for any task still matching the old prefixes.

Unchanged and preserved: the `PipelineDto` / `EditPipelineDto` shapes, `repositoryId` = `RemoteId` (not the SourceControl UUID), `projectName: ""`, `repositoryType: "git"`, `branch` as the only hard-required field, and the process-arguments model.

## Naming

The skill is `uipath-aops`, tracking the CLI tool. `uipath-governance` already owns "AOps" for `uip gov aops-policy` — a different product surface that happens to share the word — so the description carries a `→` redirect to it, and the When-to-Use section calls the split out explicitly.

## Tests

One YAML per leaf directory, per the current template naming:

| Task | Covers |
|---|---|
| `discovery/discovery_smoke` | Connection → source → repos → processes sequence |
| `pipeline-create/pipeline-create_e2e` | Full workflow through `pipeline create --file`, with skill-trigger and DTO-shape assertions |
| `dto-shape/dto-shape_field_mapping` | Field mapping and camelCase input keys |
| `repository-id/repository-id_remote_not_uuid` | Extracts `RemoteId` from a raw two-row `connection repos` payload — nothing labels which id is correct |
| `cli-migration/cli-migration_legacy_prefix` | Request stated in the retired `uip sc` / `uip cicd` vocabulary; agent must translate, not copy |

Every task adds `command_not_executed` guards on the retired prefixes. `agent:` and `sandbox:` blocks are omitted — they only restated experiment defaults (`prune-task-defaults.py --dry-run` reports 0/5 to update).

## Note on the `uip-check-skip` markers

`check-skill-verbs.py` flags this skill's deliberate references to the retired verbs. The migration guide's verb map is restructured to one row per subject so its escaped-pipe cells stop parsing as verb paths, and the lines that name `sc` / `cicd` on purpose carry the `<!-- uip-check-skip -->` marker the checker documents for historical references.

In frontmatter the marker rides a YAML comment, so the parsed `description` and `when_to_use` stay clean (611 / 453 chars) while keeping the legacy prefixes as matching signal for stale queries.

## Checklist

- [x] Folder is `skills/uipath-<kebab-case>`; `name` matches the folder
- [x] `description` under the 1024 cap — `hooks/validate-skill-descriptions.sh` passes
- [x] Critical Rules section present and numbered
- [x] No structural cross-skill dependencies
- [x] All relative links resolve
- [x] CODEOWNERS updated
- [x] Status recorded in `assets/skill-status.json` (`preview`), README table regenerated, no status marker in SKILL.md
- [x] Grouped in `skills.sh.json`
- [x] `check-skill-status.py`, `check-cli-verbs.py`, `check-skill-verbs.py`, `check-skills-sh.py`, `check-task-driver.py`, `skills:validate`, `skills:build`, `version:check` all pass locally
- [x] No secrets or personal paths

Not yet done: a coder-eval passing-run claim — the tasks need a tenant-less run to record.


mihai-mocanasu-uip (approved):
looks ok

rockymadden (approved):

PR #2806, opened 2026-08-25
docs(functions): document packOptions for controlling package contents

andreibalas-uipath:
## Why

Root-caused a `skill-functions-deploy-tenant` failure (codex, `gpt-5.6-terra`, score 0.47): the grader asserts `packOptions.directoriesExcluded` includes `data`, but `skills/uipath-functions/SKILL.md` never mentioned `packOptions` (or any package-content control). The agent excluded `data/` via `.uipathignore` — which genuinely works with the shipped CLI — and was docked. The only place `packOptions` was documented was the sibling `uipath-agents` skill, which functions can't depend on.

## What

Adds a **"What Goes Into the Package"** subsection under Step 8 (Pack and Publish) in `skills/uipath-functions/SKILL.md`, mirroring `skills/uipath-agents/references/coded/lifecycle/deployment.md`:

- what `pack` puts in the `.nupkg`
- `packOptions` example (`fileExtensionsIncluded` / `filesIncluded` / `filesExcluded` / `directoriesExcluded` / `includeUvLock`), with `data` in the exclusion list

Claims verified against real `uip function pack` output (1.202.0): without `packOptions`, `content/data/fixtures.json` is packed; with `directoriesExcluded: ["data"]` it is not; `uv.lock` is included when present; `__pycache__`/`.env` are excluded by default.

## Eval

`coder-eval run tests/tasks/uipath-functions/deploy_tenant/deploy_tenant.yaml -e experiments/default.yaml -T codex -m gpt-5.6-terra`

| Skill | Result | Notes |
|---|---|---|
| before | 1.0 (48→95 s) | **Not a real pass** — agent `rg`'d the skill for "exclude", found nothing, then read `tests/.../check_deploy_tenant.py` and copied the expected answer |
| after | 1.0 (48 s) | Agent read SKILL.md, wrote `packOptions.directoriesExcluded: ["data"]` directly; never touched `tests/` |

`npm run skills:validate` clean; no flavor overrides exist for `uipath-functions`.

## Not in this PR

- The grader still scores a working `.uipathignore`-only solution 0; an outcome-based nupkg check in `check_deploy_tenant.py` would make it robust.
- Local/tempdir runs can pre-solve tasks by reading `tests/` under `$SKILLS_REPO_PATH`.

### DIFF
--- a/skills/uipath-functions/SKILL.md
+++ b/skills/uipath-functions/SKILL.md
@@ -269,8 +269,6 @@ uip function push
 }
 ```
 
-Verify after `pack`: `unzip -l .uipath/<NAME>.<VERSION>.nupkg` must not list excluded paths.
-
 ## Important Notes
 
 - `UiPath()` must never be instantiated at module level — always inside a function body
--- a/skills/uipath-functions/SKILL.md
+++ b/skills/uipath-functions/SKILL.md
@@ -253,6 +253,24 @@ To sync to Studio Web instead of publishing to Orchestrator:
 uip function push
 ```
 
+#### What Goes Into the Package
+
+`.nupkg` produced by `pack` contains project files (source, `pyproject.toml`, `uipath.json`, `uv.lock` when present) and generated metadata (`entry-points.json`, `bindings_v2.json`, `package-descriptor.json`, `operate.json`). Control inclusion via `packOptions` in `uipath.json` — keep local-only fixtures, test data, and caches out of the published artifact:
+
+```json
+{
+  "packOptions": {
+    "fileExtensionsIncluded": [".py", ".json"],
+    "filesIncluded": ["config.yaml"],
+    "filesExcluded": ["test_*.py"],
+    "directoriesExcluded": ["data", "tests", "__pycache__"],
+    "includeUvLock": true
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


skills/uipath-functions/SKILL.md
```diff
@@ -253,6 +253,22 @@ To sync to Studio Web instead of publishing to Orchestrator:
...
 ```
 
+#### What Goes Into the Package
+
+`.nupkg` produced by `pack` contains project files (source, `pyproject.toml`, `uipath.json`, `uv.lock` when present) and generated metadata (`entry-points.json`, `bindings_v2.json`, `package-descriptor.json`, `operate.json`). Control inclusion via `packOptions` in `uipath.json` — keep local-only fixtures, test data, and caches out of the published artifact:
+
+```json
+{
+  "packOptions": {
+    "fileExtensionsIncluded": [".py", ".json"],
```
radu-mocanu, line 273: we should document here the default file extensions included https://github.com/UiPath/uipath-python/blob/fea41bb8224e62ac945379954bd719f3ecd5adf0/packages/uipath/specs/uipath.spec.md?plain=1#L91

radu-mocanu (approved):

nikhil-maryala (approved):

PR #2809, opened 2026-08-25
feat(uipath-maestro-flow): Flow→IXP design-time handoff e2e [RE-12982]

joe-prosser:
Adds the e2e for [RE-12982](https://uipath.atlassian.net/browse/RE-12982): a user in a Flow context supplies documents and asks for an extraction flow, without knowing an IXP design-time project must be built first. The prompt never says "IXP", "project", or "publish" — recognising the prerequisite and handing off to `uipath-ixp` is the behaviour under test, asserted via `skill_triggered` plus tenant-verified outcomes (project created, folder-deployed into the run folder, flow wired or correctly mock-degraded).

**Skill changes the test iteration forced:**
- `uipath-maestro-flow` Critical Rule 4 gains a narrow carve-out (IXP extraction + user-supplied documents + no covering extractor → invoke `uipath-ixp`, then resume). The flow skill resolves the target Orchestrator folder (per rule 5) and passes it in the handoff; the tenant writes need no separate consent gate because that folder choice is the one irreversible decision, and rule 5 already routes it.
- The ixp plugin docs describe the handoff outcome instead of carrying the sibling's commands, prescribe a bounded registry poll with a mock + `DeploymentName`-breadcrumb fallback (flow-registry indexing latency is org-scoped: ~25s healthy, hours observed), and now state that folder deployment — not `publish` — is what surfaces a model in the flow registry.
- `uipath-ixp` documents the deploy path (`create → list-models → deployments create`) in a Deployment Guide with a headless exit (no folder identified → stop and report, never guess or create one); the inbound-handoff exception makes the caller-supplied folder the contract; labelling is skipped only when the model unblocks a larger build in this session, or on explicit opt-out.

**Self-cleaning by design:** seed creates a fixed-name run folder (`flow-e2e-e9fbf2d0`, the `tests/tasks/uipath-platform/cleanup.py` uuid8 convention) that the prompt names; teardown deletes it, which removes the deployment and its registry node. Attribution is exact — wired into this sandbox's flow, or `FolderKey ==` the run folder — and nothing else is ever deleted or guessed at. A leftover from a failed teardown is found by exact-name lookup and healed by the next seed; a same-named folder younger than one task budget is treated as a live concurrent instance and refused (the folder API has no timestamp, so seed stamps its own into the folder Description at create). Deployments have no delete API; five domains were burned during development, the last by a teardown that read a failing `folders get` as "already gone" — fixed here, the residue removed by hand.

Evidence (current design): [33536499571](https://github.com/UiPath/skills/actions/runs/33536499571) and [33548029622](https://github.com/UiPath/skills/actions/runs/33548029622) both 1.000 on the wired path and left the tenant exactly as found (no `LEAKED`, no `NOT DELETED`); [33550441340](https://github.com/UiPath/skills/actions/runs/33550441340) on the merged head: 1.000, wired path, folder stamped and cleaned (third consecutive clean run, first against main's updated `validate_flow.py`). Earlier greens ([32968685992](https://github.com/UiPath/skills/actions/runs/32968685992), [32997068855](https://github.com/UiPath/skills/actions/runs/32997068855), [33056705206](https://github.com/UiPath/skills/actions/runs/33056705206), [33059578513](https://github.com/UiPath/skills/actions/runs/33059578513), [33162961197](https://github.com/UiPath/skills/actions/runs/33162961197)) predate the run-folder design and two of them leaked a folder. 51 grader unit tests, green in Test Helpers on the merged head (33550422547); task lint clean apart from the documented validate-only trade-off.

[RE-12982]: https://uipath.atlassian.net/browse/RE-12982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


richsilv (approved):
From my perspective, the proof is in the pudding. Conceptually, this all makes sense. I don't know if @rockymadden needs to okay the changes to the uipath-maestro-flow skill.

skills/uipath-ixp/references/project-setup-guide.md
```diff
@@ -1,6 +1,6 @@
 # Project Setup Guide
 
-Complete workflow for creating a **new** IXP project, labelling all documents, and getting initial metrics. Run all steps end-to-end automatically.
+Complete workflow for creating a **new** IXP project, labelling documents when quality matters, and getting initial metrics. Run steps 1-4 end-to-end automatically; Step 5 (deploy) only when the user's goal calls for it.
```
cezara98t: "when quality matters" - when would this be exactly? Why not all?

skills/uipath-ixp/references/project-setup-guide.md
```diff
@@ -85,6 +85,34 @@ uip ixp projects update-title <project-name> "Vendor Invoices" --output json
 
 Skip this step if the user already provided a meaningful name in Step 1.
 
-## Step 4 — Label All Documents
+## Step 4 — Decide whether labelling is needed
 
-Follow the [Label Documents Guide](label-documents-guide.md) to label every document in the project.
+**Labelling improves extraction quality; it is NOT required for a working,
```
cezara98t: Is this added only for tests? I don't think a real use case would be an extractor with low quality.

skills/uipath-ixp/references/project-setup-guide.md
```diff
@@ -85,6 +85,34 @@ uip ixp projects update-title <project-name> "Vendor Invoices" --output json
...
+callable model.** A trained version appears on its own within seconds of
+`projects create`.
+
+| The user wants… | Do this |
+| --- | --- |
+| A model an automation / Maestro Flow can call (documents + a downstream goal, not a quality target) | Skip labelling; go to [Step 5](#step-5--deploy-so-it-can-be-called). Say quality is untuned and labelling is the lever if fields come back wrong. |
+| Better extraction quality, or they named accuracy/F1/scores | Follow the [Label Documents Guide](label-documents-guide.md) for every document. Deploy (Step 5) only if they also want a callable model. |
+| Unclear | Skip labelling — cheap to add later (`deployments upgrade` moves a deployment to a newer version). Confirm before deploying: a deployment cannot be un-deployed. |
+
+## Step 5 — Deploy so it can be called
```
cezara98t: This reference file is related only to setting up a new project and deploying the model is not part of it. We can add a new reference file if needed. 
Also, deploying to a folder is not the only possible final step, you can also deploy to a project. This could be a good enough reason to create a new reference file for describing the final step (deployment).

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: Yeah maybe this is poorly phrased. Maybe it should be "when annotation quality matters" or something like that? The way I see it people will use the skill in two ways here:
- The first is where they want the skill to do everything including labelling (this is the implementation as written today) 
- The second is where they get the skill to scaffold the end to end project and then do the annotations themselves e.g. to convince themselves that the annotations are good, or to minimize claude costs.

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: This is also added for the scaffolding use case

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: Yeah fair, makes sense. Will change.

skills/uipath-ixp/references/project-setup-guide.md
```diff
@@ -87,4 +87,11 @@ Skip this step if the user already provided a meaningful name in Step 1.
 
 ## Step 4 — Label All Documents
 
-Follow the [Label Documents Guide](label-documents-guide.md) to label every document in the project.
+**Default:** follow the [Label Documents Guide](label-documents-guide.md) to label every document in the project.
+
+Labelling is optional — it produces the **project score** (`get-metrics` reports nothing until documents are confirmed) and is not required for a callable model (a trained version appears on its own within seconds of `projects create`). Skip it only when:
+
+- **The request is a model for an automation to call** — a flow, activity pack, or other runtime consumer, with no score, metrics, or accuracy target named. Deploy instead per the [Deployment Guide](deployment-guide.md).
+- **The user opts out** — says to skip labelling, or that they (or an SME) will label. Stop after Step 3 and hand over the project name; they can label in-product or via the [Label Documents Guide](label-documents-guide.md) later.
```
cezara98t: "they (or an SME) will label" -> can make the agent biased and only skip in those cases. It's been taking examples literally word-for-word lately - we had issues in our tests.
What do you think of "or that somebody else will handle labelling"?

skills/uipath-ixp/references/deployment-guide.md
```diff
@@ -0,0 +1,24 @@
+# Deployment Guide
+
+The final step of a project's lifecycle: making a trained model version available beyond the labelling loop. Two distinct targets — pick from what the user needs, they are not interchangeable:
+
+| Target | Command | What it gives you |
+| --- | --- | --- |
+| **Project** ("publish") | `uip ixp projects publish <project-name> --output json` | Marks the version published *inside* the project — version pinning, `live`/`staging` tags, rollback. Nothing outside the project can call it. |
```
cezara98t: "Nothing outside the project can call it." -> this is wrong. Once published, the extractor can be used via Framework.

skills/uipath-ixp/references/deployment-guide.md
```diff
@@ -0,0 +1,24 @@
...
+| **Orchestrator folder** | `uip ixp deployments create <project-name> --version <N> --folder-key <guid> --output json` | Makes the model callable at runtime by activity packs and Maestro Flow. Neither labelling nor `publish` is required first. |
+
+Runtime callers — Maestro Flow nodes and IXP activity packs alike — resolve the model through the **folder** target. Publish additionally when the user wants the version marked published; see the "Publish the model" row in [SKILL.md Task Navigation](../SKILL.md#task-navigation) for tags, rollback, and unpublish.
```
cezara98t: An idea for rephrasing.
> Runtime callers split by how they address the model: folder-resolving callers (Maestro Flow nodes; anything reading folder deployments) see only the folder target, and tag-resolving callers (the DU framework API and its activities) see only the project target's tags. Neither substitutes for the other — pick per consumer, or do both.

skills/uipath-maestro-flow/references/author/references/plugins/ixp/planning.md
```diff
@@ -38,7 +38,7 @@ Don't use IxP as a generic OCR. IxP is field-oriented — it extracts named fiel
 
 - `uip login` — IxP nodes only appear in the registry after authentication.
 - `uip maestro flow registry pull --force` must be run to cache IxP model node types locally.
-- A trained, published IxP extraction model must exist on the tenant. If none exists, surface it in the **Open Questions** section of the architectural plan so the user can train one while reviewing.
+- A trained IxP model must be **deployed to an Orchestrator folder** (`deployments create`) — folder deployment, not `projects publish`, is what puts it in the flow registry (publish only marks a version inside the project). If none exists, see [No published model](#no-published-model-branch-on-documents).
```
cezara98t: Suggestion for this line: drop the `projects publish` contrast. What publish does or doesn't do is the uipath-ixp skill's knowledge to own.

> A trained IxP model must be deployed to an Orchestrator folder — the flow registry lists folder deployments only. Create one via the uipath-ixp skill. If none exists, see [No published model].

Same publish-contrast appears in other places too.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+initial_prompt: |
+  I've got a batch of scanned falconry licences in ./documents/ — have a look
+  at them. Build me a UiPath Maestro Flow that reads licences like these and
+  pulls out the licence number, ring number, species, falconer and the date the
+  licence expires, then logs the ring number and expiry.
+
+  Trigger it manually. Set up a fresh Orchestrator folder for anything this
+  needs deployed and use that folder — don't put anything in the existing ones.
+  The task is not complete until `uip maestro flow validate` passes on the flow.
```
cezara98t: Do we want to mention commands in the prompts? I usually write them like a user would, which wouldn't look into cli commands.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+  - The `uip` CLI is already available and the runner is logged in.
+  - Use `--output json` on all uip commands.
+  - Do NOT run `uip maestro flow debug` or `uip maestro flow deploy`.
```
cezara98t, line 99: Nit: makes sense on a live tenant, so keep it; but it does mean we never find out whether the agent would have deployed on its own. Might be worth a small smoke test at some point.

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: Addressed in 3483841b6 — labelling is now the default for setup; it skips only on the two objective conditions in Step 4.

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: Addressed in 3483841b6 — labelling now defaults on; the only remaining skip is the automation-consumer path (the RE-12982 flow-handoff scenario), and the agent must state the model is unscored.

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: Addressed in fbc3b4d53 — deployment extracted to references/deployment-guide.md, covering both targets (project publish + Orchestrator folder).

skills/uipath-ixp/references/project-setup-guide.md
joe-prosser: Addressed in e3cb92625 — took your wording.

skills/uipath-ixp/references/deployment-guide.md
joe-prosser: Addressed in e3cb92625 — aligned with #2831's semantics.

skills/uipath-ixp/references/deployment-guide.md
joe-prosser: Addressed in e3cb92625 — adopted your text near-verbatim.

skills/uipath-maestro-flow/references/author/references/plugins/ixp/planning.md
joe-prosser: Addressed in e3cb92625 — took your sentence for planning.md, and removed the only other publish-contrast (impl.md § Open Questions).

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
joe-prosser: Addressed in 2a0dd749b — dropped the validate sentence from the prompt entirely; grading stays outcome-based. 3/3 passing runs since, with the skill validating unprompted. The remaining command mentions (--output json, the debug/deploy prohibitions) are deliberate harness guardrails.

joe-prosser:
Addressed all review feedback — labelling default-on (3483841b6), deployment guide extracted + corrected per #2831 (fbc3b4d53, e3cb92625), publish-contrast removed from the flow skill, validate command dropped from the test prompt (2a0dd749b). Validated with 3 consecutive 1.000 e2e runs on HEAD ([1](https://github.com/UiPath/skills/actions/runs/33054047267), [2](https://github.com/UiPath/skills/actions/runs/33056705206), [3](https://github.com/UiPath/skills/actions/runs/33059578513)).

cezara98t (approved):

skills/uipath-ixp/SKILL.md
```diff
@@ -63,6 +63,7 @@ Do not answer these from this skill. Re-activate `uipath-maestro-flow` and follo
 2. To create a new project: follow [Project Setup Guide](references/project-setup-guide.md)
 3. To improve an existing project: follow [Improve Prompts Guide](references/improve-prompts-guide.md)
 4. To label documents on an existing project: follow [Label Documents Guide](references/label-documents-guide.md)
+5. To deploy a model so an automation can call it from an Orchestrator folder (Maestro Flow, other folder-resolving callers): `projects create` → `list-models` → `deployments create --folder-key`. Neither labelling nor `publish` is required — see [Deployment Guide](references/deployment-guide.md).
```
rockymadden, line 69: **Major — the sibling still says "don't".**

This PR teaches `uipath-maestro-flow` to invoke this skill mid-flow, but `SKILL.md:20-28` here is untouched:

> STOP and invoke the `uipath-maestro-flow` skill instead when any of these hold: … The working context is a `.flow` file or a Maestro flow rather than a standalone IXP project.
> Do not answer these from this skill. Re-activate `uipath-maestro-flow` … This overrides Critical Rule 1.

The handoff lands in precisely that state, and the frontmatter `description` repeats it ("DO NOT TRIGGER during .flow / Maestro Flow work"). Rule 4 was a coin flip because two documents disagreed; this leaves the same disagreement one hop away.

Fix: reciprocal carve-out in *When NOT to Use* and in the `description` — something like "Exception: when `uipath-maestro-flow` hands off user-supplied documents to build a missing extractor, do the project-creation + folder-deploy work here and hand back." Cheap, and it closes the loop the 6/6 result is currently resting on.

skills/uipath-maestro-flow/SKILL.md
```diff
@@ -86,7 +86,7 @@ These rules apply across all three capabilities. Each capability index adds capa
...
 
-4. **Never invoke other skills automatically** — when a flow needs an RPA process, agent, or app, identify the gap and provide handoff instructions. Let the user decide when to switch skills.
+4. **Never invoke other skills automatically** — when a flow needs an RPA process, agent, or app, identify the gap and provide handoff instructions. Let the user decide when to switch skills. **One exception — IXP extraction with documents in hand:** when the flow needs document extraction, the user supplied sample documents, and `registry search "uipath.ixp"` shows no extractor covering them, invoke the `uipath-ixp` skill to build and deploy the model, then resume the flow ([plugins/ixp/impl.md — If the Model Does Not Exist Yet](references/author/references/plugins/ixp/impl.md#if-the-model-does-not-exist-yet)). Do NOT drive `uip ixp` project or deployment commands from this skill instead of invoking it — the sibling's guides carry guardrails this skill does not. If `uipath-ixp` is unavailable in the session, fall back to `core.logic.mock` plus an Open Questions entry, exactly as when no documents were supplied.
```
rockymadden: The carve-out itself reads well and the "do NOT drive `uip ixp` from here" clause is the right guardrail.

One gap: it assumes the sibling accepts the handoff. `uipath-ixp` currently instructs the opposite when the working context is a `.flow` (see my comment on `skills/uipath-ixp/SKILL.md`). Worth landing both halves in this PR so the contract is symmetric.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py
```diff
@@ -0,0 +1,743 @@
...
+            return 0
+
+    # Degraded acceptance: deployment exists, the flow landed the documented
+    # mock fallback, and the DeploymentName is recorded in the built artifacts.
+    has_mock = any(
+        str(node.get("type", "")) == MOCK_NODE_TYPE
+        for document in parsed.values()
+        for node in document.get("nodes", [])
+    )
+    if has_mock and deployment_names:
```
rockymadden, line 728: **Major — this branch is the only one CI has ever taken.**

Both runs cited as evidence in the PR body passed here, not through the wired branch:

```
run 32861035922 → OK (degraded)
run 32867830480 → OK (degraded)   score=1.000  duration=906s
```

So the headline outcome — flow wired to the extractor the agent just built — has not been observed green, and the 6.0-weight gate stays satisfiable by deploy + mock + breadcrumb after the registry latency clears. An agent that stops wiring the node still scores 1.000.

Fix: have `check` establish the precondition itself rather than trusting it — `registry pull --force` then `registry get <expected node type>` for this run's deployment, accept degraded **only** when the node genuinely is not served, require wiring otherwise. That keeps the tenant-weather immunity you built this for while making the fallback conditional instead of permanent.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py
```diff
@@ -0,0 +1,743 @@
...
+        else:
+            print(
+                f"IGNORING unresolvable match '{node_type}' "
+                f"(`registry get` exited {completed.returncode}) — stale deployment "
+                "residue, not a usable extractor."
+            )
+    return covered, len(nodes), len(matched)
+
+
+def require_domain_uncovered() -> None:
```
rockymadden: **Major — destructive, and unattributed.**

`resolvable_covering_nodes` matches `DOMAIN_MARKERS` (`falcon`, `raptor`, `mews`, …) as substrings against every IXP registry node's DisplayName/NodeType, then this function deletes the Orchestrator folder behind each match — recursively, with everything in it. In `pre_run`, before the agent runs.

Compare teardown, which is careful: created-after-seed (Id above watermark) **and** carrying this run's deployment. Self-heal has neither guard. Point this task at a developer's own tenant or the wrong org and it deletes a real folder because a model name happened to contain `mews`.

Fix: require residue evidence before deleting — the node's backing project absent from `uip ixp projects list` is the exact signature you describe in the docstring ("teardown deleted the project and could not delete the folder"), and it is one call. Failing that, gate the whole self-heal behind an env opt-in so only CI arms it.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,148 @@
...
+run_limits:
+  max_turns: 100
+  turn_timeout: 1500
+  task_timeout: 3000
+
+# Hook and criterion timeouts must exceed handoff.UIP_TIMEOUT_SECONDS (120), or
+# the harness kills the wrapper mid-`uip` call and the failure reads as a skill
+# defect.
+pre_run:
+  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py seed"
```
rockymadden, line 73: **Minor — 180s does not cover the self-heal path.**

The comment above reasons only about `UIP_TIMEOUT_SECONDS` (120), but the sleeps dominate: folder delete is 4 attempts × 5s, then up to 3 rechecks × 15s, each recheck re-running `registry pull --force` + search + `registry get`. Worst case is ~240s+, past this timeout.

Consequence is the misread this task works hard to avoid everywhere else: the hook is killed, no snapshot is written, and `check` fails with "`.grader_baseline.json` missing" — which reads as a skill defect. Raise to ~420s, or drop `DOMAIN_RECHECK_ATTEMPTS` to 2.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,148 @@
...
+      path: documents
+      mount_point: documents
+
+# The whole chain runs in ONE agent turn; its long poles are `projects create`
+# (~50s) and registry propagation after deploy (measured up to ~8 minutes).
+# 1500 leaves room to build the flow after a slow propagation.
+run_limits:
+  max_turns: 100
+  turn_timeout: 1500
+  task_timeout: 3000
```
rockymadden, line 64: **Minor — inert, and the harness says so.** From run 32867830480:

```
[WARNING] run_limits.task_timeout (3000s) exceeds run_limits.turn_timeout (1500s).
A larger task_timeout cannot extend the agent's single iteration; the agent budget is turn_timeout.
```

Drop it, or set it to something the harness will honour.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,148 @@
...
+    expected_skill: "uipath-ixp"
+    weight: 3.0
+    pass_threshold: 1.0
+
+  # Advisory (pass_threshold 0), NOT gates: the primary gate already proves both
+  # of these happened — a new project cannot exist without `projects create`, and
+  # a deployment cannot exist without `deployments create`. They are kept only to
+  # separate "never attempted" from "attempted and errored" when the outcome
+  # check fails, which is the difference between a skill-routing defect and a
+  # CLI/tenant problem. Weight 1.0 keeps them from moving the score.
```
rockymadden: **Minor — they do move the score.**

Advisory criteria always pass, so each adds its weight to both numerator and denominator. Your own data point proves it: 0.613 is exactly 9.5/15.5 — the two advisories counted as passing in that run. Without them the same run is 7.5/13.5 = 0.556.

Net effect is a 2.0/15.5 = 0.129 floor under every failing run, which matters for evalboard aggregates. Either fix the comment to say what they actually do, or drop the weight if the harness accepts 0.

rockymadden (commented):
Reviewed the skill changes, the task YAML and the grader. Ran the grader unit suite locally: **42 pass** (235 across the whole `uipath-maestro-flow` suite) — the PR body says 37, stale.

The grader design is the strong part: digest-based snapshot diff so the sandbox leaks no answer, per-project candidate scoping, read-deployments-before-delete ordering, folder-Id watermark, and a teardown that restores its own precondition instead of burning the fixture domain.

Three things I'd want addressed before merge, detailed inline:

1. **`uipath-ixp` still tells the agent to bounce back.** `skills/uipath-ixp/SKILL.md:20-28` (untouched by this PR) says *"STOP and invoke `uipath-maestro-flow` … The working context is a `.flow` file or a Maestro flow rather than a standalone IXP project. … This overrides Critical Rule 1."* The new Rule 4 carve-out hands off in exactly that state, so only one half of the mutual-exclusion contract moved — the same coin-flip that Rule 4 was.
2. **The wired path has never gone green in CI.** Both runs cited as evidence passed via the fallback, not the behaviour under test.
3. **`seed`'s self-heal deletes Orchestrator folders on a substring match**, with none of the attribution guards teardown has.

Plus smaller notes inline on the `pre_run` timeout budget, the inert `task_timeout`, the advisory-weight arithmetic, the missing `documents/generate.py`, and a cleanup path that one malformed `.flow` can abort.

Two things outside the diff worth a line in the PR body:

- **`--group` semantics.** Rule 13 + cli-reference flip from "FULL label path (`Invoice > Line Items`)" to "verbatim `Name` (`Line Items`)". Unrelated to RE-12982, but it resolves a real contradiction — `label-documents-guide.md` already used the leaf form throughout — so keep it, just name it so the ixp owners see what they approved.
- **Failing check.** `Run skill smoke tests` is at 82.7%, 13 failures, 12 of them unrelated flow tasks — reads ambient. But `skill-ixp-deploy-folder-by-name-smoke` scored 0.812 in a suite this PR touches. Worth a re-run; if it repeats, check whether the new `--output-filter "[?Path=='Shared'].Key"` example steers agents at a field that smoke's recorded payload (`DisplayName`/`FullyQualifiedName`) never serves.

**Style, and it is a real maintenance cost:** 110 em-dashes across ~1950 added lines, a single 234-word bullet in `impl.md`, and run-log archaeology — GH run IDs, dated latency measurements ("measured 2026-08-24", "~25s healthy, hours observed"), "environment weather" — embedded in guidance the agent reads at runtime and in grader docstrings. Compress to the rule plus one link; the forensics are already in this PR description, which is where they belong.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,148 @@
...
+  publish` is not required), then resume the flow and wire the resulting node.
+
+  The prompt never says "IXP", "project", "publish" or "deploy": recognising the
+  prerequisite is the behaviour under test. Grading is scoped to this run's
+  artifacts via a tenant snapshot-diff; handoff.py (seed / check / teardown)
+  holds the full design rationale — read it before treating a failure as a
+  skill regression. Key properties: the scenario only measures the redirect
+  while no published extractor covers the fixture domain (currently falconry
+  licences; documents/README.md records how four prior domains were burned, and
+  documents/generate.py rotates); the prompt steers the deployment into a fresh
```
rockymadden: **Minor — `documents/generate.py` is not in this PR.**

`documents/README.md` also tells the next maintainer to "regenerate matching fixtures" when the domain burns, with no generator committed — and the rendering recipe (Pillow, 1000×1350, twelve cells) is prose, not code. Commit the generator or drop both references.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
```diff
@@ -0,0 +1,148 @@
...
+    timeout: 300
+
+initial_prompt: |
+  I've got a batch of scanned falconry licences in ./documents/ — have a look
+  at them. Build me a UiPath Maestro Flow that reads licences like these and
+  pulls out the licence number, ring number, species, falconer and the date the
+  licence expires, then logs the ring number and expiry.
+
+  Trigger it manually. Set up a fresh Orchestrator folder for anything this
+  needs deployed and use that folder — don't put anything in the existing ones.
```
rockymadden: **Minor — this is a partial hint.** The PR body says the prompt never says "deploy"; it says "for anything this **needs deployed**". Worth rewording to "anything this needs set up" so the claim holds literally — the fresh-folder steer is load-bearing for teardown, the word "deployed" isn't.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py
```diff
@@ -0,0 +1,743 @@
...
+        print(f"OK: deleted IXP project '{project_name}'")
+    else:
+        detail = (completed.stdout or completed.stderr).strip()
+        print(
+            f"WARN: could not delete '{project_name}' "
+            f"(exit {completed.returncode}): {detail}"
+        )
+
+
+def wired_ixp_references() -> str:
```
rockymadden, line 818: **Minor — one malformed `.flow` aborts all cleanup.**

Bare `json.load` per file, and `teardown_main` catches `BaseException` — so a single unparseable flow (a plausible way for a run to fail) turns the whole teardown into "cleanup failed", leaking project + folder + deployment and burning the fixture domain. That's the failure this design exists to prevent.

Wrap the per-file parse in try/except and keep going; the docstring already claims this helper is "deliberately tolerant".

skills/uipath-ixp/SKILL.md
joe-prosser: Fixed in a73ba4188 — *When NOT to Use* now carries the inbound-handoff exception (documents supplied + no covering extractor → project + folder-deploy work happens here, hand back, never wire the flow). One deliberate deviation from your suggested fix: I left the frontmatter `description` untouched. The handoff arrives as an explicit skill invocation, which doesn't go through description matching, and the description's DO-NOT-TRIGGER clause is scoped to listing/wiring, which remains correct — while any description edit triggers the activation-gate recall eval. If you still want the description half, happy to take it as a follow-up with the recall run.

skills/uipath-maestro-flow/SKILL.md
joe-prosser: Both halves are in this PR now — see a73ba4188 for the uipath-ixp side.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py
joe-prosser: Two answers. First, the evidence in the PR body was stale (like the test count): the wired branch has gone green repeatedly — 32968685992 first, then 32997068855, 33056705206, 33059578513, each naming the wired deployment in the check output. Second — your structural point stood regardless, and it's fixed exactly as you suggested in a73ba4188: `check` now runs its own registry probe (pull → search → `registry get`) and rejects the mock whenever this run's node is actually served; matching is sanitization-aware so underscore-carrying names can't slip past. First green under the enforced gate: 33162961197 (1.000, wired). One consequence worth naming: the probe runs at grading time, so a node that surfaces between the agent's documented give-up and grading now fails that run — accepted; it narrows the weather window from hours to minutes and keeps the gate honest.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py
joe-prosser: Fixed in a73ba4188 with three guards, the first being your suggestion: (1) residue signature — refuse unless no live project carries a domain marker (backing project gone = the leak signature); (2) live-project cross-check against the node identity, sanitization-aware, for the case where a live extractor's DisplayName carries a marker its project name doesn't; (3) positive folder identification via `folders get`, refusing protected roots (`Shared`/`Default` at any path depth) and anything unreadable. Live validation: self-heal ran twice today — 33164712588's seed healed the previous run's leaked folder mid-seed and proceeded to a clean probe.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
joe-prosser: Your ~420 suggestion turns out to be impossible — TaskDefinition schema-caps hook timeouts at 300 (found the hard way: 33162270721 skipped the task at load with `pre_run.0.timeout <= 300`). Took your alternative instead: capped at 300 and dropped `DOMAIN_RECHECK_ATTEMPTS` to 2 so the self-heal worst case fits (aca1f3264).

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
joe-prosser: That harness warning misled us both — I dropped it per your first option and the next run inherited the experiment's 1200s whole-task cap, which is below `turn_timeout` alone (the same trap pdd_to_sdd.yaml documents). It caps end-to-end wall clock, so it's load-bearing: healthy runs here measure 780–1550s. Restored to 3000 with an accurate comment (f68f55caf).

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
joe-prosser: You're right, the comment was false. Corrected in e36c02c34 to state the real effect (a 2.0/15.5 ≈ 0.13 floor under failing runs — not partial credit). Kept weight 1.0 rather than testing the harness's tolerance for 0.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
joe-prosser: Dropped both references (e36c02c34) — the generator was deliberately not committed; the README now names its Rendering section as the fixture spec.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml
joe-prosser: Took your wording — the prompt now says "Create a fresh Orchestrator folder for anything this needs set up" — and dropped "deploy" from the description's never-says claim anyway, since the debug/deploy guardrail bullet still carries the word (e36c02c34). Green post-change: 33162961197.

tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/handoff.py
joe-prosser: Fixed in e36c02c34 — per-file try/except; cleanup proceeds and prints which file it skipped.

joe-prosser:
Addressed the full review across a73ba4188 (majors), e36c02c34 (minors), f68f55caf + aca1f3264 (hardening found while validating — task_timeout is load-bearing against the experiment default, and hook timeouts are schema-capped at 300). Validated live: [33162961197](https://github.com/UiPath/skills/actions/runs/33162961197) passes 1.000 with the conditional gate armed and the real node wired; the self-heal attribution guards ran twice on real leaks. PR body refreshed (test count, run links — the ones you flagged as stale).

rockymadden (commented):
## What it does

Adds one e2e task asserting that a Flow-context agent, handed documents and no IXP extractor, recognises the missing prerequisite and hands off to `uipath-ixp` instead of mocking.

**Test rig** (`tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/`)
- 3 synthetic falconry-licence PNGs, a 154-line task YAML, an 891-line `seed`/`check`/`teardown` grader, 42 pytest unit tests.
- Attribution is a tenant snapshot-diff: SHA-256 project-name digests plus a folder-Id watermark from a create-and-delete sentinel folder. Rationale is sound: the prompt can't say "record what you created" without leaking the answer.
- `check` passes on a wired `uipath.ixp.*` node, **or** on a `core.logic.mock` + `DeploymentName` breadcrumb while its own registry probe confirms the node isn't served.
- Teardown deletes the run-scoped Orchestrator folder, which removes the deployment and its registry node. That's the self-cleaning claim, and it's the right idea: 4 fixture domains were burned before it existed.

**Skill changes the test forced**
- `uipath-maestro-flow` Rule 4 gains an IXP carve-out (auto-invoke `uipath-ixp`).
- `uipath-ixp` gains `references/deployment-guide.md`, a task-nav entry, an inbound-handoff exception, and permission to skip labelling.
- Both ixp plugin docs branch on "documents supplied?" instead of always mocking.

## ✅ Verified good

- Cross-skill relative links resolve. The 6× `../` in `impl.md` lands on `skills/`; all anchors match real headings.
- Fixtures are genuinely synthetic. Rendered one: invented values, footer disclaimer, no PII, high-contrast so OCR isn't a variable.
- The 42 tests are collected by existing CI (`.github/workflows/test-helpers.yml` → `pytest tests/tasks/uipath-maestro-flow/`). No wiring change needed.
- Advisory `pass_threshold: 0` criteria match repo convention (72 existing uses). Not a scoring smell.
- Fail-fast on indexed dict access, read-deployments-before-delete ordering with a test locking it in, watermark correctly shields `Shared`/`Default` in teardown. Good instincts.

---

## 🔴 Raise with the author

### 1. Teardown can delete a concurrent task's IXP project — MAJOR

`cleanup()` falls back to "the only new project is ours":

```python
if not attributable and len(created_projects) == 1:
    attributable = list(created_projects)
```

`tests/tasks/uipath-ixp/e2e/full_lifecycle.yaml` is tagged `e2e` and creates an IXP project. `make e2e` runs both tasks in one invocation, and `run-coder-eval.yml:411` defaults `-j` to **4**. If this run fails to create a project while `full_lifecycle`'s appears in the diff, that project is the lone new one, isn't wired, and gets deleted mid-run.

`test_teardown_deletes_a_lone_new_project_even_when_nothing_was_wired` encodes this as intended behaviour, so it won't be caught.

Worth checking whether this explains the "unrelated sibling timeout" in the 29/30 sibling result.

**Fix:** see #3 — a deterministic run-scoped folder makes the fallback unnecessary for deployed projects, and an undeployed leftover should be reported as `LEAKED`, not guessed at.

### 2. Seed self-heal deletes folders inferred from a parsed GUID — MAJOR

`require_domain_uncovered` extracts the trailing 36 chars of a registry `NodeType` and deletes that Orchestrator folder. Deleting a folder takes everything in it (processes, assets, queues, triggers).

Guards at that point are name-only, because seed has no watermark yet:

```python
PROTECTED_FOLDER_NAMES = {"shared", "default"}
```

A pre-run hook that deletes a folder identified only by a string parsed out of a node type is the most dangerous thing in the diff. It's contained to the CI org, but a folder named anything other than `shared`/`default` is fair game.

There's also a bricking path: if an agent ignores the "fresh folder" steer and deploys into `Shared`, teardown leaks the deployment, and the next seed refuses to delete `Shared` and hard-fails. The task then needs a hand-delete or a domain rotation, permanently.

### 3. The repo already has a convention for this — MAJOR (design)

Two existing patterns:
- `tests/tasks/uipath-platform/cleanup.py` — per-task `uuid8` embedded in resource names, docstring says *"Per-task uuid8 makes this parallel-safe — siblings don't touch each other's resources."*
- `tests/tasks/uipath-ixp/_shared/cleanup_project.py` — agent writes `report.json`, cleanup deletes exactly that project.

This PR invents a third mechanism (digest diff + folder-Id watermark + NodeType GUID parse). The stated reason only rules out the second: a "record what you created" instruction leaks the answer. It does not rule out the first.

**Proposal:** have `seed` create the run folder itself, named `e2e-flow-ixp-<uuid8>`, and change the prompt from *"Create a fresh Orchestrator folder"* to *"use the Orchestrator folder `e2e-flow-ixp-<uuid8>` for anything this needs set up"*. That is no more leaky than the current steer (arguably less), and it:
- deletes the GUID-parse path — self-heal and teardown only ever touch folders matching the prefix (fixes #2)
- makes deployed-project attribution exact via `FolderKey`, so the lone-project guess goes away (fixes #1)
- removes the sentinel folder, the watermark, and `folder_created_after_seed` entirely
- guarantees the folder is deletable, so the `Shared` bricking path closes

### 4. `uipath-ixp` will stall on the folder question in CI — MAJOR

New `deployment-guide.md`:

> Ask which folder when none was identified

`uipath-maestro-flow` Rule 5 has a non-interactive fallback. `uipath-ixp` has none — `grep -n "non-interactive\|headless\|cannot be asked" skills/uipath-ixp/SKILL.md` returns nothing. A headless handoff that reaches the folder question has no documented exit.

The task only passes today because the *prompt* pre-answers it. Any other caller hits the stall. Needs either a non-interactive rule in `uipath-ixp/SKILL.md` or an explicit "the handing-off skill must supply the folder key" contract in the inbound-handoff exception.

### 5. Skip-labelling licence is wider than the PR's scope — MAJOR

`project-setup-guide.md`:

> **The request is a model for an automation to call** — a flow, activity pack, or other runtime consumer, with no score, metrics, or accuracy target named.

Almost every real flow request matches that. Labelling becomes opt-out by default across the whole `uipath-ixp` skill, and the skill's own Critical Rule 15 is built on scores meaning something. An unscored model silently shipped into a production flow is a worse outcome than a slow one.

The e2e needs this only for the inbound-handoff path. Scope it there rather than to "any automation consumer".

### 6. Rule 4 carve-out auto-creates tenant resources without consent — MAJOR (design)

Rule 4 exists so the agent doesn't switch skills unasked; Rule 5 exists so decisions go to the user. The carve-out has the flow skill invoke `uipath-ixp`, which creates an IXP project **and** a folder deployment — side-effecting tenant writes — with no confirmation gate. The PR body is candid that Rule 4 made the redirect a coin flip, so the change is justified; the missing piece is whether the *tenant writes* should still be confirmed in an interactive session. Worth a deliberate answer, not an inherited one.

---

## 🟡 Minor

**7. `check_main` matches raw, `run_node_in_registry` matches sanitized.** `handoff.py` has `sanitize_registry_segment` and uses it in the registry probe, but the wired-match is raw substring:

```python
wired = next((name for name in names if any(name in found for found in identifiers)), None)
```

A `DeploymentName` carrying `_` or uppercase won't substring-match its own lowercased-and-hyphenated `NodeType`. It survives today only via `inputs.modelName`. When it doesn't, the run prints `PROPAGATION SIGNATURE` — a flaky red blamed on the skill, the exact failure mode this grader is built to prevent. Reuse the same sanitize + casefold comparison in both places.

**8. Hook timeout budget doesn't hold at worst case.** `UIP_TIMEOUT_SECONDS = 120` against a 300s hard cap, with `seed` making up to ~8 sequential `uip` calls (`--help`, `pull`, `search`, N× `get`, `folders create`, 4× `delete` with sleeps, `projects list`). The YAML says the self-heal path *"is budgeted to fit under it"* — true at nominal latency only. Two hung calls exceed the cap, and the timeout reads as a skill defect. Drop `UIP_TIMEOUT_SECONDS` to ~45.

**9. Propagation numbers disagree across four files.** `~25s` (impl.md, module docstring), `~22s` (PROPAGATION SIGNATURE message), `up to ~8 minutes` (YAML comment), `hours` (elsewhere). Pick one measurement, state it once, link to it.

**10. `impl.md`'s "Yes" bullet is ~350 words in a single paragraph.** It carries eight separate instructions: invoke, rationale, bash block, propagation weather, retry policy, breadcrumb with example, runtime-lag aside, unavailability fallback. The retry policy (`pull --force`, then ~30s for at most 3 min, then stop) is the operative part and it's buried mid-paragraph. Split: short bullet + a subsection.

**11. Breadcrumb scan misses the solution root.** `project_files()` globs under `find_project_dir()`, so a `PLAN.md` written one level up at the solution root isn't seen. The `.flow` label instruction covers the main path; widening the glob one level removes a false negative for free.

**12. Two assertions don't assert.**
- `test_seed_accepts_an_uncovered_fixture_domain`: `"do not cover" in completed.stdout` — no such string exists in `handoff.py`. Dead branch of the `or`.
- `test_teardown_leaves_a_new_folder_owned_by_a_concurrent_run`: the `.replace(...)` is a no-op (the RUN key string can't contain the CONCURRENT key) and the `or` makes it near-tautological. Collapse to `assert f"deleted run-scoped folder {CONCURRENT_FOLDER_KEY}" not in completed.stdout`.

**13. No committed fixture generator.** `documents/README.md` gives a prose spec ("Pillow, 1000×1350 on white: dark header band…") and says *"there is no committed generator"*. Rotating the domain then means re-implementing from prose. Commit the ~20-line script.

**14. `--output-filter` needs `--limit` note is in the wrong skill.** It landed in `uipath-ixp/references/cli-reference.md` § Deployments but describes `uip or folders list`. Anyone hitting it from `uipath-platform` won't find it.

**15. The prompt steer softens the premise.** *"Create a fresh Orchestrator folder for anything this needs set up"* nudges the agent toward provisioning. It never says IXP, so the core measurement holds, but it's a thumb on the scale. #3's seed-created folder removes it.

---

## ⚠️ Documentation weight

Not slop, but heavily over-documented. `handoff.py` is 891 lines for three subcommands, with a 55-line module docstring; the YAML is roughly 40% comment.

The GH-run-ID citations (`32967816894`, `32968685992`) are the best thing in the file — real evidence, worth keeping. What isn't earning its space is the restated rationale: several comments explain a decision the adjacent docstring already made, and `describe`/`project_digest`/`run_uip` carry prose that adds nothing.

Two comments actively mislead and should be fixed rather than trimmed: the `~22s`/`~25s`/`~8min` disagreement (#9) and the "budgeted to fit under it" claim (#8).

---

## Verdict

The test design is genuinely good — snapshot-diff attribution, the enforced conditional gate with its own registry probe, `PROPAGATION SIGNATURE` labelling, and read-before-delete ordering all show someone who ran this against a real tenant and learned from it. The self-cleaning teardown is the right answer to burning fixture domains.

**Blocking:** #1 and #2 are destructive-on-a-shared-tenant. #3 fixes both with less code than is here now.

**Should settle before merge:** #4, #5, #6 — three skill-behaviour changes broader than one e2e's needs.

Everything in 🟡 is fine as follow-up.

joe-prosser:
Thanks Rocky — every finding is addressed; each is mapped to a commit below. The design went where #3 pointed, and it also caught one bug none of us had seen.

**Blocking**

- **#1 / #2 / #3** — `69f7a0043`. Seed creates the fixed-name run folder (`flow-e2e-e9fbf2d0`, the `cleanup.py` uuid8 convention) and the prompt names it. Project attribution is exact (wired into this sandbox's flow, or `FolderKey ==` the run folder); the lone-project fallback is gone and an unattributable project is reported `NOT DELETED`, never guessed at. Self-heal deletes only a folder with that exact name; the GUID-parse delete path, `PROTECTED_FOLDER_NAMES`, the sentinel and the watermark are all deleted. One correction to the proposal: `e2e-flow-ixp-<uuid8>` would have put "ixp" in the prompt, so the name is domain-neutral.

**Should settle before merge**

- **#4** — `b28bb2002`. A two-ended contract: the flow skill resolves the target folder before invoking (per its rule 5, whose non-interactive fallback already exists) and passes it; `uipath-ixp` treats it as the caller's contract, and the Deployment Guide gains the documented headless exit — stop and report the missing key, never guess or create one.
- **#5** — `b28bb2002`. The skip-labelling trigger is now "the model unblocks a larger build in this session", with the handoff as the canonical case — narrower than "any automation consumer", and caller-agnostic (caller identity isn't something an agent can observe).
- **#6** — `b28bb2002`. Deliberate no-gate, written into the carve-out: the writes fulfil the extraction request itself, and the one irreversible choice — which folder the deployment lands in — is exactly the decision rule 5 routes to the user. That sentence is also load-bearing: without it, rule 5's "destructive ops are never auto-answered" could read as a consent gate and stall a headless run.

**Minors folded in** — #7 one `name_matches` rule for the `.flow` match and the registry probe (and it fired: two green runs named the project `falconry_licences-…` with an underscore); #8 cap 120→45 with the real arithmetic replacing "budgeted to fit"; #9 one canonical propagation measurement; #12b and #15 fell out of the redesign.

**Found along the way** — the two earlier "green" runs (33162961197 / 33173123153) had each leaked a folder: teardown read a failing `folders get` as "already gone" and skipped the delete, which is what burned the falconry domain. Fixed in `f502960f4` (deletion judged by the delete's own result), residue removed by hand. Two follow-ons from the same class: the folder API has no timestamp, so the concurrency guard stamps its own into the folder Description at create (`bed16fe07`); and the CI tenant has >500 folders, so the leftover lookup is an exact-name `folders get`, not a list (`72660179e`). `1356d33dd` merges main across #2879's `references/` flattening (GitHub had reported the PR CONFLICTING since 08-28, so PR CI had silently stopped); `18025710c` takes an independent reviewer's three findings.

**Evidence** — green on the new design: 33536499571, 33548029622, and 33550441340 on the merged head (1.000, wired, stamped folder created and deleted). All three left the tenant exactly as found (no `LEAKED`, no `NOT DELETED`). Unit tests 42→51; Test Helpers green on `18025710c` (33550422547) — the first PR CI run since the CONFLICTING state began; CLI Verb Gate green too.

**Deliberately deferred (your "fine as follow-up")** — #10 impl.md paragraph split, #11 breadcrumb glob, #12a dead assertion branch, #13 committed fixture generator, #14 `--limit` note in uipath-platform, doc-weight trim; plus the wider "published"→"deployed" vocabulary pass in the ixp plugin docs (only the two causal claims were corrected here).

**Supersedes two of my 08-28 replies** — the one on `handoff.py:334` (the three self-heal guards: all three are deleted, replaced by the exact-name lookup) and the one on the YAML prompt (`"Create a fresh Orchestrator folder"` — now the seed-created folder is named instead, and the run I cited there, 33162961197, later turned out to have leaked).

rockymadden (approved):
Thank you!

PR #2813, opened 2026-08-25
fix(scorecard): count maturity carry-forward tasks in the Confluence scorecard

bai-uipath:
The Coding-Agents scorecard has been under-reporting every product row. Bianca spotted it on Licensing, which the card showed as 3/3 when 12 licensing tasks exist — but the cause is not licensing-specific and it moves the headline number too.

On a `--skip-mature` night the nightly runner does not re-run a task that has passed five runs in a row; it carries the task forward as a passing row so the denominator stays at the full authored suite size. The scorecard was discarding all of those rows, because they are the one kind of row that does not carry a variant id and the card scores a single variant. On the two most recent affected nightlies that silently removed 45% and 71% of the run before anything was counted.

The published pass rate therefore disagreed with evalboard for the same run — 90.3% vs 97.2% on 2026-08-21 — because evalboard counts carried rows and the card did not. This change makes the card agree.

The two columns are kept on different bases, because they mean different things. Tests Pass/Fail and the tier splits count a carried task as a pass on the full known basis, which is the org card's own definition of the column. Every eval score and every mean stays executed-only, because a carried row's score is a uniform synthetic 1.0 rather than a measurement, and averaging those in inflated the overall mean by up to 24.6pp.

It also makes the carried share visible rather than implicit. A product row that rests entirely on carried passes has no fresh signal that night, and the card should say so instead of presenting it as a green measurement.

The guard is permanent, not a stopgap: every run.json already in blob storage lacks the field, so any historical run scored through this command needs it regardless of what the runner writes going forward.

## Testing

No CI covers `.claude/commands/`, so this was verified by extracting the command's Phase 1 script from both `origin/main` and this branch and running each against the same real nightly `run.json`, pulled from the `coderevaltests` / `runs` blob container.

**Affected night (`2026-08-21_04-55-29`), where maturity skipping is active:**

| | `origin/main` | this branch |
|---|---|---|
| licensing | 3/3 | **12/12** (9 carried) |
| uipath-platform | 31/39 | **93/101** (62 carried) |
| headline pass rate | 318/352, 90.3% | **1174/1208, 97.2%** |
| overall mean | 94.5% | 94.5%, unchanged |
| per-skill eval cells differing from `main` | | **0 of 32** |

**Regression check, full run with zero mature skips (`2026-08-24_04-15-29`):** every number is unchanged, and 0 of 35 skills differ. On nights where maturity is not active this change is inert.

**Evalboard parity:** evalboard counts every `task_results` row with no variant filter, which computes to 1174/1208, 1162/1220 and 813/1229 across the three runs checked. This branch reproduces all three exactly; `origin/main` matches none of them.

The scope is deliberately narrow: **only Tests Pass/Fail and the tier splits move.** Every Build / Operate / Troubleshoot cell and every mean is byte-identical to what `main` produces today, on both runs.

One thing worth a reviewer's attention: the current script already prints `dropped variant None: 856 distinct tasks` on an affected run, so it detected the discard and announced it. Nuance 5 attributes the disjoint-variant case to an activation/second-phase set, which made the warning read as expected behavior rather than data loss. That is how this went unnoticed.

For context, the manually published 2026-08-24 card already reports both bases by hand (`2/2 exec · 12/12 known` for Licensing) and already keeps its means executed-only. This change encodes that existing practice into the command so it does not have to be redone each week.


uipreliga (approved):
🚢

PR #2815, opened 2026-08-25
fix(uipath-maestro-case): repair connector schema keys from their contract authority

song-zhao-25:
## The bug

`uip maestro case spec --output json` applies a PascalCase **naming policy** to every object key, including the connector-contract field names inside the request and response schemas. The documented repair — lower the first character of every key — is lossless only for camelCase contract names:

| contract name | serialized as | lowering gives | correct? |
|---|---|---|---|
| `conversationId` | `ConversationId` | `conversationId` | ✅ |
| `request_body` | `RequestBody` | `requestBody` | ❌ underscore destroyed |
| `Title` | `Title` | `title` | ❌ contract name is capital-first |

The old text claimed *"lowering restores the connector-contract names"* — true for its own `toRecipients` example, false for every snake_case field and every capital-first one. The result passes `uip maestro case validate` and dies at runtime with `RequestBody not found, did you mean request_body`.

Found by a live coder-eval run of `golden_rebuild/contract_execution` whose field-name criterion the documented procedure could not satisfy.

## The fix — one bullet on the existing rule

`§ Normalize key casing` keeps its place, its heading, and its anchor. One false clause is deleted and one bullet added, because the two rules already there — lower the first character, verify no capital-first key remains — are correct for contract keys once the restore overwrites them immediately after.

The new bullet carries only what an agent cannot derive:

- **Three authorities**, none previously named anywhere in the skill: `.Data.Outputs.ResponseFields[].Name` for the response, `.Data.Inputs.{BodyFields,QueryParameters,PathParameters,EventParameters}[].Name` for inputs (also present verbatim in `tasks.md` `input-values`, the copy that always persists), and the targeting `$ref` value for a `definitions` key.
- **Each `Name` is a dotted leaf path** — walk its segments through `properties` / `items` / `$ref`, each segment being the byte-exact key at that level; top-level keys are the de-duplicated first segments. A segment may carry an `[*]` array marker.
- **Match by name, never by index.** The collections differ in shape and length: Jira returns `Fields, Id, Key` against `fields.issuetype.name, id, key`, so index pairing writes a literal `"fields.issuetype.name":` key.
- **Skip the `Error` output** — `ResponseFields` does not describe its fixed envelope.
- **Written → authority only.** An authority name with no written key is an unused optional parameter, not a mismatch (List Emails writes 3 of 9 query parameters).
- Unmatched, or a cache truncated to `CaseShape` alone → `build-issues.md`; never invent a name.

Consumers routed: `connector-activity`, `connector-trigger`, the `event` trigger (which now restores names before writing its sidecar, since it owns the only copy every downstream consumer reads), and — via the shared § Procedure — all four condition scopes. The motivating run carries the damage in a **stage-entry** rule.

## Verified end to end on a live build

A full `cm_golden_expense` build on this branch produced:

```
caseplan.json: 100 schema property keys, capital-first: NONE
HTTP Webhook  -> request_body, request_headers          (the original defect, fixed)
Outlook       -> 39 top-level keys byte-exact, toRecipients[*] intact,
                 17/17 $ref/definitions pairs resolving, 0 dangling
```

For contrast, a pre-existing artifact in `tests/runs` has **29 dangling `$ref`s** (`Blocks[*]` as a definitions key against a `$ref` to `blocks[*]`). Spec-caches on this branch also persist `Inputs` and `Outputs` alongside `CaseShape`, where older caches kept only `CaseShape` and so destroyed the authority.

## New: a dedicated naming test

`tests/tasks/uipath-maestro-case/connector_naming/` — one stage, two connector tasks, built and validated but **never run**.

The pair spans both ways the rename can be mis-repaired. **Slack "Send Message to Channel"** is snake_case (`send_as`, `link_names`, `app_id`, `thread_ts`, `response_metadata`), which first-character lowering destroys. **Outlook "Calendar Event Created"** is capital-first (`Title`, `ID`, `CalendarID`, `Attachments[*].MIMEType`), which a "no capital-first key may remain" scan destroys. Between them the case also covers array markers, three-level nesting, `$ref`/definitions, a 102-leaf → 13-property collapse, and contract names that collide with JSON-Schema keywords (`Title`, `Type`, `Name`).

Neither task needs a resource lookup: the trigger declares no event parameters, and `send_as` / `channel` take values the SDD gives by name so the skill resolves them. The trigger does carry a three-clause `jmes` filter, which puts the same names through a third sink — they compile into a JMESPath expression where a rename silently matches nothing at runtime.

`check_connector_naming.py` derives every expectation from a live `case spec` at grade time rather than hardcoding names, so a connector version bump cannot leave a stale golden passing. It is validated three ways: it passes on a solution built by hand in the Studio Web canvas, passes on an agent build, and fails on each of three injected defects (snake lowered, camel over-lowered, a definitions key renamed until its `$ref` dangles).

The canvas build also confirmed the derivation independently — its 13 Slack and 14 calendar top-level property names match the rule exactly, and its `Error` envelope is lowercased, which is the skip rule.

## Why it went unnoticed

Across every built caseplan on one dev machine: **2623** camelCase top-level keys (lowering recovers) vs **101** snake_case (it does not). The rule is right ~96% of the time.

## One limit, documented rather than papered over

**Persist the WHOLE spec response.** `Outputs.ResponseFields` and `Inputs.*` sit outside `CaseShape`; a cache holding only `CaseShape` has already discarded the contract names. Observed in two real builds in `tests/runs`.

An earlier revision called nested `definitions` names unrepairable and flagged the nested damage as *inferred*. Both are settled: the damage is observed in a built artifact, and every one of 43 `definitions` keys examined resolves to a unique `$ref` value, which is the byte-exact name.

## Scope

Five reference files plus three test files. An earlier revision of this branch also edited `io-binding`, `global-vars`, `phased-execution`, and `implementation.md`; those are reverted — they do not own connector field names. The doc-level gaps there are real but pre-date this PR and stay tracked as known gaps: io-binding's walker errors-and-skips with no repair awareness, global-vars teaches serializer-damaged names as the contract, the in-session spec-call memo has no carve-out, and Check 12 has no sub-check comparing on-disk property names.

`npm run skills:validate` passes (26 skills, default + studioweb). Links resolve except seven already broken on `main`. No conflicts with current `main`.


song-zhao-25:
Superseded by #2838.

Same defect, rebuilt off current `main`. This branch churned through 21 commits and 13 rewrites of `§ Normalize key casing`, mostly because I kept restructuring prose against review opinions before running anything executable. The run that finally completed found two defects no review had caught.

#2838 keeps what was verified — the test (validated against a Studio Web canvas build, an agent build, and five injected defects) and the empirical findings — and replaces the guidance with one 9-line exception: schema property keys are never spliced from the response, so there is no damage to repair. That removes the restore step, the positional-region rules, the `Error` exception, and the keyword/contract collision reasoning, none of which have anything left to be about.

PR #2816, opened 2026-08-25
port(planner,case): Node ports of audit_sdd.py and audit_plan.py

roberts-cliff:
Rewritten to be **only the port**. Every added check has been removed; this changes the runtime and nothing else.

## Why

`python3` is not a runtime these skills can assume. On Windows it is commonly a Store alias stub that exits non-zero, so the auditors the docs mandate as gates silently degrade into "skip it". Node is guaranteed — `uip` is an npm-installed Node CLI, so if `uip` runs, these run.

## Two commits

1. **`7275ce921`** — add `audit-case-sdd.mjs` and `audit-plan.mjs`. Python left in place.
2. **`35ca32bb3`** — cut the docs over, delete the Python.

## Parity

Differential run against the Python originals comparing **full stdout+stderr, no exclusions**:

```
audit-case-sdd.mjs  vs audit_sdd.py     9 sdd.md fixtures + --draft   identical
audit-plan.mjs      vs audit_plan.py    6 tasks.md fixtures            identical

PARITY: PASS=15 FAIL=0
```

The earlier version of this PR could only claim parity with carve-outs, because it carried behaviour changes. This one has none, so the comparison is unfiltered.

## What was removed since your review

@AlvinStanescu — everything you objected to as an addition is gone:

- the widened `\bvars\.` lineage anchor — reverted to the Python's `=vars\.`
- the entire SLA Response Map check (section presence, column set, legal tokens, two-way closure, scope↔target)
- the `case-sdd-template.md` wording that described those checks

Those were real findings, but they are new enforcement and they belong in their own PR where they can be argued on their merits. Nothing in this branch validates anything the Python did not already validate, identically.

## Also removed: the availability escape hatches

```
planning.md              "If `python3` is unavailable, re-check ... manually"
case-design-lane-guide   "All three of python3 / python / py unavailable -> verify by hand"
case-sdd-template.md     "<py> = the first of python3, python, py that runs"
```

These existed only because `python3` could not be relied on. Under Node they describe a condition that no longer gates anything, and leaving them tells an agent to hand-verify on a false premise.

## Rename

`audit_sdd.py` -> `audit-case-sdd.mjs`, per your inline comment. It validates the Case Management SDD template specifically and sits beside six other SDD templates it knows nothing about.

## Deliberately NOT changed

The guard's exemption still scopes to `uipath-maestro-case/scripts/`, so an agent running the **planner's** auditor is flagged by `skill-case-artifact-safety-guard`. I verified this is **pre-existing, not port fallout** — the guard on `origin/main` flags `python3 <planner>/scripts/audit_sdd.py` in exactly the same way. It is a real hole, and it needs its own change with its own argument rather than riding along here.

## Review notes

Three translation details worth a look:

- `sectionBody()` emulates Python's `(?=^...|\Z)`. JS `$` under `/m` is end-of-**line**, so a naive port truncates every section at its first newline.
- `pyRepr` / `pyTupleRepr` / `stripChars` / `pySorted` reproduce Python's `repr()`, tuple repr, `str.strip(chars)` and codepoint sort — finding text is compared verbatim.
- Model facts load from `case-design-layers-guide.md` via `import.meta.url`, matching the Python's `__file__`-relative read.


tests/tasks/uipath-maestro-case/guardrails/artifact_safety_guard.yaml
```diff
@@ -197,9 +197,15 @@ success_criteria:
     tool_name: "Bash"
     # The script path must be the token immediately after the interpreter: agents wrap every
     # command in `/bin/bash -lc '...'`, so `bash\s+[^\n]*\.sh` would match any command that merely
-    # mentions a script path (e.g. `rg ... audit_plan.py`). The lookahead exempts the skill's own
-    # read-only auditor, which planning.md mandates as the final plan gate.
-    command_pattern: '([<>]{1,2}\s*\S*(caseplan\.json|tasks\.md|bindings_v2\.json|entry-points\.json)|tee\b[^\n]*(caseplan\.json|tasks\.md)|\b(node|nodejs|python[0-9]?|bash|sh)\s+(?!\S*uipath-maestro-case/scripts/)\S+\.(js|mjs|cjs|py|sh)\b)'
+    # mentions a script path (e.g. `rg ... audit-plan.mjs`). The lookahead exempts a skill's own
+    # read-only auditors, which the skill docs mandate as gates.
+    #
+    # The exemption is `uipath-<skill>/scripts/`, not `uipath-maestro-case/scripts/`: a case build
```
abhiram-vad: Let's not mention uipath planner scripts here. We can also clean up mentions of audit_sdd.py to avoid confusion.

skills/uipath-planner/scripts/audit-sdd.mjs
```diff
@@ -0,0 +1,1227 @@
...
+ *
+ * Usage:
+ *     node audit-sdd.mjs <sdd.md> [--draft <sdd.draft.md>]
+ *
+ * Read-only. Exit 0 = shape-clean. Exit 1 = numbered findings on stderr; repair
+ * the document with Write/Edit and re-run until clean. `--draft` additionally
+ * verifies the finalized document preserves the draft's ordered stage/task
+ * inventory and every draft `=js:` expression.
+ *
+ * Node port of audit_sdd.py — same CLI surface, same finding text, same exit
```
abhiram-vad: let's avoid mention of audit_sdd.py

abhiram-vad (commented):
lgtm overall. Just posting minor comments.

tests/tasks/uipath-maestro-case/guardrails/artifact_safety_guard.yaml
roberts-cliff: Done in 4a277218e. The comment no longer names the planner or its scripts — it now reads:

> The exemption is `uipath-<skill>/scripts/`, not one hard-coded skill: a single conversation can legitimately run more than one skill's read-only auditor, and each skill's docs mandate its own as a gate. Scoping the exemption to one skill penalised the agent for running a gate another skill requires. Pre-existing hole, not fallout from the Node port.

Same rationale, no cross-skill reference and no `audit_sdd.py`. The regex itself is unchanged.

skills/uipath-planner/scripts/audit-sdd.mjs
roberts-cliff: Done in 4a277218e — dropped the `Node port of audit_sdd.py` framing from the header and the four remaining `audit_sdd.py` references in the body (the divergence comment, the Response Map section header, and the two lines in its rationale block). Same treatment applied to `audit-plan.mjs` for symmetry. The Node-availability rationale stays; the comparison to the Python is gone.

Zero `audit_sdd.py` / `audit_plan.py` mentions now remain anywhere outside the `.py` files themselves, which the follow-up deletes.

roberts-cliff:
Addressed in `4a277218e` — all three findings.

**@abhiram-vad's two** — dropped every `audit_sdd.py` / `audit_plan.py` cross-reference. Zero mentions now remain outside the `.py` files themselves. Agreed on the reasoning: with the Python going away in the follow-up, pointing at it from Node docs only muddies which one is authoritative. The guard comment no longer names the planner or its scripts either — it now makes the general point (one conversation can run more than one skill's read-only auditor) without the cross-skill reference. Regex unchanged.

**Review bot's Medium** — `case-design-lane-guide.md:280` had a leftover *"All three of `python3` / `python` / `py` unavailable → verify by hand"* sentence sitting directly under a `node` command. Removed. Good catch; that one would have told an agent to hand-verify on a condition that no longer gates anything.

Script changes this commit are **comment-only**. Differential parity against the Python originals re-run and unchanged: 9/9 sdd fixtures, 6/6 `tasks.md` fixtures.

**On the lint bot's High (missing passing-run claim):** I have not run `skill-case-artifact-safety-guard` locally. A coder-eval run of it plus the 6 other previously-failing case tasks is in flight against this branch — https://github.com/UiPath/skills/actions/runs/32899817428 — and I'll post the result here rather than claim a pass I haven't seen. Worth noting one of those 7 originally failed on infra (a 503), not logic.

roberts-cliff:
## Coder-eval result: 3/7 — [run 32899817428](https://github.com/UiPath/skills/actions/runs/32899817428)

Re-ran the 7 previously-failing case tasks against this branch.

| task | result | cause |
|---|---|---|
| `aged-invoice-structural` | ✅ SUCCESS | |
| `edit-bound-task-add-remove` | ✅ SUCCESS | |
| `golden-rebuild-cm-expense` | ✅ SUCCESS | |
| `e2e-expense-runnable` | ❌ 0.600 | **infra** — `solution resources refresh` → `Failed to authenticate` |
| `e2e-outcome-external` | ❌ 0.375 | **infra** — same expired refresh token |
| `phase-0-procurement-sla-interrupts` | ❌ | checker failure, unrelated to this PR |
| `sla-sdd-response-map` | ❌ 0.250 | see below |

**The two e2e failures are the CI tenant's stored refresh token being expired**, not logic: `"Message": "Failed to authenticate", "Retry": "RetryWillNotFix"`. Both got as far as a clean `case validate` and then died at `resources refresh`. Worth a separate look at the runner's credentials.

### `sla-sdd-response-map` — my new check is necessary but not sufficient

This is the honest finding. Score is **0.250 both before and after this PR** — byte-identical failure. The new `responseMapFindings()` neither fixed nor broke it.

What the task's own `check_sdd_response_map.py` grades:

```
- sla-status-change("Assess","Assess SLA") targets 'Assess' but every SLA Response Map
  row titled 'Assess SLA' is scoped ['stage: Assess (`assess`)']
- sla-status-change("root","OrderExceptionReview Case SLA") targets 'root' but every row
  titled 'OrderExceptionReview Case SLA' is scoped ['case: root']
```

That is **scope↔target alignment** — that each map row's Scope cell resolves to the same thing its `sla-status-change` call targets. My check covers section presence, the seven columns, and legal Response tokens. It does not cover this.

So the template's promise is now *partly* kept rather than wholly unkept. I'd rather land the partial fix and be explicit about the remaining gap than quietly imply the check is complete. Happy to extend `responseMapFindings()` to cover scope↔target in this PR if you'd prefer it closed in one go — @abhiram-vad your call.

### On the lint bot's High

Still not claimable: `skill-case-artifact-safety-guard` — the task this PR actually modifies — **was not in this run**. I dispatched the 7 previously-failing tasks, and that guard task was not among them. Running it next.

roberts-cliff:
## Guard task passes — and scope↔target is now covered

**Lint bot's High is closed.** `skill-case-artifact-safety-guard` — the task this PR modifies — ran against the branch: **5/5 criteria, weighted 1.000, SUCCESS** ([run 32905385073](https://github.com/UiPath/skills/actions/runs/32905385073)). Claim added to the PR description. It wasn't in my earlier 7-task run because that run was the previously-failing set, and this guard task wasn't one of them.

## scope↔target folded in — `5eb7da4bc`

@abhiram-vad I went ahead and closed it in this PR rather than leaving a known-partial check in place. Ported the rest of the contract from `tests/tasks/uipath-planner/_shared/sla_response_map_check.py` — the eval's own ground truth — so the gate flags exactly what the task grades:

- **scope↔target** — the matched row is scoped to what the call targets (`root` → `case`; a stage target → `stage: <name>`), with its Interrupting value surviving into the conditions table rather than defaulting
- two-way closure between map rows and `sla-status-change` entries
- notify-only rows carry `—` for Target and Interrupting; other responses name a Target
- `start-task` carries `—` (a task-entry rule interrupts nothing; `No` would imply stage re-entry, re-running the breached stage's other tasks), `enter-stage`/`exit-*` carry an explicit Yes/No

Scope matching is **exact on the stage name**. `case-sdd-template.md` specifies the bare form, so `stage: Assess (\`assess\`)` — precisely what the failing eval run emitted — is a real defect, not a formatting variant. Title-only matching would also let a case-level and a stage-level SLA sharing a title satisfy closure with a row for the wrong target.

Verified against the Python checker:

| check | result |
|---|---|
| failing eval fixture | 10/10 findings identical |
| synthetic fixture (scope suffix + Interrupting mismatch) | identical |
| same fixture, defects repaired | both silent — no false positives |
| warranty SDD v1.1 | 0 Response Map findings (unchanged) |
| full parity suite | 14/14 unchanged |

One thing worth flagging for review: the first cut of this **crashed** with a `ReferenceError` on any SDD that reached the scope↔target path (`pyTupleRepr` lives in `audit-plan.mjs`, not this file). My initial differential showed 10/10 clean because that fixture fails on missing columns and returns before reaching the new code. It only surfaced when I read full output instead of grepping. Fixed with `pyListRepr` — Python prints a list repr here, not a tuple — but it's a reminder that the fixture set doesn't reach this path by default. The warranty SDD has zero graph-entry rows and zero `sla-status-change` calls, so it exercises none of it; the synthetic fixture is the only coverage.

Template promise updated to describe what is now actually enforced.

abhiram-vad (approved):

skills/uipath-planner/scripts/case/audit-case-sdd.mjs
```diff
@@ -0,0 +1,1394 @@
+#!/usr/bin/env node
```
AlvinStanescu, line 1: Should be called `audit-case-sdd`. That being said, are the rules clear for the agent without it invoking this?

AlvinStanescu (dismissed):
I do not believe in this auditing approach where we try to "validate" a textual document via regex (instead of a subagent with clear rules) and would like some clarifications on why we want to move ahead with this.

What if the user specifically wants a slightly different structure?
What if they want the PDD or SDD in a different language (e.g. German or Japanese customers use localized PDDs and SDDs and use the agent in their respective language).

Let's please consider switching this to a validation agent approach, giving that agent a couple of files to validate against. This approach is, I believe, much safer, and more reliable.

roberts-cliff:
> I do not believe in this auditing approach where we try to "validate" a textual document via regex (instead of a subagent with clear rules) and would like some clarifications on why we want to move ahead with this.
> 
> What if the user specifically wants a slightly different structure? What if they want the PDD or SDD in a different language (e.g. German or Japanese customers use localized PDDs and SDDs and use the agent in their respective language).
> 
> Let's please consider switching this to a validation agent approach, giving that agent a couple of files to validate against. This approach is, I believe, much safer, and more reliable.

this change was intended as a simple port from python to js, a couple bug fixes snuck into the set.
my intention after the execution language port is to pull this deterministic check into the cli where i think it belongs.
i had not considered i18n, as is the skill is basically blind to that consideration, i'll bring this concern to my team

skills/uipath-planner/scripts/case/audit-case-sdd.mjs
roberts-cliff: Renamed to `audit-case-sdd.mjs` in `868c1523e` — you're right, it validates against the Case template specifically and sits beside six other SDD templates it knows nothing about. All four referencing docs updated.

> are the rules clear for the agent without it invoking this?

I went looking for evidence rather than guessing. Took six coder-eval runs that produced an `sdd.md`, pulled each agent's recorded command list out of `task.json`, and checked two things: did it ever invoke the auditor, and does the document it produced conform.

| task | cmds recorded | invoked auditor? | findings |
|---|---:|---|---:|
| `aged-invoice-structural` | 22 | no | 4 |
| `e2e-expense-runnable` | 20 | no | 41 |
| `e2e-outcome-external` | 26 | no | 17 |
| `golden-rebuild-cm-expense` | 33 | no | 40 |
| `phase-0-procurement-sla-interrupts` | 8 | no | 2 |
| `sla-sdd-response-map` | 8 | no | 10 |
| | **117** | **0 of 6** | **114** |

**The rules read as clear — that is what makes this interesting.** They are written out in `case-sdd-template.md` and the design guides, the skill mandates running the gate, and across 117 recorded commands not one run invoked it. Every resulting document was non-conformant.

I checked whether 114 is one structural miss cascading. It is not — the findings are independent, per-task and per-stage:

```
72 x  task 'X' missing '<marker>'     (mostly **Activation Mode:** / **Design Rationale:**)
14 x  missing required heading
11 x  stage 'X' missing '<marker>'
```

The `**Activation Mode:**` misses are not cosmetic. Per the case skill's Rule 6, a missing entry rule only *warns* under `validate` — but it **hangs `case debug` indefinitely**. That is the class of defect this catches.

So the failure mode is not ambiguous instructions. It is documented instructions being read and not followed, and a mandated gate not being run. That is the specific thing a deterministic check is good at and an instruction-following agent is not — which is why I would rather this end up in the CLI than in more prose.

**Caveat, stated up front:** these six are the *failed* eval set, so it is a biased sample. It establishes the failure mode exists and is severe; it does not establish a base rate. Happy to audit a full nightly run if the unbiased number would change your read.

roberts-cliff: Renamed to `audit-case-sdd.mjs` — done in `7275ce921`.

On your second question (*"are the rules clear for the agent without it invoking this?"*): the branch no longer adds any rules, so it is moot for this PR. It is now a pure runtime swap — same checks, same finding text, same exit codes as the Python.

Worth answering anyway, since it is the right question. I looked at six coder-eval runs that produced an `sdd.md`, pulled each agent's recorded command list from `task.json`, and checked whether it ever invoked the auditor:

| task | cmds | invoked auditor? | findings on its SDD |
|---|---:|---|---:|
| `aged-invoice-structural` | 22 | no | 4 |
| `e2e-expense-runnable` | 20 | no | 41 |
| `e2e-outcome-external` | 26 | no | 17 |
| `golden-rebuild-cm-expense` | 33 | no | 40 |
| `phase-0-procurement-sla-interrupts` | 8 | no | 2 |
| `sla-sdd-response-map` | 8 | no | 10 |
| | **117** | **0 of 6** | **114** |

The rules read as clear — they are written out in `case-sdd-template.md` and the design guides — and across 117 recorded commands the mandated gate was invoked zero times, with every resulting document non-conformant. Not one cascade either: 72 of the 114 are independent per-task marker misses (mostly `**Activation Mode:**`, which only *warns* under `validate` but hangs `case debug`).

Caveat: those six are the *failed* eval set, so it is a biased sample — it shows the failure mode is real and severe, not its base rate.

roberts-cliff:
@AlvinStanescu — rewritten as a pure port. Your review is still holding this as CHANGES_REQUESTED, so it needs your eyes to clear; here is what changed.

**Every addition you objected to is gone.** The branch is now two commits and changes the runtime only:

| | |
|---|---|
| `7275ce921` | add `audit-case-sdd.mjs` + `audit-plan.mjs`; Python left in place |
| `35ca32bb3` | cut docs over, delete both `.py` |

Removed since your review:
- the widened `\bvars\.` lineage anchor → reverted to the Python's `=vars\.`
- the entire SLA Response Map check (section presence, columns, legal tokens, two-way closure, scope↔target)
- the `case-sdd-template.md` wording that described those checks

Those were real findings, but they are new enforcement and they belong in their own PR argued on their merits — not smuggled in on a port. Nothing here validates anything the Python did not already validate, identically.

**Parity is now unfiltered.** Previously I could only claim it with carve-outs, because the branch carried behaviour changes. Full stdout+stderr, no exclusions:

```
audit-case-sdd.mjs  vs audit_sdd.py     9 sdd.md fixtures + --draft   identical
audit-plan.mjs      vs audit_plan.py    6 tasks.md fixtures            identical
PARITY: PASS=15 FAIL=0
```

**On the regex-vs-agent question you raised** — I am not relitigating it here, because this PR no longer takes a position on it. It swaps `python3` for `node` on checks that already exist and already run. Whether deterministic auditing is the right approach at all is a live question, and Cliff's plan is to pull this into the CLI next, which is where that discussion belongs.

**On i18n** — you were right that nothing here handles it. I checked: the repo has no i18n infrastructure at all (no locale dirs, no `.po`/`.xliff`/`.resx`, all seven SDD templates 100% ASCII English), and `audit-case-sdd.mjs` is the only script in the repo with hard-coded English structural literals — 37 of them — because it is the only one that parses a natural-language document. The real boundary is the template: `case-sdd-template.md` prescribes English headings, the planner authors from it, and the auditor checks conformance to it. A German SDD fails because nothing upstream can produce or consume one, not because a regex is brittle. Swapping the regex for an agent would make the auditor tolerant of a document the rest of the pipeline still cannot handle. That is an argument for taking i18n up at the template/product layer.

One thing deliberately **not** fixed, flagged so it is not mistaken for an oversight: the guard's exemption still scopes to `uipath-maestro-case/scripts/`, so an agent running the planner's auditor is flagged by `skill-case-artifact-safety-guard`. I verified against `origin/main` that this already happens with `python3 <planner>/scripts/audit_sdd.py` — pre-existing, not port fallout, and it deserves its own change.

PR #2819, opened 2026-08-25
ci: default coder-eval runs to codex + gpt-5.6-luna

bai-uipath:
Codex is the harness the nightly runs every day; claude runs twice a week. But the two places most people actually start a coder-eval run from — the PR smoke gate and the ad-hoc `Run Coder Eval` dispatch — both still defaulted to `claude-sonnet-5`. So the gate that decides whether a skill change is good measured a twice-weekly harness rather than the daily one, and every ad-hoc run cost roughly fifteen times what it needed to.

This points both defaults at codex, with the model coming from the `CODEX_MODEL` repo variable as before. That variable moves from `gpt-5.6-terra` to `gpt-5.6-luna` alongside this PR (a repo-settings change, not visible in the diff). Codex on luna measures at pass-rate parity with `claude-sonnet-5` and `claude-sonnet-4-6` on this suite, so this is a price change rather than a capability trade.

The evaluation side is deliberately untouched. `llm_judge`, the simulated user and the LLM reviewer stay on Bedrock Claude via `BEDROCK_MODEL`, so the grader does not move with the agent and scores stay comparable across harnesses.

Both workflows already supported codex — the experiment files are harness-agnostic and already pass the codex credentials through to the sandbox. What changes here is which one you get without asking.

Out of scope, on purpose: the ADO nightly's own `CODEX_MODEL` (still terra), the judge and simulated-user defaults, and every PR review workflow.

## Testing

Both defaults were run on this branch before asking for review.

**The smoke gate on this PR is itself the test.** A `smoke-skills.yml` edit trips the infra-trigger fan-out, so it selected a 40-task deterministic sample of the non-Windows tree and ran every one on codex + `gpt-5.6-luna`.

| | |
|---|---|
| Pass rate | **97.5% (39/40)**, against a 95% threshold |
| The one miss | `skill-datafabric-smoke-complex-fields`, a `pre_run` tenant-seeding failure (`could not seed value(s) on CE_SmokeTags`) at `iterations=0` — the agent never started |
| llm_review gate | reported all 40 above 0.7, but see the caveat below |

**Ad-hoc dispatch, run with no agent input so it took the new defaults:**

```
Running: coder-eval run tasks/uipath-admin/audit/audit_events_basic_smoke.yaml
  -e experiments/nightly.yaml --type codex --model gpt-5.6-luna -j 4
skill-admin-audit-events-basic-smoke: SUCCESS
1/1 PASS (linux)
```

One thing the run surfaced that is not caused by this change and is worth someone's attention separately: **not one of the 40 tasks carried an `llm_review` score** — every row is `llm=—`, and the reviewer logged skipped-or-errored on all of them. Two recent smoke runs from main-based branches show the same (0 of 1 scored, 0 of 16), so this predates the PR: the coder_eval reviewer only supports Anthropic direct or the LLM Gateway, not Bedrock, and the shared workspace quota is drained. The practical consequence is that the hard `llm_review >= 0.7` gate is currently passing vacuously, on claude as much as on codex.

### DIFF
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -75,17 +75,23 @@ on:
       # Windows runs it in-process on the host (tempdir driver) with its own
       # delegate-stdio install, ROPC user token, and shared interop — ports of
       # coder_eval_uipath's daily.sh / daily-windows.ps1 wiring.
+      #
+      # codex is the DEFAULT: it is the harness the nightly runs daily (claude runs
+      # twice a week), and codex on the cheap tier measures at pass-rate parity with
+      # claude-sonnet-5 for a fraction of the cost. An ad-hoc run should reproduce the
+      # daily driver by default, not the twice-weekly one.
       agent:
         description: 'Agent to run.'
         type: choice
         options:
-          - claude
           - codex
+          - claude
           - antigravity
           - delegate-sdk
-        default: claude
+        default: codex
       # Overrides the model for ANY agent. Blank keeps that harness's own repo
-      # variable (CLAUDE_CODE_MODEL / CODEX_MODEL / ANTIGRAVITY_MODEL).
+      # variable (CLAUDE_CODE_MODEL / CODEX_MODEL / ANTIGRAVITY_MODEL) — which is
+      # where the default model lives, so it stays correct whichever agent is picked.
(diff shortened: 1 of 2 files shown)


uipreliga (approved):
🚢

PR #1978, opened 2026-07-10
test(uipath-maestro-flow): DF connector coder-eval batch

AditiGoyalUipath:
## Summary

Split out of #890 so that PR can stay scoped to the DF Flow skill removal + connector-skill V2 guidance. This PR adds only the test artefacts.

Coder-eval batch covering the UiPath Data Fabric connector in Maestro Flow:

- **Activation:** 1 positive + 1 negative routing test
- **Smoke:** create-all-types, update (partial-body), file activities (upload/download/delete on `file1`), CEQL query filter + sort/cursor pagination, error (nonexistent entity + parallel query), brownfield edit-and-revert on a pre-scaffolded `MovieReportFlow`
- **Integration:** all-fields round-trip (C-01) + expansionLevel 1/2/3 comparison (C-02)
- **E2E:** ContractRegistry pipeline — bulk-create 6 records with priority-based logic, sort-query for top priority, filter-query for urgent, targeted update, delete-all cleanup

Each task ships with its shape verifier (`check_*.py`), entity fixture, and `pre_run` wiring. `ensure_entity.py` idempotently creates the DF entity from a JSON definition; `scaffold_movie_report_flow.py` seeds the brownfield fixture and snapshots `pre_state.flow` so the revert verifier can diff against it.

## Test plan

- [ ] `python3 scripts/check-task-yaml.py tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/*.yaml` passes lint
- [ ] Each smoke task runs green under `coder-eval` against a tenant with a Data Fabric connection
- [ ] Verifiers exit 0 on the reference agent output for their task

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2822, opened 2026-08-26
test(uipath-maestro-flow): DF connector coder-eval follow-up fixes

AditiGoyalUipath:
Follow-up on #1978 (squash-merged). Two grading fixes surfaced by the first full runs after merge.

- smoke_update_existing_flow: min_count on the `flow node configure` trace 2 → 1. `configure ≥ 2` false-failed valid runs where the agent used `Edit` to revert (the tool is in the smoke allow-list). The deep-equal snapshot check already asserts revert integrity; `configure ≥ 1` is enough to kill the no-op-does-nothing case.
- trigger_lifecycle: `flow validate` criterion timeout 60 → 180s. The loop iterates 2 flows; each `uip maestro flow validate` fetches the ~389-entry dynamic-node manifest fresh, so 60s was tight. Ran in ~60.6s and hit the wall.


chandusailella (approved):
please add code reviewers from DF team as well - `tasks/uipath-maestro-flow/connector_features/datafabric_connector`

gabrielavaduva (approved):

PR #2827, opened 2026-08-26
fix(uipath-ixp): any Documents shortfall is REVIEW — drop the marker allowance

cezara98t:
## Problem

The shortfall smoke's missing-marker case failed three consecutive CI runs under three wordings: agents kept tagging a `Documents` 4-of-5 field REVIEW where the guide's allowance ("short by one or two is missing markers") demanded UPLOAD.

The source shows the payload cannot carry the answer. In `ixp-platform` (`mls/user-model-store/.../metrics/entity.py`), the per-field document count is *"user model documents containing at least one assigned entity"*, with no empty-value filter at the scoring layer — and no CLI read exposes per-document annotation state. Whether a marker-only document counts is not traceable from the metrics code (markers could be stored as assigned entries or as dismissed negatives), which is the point: **from the numbers alone, the agent cannot know why a document is missing from a field's count**. The allowance was a bet the data cannot back, and the smoke graded a guess as if it had one right answer.

## Change

The rule becomes decision-free, and the cost asymmetry favors it:

- **`Documents` equal to `ValidatedDocuments`** → the sample is as large as the labelled data allows → **UPLOAD**.
- **`Documents` below `ValidatedDocuments`** → documents without evidence for the field, and the payload cannot say why → **REVIEW** — the review shows which in seconds, while an unreviewed document left unfound caps the field for good.

The smoke's 4-of-5 field flips to REVIEW, and its criterion now guards the opposite mistake — rounding a small shortfall up to "complete". The `Annotations` trap (largest `Annotations` on the fewest documents) is unchanged. Rule, test, and observed agent behavior now agree, so the grading is deterministic.

Based on `main` so the fix merges independently; #2825 (the overlay consolidation) rebases on top of this afterwards.

### DIFF
--- a/skills/uipath-ixp/references/improve-prompts-guide.md
+++ b/skills/uipath-ixp/references/improve-prompts-guide.md
@@ -278,8 +278,8 @@ That is 0.2 at `Annotations` = 5 — one flipped annotation is not evidence — 
 
 **A small `Annotations` has two causes with opposite remedies.** `Annotations` counts reviewed **extractions**, not documents — one document can contribute several — so it cannot be compared against a document count directly. Compare the field's own `Documents` against the project-level `ValidatedDocuments`:
 
-- **`Documents` equal to `ValidatedDocuments`, or short by one or two** → the sample is already as large as the labelled data allows (a small shortfall is missing markers — the field legitimately absent from a document — not an unreviewed field). Tag it **UPLOAD**.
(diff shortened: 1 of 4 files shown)


misupantea-uipath (approved):

cezara98t:
C:/Program Files/Git/cherrypick release/v1.201

PR #2830, opened 2026-08-26
fix(uipath-ixp): the version follows the question — improvement baseline reads latest, graded

cezara98t:
## Problem

Two guidance sources disagree on which version an improvement loop baselines against, and the always-loaded one wins:

- **SKILL.md Critical Rule 21** (always in context): resolve the version first — `Tags[]` live, else pinned, else latest — which reads as live-first for *any* score reporting.
- **improve-prompts-guide 1a** (fetched on demand): the loop's baseline is the **latest trained model** — the model the loop iterates on.

Measured, not speculated: two CI runs of the new smoke answered with the live version (9 / 0.98) for a prompt that names the improvement loop and the word *baseline* outright. An agent obeying Rule 21 baselines the loop against the wrong starting point whenever live and latest differ — on the real project this work was verified against, that gap is live v9 at 0.98 vs latest at 0.89. Baselining against live also breaks the loop's arithmetic: the first comparison then bundles the edit under test with everything that happened across the intervening versions.

## Change

- **Rule 21 is stated intent-first** — the version follows the question: *how is it performing / production* → the live version (live tag, else highest pinned, else latest); *baseline for improving instructions* → the LATEST trained version, because that is the model the edits retrain (guide 1a). Then pass `--model-version <N>` and name the version — no exception clause, one decision with two branches.
- **1a's baseline command names its own version**: `get-metrics <project-name> --model-version latest`, with one line saying why latest is the baseline. No defended bare call; `latest` as a flag value verified live.
- **`get_metrics` smoke** grades the rule as the mirror of `metrics_live_version` on the shared fixture, and each twin's description names the other:

| smoke | intent | graded answer |
|---|---|---|
| `metrics_live_version` | "performing **in production**" | version 9, score ≥ 0.95 |
| `get_metrics` | "**baseline** for an **improvement loop**" | version 30, score = 0.91 |

The outcome check is what makes intent-driven version choice gradeable: an agent resolving live regardless of intent answers 9 / 0.98 and fails, and the values are only reachable from an actual `get-metrics` read. The Name-not-Title command criterion stays. (The twin keeps `gte 0.95` because it discriminates there; here `gte` could not separate 0.91 from 0.98, so the score asserts `equals 0.91` — the mock's exact value, copied verbatim per the skill's state-numbers-plainly rule.)


cezara98t:
Closing per the decision rule: two CI runs answered with the live version despite the prompt naming the improvement loop and its baseline outright. The failure exposed a real guidance conflict rather than a bad agent: SKILL.md Critical Rule 21's resolution ladder (live tag, else pinned, else latest) instructs live-first for ANY score reporting and is always in context, while improve-prompts-guide 1a defines the loop's baseline as the latest trained model and is only fetched on demand. Until that conflict is resolved in the docs, the test grades guidance the skill itself contradicts. Branch left in place in case the doc fix lands and the test is revived.

alexandrujircan (approved):

PR #2831, opened 2026-08-26
docs(uipath-ixp): a published tag is a runtime address — DU framework consumes project+tag

cezara98t:
## Problem

The "Publish the model" row says publishing *"makes the version usable inside the project"* — implying nothing outside can consume a published version. That's wrong for one whole runtime surface: the **DU framework API** (`…/du_/api/framework/…`) serves IXP projects **by project + tag** — `DiscoveryService` handles `ProjectType.IXP` in the project-based flow, and extraction is addressed as `/projects/{projectId}/{tag}/document-types/{…}/extraction` (DU-App, `src/framework/Discovery`). A published `live` tag is exactly what those callers — and any DU activities calling through the framework — consume. That's why the tag is called *live*.

What publish does **not** do is surface the model to callers that resolve models from an **Orchestrator folder**. Those are fed only by `deployments create`: Maestro Flow nodes (`FolderBasedDiscoveryService` → `IxpDeployment`; verified empirically — a project published + live-tagged for a month is absent from the flow registry, a folder-deployed one is listed), the framework's own folder-based v2 surface (`FolderBasedExtractionControllerV2_0`, tagged `SolutionDeployedModelExtraction`), and the agent-tool registry variants (`uipath.agent.resource.tool.ixp.*`).

## Change

Two sentences, same story:

**1. The "Publish the model" row.** Each side names its consumers plainly, hedged so neither reads as an exhaustive roster ("DU activities that call through it" / "Maestro Flow among them"):

> **Publishing does not deploy the model to an Orchestrator folder** — publish pins the version, and `--tag` moves the `live`/`staging` tag, which selects the version the DU framework serves (including to DU activities that call through it). Callers that resolve models from an Orchestrator folder — Maestro Flow among them — see only `deployments create` (the "Deploy this model to a folder" row below).

**2. The "deploy to staging or production" row** carried the same idea where it hurts most — a user asking to push to staging may mean exactly what the tag does. Its note now reads: *"`projects publish --tag staging|live` moves the tag the DU framework (and DU activities calling through it) resolve — for those consumers that IS the staging/live switch; it creates no folder deployment."*

Everything else in both rows (untagged default, `--tag`, don't chain a deploy onto a publish, folder-named routing) is unchanged.

Draft: #2809 touches the same conceptual ground (its deployment-guide says "Nothing outside the project can call it"); keeping this as draft until that review settles so the two land consistently.


andrei-uipath (approved):

skills/uipath-ixp/SKILL.md
```diff
@@ -75,7 +75,7 @@ If the user provides a taxonomy file, use `--skip-taxonomy` and `import-taxonomy
...
+| "Publish the model" / "Tag as live" | `uip ixp projects publish <project-name> --output json` — publishes the latest version, untagged. Add `--tag <live\|staging>` to also tag it. See [cli-reference](references/cli-reference.md) for `--model-version`/`--description`. **Publishing does not deploy the model to an Orchestrator folder** — publish pins the version and moves the `live`/`staging` tags — the tagged version is what the DU framework, and DU activities that call through it, resolve; `deployments create` (see the "Deploy this model to a folder" row below) is what callers resolving models from an Orchestrator folder — Maestro Flow among them — see. Do NOT chain a deploy onto a publish unless the user asked to deploy — a deploy needs a folder key and changes what runtime callers get. |
```
horeaprecup-uipath: > and moves the `live`/`staging` tags — the tagged version is what the DU framework, and DU activities that call through it, resolve

In the cli, the description of the --tag is: 'Optionally tag the published version. Allowed values: "live" or "staging".' . Instead of 'moves', shouldn't we use 'can move' or 'can set'?

cezara98t: Right — `--tag` is optional, so the unconditional "moves" overstated. Fixed in e434f121a: "publish pins the version and, with `--tag`, moves the `live`/`staging` tag — …". The staging/production row (line 126) keeps plain "moves" since it quotes `publish --tag staging|live` explicitly, making it conditional by construction.

horeaprecup-uipath (approved):

PR #2838, opened 2026-08-26
fix(uipath-maestro-case): take connector schema property names from the spec's own authority

song-zhao-25:
Fixes both legs of a failing eval run (score 0.500), which turned out to be two unrelated defects.

## Leg 1 — names destroyed by the lowering pass

`uip maestro case spec --output json` PascalCases every key, deleting underscores. Phase 3 then lowered the first character of every spliced key, breaking contract field names in both directions:

| Contract | Serializer sends | Lowering gives | Result |
|---|---|---|---|
| `app_id` | `AppId` | `appId` | wrong name, `validate` green |
| `Title` (Outlook) | `Title` | `title` | a name that was **already correct**, destroyed |

The second row is why this can't be fixed by scanning for leftover capitals — on a capital-first connector that scan *is* the defect.

**Fix:** the keys **of** a `properties` / `definitions` map are connector field names, so they're excluded from the lowering pass and taken from their authority — `properties` keys from the ordered distinct segments of `ResponseFields[].Name` / `Inputs.*[].Name` under that map's accumulated dotted path; `definitions` keys from the tail of the targeting `$ref`. The `Error` output is the platform envelope and still lowers.

Plus the three capital-first gates carved out. Not optional: in the failing run the agent read the exception and then **overruled it**, quoting the HARD GATE verbatim in its own reasoning.

## Leg 2 — properties missing, not mis-cased

The Slack node was built with **4 of 13** response properties. Not a naming bug: the agent hand-wrote `tasks/spec-cache.<elementId>.json` from reasoning, producing 8.4 KB from a 68 KB response and holding 4 of 102 `ResponseFields`. It hand-wrote it because Rule 13 banned shell redirection for that file. The naming rule then read a starved authority.

**Fix:** redirect on both connector paths, with a narrow Rule 13 exception for that one file (`trigger-spec-cache.json` stays excluded — it's a merge-by-T-number sidecar `>` would truncate). Plus an explicit instruction to copy the body whole, with the scale stated and a four-count self-check, since `validate` reports a truncated schema as `Valid`.

## Ground truth

Verified against the live tenant and against a hand-built Studio Web solution. The FE writes the **full** schema — Slack `response` is 514 keys (13 top-level + 118 nested property keys, 29 definitions, 29 `$ref`s), Outlook is 120 — not just the bound leaves. It lower-cases the `Error` envelope and preserves Outlook's capital-first names. The positional match between property maps and authority segments holds at every depth across 8 operations on 2 connectors.

## Paired test

`tests/tasks/uipath-maestro-case/connector_naming/` — one stage, two tasks spanning both conventions (Slack snake_case, Outlook capital-first), stopping at `validate`; the case never runs. The grader derives every expectation from a live `case spec` at grade time rather than a frozen golden. Verified against 14 synthetic caseplans: an ideal build passes, every injected defect fails.

## Status

**Draft — no passing run yet.** Known open risk: inside a payload body one spelling is often both a JSON-Schema keyword and a contract field name (`"Title"` occurs 134× in the Slack body, `"Type"` 161×), so `replace_all` cannot separate them and this PR gives no mechanism for that. Deliberately left out rather than guessed at — the next run decides whether it's needed.

Two unrelated bugs found along the way are **not** here: the filter-sink docs assert a compiled `body.filters.expression` that a literal-only filter doesn't produce (verified live), and the event-trigger sidecar has no slot for the `ResponseFields` it needs. Both deserve their own PRs.

Supersedes #2815.


song-zhao-25:
Superseded — closing in favour of the CLI-side fix plus a much smaller skill change.

**Root cause turned out to be a CLI bug, not a skill gap.** `OutputFormatter.success()` PascalCases every key inside `Data` unless the caller opts out via `preserveDataKeys`, and `case spec` never did. In a JSON Schema the `properties`/`definitions` keys ARE the connector's field names, so the policy renamed user data — irreversibly, because it deletes underscores (`app_id` → `AppId`, and lowering gives `appId`).

Fixed at source in the CLI: `fix/case-spec-preserve-schema-keys` (case-tool `8c7d21e86`). Verified against the built formatter — `app_id`, `Title`, `block_id`, `messageBot_profile` all survive and `#/definitions/messageBot_profile` resolves.

**So everything this PR added is obsolete**: the authority table, the positional-zip derivation, the `$ref`-tail rule, the keyword/field collision handling, the copy-then-rename mechanism, the gate carve-outs. `caseShape` now arrives disk-ready and is spliced verbatim.

Replaced by `fix/case-cli-camelcase-caseshape`, which is net deletion (3 files, +31/−40) and mostly removes what this PR was trying to get right. It also fixes read paths to the mixed shape the CLI fix produces (`.Data.CaseShape.context` — container PascalCase, contents camelCase; both spellings in the docs today are wrong).

Carried forward from here:
- the `>` redirect to the per-node spec-cache and its Rule 13 carve-out — that fixed a **different** bug (a hand-written 8.4 KB cache from a 68 KB response, 4 of 102 `ResponseFields`), unaffected by casing
- the `connector_naming` test unchanged — its grader resolves paths case-insensitively and reads its authority from the non-preserved `Data.Outputs.ResponseFields`, so it grades against both binaries and can verify the CLI fix

Worth recording why this ran long: five eval runs, each fixing a real defect and revealing the next, all of them downstream of a CLI bug none of the prose could have fixed. The three that were mine were self-inflicted — documenting where `validate` is blind, which the agent read as permission to skip the work.

song-zhao-25:
Closing — superseded by the CLI fix plus `fix/case-cli-camelcase-caseshape`. See the previous comment.

PR #2609, opened 2026-08-13
feat(uipath-insights): teach Insights RBAC reads and add an RBAC smoke eval [IN-13526]

celestinryf:
https://uipath.atlassian.net/browse/IN-13526
Also related to https://uipath.atlassian.net/browse/IN-13527: the tagged eval tasks here are that ticket's skills-side test layer.

## What

Teaches the six Insights RBAC read commands: `users list`, `users get`, `roles list`, `roles get`, `groups list`, and `groups get`.

- New `references/rbac-reads-guide.md`: per-command usage and fields, the name-to-GUID join, a Safe Output section, a Failure Branches section, and the read-shaped side effects.
- `SKILL.md`: a routing row, RBAC in the frontmatter with a `→uipath-admin` redirect, two scope-boundary rows, one anti-pattern, and Critical Rules 15 and 16, the RBAC counterparts of the alert read-only rule.
- New `tests/tasks/uipath-insights/rbac/smoke.yaml` and four activation rows (034-037). The smoke task gates each of the three command groups with its own `--help` reality check, and carries the CLI-bypass and shell-variable negatives the alerts smoke uses, with the PermissionService route strings.

## Why

The six reads ship as five stacked cli PRs, and this is the companion to all of them: [#3860](https://github.com/UiPath/cli/pull/3860) shared refactor (merged), [#3660](https://github.com/UiPath/cli/pull/3660) users list (merged), [#3861](https://github.com/UiPath/cli/pull/3861) users get, [#3858](https://github.com/UiPath/cli/pull/3858) roles, and [#3859](https://github.com/UiPath/cli/pull/3859) groups. One guide covers the family, so the packet has one skills half rather than five. These reads share vocabulary with `uip admin` but read tenant-scoped Insights role assignments from the Insights Portal backend, not org identity. Without the boundary stated an agent picks the wrong skill, which is the confusion raised on #2552.

These six run without `--output`, the one exception to Critical Rule 1. The format still resolves to json so the envelope is unchanged, but the CLI's safe projection stays on and email addresses, nested role IDs, and the role resource string never reach the transcript. The name-to-GUID join works entirely in the safe view, so nothing is lost by leaving the flag off.

Three of the six are not pure reads: `roles list` can make a cached Licensing call, `groups get` persists refreshed directory fields, and `groups list` drops a group whose directory entry does not resolve. That last one also 404s `groups get`, so a group missing from the list and failing its `get` is one cause, not two.

## Testing

[...]

PR #2839, opened 2026-08-26
feat(uipath-insights): teach queue monitoring [IN-13530]

celestinryf:
https://uipath.atlassian.net/browse/IN-13530

The commands ship in four stacked cli PRs: [UiPath/cli#3826](https://github.com/UiPath/cli/pull/3826) (summary, sla, operational-metrics), [UiPath/cli#4108](https://github.com/UiPath/cli/pull/4108) (timelines), [UiPath/cli#4109](https://github.com/UiPath/cli/pull/4109) (failure analysis), and [UiPath/cli#4110](https://github.com/UiPath/cli/pull/4110) (details and retry-outcomes). This half waits for the last of them to merge and publish `@uipath/cli@dev`, which the smoke gate installs.

## What

Teaches the `uipath-insights` skill the ten queue monitoring commands: `queues summary`, `sla`, `operational-metrics`, `completed-timeline`, `uncompleted-timeline`, `top-failures`, `failures-by-reason`, `failure-details`, `details`, and `retry-outcomes`.

- New `references/queue-monitoring-guide.md`: the catalog, the per-command flags, the output shapes, and an investigation workflow.
- `SKILL.md` gains queue routing, queue triggers, and the queue half of Critical Rules 1, 4, 5, and 11.
- Six activation rows, plus three fixed command patterns in `smoke_critical_rules.yaml`.
- New `tests/tasks/uipath-insights/queues/smoke.yaml`: eight commands driven from a prompt that names none of them, a `--help` reality gate per subcommand, and negatives for the guide's own traps. Ships skipped, see Limitations.
- `filter-discovery-guide.md` and `jobs-commands-guide.md`, stale the moment SKILL.md routes a queue question through the filter guide.

## Why

WS2.2 Q1 adds queue monitoring to the CLI. Agents need the matching guide to pick the right command, choose the right `--time-event` column on the two timelines, and read the output correctly. One guide covers the whole family, the way the RBAC guide covered its five cli PRs.

## Verification

Read from the CLI that ships the commands, `feature/insights-queue-details` at `7210bc30a` plus `feature/insights-queue-commands` at `edb0ba011` for the window clamp, and from `UiPath/Insights-monitoring` at `7ec27237d6`, whose queue files are byte-identical at the `develop` tip `6bbc79ee`.

The CLI carries named note constants in `commands/queues.ts` that ship as runtime `Instructions`, each written to stop an agent misreading a field, and the guide now agrees with them. Two show why that matters: `sla` is one row per queue name, so `ProcessName` is a representative pick and not half a key (`SnowQuery.cs:742`, `:776`), and `--timezone-offset` relabels `StartTime` and `EndTime` while bucket edges stay UTC-aligned (`SnowQuery.cs:261-262`).

The three smoke-task patterns were fixed by replaying the pinned grader, coder-eval 0.12.1, not by reading it. The `\bstart\b` negative was not anchored to `jobs` while this PR teaches `--time-event start`, so a correct queue answer false-failed a weight-2.0 criterion. Both negatives now name the forbidden shapes, and the barrier keeps the `\n` guard for the raw haystack while allowing backslash line continuations, the form two e2e tasks in this tree already use. 21 of 21 behaviours score as intended.

Every repo gate is green. The verb gate reports 0 blocking and 18 soft-stale findings, all of them queue verbs the pinned catalog does not carry yet, and all eleven `uipath-insights` task YAMLs validate against the pinned schema.

The new smoke task's twelve regexes were exercised against 74 crafted commands under `re.DOTALL`, the way the grader compiles them, with no misses and no false positives. That run caught a real defect before it shipped: `retry\b` matched the first six characters of `retry-outcomes`, so the write-verb guard would have failed every correct run. The terminator is `(?![\w-])` now. `npm run skills:test` is 56 of 56. `tests/scripts` is 270 passed and 44 skipped. Its one failure, `test_stage_shared.py`, is macOS-only and predates this branch: `stage_shared.sh` calls GNU `realpath -m`, which BSD realpath rejects. That file matches `main` byte for byte, no CI job runs its test, and CI runs the script itself on ubuntu.

## Limitations

**The queue smoke task ships skipped.** It drives eight of the ten commands and gates the command surface of all ten, but the catalog snapshot at `uip 1.203.0-dev.8703` carries `insights filter-queues` and no `insights queues` verb, so every `--help` gate fails today and would take the smoke check red on a PR whose skill content is fine. A skipped task lands in `RunSummary.skipped_tasks` at resolution time and never reaches the orchestrator (coder_eval 0.12.1, `models/tasks.py:466`, the version `tests/.coder-eval-version` pins), so it writes no `task.json` and stays out of the workflow's pass-rate denominator. Unskip in the PR that first runs against a dev build carrying the commands, which is what the four cli PRs above still hold up.

`failure-details` and `operational-metrics` are gated but not driven. The first needs an `--error-message` value read from a prior `failures-by-reason` row, which an empty tenant never produces, and the second needs a mandatory `--widget-type` that a prompt would have to supply, which grades prompt-following rather than the skill.

**The activation gate does not measure this skill, so the six queue rows are unexercised by CI.** `uipath-insights` is absent from `activation_gate.py`'s `BASELINES_PCT`, so the script prints `SKIP: no baseline for 'uipath-insights'` and returns 0. The gate did run on this head once the PR went ready, and that line is what it printed, so the green check is the skip rather than a pass. Twenty-two other skills have a baseline. Adding one needs a measured full activation run, which is why this PR does not add it.

The new queue triggers compete with `uipath-platform-018` and `uipath-troubleshoot-002`, and the gate only re-runs datasets whose own `SKILL.md` changed, so a loss there surfaces later on an unrelated PR. The `NOT for` clause and Critical Rule 11 draw the boundary; both datasets still need a run.

Rule 2 tells the agent to read `Instructions` for the window clamp, and on the cli side that field is currently wrong in a narrow band. All six queue and machine cli branches carry the "full 31 days old" note text, but only [#3826](https://github.com/UiPath/cli/pull/3826) carries the matching whole-day trigger; the rest test elapsed milliseconds, so a bound aged between 30 and 31 whole days reports a clamp that did not happen. Restacking those PRs onto #3826 fixes it, and no wording here works around it.

Combined frontmatter is 1528 of the 1536 truncation cap, so #2854 has to free space before adding machine triggers.

Based on `main` since #2609 merged. #2854, #3182, and #3358 stack on this branch and need a restack once it lands.


celestinryf:
Bot findings, answered.

The double blank line at `queue-monitoring-guide.md:128-129` is real and fixed in [`1e47fe84`](https://github.com/UiPath/skills/commit/1e47fe84c3481a4393e726e61e185ac7da06e14d). Lines 128 and 129 were both blank between the empty-`Data[]` paragraph and the one introducing `--time-event`. It was the only double blank in the file, and SKILL.md and the five sibling reference guides have none.

The frontmatter figure in the review is wrong. `description` is 559 characters and `when_to_use` is 969, so the combined value is 1528 against the 1536 cap, not ~1535 with one character of headroom. `hooks/validate-skill-descriptions.sh` prints the 559 on this head, and the Limitations section already gives 1528. Eight characters of room, and #2854 still has to free space before machine triggers go in.

On the task lint finding, no change here, which is what the lint itself recommends ("until then, no action needed"). A positive `command_executed` matching `uip\s+insights\s+queues\s+(summary|sla|operational-metrics)` would fail every run while `@uipath/cli@dev` carries no queue commands. It lands with the eval task once the four cli PRs merge and publish.

One Limitations edit alongside the fix: the PR came off draft at 22:48:58, so `activation-gate.yml:85` is no longer the reason the activation gate does not measure this skill. The gate ran on this head and printed `SKIP: no baseline for 'uipath-insights'`, which is the one reason left.

rliuup (approved):
LGTM

PR #2841, opened 2026-08-27
fix(evals): bake released Flow SDK into shared image

tmatup:
## Summary

- install the latest published `@uipath/flow-sdk` in the shared coder-eval image alongside the CLI and tool plugins
- pull and bake the connector catalog through `uip maestro registry pull`, with SDK and catalog provenance exposed to staged workspaces
- remove the separate source-built Maestro SDK Dockerfile and env overlay

## Validation

- full `tests/docker/Dockerfile` build passed and resolved `@uipath/flow-sdk@3.25.0`; its Flow source smoke check passed
- runtime workspace staging and Flow source check passed in the built image
- `uv run --with pytest pytest tests/scripts/test_stage_preview_sdk_workspace.py tests/scripts/test_stage_shared.py -q` (2 passed)
- `npm run skills:validate`
- `git diff --check`
- `npm run skills:test` (50/52 passed; two unrelated npm publish dry-run tests are rejected by the live registry because their fixture versions require an explicit tag)

🤖 Generated with Claude Code


bai-uipath (approved):
**Retarget to main and get a green smoke run on the branch, then good to go.** Deleting the source-built SDK image and its env overlay is the change we wanted, and moving connector library generation onto `uip maestro registry pull` gets it off hand-rolled scripts.

- **Retarget to main.** This stacks on the campaign branch, which only reaches main through the 185-file eval-corpus PR. The image change is independent of that work, and landing them together means any movement in flow task results can't be attributed to one or the other. Cut it against main directly.
- **Run smoke on the branch before merge.** It is the cheapest thing that exercises the new image end to end across the non-Flow tasks, which is what answers "does its presence break anything else."
- **Make the Flow steps best-effort rather than hard-fail.** Three new network-dependent steps land in the shared image build: the SDK install, the registry pull, and the maestro source check. As written, a Flow SDK or maestro CLI hiccup that today reds one custom image will instead red the shared image and therefore the entire nightly, every suite. The existing tool install already takes this shape with its minimum-tools floor. Fix: let all three fail soft and record what resolved, so an outage degrades the Flow tasks at runtime instead of blocking everyone's image.

**Fix if you agree, otherwise lgtm:**

- **Floating `latest` in a shared image.** A Flow SDK release now changes every task's image with no repo change, and paired comparison arms lose reproducibility. Consistent with how the CLI version floats, so not a blocker, but worth pinning for scheduled runs.
- **This changes what's under test, not just packaging.** Released 3.25.0 replaces the pinned remediation build, so prior preview results stop being a baseline for post-merge ones.
- **Minor:** `prepare-connector` is gone, worth grepping the preview skills and flow tasks for it; and the preview `SKILL.md` provenance pins now name a flow-builder-sdk snapshot that no longer builds anything in the image.

tmatup:
### Review response

**1. Retarget to `main`**

**Addressed** in follow-up PR #2843 (commit `4ba850f8c`), based directly on `main`. This keeps the shared-image change independent from the campaign eval-corpus work.

**2. Run branch smoke**

**Addressed.** [Smoke run 33035138793](https://github.com/UiPath/skills/actions/runs/33035138793) is green: the image build passed and the capped cross-skill sample finished at 100% (40/40). Docker/SDK-fixture changes now trigger this suite instead of being auto-skipped.

**3. Make Flow setup best-effort**

**Fixed** in `4ba850f8c`. SDK installation, registry pull, and Flow source checking now fail soft while writing requested/resolved SDK versions, catalog hash, and per-step status to `/opt/preview-flow-sdk-assets/setup-status.json`. A forced build with a nonexistent SDK version completed with SDK `unavailable` and downstream Flow setup `skipped`; the shared CLI/tools remained usable.

**4. Floating `latest` and changed baseline**

**Keeping `latest` intentionally** to match the agreed rolling CLI/package model. The status file records the concrete resolved SDK version and registry hash so each run remains attributable; the PR notes that 3.25.0 replaces the prior remediation build.

**5. `prepare-connector` and preview provenance**

**Keeping the current guidance.** A repository-wide search found no eval task invocation. The preview guide already defines `prepare-connector` as optional and environment-provided, with a curated-library fallback when absent. The `flow-builder-sdk` SHA identifies the synchronized documentation source, not the runtime package pin.

---

🤖 Generated with [Codex](https://openai.com/codex/)

PR #2843, opened 2026-08-27
fix(evals): make Flow SDK setup best effort

tmatup:
## Summary

Follow-up to #2841 and bai-uipath's review. This PR targets `main` directly so the shared-image change lands independently from the campaign eval corpus.

- install `@uipath/flow-sdk@latest` and materialize the connector registry in the shared image
- make SDK installation, registry pull, and Flow source checking best-effort; record requested/resolved versions, catalog hash, and step statuses in `/opt/preview-flow-sdk-assets/setup-status.json`
- treat Dockerfile and SDK smoke-fixture changes as smoke infrastructure changes, forcing the capped cross-skill smoke suite instead of skipping it
- keep `latest` intentionally aligned with the rolling CLI model while recording the concrete resolved SDK version

`prepare-connector` is not installed by the released package. A repository-wide check found no eval task invocation; the preview reference already marks it as optional and environment-provided. Its `flow-builder-sdk` SHA is documentation-source provenance, not a runtime image pin.

## Validation

- normal Docker build passed with SDK `3.25.0`, registry hash `f2ae49be22ec41059353d98dc668d3e924bb80a5c4cb357a5a1c44be928824fc`, and Flow source check `passed`
- forced-outage Docker build with a nonexistent SDK version passed and recorded SDK `unavailable`, registry `skipped`, and source check `skipped`; shared CLI/tools remained usable
- runtime assertions passed for both built images, including removal of the credential-bearing npm config
- `npm run skills:validate`
- `node --test tests/scripts/sync-maestro-sdk-preview.test.mjs`
- workflow YAML parse and `git diff --check`

🤖 Generated with Claude Code


bai-uipath (approved):
🛳️

dmetzgar (approved):

dmetzgar:
Resolved the conflict and pushed the merge (`aa2c6ff9d`). It was not mechanical — both sides had something the other needed.

#2756 landed the same SDK-install and registry-pull steps on `main` while this branch was open, in a **strict** form and with **dual-layout registry resolution** this branch's base predates. So the resolution keeps:

- **from this branch** — the best-effort semantics (an SDK/feed outage records a status instead of failing the image), `setup-status.json`, the `connectors` symlink and `FLOW_SDK_CONNECTORS_DIR`;
- **from main** — finding the snapshot at all. Up to maestro-tool `1.202.0-dev.8425`, `registry pull` wrote `current.json` under `$UIP_MAESTRO_REGISTRY_HOME`; from `dev.8461` (cli#3823) it delegates to the SDK and writes none. This branch read only `current.json`, so on a current CLI the lookup fails — and under best-effort that is **worse than main's hard failure**: the image goes green with no connector library and every connector-backed task fails at runtime behind a warning.

Verified against a live CLI rather than assumed:

```
$ uip maestro registry path --library-json
/Users/…/.uipath/cache/flow-sdk/library/current/library-json     # no current.json anywhere
$ ls <snapshot>
library-json   library-md
```

## One deliberate change to your logic, and why

`test -d "$registry_snapshot/connectors"` was a hard requirement for `registry_status=available`, and **the current layout does not produce that directory.** `registry pull` generates descriptors only for connectors a project references, and says so on an empty workspace:

```
registry pull: no connector referenced by this project yet, so no descriptors
were generated. Name one with `--connector <key>`, or author the step and pull again.
… generated 8035 markdown file(s) and 0 TypeScript module(s)
```

Left as-is that would have pinned `registry_status` at `failed` forever **and** taken the `library-json`/`library-md` symlinks down with it, since they share the `&&` chain — the image green, the library missing. It is now symlinked when present and never gates the library. Say the word if you'd rather it stayed a hard requirement and the layout question got solved differently.

## Verified

`docker build --check` no warnings · all five `RUN` bodies parse under `bash -n` (reassembled with continuations, comment lines dropped as Docker does) · both touched workflows valid YAML · the merged `INFRA_PATTERN` matches `tests/docker/` and `tests/fixtures/maestro-flow-sdk-smoke/` · `npm run skills:test` 55/55 · sync test 2/2 · `git diff --check` clean.

**Not verified:** an actual image build — it extends the GHCR agent base.

## Heads-up on ordering

This branch installs `@uipath/flow-sdk@${FLOW_SDK_VERSION}`, and since the rename that name is frozen at 4.0.0 (`@uipath/maestro-builder-sdk@5.0.1` is current). I deliberately did **not** change it here — #3172 owns the rename, and folding it in would make this PR do two things.

Consequence: merging this first leaves the shared image on the frozen 4.0.0 until #3172 lands, and #3172 touches the same block so it will need a conflict resolution. Happy to rebase #3172 onto this the moment it merges.

PR #2844, opened 2026-08-27
fix(uipath-maestro-case): bind group and role recipients on action tasks

Yiqi517:
## What

An action task reaches somebody only through `data.recipient` — the field the case engine turns into an Orchestrator task assignment. Two files in this plugin gave opposite instructions for the same SDD input.

`references/plugins/tasks/action/planning.md:56` said to leave the recipient off whenever sdd.md names a group or role, and to configure the assignment "separately via Actions app rules".

`references/plugins/tasks/action/impl-json.md:44` said `recipient.Type` `1` is a group ID, for exactly the SDD prefixes `UserGroup:` and `Role:`.

`impl-json.md:85` repeated the first instruction in its own verification list, so that file disagreed with itself as well.

A supplier-onboarding case built from an SDD that names a role on every human task came out with 11 of its 14 action tasks carrying no `data.recipient`. `uip maestro case validate` does not read the field, so the plan validated clean.

## Where the first instruction came from

It was correct when written. `fcec49183` (2026-04-17) introduced it alongside CLI authoring, where a task was created by a command whose recipient flag took an email and nothing else:

```
--recipient "<email>"

| `--recipient` | no | Omit for group/role assignment |
```

A group could not be expressed, so deferring it to manual configuration was the only option. `7e89db327` (2026-04-27) dropped CLI mutation for direct `caseplan.json` writes and `{ "Type": 1, "Value": … }` became expressible. Ten commits have touched this file since; none revisited the rule, and `impl-json.md`'s `Type 1` row was written for the new path without anyone noticing the two now contradicted each other.

## What the runtime actually accepts

The value is the group **name**. Three independent sources agree.

**The designer writes the name.** [`PeoplePicker.tsx:65`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/PeoplePicker.tsx#L65-L71) pushes `{ Type: GroupId, Value: <identityName> }` for a `LocalGroup` or `DirectoryGroup`, falling back to the identifier only when the name is absent. The user branch immediately below writes the identifier — a UUID. Groups and users differ here, so neither generalises to the other.

**The engine passes the value through untouched.** [`AppTasksCreateArgs.cs:187`](https://github.com/UiPath/PO.BpmnEngine/blob/bc836fe56ac2cbb97f7650404863e405f4fd42fc/src/BpmnEngine/UiPath.PO.Worker/Models/ProcessOrchestrationRuntime/AppTasksCreateArgs.cs#L187) deserialises the JSON straight onto `record Recipient(string Value, RecipientType Type)`. [`AppTasksFacade.cs:797`](https://github.com/UiPath/PO.BpmnEngine/blob/bc836fe56ac2cbb97f7650404863e405f4fd42fc/src/BpmnEngine/UiPath.PO.Worker/Facades/AppTasksFacade.cs#L797) derives the assignment criteria from `Type` alone, so the criteria string is optional. [`AppTasksFacade.cs:848`](https://github.com/UiPath/PO.BpmnEngine/blob/bc836fe56ac2cbb97f7650404863e405f4fd42fc/src/BpmnEngine/UiPath.PO.Worker/Facades/AppTasksFacade.cs#L848) hands `Value` to Orchestrator verbatim. Nothing on that path normalises a UUID into a name.

**Measured against the tenant.** One group, one call, three spellings:

| `AssigneeNamesOrEmails` value | Result |
|---|---|
| `Administrators` | assigned, task `Pending` |
| `administrators` | assigned, task `Pending` |
| `35551807-06b1-4cda-90a1-2fb84851eee7` | `ErrorCode 2482`, task `Unassigned` |

That UUID is what `uip admin groups list` returns for `Administrators`, so a lookup-then-write approach lands on exactly the value the runtime refuses.

## Why the old escape hatch never existed

Orchestrator's `swagger/v20.0` exposes assignment only as per-task commands — `AssignTasks`, `ReassignTasks`, `UnassignTasks`, `SaveAndReassignAppTasks`, `SaveAndReassignTask`. Neither `TaskDefinition` nor `TaskCatalog` carries an assignment-rule field. There is no app-level rule that routes a recipient-less task to a group.

And with no recipient the engine never reaches the assignment call at all — [`AppTasksFacade.cs:103`](https://github.com/UiPath/PO.BpmnEngine/blob/bc836fe56ac2cbb97f7650404863e405f4fd42fc/src/BpmnEngine/UiPath.PO.Worker/Facades/AppTasksFacade.cs#L103) returns before building the request.

## An assignment recipient is not an escalation recipient

The two shapes are near-twins and want opposite things:

| | SLA escalation | Task assignment |
|---|---|---|
| shape | `{ scope, target, value }` | `{ Type, Value }` |
| UUID | `target` | nowhere |
| name | `value` | `Value` |

Reusing the SLA plugin's group-to-UUID lookup here maps `Value` onto `target` rather than `value`. The rewritten section states the distinction so the analogy is not made again.

## The change

`Recipient Handling` is rewritten as three numbered branches — user email, group or role, silent — each with a stated outcome including the non-interactive path. A group or role is recorded as `recipient: UserGroup: <group name>`; the build strips the prefix and emits `Type 1` with the name. No lookup is involved.

The section also records the folder access a group needs before it can receive anything: it must exist in Orchestrator, and it must hold a role on the task's folder. A group created with `uip admin groups create` has neither. Without both the task is created and reaches nobody, which looks exactly like having no recipient, and `validate` sees neither.

## Verification

| Check | Result |
|---|---|
| `npm run skills:validate` | OK — default: 26 skills, 1729 files; studioweb: 26 skills, 174 replacements |
| `python3 scripts/check-skill-status.py` | OK — 26 skills, manifest valid |
| `hooks/validate-skill-descriptions.sh` | clean |
| Flavor overrides on either file | none |
| Live assignment, group name | assigned |
| Live assignment, group UUID | `ErrorCode 2482` |
| Remaining `Actions app rules` / `Resolve UserGroup` wording | none |
| Built caseplan, all 14 action tasks | `{"Type": 1, "Value": "<group name>"}` |

**An agent following the rewritten section writes the shape.** A coder-eval run of the supplier-onboarding task against a branch carrying this change produced every one of the fourteen action tasks with `data.recipient` set to `{"Type": 1, "Value": "<group name>"}`: no UUID, no surviving `UserGroup:` prefix, no dropped field. Run [33133032306](https://github.com/UiPath/skills/actions/runs/33133032306), artifact `SupplierOnboarding/SupplierOnboarding/caseplan.json`.

That run exercises `impl-json.md` and not `planning.md`. It finished in 250 seconds with a one-line `tasks/tasks.md`, so it reached the build without a planning pass, and all fourteen tasks carry the same group rather than the one each SDD row names. The shape is what the eleven unassigned tasks came down to and it is fixed; which group each task gets is decided in the planning branch and is not covered by this run.


Yiqi517:
Both reviews above predate the current head and describe the opposite rule. The later one was posted 2026-08-27T23:07Z; `95621a0ec` landed 2026-08-28T00:27Z, eighty minutes after. They read the branch as resolving a group to a directory UUID and recording `UserGroup:<uuid>`; the branch now records `recipient: UserGroup: Compliance` and states that this field holds a name and never a UUID. Neither approval covers what is in the diff today.

The first review's one manual-review item — whether the convention matches actual frontend and runtime behaviour — is the point that reversed the rule. It is settled in the description above with three independent sources: the designer writes the name, the engine passes `Value` through untouched, and a live assignment against one group accepts the name in either casing and rejects the directory UUID with `ErrorCode 2482`.

roberts-cliff (approved):

PR #2845, opened 2026-08-27
fix(planner,maestro-case): guard every =js: hop, and see the ones inside expressions

Yiqi517:
## What happened

A case built from an SDD faulted at runtime on its first agent task:

```
Failed to evaluate expression =js:("Registration certificate: " + vars.registrationCertificate.FullName + …)
Error: Cannot read property 'FullName' of null
ErrorCode: InputVariablesEvaluationError    BpmnExecution.cs:2254
```

`registrationCertificate` is a declared `file` In argument that the run did not supply, so it was null. The error text is the one `bindings-and-expressions.md` already predicts.

## Why the product does not hit this

Studio Web writes the guard for you. [`Tools.ts:806`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/utils/Tools.ts#L806-L819) rewrites a picked variable path into its optional-chaining form before it reaches the field: `vars.response.id` becomes `vars.response?.id`, `vars.arr[0].name` becomes `vars.arr?.[0]?.name`. It is idempotent, and it leaves the leading `vars.<name>` bare.

Nine call sites, across every surface where a person picks a variable: [`TipTapUtils.tsx:957`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/variable-picker/TipTapUtils.tsx#L957), [`:972`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/variable-picker/TipTapUtils.tsx#L972), [`TipTapEditor.tsx:456`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/variable-picker/TipTapEditor.tsx#L456), [`:473`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/variable-picker/TipTapEditor.tsx#L473), [`JsonEditorField.tsx:100`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/json-editor/JsonEditorField.tsx#L100), [`MonacoJsonEditor.tsx:255`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/json-editor/MonacoJsonEditor.tsx#L255), [`MonacoExpressionEditor.tsx:372`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/monaco-editor/MonacoExpressionEditor.tsx#L372), [`MonacoDiffEditor.tsx:138`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/components/controls/monaco-editor/MonacoDiffEditor.tsx#L138), [`UnifiedBuildVariablesProviderUtil.ts:30`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/services/uipath/variables/UnifiedBuildVariablesProviderUtil.ts#L30). An unguarded hop only reaches a case authored as a document.

[`Tools.ts:807`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/utils/Tools.ts#L807) gates the helper on `isVariablePrefix`, so it rewrites `vars.*` and nothing else — [`Tools.test.ts:1662`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/utils/Tools.test.ts#L1662-L1665) pins `iterator.item.name` and `notVars.foo.bar` as passing through untouched. The rules added here carry the same scope: the leading `vars.<name>` stays bare and every hop after it is chained. `response.*`, `metadata.*`, `bindings.*`, and `iterator.*` are unguarded on both sides.

## Why the skills did not stop it

**The design skill has no such rule.** `uipath-planner` teaches `=js:` in 21 places and matches zero times on `optional-chain`, `null-guard`, or `null-safe`. It authors the task-input expressions, so the unguarded form is born there.

**The build skill has the rule but never checks it.** `bindings-and-expressions.md` § Null-safe task-output references covers this, and Step 12 runs 15 checks, none of which look at it. Check 1 scans every `=js:` expression but only resolves the identifier — `vars.registrationCertificate` is a declared variable, so it passes, and nothing examines `.FullName`.

**Rule 6 pulls the other way.** "Preserve every Inputs row, binding mode, and value" tells the build to carry the sdd.md value through as authored. Nothing said which instruction wins.

**A second defect in the same file, found while reading it.** This one the checker could have caught and did not — it is a different failure from the crash above: the crash names a declared variable, and `sdd_check.py` checks declaration, so it would have passed the document either way. `sdd_check.py:417` anchored its reference scan on the `=vars.` prefix:

```python
refs = set(re.findall(r"=vars\.([A-Za-z]\w*)", text)) - {"X"}
```

References written inside a `=js:` string concatenation appear as `" + vars.X + "` — a space before `vars`, no `=`. On the SDD above that is 43 of 64 references never checked, and all four undeclared names were among them.

## The change

| File | Change |
|---|---|
| `skills/uipath-planner/references/case-design-layers-guide.md` | Layer 3 § Expressions states the rule for every `=js:` sink planner authors, with the null sources named |
| `skills/uipath-maestro-case/references/implementation.md` | Step 12 Check 16 scans every sink in `bindings-and-expressions.md` § Form by sink — connector `body` fields and trigger filters included, since a connector nests its inputs under `body` rather than `value` — says plainly that it outranks preserve-as-authored, and names what the guard trades away |
| `skills/uipath-maestro-case/SKILL.md` | Rule 6 names Check 16 as the one exception to preserving the value |
| `tests/tasks/uipath-maestro-case/_shared/sdd_check.py` | Reference scan matches both binding forms |
| `tests/tasks/uipath-planner/_shared/sdd_check.py` | The same edit — the two copies are kept byte-identical and `test_shared_twins.py` asserts it |
| `tests/tasks/uipath-maestro-case/_shared/test_sdd_check.py` | Three cases over the widened scan |

## Verification

Checker, against the SDD that faulted:

```
before   OK: sdd.md mechanically sound — 52 variables, 21 =vars references all resolve
after    mapping: 4 vars not declared: newExpectedDate, newExpectedDate2, newExpectedDate3, newExpectedDate4
```

Against the corrected SDD: `OK — 55 variables, 44 vars references all resolve`.

No regression: every `sdd.md` fixture under `tests/tasks/uipath-maestro-case/*/fixtures/` was run through the checker before and after. Mapping and lineage findings are identical on all ten — the fixtures that fail do so on template-shape and stage-count findings that predate this change.

- `pytest tests/tasks/uipath-maestro-case/ tests/tasks/uipath-planner/` — 188 passed, 14 skipped
- Restoring the `=vars.`-anchored pattern fails `test_reference_inside_a_js_expression_is_checked`, so the new cases guard the change rather than the fixture
- `npm run skills:validate` — OK, 26 skills, 1729 files
- `python3 scripts/check-skill-status.py` — OK, manifest valid
- `hooks/validate-skill-descriptions.sh` — clean
- No flavor override or flavor marker on any changed file
- Both new relative links resolve, and their anchors exist


Yiqi517:
On the prose-reference consideration: the conclusion holds, but not for the stated reason. SDDs authored through the template do carry prose `vars.` references — four across the eleven `tests/tasks/**/fixtures/sdd.md` documents, all in `supplier_onboarding`:

```
:641   **Recipient:** Expression:=vars.assignedBuyerEmail
:683   **Recipient:** Expression:=vars.assignedBuyerEmail
:724   **Recipient:** Expression:=vars.assignedBuyerEmail
:1955  (`vars.expectedAnnualSpend < 500000` skips it) and in its entry guard on the tier task's own
```

The last one sits inside an HTML comment, so a scan that keys on table cells would not have found it either way.

What keeps the widened pattern safe is that all four name a declared variable. The exposure is narrower than "prose mentions": it is a prose mention of a name absent from §Case Variables, and the corpus has none. A future SDD that names a renamed or example variable in prose would report it, and the finding would be a false positive rather than a lineage break — worth knowing before someone reads the reason above and assumes prose `vars.` cannot occur.

PR #2852, opened 2026-08-27
feat(uipath-api-workflow): teach files as JobAttachment references and the base64 activities

rares-baesu-uipath:
## Jira

https://uipath.atlassian.net/browse/SW-32087

## What

Teaches `uipath-api-workflow` how files work in API workflows and how to use Studio Web's **File to Base64** / **Base64 to File** activities. Pairs with UiPath/cli#3845 (CLI side: `--input-file` / `--output-dir` / `--folder-key`, base64-aware `validate`).

- **SKILL.md** — new critical rule 22: a file is a `JobAttachment` reference, never bytes; the two activities are `run.script` tasks recognised by `metadata.activityType` + the exact `$helpers.file.fileToBase64(...)` / `$helpers.file.base64ToFile({ base64, fileName?, mimeType? })` call; results are read as `$context.outputs.<Key>.output`; `serializeData()` only inline in an HTTP body / Response, and inside a JSON body only on a base64 reference (a binary ref's marker or a bare ref nested in a body fails at send time); both helpers need a signed-in run. Plus plan-table rows, triggers, reference/template rows, anti-patterns. Description kept ≤ 1024 chars.
- **New `references/files-and-base64.md`** (concepts, exact JSON, `serializeData()` rules, round trip, CLI run, limits/gotchas); sections added to `task-types.md` (§11/§12), `expressions-and-context.md`, `cli-reference.md`, `troubleshooting.md` (File & Base64 pitfalls).
- **Template** `file-base64-roundtrip-example.json` — verified end to end with the CLI branch (upload → encode → decode → download, bytes identical).
- **Studio Web flavor**: CLI-only passages live in `file-run-cli` / `file-inputs-cli` blocks, overridden with the host `RunProject` flow. `npm run skills:validate` / `skills:build` pass; built trees are marker-free.
- **Tests** (`tests/tasks/uipath-api-workflow/`): `base64/author_roundtrip` (integration, mode:build), `base64/run_roundtrip` (e2e, mode:operate, outcome-graded — re-runs the authored workflow with a fresh file and compares bytes), `diagnose/base64_helper_namespace` (integration, mode:diagnose, fixture drops the `.file.` namespace).

## Test run status (please read)

The three tasks are tagged **`path-to-ga`** because they cannot pass on the CI runner yet: `@uipath/cli@latest` predates the CLI change (`validate` rejects `FileToBase64` with `Unknown activityType`, and `--input-file` does not exist). They have **not** been run under coder-eval for that reason. What was verified instead:
- YAML parses, every `command_pattern` compiles, the embedded grading Python compiles, tags follow the taxonomy, `check-task-driver.py` passes.
- The diagnose fixture fails `validate` for exactly one reason and becomes `Valid` with the one-token fix (checked with the CLI branch build).
- The e2e grading command was executed by hand against a real tenant with the CLI branch build and passed.

Suggested order: land UiPath/cli#3845 → CLI release → run `make tags TAGS="path-to-ga uipath-api-workflow"` → merge this.

## Review

Two independent review passes. Fixed from them: YAML folding that broke the grading script (now a literal block + heredoc), two engine facts that were inverted (data-URL prefix / whitespace are *tolerated* by the engine; nested `serializeData()` only for base64 refs), unknown-size references also stream, `base64ToFile` on a binary ref is a documented no-op, regex edges in the tasks, neutral template description. Everything else in the docs was verified against the executor, the Studio Web translators and the CLI sources.


rares-baesu-uipath:
CI: fixed the `Build and inspect every skill package` failure (the flavor compose test caught CLI-only text — a templates-table row and two troubleshooting entries — leaking into the Studio Web tree; now flavored/reworded, `npm run skills:test` 52/52). The remaining red check, `path-to-ga approval`, is the policy gate for the `path-to-ga` tags: it needs an approval on the head commit from one of sherifmak, uisherif, bai-uipath, uipreliga, rockymadden.

skills/uipath-api-workflow/references/files-and-base64.md
```diff
@@ -0,0 +1,149 @@
...
+- **Namespace:** `$helpers.fileToBase64` (no `.file.`) fails with `is not a function` and fails `validate`.
+- **Invalid base64:** a `data:…;base64,` prefix and whitespace/line breaks are **tolerated** (stripped before decoding). What is rejected as `Invalid base64 string`: the URL-safe alphabet (`-` / `_`), other non-base64 characters, bad padding, and an empty string.
+- **Names:** `fileName` / `mimeType` never rename a reference input; a decoded text file loses its extension.
+- **Preview feature:** in Studio Web the two activities sit behind the `FE.EnableBase64Activities` flag and are marked "in preview". In the CLI the activities, `--input-file` / `--output-dir` / `--folder-key` and the base64-aware `validate` arrive together in the release that bundles executor 12.26+; earlier CLIs have none of them.
+- **Older CLI:** `Unknown activityType 'FileToBase64'` from `validate` (or `unknown option '--input-file'`) means the CLI predates this feature — upgrade it; the workflow itself is fine.
```
liviubobocu: i don't think this comment makes much sense, since this is added now as a feature

skills/uipath-api-workflow/references/files-and-base64.md
```diff
@@ -0,0 +1,149 @@
...
+  }
+}
+```
+
+`uip api-workflow validate` covers these activities offline: it accepts `FileToBase64` / `Base64ToFile` and rejects a task of either type whose script does not call its `$helpers.file.*` function.
+<!--skill-flavor:file-inputs-cli:end-->
+
+## 6. Limits and gotchas
+
+- **Size:** raw base64 *strings* and small references are converted in memory, capped at 50 MB per conversion; references with a declared `Metadata.Size` above 1 MB — or with **no** declared size, which is common for job inputs — stream with bounded memory and bypass the cap.
```
liviubobocu: I know I have tried with more than 50 Mb, even 100 MB, can we please double check the max size?

skills/uipath-api-workflow/SKILL.md
```diff
@@ -136,6 +137,12 @@ Do NOT use for: `.flow` Maestro flows (→ `uipath-maestro-flow`), `.xaml` / cod
...
+<!--skill-flavor:file-run-cli:start-->
```
liviubobocu, line 146: This is written chaotically. 
```
Never rename the helper (`$helpers.fileToBase64` without `.file.` fails at runtime with `is not a function`) — `validate` rejects a `FileToBase64` / `Base64ToFile` task whose script does not call its `$helpers.file.*`
```
I don't know if you can rename the helper in any way. The method itself is namespaced under file and it is somehow self explanatory that we should not call the method directly on $helpers.
This part also deserves some sections with each paragraph targetting a specific rule.

skills/uipath-api-workflow/SKILL.md
```diff
@@ -340,6 +351,8 @@ The mistakes an agent makes most often (each maps to a Critical Rule above — s
 - **Do NOT** wrap connector `bodyParameters` / `queryParameters` literals as `${'literal'}` — rule 5's wrap is **inverted** for connectors; bare literals only, or the field clears on save. See rule 16.
 - **Do NOT** ship a `<REPLACE_WITH_*>` placeholder in a workflow — StudioWeb renders it as a broken connection and it 401s. No pinged UUID → ask the user. See rule 16.
 - **Do NOT** read workflow inputs as `$input.<name>` from a non-first activity — use `$workflow.input.<name>`. See rule 13.
+- **Do NOT** treat a file input or a File to Base64 result as a string — both are `JobAttachment` references. Inline a file's content only with `<ref>.serializeData()` inside an HTTP body / Response, never in an Assign or script — and inside a JSON body field only on the File to Base64 output (`$workflow.input.document.serializeData()` nested in a body fails with "Raw bytes cannot be embedded in JSON"). See rule 23.
+- **Do NOT** write `$helpers.fileToBase64(...)` / `$helpers.base64ToFile(...)` — the helpers live under `$helpers.file.`; `validate` rejects the task and the runtime says `is not a function`. See rule 23.
```
liviubobocu, line 360: This Do Not section doubles down on the above not calling the method directly on $helpers global. I think it's worth having a single section for it.

tests/tasks/uipath-api-workflow/diagnose/base64_helper_namespace/base64_helper_namespace.yaml
```diff
@@ -0,0 +1,92 @@
+task_id: skill-api-workflow-diagnose-base64-helper-namespace
+description: >
+  Diagnose (file helpers): the workflow's File to Base64 task calls
+  `$helpers.fileToBase64(...)` — the `.file.` namespace is missing. `validate`
+  reports a single error at `.../run/script/code` ("does not call
+  $helpers.file.fileToBase64("); at runtime it would fail with `is not a function`.
+  Tests whether the skill teaches rule 22's exact helper shape and the
```
liviubobocu: rule 23 here

tests/tasks/uipath-api-workflow/base64/author_roundtrip/author_roundtrip.yaml
```diff
@@ -0,0 +1,103 @@
+task_id: skill-api-workflow-base64-author-roundtrip
+description: >
+  Skill-guided authoring of a file-handling API workflow: a `document` file input
+  is encoded with File to Base64 and decoded back with Base64 to File, and both
+  references are returned. Tests whether the skill teaches rule 22 — files are
```
liviubobocu: rule 23 here

skills/uipath-api-workflow/references/task-types.md
```diff
@@ -490,3 +492,81 @@ The simple single-expression form is fine; the designer corruption only affects
...
+  }
+}
+```
+
+For a file-reference input drop `fileName` / `mimeType`: `base64ToFile({ base64: $context.outputs.FileToBase64_1.output })`. Omit the optional keys entirely when unset — a dangling `fileName: ` is a syntax error.
+
+**Common mistakes:**
+- `$helpers.base64ToFile(...)` (namespace dropped)
+- Expecting `fileName` / `mimeType` to rename a reference input — they are ignored for references
+- URL-safe base64 (`-` / `_`), non-base64 characters, bad padding or an empty string → `Invalid base64 string` at runtime (a `data:…;base64,` prefix and whitespace are tolerated and stripped)
```
liviubobocu: The real message is "The provided value is not a valid base64 string: {{taskName}}"
Invalid base64 string is hallucinated. please check the other places where this is referenced as well.

skills/uipath-api-workflow/references/files-and-base64.md
rares-baesu-uipath: Fair point — removed it (and the matching troubleshooting entry about older CLIs).

skills/uipath-api-workflow/references/files-and-base64.md
rares-baesu-uipath: Double-checked against the executor source: references over 1 MB (or with unknown size) stream with no size cap — that's why your 100 MB test worked. The 50 MB cap only applies to in-memory inputs (raw base64 strings). Reworded the bullet to lead with that.

skills/uipath-api-workflow/SKILL.md
rares-baesu-uipath: Restructured rule 23 into per-topic sub-bullets and dropped the "never rename the helper" sentence.

skills/uipath-api-workflow/SKILL.md
rares-baesu-uipath: Agreed — removed the duplicate wording from rule 23, so the namespace mistake now lives only in this anti-pattern entry.

tests/tasks/uipath-api-workflow/diagnose/base64_helper_namespace/base64_helper_namespace.yaml
rares-baesu-uipath: Fixed.

tests/tasks/uipath-api-workflow/base64/author_roundtrip/author_roundtrip.yaml
rares-baesu-uipath: Fixed.

skills/uipath-api-workflow/references/task-types.md
rares-baesu-uipath: You're right — corrected to the exact message (verified in the executor's locales) here and in files-and-base64.md + troubleshooting.md.

liviubobocu (approved):

skills/uipath-api-workflow/references/troubleshooting.md
```diff
@@ -776,6 +776,57 @@ These are issues that surface only when a workflow is opened or run in **StudioW
...
+- **Symptom:** a marker object instead of file content
+- **Cause:** `serializeData()` was stored in a variable / returned from a script instead of being called inline in a body or Response field
+- **Fix:** call it only inside the HTTP body or `response` expression
+
+### Decoded file has no extension (`hello` instead of `hello.txt`)
+- **Symptom:** Base64 to File returns `FullName: "hello"`, `MimeType: application/octet-stream`
+- **Cause:** for a *reference* input the engine strips `.base64` and sniffs the MIME type from the bytes; plain text has no signature, and `fileName` / `mimeType` are ignored for references
+- **Fix:** acceptable for opaque payloads; when the name matters, decode a raw base64 **string** with `fileName` / `mimeType` (e.g. `{ base64: $context.variables.payload, fileName: 'notes.txt', mimeType: 'text/plain' }`)
+
+### `The provided value is not a valid base64 string: <task name>`
```
liviubobocu: <task name> is wrong (files-and-base64 §6, task-types §12, troubleshooting heading). All three invalidBase64Error(...) call sites pass BASE64_TO_FILE_FN, so the message is always literally …not a valid base64 string: base64ToFile. An agent greps for its activity key and misses — and one of the three is a troubleshooting heading, which is exactly what gets matched.

skills/uipath-api-workflow/references/files-and-base64.md
```diff
@@ -0,0 +1,148 @@
...
+Consequences:
+- A file input declared in `input.schema` (`"$ref": "#/definitions/job-attachment"`, `x-uipath-resource-kind: "JobAttachment"`) arrives as this object: `$workflow.input.document.FullName` works; there is no `.content`.
+- File bytes never enter `$context`, so context stays small no matter how big the file is.
+- Everything that reads or writes files goes through Orchestrator blob storage, so it needs a signed-in session.
+
+## 2. The two activities
+
+Both are ordinary `run.script` tasks. What makes them *these* activities — for Studio Web's designer, for `uip api-workflow validate`, and for the engine — is `metadata.activityType` plus the `$helpers.file.*` call in the script. Studio Web restores the property panel by parsing that call, so keep the exact shape.
+
+### File to Base64 (`FileToBase64`)
```
liviubobocu, line 42: The single-expression constraint is still undocumented — the one substantive gap. Shape matrix through the CLI head: multi_statement/trailing_stmt → Valid with a warning that SW "rebuilds the script … and drops the rest, which breaks the workflow". But fileToBase64(ref, {extra:1}) is Valid with no warning while SW drops the second argument. The skill is read before authoring, the warning only appears after — so for that case the skill is the only defense. It currently says just "keep the exact shape".

liviubobocu (approved):

skills/uipath-api-workflow/references/troubleshooting.md
rares-baesu-uipath: Fixed in 45e2f4dd4 — all three sites (files-and-base64 §6, task-types §12, troubleshooting heading) now carry the literal `…not a valid base64 string: base64ToFile`, with a note that the suffix is the helper name and never the task key, so grepping for the message matches.

skills/uipath-api-workflow/references/files-and-base64.md
rares-baesu-uipath: Documented in 45e2f4dd4. files-and-base64 §2 now states the constraint explicitly: the script is one `return` expression with exactly one argument; Studio Web rebuilds it from the parsed call on save and silently drops a preceding/trailing statement, a second argument (`fileToBase64(ref, { extra: 1 })`) or extra option keys. It also calls out that `validate` warns only for the extra statements and passes an extra argument as `Valid`, so the agent must check the shape before validating, and to move pre-processing into a JavaScript activity. Same point added as one sentence in rule 23, as a common-mistake bullet in task-types §11/§12, and as a new troubleshooting entry.

astridstaicu (approved):

PR #2853, opened 2026-08-27
fix(maestro-flow): raise validate criterion timeout to 180s

rockymadden:
## Problem

The `uip maestro flow validate passes on the flow file` success criterion runs under a 30s budget. That budget is inside the command's noise band, so valid flows score 0.0 on a gating criterion.

`uip maestro flow validate` refreshes the node manifest from the tenant on **every** invocation. `flow-tool`'s validate command hardcodes `validateCurrentManifests: true` when constructing `FlowValidateService`, which reaches `loadCurrentManifestMapUncached()` → `ManifestClient.getManifest()`:

```
[ManifestClient] fetchDynamicNodes ok: total=3304 ... orchestrator nodes=3277 IXP nodes=14
```

No `--governance` needed, not gated by `~/.uipath/maestro/registry.json`, and no offline flag exists. Wall time is network-bound and bimodal.

13 runs against one valid, unchanged flow:

| | seconds |
|---|---|
| typical | 8.3, 8.6, 9, 10.6, 11.2, 12.2, 12.6, 12.8, 14 |
| tail | **60.7, 61.4, 67.2** |

Roughly a quarter of runs exceed 30s. The floor alone eats 28% of the budget.

## Observed failure

`skill-flow-feet-inches`:

- criterion 1: `Exit code: -1`, `timed out after 30 seconds`, `Stdout: (empty)`
- criteria 2 and 3 (flow debug, 300s budgets): passed in 28s and 18s
- `weighted_score: 0.769`, `final_status: FAILURE`

The flow was fine. Running `uip maestro flow validate` on the preserved artifact returns `Result: Success, Status: Valid`, and the same run's agent had validated that identical file in 3.12s. The budget was below the command's variance, not above it.

## Why 30s

It was never chosen for this command. It is coder_eval's `run_command` default (`docs/TASK_DEFINITION_GUIDE.md`). #2213 then swapped the criterion from a bare CLI call to a python wrapper:

```diff
-    command: "uip maestro flow validate FeetInches/FeetInches/FeetInches.flow"
+    command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
     timeout: 30
```

adding interpreter start, a recursive `**/project.uiproj` glob, and a nested subprocess (~1-2s) without re-baselining. `tests/experiments/default.yaml` sets no per-criterion timeout, so there is no single-place fix; this is per-task.

## Change

54 task YAMLs, one line each:

```diff
   - type: run_command
     description: "uip maestro flow validate passes on the flow file"
     command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
-    timeout: 30
+    timeout: 180
```

- 53 from 30, plus `multi_node/billing_dispute_resolution` from 60, which is also under the observed 67s tail.
- The two e2e tasks already at 180 (`escalation_jira_ticket`, `customer_escalation_triage`) are untouched.
- 2.5x headroom over the worst measured run.

**Scope: `uipath-maestro-flow` only.** All 54 affected tasks, including the 6 under `context-grounding/`, `ixp/`, and `connector_features/`.

Those 6 sit in subfolders whose CODEOWNERS rules override the base maestro-flow rule, and with `requireCodeOwnerReview: true` each override is another required approval. That is why this PR requests 20 reviewers for a timeout bump. #2858 fixes the ownership gap by adding the flow team to those nested rules. Merge that first and this collapses to one flow-team approval after a re-push.

The 10 `uipath-human-in-the-loop` tasks calling the same shared script keep 30s and carry the same exposure. Different owner, worth a follow-up.

## Verification

Ran the criterion's exact command against the preserved `skill-flow-feet-inches` artifact from the failing run:

```
$ python3 tests/tasks/uipath-maestro-flow/_shared/validate_flow.py
Validating FeetInches/FeetInches/FeetInches.flow
{"Result": "Success", "Code": "FlowValidate", "Data": {"Status": "Valid"}}
exit 0
```

Two runs, 12.7s and 67.2s. Both exit 0, both inside the new 180s budget; the second would have been killed under the old one. The flow file is byte-identical to the one that scored 0.0.

Not claimed: a full `coder-eval` task run under the new timeout. This is a timeout-only change to a criterion whose command is unchanged, so the reproduction above covers the failure mode directly. Happy to add a full run if a reviewer wants it.

- `git diff -U0` removed lines are exactly `timeout: 30` (53) and `timeout: 60` (1); added lines are exactly `timeout: 180` (54). Nothing else moved.
- All 54 files parse; each has exactly one `validate_flow.py` criterion, all at 180.
- `python3 scripts/check-task-driver.py tests/tasks` passes.
- 6 of these files write the command unquoted, so a `validate_flow\.py"` anchored regex would have silently skipped 5 of them. The sweep used a quote-agnostic pattern and the count reconciles (56 referencing files, 2 already at 180, 54 changed).

## Not in scope

Follow-ups this leaves open, in priority order:

1. **Upstream CLI.** Add a request timeout inside `ManifestClient.getManifest` so a stall degrades to the warn-and-continue path that already exists (`"Could not load current OOTB node manifests; skipping stale-handle validation"`), or make the refresh cache-backed / opt-in. Schema validity should not need a tenant round trip. Once shipped, this budget can come back down.
2. **`_shared/validate_flow.py`.** It passes no `timeout=` to `subprocess.run` and has no retry, so the outer kill is the only guard and partial stdout is discarded (`coder_eval/sandbox.py` returns `(-1, "", err)` on `TimeoutExpired`). Its sibling `flow_check.run_debug` already does this right: inner cap below the criterion budget, 3 retries, 5s backoff. That asymmetry is why this failure reported empty stdout.
3. **HITL tasks.** Same script, same exposure, still at 30s.
4. **CODEOWNERS granularity.** Addressed in #2858: nested away-team rules under the flow test tree override the base rule, so mechanical sweeps fan out to every owner they touch.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/connector_features/generic_dynamic_node/generic_dynamic_node.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/generic_dynamic_node/generic_dynamic_node.yaml
@@ -48,7 +48,7 @@ success_criteria:
   - type: run_command
     description: "uip maestro flow validate passes on the flow file"
     command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
-    timeout: 30
+    timeout: 180
     expected_exit_code: 0
     weight: 3.0
     pass_threshold: 1.0
--- a/tests/tasks/uipath-maestro-flow/context-grounding/batch_transform/batch_transform.yaml
+++ b/tests/tasks/uipath-maestro-flow/context-grounding/batch_transform/batch_transform.yaml
@@ -52,7 +52,7 @@ success_criteria:
   - type: run_command
     description: "uip maestro flow validate passes on the flow file"
     command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
-    timeout: 30
+    timeout: 180
     expected_exit_code: 0
     weight: 3.0
     pass_threshold: 1.0
--- a/tests/tasks/uipath-maestro-flow/context-grounding/summarize/summarize.yaml
+++ b/tests/tasks/uipath-maestro-flow/context-grounding/summarize/summarize.yaml
@@ -54,7 +54,7 @@ success_criteria:
   - type: run_command
     description: "uip maestro flow validate passes on the flow file"
     command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
-    timeout: 30
+    timeout: 180
     expected_exit_code: 0
     weight: 3.0
     pass_threshold: 1.0
--- a/tests/tasks/uipath-maestro-flow/e2e/devcon_expense_approval.yaml
+++ b/tests/tasks/uipath-maestro-flow/e2e/devcon_expense_approval.yaml
@@ -77,7 +77,7 @@ success_criteria:
   - type: run_command
     description: "Flow validates successfully"
     command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
-    timeout: 30
+    timeout: 180
     expected_exit_code: 0
     weight: 3.0
     pass_threshold: 1.0
--- a/tests/tasks/uipath-maestro-flow/edit/add_node/add_node.yaml
+++ b/tests/tasks/uipath-maestro-flow/edit/add_node/add_node.yaml
@@ -25,7 +25,7 @@ success_criteria:
   - type: run_command
     description: "uip maestro flow validate passes on the flow file"
     command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/validate_flow.py"
-    timeout: 30
+    timeout: 180
     expected_exit_code: 0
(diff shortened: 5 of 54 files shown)


nikhil-maryala (approved):

rockymadden:
Trimmed to 48 files. The 6 tasks under `context-grounding/`, `ixp/`, and `connector_features/` moved to #2857 — each of those directories overrides the base maestro-flow CODEOWNERS rule, and with `requireCodeOwnerReview` that turned a one-line-per-file sweep into 4 required approvals across 20 reviewers. This PR now needs one approval from `/tests/tasks/uipath-maestro-flow/`.

rockymadden:
Reverted the split, this is back to all 54 files. Closed #2857.

Root of the reviewer fan-out is that `context-grounding/`, `ixp/`, and `connector_features/` override the base `/tests/tasks/uipath-maestro-flow/` CODEOWNERS rule rather than extending it, so the flow team has no review rights on 6 files in its own test tree. #2858 adds the flow team to those nested rules, additively.

CODEOWNERS is read from the base branch, so #2858 needs to land first; after that a re-push here collapses this to one flow-team approval.

PR #2854, opened 2026-08-27
feat(uipath-insights): teach machine monitoring [IN-13530]

celestinryf:
https://uipath.atlassian.net/browse/IN-13530

Also related to https://uipath.atlassian.net/browse/IN-13527: the activation rows here are that ticket's skills-side test layer.

The commands ship in two stacked cli PRs: [UiPath/cli#3849](https://github.com/UiPath/cli/pull/3849) (runtime-mix, details) and [UiPath/cli#4111](https://github.com/UiPath/cli/pull/4111) (availability-timeline, top-failures, utilization). This half waits for the second to merge and publish `@uipath/cli@dev`, which the smoke gate installs. Stacked on #2839 (queue monitoring); the diff against that branch is the three files below.

## What

Teaches the `uipath-insights` skill the five machine monitoring commands.

- New `references/machine-monitoring-guide.md` with the command catalog, shared flags, the error table, and the per-command output shapes.
- `SKILL.md` generalized from the queue family to cover machines. `description` and `when_to_use` are 551 and 943 characters, so the combined listing text is 1,494 against the 1,536-character cap. Critical Rule 5 also names where a machine investigation starts, since the family has no `summary` command.
- Four activation rows in `tests/tasks/activation/uipath-insights.jsonl`.
- New `tests/tasks/uipath-insights/machines/smoke.yaml`: all five commands driven from a prompt that names none of them, a `--help` reality gate per subcommand, and negatives for the guide's own traps. Ships skipped, see below.

## Why

WS2.2 M1 adds machine monitoring to the CLI. Agents need the matching guide to pick the right command and to read the outputs that are easiest to misreport.

## Verification

Every behavioral claim was read from the CLI on the two branches above and from `MachinesQuery.cs` in `UiPath/Insights-monitoring` at `e17accd8`, rather than inferred from field names. Six claims changed as a result.

`CurrentProcess` carries the last job event's process whatever state that job ended in, so a `Status` of `In Use` is the only running-now signal. The CLI ships that sentence in its own `Instructions`.

The `details` row spans three windows rather than two. Only the non-`In Use` `Status` values come from the fixed 30-day session scan, and `Available` is the query's fallback when no session row exists at all.

Two filters reach less than their names suggest: `--host-machine-name` narrows only the job-derived columns of `details`, and `--machine-type` never reaches `availability-timeline`'s job markers.

An absolute window whose end is also past the 30-day cap collapses to about a millisecond and returns an empty result that is an artifact of the clamp, not an absence of data.

The `N/A` serverless host placeholder cannot be fed back to `--host-machine-name` on `top-failures` or `utilization`, which compare the raw column.

`AverageUtilizationMs` divides by a robot count spanning both execution types, so the Unattended row is not a per-unattended-robot average.

The Errors section carries its own table rather than pointing at the queues guide, since SKILL.md tells an agent to read only the guides its task needs.

Ran locally, all passing on both flavors (27 skills): `npm run skills:validate`, `npm run skills:build`, `npm run skills:check-links`, `npm run skills:test` (56 tests), `hooks/validate-skill-descriptions.sh`, `python3 scripts/check-skills-sh.py`, and `npm run version:check`.

CI's `Run skill smoke tests` is green on this branch and it does not cover this change. The eight tasks it runs are the existing ones, and none of them names a machine command, `smoke_all_commands.yaml` included.

**The new machine smoke task ships skipped**, which is why the suite is still eight. The catalog snapshot at `uip 1.203.0-dev.8703` carries `insights filter-machines` and no `insights machines` verb, so every `--help` gate fails today and would take the smoke check red on a PR whose skill content is fine. A skipped task lands in `RunSummary.skipped_tasks` at resolution time and never reaches the orchestrator (coder_eval 0.12.1, `models/tasks.py:466`, the version `tests/.coder-eval-version` pins), so it writes no `task.json` and stays out of the workflow's pass-rate denominator. Unskip in the PR that first runs against a dev build carrying the commands, which needs both cli PRs above to merge.

The task validates against that pinned `TaskDefinition` schema, and its twelve regexes were exercised against 55 crafted commands under `re.DOTALL`, the way the grader compiles them, with no misses and no false positives.

The four activation rows are unexercised, and the reason is worth recording. `activation-gate.yml:85` skips the gate job on a draft PR, but that is not the only barrier: `uipath-insights` is absent from `BASELINES_PCT` in `tests/scripts/activation_gate.py`, so the script prints `SKIP: no baseline` and exits 0 even on a ready PR. A manual `workflow_dispatch` against this branch confirmed it. Twenty-two other skills have a baseline, so every green Skill activation gate this skill has ever shown was that skip. Adding a baseline needs a measured full activation run.

### DIFF
--- a/skills/uipath-insights/SKILL.md
+++ b/skills/uipath-insights/SKILL.md
@@ -1,13 +1,13 @@
 ---
 name: uipath-insights
-description: "UiPath Insights read-only monitoring via `uip insights`: job metrics, failure analysis, and process performance; queue totals, SLA risk, timelines, and failure drill-down; filter discovery of active folders, processes, queues, and machines; alert definition, history, delivery, and entitlement reads; Insights user, role, and group reads. Alert and RBAC writes are not covered. For Orchestrator job start/stop/logs→uipath-platform, root-cause analysis of specific errors→uipath-troubleshoot, RPA workflow authoring→uipath-rpa, org-level identity→uipath-admin."
-when_to_use: "User says 'job failures', 'automation health', 'job success rate', 'processing time', 'which processes fail the most', 'failure reasons', 'job trends', 'how many jobs ran', 'job metrics', 'uncompleted jobs', 'faulted jobs', 'job timeline', 'process details', 'which folders can you see', 'what processes are active', 'find the folder key', 'discover queues', 'which machines reported', 'filter-folders', 'filter-processes', 'queue backlog', 'queue SLA', 'why are queue items failing', 'queue retry success', 'Insights alert', 'which alerts exist', 'alert history', 'did an alert fire', 'alert delivery', 'alerting entitlement', 'who has Insights access', 'Insights roles', or 'Insights groups'. NOT for alert or RBAC writes, starting/stopping jobs or queue item CRUD (uipath-platform), or root-cause debugging of a specific job error (uipath-troubleshoot). Also 'uip insights', 'insights jobs', 'insights queues'."
+description: "UiPath Insights read-only monitoring via `uip insights`: job metrics, failure analysis, and process performance; queue totals, SLA risk, timelines, and failure drill-down; machine runtime mix, availability, fault ranking, and runtime minutes; filter discovery of active folders, processes, queues, and machines; alert definition, history, delivery, and entitlement reads; Insights user, role, and group reads. Alert and RBAC writes are not covered. For job start/stop/logs→uipath-platform, root-cause analysis→uipath-troubleshoot, workflow authoring→uipath-rpa, org-level identity→uipath-admin."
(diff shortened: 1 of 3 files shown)


celestinryf:
One fix and four verdicts.

Rule 5 now covers machines, in [`3c09f0ac`](https://github.com/UiPath/skills/commit/3c09f0ac6). It says the family has no summary, then names `machines runtime-mix` for the tenant-wide runtime split and `machines details` for a named machine's current state.

Not adding a `top-failures` activation row. Every row in `tests/tasks/activation/uipath-insights.jsonl` carries `expected_skill` and nothing else, so the file asserts which skill activates and never which command runs. Row 016 already routes machine fault ranking here, and a fifth row would assert the same thing a fifth time.

Both cli PRs are still open, [#3849](https://github.com/UiPath/cli/pull/3849) and [#4111](https://github.com/UiPath/cli/pull/4111), so `@uipath/cli@dev` does not carry the machine commands yet.

That is what the smoke check turns on. `Run skill smoke tests` is green here because no task under `tests/tasks/uipath-insights/` names a machine command, `smoke_all_commands.yaml` included, so the pass reaches none of the five commands. The PR body said the suite was not run, which was wrong in the other direction; it now says the suite ran and what it misses.

The `AverageUtilizationMs` claim holds against the source. In `RealtimeMonitoring/RealTimeDashboards/src/DataAccessor/Queries/MachinesQuery.cs`, the `ROBOT` CTE at lines 84 to 86 is `COUNT(DISTINCT ROBOTNAME) ... WHERE TYPE IN (0, 1)`, and lines 107 to 108 divide both `UNATTENDEDAVGUTIL` and `ATTENDEDAVGUTIL` by that one cross-joined count. The Unattended row averages over attended and unattended robots together, as the guide says.

The baseline point is unchanged and stays in the PR body. `uipath-insights` is absent from `BASELINES_PCT`, which holds 22 other skills, so the gate skips whatever these rows say.

One correction to the body while I was in it: the combined frontmatter text is 1,494 characters against the 1,536 cap, not the 49 characters of headroom it claimed.

rliuup (approved):
LGTM

PR #2858, opened 2026-08-27
chore(codeowners): own flow and BPMN via @UiPath/maestro-cli-team

rockymadden:
Unblocks #2853 and every future flow or BPMN sweep.

## Problem

CODEOWNERS matching is **last-match-wins, not additive**. A subfolder rule *replaces* the base rule rather than extending it. The file already flags this on line 97:

```
# Flow test tasks (general) — more specific subfolder rules below override this
```

Nine subfolders under the flow skill and its test tree are owned by away teams, so the flow team had no review rights on its own tree there. `main` enforces `requiredApprovingReviewCount: 1` with `requireCodeOwnerReview: true`, and GitHub requires an approval **per matching rule**, so any sweep across flow paths becomes one required approval per away team.

Concretely: #2853 is a one-line-per-file criterion timeout bump across 54 task YAMLs. Six files fell into these subfolders, which turned it into **4 required approvals and 20 requested reviewers** for a change that alters no task logic, prompt, criterion shape, or weight.

## No wildcard can fix this

Worth stating explicitly, since it is the obvious first instinct: there is **no wildcard that adds owners**. A wildcard can only replace them. `/skills/uipath-maestro-flow/** @team` placed after the subfolder rules would win for every flow file and strip the away teams' ownership entirely. That is a different and much bigger change than intended.

So each rule has to carry the team.

## Change

All 17 flow and BPMN rules now list `@UiPath/maestro-cli-team`, and every enumerated individual who is a team member collapses into that handle.

```diff
-/tests/tasks/uipath-maestro-flow/ @bai-uipath @rockymadden @gozhang2 @tmatup @akshaylive @jiyangzh @nikhil-maryala @baishalighosh @DevMomo
+/tests/tasks/uipath-maestro-flow/ @UiPath/maestro-cli-team

-/tests/tasks/uipath-maestro-flow/ixp/ @UiPath/communications-mining @constantinmuraru @vlad-crisan
+/tests/tasks/uipath-maestro-flow/ixp/ @UiPath/communications-mining @constantinmuraru @vlad-crisan @UiPath/maestro-cli-team

-/tests/tasks/uipath-maestro-bpmn/ @bai-uipath @rockymadden @gozhang2 @tmatup @akshaylive @jiyangzh @nikhil-maryala @baishalighosh @DevMomo
+/tests/tasks/uipath-maestro-bpmn/ @UiPath/maestro-cli-team
```

Coverage:

| Area | Rules |
|---|---|
| `/skills/uipath-maestro-flow/` incl. plugin guides (summarize, batch-transform, connector, connector-trigger, agent, inline-agent, ixp) | 9 |
| `/tests/tasks/uipath-maestro-flow/` incl. context-grounding, connector_features, datafabric_connector, ixp | 4 |
| `/skills/uipath-maestro-bpmn/`, `/tests/tasks/uipath-maestro-bpmn/` | 2 |
| `/preview/uipath-maestro-flow/`, `/preview/uipath-maestro-bpmn/` | 2 |

Six rules collapse to the handle alone (no away owners): all three BPMN rules, `/preview/uipath-maestro-flow/`, and `/tests/tasks/uipath-maestro-flow/`.

**Additive on the eleven shared rules.** Every existing owner keeps review rights and away teams are still auto-requested on changes to their own plugin guides and tasks. Nobody is removed from a path they own today.

Using a handle instead of a roster also stops the drift: the 9 names were duplicated across 8 lines and went stale whenever someone joined or left.

## Deliberately out of scope

Several other rules name a `@UiPath/maestro-cli-team` member individually — `song-zhao-25` on planner and maestro-case, `dushyant-uipath` on human-in-the-loop, `baishalighosh` on integration-service, `tmatup` on `/commands/` and preview-case, `bai-uipath`/`akshaylive` on `/.github/`.

Those are **left alone on purpose.** Being on this team is incidental there; those people own those paths in a different capacity, and collapsing them to the handle would hand this team ownership of another team's area. "Team member listed individually" is not the same as "should be the team handle."

## Repo access

A CODEOWNERS entry naming a team **without repo access is silently ignored** — no error, the rule just does not apply. `@UiPath/maestro-cli-team` was not a collaborator here, so it has been granted `push`, matching every other team already used in this file (`communications-mining`, `team-coded-agents`, `team-merlot`, `team-orange`, `DatafabricCodingAgent`, `llmops-team`, all `push`).

Confirmed working: the team now resolves as a review request on this PR.

## Pre-merge item

`@bai-uipath` and `@akshaylive` are named in the current flow and BPMN rules but are not yet members of `@UiPath/maestro-cli-team`. They must be added before merge, or they lose review rights on:

- `/skills/uipath-maestro-flow/`, `/tests/tasks/uipath-maestro-flow/`
- `/skills/uipath-maestro-bpmn/`, `/tests/tasks/uipath-maestro-bpmn/`
- `/skills/uipath-maestro-flow/references/plugins/connector/` and `connector-trigger/` (bai-uipath)

Every other individual this PR removes is already a team member, verified against the live roster. @rockymadden is handling the membership add.

## Verification

- Rule count unchanged at 118; none added or removed.
- All 54 files changed in #2853 resolve to a rule owning `@UiPath/maestro-cli-team`. Before: 48 of 54.
- Diffed the before/after owner set for every rule in the file: the only owners losing rights anywhere are the two named above.
- No flow or BPMN rule still names a team member individually, checked against the live roster.
- 17 insertions, 17 deletions, CODEOWNERS only.

## Sequencing

GitHub evaluates CODEOWNERS from the **base** branch, so this does not retroactively change #2853's requirements. Merge this first, then #2853 needs a push to re-evaluate and collapses to one `@UiPath/maestro-cli-team` approval.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -18,9 +18,9 @@
 /skill-flavors/studioweb/ @astridstaicu
 
 # preview skills
-/preview/uipath-maestro-flow/  @dmetzgar @tmatup
+/preview/uipath-maestro-flow/ @UiPath/maestro-cli-team
 /preview/uipath-maestro-case/ @dmetzgar @tmatup
-/preview/uipath-maestro-bpmn/ @dmetzgar @tmatup
+/preview/uipath-maestro-bpmn/ @UiPath/maestro-cli-team
 
 # Activity references (bundled inside uipath-rpa skill)
 /skills/uipath-rpa/references/activity-docs/ @DragosUnguru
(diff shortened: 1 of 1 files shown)


rockymadden:
Reworked per review: this now uses `@UiPath/maestro-cli-team` instead of enumerating individuals, and covers **all 14** flow rules rather than just the 4 test rules that blocked #2853.

Two things worth calling out for reviewers:

1. There is no additive wildcard in CODEOWNERS. A wildcard can only replace owners, so `/skills/uipath-maestro-flow/** @team` at the end would strip the away teams. Each rule has to carry the team, which is what this does.
2. The team had no repo access, and a CODEOWNERS entry for a team without access is silently ignored. It has been granted `push`, matching every other team in this file.

Pre-merge: `@bai-uipath` and `@akshaylive` need adding to the team or they lose rights on 4 rules.

rockymadden:
Condensed further: `@dmetzgar` is a team member so `/preview/uipath-maestro-flow/` collapses to the handle alone. Checked all 14 flow rules against the live roster — no rule still names a team member individually.

`/skills/uipath-maestro-bpmn/` and `/tests/tasks/uipath-maestro-bpmn/` each list the same 9 individuals, all team members, and could collapse the same way. Left out to keep this PR scoped to flow.

rockymadden:
Added BPMN. All three BPMN rules had zero away owners, so they collapse to the handle alone.

Now 17 rules across flow, BPMN, and preview. Full-file diff check: rule count unchanged at 118, and the only owners losing rights anywhere are `@bai-uipath` and `@akshaylive` pending their team membership.

One thing I deliberately did **not** touch: `song-zhao-25` (planner, maestro-case), `dushyant-uipath` (HITL), `baishalighosh` (integration-service), `tmatup` (`/commands/`, preview-case) are all on this team but own those paths in a different capacity. Collapsing them would hand us ownership of other teams' areas.

nikhil-maryala (approved):

PR #2864, opened 2026-08-27
feat(tests): route checker LLM through litellm gpt-5.6-luna via coder-eval 0.11.5

akshaylive:
Bump coder-eval to 0.11.5 and set checker_context.api_route defaults (default/smoke/smoke-windows/nightly/activation experiments) so the success-checker/judge calls litellm -> gpt-5.6-luna instead of the built-in default, resolving credentials from CODEX_BASE_URL/CODEX_API_KEY.


bai-uipath (approved):
**LGTM on the direction, but this has never run e2e against a real task.** CI died on a missing import before a single judge call went over the wire.

- **Run one real task e2e before merge.** `default.yaml` is tempdir, so the checker runs on the host: install the extra locally and run the data-fabric choiceset task that errored in CI, plus one simulation task to confirm the reject. Nothing else proves the call shape, and the bare `gpt-5.6-luna` id carries no provider prefix where the upstream docs do.

- **The litellm extra is installed nowhere, and this repo can't fix it.** Under the docker driver the checker runs in-container, so the agent image needs it, as do the coder_eval_uipath host venvs and the skills workflow installs. Fix: `--extra litellm` in coder_eval's agent Dockerfile, release, re-pin.

- **Every `simulation.enabled` task hard-errors at route resolution.** The orchestrator rejects a litellm eval route alongside simulation or `agent_judge`, so 318 of 1231 tasks die before any criterion runs, roughly 30% of the suite counted as misses. Fix: scope it at the task layer to the 54 llm_judge tasks without simulation.

Minor: `api_version` is verbatim from the upstream doc example and `CODEX_API_VERSION` is set nowhere in the pipelines; the activation experiment has no llm_judge criteria, so its block is dead.

uipreliga:
## Code review

Reviewed with three models in parallel (Gemini 3, GPT-5.6, Opus). The headline finding below I verified independently against the `coder_eval` source at tag `v0.11.3` and against the full task corpus.

### 🔴 Critical — blocking

**`route: litellm` is `llm_judge`-only, but it's applied as a blanket experiment default → 318 tasks hard-ERROR at setup.**

coder-eval 0.11.3 rejects this combination on purpose. `src/coder_eval/orchestrator.py:1508` `_reject_litellm_eval_route_if_unsupported()` is called unconditionally from `_resolve_routes()` → `_setup`, for **every** task, whether or not it has an `llm_judge` criterion:

```python
if self.task.simulation is not None and self.task.simulation.enabled:
    offenders.append("simulation.enabled")
if offenders:
    raise ValueError("checker_context.api_route.route: litellm is llm_judge-only …")
```

The guard is present at tag `v0.11.3` (not just on `main`). Scanning `tests/tasks/`:

| tier | tasks with `simulation.enabled: true` |
|---|---|
| `e2e` | 303 |
| `integration` | 14 |
| `smoke` | **1** |
| **total** | **318** |

`agent_judge` is not a problem — there are zero `agent_judge` criteria in the repo.

Blast radius: all 318 error under `nightly.yaml` (the Linux `run-coder-eval.yml` job) and under `default.yaml` (all local `make e2e` / `make all`). The PR gate is hit too — `tests/tasks/uipath-maestro-flow/interactive/solution_select.yaml` is tagged `smoke` **and** sets `simulation.enabled`, so it will ERROR in `smoke-skills.yml` the next time a maestro-flow change selects it. This PR's own CI won't surface it, since it changes no skill source.

The route needs to be scoped rather than defaulted — per-task `checker_context`, a variant-level override, or a separate experiment file.

### 🟠 High

**1. `smoke-rpa-skills.yml` runs `smoke-windows.yaml` but got neither the extra nor the credentials.** (flagged independently by all three reviewers)

- `:138-140` installs `coder-eval==<ver>` — no `[litellm]`
- `:291-315` run-step `env:` exports `API_BACKEND`/`AWS_*`/`BEDROCK_MODEL`/`ANTHROPIC_API_KEY`/`AGENT_MODEL` — no `CODEX_API_KEY`, no `CODEX_BASE_URL`
- `:363` runs `-e experiments/smoke-windows.yaml`, which is `driver: tempdir` → the checker runs on the host runner

The sibling Windows job in `run-coder-eval.yml` got both (`:634-640`, `:864-865`). Mitigating: no `windows`-tagged task currently has an enabled `llm_judge`, so the route resolves but never dispatches — latent, and it fires the moment one is added.

**2. Local development hard-fails, with no fallback.** `default.yaml` is the documented "devs locally / ad-hoc runs" config; `tests/Makefile:18,22` install coder-eval from git with **no extras**. There is no graceful degradation: `judge_litellm.py:79-87` raises `JudgeInfrastructureError` on the missing package, `:105-112` raises on an unset **or empty** env var, and `errors/judge.py` notes these are deliberately *not* converted to a scored-0.0 criterion — the row lands as `FinalStatus.ERROR`. `tests/README.md` has no mention of `CODEX_*`, `checker`, `judge`, or `litellm`, and this PR adds none.

**3. `model: gpt-5.6-luna` carries no provider prefix, while `api_version` is Azure-shaped — one of the two is wrong.** `judge_litellm.py` passes the model id verbatim (its own test comment: *"Model id travels verbatim, provider prefix and all"*). A bare `gpt-5.6-luna` resolves to provider `openai`; the framework's canonical examples are `azure/gpt-5.6-luna` (`docs/TASK_DEFINITION_GUIDE.md:1322`) and `azure_ai/gpt-5.6-luna` (`judge_litellm.py:16`, `tests/test_judge_litellm.py:67`). Under provider `openai`, `params.api_version: "2024-05-01"` is inert config that reads as though it does something.

There is a real counter-argument: `pricing.py:90` keys the rate card on the **bare** name and `_normalize_model` does not strip `azure/`, so a prefixed id would report the judge as unpriced. If that's the reason for the bare name, it deserves a comment — and `api_version` should go. Either way this needs **one live judge invocation to settle**; `CODEX_BASE_URL`'s wire shape isn't knowable from this repo.

### 🟡 Medium

- **`activation-gate.yml` — same gap, and here it's pure drift.** `:101` installs without `[litellm]`; `:104-117` exports no `CODEX_*`. `activation.yaml` declares no `sandbox:` block, and `models/sandbox.py:404` confirms `tempdir` is the default → host runner. Activation tasks use only `skill_triggered` criteria (zero `llm_judge`, zero simulation), so the block buys nothing there. Either wire it or drop it from `activation.yaml`.
- **The extra was added where it isn't used, and omitted where it is.** Under `driver: docker` the orchestrator runs in-container (`models/sandbox.py:165-166`), so the host extra is never imported by the checker. `run-coder-eval.yml:271-274` (docker), `smoke-skills.yml:359` (docker) and `smoke-skills.yml:721` (`coder-eval plan` only — never runs a checker) are all redundant, while the two tempdir jobs that genuinely need it were missed. Harmless as belt-and-braces, but the inverted pattern is what hid the two real gaps.
- **`tests/docker/Dockerfile:146` — `pip install --no-cache-dir litellm` is unbounded.** coder-eval's own extra constrains it to `litellm>=1.95.0,<2.0.0`; a rebuild of the same commit can pull a 2.x the framework excludes, silently changing judge behaviour. Two suggestions that came up and that I'd explicitly **discard**: bare `pip` is fine here (base is `python:3.13-slim` + `uv pip install --system`, same site-packages — same precedent as `uipath-langchain` on line 137), and `pip install "coder-eval[litellm]==<ver>"` would 404, since the Dockerfile configures no PyPI extra-index and coder-eval lives on the Azure Artifacts feed. A bounded `litellm` range is the right expression of intent.

### 🟢 Low

- **`same-ground-headtohead.yaml` didn't get the block — and it's a head-to-head comparison arm.** Its counterpart moves judges; it doesn't. That's an experiment-validity question rather than tidiness, so worth an explicit answer either way. (`flow-v2-preview.yaml` was also skipped, but it declares no `checker_context`, points at `skills-codex:latest` which no workflow in this repo builds, and isn't in CI — moot today.)
- **`smoke-skills.yml:473-475` is now factually wrong**: *"that is the evaluation-side model (llm_judge + the simulated user), which stays on Claude on purpose — the grader must not move with the agent."* After this PR `llm_judge` runs on `gpt-5.6-luna` via the codex agent's own endpoint and credential (`:466-467`). If `vars.CODEX_MODEL` is also `gpt-5.6-luna`, the gate self-grades. Worth an explicit note in the PR description. (`:452-457`, the ANTHROPIC_API_KEY / final-reviewer comment, is a genuinely separate path — leave it.)
- **DRY — no action needed.** I checked: `ExperimentDefinition` has no `extends`/`include`/`base`, and the 4-layer merge in `orchestration/experiment.py:176-217` starts from `default_experiment.yaml` *inside the wheel*, which isn't repo-authorable. Duplicating the block across the 5 files is the correct call; please don't build an abstraction for it. A `# keep in sync with the other experiments/*.yaml` pointer is the only cheap improvement — the duplication is exactly what let the other experiments drift.

### What's good here

- The 0.11.2 → 0.11.3 bump is load-bearing and correctly ordered — `checker_context`'s typed model landed in 0.11.3, so `params`/`env_params` would be a load-time rejection on 0.11.2.
- `env_params` (env var *names*, never values) is exactly the mechanism `LiteLLMRoute` prescribes: no credential enters a YAML, and the names are safe to record in `environment_info`.
- `smoke.yaml` / `nightly.yaml` already forwarded `CODEX_*` via `env_passthrough_extra`, so the two docker experiments that matter most were env-complete for the in-container checker.
- The Dockerfile comment at `:141-145` states the genuinely non-obvious fact (a host install doesn't reach the in-container orchestrator). More of this PR should be commented like that — the new YAML block carries no comments, in files where every other non-obvious value has one.
- The new Docker layer is placed last, so it doesn't invalidate the expensive npm/.NET layers above it.

### Summary

Intent and the version pin are right, but as written this isn't mergeable: applying `route: litellm` as an experiment-wide default collides with coder-eval 0.11.3's own `llm_judge`-only guard, and 318 tasks — including one `smoke`-tagged — will raise `ValueError` at setup rather than run. Secondary to that, the two `tempdir` jobs that actually need the host wiring were missed while three docker/plan-only sites that don't use it got it, and local `make` runs now error with no fallback and no README update.

<sub>🤖 Multi-model review (Gemini 3 · GPT-5.6 · Opus) via [Claude Code](https://claude.com/claude-code); critical finding verified against `coder_eval@v0.11.3` and the task corpus.</sub>

uipreliga (approved):
Fix what you agree with and 🚢

PR #2861, opened 2026-08-27
test(maestro-flow): outcome-graded DevCon billing-invoice-lookup e2e task

nikhil-maryala:
## What

Adds one **outcome-based** e2e task for `uipath-maestro-flow`, derived from Jiyang's DevCon **BillingInvoiceLookup** scenario (prompt reused near-verbatim), to grow our outcome-graded Flow coverage.

- **Prompt (DevCon scenario):** build `BillingInvoiceLookupOutcome` — normalize a messy `invoiceNumber` (trim, uppercase, prepend `MCS-`) and query the `BillingDisputeERP` Data Service entity; return `matchedInvoiceNumber` + `lineItemCount`.
- **Graded on outcome, not structure:** the agent builds + validates only; the checker runs `uip maestro flow debug` itself over three malformed inputs (missing prefix, wrong casing, leading whitespace) and asserts each resolves — via a **real Data Service query** — to `MCS-2026-04872` with **8 line items**. A required `uipath-dataservice.query` node + deterministic 8-line oracle block hardcoding.
- **Tags:** standardized outcome markers `path-to-ga`, `outcome-graded`.
- Registers `outcome-graded` in the `tests/README.md` tag taxonomy as a flat boolean marker.

## Success criteria
1. `command_executed` (advisory, 1.5) — agent validated the flow
2. `run_command` (3.0) — `flow validate` passes
3. `run_command` (5.0) — outcome checker: 3 malformed inputs each → `MCS-2026-04872`, 8 line items

## Validation
Dispatching `run-coder-eval.yml` from this branch to confirm the task passes end-to-end (agent build + validate + debug outcome).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2866, opened 2026-08-27
test(maestro-flow): add empty-result outcome case to billing_invoice_lookup

nikhil-maryala:
## What

Adds a **new outcome-based evaluator** to the existing `multi_node/billing_invoice_lookup` task instead of shipping a duplicate task file (context: a near-identical copy was proposed and closed — #2861). This grows outcome coverage in-place.

**Change is checker-only** (`check_billing_invoice_lookup.py`) — the pinned prompt and the task YAML are untouched:

- After the three happy-path malformed inputs, adds a **fourth case**: a normalized-but-absent invoice (` mcs-9999-00000 ` → `MCS-9999-00000`, not seeded in `BillingDisputeERP`).
- Asserts the debug run returns **0 line items** — verifying the flow handles an **empty result set** (the query matches no rows) rather than only the happy path where a row always exists.

The existing `uipath-dataservice.query` node assertion and the deterministic 8-line oracle are unchanged; the new case adds a second, distinct oracle (0 rows).

## ⚠️ Same-ground campaign — @tmatup please review

`billing_invoice_lookup` carries the same-ground campaign contract (Tao, 2026-08-10) and is compared cross-arm. Notes:

- The edit is **checker-only and loop-neutral** — both arms share this file, so parity is preserved by construction (no prompt/limit-envelope change).
- **But** it changes the intersection criteria and may **lower the measured pass rate**: the pinned prompt does not explicitly ask for empty-result handling, so agent builds that don't gracefully handle a no-match query will now fail the fourth case. That is the intended signal (a real robustness gap), but it affects your campaign measurements — please confirm you're OK with it before merge, or tell me to hold / move it to a standalone task.

## Validation
Dispatching `run-coder-eval.yml` from this branch to confirm the enhanced task still passes end-to-end (all four cases).

### DIFF
--- a/tests/tasks/uipath-maestro-flow/multi_node/billing_invoice_lookup/check_billing_invoice_lookup.py
+++ b/tests/tasks/uipath-maestro-flow/multi_node/billing_invoice_lookup/check_billing_invoice_lookup.py
@@ -2,13 +2,15 @@
 """BillingInvoiceLookup: the agent builds + validates only; this check runs
 `uip maestro flow debug` itself for three malformed invoice-number forms and
 asserts each resolves, via a real Data Service query, to invoice MCS-2026-04872
-with 8 line items.
+with 8 line items. A fourth case supplies a normalized-but-absent invoice and
+asserts the flow returns 0 line items — verifying it handles an empty result
+set, not just the happy path.
 
 The flow normalizes a raw `invoiceNumber` (trim, uppercase, ensure the "MCS-"
 prefix) and queries the seeded `BillingDisputeERP` entity. A Data Service query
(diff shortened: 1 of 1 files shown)


nikhil-maryala:
Closing. On reflection this bolted an assertion onto the existing task that grades an **unspecified contract** — the prompt only defines matchedInvoiceNumber/lineItemCount for a *matching* invoice, so asserting '0 line items on an absent invoice' tests behavior the task never asked for. It also mutated a parity-locked same-ground campaign task (@tmatup) and coupled to the primary criterion. If empty-result coverage is wanted, it should be a self-contained task whose prompt explicitly requires no-match handling. #2863 stands as the outcome-coverage addition.

PR #2869, opened 2026-08-27
fix(evals): point the Flow v2 preview experiment at the shared image

bai-uipath:
#2841 moved `@uipath/flow-sdk` and the connector registry into the shared image and taught `nightly.yaml` to stage them. `flow-v2-preview.yaml` was left on the old contract, so it still asks for a `skills-codex:latest` image that nothing on this branch builds, and still writes an npmrc for an in-sandbox `npm install` the baked SDK makes unnecessary.

This points it at the shared image and stages the SDK the same way `nightly.yaml` does, so both arms of the comparison resolve the SDK and connector library from the same place rather than one of them fetching its own at task time.

Staging is unconditional here rather than gated behind `MAESTRO_FLOW_SDK_SETUP`, since a preview run always authors through the SDK. Say the word if you'd rather both experiments share the switch.

### DIFF
--- a/tests/experiments/flow-v2-preview.yaml
+++ b/tests/experiments/flow-v2-preview.yaml
@@ -15,12 +15,10 @@ defaults:
   sandbox:
     driver: docker
     docker:
-      image: skills-codex:latest
+      image: skills-image:latest
       network: bridge
       env_passthrough_extra:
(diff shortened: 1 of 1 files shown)


uipreliga (approved):
🚢

PR #2873, opened 2026-08-27
fix(planner): a gate reads the producing field, and the validator says so

Yiqi517:
## What goes wrong

A case built from an SDD stalls right after a person makes a decision. The buyer approves, both of that stage's required tasks finish, and then nothing happens. The stage stays open, the next stage never opens, and the case sits there. There is no incident, no error, and `uip maestro case validate` calls the case valid.

The runtime shows the guard on that stage's exit evaluating to false at the moment the value it reads is present:

```json
{"expression": false, "expressionString": "vars.buyerDecision === \"approve\""}
```

Once it evaluates false, nothing moves again: the event that would trigger the next evaluation is the very one the case is waiting for.

## Why it happens

A task field lands in exactly one slot. The SDD wrote `Action -> buyerDecision`, which gives one value two names: the engine stores it under `buyerDecision`, while a task-output reference resolves to the minted `action2`. Neither name reaches every scope a guard is evaluated in, so the branch never fires.

Two frontend sites pick different slots, which is why a renaming row splits them:

- [`CaseManagementTasksEventSubprocessUtils.ts:153,155`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/services/serialization/converters/case-converter/case-mgmt/generate-adhoc-case/tasks-event-subprocess/shared/lifecycle/CaseManagementTasksEventSubprocessUtils.ts#L153-L155) hoists the value out of the task with `` `=js:vars.${output.var ?? output.id}` `` — the renamed slot whenever a row renames.
- [`CaseManagementOrchestrationTaskFlowUtils.ts:205`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/services/serialization/converters/case-converter/case-mgmt/generate-adhoc-case/main-event-subprocess/shared/utils/CaseManagementOrchestrationTaskFlowUtils.ts#L205) builds the state the rules evaluator reads with `` `=js:vars.${output.id}` `` — always the field's own slot.

A field with no renaming row keeps one name for both, so both sites agree and the guard reads a populated slot.

A runtime record bears this out. Across thirteen variable scopes in two instances of a case carrying this shape, every `id` slot is null in every scope, while the `var` name is carried by the case globals and the task-completion scope and absent from the rest. Neither half is dependable wherever a guard happens to be evaluated.

**The platform side of this is a known open defect** — renamed values are not visible to rule evaluation, tracked as MST-10838, unresolved since June. This PR does not fix that. It stops the coding agent from authoring the shape that triggers it, and the real fix belongs on the platform.

This is the third attempt at that from the skills side:

| | Fix | What it did |
|---|---|---|
| Jun 15 | [#1471](https://github.com/UiPath/skills/pull/1471) | Introduced `vars.$xref('Stage','Task','field')` so a gate could name a task's own field instead of the case variable |
| Aug 20 | [#2718](https://github.com/UiPath/skills/pull/2718) | Reorganised the case guidance after the last report of this bug, writing the § Gate on the producer section that carries the sentence below |
| this PR | | Closes the gap both left, and adds the mechanical checks so a fourth report does not depend on wording alone |

## Why the guidance produced that shape

**First, the stated reason invited an exemption.** A case built while this PR was open reached the same shape, and its build log records how. The agent adopted the reason the sentence below gives, that "the extract of the task that fired the gate has not run yet, so an `IF` that reads a case variable that task writes is stale", and concluded that a gate firing after a whole stage completes is therefore safe. It wrote that exemption down, built a classifier around it, and the classifier reported all seventeen of the case's gates clean. Two of them were the stalling shape. Timing is not the mechanism: the record above holds in every scope of both instances, so a later evaluation reaches no populated slot either. A reason phrased as "not yet" licenses an exemption for anything that fires later; the replacement states what holds at every moment.

**Second, two instructions pointed opposite ways.**

§ When to declare a Case Variables row offered "case-level state read by a condition (`IF`)" as a reason to declare a row, and never said that a task's own field is not case-level state. Declaring a row means an Outputs row moves the value into it.

§ Gate on the producer then said the `->` extract can stay, because the extract is not what the gate reads. It is exactly what the gate reads — the gate reads the slot the extract emptied.

An author following both instructions produces a renamed field with a gate pointed at it.

## What changes

- The Case-Variables rule keeps its three existing reasons and gains the boundary it was missing: a task's own field is not case-level state, and a gate reads it with `vars.$xref(...)`. State produced by a `Default`, an `=` row, or a decision button is still declared and still read by an `IF` — five variables across the existing fixtures do exactly that.
- A value that is also needed elsewhere is copied with a separate `=` row. A copy leaves the field's own slot intact; a rename does not.
- The WHEN list gains `required-tasks-completed` and `selected-stage-completed`. The scheduler recognises six event types ([`Constants.cs:35-43`](https://github.com/UiPath/dmnscheduler/blob/bfeec30a8eaa6478df2073479a781df1a6d6cc29/Scheduler/Core/Constants.cs#L35-L43)) and a stage completing is not among them — a stage completes because its last required task did, so both of these evaluate on a `TaskCompleted` pass exactly like the two already listed.
- The rule is scoped by what the `IF` reads rather than by what fires the gate. A scope worded as "a gate that fires on a task completing" excludes `selected-stage-completed` on its face, which is how an agent holding item 4's own list still stepped over it.
- The rule becomes four numbered obligations instead of one sentence carrying four.

This guidance has been restated twice and a case reached the same shape both times, so the change is not only wording. `audit_sdd.py` now reports any gate reading a renamed field, naming the line that renamed it and the shape to use instead, and reads both the bare and backticked WHEN cell styles the fixtures use. `check_reject_route.py` gains the same rule at the point where it already asserts the rejection route exists, so an SDD passes by carrying the gate **and** leaving the field unrenamed — a criterion that only asked "no finding" would have passed an agent that authored nothing at all.

Two fixtures carry the shape today. `aged_invoice_structural` gates on a field its own decision buttons already write, so the renaming row was redundant and is dropped. `case_finalize_draft` gates on a field whose renaming row is its only producer; removing it would break lineage, so it is reported for its owner rather than rewritten here.

## Verification

- All nineteen SDD fixtures audited before and after, with both script versions colocated so the model-driven checks arm identically: byte-identical output on eighteen, the nineteenth differing only by the two new findings.
- Eleven hand-built cases. Reported: a renamed field read by each of the four WHENs, in bare and in backticked cells, and an equal-name row whose field is extracted twice. Silent: an equal-name unique row, a renamed field with no reader, one read by a task input rather than a gate, one read under `case-entered`, and a dotted equal-name path.
- `check_reject_route.py` exits 0 on a finalized SDD carrying the gate, 1 with a rename spliced in, and 1 on the unfinalized draft — doing nothing does not pass.
- `load_model_facts` still returns `degraded: None` with all five facts populated; one edited line sits inside a section it parses, and a silent failure there would disarm three other checks.
- `expressionIsValidLength` ([`Tools.ts:686-692`](https://github.com/UiPath/PO.Frontend/blob/b1ed4e8b37188fd7d2482899d884abfde747dbe9/src/utils/Tools.ts#L686-L692)) rejects only an empty string, so no guard is dropped for length.
- `npm run skills:validate` OK · `check-skill-status.py` OK · `validate-skill-descriptions.sh` clean · `pytest tests/tasks/uipath-maestro-case/_shared/` 52 passed, 14 skipped · all five in-file anchors resolve · no flavor override or marker on any changed file.

### DIFF
--- a/skills/uipath-planner/references/case-design-layers-guide.md
+++ b/skills/uipath-planner/references/case-design-layers-guide.md
@@ -254,7 +254,7 @@ A row that relays one task's output to one consumer is the relay anti-pattern �
 
 ### Gate on the producer, never on a renamed field
 
-A task field lands in exactly one slot. `Action -> decision` puts the value in `decision` and leaves the `Action` slot empty; an equal-name row (`Action -> Action`) keeps one slot under one name, until a second row extracts `Action` elsewhere in the case and the allocator suffixes the later slot. A rule that fires on task completion reads the field's own slot, so a gate on a renamed field reads the empty one. The branch silently never fires, the stage stalls, and nothing errors.
+A task field lands in exactly one slot. `Action -> decision` puts the value in `decision` and leaves the `Action` slot empty; an equal-name row (`Action -> Action`) keeps one slot under one name, until a second row extracts `Action` elsewhere in the case and the allocator suffixes the later slot. A rule reads the field's own slot, so a gate on a renamed field reads the empty one, whenever that gate fires. The branch silently never fires, the stage stalls, and nothing errors.
 
 | Outputs row on `Decide` | `IF` reads | At gate time |
 |---|---|---|
@@ -262,7 +262,7 @@ A task field lands in exactly one slot. `Action -> decision` puts the value in `
 | `Action -> decision` | `vars.$xref(...,'Action')` — the emptied slot | ✗ `null` |
 | no Outputs row for `Action` | `vars.$xref(...,'Action')` — one slot, one name | ✓ populated |
 
-**Rule** — all four hold for a gate that fires on a task completing:
+**Rule** — all four hold for a gate whose `IF` reads a task's output:
 
 1. The `IF` reads the producing task's own field as `vars.$xref('<Stage>','<Task>','<field>')`.
 2. That field carries no renaming Outputs row — no row at all, or an equal-name row whose target is unique case-wide.
--- a/skills/uipath-planner/references/case-design-layers-guide.md
+++ b/skills/uipath-planner/references/case-design-layers-guide.md
@@ -254,7 +254,7 @@ A row that relays one task's output to one consumer is the relay anti-pattern â€
 
 ### Gate on the producer, never on a renamed field
 
-A task field lands in exactly one slot. `Action -> decision` puts the value in `decision` and leaves the `Action` slot empty; an equal-name row (`Action -> Action`) keeps one slot under one name, until a second row extracts `Action` elsewhere in the case and the allocator suffixes the later slot. A rule reads the field's own slot, so a gate on a renamed field reads the empty one, whenever that gate fires. The branch silently never fires, the stage stalls, and nothing errors.
+A task field lands in exactly one slot. `Action -> decision` puts the value in `decision` and leaves the `Action` slot empty; an equal-name row keeps one slot under one name â€” equal on the source path's last segment, so `Action -> Action` and `response.status -> status` both qualify, until a second row extracts `Action` elsewhere in the case and the allocator suffixes the later slot. A rule reads the field's own slot, so a gate on a renamed field reads the empty one, whenever that gate fires. The branch silently never fires, the stage stalls, and nothing errors.
 
 | Outputs row on `Decide` | `IF` reads | At gate time |
 |---|---|---|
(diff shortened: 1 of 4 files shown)


Yiqi517:
> **Runtime timing of the four WHENs.** The script and guidance assert that `required-tasks-completed`, `selected-tasks-completed`, `selected-stage-completed`, and `selected-stage-exited` all evaluate before the producing task's extract runs. This is a runtime-behavior claim that code review can't verify — worth confirming that `selected-stage-completed` shares the same evaluate-before-extract ordering.

Confirmed on a live tenant, and the ordering turns out not to be the mechanism.

A `selected-stage-completed` gate reading a renamed field was run against two recorded instances, sampled across thirteen variable scopes. The renamed name holds the decision in some scopes and is absent from others; the field's own slot is null in every scope of both runs. Waiting longer reaches neither — the renaming row splits the value between a name the engine writes and a name a task-output reference resolves to, and no scope carries both dependably.

So the rule holds for `selected-stage-completed`, but for the reason the PR now states rather than an evaluate-before-extract ordering. The scope wording was changed accordingly: the rule applies to any gate whose `IF` reads a task's output, not to gates picked out by what fires them.

The two `audit_sdd.py` observations are fixed in 630152b37 — the ineffective `re.M` is dropped, and every renaming row feeding a target is now kept and named in the finding instead of only the last one.

roberts-cliff (commented):
let's not add more python scripts, we don't know if a user will have that runtime.
you could add node scripts instead

PR #2881, opened 2026-08-28
fix(maestro-flow): fix the failing ixp coder-eval rows

joe-prosser:
The failing IxP rows ([thread](https://uipath-product.slack.com/archives/C0A2T23NJ59/p1787841394793009?thread_ts=1787824967.130479&cid=C0A2T23NJ59)) die on the clock, not the assertions: sonnet-5 lands the IxP node early, then wires ungraded detail until the turn ceiling kills the row — the dead rows' artifacts grade 1.0 when their criteria are re-run. Raising budgets doesn't help; the work expands to fill them (died at 903s under a 900s ceiling, 1203s under 1200s).

- **`routing.yaml`**: early-stop sentinel — cuts the run once the registry search has passed and a tool call writes the IxP/mock node literal. Validated: **10/10 cut rows SUCCESS at 325–682s**, including the row that previously ground to death.
- **`impl.md`**: an unnamed downstream target goes to Open Questions instead of ask-and-halt. Before: `routing/invoice-extraction` asked and built nothing. After: **built in 3/3 runs**.
- **`e2e_01`**: `turn_timeout` 1200→1500. Not effective in validation (3/3 deaths at the new ceiling) — kept for a clean-window read, happy to drop.

Validated with 6× `run-coder-eval.yml` on `agent=claude`: [1](https://github.com/UiPath/skills/actions/runs/33158931153)/[2](https://github.com/UiPath/skills/actions/runs/33158938907)/[3](https://github.com/UiPath/skills/actions/runs/33158946369), [4](https://github.com/UiPath/skills/actions/runs/33164889886)/[5](https://github.com/UiPath/skills/actions/runs/33164896580)/[6](https://github.com/UiPath/skills/actions/runs/33164902795). Full analysis on RE-13392.


richsilv (approved):

rockymadden (commented):
# 🔍 Review

Read the coder_eval early-stop internals and downloaded the artifacts from all 6 validation runs. The two validated SHAs are a clean A/B (`9fbdef83` = impl.md + budget bump, no sentinel; `c2d1e09a` = + sentinel), so the sentinel's effect is directly measurable.

## What it does

| File | Change |
|---|---|
| `references/plugins/ixp/impl.md` | An unresolved/unnamed **downstream** target is no longer a reason to ask-and-halt. Carry it in Open Questions, scaffold anyway. |
| `tests/.../ixp/routing.yaml` | Adds an early-stop **sentinel** criterion (`on_pass: stop`) that cuts the run once a tool call writes the IxP/mock node literal. Arms the registry criterion with `stop_early: {}`. Raises `task_timeout` 900 → 1200 and adds `turn_timeout: 1200`. |
| `tests/.../ixp/e2e_01_...yaml` | `turn_timeout` 1200 → 1500. |

---

## 🔴 The sentinel makes the suite worse, and the validation runs prove it

Pooled `skill-flow-ixp-routing` rows, 15 per arm:

| Arm | Row status | `.flow` artifact criterion (gate: mean ≥ 0.85) | Rows cut | Cut rows with **no node on disk** |
|---|---|---|---|---|
| `9fbdef83` no sentinel | 13 SUCCESS / 2 ERROR | **1.0** (1.0 / 1.0 / 1.0 per run) | 0 | 0 |
| `c2d1e09a` sentinel | 13 SUCCESS / 2 ERROR | **0.46** (0.75 / 0.5 / 0.2 per run) | 10 | **7** |

1. **Zero row-level benefit.** 13/15 SUCCESS in both arms. The budget bump alone already gets there.
2. **The cut destroys the artifact it is supposed to certify.** 7 of 10 cut rows ended with no IxP/mock node in any `.flow`. Every one still reported SUCCESS, because on an early-stopped run only ARMED criteria gate `final_status` (`EvaluationResult.armed_criteria_passed`), and the `run_command` artifact grep cannot be armed (it is not live-observable). The artifact criterion's suite mean falls below its own 0.85 gate in all three runs, so `suite_passed` is now False where it was True.

### Mechanism, confirmed from a transcript

`routing/receipts` in run [33164889886](https://github.com/UiPath/skills/actions/runs/33164889886):

- Tool call #50 is a `python3 <<EOF` heredoc that rewrites `ReceiptExtraction.flow` with the mock + IxP nodes.
- The sentinel matches `"type": "core.logic.mock"` at char 651 of that command's params.
- The watcher latches on `ToolStartEvent` (the tool_use *announcement*, before execution), `should_stop()` flips, `_pump_messages` breaks, the SDK subprocess dies.
- That command's `result_status` is `unknown`. The saved artifact contains `['core.trigger.manual']` and nothing else.
- Row: `SUCCESS`, `early_stop.reason = criterion_passed`, 890s elapsed, 32 turns still on the clock.

The deciding tool call is always killed. When it is the write, the node is lost. `pass_threshold: 0` protects the *sentinel's* score, not the artifact.

**Proposed fix:** drop the sentinel. If you want to keep the mechanism, it has to latch on a signal whose side effect nobody grades, for example a `uip maestro flow validate` invocation or a read-back of the `.flow`, never on a write. And per coder_eval's own docs, an early-stop flavor belongs in a separate variant (`stop_early: false` on the authoritative run), not baked into the shared task the nightly gates on.

---

## 🟠 Other issues

**`e2e_01` is not fixed and now costs more.** All six runs ERROR on it: 1202/1203/1202s at the old ceiling, 1505/1502/1503s at the new one. The change buys +300s of burn per run and nothing else. The added comment ("three consecutive sonnet-5 runs died at ~1202s with 300s of task budget left") and the pre-existing "1500s and 120 turns retain measured headroom" are both contradicted by the data now sitting in the same file. Drop the change, or drop the comment and say the ceiling is not the cause.

**The `routing.yaml` budget bump is the part that actually works, and the PR body argues against it.** The body says "raising budgets doesn't help." Baseline arm at 1200/1200 gets 13/15 SUCCESS with artifact mean 1.0. That bump is not mentioned in the body at all. Decide what the fix is and say so, since the failure path now burns 1200s instead of 900s on the two rows per run that still die (those are the large-single-write rows the sentinel never latches on anyway).

---

## 🟡 Minor

- **Wrong claim in a code comment.** "a zero-weight armed criterion collapses the pass-stop floor and would cut at the registry search" is not how it works, twice over: `BaseSuccessCriterion` rejects `weight: 0` together with any `stop_early` block, and `_floor()` is computed over the `on_pass: stop` subset only, so a zero-weight member yields `None` (no cut at all), not an earlier cut.
- **Weighted score is permanently diluted.** The sentinel carries weight 1.0 and scores 0 on non-latching rows, so a fully correct full run now reports 0.667 (see `contracts` in run 33164889886: registry 1.0, artifact 1.0, weighted 0.667). `weight: 0` is not available alongside `stop_early`, so this cannot be made score-neutral.
- **"Keep in sync with `scaffold_minimal` / `scaffold_multinode` / `routing_listing`" is now false.** Only this copy is armed. Say the divergence is intentional or the next person will propagate the arming.
- **Validated on `agent=claude` only.** The nightly default is `codex` (daily; claude runs twice a week). Latch behavior depends on tool-call shape, so codex needs its own read. Also confirm `delegate-sdk` sets `supports_cooperative_stop = True` in the private plugin, otherwise arming this task hard-errors at resolution for that agent (`EarlyStopConfigError`).
- **impl.md wording.** "Ambiguity anywhere else in the flow is never a reason to stop before scaffolding" is an absolute that collides with `CAPABILITY.md:64` (no healthy connection → STOP) and `SKILL.md:90` (non-interactive → stop and report the open decision). Scope it to the extraction node.
- **impl.md placement.** An unnamed downstream target is not IxP-specific. Connector and HITL rows will hit the same halt. `planning-arch.md` or `greenfield.md` looks like the right home.
- **impl.md evidence.** Both validated arms already contain the impl.md change, so the linked runs do not A/B it. The "3/3" claim needs the earlier runs linked.
- **Upstream.** Worth a coder_eval issue: on pass-stop the deciding call's side effect is discarded while post-hoc `run_command` criteria still grade the sandbox. The docs cover the distractor precision trade-off, not this one.

---

## 🧹 Comment hygiene

- The `routing.yaml` sentinel carries 10 lines of narrative comment for a 6-line criterion, and one paragraph of it states a mechanism that does not exist.
- "10/10 cut rows SUCCESS" is true and meaningless: SUCCESS on a cut row is decided by the armed subset, which excludes the check the suite exists to make. 7 of those 10 rows have no node on disk.
- Both YAML files now hold comments asserting numbers the runs refute.

---

## 🙋 Not my call

- Whether `uipath.ixp.*` is published on the staging tenant during these runs (real node vs mock changes what "landed" means).
- Whether `skill-flow-ixp-routing` is a release gate or advisory. If it gates, item 1 turns the suite red.
- The RE-13392 analysis. This is from the PR, the code, and the run artifacts only.

---

**Verdict:** `impl.md` is the one clean win. Land it plus the budget bump; drop the sentinel and the `e2e_01` change.

PR #2152, opened 2026-07-21
fix(uipath-coded-apps): use `uip admin external-apps` CLI for external app setup

ninja-shreyash:
## What & why

The `uipath-coded-apps` skill documented **only** a Playwright browser-automation path for creating/updating OAuth External Applications (`oauth-client-setup.md` was literally titled *"OAuth Client Management via Playwright CLI"*). A first-class CLI — `uip admin external-apps create/update/get` — already exists and supersedes it.

Fixes **PLT-106013**. Verified against the linked Slack thread:
- The CLI verbs are in `assets/uip-catalog-snapshot.json` (current) and fully documented in the sibling `uipath-admin` skill.
- The Playwright script (~400 lines of portal DOM automation with a "re-run codegen when selectors break" maintenance loop) provided **no capability the CLI lacks** — both hit the same Identity backend and need the same admin permission.

> The ticket's second symptom — `Cannot find the target partition. (#218)` from duplicate `acr_values` — is an **Identity Server** bug tracked separately in **PLT-106121** (server-side, owned by the Identity team). Out of scope here.

## Changes

| File | Change |
|---|---|
| `oauth-client-setup.md` | Full rewrite → CLI create/update. `get → union → update` documented because CLI `update` **replaces** scopes/redirects (doesn't merge). Kept a slim manual-portal fallback for the one real gap: caller lacks identity-admin permission (`403`) or isn't authenticated. Removed Chrome detection, Playwright install, the consolidated script, and codegen (~660 lines). |
| `create-web-app.md` | Create the External App via `uip admin external-apps create` instead of browser automation. |
| `create-action-app.md` | Same for the coded **action app** path (per Sandeepan's comment on the ticket). Preserves the action-app nuance (redirect URI is an inert placeholder — runs in Action Center's iframe). |
| `debug.md` | `add-redirects` / `add-scopes` / `redirect_uri_mismatch` / `invalid_scope` / 401 remediations now use `uip admin external-apps update`. The runtime **login-reproduction** harness stays Playwright — the CLI can't reproduce a browser OAuth login. |
| `sdk/conversational-agent.md` | Pointer wording: "browser automation" → CLI. |

## Fallback rationale

CLI-first, with a **manual-portal** fallback only (Playwright removed entirely). External-app management is permission-gated; when the caller lacks admin rights, the task must be handed to an admin — a short click-through one-pager serves that. Playwright didn't help there either (same permission requirement).

## Tests

[...]

PR #2882, opened 2026-08-28
docs(uipath-coded-apps): note portal picker labels vs scope names [APPS-37264]

ninja-shreyash:
## Why

The ticket asked to add Insights rows to the "Scope to Resource Mapping" table in `oauth-client-setup.md`. That table was removed in #2152, when external-app setup moved to the `uip admin external-apps` CLI. The CLI takes flat scope names and Identity resolves each to its resource, so there is no table to maintain and the CLI path already works from the scope names in `oauth-scopes.md`.

The one place still affected is the manual portal fallback, where a human picks resources in the portal picker. It did not say the picker groups scopes under resource labels that differ from scope names, and nothing flagged that Insights is split across two picker resources.

## Change

Adds a note to the Manual portal fallback section: the picker uses resource labels, not scope names, and Insights analytics/SLA needs both the Insights resource (scope `Insights`) and the Insights Real-Time Data resource (scope `Insights.RealTimeData`), plus PIMS for SLA summaries.

Docs only. The CLI flow is unchanged.

### DIFF
--- a/skills/uipath-coded-apps/references/oauth-client-setup.md
+++ b/skills/uipath-coded-apps/references/oauth-client-setup.md
@@ -96,7 +96,7 @@ A missing scope name (`scope not found` / `Not all scopes=<list> are present in 
 
 ## Manual portal fallback
 
-> **Portal picker labels vs scope names.** In the portal, scopes are grouped under **resource labels** that differ from the CLI scope names. Pick the scopes per resource from [oauth-scopes.md](oauth-scopes.md). Some services span two resources: Insights analytics and SLA reads need both the **Insights** resource (scope `Insights`) and the **Insights Real-Time Data** resource (scope `Insights.RealTimeData`); SLA summaries also need **PIMS**.
+> **Portal picker labels vs scope names.** In the portal, scopes are grouped under **resource labels** that differ from the CLI scope names. Pick the scopes per resource from [oauth-scopes.md](oauth-scopes.md). Some services require more than one resource.
 
 ### Create a new External Application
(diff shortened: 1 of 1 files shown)


deepeshrai-tech (approved):

maninder-uipath (approved):

skills/uipath-coded-apps/references/oauth-client-setup.md
```diff
@@ -96,6 +96,8 @@ A missing scope name (`scope not found` / `Not all scopes=<list> are present in
 
 ## Manual portal fallback
 
+> **Portal picker labels vs scope names.** In the portal, scopes are grouped under **resource labels** that differ from the CLI scope names. Pick the scopes per resource from [oauth-scopes.md](oauth-scopes.md). Some services span two resources: Insights analytics and SLA reads need both the **Insights** resource (scope `Insights`) and the **Insights Real-Time Data** resource (scope `Insights.RealTimeData`); SLA summaries also need **PIMS**.
```
vnaren23: 1. `Some services span two resources` vs `Some services require more than one resource`. Which is better?
2. Should we mention this? Cant we just leave it as some services require statemenet? 
> Insights analytics and SLA reads need both the **Insights** resource (scope `Insights`) and the **Insights Real-Time Data** resource (scope `Insights.RealTimeData`); SLA summaries also need **PIMS**.

ninja-shreyash: done

PR #2885, opened 2026-08-28
perf(uipath-maestro-flow): whole-bundle token optimization (−20.7%)

CarlesUIPath:
## What

Whole-bundle token-efficiency optimization of the **`uipath-maestro-flow`** skill (SkillReducer, arXiv:2603.29919, combined with the `/optimize-skill` compaction engine). Compresses the description, SKILL.md body, and **all** reference files while preserving every rule, CLI, enum, threshold, and worked example that drives agent behavior.

**Bundle: 222,786 → 193,839 tokens (−13.0%, −28,947)** · **body 5,608 → 4,013 (−28.4%)** vs current `main` (`bridge.skill_stats`, tiktoken cl100k_base).

> **Why lower than earlier drafts (−20.2% → −13.0%):** (1) the −20.2% used a heavier 225,104 baseline snapshot — re-measured consistently, that candidate was ~−17.6%; (2) the review required restoring over-dropped worked examples (+~1.8k tok); (3) **merging current `main` pulled in two new features** — #2809 Flow→IXP handoff and #2484 inline-agent encoding — and their reference files (`ixp/impl.md`, `inline-agent/impl.md`) were taken **uncompressed** to ship the features intact rather than risk a bad merge (+~10k tok). The remaining reduction is faithful and concentrated in 3 verbose files (connector, planning-arch, editing-operations-json).

## How

- **Description** — ddmin minimization (kept identity + `.flow`/CLI signals + sibling redirects).
- **Body** — taxonomy-guided relocation of detail into references, then the gated compaction loop (~−36% pre-merge); the merged body is **4,013 tok / −28.4%** vs current `main` after absorbing main's new rule-4 IXP-handoff (#2809).
- **References** — cross-file dedup + per-file compression, each edit gated on its exercising tasks (accept only if 0 regressions vs baseline).
- **Compression boundary respected** (`.claude/rules/token-optimization.md §0`): rules/action-patterns/procedures kept as full sentences; only enumerations, tables, and duplication compressed.

## Results

| Metric | Baseline (`origin/main`) | Candidate | Δ |
|---|---|---|---|
| Bundle (all files) | 222,786 | 193,839 | **−13.0%** |
| SKILL.md body | 5,608 | 4,013 | **−28.4%** |
| Cross-file redundancy | 8.2% | 6.7% | −1.5pp |

**Suite size:** 126 task *files* on disk, but the 4 ixp matrix files expand into 25 sub-cases, so the executable suite is **~140 instances** (138 graded per seed in docker; the nightly reports the same ~140). Our docker runs graded the **expanded** set — the ixp sub-cases (`routing`, `routing-listing/r01…r10`, `routing-negative/*`, `e2e-project-selection/*`) are individually covered on both arms.

**Per-task cost / usage / wall on the merged candidate — sonnet + terra (candidate vs current `main` nightly, single-seed, measured only on both-SUCCESS tasks):**

| Model | n (both✓) | Δ tokens | Δ cost | Δ wall | per-task avg (ctrl → cand) |
|---|---|---|---|---|---|
| **terra** (`gpt-5.6-terra`) | 92 | **−3.9%** | **−4.6%** | **−8.2%** | 1.01M→0.97M tok · $0.398→$0.380 · 188→173s |
| **sonnet** (`claude-sonnet-5`) | 91 | **+10.8%** | **+5.1%** | **+7.6%** | 2.87M→3.18M tok · $1.556→$1.636 · 316→340s |

On the new runs, previous gains cannot be reproduced after add-backs. 

**Honest read:** the two models **disagree in sign** (terra favorable, sonnet unfavorable this seed) — the signature of **single-seed agent/cache-read variance, not the skill compression**. The compressed skill's ~40k-token bundle delta is **<2%** of the ~1M (terra) to ~2.9M (sonnet) `total_tokens` the agent actually processes per task (dominated by cache-read of the full agentic context). So per-task cost is **not a reliable compression signal at single-seed** — consistent with the 2-seed luna result (modest, noise-bounded). **The durable, reliable win is the −13.0% bundle / −28.4% body resident reduction** (routing + context-hygiene budget), not per-task token cost.

**Conclusion:** Maestro-flow tasks require multi-turn builds where context increases, and therefore, main body/reference trimmings do not result in effective performance gains.

## Validation -- last run

Post-merge re-gate of the candidate (`b8e34971d`) vs current `main` nightlies, on the correct 0.11.6 Azure toolchain (single-seed unless noted). Verdict is **regressions**, not per-task cost (cost is noise-bounded — see above).

| Model | candidate vs control | pass /140 | verdict |
|---|---|---|---|
| **sonnet** `claude-sonnet-5` | Azure `adhoc-2026-09-02_13-59-42` vs main nightly `2026-09-02_09-33-49` | **107 vs 110** | parity within single-seed noise; **0 systematic regressions** — 5 cand-better vs 7 graded-worse + 1 env-error, all in known live/e2e/connector/ixp flaky families |
| **terra** `gpt-5.6-terra` | Azure `adhoc-2026-09-02_16-03-46` vs main nightly `2026-09-02_09-32-42` | **119 vs 119** | **dead even, 0 regressions** — 2 cand-better ↔ 2 cand-worse (`ipe-complex-array`, `ipe-jira-lifecycle`, data-fabric/jira live), 0 ERRORs |
| **luna** `gpt-5.6-luna` | 2-seed docker + tie-breaks — **development phase (pre-merge)** | n/a (pre-merge) | **Checked throughout development; no regressions caught** (every candidate-worse task tie-broke to flaky). Not re-run on the merged bundle. |

**Sonnet + terra confirm 0 systematic regressions on the merged candidate** (two independent models, parity vs `main`). Luna was validated pre-merge (2-seed, 0 regressions).

## Code review — addressed (CHANGES_REQUESTED)

All 11 findings fixed. The skill's own `.maintenance/check-all.sh` now prints **`All checks passed.`** (was 4 failing: 37 anchors, 5 template sections, 6 link-text, 2 versions):

- **Critical** — `node-output-wiring.md` restored to main's #2379 contract (`{{input.<id>__output__<var>}}`) + Voice `callContext` row (had reverted to the pre-#2379 form via a bad merge).
- **Anchors 37→0** — renamed headings restored to original text (cli-conventions §3/§6, editing-operations, editing-operations-json, planning-arch, connector); merged sections given explicit `<a id>` anchors (ixp, http, inline-agent, eval-sets) — resolves the cross-skill inbound links from `uipath-agents` too.
- **Voice #2379 discoverability** — restored Plugin Index + Port Reference rows (planning-arch), routing rows (planning-impl), "Add voice nodes" (brownfield).
- **`skill-flavor` leak** — CLI clauses removed from canonical `SKILL.md` bullets (kept inside markers); verified `compose-skill-flavor build` emits a clean studioweb SKILL.md.
- **Template 5→0** — restored `When to use this capability` / `Workflow` / `Common tasks` headings (diagnose, evaluate).
- **Worked examples** — restored connector Step 6c `--detail` payloads + `DAP-DT-_2003` signature, Step 3 objectName/`--operation` pairs, mermaid counter-examples, the `< 0.3.4` legacy CLI row.
- **link-text 6→0, versions 2→0, 78 curly quotes→straight, consent-gate phrasing restored (rules 2/6/8).**

Restorations cost **+1,809 tok**; after merging current `main` (below), the bundle is **−13.0%** vs main.

## Merge notes

Merged current `main` (twice, as it advanced): kept our compressed content, re-homed onto main's flattened `references/` layout (#2879, link checker clean), dropped internal `MST-*`/Jira identifiers (#2865/#9ab6b320a), inherited main's 180 s validate budget + retry (#2853/#2859), the new `inline-voice-agent` plugin (#2379, ships uncompressed), and the post-flatten link-drift repair (#2884).

CI: verb gate ✅, skill smoke (luna) ✅, all checks green.


skills/uipath-maestro-flow/references/shared/node-output-wiring.md
```diff
@@ -1,139 +1,83 @@
...
+| Loop nodes (`core.logic.loop`) | `inputs.collection` | **YES** |
+| Subflow nodes (`core.subflow`) | `inputs.<inputId>.source` | **YES** |
+| Script nodes (`core.action.script`) | `inputs.script` body | **NO**; the body is already JS and reads `$vars.*` directly |
+| Inline-agent prompt (`uipath.agent.autonomous` `agent.json` `messages[].content`) | Flow-node tokens | **NO**; use `{{ $vars.<flowNodeId>.output[.<field>] }}` with spaced braces and mirror it in `contentTokens[]` as `{ "type": "variable", "rawString": " $vars.<flowNodeId>.output[.<field>] " }`. `rawString` must include leading and trailing space. Never use `{{input.<id>}}` or bare `{{name}}`. See [author/references/plugins/inline-agent/impl.md § Wiring Flow Variables into Agent Prompts](../author/plugins/inline-agent/impl.md#wiring-flow-variables-into-agent-prompts). |
```
rockymadden: 🔴 **Critical — this row reverted to the pre-#2379 contract.**

This line is byte-identical to `d3ca41816^`, i.e. the version *before* #2379 landed. A merge kept the branch's stale copy of this shared file.

| | Contract |
|---|---|
| `main` today (#2379) | `{{input.<trigger>__output__<var>}}`, `rawString` brace-free and space-free |
| This line | `{{ $vars.<flowNodeId>.output[.<field>] }}`, `rawString` "must include leading and trailing space", and **"Never use `{{input.<id>}}`"** |

This PR's own `references/author/plugins/inline-agent/impl.md:41-46` teaches the `__output__` form. So the bundle now ships a shared wiring table that explicitly forbids what its own plugin doc mandates. An agent reading this table first authors a prompt that fails `uip agent validate`.

Corroborating: `check-link-text.sh` flags this same line for a stale pre-flatten label (`author/references/plugins/inline-agent/impl.md`), which is a second tell that it came through a bad merge resolution rather than a deliberate rewrite.

Also gone from this table: the **Voice `callContext`** row (`uipath.agent.voice` / `uipath.conversational.voice.end-call`, structured binding, `fieldType` object vs string) that #2379 added.

**Fix:** restore both rows from `origin/main` verbatim. Do not re-compress, then re-check against `inline-agent/impl.md`.

skills/uipath-maestro-flow/SKILL.md
```diff
@@ -1,112 +1,84 @@
...
 
-Comprehensive guide for creating, editing, validating, debugging, publishing, diagnosing, and evaluating UiPath Flow projects using the `uip` CLI and `.flow` file format. The skill is organized into four capabilities — **Author**, **Operate**, **Diagnose**, **Evaluate** — each with its own index doc.
+Guide for creating, editing, validating, debugging, publishing, diagnosing, and evaluating UiPath Flow projects with the `uip` CLI and `.flow` format.
 
-## When to use this skill
-
-**Author** — building or editing a `.flow` file. Read [references/author/CAPABILITY.md](references/author/CAPABILITY.md).
+## Capabilities
 
+- **Author** — Build and edit `.flow` files; create projects with `uip maestro flow init`; add nodes, edges, variables, subflows, transforms, and triggers; explore the registry; validate and format locally; apply node ownership; configure connectors, triggers, managed HTTP, inline-agent scaffolding, IxP/document-extraction nodes, and IxP models; plan complex flows first. Read [references/author/CAPABILITY.md](references/author/CAPABILITY.md).
```
rockymadden: 🟠 **Major — `skill-flavor` leak.** `create projects with `uip maestro flow init`` now sits in the canonical prose *and* inside the `project-creation-scope` markers on line 15.

The studioweb flavor **replaces** that block (`skill-flavors/studioweb/uipath-maestro-flow/SKILL.md`) because that host has no `uip` CLI. I built it to confirm:

```
$ node scripts/compose-skill-flavor.mjs build
$ sed -n '/^## Capabilities/,/^## Capability router/p' \
    build/skills/studioweb/uipath-maestro-flow/SKILL.md

- **Author** — ... create projects with `uip maestro flow init`; ...
- Create a new Flow project through Studio Web's solution-level project-creation capability   <- the override
```

The composed studioweb skill now contradicts itself one line apart. Base composes clean, with neither string present.

`CONTRIBUTING.md:246` is explicit: *"Keep the complete default behavior in the canonical file and wrap the smallest differing passage."* Duplicating the passage outside the wrapper defeats it.

**Fix:** drop `create projects with `uip maestro flow init`` from this bullet. It belongs only inside the markers.

skills/uipath-maestro-flow/SKILL.md
```diff
@@ -1,112 +1,84 @@
...
-- Explore available node types via the registry
-- Validate or format a Flow file locally
-- Apply Node ownership — Edit/Write for user-owned nodes, CLI for CLI-owned nodes (see Author capability)
-- Configure connector, connector-trigger, or managed HTTP nodes; scaffold inline-agent projects
-- Add IxP / document-extraction nodes to a flow, or list IxP models available to Maestro Flow through the flow registry
-- Plan a complex flow before building
-
-**Operate** — publishing, running, or managing a deployed flow. Read [references/operate/CAPABILITY.md](references/operate/CAPABILITY.md).
-
+- **Operate** — Publish, run, and manage deployed flows; push to Studio Web with `uip solution upload`; deploy to Orchestrator with `uip maestro flow pack` plus `uip solution publish`; debug real systems, trigger processes, inspect jobs/traces, and pause, resume, cancel, or retry instances. Read [references/operate/CAPABILITY.md](references/operate/CAPABILITY.md).
```
rockymadden: 🟠 **Major — same `skill-flavor` leak.** `push to Studio Web with `uip solution upload`; deploy to Orchestrator with `uip maestro flow pack` plus `uip solution publish`` is duplicated here and inside the `upload-scope-bullets` markers on lines 19-20.

The studioweb flavor **empties** that block entirely. Composed output at this PR's head:

```
- **Operate** — ... push to Studio Web with `uip solution upload`; deploy to Orchestrator
  with `uip maestro flow pack` plus `uip solution publish`; ...
```

That is exactly what emptying the block was meant to prevent.

**Fix:** remove both CLI clauses from this bullet, leave them in the markers.

skills/uipath-maestro-flow/SKILL.md
```diff
@@ -1,112 +1,84 @@
...
-   3. **Only then create/scaffold** — scaffold an inline agent, mock, or create-new-resource only when both searches return no match AND either the user explicitly asks to embed/inline/create, or no published resource can satisfy the requirement.
+1. **Use `--output json`; prefer `--output-filter` for extraction.** Filters are global and run against the `Data` envelope, so expressions start at `Data` without a `Data.` prefix. Registry search returns a flat PascalCase array (`NodeType`, `DisplayName`, `Description`, `AvailableOnTenant`), not `Data.Nodes` or lowercase fields. Example: `uip maestro flow registry search <keyword> --output json --output-filter "[*].{NodeType:NodeType,DisplayName:DisplayName,Description:Description,AvailableOnTenant:AvailableOnTenant}"`. With `--local`, omit `AvailableOnTenant`. Use `python3 -c` or `jq` only after verifying shape and when JMESPath cannot express the transform. See [cli-conventions.md §3](references/shared/cli-conventions.md#3-prefer---output-filter-for-extraction).
+2. **Do not run `flow debug` without explicit user consent.** It can cause real side effects, including emails, messages, and API calls.
```
rockymadden: 🟡 **Minor — hedging weakens a consent gate.**

- base: "debug **executes the flow for real** (sends emails, posts messages, calls APIs)"
- here: "It **can** cause real side effects"

Rule 2 is a hard consent gate. "Executes for real" is load-bearing; "can cause" reads as a possibility the agent may weigh. Same pattern on rule 6 ("auto-scaffolds a duplicate" -> "a duplicate **may be** created").

**Fix:** restore the indicative phrasing. It costs ~4 tokens.

skills/uipath-maestro-flow/SKILL.md
```diff
@@ -1,112 +1,84 @@
...
-12. **Node and edge `id`s MUST start with a letter — never a bare UUID.** Edges: `edge_<sourceNodeId>_<sourcePort>_<targetNodeId>_<targetPort>`. Nodes: descriptive camelCase. UUIDs only for the top-level flow `id`/`entryPointId`.
+7. **Narrate progress only when requested or clearly opted into.** Otherwise work silently and surface decisions, failures, consent gates, and the final result. When engaged, use one short plain-English line per logical step across CLI calls, shell builtins, edits, and searches; do not narrate flags or JSON structure. See [shared/ux-narration-and-todos.md](references/shared/ux-narration-and-todos.md) §When to engage.
+8. **Maintain a user-facing progress list only when tracking or verbosity is requested.** In silent mode no visible todo list is required. When engaged, journeys above trivial complexity get granular step-level todos; counts follow actual work, not a target. Hide registry lookups, parsing, and file reads inside their logical step. See [shared/ux-narration-and-todos.md](references/shared/ux-narration-and-todos.md) for triggers, granularity, thresholds, and pivots.
```
rockymadden: 🟡 **Minor — prohibition softened into a non-requirement.**

- base: "In silent mode **there is no** user-facing todo list (the agent MAY track privately)"
- here: "In silent mode no visible todo list **is required**"

"There is none" and "is not required" are different instructions. The second permits the behavior the first forbids.

**Fix:** restore "there is no user-facing todo list".

skills/uipath-maestro-flow/SKILL.md
```diff
@@ -1,112 +1,84 @@
...
+10. **Batch independent tool calls and chain dependent CLI calls.** A typical greenfield build is three turns: T1 scaffold, pull the registry, and add CLI-owned nodes in one chained `Bash`, alongside independent registry/file reads; T2 read the scaffold while editing/adding the End node and edges; T3 chain configure, validate, and format. Split only when later work depends on stdout or a mutation. See [author/references/greenfield.md — Three-turn execution map](references/author/greenfield.md#three-turn-execution-map).
```
rockymadden: 🟡 **Minor — stale link text, flagged by `check-link-text.sh`.**

```
BROKEN ./SKILL.md:69 -> text="author/references/greenfield.md — Three-turn execution map"
                        url=references/author/greenfield.md#three-turn-execution-map
                        reason=prefix-mismatch
```

The label is a pre-#2879-flatten path that no longer exists; the URL is correct. Five more like it elsewhere in the PR (base scored 0 / 546).

Separately, this **anchor** is fine but four others in this file are not — see the review body.

**Fix:** `author/references/greenfield.md` -> `author/greenfield.md`.

skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -1,638 +1,324 @@
...
+- **Gaps:** use `core.action.http.v2` manual mode when no connector exists; use `--local` resources when unpublished but in solution; use `core.logic.mock` when a resource is neither in solution nor published; flag connectors lacking connections after verified empty results.
 
-> **Run `registry get` for OOTB action nodes during discovery; defer for connector and resource nodes.** OOTB nodes (HTTP, Script, Transform, queue actions, etc.) have no `--connection-id` dependency — fetch their full schemas now so the planned topology references real ports and fields. Connector field metadata (required fields, enums, reference resolution) requires `registry get --connection-id` and belongs to Phase 2; resource schemas (RPA, agent, flow, API workflow) require `--local` or published resolution and also belong to Phase 2. `is connections list --all-folders` is enough to confirm connector connection availability in this phase.
-
----
+Run `registry get` for OOTB actions during discovery. Defer connector `registry get --connection-id` and resource `registry get --local` or published resolution to Phase 2.
 
 ## Plugin Index
```
rockymadden, line 84: 🟠 **Major — the voice plugin (#2379) is no longer discoverable.**

Plugin Index dropped from 25 node types to 20. Missing:

- `core.trigger.voice`
- `uipath.agent.voice`
- `uipath.conversational.voice.create-outgoing-call`
- `uipath.conversational.voice.end-call`
- `core.logic.mock` (named by `SKILL.md` rule 3; survives only in the heuristics section)

The PR body says #2379 "ships uncompressed". Its own plugin files do. Its integration points did not: this index, the Standard Port Reference below, `planning-impl.md` (4 voice mentions -> 0), `brownfield.md` ("Add voice nodes" row), and the `callContext` row in `node-output-wiring.md`.

`check-orphans.sh` still passes only because `author/CAPABILITY.md` (untouched here) links the plugin dir. But the planner reads *this* file. With no index row and no port row, it can never select a voice node. The feature ships and is unreachable.

**Fix:** restore the five rows here and the four port rows below from `origin/main`.

skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -1,638 +1,324 @@
...
 
-| Node Type | When to Select |
-| --- | --- |
-| `core.logic.mock` | Step is TBD, resource doesn't exist yet, or prototyping. Placeholder with `input` -> `output` |
-
----
+Use `core.logic.mock` for TBD steps, missing resources, or prototypes; it has `input` -> `output`.
 
-## Selecting External Service Nodes
+## Selecting External Services
```
rockymadden: 🟠 **Major — heading rename breaks `SKILL.md` rule 3.**

`## Selecting External Service Nodes` -> `## Selecting External Services`, so `#selecting-external-service-nodes` is now dead. Inbound:

- `SKILL.md` **rule 3** (the connector -> managed HTTP -> RPA ladder)
- `references/author/planning-impl.md`

The section content itself is faithful. Only the anchor moved.

**Fix:** keep the original heading text, or update both inbound links. The skill's own `.maintenance/check-anchors.sh` catches this: base `anchors_bad=0`, this PR `anchors_bad=37`.

skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -1,638 +1,324 @@
...
-1. **Pre-built Integration Service connector** — Use when a connector exists and covers the use case. See [connector](plugins/connector/planning.md).
-2. **Managed HTTP Request** (`core.action.http.v2`) — connector mode: use when a connector exists but lacks the specific curated activity. Manual mode: use for one-off API calls to services without connectors. See [http](plugins/http/planning.md).
-3. **RPA workflow node** — Use only when the target system has no API (legacy desktop apps, terminals). See [rpa](plugins/rpa/planning.md).
-
----
+1. A curated Integration Service connector ([connector](plugins/connector/planning.md)).
+2. `core.action.http.v2` connector mode when the connector lacks the activity, or manual mode for APIs without connectors ([http](plugins/http/planning.md)).
+3. An RPA workflow only when there is no API, such as a desktop app or terminal ([rpa](plugins/rpa/planning.md)).
 
 ## Standard Port Reference
```
rockymadden, line 168: 🟠 **Major — Standard Port Reference lost the voice node rows.**

35 rows -> 31. Missing `core.trigger.voice`, `uipath.agent.voice`, `uipath.conversational.voice.create-outgoing-call`, `uipath.conversational.voice.end-call`.

Same root cause as the Plugin Index above. Without port rows the planner cannot wire a voice node even if it somehow picks one.

**Fix:** restore the four rows from `origin/main`.

skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -1,638 +1,324 @@
...
 
-Prefix each with `**[REQUIRED]**` or `**[OPTIONAL]**`:
-
-- **[REQUIRED]** Which Slack channel should notifications go to?
-- **[OPTIONAL]** Should the error handler retry before terminating?
-
----
+Prefix each item with `**[REQUIRED]**` or `**[OPTIONAL]**`. Include unresolved user decisions and verified missing connections/resources.
 
 ## Mermaid Validation Rules
```
rockymadden, line 290: 🟡 **Minor — the mermaid rules lost their counter-examples.**

The rules survive; the *wrong* forms they exist to prevent do not:

| Dropped | Was in |
|---|---|
| `A -->success B`, `A --success--> B` | edge-label rule |
| `([text])` stadium, `{{text}}` hexagon | shape rule |
| `fetchData` not `fetch-data` or `fetch.data` | node-ID rule |
| "replace `>` / `<` with words like over/under" | label rule |

The PR body promises "every worked example" preserved. For LLM-generated mermaid the counter-example is the part that changes behavior. Stating "IDs contain only `[a-zA-Z0-9_]`" without "not `fetch-data`" is the compression that costs most per token saved.

Credit where due: adding "never `graph TD`" is an improvement over base.

**Fix:** re-add the negative examples inline. ~25 tokens total.

skills/uipath-maestro-flow/references/shared/cli-conventions.md
```diff
@@ -36,90 +32,78 @@ fi
...
 uip maestro flow validate <ProjectName>.flow --output json
 uip maestro flow registry list --output json
 uip maestro flow instance incidents <INSTANCE_ID> --folder-key <FOLDER_KEY> --output json
 ```
 
-> **Anti-pattern: `--format json` does NOT exist.** The flag is `--output json`. Using `--format json` produces `error: unknown option '--format'` and exit code 3 on every `uip` subcommand — not a helpful message pointing at `--output`.
+Do not use `--format json`; it does not exist and produces `error: unknown option '--format'` with exit code 3 on every `uip` subcommand. Ignore the benign `--localstorage-file` warning when it appears.
 
-The `--localstorage-file` warning that appears in some environments is benign and can be ignored.
+## 3. Extract with `--output-filter`
```
rockymadden: 🟠 **Major — rename breaks 4 inbound anchors, two of them universal rules.**

`## 3. Prefer `--output-filter` for extraction` -> `## 3. Extract with `--output-filter``, killing `#3-prefer---output-filter-for-extraction`. Inbound:

- `SKILL.md` **rule 1**
- `SKILL.md` **anti-patterns**
- `references/author/CAPABILITY.md`
- `references/author/editing-operations-json.md`

**Fix:** restore the heading text, or update all four links.

skills/uipath-maestro-flow/references/shared/cli-conventions.md
```diff
@@ -131,51 +115,45 @@ Always check `Result` first. On failure, `Message` and `Instructions` carry the
...
-Log in interactively (opens browser):
+Log in interactively by running:
 
 ```bash
 uip login
 uip login --authority https://alpha.uipath.com    # non-production environments
 ```
 
-## 6. `--folder-key` requirement
+## 6. Supply `--folder-key`
```
rockymadden: 🟠 **Major — rename breaks 3 inbound anchors.**

`## 6. `--folder-key` requirement` -> `## 6. Supply `--folder-key``, killing `#6---folder-key-requirement`. Inbound: `diagnose/CAPABILITY.md`, `operate/CAPABILITY.md`, `operate/manage.md`.

Unrelated, same file: the `< 0.3.4` legacy-prefix row (`uip flow init MyProject`) was dropped from the CLI rename table, even though `5e56da27` on this branch deliberately marked it `<!-- uip-check-skip -->` to keep it. A later merge removed it anyway.

skills/uipath-maestro-flow/references/author/editing-operations.md
```diff
@@ -1,119 +1,89 @@
 # Flow Editing Operations
 
-Strategy selection and shared concepts for modifying `.flow` files. Direct `.flow` authoring is required for non-carve-out structural edits; CLI is reserved for the documented product-managed configuration carve-outs.
+Strategy selection and shared concepts for modifying `.flow` files. Author non-carve-out structural changes directly in `.flow`; reserve CLI for documented product-managed configuration carve-outs.
 
-## Tool Selection Ladder
+## Tool Selection and Required Strategy
```
rockymadden: 🟠 **Major — rename breaks 4 inbound anchors, including `SKILL.md` rule 9.**

`## Tool Selection Ladder` -> `## Tool Selection and Required Strategy`, killing `#tool-selection-ladder`. Inbound:

- `SKILL.md` **rule 9** (node ownership, named there as the canonical per-operation ladder)
- `references/author/greenfield.md` (x2)
- `references/author/brownfield.md`

Also dropped from this file: `### Why not Python / Node / jq / sed?`. Rule 9 requires surfacing the scripting trade-offs (state bypass, opaque diff, no interruption point) before asking approval. The rule still says so; the section it pointed at is gone.

**Fix:** keep the original heading text and restore the rationale subsection.

skills/uipath-maestro-flow/references/diagnose/CAPABILITY.md
```diff
@@ -1,75 +1,73 @@
...
+> **Navigation.** Diagnose follows Operate when a run faults and points to Author for the fix. Re-running and lifecycle are in [operate/CAPABILITY.md](../operate/CAPABILITY.md); building or editing `.flow` files is in [author/CAPABILITY.md](../author/CAPABILITY.md).
 >
-> **Inherits universal rules from [SKILL.md](../../SKILL.md)** — `--output json` + prefer `--output-filter` for extraction, no `flow debug` without consent, never invoke other skills automatically, dropdown question pattern, **plain-English narration + granular progress list (opt-in — silent by default; engage when the user asks for verbosity)**. The rules below are diagnose-scoped and apply on top.
+> **Inherited rules:** use `--output json` and prefer `--output-filter` for extraction; do not run `flow debug` without consent; never invoke other skills automatically; use the dropdown question pattern; provide plain-English narration and a granular progress list only when the user asks for verbosity, and remain silent by default. These rules apply in addition to the rules below.
 
-## When to use this capability
+## When to use
```
rockymadden: 🟠 **Major — capability index no longer matches the required template.**

`.maintenance/check-template.sh` on this PR:

```
MISSING_SECTION  references/diagnose/CAPABILITY.md  ->  ## When to use this capability
MISSING_SECTION  references/diagnose/CAPABILITY.md  ->  ## Workflow
MISSING_SECTION  references/diagnose/CAPABILITY.md  ->  ## Common tasks
```

Base: `capabilities_checked=4 missing_sections=0`. This file lost all three required sections; `## When to use this capability` was shortened to `## When to use` on this line.

The compressed content underneath is fine. It is the section headings the template contract needs.

**Fix:** restore the three heading names.

skills/uipath-maestro-flow/references/evaluate/CAPABILITY.md
```diff
@@ -1,19 +1,16 @@
...
 >
-> **Inherits universal rules from [SKILL.md](../../SKILL.md)** — `--output json` + prefer `--output-filter` for extraction, no `flow debug` without consent, never invoke other skills automatically, dropdown question pattern, **plain-English narration + granular progress list (opt-in — silent by default; engage when the user asks for verbosity)**. The rules below are evaluate-scoped and apply on top.
+> **Inherits universal rules from [SKILL.md](../../SKILL.md)** — `--output json` + prefer `--output-filter` for extraction, no `flow debug` without consent, never invoke other skills automatically, dropdown question pattern, **plain-English narration + granular progress list (opt-in — silent by default; engage when the user asks for verbosity)**. The rules below apply on top.
 
 ## When to use this capability
```
rockymadden, line 9: 🟠 **Major — same template contract break.**

```
MISSING_SECTION  references/evaluate/CAPABILITY.md  ->  ## Workflow
MISSING_SECTION  references/evaluate/CAPABILITY.md  ->  ## Common tasks
```

`## When to use this capability` survives here; `## Workflow` and `## Common tasks` do not.

Two anchors into this file's neighbours also broke: `eval-sets-guide.md#--input-file-keypath` and `#--criteria`, both linked from this file, because the per-flag `###` headings in `eval-sets-guide.md` were collapsed.

**Fix:** restore the two headings here and either restore the flag headings in `eval-sets-guide.md` or repoint these two links.

skills/uipath-maestro-flow/references/author/editing-operations-json.md
```diff
@@ -324,163 +243,86 @@ Use `Edit` to add an entry to `variables.variableUpdates.<NODE_ID>`:
...
-4. Map all `out` variables on the subflow's End node `outputs`
-5. Parent-scope `$vars` are NOT visible inside the subflow — pass values via inputs
-6. Subflow node positions go in the **subflow's own** `layout.nodes` — not in the top-level `layout.nodes`. Each subflow has an independent layout scope.
-
-See [subflow/impl.md](plugins/subflow/impl.md) for the full JSON structure and rules.
-
----
+**Tool:** `Edit` (or `Write` when scaffolding from a template).
+
+1. Add a `core.subflow` parent with `typeVersion: "1.0.0"`, `display`, inputs, and error output:
```
rockymadden: 🟡 **Minor — `check-versions.sh` PROSE-PIN violation (real one).**

```
BAD ./references/author/editing-operations-json.md:295  PROSE-PIN
    version stated as current/required; reference `registry get`/`flow init` instead
```

Base carried `"typeVersion": "1.0.0"` for `core.subflow` **only inside the example JSON block** (fences are skipped by the checker). This rewrite hoisted it into the prose sentence, which the repo rule forbids.

Base scored `violations=0` on 13,316 lines.

**Fix:** drop the version from the sentence. Leave it in the JSON example below.

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
```diff
@@ -66,55 +57,46 @@ For add, delete, and wiring procedures, see [editing-operations.md](../../editin
...
-- `$vars.{nodeId}.status` — selected outcome name (e.g. `"Approve"`, `"Reject"`)
-- `$vars.{globalId}` — workflow-global alias; `globalId` is `field.variable` with `vars.` stripped. **Do not use this in scripts — always use `$vars.{nodeId}.output.{fieldId}` instead.**
-
----
+Rules:
+- Input fields use `binding: "vars.<nodeId>.output.<field>"` (raw path; no `=js:$` prefix) and no `variable`.
+- Output fields use `variable: "vars.<globalName>"` (`vars.` required) and no `binding`.
+- InOut fields use both properties in those formats.
+- Use `schemaId` (not `id`) at schema level and generate a fresh UUID.
+- Use `typeVersion: "1.0"` for this node. **Do not run `registry get` to derive it; do not use `"1.1"` or another version.**
```
rockymadden: 🟡 **Minor — `check-versions.sh` PROSE-PIN violation (phrasing only).**

```
BAD ./references/author/plugins/hitl/impl.md:65  PROSE-PIN
```

Worth being precise: the **substance is unchanged**. Base line 74 said the same thing and passed:

> `typeVersion` — always `"1.0"` for `uipath.human-in-the-loop.quick-form`. **Do not run `registry get` to derive this value; do not use `"1.1"` or any other version.**

The compressed opener "Use `typeVersion: "1.0"` for this node" is what trips the checker's pattern, not the meaning.

**Fix:** restore the base phrasing, or add a skip marker. Your call which.

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -1,169 +1,119 @@
...
 uip is resources list "<connector-key>" --connection-id "<connection-id>" --output json \
   --output-filter "[?contains(DisplayName,'<search>')].{Name:Name,Path:Path,Custom:Custom}"
 ```
 
-- `Name` — what `--detail.objectName` accepts. Case-sensitive (e.g. `"Opportunity"`, not `"opportunity"`). Don't substitute `DisplayName`.
-- `Path` — API endpoint suffix (e.g. `/Opportunity`). Pair with the operation's HTTP method (Step 3) for `endpoint`.
-- `Custom` — `true` for tenant-defined objects.
+Use case-sensitive `Name` as `objectName`, never `DisplayName`. `Path` is the endpoint suffix and `Custom` indicates tenant-defined objects.
+
+### Step 3 — Describe the resource and read the cache
```
rockymadden, line 91: 🟡 **Minor — Step 3 lost the examples that prevent its documented failure mode.**

Base spelled out the guard against deriving `objectName` from the node-type suffix:

- `…google-gmail.send-email` -> objectName `SendEmail`, **not** `send_email`
- `…teams.send-bot-direct-message` -> objectName `bot_direct_messages`

Both gone (`send_email`, `SendEmail`, `bot_direct_messages`: 1 file in base, 0 in head). The rule "never derive it by case conversion" survives, but kebab->snake is precisely the guess an agent makes, and the worked pair is what stops it.

Same in the `--operation` note: `curated_get_issue` -> `GETBYID` and `QueryEntityRecordsCurated` -> `POST` are gone.

Also dropped from the 4.0.0 section just above: the cross-link to `uipath-platform — resources.md § --activity-version`.

**Fix:** re-add the two objectName pairs and the two `--operation` pairs.

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -192,560 +142,190 @@ The `<id>` in `--connection-id "<id>"` MUST be the connection bound to **this**
...
+uip maestro flow node configure <file> <nodeId> --detail "$(cat /tmp/detail.json)" --output json
+```
 
-| Match in token | Encoded as |
-|---|---|
-| `:::` | `_sub_` |
-| `[*]` | `_array` |
-| `::` | `_sub_` |
-| `.` | `_sub_` |
+#### Step 6c — Populate api-type custom fields
```
rockymadden, line 223: 🟡 **Minor — Step 6c went 157 lines -> 22 and dropped every worked example.**

`SKILL.md` anti-patterns point readers here by name ("See connector/impl.md Step 6c"). What is gone:

- **All three `--detail` payloads** (Jira Create Issue, Snowflake Execute Query, Dataservice V3). These were the only demonstrations of the raw-key-in-`bodyParameters` vs encoded-key-in-cache pairing that the surviving prose describes abstractly.
- The 15-row api-type support table (`curated_create_issue`, `curated_soqlQuery`, `CreateEntityRecordCurated`, `list_members_curated_dynamic::members`, …).
- The encoding examples: `fields.project.key` -> `fields_sub_project_sub_key`, `items[*]` -> `items_array`. The substitution table survives; nothing shows it applied.
- Error signature `DAP-DT-_2003 refField with name <X> not found`, the tell for dropping the runtime bucket.
- `GenerateSchema` / `generateSchema` / `FetchObjectMetadataTenant` as concrete `objectActionName` values (all 1 file in base, 0 in head).

The PR body claims it preserves "every rule, CLI, enum, threshold, and worked example that drives agent behavior". That is not accurate for this section.

**Fix:** restore the three payload examples at minimum. They are the highest-value tokens in the file.

skills/uipath-maestro-flow/references/author/plugins/inline-agent/impl.md
```diff
@@ -1,176 +1,105 @@
...
+This skill covers flow-specific operations for low-code inline agent nodes: scaffolding, JSON structure, wiring, resources, and validation. For `agent.json`, `resource.json`, solution resources, prompts, and agent-side configuration, use the `uipath-agents` skill, especially `lowcode/capabilities/inline-in-flow/inline-in-flow.md`.
 
-Node type: `uipath.agent.autonomous`. The agent is bound to a local subdirectory via `inputs.source = <projectId>`. The node's BPMN type and `serviceType` (`Orchestrator.StartInlineAgentJob`) come from the definition in `definitions[]`.
+Node type: `uipath.agent.autonomous`. Bind the agent directory with `inputs.source = <projectId>`. BPMN type, `serviceType` (`Orchestrator.StartInlineAgentJob`), version, and context come from its `definitions[]` entry. Coded Python agents use the [`agent`](../agent/impl.md) plugin (`uipath.core.agent.{key}`); inline agents are low-code only.
 
-For coded (Python) agents, use the [`agent`](../agent/impl.md) plugin (`uipath.core.agent.{key}`) — inline agents are low-code only.
+## 1. Scaffold and configure the inline agent
```
rockymadden: 🟠 **Major — renumbering these headings broke 6 inbound anchors, 4 of them from another skill.**

Sections became `## 1.` … `## 8.`, killing `#registry-validation`, `#adding-resource-nodes`, `#configure-agentjson`, `#refresh-and-validate`, `#debug`, `#wiring-flow-variables-into-agent-prompts`.

In-skill inbound: `inline-agent/planning.md` (x2), `inline-voice-agent/impl.md` (x4), `shared/node-output-wiring.md`.

**Cross-skill inbound, outside this skill's checker scope:**

- `skills/uipath-agents/references/lowcode/capabilities/inline-in-flow/inline-in-flow.md` -> `#wiring-flow-variables-into-agent-prompts`, `#debug`, `#refresh-and-validate`
- `skills/uipath-agents/references/lowcode/prompting/autonomous-agent-prompting-guide.md` -> `#wiring-flow-variables-into-agent-prompts`

`CONTRIBUTING.md:63` says skills are self-contained and cannot depend on each other, but `uipath-agents` links in here today and this PR breaks those links.

**Fix:** keep the original heading text (numbering is fine as a prefix only if the slugs are preserved, which they are not here), or fix all 10 inbound links including the two files in `uipath-agents`.

skills/uipath-maestro-flow/references/shared/ux-narration-and-todos.md
```diff
@@ -1,152 +1,115 @@
...
+
+## Narration lines
+
+Keep each line to approximately 15 words or fewer, as one sentence or fragment. Use system voice, not first person. Use user terms rather than commands, flags, or JSON internals. State the actual subject when scope is ambiguous, convey new information, and avoid ceremony, recaps, and repetition.
+
+Use or adapt these patterns:
+
+| Step | Narration |
+|---|---|
+| Login probe | “Checking whether you're logged in to the UiPath tenant…” |
```
rockymadden: 🟡 **Minor — curly quotes introduced, base had zero.**

Bundle-wide: **0 -> 44** curly quotes across 5 files. The repo is otherwise 100% straight-quoted. This table is 30 of them.

Apostrophes stayed straight (`“Checking whether you're logged in…”`), so the file is now mixed.

The instances that actually matter are the literal strings the agent is told to render:

- `SKILL.md:62` — `“Create a new solution,”` and `“Something else.”` The rewrite moved the comma and period **inside** the quotes, changing the literal option labels rule 5 requires. Base had `**"Create a new solution"**` / `**"Something else"**`.
- `ux-narration-and-todos.md:114` — same to `“Let me,”` / `“I'll go ahead,”`.

**Fix:** `sed` the 44 back to straight quotes, and pull punctuation outside the quotes on the option labels.

rockymadden (changes requested):
## Summary

Whole-bundle compression of `uipath-maestro-flow`. The compression method is sound and the reported numbers are honest. I reproduced them independently:

| Metric | Measured (bytes) | PR claim (tokens) |
|---|---|---|
| Bundle | −19.2% | −20.2% ✅ |
| `SKILL.md` | −35.9% | −36% ✅ |

Also verified clean: file set unchanged (94 files both sides), CLI flags 94 → 92 (both "losses" are regex artifacts), `uip` command paths 134 → 133, all 781 relative file links resolve, connector `impl.md` Steps 1–6c all survive structurally, and Step 4 is restored verbatim as the PR body says. `upload-safety.md` is compressed ~40% with every rule intact — that one is a model for the rest.

The problems are not in the compression. They are in the merge seams and in heading rewrites that nothing validated.

## The skill's own maintenance suite goes from green to four failures

```
$ cd skills/uipath-maestro-flow && bash .maintenance/check-all.sh
```

| Check | Base `8b0ae9e96` | This PR | |
|---|---|---|---|
| `check-anchors.sh` | 0 bad / 267 | **37 bad** / 251 | 🔴 |
| `check-template.sh` | 0 missing | **5 missing** | 🔴 |
| `check-link-text.sh` | 0 broken / 546 | **6 broken** / 526 | 🔴 |
| `check-versions.sh` | 0 violations | **2 violations** | 🔴 |
| `check-depth.sh` | pass | pass | ✅ |
| `check-orphans.sh` | 0 | 0 | ✅ |
| `check-plugin-pairs.sh` | 0 | 0 | ✅ |
| `check-uip-commands.sh` | 0 unknown | 0 unknown | ✅ |
| **Overall** | **`All checks passed.`** | **`One or more checks failed.`** | |

None of this reaches CI. `.maintenance/` is not wired into `.github/workflows/` at all, and `scripts/check-skill-links.mjs:119` does `raw.split("#")[0]`, so the anchor half of every link is discarded. That is why the checks are green on a PR that breaks 37 anchors.

## Findings

Details are inline. Ranked:

| # | Severity | Finding |
|---|---|---|
| 1 | 🔴 **Critical** | `node-output-wiring.md:37` reverted to the pre-#2379 inline-agent prompt contract. It now forbids the `{{input.<id>__output__<var>}}` form that this same PR's `inline-agent/impl.md` mandates. Bundle contradicts itself. |
| 2 | 🟠 **Major** | 37 broken heading anchors (base: 0). Four of five universal `SKILL.md` rules point at anchors that no longer exist. |
| 3 | 🟠 **Major** | 4 more broken anchors inbound from **`uipath-agents`**, outside this skill's checker scope. |
| 4 | 🟠 **Major** | Voice plugin #2379 stripped from `planning-arch.md` Plugin Index + Port Reference, `planning-impl.md`, `brownfield.md`, and the `callContext` wiring row. Plugin ships, planner cannot reach it. |
| 5 | 🟠 **Major** | `skill-flavor` leak: `uip maestro flow init` / `uip solution upload` / `pack` / `publish` duplicated outside their markers, so the composed **studioweb** `SKILL.md` contradicts its own override. |
| 6 | 🟠 **Major** | `check-template.sh` 0 → 5. `diagnose/CAPABILITY.md` lost all three required sections, `evaluate/CAPABILITY.md` lost two. |
| 7 | 🟡 **Minor** | "every worked example preserved" is not accurate. Step 6c lost all three `--detail` payloads and the support table; Step 3 lost the objectName and `--operation` pairs; mermaid lost its counter-examples. |
| 8 | 🟡 **Minor** | `check-link-text.sh` 0 → 6, all stale pre-flatten or wrong-depth link labels. |
| 9 | 🟡 **Minor** | `check-versions.sh` 0 → 2. One real (version pin hoisted from a code fence into prose), one phrasing-only. |
| 10 | 🟡 **Minor** | 44 curly quotes introduced (base: 0), including inside literal option labels: `“Create a new solution,”` moved the comma inside the string. |
| 11 | 🟡 **Minor** | Rules 2, 6 and 8 hedged from indicative to "can / may / is required". Two of those are consent gates. |

### On finding 5, reproduce it with

```bash
node scripts/compose-skill-flavor.mjs build
sed -n '/^## Capabilities/,/^## Capability router/p' \
  build/skills/studioweb/uipath-maestro-flow/SKILL.md
```

Base composes with neither CLI string present. This PR emits both, directly above the override that says to use Studio Web's project-creation capability instead.

## On the eval results

The paired-arm methodology reads sound and the reporting is unusually candid — the `slack-channel-description` write-up in particular. I cannot reproduce the runs, so I am not disputing "0 content regressions".

But I do not think the suite covers these defects. None of the 126 tasks exercises a shared index, the studioweb flavor build, a heading anchor, or a cross-skill link. Zero regressions against a suite that cannot see the break is consistent with the break existing. Worth a second opinion from whoever owns the harness.

## Also

The branch is **4 commits behind `main`** (merge-base is #2740). #2935 and #2881 both touch this skill. The next merge is where findings 1 and 4 need re-verifying, not just re-merging — both were introduced by earlier merges of this kind.

## Suggested order

1. Restore from `origin/main`: the two `node-output-wiring.md` rows, the voice rows in `planning-arch.md` / `planning-impl.md` / `brownfield.md`.
2. Move two clauses back inside their flavor markers in `SKILL.md`.
3. Revert the renamed headings (cheaper than fixing 41 links), including the `uipath-agents` inbound ones.
4. Restore the three `CAPABILITY.md` section headings.
5. Merge `main` again.
6. Gate on `bash .maintenance/check-all.sh` printing `All checks passed.`

## Process ask

Please wire **`check-all.sh`** into CI, not just the anchor check. Four independent checks that this skill already owns would have blocked this PR on their own. The suite exists, it is green on `main`, and nothing runs it. #2884 added `check-anchors.sh` specifically to catch this class of drift, and it is sitting unused.

rockymadden (commented):
## Re-review

Verified against head `b8e34971d` by running the skill's own maintenance suite, rebuilding the studioweb flavor, diffing every flagged file against `origin/main`, and writing an out-of-scope cross-skill anchor checker.

**All 11 findings are fixed.** I could not find a workaround, a papered-over link, or a restoration that broke something else.

### Findings, verified

| # | Sev | Finding | Verified state |
|---|---|---|---|
| 1 | 🔴 | `node-output-wiring.md` pre-#2379 revert | **Fixed.** `{{input.<flowNodeId>__output__<field>}}` + brace-free `rawString` + Voice `callContext` row all present. The final merge *added* main's encoding note rather than re-reverting. |
| 2 | 🟠 | 37 broken anchors | **0.** `anchors_checked=258 anchors_bad=0` |
| 3 | 🟠 | 4 inbound anchors from `uipath-agents` | **0.** `inline-agent/impl.md` is now byte-identical to `main`, so all 10 inbound links resolve. My own cross-skill checker: 0 bad, both sides. |
| 4 | 🟠 | Voice #2379 unreachable | **Fixed.** 4 Plugin Index rows + 4 Port Reference rows restored; `planning-impl.md` 4 mentions (= main), `brownfield.md` "Add voice nodes" row back. |
| 5 | 🟠 | `skill-flavor` leak | **Fixed.** Rebuilt: composed studioweb `SKILL.md` contains zero `uip maestro flow init` / `uip solution upload` / `pack` / `publish`. Marker bodies are the only home now. |
| 6 | 🟠 | `check-template.sh` 5 missing | **0.** All three headings back in `diagnose/` and `evaluate/CAPABILITY.md`. |
| 7 | 🟡 | Worked examples | **Mostly fixed.** Step 6c: all 3 `--detail` payloads, 14-row api-type table, `DAP-DT-_2003`, `fields_sub_project_sub_key`, `items_array` back. Step 3: both objectName pairs and both `--operation` pairs back. Mermaid: all 4 counter-examples back. `< 0.3.4` legacy row back. `### Why not Python / Node / jq / sed?` back. Residual gaps below. |
| 8 | 🟡 | `check-link-text.sh` 6 broken | **0.** `checked=538 broken=0` |
| 9 | 🟡 | `check-versions.sh` 2 violations | **0.** `lines_checked=11899 violations=0` |
| 10 | 🟡 | 44 curly quotes | **Partial.** Curly *doubles* 0. 11 curly apostrophes remain. |
| 11 | 🟡 | Hedged consent gates | **Fixed.** Rule 2 "It executes the flow for real", rule 8 "there is no user-facing todo list", rule 6 "Never omit `cd`". Option labels straight-quoted with punctuation outside. |

`bash .maintenance/check-all.sh` → **`All checks passed.`** Same 9 checks green on `main`. GitHub CI all green.

### Numbers, re-measured

| Metric | My measurement (bytes) | PR body claim (tokens) |
|---|---|---|
| Bundle | **−13.5%** (911,867 → 789,174) | −13.0% ✅ |
| `SKILL.md` | **−28.8%** (25,457 → 18,119) | −28.4% ✅ |

Preservation sweep, `main` vs head:

- **128 / 128** long flags kept, zero loss
- **83 / 83** files, no adds or drops
- 3 apparent node-type losses, all benign: `curated_create_issue` still present (regex artifact), the concrete agent GUID became `<AGENT_UUID>` (an improvement), `core.action.slack.send` sat in a dropped narration counter-example
- 22-row connector error table → 22 bullets, **all 22 preserved**
- 8 ASCII topology diagrams → arrow one-liners, **all 8 shapes preserved**

Branch is 6 commits behind `main`, **none touching this skill**. Low-risk merge.

---

## Comments

### A. 🟠 Major: the eval evidence no longer describes this candidate

The PR body's own last section is a plan, not a result:

> **After merge and review test** — Re-run full maestro-flow suite after review to confirm the post changes do not create regressions.

Nothing has been posted. Two things changed after the validated candidate:

1. The restorations: **+1,809 tok across ~15 files**, including `connector/impl.md` Step 6c (the file whose Step 4 trim caused the original `slack-channel-description` regression).
2. The final merge took **main's uncompressed rules 3 and 4** verbatim into `SKILL.md`, which is why the body went 3,383 → 4,013 tok. Rule 4 is now a ~200-word IXP-handoff rule that no arm has been graded against.

The entire "0 confirmed regressions, candidate nets 245 vs 226" case was measured on a bundle that no longer exists.

**Fix:** run the stated 2× Luna / 1× Terra / 1× Sonnet and post the result. I have no other basis to approve, and this is the only thing blocking me.

### B. 🟡 Minor: `connector/impl.md` dropped 3 literal error strings

Table → bullets is fine, but the symptom half got abstracted into labels. Gone from the bundle entirely:

| Dropped literal | Now reads |
|---|---|
| `Expected a field name expression but got '…'` (2 → 0) | "CEQL `[102003]` field-name error" |
| `customFieldsRequestDetails.parameterValues must be an array of [key, value] tuples` (1 → 0) | "`parameterValues` object-map error" |
| `customFieldsRequestDetails has unknown keys: ObjectActionName, ParameterValues` (1 → 0) | "Unknown `ObjectActionName` or `ParameterValues`" |

Diagnosis works by pasting an error and string-matching it against the reference. A label the user will never see in their terminal is not matchable. `Unsupported value expression 'Column'` and `DAP-DT-_2003` were kept, so the pattern is inconsistent rather than deliberate.

**Fix:** put the literal error text back in the symptom half of those 3 bullets. ~30 tokens.

### C. 🟡 Minor: duplicated sentence in `ixp/impl.md:216`

> …describing the violation **— fix the `.flow` file, not the validator. Fix the `.flow` file, not the validator** (e.g. the common `inputs.model must be…`)

Compaction artifact from the last merge. Everything else in that file's 10 changed lines checks out: the example `nodeType`, the `digitizationMode` → `fileUpload` rationale, and the full `{ id, modelName, modelDisplayName, folderKey, folderName, folderPath, description }` blob all survive elsewhere in the file.

**Fix:** delete the first clause.

### D. 🟡 Minor: two flag anchors resolve to one section

`eval-sets-guide.md:43-44` stacks both anchors on the same heading:

```
<a id="--criteria"></a>
<a id="--input-file-keypath"></a>
### Inputs, Expected Values, and Criteria
```

The checker is satisfied and the links resolve. But `evaluate/CAPABILITY.md:89` says "Add a data point with file attachments → `--input-file`" and lands the reader on a section titled "Inputs, Expected Values, and Criteria". `--input-file` is the attachment flag; it is the one a reader most needs its own landing spot for. `--search-text`, `--inputs`, `--expected`, `--evaluators`, `--entry-point` lost their headings too, but nothing links those.

**Fix:** give `--input-file` back a `###` of its own, or repoint that one link to `### File and Contains Inputs` (line 79), which is where the content actually is.

### E. 🟡 Minor: `evaluate/commands-reference.md` went 9 fences → 1

Two losses worth weighing:

- The **command tree** (18 subcommands in one ASCII block). Every command still has its own `###`, so nothing is unreachable, but the one-glance map of the surface is gone.
- The only **worked `eval add` example**: `--inputs '{"name":"Alice"}' --expected '{"greeting":"Hello, Alice!"}'`. The flag table says "Input values as a JSON object; keys must be declared as Flow input variables". Nothing shows it. Same class as finding 7.

Credit: the `--type` enum is correct here, `llm-judge-` prefix intact. That was the corruption caught adversarially, and it stuck.

**Fix:** restore the `eval add` example. The command tree is your call.

### F. 🟡 Minor: 11 curly apostrophes remain

Base had 0 of any kind. Curly doubles are now 0, so the sed pass ran but missed `’`:

| File | Count |
|---|---|
| `plugins/agent/impl.md` | 6 |
| `editing-operations-json.md` | 3 |
| `plugins/merge/planning.md` | 2 |

**Fix:** one more sed.

### G. 🟡 Minor: `## When to use this skill` → `## Capabilities`

Nothing validates SKILL.md headings, and the section always was a capability list, so the new name is arguably more honest. But 20 of 27 skills in the repo use "When to Use This Skill" and this PR is the only divergence.

Also dropped from the Evaluate bullet: the evaluator type enum `(exact-match, json-similarity, contains, llm-judge-*)`. It is correct in `commands-reference.md`, so this is a summary-level trim, not a rule loss.

**Fix:** rename back, or say in the body that the divergence is intentional so the next reviewer does not re-flag it.

### H. 🟡 Minor: stale PR title

Title says **−20.7%**. Body says −13.0%. My measurement says −13.5%. The body is right.

**Fix:** retitle to −13.0%.

### I. 🟡 Minor: process ask not addressed

Unchanged at head:

- `.maintenance/check-all.sh` appears in **no** file under `.github/workflows/`
- `scripts/check-skill-links.mjs:119` still `raw.split("#")[0]`, so the anchor half of every link is still discarded

Fair to split into its own PR. Say so in the body if you plan to, so it does not get lost. Four checks this skill already owns would have blocked the original PR unaided, and the next bundle-compression PR walks into the same gap.

---

## Worth saying

The restoration pass was thorough and honest. It did not chase the checkers: renamed headings went back to their original text rather than getting 41 links rewritten, and the merged sections got explicit `<a id>` anchors instead of being papered over. The `+1,809 tok` cost is stated in the body rather than buried. `check-all.sh` green on a bundle this size, with the file set unchanged and 128/128 flags intact, is real work.

The bundle content is in good shape. What is missing is evidence that the *current* bundle behaves.

## tl;dr

All 11 findings are genuinely fixed. I re-ran the checks and rebuilt the flavor myself: everything is green and the numbers hold up (−13.5% bundle by my count, matching the body).

One thing still blocks me: the eval results in the body were measured on the *old* candidate. The fix commits and the last merge changed the bundle after that, and the body promises a re-run that has not been posted. **Post those results and I approve.** Items B through I are polish, roughly 15 minutes total.

CarlesUIPath:
Thank you very much @rockymadden for the thorough review. The requested changes have been addressed and indicated in the PR body. 

On your main blocker (re-validate on the current bundle): I re-gated the merged candidate against current main nightlies: sonnet 107 vs 110 (parity within single-seed noise) and terra 119 vs 119 (dead even), 0 systematic regressions; every candidate-worse task is in the known live/e2e flaky families. 

Unfortunately, we could not reliably reproduce token savings on the merged version. One plausible explanation is that the main body + reference is not enough for maestro-flow tasks which require multi-turn builds where context is dominated by cache-read over many turns.

rockymadden (approved):
Thank you!

PR #2925, opened 2026-08-31
fix(api-workflow): guard loop output against empty collections; fix run_execute grader

rares-baesu-uipath:
Two fixes from the 2026-08-27 codereval failures on `uipath-api-workflow`. Rebased on main (the `run_execute.yaml` conflict with #2953 is resolved by the new grader design below).

## 1. Skill bug: loop buckets are undefined after zero iterations

The `outputs.For_Each_N` bucket is created by the loop body's export. If the loop runs zero times, the bucket never exists, and **any** read of it crashes the run with `Cannot read properties of undefined (reading 'For_Each_N')` — the loop's own `output.as`, but equally a Response, `for.in`, or Assign downstream that reads `$context.outputs.For_Each_N.results`. A trailing `?? []` does not help; `.results` already threw.

This is what failed the `diagnose-unsupported-activity-type` eval: the agent copied the skill's pattern faithfully, and count=0 crashed.

**Fix:** guard every read of a loop bucket with `?.`:
- `output.as` → `${$context.outputs?.For_Each_N}` in `task-types.md` and all three loop templates (DoWhile too)
- downstream reads → `${$context.outputs?.For_Each_N?.results ?? []}`: the loop-aggregation template's `for.in` over `Do_While_1.results` and its Response's `classified`; task-types note + nesting bullet now state the rule; new troubleshooting entry for the error

**Verified on uip 1.202.0:** unguarded Response + `items: []` → `Failure … reading 'For_Each_1'`; guarded → `Success {Results: []}`; `items: [1,2,3]` → `{Results: [2,4,6]}` either way. All three templates validate; the loop-aggregation template runs.

## 2. Test bug: `run_execute` graded the run count, not the runs

The `command_executed` criterion counts Bash tool calls, not pattern occurrences. `min_count: 2` was a false negative for two runs chained with `&&` in one call; relaxing to `min_count: 1` (as this PR first did, and as #2953 did on main) is a false positive — one real run plus a hand-written `run_40.json` passes the envelope check and scores 100%, which is grading a self-report.

**Fix:** two `command_executed` criteria, one per input, each anchoring its score inside the same command segment (README segment idiom). Chained runs count once each; a fabricated second file fails the score-40 criterion.

## Checks

- `npm run skills:validate` and `npm run skills:build` pass
- CLI verb check: clean
- Regex checked against 6 chained / single / fabricated / `--input-file` / `840` samples

### DIFF
--- a/skills/uipath-api-workflow/assets/templates/conditional-workflow-example.json
+++ b/skills/uipath-api-workflow/assets/templates/conditional-workflow-example.json
@@ -147,7 +147,7 @@
                         }
                       }
                     ],
-                    "output": { "as": "${$context.outputs.For_Each_1}" },
+                    "output": { "as": "${$context.outputs?.For_Each_1}" },
                     "metadata": { "activityType": "ForEach", "displayName": "Process Transactions", "fullName": "ForEach" }
                   }
                 },
--- a/tests/tasks/uipath-api-workflow/operate/run_execute/run_execute.yaml
+++ b/tests/tasks/uipath-api-workflow/operate/run_execute/run_execute.yaml
@@ -31,17 +31,27 @@ initial_prompt: |
   Do NOT ask for approval or confirmation.
 
 success_criteria:
+  # One criterion per input, each anchored on its score inside the SAME command
+  # segment (the lookahead stops at newline / && / || / ; / | / the next uip), so
+  # two runs chained in one Bash call still count once each, and a hand-written
+  # run_40.json cannot stand in for the score-40 run (CONTRIBUTING.md: do not
+  # score self-reports). A single pattern with min_count: 2 was a false negative
+  # for && chaining (2026-09-01); min_count: 1 was a false positive for
+  # one-real-run-plus-fabrication.
   - type: command_executed
-    # min_count 1, not 2: the matcher runs one search per Bash tool call, so
-    # both runs chained with && in a single call count once (the 2026-09-01 run
-    # scored a false 1/2 that way). One visible invocation still anchors that
-    # the agent drove the runner; per-input evidence lives in the envelope
-    # checks on run_85.json / run_40.json below.
-    description: "Agent executed the workflow with the local runner (both runs may share one chained Bash call)"
+    description: "Agent ran the workflow with the local runner for score 85"
     tool_name: "Bash"
-    command_pattern: '(uip|\$UIP)\s+api-workflow\s+run'
+    command_pattern: '(uip|\$UIP)\s+api-workflow\s+run(?=(?:(?!\n|&&|\|\||;|\||\s(?:uip|\$UIP)\s).)*(?<![A-Za-z0-9])85(?![0-9]))'
     min_count: 1
-    weight: 2.0
+    weight: 1.0
+    pass_threshold: 1.0
+
+  - type: command_executed
+    description: "Agent ran the workflow with the local runner for score 40"
+    tool_name: "Bash"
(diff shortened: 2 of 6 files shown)


skills/uipath-api-workflow/assets/templates/conditional-workflow-example.json
```diff
@@ -147,7 +147,7 @@
                         }
                       }
                     ],
-                    "output": { "as": "${$context.outputs.For_Each_1}" },
+                    "output": { "as": "${$context.outputs?.For_Each_1}" },
```
liviubobocu, line 150: Guarding the loop's own output.as doesn't help any downstream consumer of the bucket, which fails on the identical input:

Response reads ${$context.outputs.For_Each_1.results}, items: []
→ "Result": "Failure" ... Cannot read properties of undefined (reading 'For_Each_1')
Guarded (${$context.outputs?.For_Each_1?.results ?? []}) → Success {seen: 0, results: []}.

tests/tasks/uipath-api-workflow/operate/run_execute/run_execute.yaml
```diff
@@ -32,10 +32,13 @@ initial_prompt: |
 
 success_criteria:
   - type: command_executed
-    description: "Agent executed the workflow with the local runner, once per input (the operate action)"
+    # min_count is 1, not 2: the counter counts Bash tool calls, not pattern occurrences,
```
liviubobocu: min_count: 1 trades a false negative for a false positive, and the stated rationale doesn't hold (medium)

The comment says the file_exists + run_command criteria "already prove both runs happened" — they don't. grep '"FAIL"' run_40.json passes on a file the agent hand-wrote. An agent that runs score 85 once and authors run_40.json by hand now scores 100%; min_count: 2 incidentally caught that, and CONTRIBUTING.md:395 ("do not score self-reports") is the rule at stake.

skills/uipath-api-workflow/assets/templates/conditional-workflow-example.json
rares-baesu-uipath: Right — reproduced on 1.202.0: unguarded Response `${{ results: $context.outputs.For_Each_1.results ?? [] }}` with `items: []` → Failure `Cannot read properties of undefined (reading 'For_Each_1')`; guarded `$context.outputs?.For_Each_1?.results ?? []` → Success `{Results: []}`. Fixed in 66ef5e253: the loop-aggregation template's two downstream reads (the For_Each `for.in` over `Do_While_1.results` and the Response's `classified`) are guarded, the task-types note and nesting bullet now say every downstream read of a loop bucket needs `outputs?.X?.results ?? []` (not just `output.as`), and troubleshooting has an entry for the error. Templates still validate and the loop-aggregation one runs.

tests/tasks/uipath-api-workflow/operate/run_execute/run_execute.yaml
rares-baesu-uipath: Agreed, the rationale was wrong — a hand-written `run_40.json` passes the envelope check. In 66ef5e253 the single pattern is replaced by two `command_executed` criteria, one per input, each anchoring its score inside the same command segment (`(?=S*85)` / `(?=S*40)` with the README segment idiom). Two runs chained with `&&` in one Bash call now count once each (the 2026-09-01 false negative), and one real run plus a fabricated second file fails the score-40 criterion (the false positive you describe). Main had already relaxed to `min_count: 1` via #2953, so this supersedes that too.

PR #2930, opened 2026-08-31
fix(skill-flavors): studioweb publish flow asks the user for a destination [PILOT-7271]

avserbanescu:
## Motivation

Companion to [UiPath/Autopilot#5454](https://github.com/UiPath/Autopilot/pull/5454) ([PILOT-7271](https://uipath.atlassian.net/browse/PILOT-7271)). Across several live Studio Web sessions, the skill text described a surface the host does not have, and the agent followed it into dead ends: publishing always landed in the personal workspace, `uip solution publish --help` performed a real publish, `uip solution deploy` was declared unavailable when it works, and feed discovery pointed at a command the browser bundle lacks.

## Changes

- **`uipath-solution` — new studioweb flavor.** Replaces the desktop `pack → publish → deploy` lifecycle (which dead-ends in the browser: `pack` is Node-only) with the browser flow. Adds the ask-first destination choice, the feed-vs-folder distinction, and the deploy guidance below.
- **`uipath-api-workflow`**: removed all four "optionally inspect `uip solution publish --help`" instructions — that probe published a real package. Documented destination selection and the bridge flag table.
- **`uipath-maestro-flow`**: publication routes through the intercepted CLI with the same ask-first flow rather than handing the user to the UI; deployment guidance corrected.
- **Deploy is available.** `solution-tool/browser.json` marks the tool `compatible` and excludes only cleanup/pack/restore/run/packager-tool, so every deploy verb ships in the browser bundle. The flavor now includes `deploy run` and drops the "deployment stays in Studio Web" claim.
- **Guard the side effect.** `deploy run` never reuses a folder name and cannot deploy into an existing folder; `--parent-folder-path` nests the new one, and without it the folder is created at the tenant root. The flavor states that publishing to the shared location puts nothing in the Orchestrator folder named `Shared` — the folder and the publish destination are different objects — which is the misreading that prompted this work.
- **Feed discovery.** `uip solution feeds list` and `publish --feed` do not exist in the browser bundle. Two narrow markers were added to the canonical `references/pack-and-deploy.md` (the feed-discovery section and the single `publish --feed` scoping row) and overridden for studioweb. Marking the whole section would have frozen it for this flavor and dropped its `--parent-folder-path` / `deploy run --feed` guidance, which is correct in the browser — those stay canonical and keep inheriting.
- **Tests**: api-workflow contract assertions updated to the new publish contract; new composed-content test for the `uipath-solution` flavor.

`uipath-agents` intentionally has no override: it carries no publish text, and the interceptor's own guidance covers `uip agent publish`.

## Reviewers

This touches one **canonical** file, `skills/uipath-solution/references/pack-and-deploy.md` (owned by @UiPath/team-merlot @UiPath/team-orange), but **only to add marker comments** — no prose changed, and the default variant composes byte-identical to `HEAD` (`diff` exit 0). Everything else is under `skill-flavors/studioweb/`.

## Validation

- `npm run skills:validate` — OK (default: 27 skills, 0 replacements; studioweb: 27 skills, 177 replacements)
- `npm run skills:build` — OK, marker-free trees; `grep -rl "skill-flavor:" build/skills/` returns nothing
- `npm run skills:test` — 51/53; the 2 failures are pre-existing npm-registry environmental failures reproduced identically on `main`
- Composed studioweb `pack-and-deploy.md` verified: no `uip solution feeds list`, no `--feed "Finance"`, and `--parent-folder-path` plus the deploy-side scoping rows all survive

## Known residual

In the composed studioweb reference, the inherited sentence "pass the same `--feed` / `--personal-workspace` there too" refers to `deploy run` (correct — deploy still takes `--feed`), but the word "same" now points back at a publish flag this flavor does not use. Left inherited rather than adding a third marker for an anaphora; the flavor's `SKILL.md` states the browser-correct rule directly.

[PILOT-7271]: https://uipath.atlassian.net/browse/PILOT-7271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


avserbanescu:
## Upstream dependency: UiPath/cli#3844

Worth knowing before merge — [UiPath/cli#3844](https://github.com/UiPath/cli/pull/3844) (open) promotes `uip solution deploy upgrade` **and** the folder-feed surface (`solution feeds list`, plus `--feed` / `--personal-workspace` on publish, deploy run, deploy list, packages list, packages delete) out of preview, and fixes upgrade so it actually upgrades rather than leaving a `Draft` record behind.

Two of this PR's statements are written to survive that landing, and one will need a follow-up.

**Survives #3844.** The guidance that `--feed` does not apply to a Studio Web publish does not depend on the preview gate: the host intercepts `solution publish` before it reaches the CLI, and the intercepted command takes `--location` against the host's own destination list. Corrected in 67002bc79 — the earlier text claimed both `feeds list` and `--feed` were absent from the browser bundle, which was wrong. They are preview-gated (`solution-tool/src/tool.ts` wraps `registerFeedsCommand` in `previewOnly`; `commands/publish.ts` guards `--feed` with `isPreviewBuild`). The "unknown command" that produced that claim came from a locally rebuilt, unstamped CLI bundle in one dev environment — a CI-published prerelease stamps its own version, so `isPreviewVersion()` is true and both surfaces register.

**Also vindicated by #3844.** This PR tells the agent not to recommend `deploy upgrade` in Studio Web (2ed2214e9). #3844 explains why that was right on the merits, not just on availability: *"The upgrade endpoint only queues the move: it creates a second deployment record (`VersionChange`, status `Draft`) and leaves the live version alone... The command reported success while the version never changed."*

**Needs a follow-up when #3844 lands and Autopilot repins.** This PR states that only two destinations are offered over the chat bridge, because the host hands it the v1 publish-location list. That stays true for `publish` (host-intercepted, bridge-supplied list). But once the feed surface is GA, `deploy run --feed` and `packages list --feed` become usable from chat, so the deploy-side guidance should gain folder-feed coverage. Not blocking this PR; worth a tracking note.

avserbanescu:
Pushed 7f47bebb5 — this PR now also touches a **second canonical skill**, `skills/uipath-maestro-flow/SKILL.md`, so `@chandusailella @mukundbayyaram @UiPath/maestro-cli-team` are pulled in as reviewers.

**What it does there:** adds an *empty* extension point immediately after Critical Rule #5 ("Always present user questions as a dropdown with a `Something else` escape hatch"):

```
<!--skill-flavor:user-question-options-extra:start-->
<!--skill-flavor:user-question-options-extra:end-->
```

The default flavor composes **byte-identical to before** (verified: `diff -r` of the composed `default` tree against HEAD is empty). Rule #5 itself is unchanged and stays shared — this is the additive extension pattern rather than a wholesale replacement, so future edits to rule #5 keep flowing to every flavor.

**Why:** rule #5 is unconditional, and it is right for the open-ended authoring decisions it was written for (#246, Apr 2026). It is wrong for one case this PR introduces. When `uip solution publish` runs with no destination, the interceptor answers with the exact set of feeds the user can publish to, and `--location` accepts only those keys or names. Offering `Something else` there is a guaranteed dead end: nothing the user types can match, the command rejects it, and they get shown the same list a second time.

The studioweb flavor fills the block with a carve-out scoped to *host-enumerated* option sets only. Everything else in rule #5 — the enumerated options, the numbered-list fallback, the non-interactive behaviour, the consent gates — is explicitly restated as still applying.

Verification: `skills:validate` OK (default 0 replacements, studioweb 178), `skills:build` clean with no marker leakage, `skills:test` 51/53 — the 2 failures are the pre-existing `npm publish --dry-run` prerelease-tag environment failures, identical on a clean tree.

avserbanescu:
## Verified end-to-end in Studio Web — every piece of guidance in this PR held up

Full lifecycle run live in the browser (alpha tenant, local build serving these branch skills from `/skills/sw/`, `@uipath/cli` `1.202.0-dev.8509` confirmed with `uip --version`). Solution `test upgrade 4`:

publish **v1.0.0** → shared tenant feed → deploy into **`Shared/test upgrade 4`** → upgrade in place to **v1.0.2** → **v1.0.3** → **v1.0.4**. Every step Successful / Active, folder and configured values preserved, **zero `4005` errors**.

Companion Autopilot PR: UiPath/Autopilot#5454 (full command-by-command table there).

### What each piece of guidance did

**Ask-first destination** (`SKILL.md`) — `uip solution publish` with no destination flags published nothing and enumerated the feeds. No silent personal-workspace default. The user had already said "shared", so the agent correctly skipped the question and went straight to `--location "Orchestrator Tenant"` — exactly the "skip the question when the user already named a destination" branch.

**The `user-question-options-extra` carve-out** (`uipath-maestro-flow`) — scoped precisely, which is the part I most wanted to see. Across the whole run "Something else" was offered exactly **once**, and on the right question:

- **Publish destination** — host-enumerated, `--location` accepts only those keys or names → offered without "Something else". Correct: a free-form answer there can only produce a rejected command and a second round-trip.
- **Deployment/folder name** — genuinely open-ended, the user invents the string → offered **with** "Something else". Also correct, and it shows the carve-out did not over-apply and quietly strip the escape hatch from open questions.

That is the distinction the exception was written to draw, behaving as intended on both sides.

**"Pass the deployment's `Key`, not its `InstallDeploymentKey`"** — this earned its place outright. The key changed after *every* version change; four distinct keys across the run:

```
7aefe1b4-1ef0-4627-a83b-b04b6f1a3c98   (after deploy run,  v1.0.0)
606e642a-18b8-40fb-84a8-de394764e738   (after upgrade,     v1.0.2)
084c0ea2-87f1-47a0-8210-2f1a278f7297   (after upgrade,     v1.0.3)
840d0149-d88e-4e66-a0de-6fa63532d319   (after upgrade,     v1.0.4)
```

The agent re-read the current `Key` from `deploy list` before each upgrade rather than reusing a cached one, and never hit `HTTP 400` / `4005`. Reusing the previous key would have failed on every upgrade after the first — so this rule is what made three consecutive in-place upgrades possible.

**"To ship a later version, upgrade it — do not `deploy run` again"** — followed correctly all three times. The agent checked `deploy list` for an existing deployment first, found one, and upgraded in place instead of creating `test upgrade 4_1`, `_2`, `_3` folders in a shared tenant.

**Removing the old "`deploy upgrade` is NOT usable here" claim** — necessary, and its cost is measurable. That text was still being served to a session earlier the same day (the frontend had been started six minutes before `82d2ecdec` landed, and skills are snapshotted at build time). That session treated `deploy upgrade` as unavailable, went hunting for workarounds, and only found the truth by probing `--help` directly. With the corrected text loaded, the agent went straight to `deploy upgrade` every time.

### One CLI-side note for reviewers

`deploy list` rejects `--output-filter` unless an explicit `--limit` is passed. The agent hit this, recovered on its own by re-probing `--help` and retrying with `--limit 1000`, and recorded it as a memory — so it cost a round-trip but not the task. I have left it out of `cli-conventions.md` for now, since it is CLI behaviour rather than solution-lifecycle guidance and is probably better fixed in the CLI. Happy to add it here if reviewers prefer.

astridstaicu (approved):

PR #2934, opened 2026-08-31
fix(uipath-maestro-case): add `number` to the case type enum, and state the leaf-name rule where it is read

roberts-cliff:
`skill-case-io-binding-matrix` has failed on **both harnesses every run since Aug 27**. Two distinct defects, one per harness. Neither is the agent being careless — both are gaps in this skill.

## 1. `number` was missing from the documented type enum

Claude emitted `type: "integer"` where the checker expects `"number"`:

```
reassigned output 'estimatedAge' -> 'estimatedAge' is not canonical:
  {'type': {'expected': 'number', 'actual': 'integer'}}
```

The task's SDD declares `estimatedAge` as **`number`** — in the Case Variables row and both Outputs rows. But the skill's type vocabulary read:

```
string | integer | float | double | boolean | datetime | date | jsonSchema | file
```

**No `number`.** Given that list, normalising the SDD's `number` onto the nearest listed value is the reasonable move. It is also wrong.

Verified against real artifacts before changing anything: `AgedInvoiceResolution/caseplan.json` carries **3 × `number` and 3 × `integer`** and passes `validate`. So `number` is legal and the enum was simply incomplete.

**Fix:** adds `number` in all three places the vocabulary is stated (`global-vars/planning.md` field semantics + § Types, `create-inline-common.md`, `entry-points-sync.md`), plus a `number` row in the entry-points property-body table. And states explicitly that **the list is not a normalisation target** — emit the SDD's declared type verbatim; if a type is not listed, keep it and raise a Review Flag rather than mapping it onto the nearest match.

## 2. The leaf-name rule was documented only where the agent never looked

Codex emitted `name: "Error.Message"` where the leaf name is required:

```
reassigned output 'Error.Message' -> 'errorMessage' is not canonical:
  {'name': {'expected': 'Message', 'actual': 'Error.Message'}}
```

That rule is stated **correctly** in `io-binding/impl-json.md` — in prose, in the emitted shape, and in a worked `Error.Message -> errorMessage` example that even calls out the wrong-type anti-pattern.

But the Codex run made **28 commands and zero `Read`s**. It never opened the file. And `io-binding/planning.md` — which *is* read during planning — described the dot-path **`type`** rule while omitting **`name`** entirely.

**Fix:** adds the `name` half to planning.md's dot-path paragraph, beside the existing `type` half, naming both wrong forms concretely and linking onward to the worked example.

## Why not more prose in impl-json.md

This is the third time this week a correct, well-written rule has been ignored because of *where* it lives or *what it omitted*, not what it said. Claude read `impl-json.md`, had the worked example in context, and still failed — because the enum it was reconciling against was wrong. Codex never read it at all. Adding more text to that file would have fixed neither.

## Scope

Docs only, 4 files, 8 insertions. No behaviour claimed for the third harness, and **not** a fix for `golden-rebuild`'s surviving `$xref(` — that is a separate defect in the same Rule 10 cluster (Step 11.5 / Check 4 going unexecuted, which `validate` cannot see).

## Not claimed

`io_binding` has **not** been re-run against this branch yet. Happy to dispatch before merge.


roberts-cliff:
## Passing run — codex, green

`skill-case-io-binding-matrix` — [run 33412889159](https://github.com/UiPath/skills/actions/runs/33412889159) against `961b68261`:

```
skill-case-io-binding-matrix: SUCCESS
Success criteria: 2/2 passed, weighted score: 1.000
1/1 PASS (linux)
```

This task had failed on **both** harnesses every run since Aug 27.

The produced `caseplan.json` confirms all three fixes independently of the checker verdict:

```
dotted names      : 0          (was name="Error.Message")
integer types     : 0          (was type="integer"; now 4 x number)
custom rows       : 6  ->  blank target + body present on 6/6
non-custom rows   : 5  ->  all carry a real target
```

That last line matters: an intermediate commit of mine over-applied the blank-`target` rule to bare auto-mint rows, which the runs caught. The final scoping keeps blanks on Scenario E only.

## Claude is not green yet

Same commit, [run 33412901810](https://github.com/UiPath/skills/actions/runs/33412901810): the artifact shows **zero** violations of the three contracts this PR fixes, but the task still fails — now on `API binding resourceSubType is None`, a `bindings_v2` defect unrelated to output shape. A local `--driver tempdir` run reproduced the same picture and then failed on `case debug` returning `Faulted`, which is the current alpha platform outage (Orchestrator SQL at ~38% failure today), not the skill.

So the claim is narrow and deliberate: **these three contract violations are eliminated, verified in the artifacts on both harnesses, and codex now passes outright.** It is not "the task passes everywhere" — there are further defects behind the ones this PR addresses.

StephenHanzlik (approved):
Nice!

charlesliu9 (approved):

Yiqi517 (approved):
Nice!

PR #2939, opened 2026-08-31
fix(uipath-maestro-case): make the references readable on a harness with no Read tool

roberts-cliff:
Codex reads reference files, then stops before the end — and the rules that go unread are the ones we keep seeing violated. This PR fixes the reading. **Scope was cut back on 2026-09-02**: everything past the read fixes (skeleton-then-Edit cadence, `apply_patch` naming, ~20 per-assertion io_binding/golden fixes) measured no gain and was dropped; see "What was dropped" below. **7 commits: six read fixes + a merge of `main` (#2943, #3006).**

## The original measurement

Codex has **no `Read` tool**. Every file access goes through Bash. Across one nightly's codex tasks:

```
166 ranged reads;  86 of them exactly `sed -n '1,240p'` or `1,260p`;  124 start at line 1
13 of 84 file-reads left content unread
4x  grep -n 'END:'   — searching FOR the END marker instead of reading to it
```

There is no hard limit — the largest file ever read to EOF is 48,309 B, the two never read whole were 58,902 B and 67,815 B. Agent judgement, not a cap.

## What changed — the skill

**1. Name the tool the harness actually has.** Rule 24: use `Read`, or with no `Read` tool a single `cat`. Verified `cat` is safe at these sizes (linear to 67,815 B / 16,830 tokens, no plateau).

**2. Split the two files never read whole**, at their own headings: `implementation.md` 59,210 B → Phase 2 / Phase 3 / Phases 4–7; `case-editing-operations.md` 67,815 B → operations / primitives / composites. Each split file carries its own END marker; Rule 24 applies per file.

**3. Required-read markers** on the reference links agents were observed skipping (io-binding planning → `bindings-and-expressions.md`, `impl-json.md`).

**4. Per-phase read manifest** (Step 5.9 for Phase 2, Step 9.65 for Phase 3), derived from `sdd.md`: a fixed core plus one `impl-json.md` per task/trigger/condition type the design contains. Framed as a floor, not a ceiling — the manifest cut one task's reads 25 → 18 and dropped `global-vars/impl-json.md` (n=1), hence the wording. Exact-path gating: `planning.md` does not satisfy an `impl-json.md` entry.

**5. Merge of `main`.** #2943's `tasks.md` removal and element terminology win everywhere they conflict; this branch wins on file structure. #2943's Phase 3 / Step 12 completeness-gate edits are ported into the split files they now live in. The write cadence is main's (per-section, `<10` / `≥10` elements) — unchanged by this PR.

## Evidence

- **codex, golden:** refs read to EOF 11 → 44 after the read fixes; the `validate`-green-zero-task-nodes failure (no task `impl-json.md` ever read) stopped occurring.
- **Skill smoke: 6/6** on the pre-cut head; re-running on the cut head now.
- **`uip maestro case validate` returns `Valid` for a caseplan whose stages contain zero tasks**, with unresolved `$xref` markers, a hoisted `conditionExpression`, or a hand-composed connector context. It is a shape gate, not a completeness gate — which is why a skipped read fails silently.

## What was dropped, and why

Branch was reset from 27 commits to the six read fixes (backup: `backup/case-readability-pre-reset`). Dropped:

- **Skeleton-then-Edit write cadence + `apply_patch` naming.** Motivated by codex composing `caseplan.json` in a `python` heredoc; judged a dead end — it fights the harness rather than the reading problem this PR is about.
- **~14 per-assertion io_binding fixes** (bare-row format, `vars.` scoping, nested-leaf names, the Scenario E table and its revert). io_binding all-pass sat at ~1-in-7 before and after: ~30 conjunctive assertions at ~0.95 each ≈ 21%. Per-assertion prose fixes cannot move that number.
- **Golden Phase 3 fixes** (SLA count gate, Step 10.5 UUID count, Step 11.5 `$xref` grep, connector-context copy rules). 0 for 9 runs across three commits executed all four of Steps 9.7/9.8/10.5/11.5; failures moved between runs rather than vanishing. Self-administered gates were never run (grep gate: 0/66 calls).
- **The TodoWrite-mechanism claim** for phase-tail omission. Falsified — claude-code with 16 TaskCreate/TaskUpdate calls dropped the identical tail.

The Rule 2 scope fix (design-not-schema; the `Action` output never minted) is also in the backup and may return as its own PR.

## Prior work

Everything above is an instance of documented 2025–26 results:

- **Curse of instructions** (ManyIFEval, Harada 2025): all-instruction success = single-instruction success ^ N. Explains io_binding's ~1-in-7.
- **Omission dominates** (IFScale 2507.11538; 2608.02639): at density the errors are dropped instructions, not wrong ones.
- **Procedures fail at the tail** (2605.00817): under-execution rises with step count — our Phase 3 tail.
- **Corrupt success** (2603.03116): 27–78% of benchmark successes conceal procedural violations — our `validate`-green-zero-tasks case.
- **Guardrails beat guidance** (2604.11088; ContextCov 2603.00822: 88.3% vs 67.0% prompt-only; PolicyGuard 2606.29225): enforcement outside the document moves compliance. #2943's `audit_caseplan.py` is the right shape; the CLI `validate --strict` spec (UiPath/cli#3927) is the follow-on.
- **94% of SKILL.md files have a Rationalization Loophole** (2607.01456): no guard against skipping a required step.


roberts-cliff:
Coder-eval dispatched on this branch, both harnesses:

| harness | run |
|---|---|
| claude | https://github.com/UiPath/skills/actions/runs/33429998424 |
| codex  | https://github.com/UiPath/skills/actions/runs/33429984907 |

Ref `fix/case-readability-full-reads`, `-j 2`. Three tasks, chosen because they exercise the content that moved:

- `golden_rebuild/cm_golden_expense` — fails on surviving `$xref(` markers. Step 11.5 and Check 4 are exactly what moved into `implementation-phase-3.md`, out of the region codex stopped reading at (~line 230 of 404).
- `io_binding/io_binding` — the current flakiest in the Rule 10 cluster.
- `edit/bound_task_add_remove` — edit mechanics, reads the `case-editing-*` files.

**What a green run would and would not prove.** This PR moves text; it does not change a single rule. So a pass here is consistent with the split helping, but it is not proof — these tasks vary run to run, and the same file was sometimes read whole and sometimes not before any of this. The stronger signal will be in the artifacts: whether the agent now reads `implementation-phase-3.md` to its END marker, and whether `$xref(` survivors disappear from the emitted caseplan. I will check that directly rather than reporting the verdict.

**Also worth stating up front:** alpha is degraded right now — Orchestrator SQL is at roughly 38% failure and `case debug` runs are returning `Faulted`. `cm_golden_expense` and any other task with a debug step may fail for that reason alone. I will separate environmental failures from real ones when I post results.

StephenHanzlik:
Read the thing codex

StephenHanzlik (approved):
Looks good.  Keeping an eye on evals.

PR #2943, opened 2026-08-31
refactor(uipath-maestro-case): drop the tasks.md; gate completeness o…

charlesliu9:
refactor(uipath-maestro-case): drop the tasks.md; gate completeness on caseplan.json

The build pipeline was `sdd.md` → `tasks.md` → `caseplan.json`. The middle
file was a lossless restatement of the SDD in T-numbered form. On real runs it
came out at 4–96KB (median ~27KB), written once via section-batched appends
and re-read two or three times more — roughly 15–25% of a large build's output
tokens and turns, spent before a single line of `caseplan.json` existed.

That price bought one thing worth keeping: enumeration discipline. Writing one
T-entry per SDD row forced the agent to visit every declaration, which is the
defense against the expensive silent-omission class — a dropped stage,
variable, or task entry rule. `uip maestro case validate` cannot see any of
them, and a task with no entry rule only *warns* there, then hangs
`case debug` indefinitely rather than faulting.

But `tasks.md` defended that badly, in two ways:

- **It graded the plan, not the artifact.** A perfectly enumerated plan could
  still yield an incomplete caseplan, and nothing checked the gap.
- **It added a failure class of its own.** Plan-shape drift — merged SLA title
  lines, invented header synonyms, re-quoted SDD cells, heading levels
  `audit_plan.py` silently counted as zero — failed runs whose
  `caseplan.json` was fine.

So: keep the discipline, move the gate to the artifact, delete the file.

## What

`sdd.md` stays the plan. Phase 1 becomes resolution only, producing tenant
identities in `tasks/registry-resolved.json`. Phases 2–3 read `sdd.md` plus
that ledger and write `caseplan.json` directly. Step 12 runs a new
deterministic cross-check that must print `AUDIT OK` before Phase 4.

Plan-only mode is retired: a request that stops before the build now delivers
`sdd.md` alone. The ledger deliberately carries no structure, activation
modes, entry rules, I/O, or rationale — duplicating the contract there would
recreate the very drift being removed.

## How

**New gate — `scripts/audit_caseplan.py` (485 lines).** Diffs
`caseplan.json` against `sdd.md`, exiting non-zero on any MISSING finding:

    dropped stage / task / variable / condition / SLA row
    task type mismatch against the SDD
    empty entryConditions          (validate only warns; debug hangs)
    placeholder standing in for an already-resolved resource
    bare resourceKey with no `.<entry>` leaf suffix
    api-workflow binding missing resourceSubType

WARN findings — extra caseplan elements, legitimate placeholders, surviving
`<UNRESOLVED>` markers — report but do not block. The parser resolves task
names through `displayName`, `label`, or the `name` binding default, and
tolerates `Secondary Stage:` headings, `Name (slug)` stage titles, and
disambiguated duplicate task names.

**`references/planning.md`: 399 → 184 lines (46KB → 22KB).** Retitled
"Phase 1 — Resolution". Steps 0–3 kept; the Step 4 authoring spec and Step 5
plan gate deleted. Survivors were relocated, not dropped:

- execution order, the activation-mode audit, and lane-grouping rules →
  `implementation.md` Step 9, where tasks are actually written
- the completeness principle → `implementation.md` Phase 2 preamble and
  SKILL.md Rule 6
- the "Not Covered" section → the `phased-execution.md` completion report

**`SKILL.md`.** Description now reads "Resolves tenant resources then authors
or edits caseplan.json directly" (486 chars; every `→` redirect preserved).
Rule 6 becomes losslessness against the SDD, enforced by the new gate; Rule 7
becomes a build gate; the plan-only branches in Rules 1 and 3 become
design-only; Rule 8 moves `<UNRESOLVED>` markers to the ledger; Rule 13 drops
`tasks.md` from the artifact list.

**60 skill files retargeted**, including all 45 plugin references. The
io-binding projection keeps its delimiter-strip and placeholder-rejection
rules verbatim — those encode real, repeatedly-observed defects — but now
projects to caseplan shape, with the intermediate notation demoted to an
explicitly in-memory "canonical binding form". T-numbers survive only as
execution-order labels in `id-map.json` and TodoWrite, never as a file.

Deleted `scripts/audit_plan.py` and `assets/templates/tasks-md-template.md`,
and fixed the 12 anchor links pointing into renamed sections. uipath-planner's
`case-sdd-template.md` loses its "Tasks file" row; its own Lane A
`<process>-tasks.md` is a different artifact and is untouched.

## Tests

`tests/tasks/uipath-maestro-case/athena_cm_event` is migrated as the first
case. `check_athena_cm_event_plan.py` → `check_athena_cm_event_rules.py`
grades the same regression — an authored entry rule must survive the build
unchanged — against `caseplan.json` instead of `tasks.md`, and widens it:

- all 7 tasks, not the 2 the caseplan checker already covered
- both `selected-tasks-completed` gates resolved to their predecessor's id
- no task left with empty `entryConditions`

The YAML swaps the checker and records `audit_caseplan.py` adherence as an
advisory `command_executed`. A hard `run_command` would have to locate a
script inside the installed plugin, and the outcome is already graded
deterministically by the three checkers.

Migrating this task exposed a real bug in the new gate: caseplans that name
tasks only through bindings would have reported every task MISSING. Fixed,
with a regression test.

## Verified

28 unit tests pass (19 for the gate, 9 for the Athena checkers). The gate
returns `AUDIT OK` on a real matched SDD/caseplan pair, and correctly reports
genuine omissions on another — RetailAccountOpening is missing two declared
variables and the entry/exit conditions on one stage, precisely the class this
change exists to catch.

`skills:validate`, `skills:build`, `validate-skill-descriptions.sh` and
`check-skill-status.py` are clean. `skills:pack` fails, but it reproduces with
these changes stashed and is pre-existing.

## Continued tests

52 further files under `tests/tasks/uipath-maestro-case/` referencing
`tasks.md`(including `check_tasksmd_shape.py` and 35 task YAMLs) were migrated and tests passed. Except a few tests also failed in master branch, means not introduced by this PR, and will be fixed separately.


roberts-cliff (approved):

skills/uipath-planner/assets/templates/case/case-sdd-template.md
```diff
@@ -29,7 +29,7 @@
 | **Project SDD role** | child |
 | **Independently executable** | no — Lane A derives tasks only via the Solution root |
 | **Project list section** | §2: Stages & Tasks + §4: Integrations |
-| **Tasks file** | `tasks/tasks.md` — the build skill's contract; planner Lane A task derivation writes its own `<CASE_NAME_KEBAB>-tasks.md` per §Next Steps |
+| **Build handoff** | this SDD is the build skill's contract — it resolves resources into `tasks/registry-resolved.json` and writes `caseplan.json` directly; planner Lane A task derivation writes its own `<CASE_NAME_KEBAB>-tasks.md` per §Next Steps |
```
abhiram-vad: Does this mean we still use tasks.md ? I might have missed it when I pushed the initial template, but I feel we don't need to mention this information in uipath planner. UiPath planner can just generate SDD and leave it there? iirc, the phase 0 contract in uipath-maestro case mentions that.

charlesliu9: good catch, I'll remove it

abhiram-vad (approved):
lgtm

uipreliga (approved):
Activation part looks OK, but please check the bot-identified issues.

uipreliga:
## Code Review — `feat/case-remove-tasks-md`

Multi-reviewer pass (3 independent reviewers, findings deduplicated). **Every finding below was verified by running the code**, not inferred from reading it — repro snippets included.

**Baseline:** test suite green — `202 passed, 14 skipped` across `tests/tasks/uipath-maestro-case/` and `tests/tasks/uipath-planner/` (skips unrelated: `jmespath`/`PyYAML` missing). The two `sdd_check.py` twins are byte-identical. All relative links in the 58 changed Markdown files resolve, and nothing references the deleted `tasks-md-template.md`.

This PR makes `audit_caseplan.py` the **sole** enforcement point for losslessness, replacing everything `tasks.md` used to guarantee. So the gate's trustworthiness *is* the PR. That's where the findings concentrate.

---

### 🔴 Critical

#### C1 — The gate prints `AUDIT OK` on an SDD it parsed nothing from
`skills/uipath-maestro-case/scripts/audit_caseplan.py:303-392`

Every MISSING check has the shape *"SDD declares X → caseplan must have X."* There is no assertion that the parse produced anything. Verified with unrelated prose as the SDD, against a real caseplan:

```
MISSING: [] -> exit 0
WARN   : ["stage 'Intake': in caseplan.json, no matching SDD stage"]
```

Stage headings at `####` instead of `###` do the same. Any SDD the parser doesn't recognize degrades silently into a no-op that still prints `AUDIT OK: caseplan.json covers every SDD element` — which `implementation.md:367` instructs the agent to quote as evidence.

**Fix (minimal):** fail in `main()` when `sdd["stages"]` is empty, or print the parse census (`stages=N tasks=N vars=N`) on the OK line so a zero is visible.

---

### 🟠 High

#### H2 — `TASK_HEADING` misses every secondary-stage task detail block
`audit_caseplan.py:34` — `r"^Task\s+[\d.]+\s*[:.]"` cannot match the `S`-prefixed numbering this repo's own fixtures use.

```
'Task 1.1: SLA warning timer'      -> True
'Task S1.1: SLA warning timer'     -> False
'Task S2.1: Create incident record'-> False
```

Present in four staged fixtures, **including both that this PR wires the gate into**:

```
aged_invoice_structural/fixtures/sdd.md:468  ##### Task S1.1: SLA warning timer
aged_invoice_structural/fixtures/sdd.md:529  ##### Task S2.1: Create incident record
sla_from_sdd/fixtures/sdd.md:554, :627
```

Running the parser on `aged_invoice_structural` confirms both tasks return `entry_rows=0, identity_resolved=False` — which downgrades the placeholder-vs-resolved MISSING check to a WARN and skips the entry-rules check entirely. The interrupting exception lanes the skill emphasizes most are exactly the ones that lose their strongest checks.

**Fix:** `TASK_HEADING = re.compile(r"^Task\s+[A-Z]?[\d.]+\s*[:.]\s*(.+)$", re.I)`

#### H3 — Two SDD tasks can bind to one caseplan task, hiding a dropped task with no signal at all
`audit_caseplan.py:291-300` — the exact-match branch never consults `seen`.

Verified: SDD declares `Check` and `Check Order`; caseplan contains only `Check Order`:

```
MISSING: []
WARN:    []
```

`Check` matches `Check Order` through the prefix branch, `Check Order` then exact-matches the same node, and the genuinely absent task is never reported — not even as a WARN.

A second door to the same defect: `norm()` strips all punctuation *and* drops trailing parentheticals, so `Approve (final)` and `Approve (initial)` collapse to one key in `plan_tasks` (`compare:329`). Verified end-to-end — a caseplan missing `Approve (initial)` prints `AUDIT OK`, exit 0.

**Fix:** guard the exact branch with `seen` too. Note that simply rejecting duplicate normalized keys would wrongly block distinct valid labels — match on exact display name first, then fall back to normalized/prefix only when the candidate is unique *and* unseen, with `seen` holding indices rather than names.

#### H4 — A heading or column-header variation silently deletes a whole check class
`audit_caseplan.py:129-141` (headings) and `:100-105` (columns) match by exact `norm()` equality.

Verified against a fixture in the tree today — `io_binding/fixtures/sdd.md:27` writes `### 1.5 Case Variables`, normalizing to `"1 5 case variables"`:

```
io_binding fixture -> variables parsed: []
```

All seven declared variables go unaudited, with no warning. Same class of failure for a Tasks table headed `| # | Task Title | … |` — zero tasks parsed, and deleting every task from the caseplan then yields `MISSING: []`.

**Fix:** suffix/containment match for headings; and at minimum, `name_at is None` after a *matched* section heading should never be silent.

#### H5 — `registry-resolved.json`'s documented shape is still the retired Markdown T-entry
`references/placeholder-tasks.md:82-104` plus ~20 plugin `planning.md` files (58 `- inputs:` / `- activation-mode:` / `- lane:` / `- verify:` lines under the renamed `## Fields to Resolve` heading).

The rename was mechanical; the semantics weren't migrated. This contradicts three things at once:

1. **The file is JSON.** All three `registry_handoff_*/check_registry_handoff.py` do `json.loads()` + `assert isinstance(entries, list)` — an agent following `plugins/tasks/process/planning.md` writes Markdown and crashes them on line one.
2. **Rule 9** (`SKILL.md:41`) mandates exact JSON keys `stage` / `task` / `taskType` / `cacheFile` / `searchQuery` / `matches` / `selected` / `rationale`.
3. **This PR's own rule** (`planning.md:163`) forbids restating activation modes, entry rules, inputs and outputs in that ledger — precisely what these blocks instruct.

`placeholder-tasks.md:82-104` still carries a ` ```markdown ` fence, an H1 `# process task "…"`, and rules about "wiring lines rendering as markdown H1 headings" and keeping `order` (which was deleted from the example directly above it).

#### H6 — Presence checks stand in for count checks
`audit_caseplan.py:318, 321, 352, 377, 384`

Row counts are parsed (`entry_rows`, `exit_rows`, `case_exit_rows`, `triggers`) and then discarded — only truthiness of the caseplan side is tested. A caseplan keeping 1 of 3 stage-exit rows, 1 of 4 `metadata.caseExitRules`, or 1 of 3 declared triggers passes clean. That is the archetypal lossy build this gate exists to catch.

Caveat worth encoding in a comment either way: SDD condition rows can legitimately group into composite rule arrays, so `>=` on totals is safer than strict equality. As written, the silence reads as unintentional.

---

### 🟡 Medium

- **`first_table()` absorbs the next table after a header-only table** (`:80-97`) — the `break` requires `header and rows`, so an empty declared table runs into the following table and reports its header + data as rows. Inflated counts → false MISSING that hard-blocks a valid build. Verified.
- **`first_table()` splits on escaped pipes** — the SDD template uses `\|` heavily. Verified: `| 1 | Route \| Escalate | action |` → 4 cells, every later column shifted, value truncated to `'Route \'`.
- **`sla_reference_findings()` scans raw Markdown including fenced code blocks** (`:420-437`), and its findings land in `missing` (exit 1). Verified: a fenced example containing `sla-status-change("Case", "Breached")` hard-blocks the build. Inverse gap: `if args and len(args) not in (2,3)` skips the arity check entirely for an unquoted call.
- **`main()` raises a raw `IndexError` on a missing flag value** (`:454`). Reachable from CI — all three graded YAMLs build the argument as `--sdd $(find . -name sdd.md … | head -1)`, so an empty `find` yields a traceback instead of a diagnosable failure.
- **`field()` returns `""` when the value is on the next line** (`:112-114`) — falsy, so `identity_resolved` silently becomes `False` and the placeholder check downgrades to a WARN. A rendering variation weakens the gate rather than being reported.
- **Deleted SLA validation has no replacement.** `_tasks_frontend_issues()` enforced *escalation needs ≥1 recipient* and *at-risk needs `at-risk-percentage`*. `grep -n "recipient\|at-risk\|percentage"` over `sdd_check.py` now returns **only the docstring line at `:33` that still advertises them.** Duration/expression/title checks survive at `:219-283`; those two exist nowhere. Applies to both twins. Either re-home them onto the SDD's escalation table or onto `slaRules[]`/`escalationRule[]` in `audit_caseplan.py` — but not nowhere, and fix the docstring either way.
- **`resourceSubType` truthiness** (`:414`) — checks only non-empty while the message and contract require exactly `"Api"`; `"Process"` passes.
- **Duplicate `## Step 12` heading** — `implementation.md:295` (Phase 3 validator, Checks 1–15) and `:357` (the `audit_caseplan.py` gate, inside Phase 4). `SKILL.md:38` says the gate must pass *"before Phase 4"* while `implementation.md:357` places it *inside* Phase 4. Both `#step-12--…` anchors are now ambiguous.
- **`SKILL.md:80` and `:87` still describe the retired workflow.** `:80` reads `→ Phase 1 Planning →` against `SKILL.md:103 ### Phase 1 — Resolution`. `:87` — the **user-facing** greenfield kickoff — still promises *"**Planning** — I draft a task plan from the spec and continue; ask up front if you want to review it first."* That is the first thing the agent says to the user, and it is exactly what `SKILL.md:216` now forbids.
- **Unresolved-marker contract is narrower than documented.** `check_issue_journal.py:57` and `registry_handoff_action_case_sdd_only/check_registry_handoff.py:146` assert on the literal `<UNRESOLVED`, but `registry-discovery.md:256` documents `[REGISTRY LOOKUP FAILED: …]` and `SKILL.md:41` documents `selected: null` as the empty-lookup signals.
- **Test coverage is happy-path + single deletions.** `test_audit_caseplan.py` covers the deletion classes well, but touches none of C1, H2, H3, H4, H6, the CLI edge, or the code-fence case — every one of which is a **false pass**, the failure mode a gate must never have. `test_secondary_stage_heading_is_parsed` is the right instinct but stops at the stage heading and never reaches the task heading, which is the one that breaks.

---

### 🟢 Low

- **Dead code** — `head["own"]` (`:76`, also subtly wrong: slices to the next heading of *any* level), `stage["required"]` (`:155`), `stage["detail_names"]` (`:218`), `task["mode"]`, `task["required"]`, `task["detail_type"]`. Each appears exactly once, at its assignment. `is_blank()` (`:55`) rebuilds its set on every call.
- **`STAGE_NODE_TYPES` accepts `case-management:ExceptionStage`** (`:37`) — removed at schema v22 per `case-schema.md:214` ("MUST NOT be emitted").
- **`binding_findings()` message is misleading** (`:408-412`) — says the key needs the *"`.<entry name>` leaf suffix"*, but per `implementation.md:314` it is the `name`-binding default. The heuristic is sound; the remediation text sends the agent hunting for the wrong thing.
- **Two hitl criteria can fail on a match** — `hitl/quality_01_action_schema_design.yaml:229,237,245` grep across two files, one of which may not exist on either build path, and assert `expected_exit_code: 0`. Verified: grep exits 2 when a named file is missing, even with a match in the other. `quality_02` got this right with `grep A || grep B`.
- **Dangling `tasks.md` / T-numbering leftovers** — `connector-integration.md:59` (*"mark the task `<UNRESOLVED: …>` in `tasks.md`"* — three sibling lines in the same file were fixed, this one missed) and its `:170` heading `## Output Contract to Tasks.md`; `plugins/conditions/stage-entry-conditions/planning.md:26` and `task-entry-conditions/planning.md:25` cite **§4.8**, a heading that no longer exists anywhere; `plugins/sla/planning.md:70,176,189` + `sla/impl-json.md:17` use `attach-to: default | T<m>` with no T-numbered file left to reference; `plugins/variables/global-vars/planning.md:54` still says *"Place after the case file (T01)… T-number for the first variable depends on trigger count."* Also `uipath-planner/references/case/case-design-lane-guide.md:13`.
- **`phased-execution.md:172` says "Checks 1-11"** while `implementation.md:299` says 1–15 — and `:174`, two lines later, names Checks 12 and 15.
- **`SKILL.md:38` omits `--registry`** from the gate invocation it quotes, as do all three graded YAMLs — so `registry_findings()` never runs in CI.
- **Stale generated report** — `tests/reports/uipath-maestro-case.md:215,269` cite the deleted `check_tasksmd_shape.py` / `check_task_dependency_chain_tasksmd.py`.
- **Vestigial "mode vs lane conflict"** (`implementation.md:181`) — described a separately-authored plan value; nothing can conflict now.

---

### Known-deferred / out of scope

- `guardrails/artifact_safety_guard.yaml` still says *"MUST assemble caseplan.json / tasks.md"* (`:4`), *"post-tasks.md"* approval (`:150`), and names `audit_plan.py` (`:200`). Commit `9021b412f` defers this deliberately — but as it stands the guardrail prompt instructs the agent under test to do what `SKILL.md:216` forbids, so the follow-up should land with or before this. The regex lookahead `(?!\S*uipath-maestro-case/scripts/)` still exempts the new script correctly; only the prose is stale.
- `uipath-planner`'s `<process>-tasks.md` hits are its own unrelated Lane A artifact — correctly untouched.
- `preview/uipath-maestro-case/SKILL.md` is an upstream `flow-builder-sdk` snapshot with its own contract — correctly untouched.

---

### ✅ What's good here

1. **The direction is right.** Gating the finished artifact beats grading a plan file that can drift from both its input and its output. `implementation.md:13` and `planning.md:163` argue it well.
2. **`implementation.md` doesn't just find-and-replace** — it *adds* the Activation-mode audit and Lane-grouping contract (`:151-175`), genuinely re-homing the semantics `tasks.md` used to carry. That's the substantive part of the refactor.
3. **The gate targets the right failure.** The `no entryConditions` → MISSING check, with its rationale that *"`validate` only warns … a task with no entry rule hangs `case debug` indefinitely"*, is exactly what a deterministic gate should own — and it tells the agent why a clean `validate` isn't evidence.
4. **MISSING/WARN split is well judged**, WARNs print even on a clean exit, and the repair loop has a real stop condition (max 3 rounds → `AskUserQuestion`, plus a no-`python3` fallback).
5. **Test migration is a genuine upgrade where it happened.** `check_athena_cm_event_rules.py` now grades the caseplan's real `entryConditions` and resolves `selectedTasksIds` to the predecessor's actual `id` — something the `tasks.md` version couldn't. `check_io_binding_case.py` kept equivalent caseplan-level coverage of every row it dropped. `build_from_planner_sdd.yaml:70-71` inverted its criterion correctly, now proving *no* substitute plan file was written via a repo-wide `^## T[0-9]+` sweep.
6. **No orphaned checkers**, and both `sdd_check.py` copies stayed in lockstep.
7. **485 lines of hand-rolled parsing is proportionate here**, and it correctly does *not* import from the test tree — skills must stay self-contained per `CLAUDE.md`. Please don't consolidate it with `sdd_check.py`.

---

### Overall

The direction is sound and the doc/test migration is largely complete, but `audit_caseplan.py` isn't yet trustworthy enough to carry the guarantee the rest of the PR now delegates to it: it passes silently on an SDD it failed to parse, skips its strongest checks on every secondary-stage task in the repo's own fixtures, can hide a dropped task with no output at all, misses an entire variable table on a fixture in the tree today, and never compares counts — and the tests exercise none of those paths.

Suggested gating set: **C1–H6**, plus the user-facing `SKILL.md:87` leftover. The Medium docs cleanup (H5's plugin blocks, the duplicate Step 12) and the SLA-validation re-homing are the right immediate follow-ups.

dmetzgar (approved):

uipreliga (commented):
# Review: UiPath skills (uipath-maestro-case: drop tasks.md, gate completeness on caseplan.json) — pr:2943

PR #2943 by @charlesliu9 · feat/case-remove-tasks-md → main · OPEN · reviewed against 8a88d4e

**Change class:** complex — removes a documented skill artifact (tasks.md) from the maestro-case contract, renames the audit script consumed by task YAMLs, and changes the Step 12 completeness gate; correctness requires reasoning about what still references the retired artifact

The tasks.md-to-caseplan.json consolidation is architecturally sound and ships with clean security, tight code, and real new negative-control coverage (8.5/10), but the single gate the whole design now rests on, audit_caseplan.py, prints "AUDIT OK" in at least four situations where it verified nothing (zero-parse SDD, every secondary-stage task, empty-rules entryConditions, partial row drops) while two eval tasks grade so weakly that a no-op agent scores full marks, so the losslessness guarantee this PR advertises is not yet real and the correctness fixes below should land before merge.

## Summary

| 1. Correctness & Logic | 6.9 / 10 | 0 | 3 | 0 | 1 | audit_caseplan.py's TASK_HEADING regex rejects the documented secondary-stage form `##### Task S<K>.<M>:`, so placeholder-honesty and entry-rule-content checks silently no-op for every exception-lane task |
| 2. Code Quality & Simplicity | 9.4 / 10 | 0 | 0 | 1 | 1 | Six computed-but-never-read fields in the new 485-line audit script, two of which advertise checks it does not perform |
| 3. Types & Contracts | 8.3 / 10 | 0 | 0 | 3 | 2 | Completeness gate checks conditions/triggers/SLA for non-emptiness only, so partial row drops print AUDIT OK — and SKILL.md:216 overclaims it as proof nothing was dropped |
| 4. Test & Validation Health | 8.2 / 10 | 0 | 1 | 1 | 3 | build_from_planner_sdd grades only absence/no-touch after the PR — a no-op agent scores the full 11/11 |
| 5. Security | 9.8 / 10 | 0 | 0 | 0 | 2 | audit_caseplan.py interpolates raw `<UNRESOLVED…>` marker text from an agent-written file into the gate's own WARN report unescaped |
| 6. Architecture & Design | 8.5 / 10 | 0 | 1 | 1 | 0 | The new mandatory Step 12 gate replaces a same-skill contract with an unpinned cross-skill one and passes silently when the SDD grammar does not match |
| 7. Error Handling & Resilience | 8.9 / 10 | 0 | 1 | 0 | 1 | Completeness gate prints "AUDIT OK" when it parses zero elements out of the SDD — the empty-parse guard the deleted audit_plan.py had was not carried over |
| 8. Interface, Docs & Compatibility | 8.2 / 10 | 0 | 1 | 1 | 3 | case-sdd-template.md drops the `Tasks file` handoff row that planner Rule 5 and this PR's fixtures still require |

**Overall Score**: 8.5 / 10 · **Weakest Axis**: Correctness & Logic at 6.9 / 10
**Totals**: 🔴 0 · 🟠 7 · 🟡 7 · 🔵 13 across 8 axes reviewed.

## Blockers (🟠 High — 0 Critical)

1. **[Axis 1] audit_caseplan.py's TASK_HEADING regex rejects the documented secondary-stage form `##### Task S<K>.<M>:`, so placeholder-honesty and entry-rule-content checks silently no-op for every exception-lane task** — `skills/uipath-maestro-case/scripts/audit_caseplan.py:34`

`TASK_HEADING = re.compile(r"^Task\s+[\d.]+\s*[:.]\s*(.+)$", re.I)` (line 34) only accepts digits and dots after `Task`. The SDD template this gate reads mandates a different form for secondary stages: skills/uipath-planner/assets/templates/case/case-sdd-template.md:438 — "Secondary-stage task headings use `##### Task S<K>.<M>: <Task Name>`". `Task S1.1: ...` does not match, so `parse_stage` (line 186) records no detail block, and every secondary-stage task falls back to `entry_rows=0`, `identity_resolved=False`, `resolved_resource=None`.

Verified on the repo's own fixture (tests/tasks/uipath-maestro-case/aged_invoice_structural/fixtures/sdd.md, which is graded by the new `run_command` gate at aged_invoice_structural.yaml:92-100):
```
SLA Escalation | detail_names: []
    SLA warning timer entry_rows= 0 identity_resolved= False
Automation Incident | detail_names: []
    Create incident record entry_rows= 0 identity_resolved= False
```
Consequence, reproduced end to end: an SDD secondary-stage task with `**Resolved Resource:** EscalationWorkflow` / `**Resource Identity:** wf-9`, written into the caseplan as `"data": {}` with `entryConditions: [{"rules": []}]`, produces only `WARN: ... placeholder task (empty \`data\`)` and then `AUDIT OK: caseplan.json covers every SDD element` (exit 0) — because the blocking branch at line 357 (`if is_placeholder and task.get("identity_resolved")`) can never fire without a parsed detail block, and implementation.md:365 says WARN lines "do not block".

Fix: widen the regex to the documented form, e.g. `r"^Task\s+S?[\d.]+\s*[:.]\s*(.+)$"`, and add a regression case to tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py using a `### Secondary Stage:` + `##### Task S1.1:` SDD. Also consider failing (not warning) when a stage in the SDD yields zero parsed detail blocks but its Tasks table has rows — that turns this whole class of parse miss into a loud error instead of a silent downgrade.

2. **[Axis 1] Task entry-rule gate accepts an empty-rules `entryConditions` envelope, the exact shape Check 15 forbids** — `skills/uipath-maestro-case/scripts/audit_caseplan.py:347`

Lines 347-353 read:
```python
if not found["entry"]:
    missing.append(... "no entryConditions in caseplan.json" ...)
elif task["entry_rows"] and not any((c.get("rules") or []) for c in found["entry"]):
    missing.append(f"{label}: entryConditions carry no rules")
```
implementation.md:332 (Check 15) states the contract with no exemptions: "Each task's top-level `entryConditions` MUST be a non-empty array whose first element has a non-empty `rules[][]`." Two shapes that violate Check 15 pass this gate clean (both reproduced against the script):
1. `entryConditions: [{"id": "c1", "rules": []}]` on a task whose SDD detail declares only an **Activation Mode** and no **Entry Condition** table — an explicitly supported SDD shape (implementation.md:334 step 2: "If the task has no Entry Condition table, fall back to its Activation Mode"). `entry_rows` is 0, so the `elif` never runs → `compare()` returns `([], [])`.
2. `entryConditions: [{"id": "c1", "rules": [[]]}]` even when `entry_rows` is 1 — `(c.get("rules") or [])` evaluates to `[[]]`, which is truthy, so `any(...)` is True.
Both are exactly the failure implementation.md:332 calls "worse than most Step 12 findings" (`case debug` hangs indefinitely rather than faulting).

Fix: make the rules check unconditional and inspect the inner group, e.g. replace the `elif` with `elif not any(any(group) for c in found["entry"] for group in (c.get("rules") or [])): missing.append(...)`, and drop the `task["entry_rows"]` guard so Activation-Mode-only tasks are covered. Add both shapes to tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py — the existing `test_missing_task_entry_conditions_fail` only sets `entryConditions = []`, which is why the suite is green.

3. **[Axis 1] registry-resolved.json is a JSON ledger but is still documented with the retired markdown T-entry shape — placeholder-tasks.md § Entry Shape, the 20 planning `Fields to Resolve` key lists, a stray ```markdown fence, and "markdown Edit-append" wording** — `skills/uipath-maestro-case/references/placeholder-tasks.md:82`

The heading was renamed from "`tasks.md` Planning-Entry Shape" to "`registry-resolved.json` Entry Shape" (line 82) but the body was left in Markdown:
```markdown
# process task "Validate Submission Completeness" in stage "Submission Review"
- taskTypeId: <UNRESOLVED: processOrchestration-index.json empty in tenant>
- folder-path: <UNRESOLVED>
- runOnlyOnce: false
- isRequired: true
- verify: Confirm Result: Success, capture TaskId ...
```
plus a rule at line 103 that reasons about Markdown rendering ("they would render as markdown H1 headings") and a rule at line 104 that says "Keep every other field — order, verify, is-required, run-only-once, display-name" (`order` was even deleted from the example above it in this PR).

This contradicts three in-scope contracts for the same file: (a) SKILL.md Rule 9 fixes the ledger's keys as `stage`, `task`, `taskType`, `cacheFile`, `searchQuery`, `matches`, `selected`, `rationale`; (b) references/planning.md:154 says the wiring goes into "the entry's `wiringNotes` string array" and then points here for the shape — but here it is a fenced ```text block; (c) references/planning.md:163 says "This ledger holds **only what registry lookups produced** ... do not restate stage/task structure ... inputs, outputs", while `runOnlyOnce` / `isRequired` / `display-name` / `verify` are exactly that. An agent following this section emits non-JSON into `registry-resolved.json`, which Phase 3 later reads as structured data (e.g. plugins/tasks/connector-activity/impl-json.md: "Construct the input-details object from `registry-resolved.json`" including a literal `filter` JSON object).

Fix: rewrite the block as an actual JSON object using Rule 9's keys plus `wiringNotes: ["..."]`, and drop the Markdown-rendering rules. The same rename-without-rewrite exists in two sibling in-scope files and should be fixed in the same pass: references/connector-integration.md:170 still carries the heading "## Output Contract to Tasks.md" over a ```text hyphen-bullet entry, and references/connector-trigger-planning.md:274 ("## registry-resolved.json fields (resolution)") still shows a ```markdown hyphen-bullet block.

4. **[Axis 4] build_from_planner_sdd grades only absence/no-touch after the PR — a no-op agent scores the full 11/11** — `tests/tasks/uipath-maestro-case/build_from_planner_sdd/build_from_planner_sdd.yaml:52`

The PR deletes the only three criteria that required the agent to DO something (`file_exists tasks/tasks.md` w2.0, the T-entry grammar grep w3.0, and the `grep -Fq 'FinancialPostingFunction' … 'resolve at build'` grep w2.0). All five surviving criteria are pure negations:

  `cmp -s sdd.md $TASK_DIR/fixtures/sdd.md`
  `! ls ./*-tasks.md 2>/dev/null | grep -q .`
  `! test -f tasks/tasks.md && ! find . -name '*.md' … -exec grep -lE '^## T[0-9]+' {} + 2>/dev/null | grep -q .`
  `! test -f tasks/registry-resolved.json`
  `! find . -name caseplan.json -not -path '*/.venv/*' | grep -q .`

Every one is satisfied by an agent that reads nothing, writes nothing and answers "I won't do that" — the staged fixture sdd.md is byte-identical because it was never touched, and no file was created. The task scores 11/11 without loading the skill. There is no `skill_triggered` criterion and no `llm_judge` on the reply, so nothing anchors the run to real behavior. This is the anchor case for "a test that passes regardless of the behavior".

Restore a positive signal that survives the tasks.md removal — the task's own prompt asks the agent to "Tell me in your reply what stays unresolved", so grade that: add a `skill_triggered` criterion for `uipath-maestro-case`, plus an `llm_judge` (or a `file_contains` on an agent-written report) asserting the reply names `FinancialPostingFunction` and `EmailDrafter` as deferred until the build machine is online. Keep the five negations as the regression guards they are, but they must not be the whole grade.

5. **[Axis 6] The new mandatory Step 12 gate replaces a same-skill contract with an unpinned cross-skill one and passes silently when the SDD grammar does not match** — `skills/uipath-maestro-case/scripts/audit_caseplan.py:121`

`audit_plan.py` (deleted) parsed `tasks/tasks.md`, whose shape the same skill owned via `assets/templates/tasks-md-template.md` (also deleted). Its replacement `audit_caseplan.py` parses `sdd.md`, whose shape is owned by a *different* skill — `skills/uipath-planner/assets/templates/case/case-sdd-template.md`. `parse_sdd` (line 121-144) hard-codes that template's heading strings and levels verbatim: `head["level"] == 3 and norm(title) == norm("Case Exit Conditions")`, `"Case Triggers"`, `"Case-Level SLA Escalation Rules"`, `"Case Variables"`, and `STAGE_HEADING`/`TASK_HEADING` at levels 3 and 5; `parse_stage` (line 147-217) adds level-4 `"Stage Entry Conditions"`, `"Stage Exit Conditions"`, `"Stage SLA"`, `"Tasks"`. They agree today (I checked every one against the template's heading list) but nothing binds them: the two files sit under different CODEOWNERS lines (122 vs 91), and `tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py:28` pins the parser against a hand-written inline `SDD = """..."""` literal, not against the planner template, so a heading rename on the planner side would keep all 125 tests green. The failure is silent rather than loud: when nothing matches, `parse_sdd` returns `{"stages": [], "case_exit_rows": 0, "triggers": 0, "variables": [], "sla_case": False}`, `compare` appends no `missing`, and `main` prints `AUDIT OK: caseplan.json covers every SDD element` — while SKILL.md:216 tells the agent this gate "proves nothing was dropped" and Rule 6 requires it to print `AUDIT OK` before Phase 4. Fix: make the parse assert it saw something (exit non-zero with `AUDIT FAIL -- parsed 0 stages / 0 tasks from <sdd>; heading grammar does not match the case SDD template`) so a template drift is loud, and add a parity fixture that runs `parse_sdd` over a filled-in copy of `skills/uipath-planner/assets/templates/case/case-sdd-template.md` so the two skills' heading grammars are pinned to each other by a test rather than by hope. Cross-axis: the silent-pass half also belongs to Axes 1 and 4.

6. **[Axis 7] Completeness gate prints "AUDIT OK" when it parses zero elements out of the SDD — the empty-parse guard the deleted audit_plan.py had was not carried over** — `skills/uipath-maestro-case/scripts/audit_caseplan.py:481`

`compare()` derives every MISSING finding from what `parse_sdd()` extracted, and `main()` ends with an unconditional `print("AUDIT OK: caseplan.json covers every SDD element")` (line 481) whenever `missing` is empty. Nothing checks that the SDD yielded anything at all, so an SDD the parser does not recognise produces a false all-clear against an empty caseplan. Verified by running the shipped script:

  $ printf '# SDD\n\nSome prose, no structured sections.\n' > sdd.md
  $ printf '{"nodes":[],"variables":{},"metadata":{}}' > caseplan.json
  $ python3 skills/uipath-maestro-case/scripts/audit_caseplan.py caseplan.json --sdd sdd.md
  AUDIT OK: caseplan.json covers every SDD element   (exit 0)

and again with a near-miss that only shifts heading depth (`## Stage 1: Intake` instead of the template's `### Stage <N>:`, which `parse_sdd` requires at `head["level"] == 3`, line 142) — same `AUDIT OK`, exit 0, against a caseplan with no nodes at all. Same for a `--sdd` pointed at the wrong existing markdown file.

This matters because the gate is the sole enforcement of losslessness now that tasks.md is gone: SKILL.md:38 states the gate "must print `AUDIT OK` before Phase 4", and planning.md:5 says "Losslessness is enforced against the finished artifact by the Step 12 `audit_caseplan.py` gate". A vacuous OK green-lights an empty or truncated caseplan straight through Phase 4 to publish.

The script this replaces had exactly this guard — `skills/uipath-maestro-case/scripts/audit_plan.py` (deleted here) opened `audit()` with:

    headings = list(re.finditer(r"(?m)^## (T\d+)[^\n]*$", text))
    if not headings:
        findings.append("no `## T{N}:` entries found — the compact plan uses T-numbered H2 entries")
        return findings

Fix: add the equal-and-opposite guard in `main()` before reporting success — if `parse_sdd()` returns no stages AND no variables AND no triggers AND no case-exit rows, emit a MISSING finding naming the SDD path and the expected `### Stage <N>: <name>` / `#### Tasks` heading shapes, and exit non-zero rather than printing AUDIT OK. Cover it with a test in `tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py`, which currently has no case for an SDD that parses to nothing. Note this also removes a silent-scoring hazard in the three task YAMLs that invoke the gate with `--sdd $(find . -name sdd.md ... | head -1)` (aged_invoice_structural.yaml:95, athena_cm_event.yaml:91, expense_runnable_e2e.yaml:97): when `head -1` picks a non-conforming sdd.md the criterion silently awards its full weight 3.0.

7. **[Axis 8] case-sdd-template.md drops the `Tasks file` handoff row that planner Rule 5 and this PR's fixtures still require** — `skills/uipath-planner/assets/templates/case/case-sdd-template.md:31`

The PR deletes the `| **Tasks file** | ... |` row from the Planner Handoff header table (it sat between line 31 `| **Project list section** | ...` and line 32 `| **Generated by** | uipath-planner |`) without replacing it. Three in-scope artifacts assume it was renamed, not removed: tests/tasks/uipath-maestro-case/sla_from_sdd/fixtures/sdd.md:25 and tests/tasks/uipath-maestro-case/timer_connector_from_sdd/fixtures/sdd.md:23 now carry `| **Build handoff** | this SDD is the build skill's contract — it resolves resources into `tasks/registry-resolved.json` ... |`, and tests/tasks/uipath-maestro-case/build_from_planner_sdd/build_from_planner_sdd.yaml:13 instructs "do not modernize it to the template's `Build handoff` row" — but `grep -rn 'Build handoff' skills/uipath-planner/assets/templates/` returns nothing, so no such row exists in the render contract. The removal also breaks the planner's own unchanged contract: references/sdd-generation-guide.md:473 lists `| **Tasks file** | ... ← planner writes here on first run. Solution scope: root AND every child carry the SAME canonical `<SOLUTION_NAME_KEBAB>-tasks.md` — exactly one tasks file per solution, never one per project`, and references/pdd-driven-lane-guide.md:29 then defaults the missing field to `<sdd-basename-without-sdd>-tasks.md` and surfaces it to the user as a defaulted field — for a solution-scope case child that yields a per-project tasks file, which that guide forbids. The five sibling templates (rpa/flow/agent/api-workflow/bpmn/coded-app) all still carry the row, so the case template is now the lone divergent mirror. Add the row back under one agreed name — either keep `Tasks file` with the canonical planner value, or land the `Build handoff` rename in the template AND in sdd-generation-guide.md:473 and pdd-driven-lane-guide.md:29 in the same change.


## Non-blocking, but please consider before merge (🟡)

1. **[Axis 2] Six computed-but-never-read fields in the new 485-line audit script, two of which advertise checks it does not perform** (`skills/uipath-maestro-case/scripts/audit_caseplan.py:76`) — `compare()` is the only consumer of `parse_sdd()`/`parse_stage()`/`section_blocks()` output, and it never reads any of these:

- L76 `head["own"] = text[head["body_at"]:heads[index + 1]["start"]] if index + 1 < len(heads) else text[head["body_at"]:]` — computed for every heading in both the SDD pass and each nested `section_blocks()` call, never read anywhere. Pure O(n) slicing per heading, and the `own` vs `body` naming implies a semantic distinction the script does not have.
- L155 `"required": field(body, "Required for Case Completion")` — parsed per stage, never compared against the caseplan.
- L180 `"mode": cell(row, mode_at)` and L181 `"required": cell(row, req_at)` — parsed per task, …
2. **[Axis 3] Completeness gate checks conditions/triggers/SLA for non-emptiness only, so partial row drops print AUDIT OK — and SKILL.md:216 overclaims it as proof nothing was dropped** (`skills/uipath-maestro-case/scripts/audit_caseplan.py:318`) — compare() gates only on emptiness: `if stage["entry_rows"] and not target["entry"]` (line 318), `if stage["exit_rows"] and not target["exit"]` (line 322), `if stage["has_sla"] and not target["sla"]` (line 326), `if sdd["case_exit_rows"] and not plan["case_exit_rules"]` (line 377), and per task `elif task["entry_rows"] and not any((c.get("rules") or []) for c in found["entry"])` (line 352). The SDD row counts it painstakingly parses are used only as booleans. The mapping is documented as strictly 1:1 — references/plugins/conditions/stage-entry-conditions/impl-json.md:28 states "**One row = one condition object.** Each entry-condition row in the SDD maps to a **separate** object in …
3. **[Axis 3] quality_01 multi-operand grep exits 2 when either file is missing, grading the promised either-artifact OR as an AND** (`tests/tasks/uipath-maestro-case/hitl/quality_01_action_schema_design.yaml:229`) — Lines 229, 237 and 245 replace single-file greps over `tasks/tasks.md` with `grep -ri '<pattern>' tasks/registry-resolved.json VendorPOReviewCase/VendorPOReviewCase/caseplan.json` under `expected_exit_code: 0`, and the descriptions promise OR semantics ("caseplan on the resolved path, registry audit on the skeleton path"). A multi-file grep does not implement that: grep's exit status is 2 when any operand is missing, and the -q escape hatch ("if -q is used and a line is selected, the exit status is 0 even if an error occurred") is not in play here because -q is absent. I confirmed the behaviour locally: `grep -ri 'taskTitle' a.json missing.json` printed the match from a.json and still …
4. **[Axis 3] Task-entry planning requires an `entry-rule:` field the SDD template does not define, and tells the agent to edit the SDD** (`skills/uipath-maestro-case/references/plugins/conditions/task-entry-conditions/planning.md:67`) — The rewrite moved the `activation-mode:` / `entry-rule:` pair (shown at lines 62-65) from "the task's own §4.6 T-entry" onto "the task's own SDD row", and line 67 now ends: "if the task's own SDD row is missing `entry-rule:`, go back and add it there." Neither field exists in the SDD contract. skills/uipath-planner/assets/templates/case/case-sdd-template.md's Tasks table header is `| # | Task Name | Type | Activation Mode | Starts When | Required | Run Only Once | Persona | SLA |` (line 285) and the per-task detail block carries `**Type:**`, `**Activation Mode:**`, `**Design Rationale:**`, `**Description:**` and an `**Entry Condition:**` table with columns `WHEN | IF | Display Name` (lines …
5. **[Axis 4] New Step-12 completeness gate ships with 6 untested MISSING branches** (`tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py:130`) — The 284-line test file has real negative controls for dropped stage, dropped task, dropped variable, dropped caseExitRules, missing task entryConditions, type mismatch, bare resourceKey, missing resourceSubType, and the placeholder/resolved pair — good. But six `missing.append(...)` branches in `skills/uipath-maestro-case/scripts/audit_caseplan.py` have no test that exercises their failure:

  - stage entry rows declared, `caseplan has none` (audit_caseplan.py `compare`, "SDD declares {n} entry condition row(s)")
  - stage exit rows declared, caseplan has none
  - `SDD declares a Stage SLA, caseplan has no slaRules` (only the *inverse* is tested, by …
6. **[Axis 6] Sweep stragglers in connector-integration.md and stages/impl-json.md: a surviving `tasks.md` write instruction, a stale section heading, and a tautological input-spec row** (`skills/uipath-maestro-case/references/connector-integration.md:59`) — `connector-integration.md` was swept in four places by this PR (lines 78, 136, 172 and its fenced example) but line 59 was missed and still reads: "**Skip / create declined or failed / non-interactive run** → mark the task `<UNRESOLVED: no IS connection for <connectorKey>>` in `tasks.md` and omit `input-values:`" — a write instruction naming an artifact SKILL.md:216 now forbids ("Do not author `tasks/tasks.md`, a T-numbered task list, or any other restatement of the SDD"). `grep -rn 'tasks\.md' skills/uipath-maestro-case/` shows this is the only surviving write-target reference in the shipped skill; every other hit is a deliberate prohibition. Fix: change it to `registry-resolved.json`, …
7. **[Axis 8] registry-discovery.md:256 retargets the `[REGISTRY LOOKUP FAILED: ...]` marker into registry-resolved.json — a second marker vocabulary the mandated `<UNRESOLVED: ...>` rule and the new audit gate's regex both miss** (`skills/uipath-maestro-case/references/registry-discovery.md:256`) — The line was retargeted from `tasks.md` to the JSON ledger: "mark it in `registry-resolved.json`: `[REGISTRY LOOKUP FAILED: <name> in <folder>]`". Two problems. (1) SKILL.md:40 Rule 8 fixes the marker as `<UNRESOLVED: ...>` ("Keep `<UNRESOLVED: ...>` in `tasks/registry-resolved.json`") and planning.md:151 repeats it ("Record `<UNRESOLVED: <reason>>` in that entry's `taskTypeId` / `typeId` / `connectionId` slot ... and set its `selected` to `null`"), so this introduces a second, undocumented marker vocabulary for the same state. (2) The new gate only recognises the first form — skills/uipath-maestro-case/scripts/audit_caseplan.py:38 defines `UNRESOLVED = re.compile(r"<UNRESOLVED[^>]*>", …

## Nits (🔵)

- **[Axis 1]** audit_caseplan.py node-type sets contain strings the skill's JSON recipes never emit, including `case-management:ExceptionStage` removed at schema v22 (`skills/uipath-maestro-case/scripts/audit_caseplan.py:37`)
- **[Axis 2]** Hand-rolled flag parsing with a string-keyed setter, where every other Python script in the repo uses argparse (`skills/uipath-maestro-case/scripts/audit_caseplan.py:451`)
- **[Axis 3]** authored_rule_text docstring claims it reads `**Key:** value` lines; it silently skips them (`tests/tasks/uipath-maestro-case/design_delegation_regressions/check_case_reasoning_regressions.py:31`)
- **[Axis 3]** New audit fixture puts the `<UNRESOLVED>` marker in `selected`, which the contract says must be null (`tests/tasks/uipath-maestro-case/_shared/test_check_issue_journal.py:66`)
- **[Axis 4]** SLA escalation recipient / at-risk-percentage invariants deleted with no new home; sdd_check docstring still advertises them (`tests/tasks/uipath-maestro-case/_shared/sdd_check.py:33`)
- **[Axis 4]** New audit criterion compares two agent-writable artifacts with no fixture-equality guard on the staged sdd.md (`tests/tasks/uipath-maestro-case/e2e_expense_runnable/expense_runnable_e2e.yaml:97`)
- **[Axis 4]** Activation row 003 rewritten into a near-duplicate of row 002 (`tests/tasks/activation/uipath-maestro-case.jsonl:3`)
- **[Axis 5]** audit_caseplan.py interpolates raw `<UNRESOLVED…>` marker text from an agent-written file into the gate's own WARN report unescaped (`skills/uipath-maestro-case/scripts/audit_caseplan.py:443`)
- **[Axis 5]** Rule 13's Bash allow-list was not updated for the mandatory `audit_caseplan.py` gate Rule 6 now requires, leaving the artifact-I/O guardrail self-contradictory (`skills/uipath-maestro-case/SKILL.md:45`)
- **[Axis 7]** check_issue_journal grader swallows read and parse errors, silently downgrading a build to "nothing unresolved" (`tests/tasks/uipath-maestro-case/_shared/check_issue_journal.py:63`)
- **[Axis 8]** SKILL.md Rule 6 documents the Step 12 gate without `--registry`, diverging from the canonical invocation and dropping the unresolved-marker WARN (`skills/uipath-maestro-case/SKILL.md:38`)
- **[Axis 8]** evals.json cases 1–2 still expect a stop-after-Phase-1 for prompts the rewritten Rule 7 no longer treats as stop conditions (`skills/uipath-maestro-case/references/evals/evals.json:7`)
- **[Axis 8]** planning.md still cross-references "the explicit plan-only / no-build path in SKILL.md Rule 3" after that exception was renamed "Design-only exception" (`skills/uipath-maestro-case/references/planning.md:51`)

## What's Missing

**Parallel paths & mirrors:**
- 🟡 `tests/tasks/uipath-maestro-case/guardrails/artifact_safety_guard.yaml` — the eval that grades Rule 13, the rule this PR's new Python gate sits on — was not swept with the rest of the suite. Line 4 still says the agent "MUST assemble caseplan.json / tasks.md"; line 150 still names a "post-tasks.md" approval step; the line-196 criterion description still lists `| tee tasks.md`; and the comment at lines 200-201 still names the deleted `skills/uipath-maestro-case/scripts/audit_plan.py` and credits "planning.md" for mandating the auditor as "the final plan gate" (it is now `references/implementation.md` § Step 12). The exemption lookahead `(?!\S*uipath-maestro-case/scripts/)` still functions for `audit_caseplan.py`, so nothing breaks today — but this file is owned by a different team (`CODEOWNERS:124`), so the sweep needs a lockstep change there rather than being left as the one un-updated mirror of the retired vocabulary. _(trigger: skills/uipath-maestro-case/scripts/audit_caseplan.py)_ _(restates: Axis 6: Sweep stragglers in connector-integration.md and stages/impl-json.md)_
- 🔵 The prompt reword landed in exactly one task: `tests/tasks/uipath-maestro-case/multi_stages_tasks/task_dependency_chain/task_dependency_chain.yaml:254` now reads "Treat all build gates as pre-approved" (was "Treat the generated tasks.md plan as pre-approved"). Eight sibling case prompts keep the now-artifact-less phrasing "Treat the generated plan as pre-approved" — including three files this PR edited: `tests/tasks/uipath-maestro-case/hitl/quality_01_action_schema_design.yaml:184`, `.../hitl/quality_02_result_downstream.yaml:159`, plus `.../hitl/quality_03_boolean_decision.yaml:224`, `.../hitl/quality_04_skeleton_completion_report.yaml:186`, `.../hitl/smoke_02_stage_exit_on_action.yaml:138`, `.../hitl/smoke_03_multi_outcome_routing.yaml:153`, `.../logging/logging_issue_log.yaml:124`, `.../guardrails/artifact_safety_guard.yaml:149`. "The generated plan" no longer names anything the agent produces; apply the same substitution everywhere or revert it in the one place. _(trigger: tests/tasks/uipath-maestro-case/multi_stages_tasks/task_dependency_chain/task_dependency_chain.yaml)_

**Tests:**
- 🟡 Nothing grades that the **agent** runs the new mandatory Step 12 gate. `skills/uipath-maestro-case/SKILL.md:38` (Rule 6) requires `scripts/audit_caseplan.py` to print `AUDIT OK` before Phase 4, and `skills/uipath-maestro-case/references/implementation.md:359-367` adds a repair loop (max 3 rounds), an AskUserQuestion stop, and a "quote the final AUDIT OK line as evidence" reporting duty. The only appearances of the gate in the eval suite are three harness-side `run_command` criteria that invoke it themselves — `tests/tasks/uipath-maestro-case/aged_invoice_structural/aged_invoice_structural.yaml:95`, `.../athena_cm_event/athena_cm_event.yaml:91`, `.../e2e_expense_runnable/expense_runnable_e2e.yaml:97`; `grep -rn audit_caseplan tests/` returns no `command_executed`/`command_pattern` criterion. An agent that never runs the gate, never repairs a MISSING finding and never reports the WARN Open Items scores identically to one that follows Step 12. Note this interacts with the Rule 13 conflict on Axis 5 — settle the carve-out before adding the criterion. _(trigger: skills/uipath-maestro-case/SKILL.md)_
- 🟡 `wiringNotes` is a new field introduced by this PR (`git grep -c wiringNotes origin/main -- skills/` returns nothing) and is now the sole carrier of placeholder wiring intent into the Phase 5 completion report, required by four files: `skills/uipath-maestro-case/references/planning.md:154`, `.../references/implementation.md:185`, `.../references/phased-execution.md:222`, `.../references/plugins/variables/io-binding/planning.md:97`. It has no grader anywhere — no eval YAML or checker mentions it, and the three `check_registry_handoff.py` copies assert only a superset of the Rule 9 key set (`tests/tasks/uipath-maestro-case/registry_handoff_sdd_only/check_registry_handoff.py:18-26` and `:54`), so a `registry-resolved.json` entry with no `wiringNotes` passes clean. Add an assertion in the placeholder-producing registry-handoff task, and name the key in SKILL.md Rule 9's explicit key list (`skills/uipath-maestro-case/SKILL.md:41`) instead of leaving it to "plus resolved I/O/review metadata". _(trigger: skills/uipath-maestro-case/references/planning.md)_
- 🟡 `tests/tasks/uipath-maestro-case/design_delegation_regressions/check_case_reasoning_regressions.py` was rewritten in this PR (-81 lines, its parser retargeted from tasks.md T-entries to SDD detail cards) and its companion `check_tasksmd_shape.py` deleted, leaving that directory with no test file at all — while every sibling checker directory ships one: `design_delegation_procurement/test_check_procurement_sla_interrupts.py`, `athena_cm_event/test_checkers.py`, `timer_connector_from_sdd/test_check_timer_connector.py`, `_shared/test_check_issue_journal.py`, `_shared/test_sdd_check.py`. The rewritten checker gates a live criterion at `weight: 4.0` (`case_reasoning_regressions.yaml:55-59`), and a unit test over `authored_rule_text` (line 28) would have caught the docstring/regex mismatch reported on Axis 3. _(trigger: tests/tasks/uipath-maestro-case/design_delegation_regressions/check_case_reasoning_regressions.py)_ _(restates: Axis 3: authored_rule_text docstring claims it reads `**Key:** value` lines; it silently skips them)_
- 🔵 The gate's `--registry` argument has zero coverage on both sides. `skills/uipath-maestro-case/references/implementation.md:362` documents the canonical invocation as `... --sdd sdd.md --registry tasks/registry-resolved.json`, but all three eval invocations omit it (`aged_invoice_structural.yaml:95`, `athena_cm_event.yaml:91`, `expense_runnable_e2e.yaml:97` each pass only `--sdd $(find . -name sdd.md ... | head -1)`) and `tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py` never passes it either — so `registry_findings()` (`audit_caseplan.py:439-444`) and the surviving-`<UNRESOLVED>` WARN that Rule 8 and the Phase 5 Open Items depend on are never executed by anything. Either add `--registry` to the three eval commands or unit-test the function. _(trigger: skills/uipath-maestro-case/scripts/audit_caseplan.py)_ _(restates: Axis 4: New Step-12 completeness gate ships with 6 untested MISSING branches)_

**Downstream consumers:**
- 🟡 Retargeting the `[REGISTRY LOOKUP FAILED: <name> in <folder>]` marker into `registry-resolved.json` (`skills/uipath-maestro-case/references/registry-discovery.md:256`) left **two** unresolved-marker consumers un-updated, not one. Besides the gate's `UNRESOLVED = re.compile(r"<UNRESOLVED[^>]*>", re.I)` at `skills/uipath-maestro-case/scripts/audit_caseplan.py:39`, the eval grader `tests/tasks/uipath-maestro-case/_shared/check_issue_journal.py:60` now scans the same file with `re.findall(r"<UNRESOLVED", ...)` — so a build whose only unresolved work is recorded in the second marker vocabulary is scored as carrying nothing the operator must be told about (it survives only via the checker's separate `data: {}` placeholder count). Collapsing the marker to the Rule 8 form fixes both consumers at once. _(trigger: skills/uipath-maestro-case/references/registry-discovery.md)_ _(restates: Axis 8: registry-discovery.md:256 retargets the `[REGISTRY LOOKUP FAILED: ...]` marker into registry-resolved.json)_

**Docs & config:**
- 🟠 Deleting the `| **Tasks file** | ... |` row from `skills/uipath-planner/assets/templates/case/case-sdd-template.md` (between lines 31 and 32) left every planner doc that specifies that row untouched: `skills/uipath-planner/SKILL.md:39` (Rule 5 still enumerates `Tasks file` among the required handoff fields for every SDD), `skills/uipath-planner/references/sdd-generation-guide.md:473` (canonical header block, whose `Project list section` line explicitly covers case), and `skills/uipath-planner/references/pdd-driven-lane-guide.md:25`, `:29`, `:56`, `:181` (reads the field, defaults it to `<sdd-basename>-tasks.md` when missing, surfaces it as a defaulted field, and forbids the per-project tasks file that default produces for a case child). Neither planner reference is in this PR's changed-file set. Land the rename in the template *and* those four documents in one change, or restore the row. _(trigger: skills/uipath-planner/assets/templates/case/case-sdd-template.md)_ _(restates: Axis 8: case-sdd-template.md drops the `Tasks file` handoff row that planner Rule 5 and this PR's fixtures still require)_

**Rollout impact:**
- 🟡 The new unit suite guards a file the CI trigger cannot see. `tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py:15-25` loads its subject by path from `skills/uipath-maestro-case/scripts/audit_caseplan.py`, but the `pytest-case-check` job that runs it (`.github/workflows/test-helpers.yml:55-69`) is gated by a `paths:` filter (lines 5-14) listing only `tests/tasks/...`, `tests/scripts/**`, `scripts/**`, `hooks/**` and the workflow itself — no `skills/**`. A future PR that edits only the gate script runs no test at all, and `test-helpers.yml` is the sole workflow invoking pytest. The filter already carries the precedent comment `- 'scripts/**'  # guarded scripts (e.g. build-uip-catalog.py) must trigger their tests`; add `skills/uipath-maestro-case/scripts/**` alongside it in this PR, since this PR is what creates the cross-tree dependency. _(trigger: tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py)_
- 🔵 Blast radius on the activation gate is unstated. This PR rewrites the skill's frontmatter `description` (`skills/uipath-maestro-case/SKILL.md:3`) *and* mutates one of the 50 measured positives (`tests/tasks/activation/uipath-maestro-case.jsonl:3`), while `tests/scripts/activation_gate.py:69` keeps `"uipath-maestro-case": 90` — a figure the surrounding comment (lines 30-38) pins to a 2026-06-17 full-set measurement of the *previous* row set. The gate will run (activation-gate.yml has no paths filter), but a pass/fail against a baseline measured over a different set of prompts is no longer apples-to-apples. Say in the PR whether a re-baseline is needed, or leave the row set alone when the description changes. _(trigger: tests/tasks/activation/uipath-maestro-case.jsonl)_ _(restates: Axis 4: Activation row 003 rewritten into a near-duplicate of row 002)_

## Guardrails & Automation

**Static checks (lint / repo-local):**
- Retired-vocabulary registry + scoped gate, cloning the repo's existing `cli-renames` mechanism exactly. NOTE on tier: rubric tiers 1/2/4 are unavailable here — no linter and no typechecker are configured for this repo's Python or Markdown in any workflow, and CodeQL is not enabled on UiPath/skills — so every static … _Prevents:_ A6-medium `connector-integration.md:59` surviving `tasks.md` write instruction and `:170` `## Output Contract to Tasks.md` heading; A8-medium …
- Cross-skill SDD heading-grammar contract guard, as a `tests/scripts/test_*.py` static guard in the exact shape of `tests/scripts/test_runtime_payload_key_casing.py` (regex scan over source files, module docstring stating the background and the run command, no execution of the product). Add … _Prevents:_ A1-high `audit_caseplan.py:34` `TASK_HEADING` rejecting the documented `##### Task S<K>.<M>:` secondary-stage form (caught statically, today); A6-high the …
- Deterministic task-YAML criteria gate, following the `scripts/check-task-driver.py` + `.github/workflows/task-driver-gate.yml` pattern (stdlib + PyYAML, sub-second full scan of `tests/tasks`, no `paths:` filter so it can be a required check, exit 1 with paths printed). Add `scripts/check-task-criteria.py` with two … _Prevents:_ A4-high `build_from_planner_sdd/build_from_planner_sdd.yaml:52` where all five surviving criteria are negations and a no-op agent scores 11/11; A3/A4-medium …
- Orphan reachability pass in the existing link checker. `.claude/rules/skill-structure.md` already states the rule — "No orphaned files (every file is reachable from SKILL.md)" — and `.claude/rules/skill-review.md` lists "unreachable files" as a Structure red flag, but nothing enforces it. … _Prevents:_ A8-low `skills/uipath-maestro-case/references/evals/evals.json` carrying stale `plan-only` expectations while being unreachable from SKILL.md and from any …
- JSON-artifact fence-language guard, as a `tests/scripts/test_*.py` static guard in the `test_runtime_payload_key_casing.py` shape. Scan `skills/**/*.md`; for each fenced code block, look at the nearest preceding heading and the sentence introducing the fence — when either names a `*.json` artifact and describes its … _Prevents:_ The five-axis merged finding at `references/placeholder-tasks.md:82` (`## \`registry-resolved.json\` Entry Shape` over a ```markdown H1 + hyphen-bullet block, …
- argparse-when-flagged convention guard. The repo convention is argparse for any script that accepts flags — `build-uip-catalog.py`, `check-cli-verbs.py`, `check-skill-status.py`, `check-skill-verbs.py`, `check-skills-sh.py`, `prune-task-defaults.py` all use it; the two that do not (`check-task-driver.py`, … _Prevents:_ A2-low `audit_caseplan.py:451` hand-rolled `--sdd`/`--registry` parsing with a string-keyed setter, which also loses argparse's free behavior: …
- Documented-invocation parity for repo-owned skill scripts. Add a `tests/scripts/` guard that, for each `skills/<skill>/scripts/<name>.py`, collects every invocation of that script shown in the skill's own Markdown (`SKILL.md` and `references/**`) and every invocation shown in the eval task YAMLs under … _Prevents:_ A8-low `SKILL.md:38` Rule 6 documenting the mandatory Step 12 gate as `scripts/audit_caseplan.py <caseplan.json> --sdd sdd.md` while …

- Empty-parse fail-loud, plus the regression test that pins it. In `skills/uipath-maestro-case/scripts/audit_caseplan.py`, add a guard in `main()` before the success print at line 481: when `parse_sdd()` yields no stages AND no variables AND no triggers AND no case-exit rows, exit non-zero with `AUDIT FAIL -- parsed 0 … _Prevents:_ A7-high `audit_caseplan.py:481` printing `AUDIT OK: caseplan.json covers every SDD element` (exit 0) against a completely empty caseplan when the SDD parses to …
- Golden parity fixture that runs the parser over the planner template. Add to `tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py` (or a new `test_sdd_template_parity.py` next to it) a test that fills the placeholders of `skills/uipath-planner/assets/templates/case/case-sdd-template.md` into a temp SDD — … _Prevents:_ A1-high `audit_caseplan.py:34` — the `TASK_HEADING` regex silently downgrading every exception-lane task to `entry_rows=0`, `identity_resolved=False`, which …
- Negative control per MISSING branch, with a count guard so the set cannot rot. `compare()` in `audit_caseplan.py` has 15 `missing.append(` sites; 9 have negative controls in `tests/tasks/uipath-maestro-case/_shared/test_audit_caseplan.py` and 6 do not (stage entry rows declared/caseplan has none, stage exit rows, … _Prevents:_ A4-medium — 6 untested MISSING branches in the new Step-12 gate, including dropped stage-entry rules and a dropped trigger, which are precisely the …
- Assert every key the parser writes is consumed by `compare()`. Add a test that introspects the dicts returned by `parse_sdd`/`parse_stage`/`section_blocks` on the shared fixture and asserts their key sets are a subset of the keys `compare()` reads (extractable from the script source, or maintained as an explicit … _Prevents:_ A2-medium `audit_caseplan.py:76` — six computed-but-never-read fields, two of which (`Activation Mode`, `Required`) advertise checks the gate does not perform, …
- Fixture-equality guard on every task that grades one agent-writable artifact against another. `audit_caseplan.py <caseplan> --sdd $(find . -name sdd.md … | head -1)` compares two files the agent holds Write/Edit on, so trimming a stage out of the staged SDD makes the gate report `AUDIT OK` on an incomplete caseplan. … _Prevents:_ A4-low — the new `audit_caseplan.py` criterion (weight 3.0) in `expense_runnable_e2e.yaml:97` and `aged_invoice_structural.yaml:95` being satisfiable by …
- Near-duplicate review pass over the activation recall sets. When a row in `tests/tasks/activation/<skill>.jsonl` is rewritten rather than removed, check the rewritten prompt against the other rows in the same file for signal collapse, and either retarget it at vocabulary no other row covers or drop it and note the … _Prevents:_ A4-low `tests/tasks/activation/uipath-maestro-case.jsonl:3` — row 003 rewritten from "Generate tasks.md from my sdd.md" into "Generate caseplan.json from my …
- Advertised-coverage parity for `_shared` checker modules. When a check function is deleted from a shared checker, the module docstring's coverage sentence must be updated in the same change, and any invariant that has no surviving home must be either re-homed or explicitly recorded as dropped. Concretely for … _Prevents:_ A4-low `_shared/sdd_check.py:33` — two deleted SLA invariants with no new home while the module docstring still claims to cover them, so a future reader trusts …
- Close the Rule 6 / Rule 13 contradiction inside `skills/uipath-maestro-case/SKILL.md` and drop the harness's compensating hack. Rule 6 (`SKILL.md:38`) now makes a Python script that reads `caseplan.json` a mandatory pre-Phase-4 step, while Rule 13 (`SKILL.md:45`) bans Python, scripts and shell redirection over exactly … _Prevents:_ A5-low `SKILL.md:45` — the artifact-I/O guardrail telling the agent not to do the thing Rule 6 requires, which teaches the agent that the repo's primary …

## Top 5 Priority Actions

1. Restore the empty-parse guard the deleted audit_plan.py had at skills/uipath-maestro-case/scripts/audit_caseplan.py:481 — exit non-zero naming the SDD path and the expected '### Stage <N>:' / '#### Tasks' heading shapes when parse_sdd() yields no stages, variables, triggers or case-exit rows, since today a prose-only or one-heading-level-shifted SDD prints AUDIT OK against a completely empty caseplan and green-lights it through Phase 4 to publish.
2. Widen TASK_HEADING at skills/uipath-maestro-case/scripts/audit_caseplan.py:34 to accept the template-mandated '##### Task S<K>.<M>:' form (r"^Task\s+S?[\d.]+\s*[:.]\s*(.+)$"), because on the PR's own graded fixture every secondary-stage exception-lane task parses as entry_rows=0 / identity_resolved=False, making the placeholder-honesty and entry-rule-content branches unreachable dead code.
3. Make the rule and row checks actually check content in skills/uipath-maestro-case/scripts/audit_caseplan.py:347 and :318 — drop the task["entry_rows"] guard and inspect the inner rule group, and convert the stage entry/exit and case-exit presence tests to strict count comparisons — since 'rules: []', 'rules: [[]]' and silently dropped condition rows all pass clean today despite the 1:1 no-merge contract in implementation.md:13 and the SKILL.md:216 claim that the gate proves nothing was dropped.
4. Repair the two eval regressions that let non-work score full credit: add a skill_triggered plus llm_judge anchor to tests/tasks/uipath-maestro-case/build_from_planner_sdd/build_from_planner_sdd.yaml:52, whose five surviving criteria are all negations so a refusing agent scores 11/11, and rewrite the multi-operand greps at tests/tasks/uipath-maestro-case/hitl/quality_01_action_schema_design.yaml:229/237/245 into the 'grep -q A || grep -q B' form used by quality_02:202, since grep exits 2 when either operand is missing and turns a promised OR into an AND worth 5.0 weight.
5. Finish the tasks.md sweep so the documented artifact shape matches the JSON contract: rewrite the entry-shape block at skills/uipath-maestro-case/references/placeholder-tasks.md:82 as a real JSON object using SKILL.md Rule 9's keys plus wiringNotes (and fix the same stale headings at connector-integration.md:170 and connector-trigger-planning.md:274), change the surviving 'mark the task ... in tasks.md' write instruction at connector-integration.md:59 to registry-resolved.json, unify the second marker vocabulary at registry-discovery.md:256 onto '<UNRESOLVED: ...>', and land the Tasks-file/Build-handoff row decision in skills/uipath-planner/assets/templates/case/case-sdd-template.md:31 together with SKILL.md Rule 5 and sdd-generation-guide.md:473.

---

**Change class:** complex — removes a documented skill artifact (tasks.md) from the maestro-case contract, renames the audit script consumed by task YAMLs, and changes the Step 12 completeness gate; correctness requires reasoning about what still references the retired artifact

**Stats:** 0 🔴 · 7 🟠 · 7 🟡 · 13 🔵 across 8 axes · **Overall 8.5/10** · weakest **Correctness & Logic 6.9/10**.
**Verification:** 31 medium+ findings adversarially re-checked · **9 dropped as false positives** (29%) · 20 corrected in place · 15 low passed through unverified.

<sub>Generated by `/code-review-wf` (8-axis workflow review, 43 agents, per-finding adversarial verification, deterministic scoring). Medium/nit/guardrail entries are abridged here to fit GitHub's 65k comment limit; the unabridged per-axis reports were produced alongside this comment.</sub>

RaduAna-Maria (approved):

PR #2955, opened 2026-09-01
Stop grading the consumables diagnose task on an optional flag

biftime:
## Problem

[`skill-platform-licensing-diagnose-consumables-scope`](https://coder-evalboard.uipath-dev.com/runs/2026-08-27_04-15-19/skill-platform-licensing-diagnose-consumables-scope) failed at **0.692** with a complete, correct diagnosis on disk.

Two criteria required a literal `--mode summary` in the shim log. `summary` is the CLI default mode, and the skill explicitly tells the agent to omit it:

> Summary report: omit `--mode` (summary is the default) or pass `--mode summary`.
> — `skills/uipath-platform/references/licensing/consumables-report.md:17`

The agent ran all three commands the task wanted, wrote a correct `verdict.md`, and scored 0 on both — docked for following its own skill. This is the case `.claude/rules/test-writing.md` warns about: grade the outcome, not the literal flag.

## Changes

| Line | Change |
|---|---|
| 42, 50 | Identify summary mode by excluding `daily\|folders` rather than requiring `--mode summary` — same idiom as the sibling smoke task `consumables_summary.yaml:31` |
| 71 | Observation-1 criterion no longer accepts bare `scoped`; requires a causal word (`suppress`, `deliberat`, `by design`, `no tenant attribution`, …) |
| 86 | Verdict criterion requires an explicit not-broken statement; dropped standalone `correct\|expected\|by design` |
| 37 | Dropped `Use --output json on every uip command` from the prompt — no criterion grades it and the skill teaches the convention |

The last three are pre-existing gameability gaps found by `/lint-task` while fixing the first.

## Verification

No new coder-eval run. I replayed the full criteria set locally against two fixtures:

| Fixture | Before | After |
|---|---|---|
| Real artifacts from run `2026-08-27_04-15-19` (`.uip-cmdlog` + `verdict.md`, pulled from blob storage) | 0.692 | **1.000** |
| Deliberately wrong verdict — "reporting IS broken", using the words `not correct`, `expected`, `scoped to the tenant` | 0.846 | **0.077** (only `file_exists` passes) |

The wrong-verdict fixture is load-bearing: my first tightening of criterion 71 still matched it, because "scoped to the tenant" is vocabulary any consumables write-up uses regardless of conclusion. Dropping the scope-wording alternatives fixed it. The old criterion 86 also passed that fixture, confirming it was a real hole.

Also checked that a log containing only `--mode folders` / `--mode daily` still fails both summary criteria, so the exclusion didn't make them trivially passable, and that the `--mode=summary --tenant=default` equals-form still matches.

## Note

`scripts/check-cli-verbs.py` crashes on Python 3.9/3.10 (`sre_parse.POSSESSIVE_REPEAT` is 3.11+), so `/lint-task`'s CLI-verb axis silently skips. Vacuous for this task (no `command_executed` criteria), but worth a separate fix.

### DIFF
--- a/tests/tasks/uipath-platform/licensing/diagnose_consumables_scope.yaml
+++ b/tests/tasks/uipath-platform/licensing/diagnose_consumables_scope.yaml
@@ -83,8 +83,8 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: run_command
-    description: "Verdict concludes reporting is NOT broken. Must state it explicitly — 'correct'/'expected'/'by design' alone are too weak, they appear in wrong verdicts too."
-    command: "grep -Eiq 'not (actually )?broken|isn.?t broken|working as (designed|intended|documented|expected)|not a (bug|defect)' verdict.md"
+    description: "Verdict concludes reporting is NOT broken. Needs a negative-polarity statement (hedges like 'not shown to be broken' count); bare 'correct'/'expected'/'by design' are too weak — they appear in wrong verdicts too."
+    command: "grep -Eiq 'not( (shown|demonstrated|proven|actually|really|truly|clearly|evidently|apparently|to|be))* broken|is ?n.?t broken|no evidence (of|that)[^.]{0,60}(broken|defect|bug)|(is|are) working as (designed|intended|documented|specified)|works? as (designed|intended|documented|specified)|not a (bug|defect)' verdict.md"
     timeout: 10
     expected_exit_code: 0
     weight: 1.0
--- a/tests/tasks/uipath-platform/licensing/diagnose_consumables_scope.yaml
+++ b/tests/tasks/uipath-platform/licensing/diagnose_consumables_scope.yaml
@@ -34,20 +34,19 @@ initial_prompt: |
   - The `uip` CLI is available but is NOT connected to a live tenant in this
     environment. Commands may fail with auth errors — that is expected. Run
     each command once and move on. Do NOT retry or attempt to login.
-  - Use `--output json` on every uip command.
 
 success_criteria:
   - type: run_command
-    description: "Agent ran a tenant-scoped consumables summary to reproduce observation 1 (.uip-cmdlog shim log)"
-    command: "grep -E 'uip platform licenses consumables get' .uip-cmdlog | grep -E -- '--mode[= ]summary' | grep -Eq -- '--tenant[= ]default'"
+    description: "Agent ran a tenant-scoped consumables summary to reproduce observation 1 (.uip-cmdlog shim log). Summary is the CLI default mode, so an omitted --mode counts."
+    command: "grep -E 'uip platform licenses consumables get' .uip-cmdlog | grep -Ev -- '--mode[= ](daily|folders)' | grep -Eq -- '--tenant[= ]default'"
     timeout: 10
     expected_exit_code: 0
     weight: 2.0
     pass_threshold: 1.0
 
   - type: run_command
-    description: "Agent also ran the summary UNSCOPED, so scoped and account-wide figures could be compared — the diagnostic move, not just an assertion about it"
-    command: "grep -E 'uip platform licenses consumables get' .uip-cmdlog | grep -E -- '--mode[= ]summary' | grep -v -- '--tenant' | grep -q ."
+    description: "Agent also ran the summary UNSCOPED, so scoped and account-wide figures could be compared — the diagnostic move, not just an assertion about it. Summary is the CLI default mode, so an omitted --mode counts."
+    command: "grep -E 'uip platform licenses consumables get' .uip-cmdlog | grep -Ev -- '--mode[= ](daily|folders)' | grep -v -- '--tenant' | grep -q ."
     timeout: 10
     expected_exit_code: 0
     weight: 2.0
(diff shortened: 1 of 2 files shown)


biftime:
CI run [33513707549](https://github.com/UiPath/skills/actions/runs/33513707549) moved this task **0.692 → 0.923** — all four criteria from the first commit passed against a live agent. Two failures remained; both were false negatives on over-strict criteria, now fixed in 5159562.

### 1. My own over-correction (`diagnose_consumables_scope.yaml`)

The agent's verdict read:

> Consumption reporting **is not shown to be broken**. Both observations are explained by documented report semantics.

Correct conclusion — but my tightened regex required `not broken` adjacently, so the hedge in the middle failed it. Now filler words are allowed between the negation and `broken`, drawn from a whitelist (`shown`, `demonstrated`, `to`, `be`, `actually`, …) rather than an open `.{0,N}` gap, so a wrong verdict cannot bridge the two words with unrelated prose. The `working as designed` alternative is guarded behind `is|are|works` so `not working as intended` can't satisfy it.

Re-verified across six fixtures — two real agent verdicts, one alternate correct phrasing, three adversarial wrong verdicts:

| Fixture | Want | Result |
|---|---|---|
| CI run verdict — "is not shown to be broken" | accept | accept |
| Blob run verdict — "is not broken" | accept | accept |
| "reporting works as designed" | accept | accept |
| "reporting IS broken", uses `not correct` / `expected` | reject | reject |
| "not correct **and** reporting is broken" | reject | reject |
| "is **not** working as intended" | reject | reject |

### 2. Pre-existing, unrelated to this PR (`users_licenses_get.yaml`)

Not caused by this change — the workflow runs the whole `licensing/` glob, so this PR's diff surfaced it. The agent's reply listed:

> - Attended - Named User (`ATTUNU`)

That **is** the required `<Friendly Name> (<CODE>)` pairing. The `llm_judge` scored it 0.0 with the rationale: "Each listed code is enclosed in backticks inside the parentheses, so the required format is not met." Failed on markdown typography alone.

The judge prompt now says to grade the pairing, not the typography, with that exact reply as a worked 1.0 example. This is a second logical change in one PR — happy to split it out if you'd rather, but the job can't go green without it.

biftime:
Converted the `users_licenses_get` format check to a deterministic artifact check (132c9cb), replacing the markdown-tolerance patch from 5159562.

The criterion was an `llm_judge` over `include_agent_output` — grading the agent's monologue, which `.claude/rules/test-writing.md` rules out ("Use side effects… Never grade on agent monologue"). Loosening the judge prompt fixed the false negative but left the criterion type wrong.

**Now:** the prompt asks for the answer in `bundles.md`, and a regex grades the artifact — the same pattern the sibling diagnose tasks (`verdict.md`, `analysis.md`, `findings.md`) already use.

```
(Attended|Unattended|Developer|Tester|Named User|Citizen|Automation)[A-Za-z .*`-]{0,40}\( *[`*]* *[A-Z][A-Z0-9]{3,} *[`*]* *\)
```

It requires a friendly name adjacent to its parenthesised code, tolerating markdown around either half. Verified by running the criterion exactly as the YAML invokes it:

| Fixture | Want | Result |
|---|---|---|
| The real CI reply — ``- Attended - Named User (`ATTUNU`)`` | accept | accept |
| Plain text — `Automation Developer Named User (RPADEVPRONU)` | accept | accept |
| Bold — `**Attended Named User** (**ATTUNU**)` | accept | accept |
| Codes only, no friendly names | reject | reject |
| Friendly names only, no codes | reject | reject |
| `bundles.md` absent | reject | reject |

The last two rejections are the exact failure modes the old judge prompt described, so the deterministic check preserves its intent while removing the LLM from the loop.

Weights: the judge was 2.0; now `file_exists` 1.0 + format regex 2.0.

biftime:
Smoke gate [33515416556](https://github.com/UiPath/skills/actions/runs/33515416556) caught a gap in my own conversion. Fixed in 8fcf726.

The agent wrote `bundles.md` (that criterion passed) with a correct, honest answer:

> No bundles could be verified. The single `uip platform users licenses get …` lookup returned: *No directory user found matching `dan.dinu@uipath.com`.*
>
> The friendly-name/code mappings consulted are recorded in [license-codes-evidence.json](./license-codes-evidence.json).

Nothing wrong with that. The format criterion simply had no pairs to match, because the mappings went to the evidence file — exactly where the prompt asked for them.

**Root cause:** the offline-fallback bullet says to resolve the name → code mappings "so you can demonstrate the required reporting format" but never names a destination. That was harmless while the criterion graded the chat reply, where the demonstration naturally landed. Moving the answer into an artifact left the bullet pointing nowhere.

The bullet now names `bundles.md` as the destination, so the format is demonstrated whether or not the lookup returns bundles.

Re-verified against the real artifact from that run:

| Fixture | Want | Result |
|---|---|---|
| `bundles.md` exactly as the agent wrote it (no pairs) | reject | reject |
| Same answer with the mappings listed | accept | accept |

The first row is the important one: the criterion is still rejecting the artifact that failed, so this is a prompt fix, not a criterion loosened until it passed.

Worth flagging that this task is inherently a format demonstration — the sandbox has no live tenant, so the lookup always fails and there are never real bundles to report. That's pre-existing task design, not something this PR changed.

dinud-uipath (approved):

PR #2957, opened 2026-09-01
feat: add HOL Guard skill for protected UiPath agent sessions

kantorcodes:
## What

Adds `uipath-hol-guard`, a skill for using the published `hol-guard` CLI around supported local coding-agent sessions before state-changing `uip` operations.

## Scope

- checks for the published `hol-guard` CLI and installs it only with user approval
- detects a supported local harness before wiring protection
- uses `hol-guard install <HARNESS>`, status verification, and a dry-run check before mutation-bearing work
- requires the next mutation-bearing coding-agent session to launch through `hol-guard run <HARNESS>`
- keeps UiPath authentication, tenant permissions, confirmations, validation, and native agent guardrails authoritative
- does not claim HOL Guard runs inside UiPath Cloud or intercepts server-side UiPath APIs
- explicitly forbids recreating Guard as a custom `uip` wrapper or substitute security layer

## Repository integration

- registers the skill as `preview`
- adds it to the Platform & Operations grouping
- updates the generated lifecycle-status row in README
- adds CODEOWNERS coverage
- adds one smoke and one e2e coder-eval task

## Validation

- parsed both task YAML files successfully
- parsed SKILL frontmatter successfully
- parsed the JSON registration files successfully
- verified the upstream-facing diff contains only the intended seven files

Full repository CI and coder-eval were not run locally in this environment.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -202,3 +202,7 @@
 # StudioAdmin AOps skill (SourceControl + CICD pipelines)
 /skills/uipath-aops/ @AndreiRR24 @UiPath/xworks
 /tests/tasks/uipath-aops/ @AndreiRR24 @UiPath/xworks
+
+# HOL Guard skill (local coding-agent execution protection)
+/skills/uipath-hol-guard/ @gabrielavaduva @AlvinStanescu @marius-bughiu @RaduAna-Maria @uipreliga @dmetzgar @rockymadden @gozhang2 @smflorentino
+/tests/tasks/uipath-hol-guard/ @gabrielavaduva @AlvinStanescu @marius-bughiu @RaduAna-Maria @uipreliga @dmetzgar @rockymadden @gozhang2 @smflorentino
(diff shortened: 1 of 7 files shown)


uipreliga:
Thanks for the careful work here — the repo integration is genuinely well done (status manifest, skills.sh grouping, regenerated README row, CODEOWNERS, correct tag taxonomy on both tasks). This isn't a quality decline.

We're closing on scope. `skills/` ships skills for UiPath product surface, maintained by UiPath owners. `hol-guard` is a third-party runtime-protection CLI, and merging this would put a `pipx install hol-guard` instruction in front of every consumer of the plugin — a dependency and vendor-trust decision that needs a named internal sponsor and a security review, not a docs review. The proposed CODEOWNERS also assigns nine UiPath engineers as owners of a tool none of them maintain.

A skill like this fits better in HOL's own plugin repo, where you can version it against your CLI directly.

Two notes if you take it there: both task YAMLs gate their only positive criterion on `hol-guard detect --json`, which we couldn't find in the published command surface — and as fork PR workflows were never approved, only 3 of the required checks ran here, so neither task has actually been executed.

PR #2960, opened 2026-09-01
fix(uipath-maestro-flow): bound run_debug by a deadline, fund its retry, enforce both

rockymadden:
## Problem

`run_debug(timeout=N)` caps **one subprocess**. The task YAML's `timeout:` caps the **whole check process**. Nothing reconciled the two, so a retry runs past the budget the criterion granted and the grader SIGKILLs the check mid-attempt.

That kill preempts `_fail_with_capture`, so nothing survives: no payload, no `instanceId`, no CLI envelope. A live-tenant hiccup scores a gating criterion 0 and reads as "the agent built the flow wrong".

## Evidence

`skill-flow-api-workflow`, 2026-09-01 (FAILURE, weighted 0.375, criterion 1 of 2 passed):

```
05:23:10 criteria.run_command: Running command ... check_api_workflow_flow.py
05:28:10 sandbox: Command 'python3 $TASK_DIR/check_api_workflow_flow.py' timed out after 300 seconds
```

`check_api_workflow_flow.py:17` calls `run_debug(timeout=240)` under a `timeout: 300` criterion. One retry needs `240 + 5 + 240 = 485s`. The child was still alive at 300s, and only a **second** `subprocess.run` explains that: a single stalled attempt raises `TimeoutExpired` at 240s and exits as a graded FAIL carrying the partial output.

## Prior art

**This is not a new class of failure.** `single_node/subflow/test_check_subflow_flow.py` guarded the same relation for one task, added after an RCA on 2026-05-29 describing the identical symptom — criterion budget below the subprocess timeout, sandbox kills the process, exit `-1` with empty output. It covered 1 task out of 44, and `api_workflow` hit the same wall four months later.

That per-task test is absorbed here, and its RCA text preserved in the guard's docstring since the doc it referenced is no longer in the tree.

## Change

**1. Separate the attempt from the deadline.** `timeout` still caps one attempt, so no call site changes. New `budget` is the deadline for the whole call, defaulting to `debug_budget()`: `retries` full attempts plus backoff. An attempt gets `min(timeout, remainder)`, so the call cannot outlive its deadline however the retries fall. A retry the remainder cannot fund is refused, and the run fails through `_fail` carrying the CLI's own envelope.

`_DEFAULT_RETRIES` drops 3 → 2 so `debug_budget` funds every attempt `retries` promises without moving any ceiling. A fast 5xx now gets one retry rather than two, matching the CLI's own "retry once before reporting".

**2. Raise the criterion ceilings** so the retry is affordable rather than nominal. 43 raised (a 44th, `escalation_orchestrator_paths`, keeps its value and gains only the `manual x7` annotation):

| was | now | count |
|---|---|---|
| 300 | 600 | 25 |
| 320 / 360 | 720 | 4 |
| 540 | 1080 | 1 |
| 600 | 720 | 2 |
| 600 | 1200 | 4 |
| 660 | 1320 | 2 |
| 900 | 1080 | 2 |
| 900 | 1800 | 1 |
| 1200 | 1320 | 2 |

Raising is close to free: the criterion timeout is a cap, not a sleep. A passing debug returns in 18-26s (measured in #2776) and the criterion exits with it. Higher ceilings only cost wall clock on runs that were already failing.

**3. Add four `task_timeout` declarations.** `wiki_pageviews`, `feet_inches`, `rpa` and `decision` inherit 1200s from `experiments/default.yaml`, and their grading alone now exceeds it, so no agent run could fit. Each declares a cap with its arithmetic inline.

**4. Enforce both relations** in `_shared/test_criterion_budgets.py`, across 120 criteria and 126 tasks:

```
criterion timeout >= debug_budget(...) + CRITERION_MARGIN_SECONDS
task_timeout      >  worst-case grading
```

The first is computed statically from each check's AST: straight-line `run_debug` calls sum, exclusive branches take the worst arm, helper calls cost inline at the call site, and a loop over a module-level literal multiplies out. The second resolves inherited caps from the experiment defaults, so the 56 tasks that declare no `task_timeout` are covered rather than skipped.

Anything the guard cannot count is **refused, never assumed**. A runtime-seeded loop must declare `# budget-guard: manual xN` on its timeout line, and that count is cross-checked against the task's `seed.py`. A bare marker, a stale one, an uncorroborated count and a script with two uncountable loops all fail.

## Scope

38 of 44 `run_debug` criteria could not survive one retry. `_POLL_TIMEOUT_ATTEMPTS = 2` was sized in #2776 against "the 600s criterion budget in the task YAMLs"; most used 300s.

`uipath-maestro-case` is unaffected — its own `run_debug` (`_shared/case_check.py:458`) has no retry loop. It has no equivalent guard either; filed as #2962.

## Verification

- `pytest tests/tasks/uipath-maestro-flow/` → **512 passed**
- `pytest tests/scripts/test_runtime_payload_key_casing.py` → 4 passed
- Rest of `tests/tasks` → 739 passed, 14 skipped
- Two `uipath-maestro-case` collection errors are pre-existing on `origin/main`, confirmed by stashing this diff

Every guard rule has a negative control: dropping a criterion below its budget, removing a `manual` marker, understating its count against the seed, or reverting a `task_timeout` each fails with the exact arithmetic.

## Known limits

- The criterion floor is a floor. `CRITERION_MARGIN_SECONDS = 60` is a chosen value matching the gap the suite already used most often, not a measured one.
- `task_timeout > grading` is an impossibility check, not a sufficiency one. It cannot prove a task fits, only catch ones that provably cannot — anything stronger needs the agent's real wall clock.
- A task that would rather not pay for a retry passes `retries=1`; `debug_budget` drops to a single attempt and both guards accept the smaller ceiling.

## Deliberately not included

The four live Jira checkers spend 480-960s outside `run_debug` that the 60s margin does not cover, so three of their criteria remain under-funded. Fixing that requires bounding probe loops in checkers that create and delete issues in a shared project and only run against a live tenant. That work was attempted, found to be bug-prone, and reverted rather than merged unexercised — filed as #2968 with a note that it needs a live-tenant run.

Those criteria are therefore funded exactly as `main` funds them today. Pre-existing gap, not a regression.

## Follow-up

The `variables.nodes[]` / "declare `error` only" doc defect that the same eval exposed is a `skills/uipath-maestro-flow/` change with different CODEOWNERS. It ships separately so its eval evidence is not measured through this flake.


tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py
```diff
@@ -0,0 +1,170 @@
...
+    None when the path runs no debug at all."""
+    tree = ast.parse(open(path).read())
+    scope: ast.AST = tree
+    if subcommand:
+        named = [
+            n
+            for n in ast.walk(tree)
+            if isinstance(n, ast.FunctionDef) and n.name == subcommand
+        ]
+        if not named:
```
rockymadden: 🔴 **Critical: silent skip defeats the guard.**

This only resolves a subcommand that is a `FunctionDef`. `check_wiki_pageviews_flow.py:74` dispatches with `if/elif` on `sys.argv[1]` inside `main()`, so `named` is empty, this returns `None`, and `test_criterion_clears_the_debug_budget` returns without asserting.

The guard's own resolution table:

```
multi_node/wiki_pageviews.yaml  sub='uipath_success'  crit=600  budget=None
multi_node/wiki_pageviews.yaml  sub='invalid_error'   crit=600  budget=None
```

Both call `run_debug(timeout=300)`. They need **665s and have 600s**. So this PR ships a live instance of the exact bug it claims to fix, hidden by the mechanism meant to catch it.

This is also the failure mode `test-helpers.yml` already warns about: *"a synthetic payload shape kept 29/29 green while the real parser read nothing (PR #2667)"*.

**Fix:** an unresolvable subcommand on a module that contains `run_debug` must `pytest.fail`, never return `None`. Charge the whole module when dispatch cannot be resolved statically. Then raise wiki_pageviews to 720.

tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py
```diff
@@ -0,0 +1,170 @@
...
+        ]
+        calls = [c for b in bodies for c in _calls_in(b)]
+    else:
+        calls = _calls_in(scope)
+    if not calls:
+        return None
+    budgets = [_budget_of_call(c) for c in calls]
+    if any(b is None for b in budgets):
+        pytest.fail(f"{path}: run_debug called with a non-literal timeout/budget")
+    return max(budgets)
```
rockymadden: 🔴 **Critical: `max` where it needs `sum`.**

`check_decision_flow.py:26,30` makes two straight-line `run_debug` calls in one `main()`:

```python
payload = run_debug(inputs={in_vars[0]: 90})
payload = run_debug(inputs={in_vars[0]: 50})
```

Neither is in a loop or a branch. Both always run. `max()` returns 485 so the guard demands 545, but the true worst case is 970, needing 1030. `decision.yaml` was set to **600, under by 430s**, and the guard passes it.

The module docstring's "a check that LOOPS pays per iteration, so this is a floor" does not cover this: both calls are unconditional and statically countable.

**Fix:** sum the calls on an execution path instead of taking the max. Keep the loop caveat for genuine loops. Raise decision to 1080.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -148,37 +160,56 @@ def _as_text(raw: bytes | str | None) -> str:
...
+    timeout: int = 240, retries: int = 3, backoff_seconds: float = 5.0
+) -> int:
+    """Worst-case wall clock for one :func:`run_debug` call.
+
+    A poll timeout is the expensive failure: it burns a whole attempt and is
+    the only transient we retry at full length, capped at
+    :data:`_POLL_TIMEOUT_ATTEMPTS`. ``retries=1`` opts out of retrying, so its
+    budget is a single attempt. Exposed so test_criterion_budgets.py can hold
+    every task YAML to the same arithmetic instead of eyeballing it."""
+    attempts = 1 if retries <= 1 else _POLL_TIMEOUT_ATTEMPTS
```
rockymadden: 🟠 **Major: `retries=3` now silently means 2.**

This sizes on `_POLL_TIMEOUT_ATTEMPTS` (2), but `run_debug`'s loop is `for attempt in range(retries)` with `retries=3`. A poll timeout is capped at 2 by design. A **slow 5xx** is not, yet its third attempt is unfunded and the deadline kills it:

```
retries=3 budget funds attempts: 2 but loop allows 3
```

That quietly narrows the 5xx retry #2776 added for customer-escalation-triage, and the parameter now lies about what it does.

**Fix:** either size on `retries` (default budget 485 -> 730, which cascades into every ceiling), or change the default to `retries=2` and document it. I'd take the second: nothing in the repo justifies a third full-length attempt.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -148,37 +160,56 @@ def _as_text(raw: bytes | str | None) -> str:
...
+    return timeout * attempts + int(backoff_seconds) * (attempts - 1)
```
rockymadden: 🟡 **Minor: `int()` truncates the backoff.**

`debug_budget(240, 3, 2.5)` returns 482; the real worst case is 482.5. Sub-second and no current caller passes a fractional backoff, but this function's entire job is an arithmetic guarantee, so it should round the safe way.

**Fix:** `math.ceil(backoff_seconds)`.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -203,17 +232,36 @@ def run_debug(
...
+    # that is smaller.
+    deadline = time.monotonic() + (
+        budget
+        if budget is not None
+        else debug_budget(timeout, retries, backoff_seconds)
+    )
+    out_of_budget = False
+
     for attempt in range(retries):
+        attempt_cap = min(timeout, int(deadline - time.monotonic()))
```
rockymadden, line 279: 🟡 **Minor: nothing floors `attempt_cap` against the CLI minimum.**

A small explicit `budget` inverts the invariant #2776 established:

```
budget=25: subprocess cap=25, cli --timeout=30  <-- inverted
```

The CLI would then be SIGKILLed before it could self-terminate with a parseable envelope, which is the original bug. Unreachable from current call sites since the default budget is `2*timeout+5`, but nothing prevents it.

**Fix:** reject `budget < _MIN_RETRY_BUDGET_SECONDS` up front.

tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py
```diff
@@ -0,0 +1,170 @@
...
+                    int(timeout.group(1)) if timeout else None,
+                )
+
+
+_CASES = sorted(_criteria(), key=lambda c: (c[0], c[2] or ""))
+
+
+def test_the_suite_was_actually_discovered():
+    """A broken walk would make every assertion below vacuously pass."""
+    assert len(_CASES) > 40
```
rockymadden: 🟡 **Minor: this does not guard what it needs to.**

It proves the YAML walk found criteria, but not that any of them resolved a budget. Combined with the silent-skip above, every case could be `None` and this still passes green.

**Fix:** also assert a floor on the number of cases with a non-`None` budget.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -69,14 +69,25 @@
...
+# The smallest remaining budget worth spending on another attempt. Below this a
+# retry cannot even fund the CLI floor plus its headroom, so it would be killed
+# mid-run and yield nothing to diagnose from.
+_MIN_RETRY_BUDGET_SECONDS = _MIN_CLI_TIMEOUT_SECONDS + _CLI_TIMEOUT_HEADROOM_SECONDS
+
+# What a check spends outside `run_debug`: interpreter start, the static
+# `.flow` asserts, and solution teardown. The criterion `timeout:` in the task
+# YAML must clear :func:`debug_budget` by at least this much — see
+# test_criterion_budgets.py, which enforces it across every task.
+_CRITERION_MARGIN_SECONDS = 60
```
rockymadden: 🟡 **Minor: inconsistent visibility.**

`debug_budget` is public but this is private, and `test_criterion_budgets.py` reaches in for `flow_check._CRITERION_MARGIN_SECONDS`. The two are one contract and should share one visibility.

**Fix:** make the margin public alongside `debug_budget`.

tests/tasks/uipath-maestro-flow/_shared/test_flow_check.py
```diff
@@ -793,28 +794,41 @@ def _cp(returncode, stdout="", stderr=""):
...
         calls["timeout"] = kwargs.get("timeout")
+        calls["caps"].append(kwargs.get("timeout"))
+        calls["clock"] += min(attempt_seconds, kwargs.get("timeout"))
         result = queue.pop(0)
         if isinstance(result, BaseException):
             raise result
         return result
 
+    monkeypatch.setattr(flow_check.time, "sleep", fake_sleep)
+    monkeypatch.setattr(flow_check.time, "monotonic", lambda: calls["clock"])
```
rockymadden, line 830: 🟡 **Minor: this patches the global `time` module.**

`flow_check.time` is the stdlib module, so this swaps `time.monotonic` process-wide for the duration of the test. It follows the existing `time.sleep` precedent directly above and monkeypatch restores it, but `monotonic` has more callers than `sleep` does.

Call it out in the docstring at minimum so the next person does not debug a mystery.

rockymadden (commented):
**Self-review, treated as request-changes.** GitHub will not let an author formally request changes on their own PR, so the PR is converted to draft to block the merge until 1-3 are addressed.

Reviewed cold. The `run_debug` change is sound and the diagnosis is well-evidenced. **The guard is not.** It has two holes, and both bite in this PR: three criteria are still under-budgeted after a change whose whole purpose was to stop that. The "this stops the class of bug" claim does not hold yet.

Blocking on 1-3. The minors can follow.

tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py
rockymadden: Fixed in 94f5356. `_entry` now falls back to `main` when the subcommand names no function, so `if/elif` dispatch is priced instead of skipped, and `_cost` picks the branch the criterion actually runs (605s, not both branches at 1210s).

wiki_pageviews raised 600 -> 720. Negative control, putting it back to 600:

```
AssertionError: multi_node/wiki_pageviews/wiki_pageviews.yaml grants 600s to
check_wiki_pageviews_flow.py invalid_error, which can spend 605s in run_debug.
Raise the criterion to >= 665s, or lower the check's timeout / pass retries=1.
```

The fallback also caught `feet_inches` f2i/i2f, which the old resolver skipped for the same reason. Both raised 300 -> 600.

tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py
rockymadden: Fixed in 94f5356. `max()` is gone, replaced by a branch-aware `_cost` walk: straight-line calls sum, `If`/`IfExp`/`Try` arms take the more expensive branch, a loop body counts once (still the documented floor).

decision now prices at 970 and is raised 600 -> 1080. Negative control:

```
AssertionError: single_node/decision/decision.yaml grants 600s to
check_decision_flow.py, which can spend 970s in run_debug. Raise the criterion
to >= 1030s, or lower the check's timeout / pass retries=1.
```

Summing alone would have over-charged wiki_pageviews for both of its mutually exclusive branches, which is why this is a branch walk rather than a plain `sum`.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Fixed in 94f5356, taking the second option as suggested.

`debug_budget` is now sized on `retries`, so it funds every attempt the caller asked for. The default drops to `_DEFAULT_RETRIES = 2`, which keeps every ceiling exactly where it is. `_POLL_TIMEOUT_ATTEMPTS` stays 2 and is now documented as independent: raising `retries` does not buy more poll-timeout attempts.

The defaults are named constants (`_DEFAULT_RETRIES_TIMEOUT` / `_DEFAULT_RETRIES` / `_DEFAULT_BACKOFF_SECONDS`) shared by `run_debug` and `debug_budget`, so the pricing function cannot drift from the function it prices.

Regression lock added: `test_default_budget_funds_every_retry_the_caller_asked_for` asserts `debug_budget(200, retries=3) == 610` and that all three attempts actually run.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Fixed in 94f5356: `math.ceil(backoff_seconds)`. `debug_budget(240, 3, 2.5)` now returns 483 rather than 482.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Fixed in 94f5356. `run_debug` now rejects `budget < _MIN_RETRY_BUDGET_SECONDS` up front, before spawning anything:

```
FAIL: run_debug budget of 25s is below the 90s floor (the CLI timeout minimum
plus its headroom); raise `budget` or `timeout`
```

Covered by `test_budget_below_the_cli_floor_is_rejected`, which also asserts no subprocess was started.

tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py
rockymadden: Fixed in 94f5356. It now asserts both that criteria were discovered and that more than 40 of them resolved a non-`None` budget, so the silent-skip failure mode above cannot keep this green.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
rockymadden: Fixed in 94f5356: renamed to `CRITERION_MARGIN_SECONDS`, public alongside `debug_budget`. The guard imports it normally instead of reaching into a private name.

tests/tasks/uipath-maestro-flow/_shared/test_flow_check.py
rockymadden: Documented in 94f5356 rather than changed. `_stub_debug` now says both clock stubs land on the stdlib `time` module, that monkeypatch restores them, and that no real sleeps should be added under the fake clock.

Left as-is otherwise: injecting a clock would mean threading a parameter through `run_debug` purely for tests, and this matches the `time.sleep` stub that has been in the file since #2776.

rockymadden:
All 8 review threads addressed in 94f5356 and resolved.

**The two criticals were real and both were live in this PR.** The guard was skipping `wiki_pageviews` entirely and under-pricing `decision` by 430s, so the PR shipped the exact bug it claims to fix, hidden by the mechanism meant to catch it. Replacing `max()` and the silent skip with a branch-aware cost walk caught **5** more under-budgeted criteria, including `feet_inches`, which the original resolver never even looked at.

| Newly caught | was | now |
|---|---|---|
| `wiki_pageviews` (x2) | 600 | 720 |
| `decision` | 600 | 1080 |
| `feet_inches` f2i / i2f | 300 | 600 |

**The major** took the second option: `debug_budget` sizes on `retries`, default drops to 2, so no ceiling moved and `retries` no longer lies about what it funds.

Negative controls for both criticals fail with the exact arithmetic, so these are locked, not just fixed.

Marking ready for review. 356 maestro-flow tests, 583 across the rest of `tests/tasks`, casing guard green.

rockymadden:
## Review #2 addressed in 9960dd8

**The critical was real and, again, live in this PR.** The `_cost` model priced loop-driven checks at one pass, so two criteria stayed green that will be SIGKILLed exactly as `skill-flow-api-workflow` was.

| task | iters | true cost | had | now |
|---|---|---|---|---|
| `billing_invoice_lookup` | 3 (static `CASES`) | 1095 | 600 | **1200** |
| `customer_escalation_triage` | 2 (seed.json) | 1210 | 720 | **1320** |

`billing_invoice_lookup` is the damning one: the count is a module-level literal sitting in the AST, and the criterion was in the original audit's OK bucket, so nothing in this PR had ever looked at it. `customer_escalation_triage` is worse in a different way, since 720 was a number **this PR set**, which made it read as reviewed.

**Root cause of the miss:** helper costs were summed separately from the entry point. The escalation checks put `run_debug` inside `verify_case` and the loop inside `main`, so the loop detector saw no debug in the loop at all. Costing helpers **inline at the call site** fixes both the pricing and the detection.

**The rule is now enforced, not disclosed:**
- `for` over a module-level literal (through `enumerate` / `sorted` / `list`) multiplies out
- a runtime-seeded loop must be sized by hand and marked `# budget-guard: manual`, which records that a human did the arithmetic. The one-pass floor still applies on top.

`escalation_orchestrator_paths` keeps 3000 and gains the marker with its 7-case arithmetic. `escalation_slack_alert` and `escalation_jira_ticket` run one debug each, not a loop, and are correctly unaffected.

**Minors 2-6** all done: documented the `retries` 3 → 2 narrowing as a behaviour change rather than a ceiling saving, admitted `CRITERION_MARGIN_SECONDS` is a chosen floor and not a measured one, dropped the `or tree` fallback for a loud failure, handled `ast.Match`, noted the helper-branch over-charge.

**8 new cost-model tests** run against a synthetic module rather than a real task, so they cannot rot when a task is edited. Both negative controls fail with the exact arithmetic.

364 maestro-flow tests, 591 across the rest of `tests/tasks`, casing guard green.

rockymadden:
## Review #3 addressed in bdf180db

Both findings were in the escape hatch review #2 built.

### 1. The marker verified nothing

Confirmed exactly as reported: marker on `customer_escalation_triage` + criterion back to **720** → guard green, true need 1270. And the guard's own failure message told authors to add the marker, so following the printed remediation switched the check off.

The marker now carries its arithmetic and the guard checks it:

```yaml
timeout: 1320  # budget-guard: manual x2. seed.py emits 2 cases x debug_budget(300) = 1210 + margin
```

- bare `manual` → refused, "write `manual xN` so the number can be checked rather than trusted"
- marker on a criterion with no uncountable loop → refused as stale
- **N is cross-checked against the task's `seed.py`** where it states cases literally. Both live cases resolve (2 and 7), so an understated `x1` now fails too. That closes the last trust point rather than just narrowing it.

### 2. Detector and cost model disagreed about scope

Rather than teach `_unsized_loops` to walk the same tree, both now come out of **one traversal** returning `_Price(seconds, unsized)`. They cannot disagree about scope because there is only one scope. Exclusive branches report every arm's guesses, since any arm might be the one that runs.

### Minors 3-5

Timeout and annotation read off the same line so they cannot drift. The non-literal check moved into the traversal, so it follows reachable code and a dead helper no longer fails a task. `_module_lengths` drops any name a function rebinds.

### Nit 6

Filed as #2962. `uipath-maestro-case` is not broken today (no retry, so budget == timeout, and every criterion clears it) but has no guard keeping it that way. Separate PR, different CODEOWNERS.

### Verification

9 more regression tests on synthetic modules. Four negative controls: 720-with-marker fails, bare marker fails, `x1` against a 2-case seed fails, stale marker fails.

373 maestro-flow tests, 600 across `tests/tasks`, casing guard green.

### Worth recording

Four rounds, four live under-budgeted criteria, one root cause each time: **something the guard could not compute was treated as fine.** wiki_pageviews behind an unresolvable subcommand, decision behind `max()`, billing_invoice_lookup behind an uncounted loop, customer_escalation_triage behind an unverified annotation. That sentence is now in the module docstring with the four examples, because the next person to extend this will feel the same pull.

rockymadden:
## Review #4 addressed in 45e2c21

### The framing was wrong, and that is the important finding

`single_node/subflow/test_check_subflow_flow.py` already asserted this relation **for one task**, added after an RCA on **2026-05-29** describing the identical failure:

> The criterion budget (120s) was *below* `flow_check.run_debug`'s own `subprocess.run` timeout (240s default). That inversion means the checker can NEVER surface the underlying CLI error: the sandbox kills the whole process first, yielding exit `-1` with empty output instead of the informative exit `1` + traceback.

Four months later `api_workflow` hit the same wall, because the guard covered **1 task out of 44**.

So this is not a newly discovered class. It is a known one that was never generalized, and the PR description said otherwise. The per-task test is deleted, this module owns the rule suite-wide, and the RCA's own words are quoted in the docstring — the doc it cited (`docs/uipath-skills/2026-05-29-skill-flow-subflow-root-cause.md`) is **no longer in the tree**, so that memory would have been orphaned a second time. PR description corrected.

### Two uncountable loops under-charged

Confirmed: loops running 2x and 7x, declared `x2`, priced **1940** against a real **4365**. The docstring called applying one N to every loop "the safe direction", which only holds when N is the largest count. Refused outright now — one number cannot describe two loops.

### A count nothing corroborates is refused

Both live markers cross-check against their task's `seed.py`, so requiring it costs nothing today and stops the next task inheriting the trust-by-default that the previous three rounds each found a live bug behind.

### Untested branches

Two refusal paths had no live task to exercise them. The decision moved into `_annotation_error`, a pure function, and all **8** paths are table-tested: ordinary criterion, corroborated count, stale marker, two loops, unannotated loop, bare marker, uncorroborated count, understated count. An untested branch is where every earlier round of this guard went wrong.

### Minors

Dropped two dead ternaries — `_merge()` with no arguments already returns `_FREE`, and both call sites built their child list twice to ask a question the function answers.

381 maestro-flow tests, 608 across `tests/tasks`, casing guard green.

rockymadden:
> [!WARNING]
> **Superseded by `ea5b223` (review #6). Two claims below are wrong.**
>
> 1. "Grading counts every command timeout including `pre_run` / `post_run`, since all run inside the watchdog" — **false**. `orchestrator.py` runs `pre_run` at 468 and `post_run` at 579, outside the `ThreadedWatchdog` block at 484-499. Grading now counts `success_criteria` only.
> 2. The relation `task_timeout >= sum(command timeouts) + turn_timeout` was **invented and does not hold** — `smoke.yaml` sets `task_timeout == turn_timeout == 900`, making it impossible for any task with any grading. Replaced with the provable floor `task_timeout > grading`.
>
> Consequently the **16 raises announced below were reverted**. Four `task_timeout` declarations survive, on the tasks whose grading alone exceeded their inherited cap.
>
> The underlying finding — that raised criteria eat into `task_timeout`, and firing it reports the whole task `TIMEOUT` — was real and is still fixed. Only the relation and the scope were wrong.

---

## Review #5 critical addressed in 82b4dc0

**The finding was in the part of this PR nobody had looked at since round one.**

Raising 41 criterion ceilings roughly doubled worst-case grading time, and nothing checked that against `task_timeout`. It is not advisory — `coder_eval/orchestrator.py:475-508` wraps `_evaluation_loop` (agent turns **and** grading) in a `ThreadedWatchdog` that SIGKILLs the agent and reports the task `TIMEOUT`. That loses every criterion including the ones that already passed, which is strictly **less** diagnosable than the single failed criterion this suite's timeouts exist to produce.

Overshooting the fix reproduced the fix's own bug one level up.

### Scope

**6 tasks pushed over by this PR:** `jira_create_issue`, `jira_get_issue`, `customer_escalation_triage`, `cli_dice_roller_simulated`, `billing_invoice_lookup`, `billing_dispute_resolution`.

**10 already were,** including `drive_to_slack` and two `ixp` tasks whose `turn_timeout` alone equals their `task_timeout`.

All 16 raised. Exempting the pre-existing ones while adding a guard would be the same "documented limitation" this PR has spent four rounds removing.

### The relation

```
task_timeout >= sum(command timeouts) + turn_timeout
```

Deliberately weak: real agent time is many turns and unknowable statically, so this cannot prove a task fits, only catch ones that provably cannot.

**The principle is already written down in this repo** — `experiments/activation.yaml` applies it to its own budget, and its comment is now quoted where the floor is defined:

> a `task_timeout` below `max_turns x turn_timeout` silently drops the row from every criterion denominator instead of grading it

Work-loop tasks run `max_turns: 200`, so that product is not usable here; one turn is the floor. Second time this review has found the repo already knew the rule and applied it to exactly one place.

Grading counts every command timeout including `pre_run` / `post_run`, since all run inside the watchdog. The scan is anchored so it cannot mistake `turn_timeout:` for a command budget — pinned by a test.

Negative control: reverting `customer_escalation_triage` to 2400 fails naming the 750s left for a 1200s turn.

452 maestro-flow tests, 679 across `tests/tasks`, casing guard green.

### On finding 2 (splitting the PR)

Not actioned — that is maestro-cli-team's call. The four parts and their sizes are laid out in the review if you want to split; the `run_debug` deadline fix has been correct and unchanged since 5e4e7512 and could land alone.

rockymadden:
## Review #6 criticals addressed in ea5b223

Both findings were in the fix I added last round, and one of them repeated the defect it was fixing.

### 1. The guard skipped every task that inherits its cap

```python
if not task_timeout:
    continue  # no cap declared, nothing to overrun
```

That comment was false. Those tasks inherit **1200s** from `experiments/default.yaml`. 58 were silently exempt, and four had **grading alone exceeding the cap**, so no agent run can fit however fast it is:

| task | grading | inherited cap |
|---|---|---|
| `wiki_pageviews` | 1620 | 1200 |
| `feet_inches` | 1380 | 1200 |
| `rpa` | 1380 | 1200 |
| `decision` | 1260 | 1200 |

This PR raised the criteria on all four. The guard built to catch "criterion raises overrun the task cap" was not looking at the four tasks where this PR most clearly did that.

Defaults are now resolved and every task with `success_criteria` is checked. The discovery test asserts inherited caps are actually being resolved, not just that tasks were found — skipping the inheritors is what made this invisible.

### 2. The relation was invented, and wrong

`task_timeout >= grading + turn_timeout` does not survive a second experiment. `smoke.yaml` sets `task_timeout == turn_timeout == 900`, so **every smoke task with any grading would be flagged impossible** while being fine. 13 flow tasks are smoke-tagged.

Replaced with what the numbers prove:

```
task_timeout > worst-case grading
```

An impossibility check, not a sufficiency one. It flags exactly the four above and nothing else, so **the 16 ceilings raised last round are reverted** — over-raising is not free, a hung task holds a runner for the full ceiling. Net YAML effect of this round: 4 added declarations, 16 reverts.

The four carry their arithmetic inline, sized at grading plus one turn. That margin is a judgment call and is written as one.

### 3. My commit message was wrong about coder_eval

It claimed `pre_run` and `post_run` "run inside the watchdog". They do not — `orchestrator.py` calls them at **468** and **579**, while the `ThreadedWatchdog` wraps only `_evaluation_loop` at **484-499**. Grading now counts `success_criteria` alone, and a test pins the exclusion.

512 maestro-flow tests, 739 across `tests/tasks`, casing guard green.

### On splitting

Re-raising round 5's finding 2, since this round makes the case sharper: two of the last three findings were **self-inflicted, in the guard rather than in the fix**. The `run_debug` deadline change has been correct and untouched since 5e4e7512 and is what actually stops the flake. If maestro-cli-team wants it merged now, `5e4e7512` plus the criterion raises stands alone, and the guard follows separately.

rockymadden:
## Copilot review: 1 of 2 findings valid, fixed in df3d4824e

### ✅ Valid — `retries=0` crashes instead of grading

> `debug_budget()` treats `retries=0` as 1 attempt but `run_debug` currently executes 0 attempts (`range(retries)`), which would leave `r` undefined.

Confirmed:

```
debug_budget(240, retries=0) = 240   (prices 1 attempt)
run_debug    range(0)        = []    (runs 0)
-> UnboundLocalError: cannot access local variable 'r'
```

A checker dying with `UnboundLocalError` instead of failing the criterion it is grading is exactly the outcome this PR exists to prevent. No call site passes `retries=0`, but the two functions disagreeing is the drift the guard is meant to stop, and `max(1, retries)` in `debug_budget` is mine — I created the disagreement.

Fixed by clamping the loop to the same `max(1, retries)`, with a regression test asserting both functions agree.

### ❌ Not valid — `attempt_cap` cannot reach 0, and `timeout=0` is caught

> `attempt_cap = min(timeout, int(deadline - time.monotonic()))` can become 0 and `subprocess.run(timeout=0)` raises `ValueError`.

Two problems with this.

**Unreachable.** The budget floor added for #2776 rejects any `budget < 90s` up front, so the first attempt gets `min(timeout, >= 90)`, and a retry only starts once the remainder clears 90s. Reaching 0 needs `timeout <= 0`, which the floor check already fails cleanly:

```
run_debug(timeout=0)
-> SystemExit: FAIL: run_debug budget of 5s is below the 90s floor ...
```

**Wrong exception.** `subprocess.run(timeout=0)` raises `TimeoutExpired`, not `ValueError`:

```
subprocess.run(["sleep","1"], timeout=0)
-> TimeoutExpired: Command '['sleep', '1']' timed out after 0 seconds
```

`run_debug` already catches `TimeoutExpired` and routes it through `_fail_with_capture`, so even if it were reachable it would be a graded failure with the partial output preserved, not a crash.

No change for that half. 513 maestro-flow tests pass.

tests/tasks/uipath-maestro-flow/_shared/flow_check.py
```diff
@@ -203,17 +238,46 @@ def run_debug(
...
+    # Clamped so `debug_budget`, which prices `max(1, retries)`, cannot promise
+    # an attempt this loop never makes: `range(0)` left `r` unbound and crashed
+    # with UnboundLocalError instead of failing the criterion.
+    attempts = max(1, retries)
+
+    for attempt in range(attempts):
+        attempt_cap = min(timeout, int(deadline - time.monotonic()))
+        cli_timeout = max(
+            _MIN_CLI_TIMEOUT_SECONDS, attempt_cap - _CLI_TIMEOUT_HEADROOM_SECONDS
+        )
```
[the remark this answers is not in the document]
rockymadden: Checked this one carefully. The imprecision is real but bounded at **1 second**, has no consequence, and the proposed remedy would introduce a crash.

**`math.floor` is a no-op here.** `int(x) == math.floor(x)` for the positive case, which is the only case reachable — `int(89.999) == floor(89.999) == 89`.

**The slip is 1s, not unbounded.** Between the pre-sleep check and the next `attempt_cap` there is only `time.sleep(backoff_seconds)` (already subtracted in `left`) plus a few statements. `time.sleep(5)` returns at >= 5s, so the drift is the overshoot: microseconds. Worst case `left == 90.0` exactly gives `attempt_cap == 89`.

**89 is harmless.** `_MIN_CLI_TIMEOUT_SECONDS` floors the CLI timeout at 30, so at `attempt_cap = 89` the CLI gets `--timeout 30` and self-terminates 59s inside our cap. The inversion the floor exists to prevent (`cli_timeout > attempt_cap`) needs `attempt_cap < 30`, i.e. **more than 60s of drift between two adjacent statements**.

**`ValueError` is wrong**, same as the previous review. Both raise `TimeoutExpired`, which `run_debug` already catches and routes through `_fail_with_capture`:

```
subprocess.run(["sleep","1"], timeout=0)   -> TimeoutExpired
subprocess.run(["sleep","1"], timeout=-1)  -> TimeoutExpired
```

**The suggested fix would break attempt 0.** Checking `remaining < _MIN_RETRY_BUDGET_SECONDS` at the top of each attempt fires spuriously when `budget == 90` exactly, since `floor(90 - epsilon) == 89 < 90`. That breaks before `r` is assigned — the `UnboundLocalError` your previous review correctly flagged, reintroduced.

Moving the check after the sleep also overruns the deadline by `backoff_seconds`. Today it refuses *before* sleeping, so the backoff is never spent on a retry we cannot afford; `test_run_debug_never_exceeds_its_budget` pins that (`clock <= 405`, which becomes 410 under the proposed shape).

No change. The invariant that matters — subprocess cap always above the CLI timeout — is held by `_MIN_CLI_TIMEOUT_SECONDS`, not by the retry floor.

rockymadden:
## Second Copilot review: finding examined, no code change

> `run_debug` decides whether a retry is fundable using a float ... but then floors the per-attempt cap with `int(...)`. This can start a retry with `attempt_cap` < `_MIN_RETRY_BUDGET_SECONDS` (or even <= 0) ... risking a `ValueError`.

The imprecision is real. Everything downstream of it is not.

| claim | finding |
|---|---|
| use `math.floor` for consistency | no-op: `int(x) == math.floor(x)` for positive x, the only reachable case |
| cap can fall below the floor | true, by **exactly 1s** (90 → 89) |
| cap can reach <= 0 | needs >60s of drift between two adjacent statements |
| raises `ValueError` | no: `timeout=0` and `timeout=-1` both raise `TimeoutExpired`, already caught |

**Why 1s is the bound.** Between the pre-sleep check and the next `attempt_cap` there is only `time.sleep(backoff_seconds)`, already subtracted in `left`, plus a few statements. `sleep(5)` returns at >= 5s, so the drift is the overshoot.

**Why 89 is harmless.** `_MIN_CLI_TIMEOUT_SECONDS` floors the CLI timeout at 30. At `attempt_cap = 89` the CLI gets `--timeout 30` and self-terminates 59s inside our cap. The inversion the floor exists to prevent needs `attempt_cap < 30`.

**Why the suggested fix is worse.** Checking `remaining < _MIN_RETRY_BUDGET_SECONDS` at the top of every attempt fires spuriously when `budget == 90` exactly — `floor(90 - epsilon) == 89` — breaking before `r` is assigned. That is the `UnboundLocalError` the *previous* Copilot review correctly found, reintroduced. Moving the check after the sleep also overruns the deadline by `backoff_seconds`; `test_run_debug_never_exceeds_its_budget` pins that at `clock <= 405`, which becomes 410.

The invariant that matters — subprocess cap always above the CLI timeout — is held by `_MIN_CLI_TIMEOUT_SECONDS`, not by the retry floor. Thread replied to and resolved.

513 tests pass.

rockymadden:
## Third Copilot review: 4 findings, all valid, fixed in d7b8715

Best of the three reviews. Two are real defects in the scan, one is a hollow test, two are stale comments.

### 1. Criterion scan read the wrong number — two ways

`_TIMEOUT` was **unanchored**, so it matched inside the limit keys:

```
_TIMEOUT.search("  turn_timeout: 900")   -> matched "timeout: 900"
_TIMEOUT.search("  task_timeout: 1200")  -> matched "timeout: 1200"
```

And blocks split on `run_command` rather than on any criterion, so the last block ran through `post_run`. Together: a criterion declaring no timeout could borrow one from a neighbour or from cleanup.

Blocks now split on any criterion and stop at the next top-level **key** (not any column-0 char, since comments live there), both patterns are anchored, and a criterion without a timeout costs coder_eval's 30s default (`models/criteria.py:185`) instead of 0. 33 criteria were being counted as free.

### 2. The task-budget fixtures asserted their own shape, not the rule

Correct, and it is the criticism this guard has attracted six times over — an assertion that cannot fail. `test_impossible_task_is_rejected` ended in:

```python
assert int(_TASK_TIMEOUT.search(text).group(1)) <= grading  # the rejected shape
```

That is a statement about the fixture. Deleting `test_grading_alone_fits_the_task_timeout` would not have turned it red.

Policy extracted into `_task_budget_error`, mirroring the existing `_annotation_error`. The suite-wide test and every fixture now call it. Two more fixtures pin the scan fixes: a criterion without a timeout costs 30, and one cannot take the number from the item below it.

### 3 & 4. Two comments describing values this PR changed

- `cli_dice_roller_simulated.yaml` still explained the old **660s** ceiling under a value of 1320
- `subflow.yaml` still said "Guarded by test_check_subflow_flow.py" — **a file this PR deletes**, and described the old one-attempt relation

Both rewritten. Checked the tree for other references to the deleted file: none.

### Verification

Grading totals move for the 33 criteria that declare no timeout. Every task stays clear of its cap, tightest being `file_attachment` at 810s under 1200s. Both negative controls still fail with the exact arithmetic.

515 maestro-flow tests, 742 across `tests/tasks`.

---

Running tally across the three Copilot reviews: **5 valid findings, 1 not** (the `attempt_cap`/`ValueError` one, answered [here](https://github.com/UiPath/skills/pull/2960#discussion_r3907458824)). The suppressed-comment mode is worth noting — this round it surfaced 4 findings it chose not to post as comments, and all 4 were real.

rockymadden:
## Fourth Copilot review: both findings valid, fixed in 3394e26

### 1. The floor guarded the aggregate, the inversion is per-attempt

Correct, and it is the same defect the *first* Copilot review circled without landing on. I validated `budget` and assumed the per-attempt cap followed. It does not:

```
timeout= 20 retries=10 -> budget=245 passes_floor=True  cap= 20 cli= 30 INVERTED
timeout= 29 retries= 4 -> budget=131 passes_floor=True  cap= 29 cli= 30 INVERTED
```

The subprocess cap lands under the CLI’s own `--timeout` minimum, so we SIGKILL it before it can return an envelope — the exact diagnostic loss #2776 fixed, reintroduced through a floor I wrote to prevent it.

`timeout` is now floored against `_MIN_CLI_TIMEOUT_SECONDS` independently of the total. Both floors stay: one keeps a retry fundable, the other keeps the subprocess outliving the CLI. Boundary now clean at 30/31:

```
timeout= 30 -> rejected   |  timeout= 31 -> cap=31 cli=30, not inverted
```

Not reachable from the suite — the smallest `timeout` any call site passes is 180 — but the check was measuring the wrong quantity.

### 2. Recursion was priced, not refused

Also correct, and sharper than it looks: `test_recursion_does_not_hang` was **asserting the underpricing**. It pinned an unbounded recursive `main()` at one 485s attempt, which is meaningless for a function that never returns, and contradicted the contract that what the guard cannot count it refuses.

A recursive call whose target runs a debug now fails explicitly:

```
line 3: `main` recurses and runs a debug, so its cost cannot be counted.
Flatten the recursion or bound it in a loop this guard can size.
```

Mutual recursion (`a -> b -> a`) is covered, and a recursive helper that runs no debug still costs nothing rather than failing spuriously. Three tests replace the one that encoded the old behaviour.

### Verification

518 maestro-flow tests, 745 across `tests/tasks`, casing guard green. No task changes — nothing in the suite was hitting either path.

---

Tally across four reviews: **7 valid, 1 not**. Both of these were in code I added in response to earlier rounds, which is the honest pattern here — the fixes have been the risk, not the original change.

rockymadden:
## Fifth Copilot review: all 5 valid, fixed in 9e26635

Four of the five are one defect: **the guard priced `run_debug` and charged a flat 60s for everything else.**

### The real cost of the jira checks

`jira_is._run` is `subprocess.run(..., timeout=120)`. `connection_id()` is **two** of those before a single issue read:

| task | debug budget | non-debug | was | now |
|---|---|---|---|---|
| `jira_search_triage` | 1205 | 480 | 1320 | **1800** |
| `jira_lifecycle` | 1205 | 600 | 1320 | **1920** |
| `jira_create_issue` | 965 | 480 | 1080 | **1560** |
| `escalation_jira_ticket` | 480 | 720 | 1600 | 1600 (already above) |

`escalation_jira_ticket` needed 1260 and has 1600. The guard sets a floor; lowering a generous ceiling is not this PR's business.

`CRITERION_MARGIN_SECONDS` was documented as "a CHOSEN floor, not a measured one" — honest, but it was still certifying budgets that cannot hold.

### The fix follows the contract this guard already has

Bounded non-debug work is now **detected** (`subprocess.run` with a literal timeout, in the check or a sibling module it imports) and must be **declared**:

```yaml
timeout: 1800  # budget-guard: overhead 480. connection_id (2) + 2 seeded issue reads, 120s each
```

The count comes from runtime data — candidate keys parsed out of the debug payload — so it cannot be derived. But it cannot be forgotten either: undeclared fails naming the call sites, declared-with-no-such-work fails as stale. Same shape as `manual xN`.

I chose the annotation over teaching the cost model to price arbitrary `subprocess` calls. Pricing `jira_is` internals means resolving cross-module helpers and loops over runtime lists, which is the "cannot count" case again, one layer down.

### 5. Misleading remedy

Correct. The budget-floor message said "raise `budget` or `timeout`" even when an explicit `budget` caused it. Now it names the one that applies:

```
explicit budget -> raise `budget`, or omit it so debug_budget(timeout, ...) derives the deadline
derived budget  -> raise `timeout`
```

### Verification

Three new tests cover detection through sibling imports, that `run_debug` and untimed calls are excluded, and the annotation parse. Both negative controls fire: removing an annotation, and adding a stale one.

521 maestro-flow tests, 748 across `tests/tasks`.

---

Tally across five reviews: **12 valid, 1 not**. This round found the flat margin, which is the assumption the guard was built on and the one I had flagged as unverified without acting on it.

rockymadden:
## Sixth Copilot review: all 3 valid, fixed in b846962

### 1. The `overhead` annotation was declared but not bounded

Correct, and it is the same gap `manual xN` had before the seed cross-check: a number nobody enforces. `_nondebug_calls` finds syntactic call sites and knows nothing about loops, so `check_jira_create_issue.py` probing one issue per candidate key accepted `overhead 480` while a malformed flow emitting fifty keys would spend 6000s.

Fixed at the source rather than in the guard — the loops are now bounded in the checkers, which is what makes the declaration true whatever the agent produced:

```python
MAX_ISSUE_PROBES = 2
cands = list(dict.fromkeys(cands))[:MAX_ISSUE_PROBES]
```

`jira_create_issue` 2, `jira_lifecycle` 3, `escalation_jira_ticket` 2 on both its success and cleanup paths. The guard additionally requires the declared total to be **whole calls at the price the code uses** — `overhead 500` against 120s calls now fails — which catches a number picked out of the air.

The count still comes from the cap rather than the cost model. Pricing arbitrary cross-module subprocess work through loops is the "cannot count" problem one layer down, and I would rather the bound live in the code that has to honour it.

While checking this: `escalation_jira_ticket` was over-declared at 720 because I assumed both jira paths could run. The `except` handler runs *instead of* the success path, so it is 480.

### 2. The two budget floors were conflated

`_MIN_RETRY_BUDGET_SECONDS` is the threshold for funding **another** attempt. Flooring the total against it meant:

```
run_debug(timeout=45, retries=1)  -> budget 45, rejected — but 45 is a supported cap
run_debug(timeout=45, retries=2)  -> budget 95, accepted — but the retry can never fire
```

The total is now floored at the CLI minimum, which is what it actually guards. The retry threshold stays on the retry gate where it belongs.

### 3. My comment double-counted its own arithmetic

`# 1320 funds the full retry budget: 2 x debug_budget(600) + margin` — `debug_budget(600)` is *already* two 600s attempts plus backoff (1205). 2x would be 2470. Corrected.

### Verification

523 maestro-flow tests, 750 across `tests/tasks`. Negative controls: a non-multiple `overhead`, and a criterion dropped below its funded total, both fail with the arithmetic.

---

Tally across six reviews: **15 valid, 1 not**. Every finding since round three has been in machinery added to answer an earlier round. The `run_debug` deadline change itself, `5e4e7512`, still has none.

rockymadden:
## Seventh Copilot review: all 3 valid, fixed in 4315098

### 1. Recursion refusal only inspected the name that closed the cycle

Correct, and my own `test_mutual_recursion_is_refused` happened to put the debug in the leg that works. The inverse:

```python
def a():  b()
def b():  run_debug(timeout=240); a()
def main(): a()
-> priced at 485   # should have been refused
```

The repeated name is `a`, `a` has no debug, so the check passed. `seen` is now an ordered stack rather than a set, so the **cycle** can be inspected: `a -> b -> a`, and any function in it that runs a debug refuses the lot. The inverse case is now a test.

### 2. The validated cap was not the cap the loop computes

Validation used `min(timeout, budget)`; the loop uses `int(deadline - time.monotonic())`. The deadline is read microseconds after it is set, so an **integer budget always floors one second short**:

```
budget=31 -> accepted, real cap 30 vs cli 30  -> no room for the CLI to report
budget=32 -> accepted, real cap 31 vs cli 30  -> fine
```

Validation now uses `min(timeout, budget - 1)`, which is what the loop will produce. Both boundary values are pinned by a test.

### 3. The overhead multiple did not constrain the count

Your example reproduces exactly. Raising `MAX_ISSUE_PROBES` from 3 to 10 left `overhead 600` green while the checker could spend 1440s:

```
jira_lifecycle.yaml: declares 5 calls of 120s but caps its probes at 10,
which it can spend on top of the fixed calls. Raise the overhead with the cap.
```

The declared count is now cross-checked against the checker's own `MAX_*` constant, so the cap and the annotation move together. `jira_search_triage` gained a cap too — it loops over seeded keys, which nothing bounded if the seed grew.

That is the third layer this annotation has needed: declared, then whole-calls, now bound to the code's own limit. Worth saying plainly that it is enforced against the cap, not derived from the call graph.

### Verification

526 maestro-flow tests, 753 across `tests/tasks`. Negative controls for all three, including your cap-raise example verbatim.

---

Tally across seven reviews: **18 valid, 1 not**.

rockymadden:
## Eighth Copilot review: all 3 valid, plus a self-audit — fixed in 2a9b3ee

### The leak was the serious one, and it was mine

Capping `cands` bounded the tenant reads but also truncated **the list teardown works from**. A malformed flow creating more than the cap left the extras in the shared CE project with nothing able to find them. I introduced that two commits ago while fixing a budget problem.

The cap now applies to the probe loop only:
- `jira_lifecycle` records every candidate *before* probing
- `escalation_jira_ticket` derives its attributable node keys from the full list
- its recovery path is **uncapped** — it only runs after the debug already failed, and cleanup correctness beats a bounded probe on a run that is failing anyway

### `run_debug(**opts)` priced at the default

Correct. Keyword splats hide their arguments, so the call is now unpriceable rather than silently priced at 240s.

---

## Then I swept the diff myself instead of waiting for round nine

Four more shapes the scan read as nothing or as one pass. **None live in the suite**, all the same class:

| shape | was | now |
|---|---|---|
| `mod.run_debug(...)` | criterion skipped whole | priced |
| `from flow_check import run_debug as rd` | criterion skipped whole | refused loudly |
| `async def main()` | "no entry point" failure | priced |
| `[run_debug(...) for _ in load()]` | one pass, unreported | sized and reported like `for` |

And one live sizing problem I caused: **`jira_create_issue` had 210s left for the agent.** The overhead raises grew its grading by 660s, and every other task in the suite keeps at least 570. The guard's floor was satisfied — this is sizing, not a rule violation — so `task_timeout` goes 1800 -> 2400, restoring roughly the headroom the task author had chosen.

I also checked and rejected a wider change: applying `grading + turn_timeout` across 29 tasks would have re-imposed the invented relation review #6 removed. The provable floor holds everywhere; only that one outlier needed a judgment call.

### Verification

532 maestro-flow tests, 759 across `tests/tasks`. Sweep categories run clean: no comprehension, alias, async or attribute-style call live in the suite; every criterion sits within its funded need; every task clears its cap.

Tally: **21 valid, 1 not**.

rockymadden:
## Ninth Copilot review: all 4 valid, fixed in 9e7bef7

**Three of the four are regressions I introduced in the last two commits.**

### 1. Recording every candidate was worse than the leak it fixed

`teardown_jira.py` **deletes** what it is given. A `CE-*` key can reach the raw debug envelope through inputs or an unrelated operation, so my "record before probing" change could have deleted an issue this run never created, in a shared project. Recording moves back inside the probe loop, where a tenant read has confirmed the issue.

I traded a leak for a destructive action and did not notice.

### 2. jira_create_issue never got the leak fix

It still truncated the candidate list before verification *and* teardown — while carrying the new comment claiming only probes were capped. I fixed two of three files and updated the comment on all three. The cap now sits where the comment always said.

### 3. escalation's failure path still probed every candidate

Correct, and my reasoning for uncapping it was wrong. Uncapping does not buy cleanup: it moves where the grader kills the check, and **a killed check cleans up nothing**. Both paths bounded.

### 4. The comprehension model was wrong twice

Written in my own sweep last commit, so this one is doubly mine:

```
[run_debug() for a in A for b in B]   priced len(A), not len(A) * len(B)
[x for x in A if run_debug()]         priced ZERO -> criterion skipped entirely
```

Every generator now multiplies and every filter is priced at its own nesting depth. Both shapes are tests.

### Suppressed: the discovery guards proved nothing

Also right. `> 40` criteria and `> 90` tasks against **120** and **126** eligible meant `_criteria()` could drop most of the suite and stay green. They now count eligibility independently and assert equality, so a parser regression that loses one criterion fails.

### Verification

534 maestro-flow tests, 761 across `tests/tasks`.

---

Tally: **26 valid, 1 not**. Worth being plain that the last two rounds have been almost entirely self-inflicted — each fix added machinery I then tested only against the case in front of me. The exact-count discovery assert is the first thing here that would have caught a whole class of that on its own.

rockymadden:
## Deep self-review, done properly — 04217117

Nine rounds happened because I kept testing each fix against the case in front of me. Replaced that with three techniques that don't depend on me thinking of the case.

### 1. Differential testing of the cost model

Generate a program, price it statically, then **execute it** with a counting `run_debug` and compare.

- branch-free programs must agree **exactly** (worst case IS the actual case)
- branching ones must never come in **under** what a run really spends

**800 randomised programs locally: 0 violations.** Nested loops, comprehensions with multiple generators and filters, helper chains, transparent iterators. 39 fixed shapes ship in the suite.

This is what would have caught last round's two comprehension bugs without Copilot.

### 2. Property test on `run_debug`

Random `(timeout, budget, retries, backoff)` against random attempt durations and outcomes, asserting what inspection kept missing:

- the call never outlives its budget
- every attempt gets a positive cap that **strictly** outlives the CLI's own timeout
- nothing escapes as a crash instead of a graded failure

**600 random schedules locally: 0 violations.** 25 seeds ship.

That's the invariant three separate Copilot findings circled (`attempt_cap` drift, the 31s boundary, `retries=0`), now checked mechanically.

### 3. Ground-truthing the YAML scan against PyYAML

All 126 task files, comparing my regex extraction to a real parser. **Found a real hole:**

```yaml
command: 'python3 "$TASK_DIR/check_inline_agent.py" "x/*.json"'
```

`_COMMAND` stopped at the first inner quote, so the criterion looked like it had no script and was **dropped from enforcement entirely** — `smoke/inline_agent_robust` and `evaluate/inline_agent_eval`. Neither runs a debug, so nothing was under-budgeted, but the scan was quietly narrower than it claimed.

Fixed: discovery goes **120 → 122**, and now matches PyYAML exactly on both the criterion set and the grading sum. PyYAML isn't in CI, so the shape ships as a unit test.

### Where that leaves it

600 maestro-flow tests, 827 across `tests/tasks`. The three engines that kept producing findings — the cost model, the deadline arithmetic, the YAML scan — are now checked against an independent reference rather than against my judgement.

Remaining known-and-accepted limits, stated rather than buried: the overhead annotation is *enforced against* a checker's `MAX_*` cap, not derived from its call graph; `CRITERION_MARGIN_SECONDS` is a chosen floor; the task-budget rule is an impossibility check, not a sufficiency one.

rockymadden:
## Tenth Copilot review: all 4 valid, fixed in 0842aeb

Three were one root cause: **capping the candidate list made ordering load-bearing.** A key scraped from the raw debug envelope can precede the one the flow actually returned, so a bounded probe could miss it, fail a correct flow, and leave the created issue unrecorded. Declared outputs are probed before scrapings now, and the caps are raised so a malformed flow has real headroom.

### That made the overhead derivable, so the annotation is gone

Every probe loop is `x[:MAX_ISSUE_PROBES]` and every fixed call is straight-line. The same model that prices the debug now prices the subprocess work — sibling module defs merged, slices by a module constant counted. It reproduces every number the annotations declared (480 / 480 / 600 / 0) and deletes the whole declared-but-unverified surface: the marker, the whole-calls rule, the cap cross-check, and the fixed-call gap this review also flagged.

That is three rounds of findings removed by deriving the value instead of trusting it.

### Comprehension generators

`[... for a in load_a() for b in load_b()]` collapsed to the comprehension's line, so one `manual xN` covered two independent loops. Unsized loops are keyed by **(line, column)** now — both generators share a source line, so line alone could not tell them apart.

### And a NameError this review did not find

While verifying, a new static undefined-name check over all 92 checkers caught `collect_outputs` used but never imported in `check_jira_lifecycle.py` — introduced by my own fix last round. It would only have surfaced hours into a live-tenant run. That check now ships.

### Verification

693 maestro-flow tests, 920 across `tests/tasks`. Differential test re-run at 300 programs, 0 mismatches. YAML scan still matches PyYAML exactly across all 126 files.

---

**Merge readiness, stated plainly:** the `run_debug` change and the criterion raises are statically verified and property-tested. The four `e2e/jira_*` checkers are not — they create and delete issues in a shared project, only execute against a live tenant, and I have introduced and fixed four separate bugs in them over three rounds. I would not merge those unexercised. See the next comment.

rockymadden:
## Reverted the live-tenant work — 2713d7b

The four `e2e/jira_*` checkers and the overhead machinery built for them are gone. They were the only part of this PR that could not be verified before merging, and three rounds of review found four separate bugs in them: unrecorded tickets, keys recorded without an ownership check (which teardown would have **deleted** from the shared `CE` project), a missing import, and a probe order that could fail a correct flow. None of it had ever been executed.

The bounds existed only so the guard could derive a number. That traded correctness in code that creates and deletes real Jira issues for tidier arithmetic. Wrong way round.

**Reverted:** the four checkers to `origin/main` verbatim, their criterion and task timeouts to this PR's pre-overhead values, and `_overhead_seconds` / `_subprocess_price` / `_nondebug_calls` / `_probe_caps` / `_sibling_modules`.

Those four criteria are now funded exactly as `main` funds them today — a pre-existing gap, not a regression. Filed as #2968, which says plainly that it needs a live-tenant run to land.

`_cost` keeps the `price` hook the derivation introduced; it costs nothing and is what would make that follow-up cheap.

## What is left

**Tests only. Zero live-tenant checkers touched.**

| part | effect | verification |
|---|---|---|
| `run_debug` deadline + budget | grading runtime | property test, 600 random schedules; 17 reviews, 0 findings |
| 44 criterion + 4 task_timeout raises | ceilings only | matches PyYAML exactly across 126 files |
| `test_criterion_budgets.py` | CI only | differential-tested against execution, 800 programs |
| deleted `test_check_subflow_flow.py` | none | superseded by the suite-wide guard |

688 maestro-flow tests, 915 across `tests/tasks`.

## Two behavioural changes to accept knowingly

1. `_DEFAULT_RETRIES` 3 → 2. A fast 5xx now gets one retry instead of two, matching the CLI's own "retry once before reporting". Deliberate; `retries=3` still works and is now funded.
2. Worst-case grading per debug criterion goes from one 240s attempt killed at 300s, to 485s of real budget. Ceilings are free on green runs; a genuinely hung debug now takes longer to fail, but fails *diagnosably* instead of being SIGKILLed.

rockymadden:
## Eleventh Copilot review: all 5 valid, fixed in f041683

All five are in the guard, which is test-only — worst case a red CI check, no eval impact. Two of them, though, would have let a criterion be skipped outright.

### Priced a debug at ZERO

```python
case _ if run_debug(...)      # guard never visited -> 0
while run_debug(...): pass    # test charged once, loop not flagged unsized
```

Guards run to choose an arm, so they all execute; they are summed now before the worst body. And `while` is split from `for`: it is never countable, its test runs every iteration, and a priced call in **either** part requires sizing.

### Priced the wrong count

- A slice was read as its upper bound, so `A[1:3]` counted **3** items instead of 2. Only a bare `[:N]` with a non-negative bound is a length now; a lower bound, a step, or a negative bound needs sizing. (`A[:-1]` was already safe — `-1` parses as a `UnaryOp`, not a `Constant` — but the `lower`/`step` half was real.)
- An inner generator's iterable is evaluated once per combination of the generators to its left, so `[x for a in A for x in helper(a)]` under-priced a debug inside `helper` by `len(A)`.

### The name check was scope-blind

Sharpest of the five. It subtracted one module-wide set of bound names, so **a parameter or local anywhere in the file hid a genuine missing import elsewhere** — precisely the `NameError` it was added to catch two commits ago. It walks lexical scopes via `symtable` now and only checks symbols resolving to a module global, so a closure over an enclosing local is not a false positive.

### Verification

Re-ran all three engines after the change, with the inner-generator shape added to the corpus:

```
differential:  300 programs, 0 mismatches
yaml vs PyYAML: 0 mismatches / 126 files
undefined names: 0 across all 92 checkers
```

699 maestro-flow tests, 926 across `tests/tasks`.

---

Worth noting these landed on the guard *after* the risky live-tenant work was reverted, so none of them touch anything that runs against a tenant.

rockymadden:
## Twelfth Copilot review: all 3 valid, fixed in 41027e7

All in the guard. One of them silently **exempted** a criterion rather than under-pricing it, which is the worse failure.

### 1. A literal length stayed trusted through mutation

```python
CASES = [1]
CASES.extend([2, 3])      # also: +=, .append, a second assignment
for c in CASES: run_debug(...)   # priced 1 attempt, runs 3
```

A name is dropped now if anything could change it after the literal: a rebinding in any form, an augmented assignment, or a mutating method call with it as receiver. Untouched literals still count, so nothing real in the suite moved.

### 2. Assignment aliases exempted the criterion entirely

`_reject_aliased_debug` only knew `from ... import run_debug as rd`. But:

```python
rd = run_debug                # or: rd = flow_check.run_debug
rd(timeout=600)
```

`_calls_in` sees no `run_debug` call, `_price_script` returns `None`, and the criterion drops out of the guard **completely** — not under-funded, unenforced. Both forms are refused now.

Rebinding to the *same* name is still allowed, which matters: `check_customer_escalation_triage.py` does `run_debug = _flow_check.run_debug`, and its calls stay visible. Covered both ways by tests.

### 3. `while` condition runs N+1 times

A terminating loop evaluates its condition once more than its body — the check that ends it. Charging N left the criterion short by a whole `debug_budget`. `break`/`return` only lower that, so N+1 stays an upper bound.

### Verification

```
differential:    300 programs, 0 mismatches
yaml vs PyYAML:  0 mismatches / 126 files
undefined names: 0 across all 92 checkers
```

708 maestro-flow tests, 935 across `tests/tasks`.

---

Same note as last round: this is all CI-only static analysis, landing after the live-tenant work was reverted. The failure mode for every one of these is "the guard under-enforces", never "an eval breaks".

PR #2971, opened 2026-09-01
fix(tests): point the three RPA-backed tasks at an offline fixture

bai-uipath:
`skill-flow-rpa` and `skill-case-single-rpa` both drove the ProjectEuler RPA workflow, which scraped projecteuler.net. That host now returns a hard 403 to datacenter IPs while still serving a normal page from a residential address, so every run failed in the browser step. Nothing about the tasks was wrong: the fixture they drove had simply stopped being reachable from the eval tenant.

Both tasks now drive a HelpDeskLookup RPA workflow that resolves a ticket title from a fixed in-process lookup. The argument shape is unchanged, one numeric input and one string output, so the tasks still exercise RPA resource node discovery, registry resolution and node wiring. What goes away is the entire network failure class, and a run that used to need minutes of browser work now finishes in seconds.

The case task also stops reaching into the flow suite's Orchestrator folder and resolves its process from its own `uipath-maestro-case` folder instead.

`skill-case-e2e-expense-runnable` carried the same breakage and moves across too. Its Payment stage holds a required `rpa` task on the graded happy path, so `uip maestro case debug` dispatched a real ProjectEuler job on every run, and that debug criterion is weight 6 of 15. It now binds HelpDeskLookup on ticket 123. The Rejected lane's RPA task moves with it, though that lane stays dormant at runtime.

Both releases are named `HelpDeskLookup` rather than after their inner project. `uip maestro flow registry search` returns only a node type, display name and description: no folder path, no resource key. Seven tenant releases already answered to the display name `RPA Workflow`, so a fixture carrying that name is indistinguishable from the rest, and an agent asked for it binds one of the others at random. A first paired run confirmed exactly that, drawing ProjectEuler's alias and faulting on the same 403 this change exists to remove. ProjectEuler was only ever discoverable because its own release carries a distinctive name.

The fixture itself, its provisioning recipe entries and its readiness pins live in `coder_eval_uipath`; this change is the task side only.

## Verification

Paired codex / `gpt-5.6-luna` run on the eval VM, tempdir driver, Bedrock backend, against this branch at `7a5888af6` with `uip 1.201.0-dev.8272`.

| Task | Status | Score | Duration |
|---|---|---|---|
| `skill-flow-rpa` | SUCCESS | 1.000 | 156.8s |
| `skill-case-single-rpa` | SUCCESS | 1.000 | 194.9s |
| `skill-case-e2e-expense-runnable` | SUCCESS | 1.000 | 442.7s |

3/3 passed, every criterion green, one iteration each.

One environment note for whoever provisions this: renaming an Orchestrator release leaves `~/.uip/case-resources/process-index.json` stale on the eval host, and `uip maestro case registry pull` honors the cache rather than refreshing it. A case task then binds the old release name and the run fails at dispatch with `[170007] Failure to start the Orchestrator job`, which points nowhere near the cache. `uip maestro case registry pull --force` clears it.

### DIFF
--- a/tests/tasks/uipath-maestro-case/e2e_expense_runnable/check_expense_runnable_structure.py
+++ b/tests/tasks/uipath-maestro-case/e2e_expense_runnable/check_expense_runnable_structure.py
@@ -54,7 +54,7 @@ REQUIRED_EXTERNAL_BINDINGS = {
     "Shared/uipath-agents/WeatherAPI.WeatherAPI": "WeatherAPI",
     "Shared/uipath-maestro-flow/CountLetters CodedAgent.CountLetters": "CountLetters",
     "Shared/uipath-agents/ProcurementProcess.ProcurementProcess": "ProcurementProcess",
-    "Shared/uipath-maestro-flow/ProjectEuler RPA.RPA Workflow": "RPA Workflow",
+    "Shared/uipath-maestro-flow/HelpDeskLookup RPA.HelpDeskLookup": "HelpDeskLookup",
     "Shared/uipath-maestro-case/CaseTest.Maestro Case": "Maestro Case",
 }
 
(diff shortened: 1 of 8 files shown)


charlesliu9 (approved):

rockymadden (approved):
Thank you much!

PR #2974, opened 2026-09-02
feat(uipath-maestro-flow): add the conversational-agent plugin — text chat flows

andrewwan-uipath:
Adds a `conversational-agent` plugin to `uipath-maestro-flow`, plus registration rows, an offline eval task, and CODEOWNERS.

Closes [JAR-10152](https://uipath.atlassian.net/browse/JAR-10152).

## What

Teaches agents to build a text chat: `core.trigger.conversation` → wait for message → a conversational agent → back to waiting, with `send-message` and `get-conversation-context` alongside.

**The agent can be any of three flavors, and the plugin says to ask rather than default.** That was deliberate — the first draft was named `inline-conversational-agent` and every routing table pointed at inline before a reader got to a word of prose. It now leads with the choice, because it decides the node type, the ports, and whether anything is scaffolded at all:

| Flavor | Node type | Port | Scaffold |
| --- | --- | --- | --- |
| Inline | `uipath.agent.conversational` | `success` | `agent init --inline-in-flow --conversational` |
| In-solution | `uipath.core.agent.<projectId>` | `output` | none — `registry list --local` |
| Published | `uipath.core.agent.<guid>` | `output` | none — `registry search` |

It also carries the rule the `agent` plugin already states for autonomous agents: **if the user names an existing agent, it is not inline** — search before scaffolding.

## Three traps that pass `flow validate` silently

The plugin exists as much for these as for the happy path. All three were found by building real flows against a tenant, not by reading source:

1. **Binding `context` and `conversationId` and dropping the other three settings keys.** Serialization reads all four runtime fields; validation requires one. The result validates clean and ships an agent with no chat history.
2. **Invented output paths.** `waitForMessage1.output.exchangeId` and the agent's `output.response` do not exist — the real path is `output.conversationContext.latestExchangeId`, and the agent declares no outputs because it streams its reply.
3. **`=js:` string bindings.** The pre-1.3 form. Studio Web renders it as literal text rather than a binding, and nothing warns.

## Registration

`CAPABILITY.md` (node inventory, task row, plugin list), `planning-arch.md` (plugin index, ports, a new "I need to have a text conversation" heuristic), `planning-impl.md`, `brownfield.md`, `shared/node-output-wiring.md`, and `operate` + `diagnose` capability rows for the Studio Web / VS Code debug hand-off.

The `agent` plugin gains a **Conversational Agents** section for the node type it already owns (`uipath.core.agent.{key}`) and points here for the loop. The settings block lives in one place.

## Tests

`tests/tasks/uipath-maestro-flow/conversational/` — one offline task plus `check_conversational_flow.py` with three checks (`settings`, `loop`, `agent-json`).

The checker was verified both ways against real projects, not just written:

| Case | Result |
| --- | --- |
| Real inline build, all three checks | pass |
| Real in-solution build, all three checks | pass |
| `context` + `conversationId` only | **fails** — names the three missing keys |
| `=js:` string binding | **fails** |
| `.exchangeId` instead of `.latestExchangeId` | **fails** |
| Inline agent wired on `output` | **fails** |
| In-solution agent wired on `success` | **fails** |
| `maxIterations: 25` | **fails** |

Writing it surfaced two bugs in my own checker that only running it exposed: `.flow` edges are `sourceNodeId`/`targetNodeId`, not `source`/`target`; and an in-solution node's id is a **solution resource key**, not `agent.json`'s `projectId`, so id-matching could never have worked.

## Verification and limits

- `.maintenance` re-run after merging `main`: 960 links, 613 link texts, 314 anchors, 82 files for orphans, 28 plugin pairs, 4 capability templates, 13351 lines for version drift — all zero violations. Two checks do report failures, both reproducing on `main` and in files this PR does not touch: one bad anchor in `diagnose/troubleshooting-guide.md`, and 81 unrecognised `uip` commands in the `ixp` and `connector-trigger` plugins.
- Every factual claim checked against a live CLI and tenant: node types and ports read from the registry, the settings block from a packed `.nupkg`, the debug hand-off from a real upload.
- **The eval task passes.** `skill-flow-conversational-chat-loop` scored 9/9 criteria (weighted 1.000) in this branch's own smoke run — [run 34296538031](https://github.com/UiPath/skills/actions/runs/34296538031), 417s, one iteration, CLI 1.203.0-dev.8572. That run was green overall at 96.3% (52/54). `expected_turns: 20` is the observed command count rather than an estimate.
- **In-solution and published are documented but ungraded.** The task pins inline; a sandbox has no user to ask.
- I could not read JAR-10152 while writing this (Atlassian MCP timed out repeatedly), so this is built from the Confluence checklist and review conversations. Worth a check that nothing in the ticket is missed.

[JAR-10152]: https://uipath.atlassian.net/browse/JAR-10152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
```diff
@@ -0,0 +1,119 @@
...
+| `uipath.conversational.send-message` | Writes a message the **flow** composes — a greeting, a handoff notice. Not the agent's reply. |
+| `uipath.conversational.get-conversation-context` | Reads recent exchanges without waiting. Rarely needed — wait-for-message already returns the context. |
+
+### Pick the agent flavor before you build
+
+The trigger and message nodes are identical whichever you pick. Only the agent node changes — and **ask the user rather than defaulting**, because the answer decides the node type, the ports, and whether you scaffold anything at all:
+
+| Flavor | Node type | Where the agent lives | Choose it when |
+| --- | --- | --- | --- |
+| **Inline** | `uipath.agent.conversational` | A UUID subdirectory inside this flow project | The agent exists only to serve this chat, or you need [structured outputs](impl.md#structured-outputs) to route on — only inline has them. You scaffold it with `agent init --inline-in-flow`. |
```
JoshParkSJ: does this have to be 
```
agent init --inline-in-flow --conversational
```

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
```diff
@@ -0,0 +1,119 @@
...
+| **In-solution** | `uipath.core.agent.<projectId>` | A sibling project in the same solution | The agent is its own project, versioned separately, maybe reused by other flows in the solution. Discover it with `registry list --local`. |
+| **Published** | `uipath.core.agent.<guid>` | The tenant, already published | The user names an existing agent, or one is already deployed. Discover it with `registry search`. Nothing to scaffold. |
+
+**If the user names an existing agent, it is not inline.** Run `uip maestro flow registry search "<name>" --output json` (and `registry list --local` for solution siblings) before scaffolding anything — the same rule the [agent](../agent/planning.md) plugin states for autonomous agents.
+
+In-solution and published share one node type and one set of ports; inline differs on both (see [Ports](#ports)).
```
JoshParkSJ: **Contradicts your own ports table further down.**

This says the two non-inline options are interchangeable. The table below shows they are not — the published one has an extra error branch that the in-solution one does not. I checked against the live registry and the table is the correct one, so it is this sentence that needs changing.

Minor on its own, but someone who reads the sentence and skips the table will wire up an error path that does not exist.

Either say they differ only in whether there is an error branch, or drop the sentence and let the table speak.

skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -191,7 +197,11 @@ Every edge requires `sourcePort` and `targetPort`.
...
 | `core.logic.mock` | `input` | `output` |
 | `uipath.agent.autonomous` | `input` | `success`, `error`, `tool`, `context`, `escalation` |
+| `uipath.agent.conversational` | `input` | `success`, `escalation`, `context`, `tool` |
 | `uipath.agent.voice` | `input` | `success`, `error`, `tool`, `context`, `escalation` |
+| `uipath.conversational.wait-for-message` | `input` | `output` |
+| `uipath.conversational.send-message` | `input` | `output` |
+| `uipath.conversational.get-conversation-context` | `input` | `output` |
 | `uipath.conversational.voice.create-outgoing-call` | `input` | `success`, `error` |
 | `uipath.conversational.voice.end-call` | `input` | `success`, `error` |
 | `uipath.core.agent.*` | `input` | `output`, `error` |
```
JoshParkSJ: **This row is now wrong, and it is the one an agent reads when it is *not* in your plugin.**

The row predates your PR, but your PR is what makes it contradictory. It claims every agent node of this kind has an error branch. The ports table you added in the plugin says otherwise, and I confirmed against the registry that yours is right — in-solution agents have no error branch.

Why it matters: this file is the general reference. Someone who never opens the plugin will trust this row, try to connect that branch, and hit a hard failure.

Either qualify it (error branch on published agents only) or add a row next to the conversational one you already added a few lines up.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
andrewwan-uipath: good catch

CODEOWNERS
```diff
@@ -144,6 +144,10 @@
 /skills/uipath-maestro-flow/references/author/plugins/ixp/ @UiPath/communications-mining @constantinmuraru @vlad-crisan @UiPath/maestro-cli-team
 /tests/tasks/uipath-maestro-flow/ixp/ @UiPath/communications-mining @constantinmuraru @vlad-crisan @UiPath/maestro-cli-team
 
+# Chat — inline conversational agent node + the conversation trigger and message nodes
+/skills/uipath-maestro-flow/references/author/plugins/conversational-agent/ @UiPath/jarvis @joshparksj @maxduu @scottcmg @andrewwan-uipath @norman-le
+/tests/tasks/uipath-maestro-flow/conversational/ @UiPath/jarvis @joshparksj @maxduu @scottcmg @andrewwan-uipath @norman-le
```
maxduu: Nice thanks - doesn't @ uipath/jarvis already contain all of us though?

tests/tasks/uipath-agents/lowcode/conversational/scaffold/check_conversational_scaffold.py
```diff
@@ -52,13 +52,17 @@ def assert_conversational_essentials(agent: dict) -> None:
...
-            f'(CLI omits it; conversational has no iteration cap), '
-            f'got {settings.get("maxIterations")!r}'
+            f"FAIL: settings.maxIterations must be 1-8 for conversational, "
+            f"got {iterations!r}"
         )
-    print("OK: settings.maxIterations absent")
+    print(f"OK: settings.maxIterations == {iterations}")
 
     output_props = (agent.get("outputSchema") or {}).get("properties") or {}
     if output_props:
```
maxduu: I think we're good to remove these changes now that Andrei Balas made this fix. Could you maybe rebase your branch? https://github.com/UiPath/skills/pull/3067/changes#diff-ba50f5d96543e61937294b61e55d84a5516bf6fc5b17b87ec35c65ca42a926cf

CODEOWNERS
andrewwan-uipath: oh right, made the change for voice as well

skills/uipath-maestro-flow/references/author/brownfield.md
```diff
@@ -43,6 +43,7 @@ For edits touching multiple top-level arrays, follow [parallel same-file Edit ru
...
 | **Add a resource node** | Discover through the registry (`--local` for in-solution, or tenant registry for published); add with `Edit`; wire edges. Use the relevant plugin's `impl.md` and [editing-operations-json.md](editing-operations-json.md). |
 | **Add an inline agent node** | Embed `uipath.agent.autonomous` with an inline agent definition in the flow project. See [inline-agent/planning.md](plugins/inline-agent/planning.md) for inline versus published selection and [inline-agent/impl.md](plugins/inline-agent/impl.md) for scaffolding, JSON, and validation. |
+| **Add chat nodes** | Turn a flow into a text conversation: a chat agent looping with wait-for-message, started by `core.trigger.conversation` — the trigger is what marks the package conversational. Settle inline versus in-solution versus published first. See [conversational-agent/planning.md](plugins/conversational-agent/planning.md) for the rules and when chat beats voice, and [conversational-agent/impl.md](plugins/conversational-agent/impl.md) for node JSON and the `conversationalAgentSettings` wiring. |
```
maxduu: Hm do you think "Settle inline versus in-solution versus published first" is worth mentioning? Curious why this is a top thing to settle given that is on the "conversational-agent node" layer rather than the "chat nodes as a whole"

skills/uipath-agents/references/lowcode/agent-definition.md
```diff
@@ -219,7 +220,7 @@ Runtime note: attachments cannot be supplied via `uip` CLI. Test from Studio Web
 | `storageVersion` | Managed by `uip agent refresh` — do not edit |
 | `isConversational` | `false` for autonomous agents, `true` for conversational agents. Do not edit. |
 | `showProjectCreationExperience` | `false` |
-| `targetRuntime` | `"pythonAgent"` for autonomous. **`null` for conversational** — conversational agents are not yet in PROD, so the field is intentionally left null until the runtime value is finalized. |
+| `targetRuntime` | `"pythonAgent"` for autonomous and conversational. |
```
maxduu: Nice thanks

maxduu (approved):

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -0,0 +1,287 @@
+# Chat (Text-based Conversation) Nodes — Implementation
+
+Wire whatever shape the conversation needs, inside the rules in [planning.md](planning.md#critical-rules-any-conversational-flow-must-follow).
+
+For a phone conversation, use [inline-voice-agent/impl.md](../inline-voice-agent/impl.md) instead.
```
JoshParkSJ, line 5: nice thanks for this!

JoshParkSJ (approved):

CODEOWNERS
```diff
@@ -145,9 +145,13 @@
 /skills/uipath-maestro-flow/references/author/plugins/ixp/ @UiPath/communications-mining @constantinmuraru @vlad-crisan @UiPath/maestro-cli-team
 /tests/tasks/uipath-maestro-flow/ixp/ @UiPath/communications-mining @constantinmuraru @vlad-crisan @UiPath/maestro-cli-team
 
+# Chat — inline conversational agent node + the conversation trigger and message nodes
+/skills/uipath-maestro-flow/references/author/plugins/conversational-agent/ @UiPath/jarvis
```
rockymadden: Can you add the maestro CLI team to these too please (all of the folders below)

tests/tasks/uipath-maestro-flow/conversational/conversational_chat_loop.yaml
```diff
@@ -0,0 +1,127 @@
...
+
+run_limits:
+  expected_turns: 30
+  max_turns: 80
+  # Same 900 both default.yaml and smoke.yaml set. This build is shorter than
+  # the voice topologies — three nodes and one loop edge — but the settings
+  # block is hand-authored JSON, which is where the turns go.
+  turn_timeout: 900
+
+initial_prompt: |
```
rockymadden, line 26: 🔴 **Critical — missing the canonical headless preamble. CI is red on this.**

`test_every_zero_shot_task_states_the_run_is_headless` fails: every zero-shot flow task must carry the exact paragraph in `tests/tasks/uipath-maestro-flow/_shared/test_headless_preamble.py:22`, verbatim (a second test asserts the wording is identical everywhere).

Paste it into `initial_prompt`.

tests/tasks/uipath-maestro-flow/conversational/conversational_chat_loop.yaml
```diff
@@ -0,0 +1,127 @@
...
+  user: it starts when a conversation begins, waits for the user's message, and
+  has an AI agent answer. After answering it must go back to waiting, so the
+  user can keep the conversation going.
+
+  Build the agent inside the flow project rather than as a separate project, and
+  give it a system prompt telling it to act as a friendly support assistant and
+  keep answers short.
+
+  Registry reads are fine, but do NOT pack, deploy, or upload, and do not run
+  `uip maestro flow debug`.
```
rockymadden: 🔴 **Critical — this phrase fails `test_forbidden_prompt_phrases_match_the_temporary_allowlist`, and it hands the agent a graded answer.**

The prompt matches `do not [^.]*\bdebug\b`, so the task lands in the offenders set. Fix by deleting the sentence, not by adding the task to the allowlist — that list is explicitly shrinking.

🟠 It also defeats the `command_not_executed` criterion below (weight 2.0). That criterion is meant to test whether the *skill* teaches that `flow debug` hands off; telling the agent in the prompt grades obedience instead. The sibling voice task says so outright:

> `voice/voice_inbound_call.yaml:20` — "The prompt deliberately does not SAY 'don't run flow debug' — the whole question is whether the agent knows on its own that debug is not the inbound test loop."

Drop the sentence, keep the criterion. One edit fixes the test and the grading.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,522 @@
...
+    "you are a chatbot.",
+    "system prompt",
+    "todo",
+}
+# Only to catch one-word junk; the filler set above does the real work.
+MIN_PROMPT_LEN = 15
+
+# Generated/staging trees the CLI writes beside the sources. Same exclusion set
+# the other maestro-flow checkers use.
+EXCLUDED_PARTS = {".cli-stage", ".v1stage", ".agent-builder", "_outputs", "v1stage"}
```
rockymadden: 🟡 The comment above this says "Same exclusion set the other maestro-flow checkers use" — it isn't. `voice/check_voice_flow.py:60` also excludes `node_modules`. Either add it or drop the parity claim.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,522 @@
...
+            problems.append(
+                f"{node_id}: settings read from more than one source {sorted(roots)} "
+                "— every key derives from the same wait node's context"
+            )
+        wait_ids = {n.get("id") for n in _nodes_of(flow, WAIT_FOR_MESSAGE)}
+        for root in roots:
+            root_id = _root_node_id(root)
+            if root_id not in wait_ids:
+                problems.append(
+                    f"{node_id}: settings are rooted at {root_id!r}, which is not a "
```
rockymadden: 🟡 **The grader rejects a topology your own planning.md calls legal.**

Verified against a synthetic build:

```
FAIL: conversationalAgent1: settings are rooted at 'getCtx1', which is not a
uipath.conversational.wait-for-message node in this flow
```

But `planning.md § Get Conversation Context` says the node "may be useful for cases when conversational agent and send-message nodes are chained together (to re-obtain the chat-history between them)" — i.e. feeding an agent from a `get-conversation-context` output is documented as legitimate.

Accept a `get-conversation-context` root as well, or drop the sentence from planning.md.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,522 @@
...
+    if metadata.get("isConversational") is not True:
+        problems.append(
+            f"{label}: metadata.isConversational is "
+            f"{metadata.get('isConversational')!r}, expected true"
+        )
+    iterations = settings.get("maxIterations")
+    if (
+        isinstance(iterations, bool)
+        or not isinstance(iterations, int)
+        or not 1 <= iterations <= 8
```
rockymadden: 🟠 **Major — this fails when the field is absent, and the agents-side checker deliberately refuses to grade the same field.**

`tests/tasks/uipath-agents/lowcode/conversational/scaffold/check_conversational_scaffold.py:58`:

> "`settings.maxIterations` is NOT graded. The released CLI (1.202.0) omits it for conversational scaffolds; the `dev` line … emits `maxIterations: 8`. The field is written by `uip agent init --conversational` itself, not by the agent under test, so asserting either way grades the CLI build, not the skill."

`impl.md` only says *"keep what the scaffold wrote"*, so on a runner whose CLI omits it, a perfectly-followed build scores 0 on a 1.5-weight criterion:

```
FAIL: …/agent.json: settings.maxIterations is None; conversational expects 1-8
```

Either grade it only when present (matching the agents-side treatment), or make the doc say "write `8` if the scaffold didn't".

FWIW the 1–8 range itself is right — I read it off the node input schema: `maxIterations: {"type":"number","minimum":1,"maximum":8}`.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,522 @@
...
+            if os.path.basename(os.path.dirname(path)) == source
+        ]
+        if not matches:
+            problems.append(f"no agent.json directory named {source} under {root}")
+            continue
+        for path in matches:
+            if _grade_agent_json(path, problems):
+                graded.add(path)
+
+    if any(not _is_inline(n) for n in agents) and not graded:
```
rockymadden: 🟡 **Mixed-flavor hole.** `and not graded` means that when the flow has an inline agent *and* an in-solution one, the in-solution `agent.json` is never graded — the inline pass already populated `graded`.

Verified: a sibling project with `engine: basic-v2`, `isConversational: false`, prompt `"todo"` passes cleanly alongside a good inline agent.

Grade project-backed agents whenever one is present, not only when nothing else was graded.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -0,0 +1,287 @@
...
+
+The conversational agent node requires a login; the trigger and the tool nodes resolve from the bundled catalog.
+
+## The `conversationalAgentSettings` Wiring Rule
+
+This is the one that goes wrong silently. The agent reads the conversation through `inputs.conversationalAgentSettings`, which holds **five** keys — a `context` binding plus four fields derived from it:
+
+```json
+"conversationalAgentSettings": {
+  "mode": "simple",
```
rockymadden, line 85: 🟡 `"mode": "simple"` shows up in every settings example here with no explanation, and the checker doesn't grade it — yet `mode` is what selects the binding contract.

`preview/skills/uipath-maestro-flow/references/conversational.md:56` documents it: `simple` (default) takes one `context` binding and the platform derives the rest; `custom` binds each field yourself and requires `conversationId`. That is directly relevant to a section whose whole point is "write all five".

Say what `mode` does, or drop it from the example.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -0,0 +1,287 @@
...
+## Node JSON
+
+Editing the `.flow` directly carries the usual obligations — chiefly a `variables.nodes[]` entry for every data-producing node, which is what makes `$vars.<id>.output` resolve at all. See [editing-operations-json.md](../../editing-operations-json.md). `node add` writes those entries for you.
+
+### Conversation trigger
+
+Replace the default manual trigger — `flow init` scaffolds `core.trigger.manual`, and the conversation trigger is what makes the packaged flow conversational and exposed as a chat experience.
+
+```bash
+uip maestro flow node delete ChatFlow/ChatFlow.flow start
```
rockymadden: 🟡 The skill's canonical spelling is `node remove` (`editing-operations-cli.md § Remove a node`), and there's already a "Replace manual trigger with connector trigger" recipe at `editing-operations-cli.md:155` that this bypasses. `node delete` is a real alias so nothing breaks — but consistency is free here.

Same for `-i` vs the documented `--input`.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -0,0 +1,287 @@
...
+  }
+}
+```
+
+### The agent — inline
+
+`inputs.source` is the scaffolded UUID; `conversationalAgentSettings` is the five-key block above.
+
+```bash
+uip maestro flow node add ChatFlow/ChatFlow.flow uipath.agent.conversational \
```
rockymadden, line 153: 🟠 **Major — this contradicts the inline-agent plugin for the same class of node.**

`inline-agent/impl.md:521`: *"Do not use Flow CLI `node add`, `edge add`, or `variable` commands for inline-agent graph edits — inline-agent node, edge, variable, layout, and tool-resource node changes are non-carve-out structural `.flow` mutations and must be authored directly with `Edit` / `Write`."*

I verified your version works: `--source` is a real, documented `node add` flag ("Use the agent's ProjectId (inline agent nodes)"), and I reproduced this whole recipe end to end on CLI 1.203.0 — `node delete start` → trigger → wait → agent w/ `--source` → three edges → `flow validate` returns `Status: Valid`. So the CLI path is current and the inline-agent rule is stale for this node.

The problem is that an agent reads both files. Pick which surface wins for an inline conversational agent and add a line to the other saying so.

skills/uipath-agents/references/lowcode/agent-definition.md
```diff
@@ -85,12 +85,13 @@ Primary configuration file for conversational agent. Edit directly.
 
 ```json
 {
-  "version": "1.1.0",
+  "version": "1.2.0",
```
rockymadden, line 88: 🟠 **Major — the version bump is half done, so the file now contradicts itself.**

You bumped the conversational example to `1.2.0`, but the same file still says:
- `:27` — the autonomous example: `"version": "1.1.0"`
- `:210` — `| version | "1.1.0" — always scaffolded at this version |`

I checked against CLI 1.203.0: **both** scaffolds now write `1.2.0` (standalone conversational, standalone autonomous, and inline-in-flow conversational). Fix all three places in one pass.

Related: `tests/tasks/uipath-agents/lowcode/conversational/scaffold/scaffold.yaml:8` still describes "no maxIterations, no targetRuntime" as the PROD-canonical shape. Your correction here contradicts it.

skills/uipath-agents/references/lowcode/agent-definition.md
rockymadden: 🟡 Verified correct for a standalone conversational agent — `uip agent init --conversational` writes `"targetRuntime": "pythonAgent"`.

But `uip agent init --inline-in-flow --conversational` writes **no** `targetRuntime` at all (I ran both), which matches `capabilities/inline-in-flow/inline-in-flow.md:48` — "No `metadata.targetRuntime`". The flat wording loses that carve-out. Worth a pointer, since this plugin's main flavor is the inline one.

CODEOWNERS
```diff
@@ -145,9 +145,13 @@
...
+/tests/tasks/uipath-maestro-flow/conversational/ @UiPath/jarvis
+
 # Voice — inline voice agent node + the trigger, dial, and end-call nodes
-/skills/uipath-maestro-flow/references/author/plugins/inline-voice-agent/ @UiPath/jarvis @joshparksj @maxduu @scottcmg @andrewwan-uipath @norman-le
-/tests/tasks/uipath-maestro-flow/voice/ @UiPath/jarvis @joshparksj @maxduu @scottcmg @andrewwan-uipath @norman-le
+/skills/uipath-maestro-flow/references/author/plugins/inline-voice-agent/ @UiPath/jarvis
```
rockymadden: 🟡 Dropping `@joshparksj @maxduu @scottcmg @andrewwan-uipath @norman-le` from the voice plugin is unrelated to adding a chat plugin, and it changes who must approve future voice changes.

Everyone removed appears to have signed off here, so it's not a blocker — but it belongs in its own PR.

rockymadden (changes requested):
## 🎯 Summary

Adds a `conversational-agent` plugin for text-chat flows (`core.trigger.conversation` → `wait-for-message` → chat agent → back), the registration rows across the skill, corrections to `uipath-agents/agent-definition.md`, and an eval task plus a 522-line structural checker.

## ✅ What I verified against a live CLI (1.203.0) and tenant registry

I rebuilt the documented recipe end to end and ran the checker against the result.

| Claim | Result |
| --- | --- |
| Node types, ports (`success`/`escalation`/`context`/`tool`; no `error`, no `memory`), `output.conversationId`, `conversationContext.latestExchangeId`, `uipath__agent_response_messages` | ✅ all match `registry get` |
| `send-message` requires all five inputs; `role`/`mimeType` single-valued | ✅ |
| Full impl.md recipe (`node delete start` → trigger → wait → agent w/ `--source` → 3 edges) | ✅ `Status: Valid` |
| `edge add` without `--source-port success` fails | ✅ |
| **The headline trap** — bind `context` + `conversationId`, drop the other three | ✅ still validates clean; the premise is real |
| Checker on a real build | ✅ all three checks pass, and it fails correctly on each documented mistake |
| `agent init --conversational` writes `1.2.0` / `maxIterations: 8` / `targetRuntime: pythonAgent` | ✅ confirms the `uipath-agents` fixes |
| Link checker | ✅ 6754 links resolve |

Unusually well-evidenced work. Nearly every problem below is in the graded task, not the docs.

## 🚩 Findings

Details are inline. Ranked:

**🔴 Critical**
1. **CI is red** — `maestro-flow checker unit tests` fails two corpus tests on the new task file: missing headless preamble, and a forbidden `do not … debug` prompt phrase.

**🟠 Major**
2. The `do not run uip maestro flow debug` sentence makes a 2.0-weight `command_not_executed` criterion free — the voice task deliberately does the opposite.
3. `agent-json` fails when `settings.maxIterations` is **absent**, while the agents-side checker explicitly refuses to grade that field because the released CLI omits it.
4. The `1.1.0` → `1.2.0` bump is half done: the autonomous example and the "always scaffolded at this version" row are now wrong (both scaffolds write `1.2.0`).
5. `impl.md` uses `node add` / `edge add` for the inline agent; `inline-agent/impl.md:521` forbids exactly that. Your version is the current one — but resolve the conflict, an agent reads both.

**🟡 Minor**
6. The checker rejects settings rooted at `get-conversation-context`, which planning.md calls legal.
7. Mixed-flavor hole: an in-solution agent alongside an inline one is never graded.
8. `EXCLUDED_PARTS` parity comment is inaccurate (`node_modules` missing).
9. `node delete` / `-i` vs the skill's `node remove` / `--input`.
10. `mode: "simple"` unexplained everywhere it appears.
11. CODEOWNERS scope creep on the voice plugin.
12. `targetRuntime` wording loses the inline carve-out.

## 🧹 Slop check

Mostly clean — the prose is specific and earns its length. Two things:

- The five-key rule is stated in **five** places (planning.md, impl.md, node-output-wiring.md, "What NOT to Do", the task description). Two is enough; the rest should link.
- The two new `node-output-wiring.md` rows are paragraphs inside table cells, longer than the voice row beside them. Compress to the rule plus a link.

## 📌 Also worth knowing

- Smoke sampled 14 tasks this run and **did not** include `skill-flow-conversational-chat-loop`, so it is still ungraded in CI — consistent with your own note.
- The smoke sandbox **is** authenticated (mounted `.auth`, alpha tenant), so the login requirement is not a smoke-eligibility problem.
- `#3827` resolves to neither an issue nor a PR in `UiPath/skills` — worth correcting the reference so the dependency is trackable.

## 🧭 Recommendation

**Request changes.** CI is red on two tests this PR's own task file causes, and two graded criteria don't grade what they claim. Every one of those is a small edit. The documentation itself checks out against a live CLI, so once the task file and the version rows are fixed this is an easy approve.

---

**tl;dr** — Good plugin; I reproduced the docs against a real CLI and the "silent trap" it is built around is genuine. Blockers are in the test task, not the docs: two CI tests fail, the prompt gives away a graded answer, the `maxIterations` check contradicts the agents-side checker, and the `1.1.0` → `1.2.0` bump is only one-third done.

CODEOWNERS
andrewwan-uipath: the people removed are all under jarvis so this doesn't affect anything

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Pick the agent flavor before you build
+
+The trigger and message nodes are identical whichever you pick. Only the agent node changes — and **ask the user rather than defaulting**, because the answer decides the node type, the ports, and whether you scaffold anything at all:
+
+| Flavor | Node type | Where the agent lives | Choose it when |
+| --- | --- | --- | --- |
+| **Inline** | `uipath.agent.conversational` | A UUID subdirectory inside this flow project | The agent exists only as part of this conversational flow, or you need [structured outputs](impl.md#structured-outputs) to route on — only inline has them. You scaffold it with `agent init --inline-in-flow --conversational`. |
+| **In-solution** | `uipath.core.agent.<projectId>` | A sibling project in the same solution | The agent is its own project, versioned separately, maybe reused by other flows in the solution. Discover it with `registry list --local`. |
```
rockymadden: 🟠 **Major — this node type does not resolve. Verified on CLI 1.203.0.**

Three distinct UUIDs exist for one in-solution agent project, and only the third one is the node-type suffix:

| Source | Value |
| --- | --- |
| `agent.json` `projectId` | `991fc386-5784-493b-86e1-a9509d0161bd` |
| solution manifest project `Id` | `35cadd9b-b9de-4aca-8f8b-095157c25eec` |
| `resources/solution_folder/process/agent/ChatAgent.json` → `key` | `685da0af-cef8-4cf9-b348-8aa089c04cfb` |

```
$ uip maestro flow registry list --local --output json
  "NodeType": "uipath.core.agent.685da0af-cef8-4cf9-b348-8aa089c04cfb"

$ uip maestro flow registry get uipath.core.agent.991fc386-… --local
  "Result": "Failure"
  "Message": "Node not found"

$ uip maestro flow registry get uipath.core.agent.685da0af-… --local
  "Result": "Success"
```

Everything else in this PR already agrees it is the resource key — `planning-arch.md:145,336`, this file's own Planning Annotation section (`uipath.core.agent.{key}`), `agent/impl.md:7,92,109,112`, `check_conversational_flow.py`'s `check_agent_json` docstring, and the PR body:

> an in-solution node's id is a **solution resource key**, not `agent.json`'s `projectId`, so id-matching could never have worked.

This is the flavor table, so it is the first thing a planning agent reads. Use `<resourceKey>`.

Same substitution needed at `conversational-agent/impl.md:164` and `agent/impl.md:199`.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -0,0 +1,291 @@
...
+  "mode": "simple",
+  "context":        { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext",                  "fieldType": "object" },
+  "conversationId": { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.conversationId",   "fieldType": "string" },
+  "exchangeId":     { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.latestExchangeId", "fieldType": "string" },
+  "messages":       { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.messages",         "fieldType": "array"  },
+  "userSettings":   { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.userSettings",     "fieldType": "object" }
+}
+```
+
+**`mode` selects the contract.** `simple` (the default) takes the single `context` binding and lets the platform derive the rest; `custom` means you bind each field yourself and `conversationId` becomes mandatory. The examples here are all `simple`.
```
rockymadden: 🟠 **Major — this contradicts the paragraph two lines below, and it contradicts it in the direction that breaks flows.**

- here: `simple` "lets **the platform derive the rest**"
- `:96`: "Write all five … that derivation only runs in the editor — **nothing derives them** when the file is authored from the CLI"

An agent reading both takes the first: `mode: "simple"` is in every example in this file, so "simple means the platform derives the rest" reads as permission to write one key. That is the exact failure this plugin exists to prevent.

The source says it more carefully — `preview/skills/uipath-maestro-flow/references/conversational.md:57-66`:

> `'simple'` mode (the default) takes ONE binding, `context`, and the platform derives the conversation id, exchange id, transcript and user settings from it. […] Either way the four derived fields are what the runtime reads … **so the SDK emits them from your `context` binding rather than leaving them out.**

So the deriving is done by the *authoring surface* — Studio Web's panel, or the SDK at compile time. Never at run time. "Platform" in that sentence is loose, and this file dropped the correcting half.

Suggested rewrite:

> **`mode` selects the contract.** `simple` (the default) is authored from a single `context` binding — Studio Web's panel and the SDK both expand it into the other four at author time. Nothing expands it at run time, so a hand-authored `.flow` writes all five whichever mode it declares. `custom` is for a turn assembled from more than one source, and makes `conversationId` mandatory.

Two smaller things while you are here: `mode` is optional (I removed it and `flow validate` still returned `Status: Valid`), and nothing grades it.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,549 @@
...
+        ]
+        if not matches:
+            problems.append(f"no agent.json directory named {source} under {root}")
+            continue
+        for path in matches:
+            if _grade_agent_json(path, problems):
+                graded.add(path)
+
+    if any(not _is_inline(n) for n in agents):
+        # Project-backed agents: grade every conversational agent.json present.
```
rockymadden: 🟡 **The mixed-flavor hole is narrower now, but `engine` is still unenforceable on this path.**

The candidate list below is filtered on `engine == "conversational-v1"`, and `_grade_agent_json` then asserts that same field — so the assertion can never fail for a project-backed agent. A misconfigured sibling is dropped from the set rather than failed.

Verified: good inline agent + an in-solution node whose sibling `agent.json` has `engine: "basic-v2"`, `isConversational: false`, prompt `"todo"`:

```
OK: 1 conversational agent.json file(s) configured
exit=0
```

The rest of your fix does work — with `engine` correct, `isConversational: false` and a `"todo"` prompt both fail properly, and the pure in-solution case fails on the empty candidate list. Only this one combination slips.

Cheapest close: compare `len(graded)` against the number of agent nodes and fail on a shortfall, so a filtered-out sibling surfaces instead of vanishing. The `OK:` line already prints the count that would give it away — it says 1 while the flow has 2 agent nodes.

Low priority: the task pins inline, and in-solution is documented-but-ungraded by design.

tests/tasks/uipath-agents/lowcode/conversational/scaffold/scaffold.yaml
```diff
@@ -5,7 +5,7 @@ description: >
   Tests whether the skill teaches the 7-field Init Defaults to Fix
   workflow and the PROD-canonical conversational agent.json shape
   (anthropic.claude-sonnet-4-5, empty inputSchema, empty outputSchema,
-  static user-message placeholder, no maxIterations, no targetRuntime).
+  static user-message placeholder, maxIterations 8, targetRuntime pythonAgent).
```
rockymadden: 🟡 **This is now right, and it contradicts the checker sitting next to it.**

I verified on released CLI 1.203.0 — a standalone conversational scaffold writes `"maxIterations": 8` and `"targetRuntime": "pythonAgent"`, so this description is accurate.

But `check_conversational_scaffold.py` (untouched by this PR) still says the opposite:

> `:18` — "released and dev CLI builds disagree on whether the conversational scaffold carries it"
> `:58` — "The released CLI (1.202.0) omits it for conversational scaffolds; the `dev` line … emits `maxIterations: 8`"

That was true at 1.202.0 and is not any more. Update the comment in the same pass, or the next reader believes whichever half they open first. Keeping the field ungraded is still correct — it comes from `agent init`, not the agent under test.

tests/tasks/uipath-maestro-flow/conversational/conversational_chat_loop.yaml
```diff
@@ -0,0 +1,138 @@
...
+  trigger, which is the only thing that marks the packed flow conversational —
+  an agent on a manual trigger packs clean and is never listed. Also catches
+  the `=js:` string form, which renders as literal text rather than a binding.
+  Needs a login: flow-core emits inline agent nodes only for an authenticated
+  client, so `uipath.agent.conversational` cannot be resolved — and therefore
+  the flow cannot be validated — without one.
+tags: [uipath-maestro-flow, smoke, mode:build, lifecycle:generate, shape:multi-node, feature:conversational]
+
+run_limits:
+  expected_turns: 30
```
rockymadden: 🟡 `30` was an estimate when you wrote it. There is real data now — this task ran in this branch's own smoke job:

```
weighted_score: 1.0   final_status: SUCCESS
duration: 417s   commands: 20   iterations: 1
cli: 1.203.0-dev.8572
```

All nine criteria scored 1.00. Worth replacing the estimate with the observed figure, and dropping these two limitations from the PR body:

> **The eval task has never been run through coder-eval** — it cannot be until #3827 is in the sandbox image. `expected_turns: 30` is an estimate.

That also discharges the only High finding from the advisory task lint, which asked for a passing-run claim before merge. You have one.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+core.trigger.conversation → wait-for-message → conversational agent ──┐
+                                   ▲                                   │
+                                   └───────────────────────────────────┘
+```
+
+That is a starting point, not the only supported shape. Add whatever the conversational flow needs: additional agent and send-message nodes, a decision on the agent's outputs, handoffs, parallel branches to execute behind-the-scenes tasks, an escalation to [hitl](../hitl/planning.md), a connector or RPA call between turns, or no loop back at all when the conversation should end.
+
+### Get Conversation Context
```
rockymadden, line 77: 🟡 Third near-identical statement of the same point. `planning.md:15` (node table) and `impl.md:193` say it too, each landing on "wait-for-message already returns the context, so only reach for this when truly needed."

This section is the right home — it has the detail the other two lack (the no-exchange-yet constraint). Cut the other two to a link.

rockymadden (changes requested):
## 🎯 Re-review

Five commits since my last pass. **All 12 findings addressed** — 10 fully, 2 partially. CI is green (992 shared checker tests pass locally), and **the eval task now runs and scores 1.0**.

## ✅ Verified fixed

| # | Finding | How I checked |
| --- | --- | --- |
| 1 🔴 | CI red — headless preamble + forbidden `do not … debug` phrase | Preamble is verbatim; `pytest tests/tasks/uipath-maestro-flow/_shared/` → 992 passed |
| 2 🟠 | The `debug` sentence handed the agent a graded answer | Sentence gone, 2.0-weight criterion kept. It scored 1.00 on the real run without being told |
| 3 🟠 | `agent-json` failed when `maxIterations` was absent | Now range-checked only when present. Absent → pass; `25` → `FAIL: … expects 1-8 when present` |
| 4 🟠 | `1.1.0` → `1.2.0` bump half done | All three sites. Released CLI 1.203.0 writes `1.2.0` for standalone and inline, both flavors |
| 5 🟠 | `node add` conflicted with `inline-agent/impl.md:519` | Scoped on both sides with reciprocal links; all four new anchors resolve |
| 6 🟡 | Grader rejected a `get-conversation-context` root | Accepted now. Invented root still fails; `=js:` and the three-missing-keys trap still fail |
| 8 🟡 | `EXCLUDED_PARTS` parity claim | `node_modules` added |
| 9 🟡 | `node delete` / `-i` | `node remove` / `--input`, plus the pointer to `editing-operations-cli.md § Replace manual trigger` |
| 11 🟡 | CODEOWNERS voice scope creep | Confirmed via the org API that all five removed individuals are `@UiPath/jarvis` members — no effective ownership loss, and `@UiPath/maestro-cli-team` is added as asked |
| 12 🟡 | `targetRuntime` lost the inline carve-out | Stated now. `agent init --inline-in-flow --conversational` writes no `targetRuntime`; standalone writes `pythonAgent` |

### The eval task ran

From the smoke artifact on this branch's own run — so the PR body's "never been run through coder-eval" is stale:

```
weighted_score: 1.0   final_status: SUCCESS   duration: 417s
cli: 1.203.0-dev.8572   commands: 20
all 9 criteria: 1.00
```

I also rebuilt the documented recipe by hand on CLI 1.203.0 (`node remove start` → conversation trigger → wait → agent with `--source` → three edges) and got `Status: Valid`, with all three checks passing.

## 🚩 Two left, both one-liners

**🟠 Major — `<projectId>` is the wrong id for an in-solution node type**, in three places. Verified against a real solution:

```
registry get uipath.core.agent.991fc386…  (agent.json projectId)  → "Node not found"
registry get uipath.core.agent.685da0af…  (solution resource key) → Success
```

The suffix is the `key` in `resources/solution_folder/process/agent/<Name>.json`. Your own checker docstring, `planning-arch.md:145,336`, `agent/impl.md:7,92,109,112`, this file's Planning Annotation section, and the PR body all say resource key. Copilot flagged it six times. Details inline.

**🟠 Major — the new `mode` sentence contradicts "Write all five" two lines below it.** `impl.md:94` says `simple` "lets the platform derive the rest"; `impl.md:96` says nothing derives them outside the editor. As written it licenses skipping the one thing the plugin exists to teach. Details inline.

## 🟡 Minor

3. Mixed-flavor hole is narrowed, not closed — `engine` is unenforceable for project-backed agents because it is also the selector.
4. `scaffold.yaml:8` now contradicts the untouched checker comment sitting beside it.
5. PR body limitations are stale, and `expected_turns: 30` can be replaced with real data.
6. `get-conversation-context` is described three times in near-identical words.

## 🧹 Slop check

Clean, and unusually well-evidenced — I re-verified node types, ports, `latestExchangeId`, the five-key trap, and the whole recipe, and everything held. The only residue is the triple restatement in minor 6.

## 📌 Recommendation: **request changes**

Everything I blocked on last time is fixed and verified. Holding only on the two majors: the flavor table is the first surface a planning agent reads and it hands over a node type that returns "Node not found", and the `mode` sentence undercuts the plugin's headline rule. Both are single-line edits.

## tl;dr

You fixed all 12. CI is green and the eval scores a perfect 1.0. Two wrong sentences left: an in-solution agent's node type needs `<resourceKey>`, not `<projectId>` (I proved `<projectId>` returns "Node not found"), and the new `mode` explanation says the platform fills in the four settings fields when the next line correctly says nothing does. Fix those two lines and this ships.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
```diff
@@ -0,0 +1,133 @@
...
+
+| Flavor | Node type | Where the agent lives | Choose it when |
+| --- | --- | --- | --- |
+| **Inline** | `uipath.agent.conversational` | A UUID subdirectory inside this flow project | The agent exists only as part of this conversational flow, or you need [structured outputs](impl.md#structured-outputs) to route on — only inline has them. You scaffold it with `agent init --inline-in-flow --conversational`. |
+| **In-solution** | `uipath.core.agent.{key}` | A sibling project in the same solution | The agent is its own project, versioned separately, maybe reused by other flows in the solution. Discover it with `registry list --local`. |
+| **Published** | `uipath.core.agent.{key}` | The tenant, already published | The user names an existing agent, or one is already deployed. Discover it with `registry search`. Nothing to scaffold. |
+
+`{key}` is a solution resource key — the `key` in `resources/solution_folder/process/agent/<Project>.json` for an in-solution agent, the Orchestrator-assigned one for a published agent. It is **not** `agent.json`'s `projectId`. See [agent/impl.md](../agent/impl.md).
```
rockymadden: **🟠 major** The `{key}` fix over-reached onto the published flavor.

> `{key}` is a solution resource key ... the Orchestrator-assigned one for a published agent.

A published agent's suffix is not a resource key. `agent/impl.md:81`, the file this sentence links to, says the `type` suffix is "the Orchestrator-assigned **UUID**, per ring and agent". A published agent does have a `resourceKey`, but it is a *different* field: composite `<FolderPath>.<AgentName>`, read from `model.bindings.resourceKey` on the manifest, never the node type. The row said `<guid>` before, which was right.

The wrong action this licenses: writing `uipath.core.agent.<FolderPath>.<AgentName>` as a published node type.

Fix, one row plus one sentence (also `planning.md:89` and `impl.md:164`):

```
| **Published** | `uipath.core.agent.<agentUuid>` | ... |

`{key}` is the in-solution agent's solution resource key, the `key` in
`resources/solution_folder/process/agent/<Project>.json`, and **not**
`agent.json`'s `projectId`. A published agent's suffix is the
Orchestrator-assigned UUID instead. Read either off `registry get`.
```

The in-solution half is correct now, so this is narrower than last round.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -0,0 +1,291 @@
...
+  "mode": "simple",
+  "context":        { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext",                  "fieldType": "object" },
+  "conversationId": { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.conversationId",   "fieldType": "string" },
+  "exchangeId":     { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.latestExchangeId", "fieldType": "string" },
+  "messages":       { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.messages",         "fieldType": "array"  },
+  "userSettings":   { "type": "jsExpression", "expression": "$vars.waitForMessage1.output.conversationContext.userSettings",     "fieldType": "object" }
+}
+```
+
+**`mode` selects the contract.** `simple` (the default) is authored from a single `context` binding, which Studio Web's panel and the SDK expand into the other four at author time. `custom` is for a turn assembled from more than one source, and makes `conversationId` mandatory. The field is optional and ungraded; the examples here all declare `simple`.
```
rockymadden: Two problems in one sentence.

> `custom` is for a turn assembled from more than one source, and makes `conversationId` mandatory. The field is optional and ungraded

**🟠 major** `conversationId` is mandatory in *both* modes. Your own line 98 says so unconditionally ("Validation only requires `conversationId`"), and `flow-builder-sdk` `typescript/sdk/src/serialize.ts:4987` confirms it: "its validator refuses a node without `conversationId` even when a context binding is present". As written, a reader in `simple` mode infers it is optional.

**🟡 minor** "ungraded" is eval-harness vocabulary in an authoring doc. It is true (`mode` is absent from `SETTINGS_KEYS`), but a reader authoring a flow grades nothing. "Optional" already says what to do.

Shorter replacement, straight from the platform editor via that same SDK comment:

```
**`mode` and `context` exist only to restore the editor UI.** The four
individual fields are what the runtime reads, in either mode, and
validation requires `conversationId` regardless. `custom` is for a turn
assembled from more than one source. `mode` is optional; the examples
here all declare `simple`.
```

The rest of this fix is right, and better sourced than before.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,537 @@
...
+            attempted.add(path)
+            if _grade_agent_json(path, problems):
+                graded.add(path)
+
+    if any(not _is_inline(n) for n in agents):
+        # No `engine` filter here: that is the field `_grade_agent_json` asserts,
+        # so filtering on it would drop a misconfigured sibling instead of
+        # failing it. A published agent has no project in the solution, so it
+        # contributes no file and is correctly absent.
+        for path in _agent_json_files(root):
```
rockymadden: **🟠 major** This sweep fails a correct in-solution build when any non-conversational agent shares the solution.

Reproduced with a fixture (correct in-solution chat flow, one autonomous sibling project):

```
A) autonomous sibling present     settings PASS  loop PASS  agent-json FAIL
   FAIL: InvoiceAgent/agent.json: settings.engine is 'autonomous-v1', expected
   "conversational-v1"; metadata.isConversational is None; maxIterations is 20
B) sibling deleted                settings PASS  loop PASS  agent-json PASS
```

The previous checker (`f8e8006e9`) **passed** fixture A. So last round's minor traded a narrow gap (a misconfigured sibling silently skipped) for a broader false failure, and mixed solutions are the normal case: `planning.md:24` sells in-solution as "maybe reused by other flows in the solution".

Fix: grade the agent the flow references. The in-solution suffix *is* the solution resource key, so resolve it through `resources/solution_folder/process/agent/*.json` (`key` to project name) rather than sweeping. Cheaper alternative: keep the sweep, fail only when **no** sibling grades clean, and report the rest as `INFO`.

Latent, not a red build: the task pins inline, so this path never runs in CI today. It is the one open finding I would not leave, since the code it replaced was correct here.

tests/tasks/uipath-maestro-flow/conversational/check_conversational_flow.py
```diff
@@ -0,0 +1,537 @@
...
+            if path in attempted:
+                continue
+            attempted.add(path)
+            if _grade_agent_json(path, problems):
+                graded.add(path)
+        if not graded:
+            return _fail(
+                f"no conversational agent.json found under {root} — this task "
```
rockymadden: **🟡 minor** This message lost its honest caveat. It was:

> ...so one must exist on disk (a published agent would have none, but the sandbox has no tenant)

The parenthesis was the part explaining that this failure is expected rather than a bug, and it still fires: a published-flavor flow with no agent project on disk fails `agent-json` outright (reproduced). Restore it.

tests/tasks/uipath-agents/lowcode/conversational/scaffold/scaffold.yaml
rockymadden: **🟡 minor** Still claims a field the checker refuses to grade.

This line lists "maxIterations 8" among what the task tests; `check_conversational_scaffold.py:24` says it "is reported but not graded". Half-fixed from last round: the checker comment was reworded, this line was not.

Either drop it from the description or grade it, since the same commit asserts the CLI is now consistent at 1.203.0.

rockymadden (approved):
## 📋 Round 3

Four commits since my last pass. **Both majors I blocked on are addressed**, and all four minors are touched. Two of the fixes over-reached and introduced new problems, all called out inline.

## ✅ Verified fixed

| Last round | Now |
|---|---|
| 🟠 `<projectId>` was the wrong id for an in-solution node type | In-solution is correct: `{key}` = solution resource key, sourced to `resources/solution_folder/process/agent/<Project>.json`. Wrong for **published** now, see the `planning.md:27` comment |
| 🟠 The `mode` sentence contradicted "Write all five" two lines below | Contradiction gone, and "Write all five" is now confirmed by an independent source. A different false clause remains, see the `impl.md:94` comment |
| 🟡 Mixed-flavor grading hole | Closed, but by over-grading. See the `check_conversational_flow.py:506` comment |
| 🟡 `scaffold.yaml:8` vs the checker comment | Checker comment reworded, `scaffold.yaml:8` not. Still contradictory |
| 🟡 PR body stale, `expected_turns: 30` | Body rewritten, `expected_turns: 20` matches the observed command count |
| 🟡 `get-conversation-context` said three times | One canonical section plus two links. Both new anchors resolve |

## 🔬 What I confirmed this round

- **The eval task passes on this head.** `skill-flow-conversational-chat-loop`: `status=SUCCESS score=1.000`, 9/9 criteria, 320s, 1 iteration, in run [34396685605](https://github.com/UiPath/skills/actions/runs/34396685605). Second consecutive clean run.
- **The debug hand-off section is accurate.** CLI 1.203.0 `debug --help` states it in the same terms as `impl.md:266`: "handed off, since only a chat UI can drive it. No debug session is started", `handedOff true`, and `--open-in-browser` exists.
- **"Write all five" is right**, and there is a source that proves it. `flow-builder-sdk` `typescript/sdk/src/serialize.ts:4984`: *"The FOUR individual fields are the runtime's source of truth, the platform's own editor says so ('they always reflect the current configuration regardless of mode; `mode` and `context` exist only to restore the editor UI')."* The earlier "lets the platform derive the rest" was the wrong version.
- Link checker: 6773 links resolve. `_shared` checker tests: 997 passed.

## 🚩 Overall findings

Line-level findings are inline. These have no single line to sit on:

**🟢 CI red is not yours.** `Run skill smoke tests` fails at 94.4% < 95%, from three `skill-flow-datafabric-smoke-*` tasks. Those task files are byte-identical to `origin/main` on this branch (empty diff), and the failures are Data Fabric entity-record assertions, nothing conversational. It still blocks merge until main is green or the run is retried.

**🟢 Ticket check, which your PR body asked for.** JAR-10152 is a stub pointing at Confluence [Conversational / Voice Flows, CLI / Skills Changes](https://uipath.atlassian.net/wiki/spaces/JAR/pages/91016397010). Your rows (5 conversational-inline, 6, 7, 9, 10, 11, 12) are all delivered, including row 12's keyword row (`chat`, `chatbot`, `voice`, `phone` are in the SKILL.md frontmatter). **Nothing in the ticket is missed.**

**🟡 minor, one deliberate deviation to confirm.** Checklist rows 1 and 3 ask for separate `conversation-trigger/` and `conversational/` plugin folders. You folded all five nodes into one `conversational-agent/`. That reads like the better call, the loop only works as a unit, and those rows are marked "Claude generated, may need validation". Confirm it was deliberate so the checklist can be closed rather than left looking unfinished.

## 🧹 Slop check

Clean. Prose is specific and sourced, and both things I flagged last round are gone: the triple restatement is one section plus links, and the `node-output-wiring.md` rows are down to the rule plus a link. Only residue is the "ungraded" clause in the `impl.md:94` comment.

## 🧭 Recommendation

**Approve.** The documentation is verified against a live CLI, the eval scores a perfect 1.0, and the ticket is fully covered.

Three findings are still open and land as follow-ups rather than blockers: two wrong sentences and one checker predicate. The checker one (`check_conversational_flow.py:506`) is the one I would not leave: it fails a correct in-solution build, and the code it replaced passed that case.

Being straight about the treadmill: this is round 3 on the same flavor table and the same `mode` sentence. Both fixes were directionally correct and both over-reached by one clause. Worth a slower read of the two replacements inline than of the originals.

---

**tl;dr** — Approved. Both blockers fixed, eval passes 9/9, CI red is main's Data Fabric tasks and not yours, and the ticket is fully covered. Three follow-ups inline: published agents take an Orchestrator UUID, not a resource key; the checker now fails a correct in-solution build whenever an unrelated agent shares the solution (I proved the old code passed that case); and `conversationId` is mandatory in both modes, not just `custom`.

maxduu (approved):

PR #2976, opened 2026-09-02
fix(maestro-case): keep the SDD's output rows on connector tasks, and re-escape on the way into JSON

Yiqi517:
Two defects on the connector write path, both found in supplier-onboarding coder-eval runs. `uip maestro case validate` returns `Valid` for both, so neither shows before the case runs.

## 1. Declared output extracts are dropped

`case spec --type activity` returns only the connector's own top-level outputs. For Outlook Send Email that is exactly two entries:

```json
"outputs": [
  {"name": "response", "type": "jsonSchema",
   "properties": ["status", "id", "conversationId", "internetMessageId"]},
  {"name": "Error", "type": "jsonSchema"}
]
```

An SDD row like `response.status -> lastEmailStatus` is not in that list and never will be. Copying `caseShape.outputs` into `data.outputs` therefore drops every declared extract: six connector tasks in one run shipped with no `lastEmailStatus` binding at all.

The write sites now say the output set is what [`io-binding/impl-json.md` § Output Binding Shapes](https://github.com/UiPath/skills/blob/main/skills/uipath-maestro-case/references/plugins/variables/io-binding/impl-json.md) produces from `caseShape.outputs` plus the SDD's `->` and `=` rows: `implementation.md` Step 9.7 Phase B and `phased-execution.md` Phase 3 step 1.

## 2. An `=js:` string literal loses one escape level

An sdd.md cell holds JavaScript source, so a line break inside a quoted literal is already the two characters `\` and `n`. Every JSON carrier below it needs one more level. Copying the cell character-for-character into `input-values` writes `\n`, JSON decodes that to a raw line break inside the `"` literal, and the element faults on first evaluation.

| carrier | text it holds |
|---|---|
| sdd.md cell | `\n` |
| `input-values` in `registry-resolved.json` | `\\n` |
| `--input-details` JSON | `\\n` |
| `caseplan.json` on disk | `\\n` |
| the JavaScript the runtime parses | `\n` |

The level is added at exactly one hop, so the rule lives at that hop: `connector-activity/planning.md` § 8 Build input-values. Every later hop is JSON to JSON, and `connector-activity/impl-json.md` Step 1.c now says to copy verbatim and re-escape nothing. Adding a level there writes `\\\\n`, which prints a literal backslash-n in the message body with no error anywhere.

Step 2's own command template single-quotes `--input-details`. Bash strips one backslash level inside double quotes, so the double-quoted form delivers `\n` and reintroduces the fault.

## Why nothing catches it today

The runtime hands everything after `=js:` to Jint: [`JavaScriptEvaluator.cs#L120`](https://github.com/UiPath/PO.BpmnEngine/blob/8ec541365/src/BpmnEngine/UiPath.PO.BpmnEngine/Expressions/JavaScriptEvaluator.cs#L120) calls `engine.Evaluate(expr[4..])`, which throws `Invalid or unexpected token`. Validating an artifact carrying 46 raw line breaks reports 16 errors, none of them about the expressions: `validate` does not parse `=js:` bodies. The packer copies the break into the `.bpmn` byte for byte.

[UiPath/cli#3935](https://github.com/UiPath/cli/pull/3935) closes that gap with a `--strict` rule that compiles every `=js:` value under the same grammar Jint uses.

## Output ids: `camelCase(name)`, not `v` + 8

`connector-trigger-impl.md` and `connector-activity/impl-json.md` stated the `v` + 8 form for both inputs and outputs. That is the input rule, and the product's own code says why it is input-only:

```ts
// for inputs always generate a random ID to avoid conflicting with outputs and other variables
const id = input.id ?? input.var ?? generateNextId("v", 8);
```

[`CanvasTaskElementUtils.tsx#L62`](https://github.com/UiPath/PO.Frontend/blob/e7f32d8b1c70c5d5768cf51ce9fd8a52a6618937/src/components/canvases/shared/utils/CanvasTaskElementUtils.tsx#L62). Twenty lines below, [outputs](https://github.com/UiPath/PO.Frontend/blob/e7f32d8b1c70c5d5768cf51ce9fd8a52a6618937/src/components/canvases/shared/utils/CanvasTaskElementUtils.tsx#L83) take `generateNewVariableId(existingVariableIds, output.name)`, which resolves to [`toUniqueCamelCase(name, taken)`](https://github.com/UiPath/PO.Frontend/blob/e7f32d8b1c70c5d5768cf51ce9fd8a52a6618937/src/utils/VariableBindingsUtils.ts#L43): camelCase the name, append `2`, `3`, ... on collision. `v` + 8 is only the fallback for an entry with no name. Every other id-minting site in the frontend does the same, including the `->` rename path in [`useOutputVariableReassignment.tsx#L264`](https://github.com/UiPath/PO.Frontend/blob/e7f32d8b1c70c5d5768cf51ce9fd8a52a6618937/src/hooks/useOutputVariableReassignment.tsx#L264).

Both files now split the convention by direction, which also matches what `io-binding/impl-json.md` has always said.

## Two smaller corrections found in the same files

`task-entry-conditions/planning.md` carries the `activation-mode` to rule-type table. `parallel-after-predecessor` is used in 14 files across the planner templates and the task plugins but was missing from that table and from the plan-presentation enum, so an agent reading the table alone had no mapping for it. It maps to `runs-sequentially`, the rule type the product uses for a task transition.

The same file scoped the ordered-run heuristic to design-lane authoring (`uipath-planner`). Phase 1 does not re-author a supplied SDD, so the heuristic never applies there.

## Verification

6 files, +33/-8, all reference Markdown. No code, no fixtures, no CODEOWNERS change.

- `case spec` output shape read from a real `registry-resolved.json`
- Escape levels measured on two artifacts: one carrying `\\n` x87 with zero raw newlines (runs), one carrying `\n` x46 that decodes to 46 raw newlines (faults). Both keep their byte form through the packer
- `validate` on the faulting artifact reports 16 errors, zero about the expressions
- Bash quoting checked directly: `"a\\nb"` delivers `a\nb`, `'a\\nb'` delivers `a\\nb`
- Every frontend caller of `generateNewVariableId` and `toUniqueCamelCase` enumerated; none mints `v` + 8 for an output


Yiqi517:
Both manual-review items check out. Measurements below, no change needed.

**1. Does the packer pass `\\n` through unchanged?** Yes. `serializeCaseDiskJsonToBpmnAndRulesJson` builds the XML from parsed objects and never re-encodes expression strings. Measured on two supplier-onboarding artifacts, one correct and one broken:

| run | caseplan file bytes | what the JS source receives | packed `.bpmn` | outcome |
|---|---|---|---|---|
| ev5 | `\\n` x87, `\n` x0 | `\n` escape x87, raw newline x0 | `\\n`, unchanged | runs |
| ev3 | `\n` x46, `\\n` x0 | raw newline x46 | `\n`, unchanged | faults |

That is the invariant the new text states: write `\\n` in the JSON so the JavaScript source carries `\n`.

**2. Is `camelCase(name)` the right output mint shape?** Yes, and it is the product's own behaviour, not just this skill's convention. `CanvasTaskElementUtils.tsx` builds a task's `data.outputs[]` and, for an entry with no id yet, calls `generateNewVariableId(existingVariableIds, output.name)`. That resolves to `toUniqueCamelCase(output.name, existing)` in `VariableBindingsUtils.ts`: camelCase the name, append `2`, `3`, ... on collision. `generateNextId("v", 8)` is only the fallback for an entry with no name at all.

Inputs are the deliberate opposite, and the code says why: `// for inputs always generate a random ID to avoid conflicting with outputs and other variables`. So `v` + 8 is the input rule, and this PR was reading it onto outputs.

This skill already said the same thing in `references/plugins/variables/io-binding/impl-json.md` § Output Binding Shapes: bare auto-mint is `id: <camelCase(name)>, var: <id>`, a `->` reassign takes `baseId = camelCase(leaf segment)`, and an unreferenced schema field falls back to auto-mint. Across the caseplans I have on disk, 663 task outputs use that form and 28 use `v` + 8. So the change removes a divergence rather than creating one.

BenHorblit-UiPath (approved):
Ah this is interesting

charlesliu9:
Which tests have you run?

Yiqi517:
> Which tests have you run?

@charlesliu9 two things:

**Both defects come from one real supplier-onboarding run.** The sdd.md cell holds `\` and `n` as two characters, and the caseplan it produced holds a real line break instead, in all six email bodies. `lastEmailStatus` appears zero times in that run's 496 messages, so the declared extract was never considered at all. I re-ran supplier-onboarding with this PR merged into the test branch three times ([33638762009](https://github.com/UiPath/skills/actions/runs/33638762009), [33630358431](https://github.com/UiPath/skills/actions/runs/33630358431), [33623354811](https://github.com/UiPath/skills/actions/runs/33623354811)) and neither defect appeared in any of them.

**Nothing on main would catch either one, so I added the coverage on a separate branch.** No case task's SDD carries a `=js:` value with a `\n`, and the only `->` extract on a connector task sits on an `<UNRESOLVED>` connector, so neither path is ever entered and every current task grades green whether this is fixed or not. That branch adds two SDD rows and two checker assertions to `connector_features/connector_activity`: [run 33696289581](https://github.com/UiPath/skills/actions/runs/33696289581), 1/1 pass.

charlesliu9 (approved):

roberts-cliff (approved):

StephenHanzlik (approved):

PR #2987, opened 2026-09-02
feat(ontology): add coded actions for writes SQL cannot express [DS-9390]

nathanverghese-uipath:
Adds coded actions to the ontology skill family: an author states a business rule in prose, the
agent decides whether it is expressible as declarative SQL, and generates a TypeScript job plus a
coded-action TTL when it is not. Extends `uipath-ontology-authoring` and `uipath-ontology-modeler`,
adds one new skill for the Orchestrator leg, and adds a shared validator.

Base: `feat/ontology-onboarding`. Not for `main` while the ontology product is in development.

## What

An Ontology action today is one authored SQL statement with `{{Entity.field}}` placeholders. The
write surface is literals-only by design, so an action whose value has to be computed cannot be
expressed at all. Coded actions move that computation into a job on Orchestrator and keep the commit
boundary declarative, so the mutation surface stays as bounded as it was.

Authoring classifies each extracted write operation on a four-test rubric, first match wins:

| Test | Verdict |
| --- | --- |
| The value must be computed from stored data, the clock, or string construction | CODED |
| It reads before writing, branches, loops over rows, writes several rows, or may write nothing | CODED |
| A rule depends on a fact the caller must not be trusted to assert | CODED |
| Otherwise: single entity, single record, caller-supplied literals | SQL |

Ambiguous cases default to SQL. `setStatus(id, status)` stays declarative; `tagOverdueTicket`
(clock arithmetic, tag-list append, converged no-op) does not.

## The new skill

`uipath-ontology-coded-action-deploy` owns the Orchestrator leg only and never touches `uip ont`,
mirroring the modeler's never-upload rule. Six phases: scaffold a `{name}-jobs` Solution, stage each
job into its project's `functions/`, pack and publish and deploy, resolve the numeric folder id,
patch it into the TTL over the `PENDING_DEPLOY` placeholder, hand the patched paths back. Three
Python scripts, dry-run by default with `--execute` gating every mutating path.

Authoring gains one sequencing step between the modeler's return and the Tier 2 upload: when coded
actions exist, delegate here, then re-run the validator on the patched TTLs before uploading. A
SQL-only inventory skips the whole leg.

## Packaging: the manifest is derived

`uip solution pack` requires each project's `entry-points.json` and never produces one. The command
that would produce it, `uip functions pack`, cannot lower the `type<T>()` contract idiom on any SDK
version (0.4.4, 0.5.0, 0.6.4): only Studio Web's packer carries that derivation walker.

The verified pipeline sidesteps it. The ontology repo's working skill runs `uip solution pack`,
`publish` and `deploy run` and never calls `uip functions pack`; its solution came from a Studio
Web export whose manifest was already derived, and `solution pack` only zips a directory and reads
no TypeScript. `tools/entry_points.py` does that derivation here, reproducing byte-for-byte what
Studio Web produced for both verified jobs. Those manifests are committed as goldens and are the
only evidence of what the platform accepts.

So jobs keep the `type<T>()` interfaces the existing jobs use, the interfaces stay the single
source of truth, the manifest is regenerated on every stage so the two cannot drift, and nothing
in stage, pack, publish or deploy runs an installer. A contract outside the lowerable grammar is
refused rather than approximated, and the validator runs the same deriver as a gate so that fails
at authoring time instead of at pack time.

The spike also settled that a CLI-scaffolded Solution yields a real job release
(`ProcessType: Function`, `ProcessKey: {Solution}.Function.{Name}`), not the HTTP-endpoint
flavour, which confirmed CLI scaffolding as the primary path and left the shipped template
skeleton as a fallback.

## Federated entities are readable and writable

The branch said federated classes were read-only and could not be targeted by write actions. That
is wrong, and the modeler's own reference guides already said so. Corrected in authoring's federated
rules, the modeler's standalone-mode paragraph, and the mapping guide. `readOnly` survives in
CLASS_MAP as an explicit per-source exception the author states, never inferred from federation. The
source system stays the authority on whether a write succeeds, and a rejection surfaces as an
upstream error on the failing step.

`skills/uipath-platform/references/data-fabric/` still states federated rows are read-only. That is
the `uip df records` surface rather than the FQS path, so it was left alone; worth a word with that
skill's owners on whether it holds there.

## Validator

`tools/coded_action_preflight.py`, dependency-free, JSON shaped like `ontology_preflight.py`. Nine
gates over each TTL and job pair: marker resolution, input equality against the marker, writes
attribution per edit, fields existing in the local `.ofn`, zod strictness, `PENDING_DEPLOY` state,
and an optional typecheck that reports `skipped` when npx is absent. The gate that matters most is
writes attribution: `ont:writes` is the union over every branch the job could take, and an edit
outside it is refused whole at `Preparing write statement` after the job has already run. The guide
says to declare the worst case, because over-declaring is free and under-declaring fails closed at
invoke time.

Unrecognized property shapes route to `unresolved` and fail the gate rather than reading as no
writes, which is the one place a validator like this can quietly do real damage.

## Tests

`tests/coded_action_preflight/`, 15 tests, real subprocess and real files, no mocks. 14 pass and the
typecheck test skips without `tsc`. Two assert the fail-closed posture directly: an absent schema
skips the field gate rather than passing it, and untraceable edit properties fail rather than
reading as no writes.

## Known gaps

Found by a review pass over the finished work, all reproduced, none yet fixed:

| Gap | Effect |
| --- | --- |
| `folder-id` and `await` resolve the solution name before dispatch | Phase 4 exits 1 without `SOLUTION_SRC`, contrary to its own docs |
| Read-vs-param classification is computed but never consumed by a gate | A read declared as a scalar passes every gate |
| Computed property keys and object spreads are not detected | An undeclared write passes, then is refused at invoke time |
| `deployments()` returns `[]` on a failed `deploy list` | The idempotence guard fails open and would create a second deployment |
| `folder-id-status` never fails | Authoring's "require no `PENDING_DEPLOY`" is not enforceable by the command it prescribes |
| `ont:writes` written as repeated predicate lines | Valid Turtle, rejected with a message that blames the job |

Also outstanding: the coded-action pair tests sit at `tests/coded_action_preflight/` where the
sibling tool's live at `tests/scripts/`, so no CI job picks them up (the deriver's tests are at
`tests/scripts/` and are triggered). The new skill has no activation dataset where every sibling
ontology skill has one, and no ontology skill yet appears in `assets/skill-status.json`, the README
table, or CODEOWNERS; the family needs registering before this merges.


skills/uipath-ontology-authoring/SKILL.md
```diff
@@ -14,6 +14,7 @@ user-invocable: true
 - Artifact folder intended for a new deployment or clone → this skill.
 - Plain-language domain description with no files → `uipath-ontology-modeler`.
 - Existing ontology CRUD, API, SDK, or artifact operations → `uipath-ontologies`.
+- Business rules that require computation route through this skill's classification of write operations; the coded leg's Orchestrator work belongs to `uipath-ontology-coded-action-deploy`.
```
vikasjayaraman: We don't yet have a clear definition of Business rule or constraint. So, we should not classify coded-leg vs not based on it. Try to build the reasoning around need for turing complete capabilities not expressible via SQL (and expanding to PL/SQL). Most simplest one is anything that spans beyond basic SQL and needs any computation (not restricted to business rules/constraints) - should get here.

skills/uipath-ontology-authoring/SKILL.md
```diff
@@ -37,15 +38,17 @@ Runtime delegation to `uipath-ontology-modeler` is intentional and is the only s
 ONTOLOGY_NAME: exact slug
 ONTOLOGY_IRI: https://ontology.uipath.com/{name}#
 WORKDIR: dedicated {name}/ output directory
-CLASS_MAP: class -> entityName, entityId, folderId, readOnly (federated only)
+CLASS_MAP: class -> entityName, entityId, folderId, readOnly (optional, rare)
```
vikasjayaraman: what do we mean by optional, rare?

skills/uipath-ontology-authoring/SKILL.md
```diff
@@ -37,15 +38,17 @@ Runtime delegation to `uipath-ontology-modeler` is intentional and is the only s
...
 MAPPING_STATUS: supplied | generate
 DOMAIN_MODEL: confirmed classes, properties, relationships, rules
 ANNOTATIONS: confirmed labels, comments, synonyms, value domains, and grain
-OPERATIONS: grouped query operations and structured write actions, if any
+OPERATIONS: grouped query operations and structured write actions, if any; every write action carries kind: SQL | CODED, and every CODED action also carries its reads, its writes union, and its process name
 DEPLOYMENT_MODE: delegated; modeler returns local-preflighted artifacts; authoring validates and uploads the inventory, with mapping as the final deploy trigger
 PREFLIGHT_HANDOFF_JSON: machine-readable JSON with CLASS_MAP, FIELD_METADATA, and explicit RELATIONSHIPS ([] when none)
 ```
 
+The modeler emits CODED actions per its coded-action contract guide, with the TTL carrying `ont:language "IMPERATIVE"` and `ont:processFolderId "PENDING_DEPLOY"`.
```
vikasjayaraman: Should we rather call this TS rather than imperative? This is not imperative.

vikasjayaraman: Also, the 2 options don't seem logical comparable between SQL Vs Imperative.

skills/uipath-ontology-authoring/SKILL.md
```diff
@@ -96,6 +100,26 @@ While reading the SDD, also scan for:
...
 
+Classify each recorded write operation as `kind: SQL` or `kind: CODED`. Apply these tests in order; first match wins:
+
+1. CODED if the new value must be computed from stored data, the clock, or arithmetic/string construction. The SQL surface is literals-only: no `GETDATE`, no expressions, enforced by Data Fabric's own parser (`SET x = x + 0` and `SET x = CASE...` are both rejected with "Only literal values supported").
+2. CODED if the operation reads before writing, branches on what it finds, iterates over rows or items (any per-row loop), may write several rows or several entities, or may legitimately write nothing (converged no-op).
+3. CODED if a rule depends on a fact the caller must not be trusted to assert. Values derived from data are computed from declared reads, never from caller parameters: a caller can lie about any fact concerning the data.
+4. Otherwise SQL: single entity, single record, caller-supplied literal values. Ambiguous cases default to SQL. Declarative by default; a job in a high-level language (currently supported: TypeScript) only when the edit cannot be expressed before it is computed.
```
vikasjayaraman, line 184: One more point to add here is:
Any action requiring reads in between writes or reads after writes are not supported and should fail at authoring gate itself for now.

skills/uipath-ontology-authoring/SKILL.md
```diff
@@ -96,6 +100,26 @@ While reading the SDD, also scan for:
...
+Worked classifications:
+
+| Operation | Verdict | Why |
+|---|---|---|
+| `updateAccountDescription(id, text)` | SQL | one row, caller supplies the value |
+| `setStatus(id, status)` | SQL | same |
+| `tagOverdueTicket(ticketId)` | CODED | clock arithmetic, list append, converged no-op |
+| `flagBigOrder(invoiceIds)` | CODED | per-row loop over lines, multi-row write |
+| `setInvoiceDecision(id, decision, approver)` | CODED | boundary case: looks like a parameter write, but composes a rationale string and screens caller-asserted facts |
+| `approveIfUnderLimit(id, amount)` | CODED | rule 3: the amount must come from a read, not the caller |
```
vikasjayaraman, line 202: Where are these examples? Are these purely to be inferred by the name and intent?

skills/uipath-ontology-modeler/SKILL.md
```diff
@@ -31,11 +32,11 @@ When called by ontology authoring, require one complete handoff object. Do not r
 ONTOLOGY_NAME: exact slug
 ONTOLOGY_IRI: https://ontology.uipath.com/{name}#
 WORKDIR: dedicated {name}/ output directory
-CLASS_MAP: class -> entityName, entityId, folderId, readOnly (federated only)
+CLASS_MAP: class -> entityName, entityId, folderId, readOnly (optional, rare)
```
vikasjayaraman: same as above - what does optional, rare mean?

tools/coded_action_preflight.py
vikasjayaraman, line 1: Can this be split into logical parts which tie to the Skill instructions? It's hard to review a 1200 line utility script without context of what each function is needed for.

skills/uipath-ontology-authoring/SKILL.md
nathanverghese-uipath: We can call this Coded instead of Imperative, I will make the change in the ontology PR accordingly.
Calling it TS would mean that we would have to extend the logic for each type of coded language, but for the same task, so it makes sense to house it all in one type.

skills/uipath-ontology-authoring/SKILL.md
nathanverghese-uipath: readOnly is rare now. In case we don't want this I can change it.

skills/uipath-ontology-authoring/SKILL.md
nathanverghese-uipath: Only added two examples as more may not be required for context and will be overkill

PR #3001, opened 2026-09-02
fix(tests): repair the phase-0 case eval family — 13 harness defects

abhiram-vad:
# fix(tests): repair the phase-0 case eval family — 13 harness defects

## TL;DR

The latest nightly's four `uipath-planner` failures were all case tasks, and all four were **instrumentation rather than model quality**. All four now pass:

| Nightly failure | Root cause | 
|---|---|
| loan-origination 0.53 — *"no SDD artifact was found"* | The agent-facing prompt never named the output file. Every `sdd.draft.md` mention lived in the task description, a YAML comment, or grader config — none of it reachable by the agent — so it invented a name and the exact-path grader found nothing. |
| candidate-interview 0.65 — judge **0.000** | Same missing filename; the judge received no file at all. |
| finalize-draft 0.86 — judge 0.200 | The judge graded the *draft's* design quality, which the prompt explicitly forbids the agent from changing (*"the design is settled, so don't redo it"*). |
| finalize-draft-loan 0.78 — judge **0.000** | Same, scored on the fixture's own inverted task order — *"funding before recording"* is staged input, not a model defect. |

**No threshold, weight, or judging bullet was loosened.**

## The change that mattered most: the judge's scale

The design-coherence judge scored the same defect anywhere from **0.48 to 0.78**, with the 0.70 pass bar sitting inside that band. Its own rationales on drafts it **passed** — *"one material routing gap and a couple of incompletely specified event/fan-in behaviors"* — are near-identical to those on drafts it **failed** — *"the offer path has a material downstream gap"*.

Measured support: across 21 collected drafts the 4 passing ones were structurally indistinguishable from the 17 failing ones — same stage count (10), comparable tasks (19 vs 16), variables (14 vs 13), guards (16.8 vs 20.6). The variance was the judge's severity assignment, not the draft, which is why five consecutive draft-side fixes moved nothing.

Cause: a 3-band scale (1.0 "no major flaws" / 0.5 "minor gaps" / 0.0) with the bar in the gap between the top two bands. The scale now carries anchors written **from the four drafts this judge itself scored 0.72–0.78**. `pass_threshold` stays 0.7, no weight changed — and the judge subsequently **failed a genuinely broken draft at 0.5** (an unroutable Debrief), so it discriminates rather than waving things through.

## Skill changes — four, each measured to help

1. **`SKILL.md` no-search rule** — paths in the skill are already known; never `Glob`/`find`/`ls` for a reference, template, or script. Three tasks had been spending their entire 60-minute budget searching for paths they already held.
2. **`audit_sdd.py` picker-route check** — a decision button's deterministic route cannot key a `user-selected-stage` lane. `reject-route` 0.385 → 1.000.
3. **`audit_sdd.py` ambiguous-persona check** — a task runs as exactly one persona, so `| Underwriter or Credit Analyst |` is a routing rule stated but never modelled. Calibrated against **144 collected artifacts** before being written: fires on 3, all failing drafts, and on **zero** passing artifacts. Only the persona column is inspected — scanning every cell also flagged legitimate task names like "Await Signed Offer or Decline".
4. **Layers-guide checklist item 2 now walks each conditional entry's FALSE branch.** A draft gave Onsite Loop entry `IF isEligibleForOnsite === true` and Debrief a single entry keyed on Onsite Loop's completion, so no non-Engineering candidate could reach Debrief. The draft papered over it as *"auto-completes as skipped"* — but a stage that does not enter cannot complete.

Also surfaced: `return-to-origin` pairs only with `required-tasks-completed` or `wait-for-connector`, and resumption decides `enter-stage` vs `exit-case`. `uipath-maestro-case` gains rules 6a/6b/6c (T-entry heading shape, entry-body field shape, plan gate).

## Two decisions worth a reviewer's attention

**The rule 13 carve-out.** `uipath-maestro-case` rule 13 bans Python/shell against artifacts; that ban was making the agent skip the read-only audit gates. The rule now says explicitly that those gates are required, not banned — they only report, never write. This widens an existing prohibition, so it deserves a second opinion.

**No static reachability check, deliberately.** Five structural hypotheses were tested against all collected artifacts and every one fired on passing drafts too, including ones scored 0.90 and 0.99. Required-plus-conditional entry is the accepted norm for terminal stages (`Hired`, `Resolved`), which complete through one of several terminal lanes. A missing `### Case Business Rules` section died the same way — 69 passing artifacts also lack it.

**No new prose in the Case Design Lane** beyond item 2's zero-line fold. That lane reads ~158 KB, and three separate additions each pushed tasks into their turn walls. Both new checks therefore live in the gate, which costs the agent no reading.

## Verification

```
pytest tests/tasks/uipath-planner          88 passed
pytest tests/tasks/uipath-maestro-case    112 passed, 14 skipped
pytest tests/scripts                      112 passed
```

Run `33553334389` (`parallelism=1`): `candidate-interview` SUCCESS 0.937 (judge 0.8) and `loan-origination` SUCCESS 0.937 (judge 0.8) **in the same run**. `finalize-draft` SUCCESS at judge 0.99 and 0.90; `finalize-draft-loan` SUCCESS at judge 0.94, 0.91, 0.90.

## Not verified

- **All 9 case tasks as a single run** — evidence is per-task across several runs.
- **The 25 non-case planner tasks.** The `SKILL.md` rule is skill-wide, reviewed by inspection only: it is scoped to this skill's own `references/`, `assets/`, and `scripts/` paths, and the two discovery-adjacent tasks (`filesystem_routing`, `smoke_file_reading`) name their inputs explicitly.

## Measurement note for whoever runs these next

Heavy design tasks must be measured at `parallelism=1`. At higher parallelism they ERROR at the turn wall with identical turn and command counts and *lower* token output — 273s/turn versus 49s/turn for the same work. That is runner latency, not deliberation, and an ERROR at the wall is a no-measurement, never a failing score. A single sample cannot distinguish a fix from that variance.


skills/uipath-planner/references/case/case-design-lane-guide.md
```diff
@@ -274,10 +274,16 @@ If the user explicitly asks to finalize the existing draft, choose `Use the draf
 10. Write, gate, flip per §Terminal step (the draft on disk is also a recovery point, so a compaction means re-finalizing from it), then the audit with the draft comparison:
 
     ```bash
-    python3 "<this skill's folder>/scripts/case/audit_sdd.py" <final SDD path> --draft <draft path>
-    # python3 absent (common on Windows) → retry the same line with `python`, then `py`
+    python3 "<skill folder>/scripts/case/audit_sdd.py" <final SDD path> --draft <draft path>
```
RaduAna-Maria: Wasn't this changed to be run with `node` instead of python? Or is this planned for a future pr?

skills/uipath-planner/references/case/case-design-lane-guide.md
```diff
@@ -274,10 +274,16 @@ If the user explicitly asks to finalize the existing draft, choose `Use the draf
...
     ```
 
+    `<skill folder>` is the directory you read THIS guide from, minus `/references` — you
```
RaduAna-Maria: Please remove

skills/uipath-planner/scripts/case/audit_sdd.py
RaduAna-Maria, line 1: See the comment above, please confirm when the python dependency will be removed

RaduAna-Maria (approved):

skills/uipath-planner/references/case/case-design-lane-guide.md
abhiram-vad: This is planned in future PR. I'm currently working on entirely removing this audit step.

skills/uipath-planner/scripts/case/audit_sdd.py
abhiram-vad: Yes, it will be removed. Will take a look at this today.

charlesliu9 (approved):

PR #3005, opened 2026-09-02
fix(uipath-maestro-flow): catalog the debug failure that happens before the flow runs

rockymadden:
## What

Adds a `failure-modes.md` entry for a `flow debug` that fails at the Studio Web upload step, before any element executes. Plus two small wiring changes: a discriminator on the "validate passes, debug faults" entry, and the voice-agent row repointed at the new entry instead of restating its remedy.

## Why

`failure-modes.md` catalogs ten patterns and every one is a *runtime* fault. A debug that never reaches the flow had no entry, and the envelope actively misleads: the CLI's generic `Instructions` tell you to check the flow project, the folder, and Studio Web availability, and all three are fine.

In the 2026-09-02 `skill-flow-slack-weather-pipeline` failure the agent hit exactly this, opened `failure-modes.md` looking for the pattern (it read most of the file), found nothing, and improvised: it cloned the Flow project into a second solution to get a green run. That left two `project.uiproj` in the tree, project discovery refused to pick between them, and a flow that had already completed correctly twice scored 0.0.

So this is the on-path file the failing agent demonstrably read. That is where the entry belongs.

## Verified before writing

- **Overwrite is the normal path and it works.** #3847 hand-verified it on alpha on 2026-08-31, two days before this run. The entry therefore leads with "a rejection is the exception" so nobody learns to reset the id on a first failure.
- **The remedy ladder is ordered cheapest-first.** Deleting the cloud solution the stale id names is better than resetting the id: the CLI's existing import-on-404 fallback then does everything, nothing is orphaned, and the `.uipx` is rewritten by the CLI. The GUID reset is the last resort for when the delete is refused too, which is what happened in the observed run.
- **The GUID reset does work.** `uploadOrOverwriteSolution` falls back to import only on 404 and rethrows everything else, so an id Studio Web does not know takes the import path. Demonstrated in the failing run: the first two debugs carried the local `solution init` GUID, logged "not found on Studio Web. Importing as new solution...", and both completed.
- **No flag forces the import path.** `flow debug` has `--folder-*`, `--poll-interval`, `--timeout`, `--inputs`, `--attachment`, `--open-in-browser`, `--login-validity`. `solution upload` has `--force`, which only skips the existence probe before calling the same Overwrite endpoint, and `--no-snapshot`.
- **`createSnapshot` is not the missing argument.** The debug push sends no snapshot and `solution upload` opts in; both were rejected identically in the observed run.
- **Replace, do not delete, the field.** Removing `SolutionId` fails `.uipx` validation, as the voice-agent doc already noted.
- Every recognition signal is a CLI-side string or envelope field (`Attempting to overwrite existing solution`, absent `elementExecutions`, `Retry: "RetryWillNotFix"`), not a backend error code, so the entry does not rot when a service message changes.

## The underlying bug is not fixed here

Leading suspect is CLI-side, in `solution-sdk/src/upload-service.ts`, and it belongs in the CLI repo rather than this PR. Import and Overwrite build materially different requests against the same backend:

| | Import | Overwrite |
|---|---|---|
| query | `?desktopJit=true&useFileRest=true&enableUnifiedBuild=true` | none |
| file field | `uploadFile` | `UploadFile` |
| telemetry | `telemetryData` | absent |

`enableUnifiedBuild` / `useFileRest` change how the server reads the uploaded archive, so an Overwrite that omits them plausibly gets the package parsed in the wrong mode, which is the shape of an "argument had an invalid value" rejection. Consistent with #3847's verification passing: that exercised a re-uploaded *downloaded* archive, while the failing case is a package the CLI built itself. The e2e mock matches on path alone and deliberately ignores the query string, so no test can catch the asymmetry.

Whatever the cause, `uploadOrOverwriteSolution` falling back to import on any non-404 rejection would remove the need for this entry. Until then the entry is the documented way out.

- Checks: `.maintenance/check-all.sh` 9/9, `npm run skills:validate`, `npm run skills:check-links`.


rockymadden:
This is not correct from what I can tell, closing.

PR #3006, opened 2026-09-02
test(uipath-maestro-case): give seeded fixture rules unique names for CLI 1.202

roberts-cliff:
The nightly image moved from @uipath/cli 1.202.0-dev.8414 to -dev.8468 between
2026-09-01 and 2026-09-02, and the new build validates condition displayName
uniqueness case-wide. Every LinearThreeStages / SlaResponse seed in this tree
carried "Entry Rule 1" on several conditions and "Complete Rule 1" on one exit
per stage plus the case-exit rule, so all eight now fail `uip maestro case
validate` before an agent touches them:

  diagnose/extract_alias_collision      Entry Rule 1 ×6, Complete Rule 1 ×4
  diagnose/formal_arg_slot_ids          ×6 / ×4
  diagnose/missing_task_entry_rule      ×5 / ×4
  diagnose/undeclared_gate_vars         ×6 / ×4
  edit/bound_task_add_remove            ×7 / ×4
  edit/templates (shared by edit/*)     ×6 / ×4
  sla_response/repair_any_escalation    ×6 / ×5
  sla_response/templates                ×5 / ×4

In the 2026-09-02 nightly this surfaced as four new case failures whose
agents did nothing wrong: the three diagnose-* tasks report "validate still
passes after the repair" = 0, and edit-add-task-run-once-default fails its
validate criterion with ten "Rule name ... is not unique" errors. The same
error is the single red task on PR #2939's smoke run.

Renumber each kind sequentially in document order (Entry Rule 1..N, Complete
Rule 1..N, Exit Rule 1..N) — the FE's own numbering scheme — via text-level
substitution so formatting is untouched: 64 displayName lines change, nothing
else. No checker asserts on these names (grep of diagnose/, edit/,
sla_response/, _shared/ is empty).

Verified on uip 1.202.0: the seven fixtures that should be clean are Valid;
diagnose/undeclared_gate_vars keeps exactly its two intended
"Variable ... does not exist" errors; diagnose/fix_broken_caseplan is untouched
and keeps its intended migration error.

---
Test-only. Recovers 4 case tasks in the nightly (`diagnose-extract-alias-collision`, `diagnose-formal-arg-slot-ids`, `diagnose-missing-task-entry-rule`, `edit-add-task-run-once-default`) and the single red smoke task on #2939.

### DIFF
--- a/tests/tasks/uipath-maestro-case/diagnose/extract_alias_collision/fixtures/LinearThreeStages/LinearThreeStages/caseplan.json
+++ b/tests/tasks/uipath-maestro-case/diagnose/extract_alias_collision/fixtures/LinearThreeStages/LinearThreeStages/caseplan.json
@@ -174,7 +174,7 @@
         "exitConditions": [
           {
             "id": "Condition_Fy9pWb",
-            "displayName": "Complete Rule 1",
+            "displayName": "Complete Rule 2",
             "type": "exit-only",
             "marksStageComplete": true,
             "rules": [
@@ -235,7 +235,7 @@
         "entryConditions": [
           {
             "id": "Condition_Bx5rNm",
-            "displayName": "Entry Rule 1",
+            "displayName": "Entry Rule 2",
             "isInterrupting": false,
             "rules": [
               [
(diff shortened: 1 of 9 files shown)


charlesliu9 (approved):

roberts-cliff:
Good catch on `fix_broken_caseplan` — the duplicates were masked in my check because the fixture fails on its intended V20→V21 migration error before uniqueness is evaluated. Renumbered it the same way in the follow-up commit; its two deliberate defects are untouched and it now stays at exactly two once the agent clears the migration.

PR #3007, opened 2026-09-02
perf(uipath-api-workflow): whole-bundle token reduction −27% (SkillReducer; body −55%)

CarlesUIPath:
## Summary

Whole-bundle token optimization of the **`uipath-api-workflow`** skill via the SkillReducer pipeline (arXiv 2603.29919 + `/optimize-skill` compaction). The skill behaves identically to the baseline; only its token footprint shrinks. All `skill-flavor:*` composition markers are preserved (composer contract test + 50 flavor tests green), so the Studio Web flavor build is unaffected.

> **Rebased on `main`, now covers the base64 feature.** Since this PR first opened, `main` merged the **files-as-`JobAttachment` / base64 activities** feature (#2852, ~8.0k tokens, unoptimized). This PR now (a) merges that feature and (b) optimizes it too — scoped strictly to the new base64 content, leaving the already-validated shared trims and @rares-baesu-uipath's marker-scope fixes intact. Upstream non-optimization fixes on `main` are preserved: @rares-baesu-uipath's marker-scope corrections and #3029's packaging/run-consent clarifications (rule 21 gates `run`, not `pack`/`publish`). The headline bundle % is a little lower than the original −30% **only because the baseline grew** by that feature; the candidate is smaller in absolute terms than ever.

### Static token reduction (tiktoken cl100k, vs current `main`)

| metric | `main` (incl. base64) | candidate | Δ |
|---|---:|---:|---:|
| resident (name + desc) | 290 | 290 | 0% |
| body (always-loaded) | 13,721 | 6,144 | **−55.2%** |
| references (Σ) | 70,514 | 55,065 | **−21.9%** |
| **raw bundle** | **84,525** | **61,499** | **−27.2%** |

Biggest ref cut: `connector-activity-discovery` −67% (15,598→5,094). The base64 re-optimization this cycle is scoped to the new feature only: **rule 23 deduped 696→544** (detail relocated to its reference, mistake-preventers + `file-run-cli` marker kept) + Phase-1 base64 table folded into prose, and **`files-and-base64.md` 2,786→2,427 (−359)**. The base64 additions in `cli-reference` / `expressions-and-context` / `task-types` / `troubleshooting` are left untouched — no task in the suite exercises them (a coverage gap, not an optimizer limit). Description unchanged.

### Dynamic per-task cost (measured, control vs candidate, median/task)

| model | tokens/task | USD/task | wall-clock/task |
|---|---:|---:|---:|
| gpt-5.6-luna | **−18.3%** | **−12.7%** | +1.7% |
| gpt-5.6-terra | **−25.1%** | **−20.7%** | +0.9% |
| claude-sonnet-5 | **−9.9%** | **−14.4%** | −0.9% |

Wall-clock is flat across models — task runtime dominates the agentic loop, so the win is resident/context budget, not per-task latency. (Dynamic figures measured on the shared-content candidate; the base64 delta only adds to the resident/context saving.)

## Validation

**Shared-content candidate — full 34-task suite (local docker + ADO combined):** validated across **luna + terra + Sonnet-5**, each task's verdict taken across both environments (local docker has a known env confound — ~13 tasks fail all three models locally yet pass on ADO; the two environments cover each other's blind spots).

| model | candidate (local + ADO) | v000 nightly (reference) |
|---|---:|---:|
| claude-sonnet-5 | **33 / 34** | 31 / 34 |
| gpt-5.6-terra | **32 / 34** | 33 / 34 |
| gpt-5.6-luna (local, symmetric) | == control, identical pass-sets | — |

**Base64 re-optimization — new-feature gate (docker, 2 seeds each):** the three base64 tasks pass cleanly with the optimized skill.

| task | tier | luna | terra |
|---|---|---|---|
| `base64-author-roundtrip` | integration | ✅ 2/2 | ✅ 2/2 |
| `base64-run-roundtrip` | e2e (signed-in) | ✅ 2/2 | ✅ 2/2 |
| `base64-diagnose-helper-namespace` | integration | ✅ 2/2 | ✅ 2/2 |

**ADO sanity — all api-workflow tasks, candidate vs same-day nightly control (2026-09-09_04-18):** 0 regressions on either model, task-for-task.

| model | candidate | nightly control | regressions |
|---|---:|---:|---:|
| gpt-5.6-terra | **37 / 37** | 37 / 37 | **none** |
| claude-sonnet-5 | **36 / 37** | 36 / 37 | **none** |

Terra is at full parity (37/37). Sonnet-5 has identical pass-sets to its control; its single miss — `diagnose-cloud-job-fault` — **also errors on the control nightly** (symmetric, so not attributable to this PR): an `agent_timeout` (600s single-turn) on both arms, while Terra passes that same task 1.0 — the task itself is healthy.

## `diagnose-cloud-job-fault` — resolved

Previously this task **errored during pre-run** because the seed deployed a constant package identity that collided in the shared tenant feed. **Fixed on `main` by #3133** (unique per-run solution name), merged here — it now deploys a unique identity and **passes** (Terra 37/37, score 1.0). On the Sonnet-5 arm it hit an `agent_timeout`, but the **same-day nightly control errors on it too**, so it is symmetric and not a regression from this PR.

## Report

Full validation writeup (static reduction, env-confound proof, per-task resolution, dynamic cost, base64 re-optimization): Confluence **"uipath-api-workflow — whole-bundle token optimization (SkillReducer pipeline)"** (page 91081277489).


skills/uipath-api-workflow/SKILL.md
```diff
@@ -7,358 +7,229 @@ allowed-tools: Bash, Read, Write, Edit, Glob, Grep
...
+
 <!--skill-flavor:response-roundtrip-validation:start-->
-    - **On-disk is authoritative.** Even with the single-expression workaround, every StudioWeb designer save can re-trigger normalization passes that may corrupt the Response shape. After any designer roundtrip, re-validate with `uip api-workflow run --no-auth` and re-apply the workaround if needed. Until the designer fix ships, treat the file on disk as truth, not what the designer renders.
+15. **Response.** `markJobAsFailed` is a sibling of `response`. Always use `then: "end"`; `then: "exit"` is for branches/loops. Object responses use one expression, e.g. `"${{ key: $context.variables.value }}"`, not independently interpolated fields. Single values may use `"${$context.outputs.Activity}"` or `"${'done'}"`. `${ { ... } }` and `${{ ... }}` are both valid; stay consistent. After Studio Web saves, treat disk as authoritative and rerun `uip api-workflow run --no-auth` after reapplying needed workarounds.
```
rares-baesu-uipath: This marker used to wrap only the last bullet of rule 15 ("On-disk is authoritative"). Now it wraps the whole rule, but the Studio Web override still only replaces that one sentence, so rule 15 disappears from the studioweb build. Please move the marker back around just the trailing "After Studio Web saves, treat disk as authoritative…" sentence.

skills/uipath-api-workflow/SKILL.md
```diff
@@ -7,358 +7,229 @@ allowed-tools: Bash, Read, Write, Edit, Glob, Grep
...
+20. **Static validation.** Run `uip api-workflow validate <Workflow.json> --output json` as the last autonomous command in every author/edit cycle. On `Result: "Failure"`, read `Instructions`, fix the activity at its JSON path, and repeat until `Data.Status: "Valid"`. Prioritize semantic-tail errors over duplicate `oneOf` noise. Validation catches malformed JSON, unknown types, required-field errors, bad evaluate settings, duplicate/empty variables, and empty task lists; not broken connections, wrong resource IDs, runtime expression errors, unwrapped literals, or multi-key Assign sets.
```
rares-baesu-uipath: Same issue as rule 15: this marker used to wrap only the "What validate catches / does NOT catch" paragraph. Now it wraps all of rule 20, so the studioweb build loses the rule entirely (rule 21 there still says "once static validation passes" with nothing to refer to). Shrink it back to the catches / does-not-catch clause.

skills/uipath-api-workflow/SKILL.md
```diff
@@ -7,358 +7,229 @@ allowed-tools: Bash, Read, Write, Edit, Glob, Grep
...
+> **TDD gate — read first (rule 22).** For create/edit requests, check `<project>/evals/` beside `Workflow.json`, never at workspace or solution root. If absent, Evaluations is off: do not ask about tests or loop mode; author normally. If present, stop before authoring and ask whether existing rows remain/change or empty tests should be added, and whether to run/retry until all pass or author once. After both answers, declare schemas, write/update evals first, then author. A request not to ask skips these questions, not runtime consent; run rows only when requested. Later behavior changes require asking whether affected expectations should change.
 
 ## When to Use This Skill
 
-- User wants to **create or edit** an API workflow JSON file
+Use for API workflow JSON creation/editing, local runs, validation, build/packaging, publishing, operating published workflows, and activities including Sequence, Assign, JavaScript, If, ForEach, DoWhile, Break, TryCatch, Wait, Response, HTTP Request, and connector activities. Use the connector and testing references for Studio Web connector workflows and project `evals/` layouts.
```
rares-baesu-uipath: "local runs, validation, build/packaging, publishing, operating published workflows" used to live inside the `surface-lifecycle-scope` / `surface-operations-scope` blocks below, which Studio Web overrides. Moving it into this paragraph means the studioweb build now shows the CLI wording *and* the override bullets. Please keep that text inside the blocks.

skills/uipath-api-workflow/references/connector-activity-discovery.md
```diff
@@ -141,737 +74,189 @@ uip api-workflow registry stub <activity-type-id> \
...
 
-**Non-Solutions projects (a `project.json` at the top level, no `Solution/` wrapper).** Skip this step — the file structure has no `Solution/resources/` tree to write into, and the properties-panel error doesn't fire because StudioWeb knows there's no Solution to consult.
+If offline hand-authoring is unavoidable, write `Solution/resources/solution_folder/connection/<connector-key>/<connection-name>.json`, starting from [assets/templates/solution-connection-resource-template.json](../assets/templates/solution-connection-resource-template.json). Use the exact connection `Name`, `ConnectorKey`, `ConnectorName`, connector version from `essentialConfiguration.connectorVersion` (or `"1.0.0"` if unparseable), pinged UUID as `resource.key`, and the existing resource folder's `folders[0].fullyQualifiedName` (default `"solution_folder"`). The key must equal both workflow connection IDs and the binding key. Write one file per unique UUID. Do not hand-author `bindings_v2.json` or debug overwrites when the CLI commands are available.
 <!--skill-flavor:solution-metadata:end-->
+<!--skill-flavor:worked-example-solution-metadata:start-->
```
rares-baesu-uipath: This marker was inside the (now deleted) Outlook worked example's bash block. The override text is a bash comment, so in the studioweb build it composes as a top-level heading: `# 5. Use Studio Web's host resource capability…`. Drop the pair here and the block in the override.

skills/uipath-api-workflow/references/connector-activity-discovery.md
```diff
@@ -141,737 +74,189 @@ uip api-workflow registry stub <activity-type-id> \
...
-
-Examples:
-
-```javascript
-// In an If when:
-"when": "${$context.outputs.getNewestEmail_1?.content?.subject?.length > 15}"
+## Response shape and field rules
 
+Both Http and IntSvc outputs are envelopes: `{ statusCode, statusText, headers, ok, request, content, vendorProcessingTimeMs }`. Read payloads under `.content`; IntSvc list payloads may be arrays directly, not `.content.value[]`. Inspect `optionalConfiguration.fieldsContainer.outputJsonSchema`: `type: "object"` means a single item; `type: "array"` means a list. Local CLI may return `content` as a JSON string while cloud returns a parsed value; normalize defensively in scripts. Use optional chaining and log the full output once if the expected shape is absent.
 <!--skill-flavor:runtime-content-normalization-comment:start-->
```
rares-baesu-uipath: This marker was inside a deleted JS code fence. Its override is a `// …` comment, which now lands as a bare line of JavaScript in the middle of the prose in the studioweb build. Drop the pair + override block, or restore a minimal fence.

skills/uipath-api-workflow/references/connector-activity-discovery.md
```diff
@@ -141,737 +74,189 @@ uip api-workflow registry stub <activity-type-id> \
...
+- **Do NOT skip the Solution catalogue sync in Solutions-mode projects.** Two files MUST exist: the catalogue resource (`Solution/resources/solution_folder/connection/<connector-key>/<name>.json`) AND the per-user debug overwrites (`Solution/userProfile/<guid>/debug_overwrites.json`). Without both, the properties panel flags the activity with "to debug this resource, select a connection for it from the resource definition page" and clicking the activity nulls `with.connectionId`. Run `uip api-workflow bindings sync --workflow <Workflow.json>` followed by `uip solution resources refresh --solution-folder <path>` to write both. See [Step 5](#step-5--solutions-mode-intsvc-connection-synchronization).
 <!--skill-flavor:solution-metadata-antipattern:end-->
+
+<!--skill-flavor:http-example-execution-proof:start-->
+Verified end-to-end: `uip api-workflow run --no-auth` on the resulting workflow returns `statusCode: 200`, `content.fact: "..."`. StudioWeb's designer renders the activity as the unified HTTP Request card. See [../assets/templates/connector-call-example.json](../assets/templates/connector-call-example.json) for a complete stub-generated workflow.
```
rares-baesu-uipath: "on the resulting workflow" — the HTTP worked example this referred to was removed, and the block now sits after the anti-patterns with no heading. Either move it under the Http kind section with a one-line lead-in, or drop it.

skills/uipath-api-workflow/references/workflow-file-format.md
```diff
@@ -100,34 +61,19 @@ Variables live at `document.metadata.variables.schema.document` as a JSON Schema
...
-    "format": "json",
-    "document": {
-      "type": "object",
-      "properties": { "userId": { "type": "string" } },
-      "title": "Inputs"
-    }
-  }
-}
-```
+Declare workflow I/O at the root with JSON Schema. Outputs are read from the final `Response` task's `response` value.
```
rares-baesu-uipath: "Outputs are read from the final `Response` task's `response` value" is repeated inside the `runtime-input-output` block at L81, which Studio Web overrides. Keeping it here too leaks the CLI wording into the studioweb build. Remove it from this line.

rares-baesu-uipath (commented):
Inline notes on the flavor markers (details in my summary comment above).

skills/uipath-api-workflow/references/testing-and-evals.md
```diff
@@ -57,168 +52,74 @@ Because the tests are files, the agent can author them, run them, fix the workfl
...
-- **Inside Studio Web** (no local CLI): rows run through the host instead — ask the user to press **Run** in the Evaluations panel (its verdicts are authoritative), or, with explicit consent (rule 21), invoke `RunProject` once per row with that row's `inputs`; the host result is the raw output. Derive `expectedOutput` from it or from `output.schema`, exactly as above.
-- `--no-auth` is fine for control-flow-only workflows and `Http` activities using `connectionId: "ImplicitConnection"`. **IntSvc (vendor connector) activities need auth** and have real side effects (emails sent, tickets created) — rules 20–21 in SKILL.md apply: never run autonomously, never with auth without an explicit "yes".
+Compare that raw object with `expectedOutput` under the declared evaluator's `targetOutputKey`. A non-`Success` run is an `ERROR`, not a score. Derive expected output from `output.schema`; declare the schema first if `output: null`. The `Response` must emit exactly those keys and casing.
 
----
+Run every row after each change with one outside-project comparator command:
 
-## 3. The test-until-green loop (interactive protocol)
+```bash
+python3 /tmp/score-evals.py <PROJECT>
```
rares-baesu-uipath: This file no longer exists anywhere in the doc — the ~60-line script that used to be right here was removed. An agent reading this will try to run a file it never got. Either put the script back (and trim only the prose around it), or change this line to say something like "write a comparator script outside the project (see below), then run it".

rares-baesu-uipath (commented):
Notes on the removed `score-evals.py` comparator.

skills/uipath-api-workflow/references/testing-and-evals.md
```diff
@@ -57,168 +52,74 @@ Because the tests are files, the agent can author them, run them, fix the workfl
...
+   > “Do you want me to work in **loop mode** — run the tests and re-try (fix the workflow, re-run) until every case passes? Or author once and you verify manually?”
 
----
+   End the turn. Ask on the first authoring prompt, whether creating or editing. Before the first run, ensure the workflow is in an `init`-scaffolded project (SKILL.md rules 19–19a).
+4. **Author tests first, then run.** Once answered:
+   1. Declare `input.schema` and `output.schema` in `Workflow.json`, including exact property casing.
+   2. Write or update the eval set from those schemas.
+   3. Author the workflow so `Response` emits exactly `output.schema`.
+   4. In loop mode, run and score every row in one command: `python3 /tmp/score-evals.py <PROJECT>`. Loop mode consents to repeated `--no-auth` runs; authenticated or side-effecting connector runs still require their own “yes”.
```
rares-baesu-uipath: Same as above: `/tmp/score-evals.py` doesn't exist anymore. Either restore the script or reword this step to "run the comparator you wrote".

skills/uipath-api-workflow/references/testing-and-evals.md
```diff
@@ -57,168 +52,74 @@ Because the tests are files, the agent can author them, run them, fix the workfl
...
+Write the comparator outside the project, never under `evals/`. It must run `uip api-workflow run` once per row with `--input-arguments`, `--no-auth`, `--output json`, `--log-level debug`, and `--log-file`; require `Result == "Success"`; extract the last `Response task evaluated successfully` response; wrap non-object outputs as `{ "result": raw }`; match evaluator refs by file base name; require `uipath-exact-match`; require `evaluationCriterias[ref].expectedOutput`; honor `targetOutputKey` values `"*"`, dotted strings, and arrays; use strict JSON deep equality (key order ignored, shapes, casing, types, and values equal); print `[PASS]`, `[FAIL]`, or `[ERROR]` per row; and exit 0 only when at least one row exists and every verdict is `PASS`.
```
rares-baesu-uipath: Since the agent now has to write this script itself from this paragraph, it needs to be a bit more explicit. The old script did a few things this description doesn't mention, and each one was there because of a real failure:

- `true` and `1` are **not** equal (compare types, not just values — JSON bools vs numbers).
- Give each `uip api-workflow run` a timeout (the script used 120s) so one hanging row doesn't stall the loop.
- If the CLI output isn't valid JSON, report the row as `ERROR` and print the first lines of stdout, don't crash.
- If the evaluator ref doesn't match any file, say so and hint that refs are file base names.
- Print one line per row with the row name, `PASS`/`FAIL`/`ERROR`, and on failure both `expected=` and `actual=`.
- Exit 0 only if every row passed.

A plain bulleted list like that would be enough. Right now a looser home-made comparator will say "all green" while the strict `uipath-exact-match` grader fails the same rows.

skills/uipath-api-workflow/references/control-flow-patterns.md
```diff
@@ -1,265 +1,201 @@
...
-└─ If_2#Wrapper                ← _2 because If_1 was already used inside the loop
-    ├─ If_2
-    ├─ If_2#Then
-    └─ If_2#Else
+│  └─ For_Each_1#Body
+│     ├─ If_1#Wrapper
+│     │  ├─ If_1
+│     │  ├─ If_1#Then
+│     │  └─ If_1#Else
+│     │  └─ Try_Catch_1
```
rares-baesu-uipath: This diagram is wrong now. `Try_Catch_1` is drawn one level too deep — as a child of `If_1#Wrapper`, next to `If_1#Else`. An If wrapper only ever holds three things (`If_1`, `#Then`, `#Else`), so the picture teaches an invalid shape.

In the original, `Try_Catch_1` sat directly under `For_Each_1#Body`, as a sibling of `If_1#Wrapper`:

```text
│       ├─ If_1#Wrapper        ← inside the loop body, but uses _1 because no other If yet
│       │   ├─ If_1
│       │   ├─ If_1#Then
│       │   └─ If_1#Else
│       └─ Try_Catch_1         ← _1 because no other TryCatch yet
└─ If_2#Wrapper                ← _2 because If_1 was already used inside the loop
```

Please restore that indentation, and ideally the three short `←` notes too — they're the only thing that explains *why* the outer If is `_2`, which is what this section exists to teach.

rares-baesu-uipath (commented):
Key-numbering diagram.

rares-baesu-uipath (approved):

PR #3009, opened 2026-09-02
test: add num_retries to checker_context api_route

akshaylive:
## Summary
- Add `num_retries: 5` under `checker_context.api_route.params` in the five experiment files that use the litellm judge route (activation, default, nightly, smoke, smoke-windows), so transient judge API failures get retried instead of failing the row.

## Test plan
- [ ] Confirm YAML parses correctly with existing experiment loader

### DIFF
--- a/tests/experiments/activation.yaml
+++ b/tests/experiments/activation.yaml
@@ -56,6 +56,7 @@ defaults:
       model: gpt-5.6-luna
       params:
         api_version: "2024-05-01"
+        num_retries: 5
       env_params:
         api_base: CODEX_BASE_URL
         api_key: CODEX_API_KEY
--- a/tests/experiments/default.yaml
+++ b/tests/experiments/default.yaml
@@ -29,6 +29,7 @@ defaults:
       model: gpt-5.6-luna
       params:
         api_version: "2024-05-01"
+        num_retries: 5
       env_params:
         api_base: CODEX_BASE_URL
         api_key: CODEX_API_KEY
--- a/tests/experiments/nightly.yaml
+++ b/tests/experiments/nightly.yaml
@@ -47,6 +47,7 @@ defaults:
       model: gpt-5.6-luna
       params:
         api_version: "2024-05-01"
+        num_retries: 5
       env_params:
         api_base: CODEX_BASE_URL
         api_key: CODEX_API_KEY
--- a/tests/experiments/smoke-windows.yaml
+++ b/tests/experiments/smoke-windows.yaml
@@ -31,6 +31,7 @@ defaults:
       model: gpt-5.6-luna
       params:
         api_version: "2024-05-01"
+        num_retries: 5
       env_params:
         api_base: CODEX_BASE_URL
         api_key: CODEX_API_KEY
--- a/tests/experiments/smoke.yaml
+++ b/tests/experiments/smoke.yaml
@@ -47,6 +47,7 @@ defaults:
       model: gpt-5.6-luna
       params:
         api_version: "2024-05-01"
+        num_retries: 5
       env_params:
         api_base: CODEX_BASE_URL
         api_key: CODEX_API_KEY


tests/experiments/activation.yaml
```diff
@@ -56,6 +56,7 @@ defaults:
       model: gpt-5.6-luna
       params:
         api_version: "2024-05-01"
+        num_retries: 5
```
uipreliga, line 59: Shouldn't it be by default something like 3 anyway? 
Also, API version should be some kind of default (i.e. not specified there)

uipreliga (approved):
Look OK, but I would think that we should have API version and num_retries specifed by default in the coder eval codebase, and here likely not enforced.

tests/experiments/activation.yaml
akshaylive: Litellm API is pretty flexible and works with a wide range of APIs. I decided to let the litellm have no defaults (and let the user configure it). Either way this does not affect this PR so I merged it. Will sync up with you offline to decide if a coder_eval update is needed..

uipreliga (commented):
# Review: UiPath skills — pr:3009 (5 files) axis:1,3,6,7,8

PR #3009 'test: add num_retries to checker_context api_route' by akshaylive · akshaya/judge_retry -> main · OPEN

**Change class:** Config tuning — adds num_retries: 5 to checker_context.api_route.params in 5 experiment YAMLs; no code or schema changes.

The change is small, structurally consistent, and clean on correctness and error handling (9.8/10 overall, zero Critical/High), but its one real risk is a contract risk rather than a logic one: `num_retries: 5` is an aliased, unvalidated litellm passthrough that expands one judge criterion from 3 to 11 provider calls (~22 min worst case) inside a 900s `task_timeout` — able to silently turn gradeable rows into not-evaluated results — and it is documented nowhere, so bound the attempt budget and sync `tests/README.md` before merge and the PR is otherwise ready.

## Summary

| 1. Correctness & Logic | 10 / 10 | 0 | 0 | 0 | 0 | — |
| 3. Types & Contracts | 9.4 / 10 | 0 | 0 | 1 | 1 | num_retries: 5 is an aliased knob — it sets the provider SDK's max_retries AND litellm's own tenacity retry loop, so one llm_judge criterion goes from 3 to 11 HTTP calls (~22 min worst case) inside a 900s task_timeout |
| 6. Architecture & Design | 9.9 / 10 | 0 | 0 | 0 | 1 | Sixth knob added to a 5-way copy-pasted checker_context block with no sharing mechanism; mirror set has already drifted once |
| 7. Error Handling & Resilience | 10 / 10 | 0 | 0 | 0 | 0 | — |
| 8. Interface, Docs & Compatibility | 9.5 / 10 | 0 | 0 | 1 | 0 | tests/README.md's authoritative api_route sample is now a strict subset of the real config — num_retries: 5 is documented nowhere, so a copied sample silently loses the retry behavior |

**Overall Score**: 9.8 / 10 · **Weakest Axis**: Types & Contracts at 9.4 / 10
**Totals**: 🔴 0 · 🟠 0 · 🟡 2 · 🔵 2 across 5 axes reviewed.

None.

1. **[Axis 3] num_retries: 5 is an aliased knob — it sets the provider SDK's max_retries AND litellm's own tenacity retry loop, so one llm_judge criterion goes from 3 to 11 HTTP calls (~22 min worst case) inside a 900s task_timeout** (`tests/experiments/smoke.yaml:tests/experiments/smoke.yaml:50`) — Choose the value against total attempts, not the label — `num_retries: 2` yields 5 provider calls (verified), `num_retries: 5` yields 11. Because `params` is applied after the judge's own kwargs (`judge_litellm.py:125` sets `"timeout": timeout_seconds`, and line 135 then does `kwargs.update(route.params)`), you can bound each attempt from the same map, e.g. add `timeout: 30` next to `num_retries`. Note `env_params` applies last (line 136) but sets only `api_base`/`api_key`, so neither key is clobbered. Add a one-line comment recording the intended attempt budget, since `num_retries` is aliased and the YAML value is not the call count. Caveat: the counts can shift with a litellm minor bump — coder_eval's `uv.lock` pins only `litellm ... specifier = ">=1.95.0,<2.0.0"`, and `tests/Makefile:18,22` installs `coder-eval[litellm]` from git main, not from the `0.11.5` in `tests/.coder-eval-version`. Raise to High if a live run confirms the retry chain exhausting `task_timeout` — no eval was executed, so that last hop is read, not observed.
2. **[Axis 8] tests/README.md's authoritative api_route sample is now a strict subset of the real config — num_retries: 5 is documented nowhere, so a copied sample silently loses the retry behavior** (`tests/experiments/smoke.yaml:50`) — Add `num_retries: 5` to the fenced sample's `params:` block in tests/README.md (line 250-251) and one sentence of prose alongside the existing explanations at L243/L257/L261-262 saying what it does and why 5 — e.g. "`num_retries` retries the judge call on transient litellm failures instead of failing the row outright; 5 is the experiment default." One line each, per .claude/rules/token-optimization.md ("a one-line addition, not a paragraph").

Verified, not estimated:
- New real block, `git show pr-3009:tests/experiments/smoke.yaml | grep -n` → `48:      params:` / `49:        api_version: "2024-05-01"` / `50:        num_retries: 5`. Identical addition in activation.yaml, default.yaml, nightly.yaml, smoke-windows.yaml.
- Documented sample at PR HEAD, `git show pr-3009:tests/README.md | awk 'NR>=245&&NR<=255'` → `250:     params:` / `251:       api_version: "2024-05-01"` / `252:     env_params:`. The `params` map documented there is now a strict subset of the shipped one.
- `git show pr-3009:tests/README.md | grep -c num_retries` → `0`. Repo-wide, `num_retries` occurs only in the 5 changed YAMLs, so no other doc carries the key either.

Why this matters here specifically: tests/README.md is the repo's authoritative test-framework reference (CLAUDE.md and .claude/rules/test-writing.md both route authors to it before touching tests/), and L243 introduces this exact fenced block as how "This repo's experiment defaults (`default`/`smoke`/`smoke-windows`/`nightly`/`activation`) route `llm_judge` through `litellm`" — i.e. the sample claims to describe precisely the 5 files this PR changed. The README's own prose at L257 also instructs a reader to hand-write an `api_route` override on a task ("if one is added, override `checker_context.api_route` back to `bedrock`/`direct` on that specific task"), so hand-authoring from this sample is a documented workflow. An author copying it gets a judge route without retries — a silent behavioral difference from every shipped experiment, in exactly the failure mode (transient judge failures) the commit message says it is fixing.

Severity rationale: Axis 8 🟡 anchor "a stale example". Capped at Medium, not High: no documented command fails, nothing breaks for existing consumers, and the sample was already not a byte-exact mirror before this PR (L249 reads `model: azure/gpt-5.6-luna` while all 5 real files read `model: gpt-5.6-luna` — pre-existing divergence, not filed, out of scope). This PR added a second divergence to that sample, which is what makes it filed rather than pre-existing debt.

The commit message does record the why (`git log -1 --format=%B pr-3009` → "Ensures the litellm judge route retries on transient failures instead of failing a row outright"), so .claude/rules/pr-review.md's "describe the why not just the what" is satisfied; the gap is that the rationale and the value 5 live only in git history, not anywhere a future reader tuning the CI budget will look. The README sentence above closes both halves with one edit.

Static guard that would prevent recurrence: a repo-local check (scripts/ pattern, dependency-free Node .mjs like the existing validators) that extracts each fenced ```yaml sample under tests/README.md's "Checker Context" heading and asserts its `checker_context.api_route.params` key set equals the key set in tests/experiments/*.yaml that define the block. Statically decidable — both sides are YAML in-repo, no runtime needed.

## Nits

1. **[Axis 3] checker_context.api_route.params is an unvalidated dict[str, Any] passthrough — a misspelled retry key is a verified silent no-op across all 5 experiments** (`tests/experiments/default.yaml:32`) — `params` is typed `params: dict[str, Any] | None = None` (~/src/coder_eval/src/coder_eval/models/routing.py:161, checkout at `v0.11.5-2-gd679ff2`) and its only consumer is `kwargs.update(route.params)` (evaluation/judge_litellm.py:134-135) — coder_eval applies no key allowlist, and nothing in this repo validates experiment `params` either (`grep -rn 'num_retries' .` returns only the 5 changed YAMLs).

Verified with the same stub-server harness: the misspelled key `num_retires=5` returned `requests=3 elapsed=1.5s` — byte-identical to the no-retry baseline, with no warning or error — while `num_retries=5` returned `requests=11 ... LiteLLM Retried: 5 times`. So a typo, or a future litellm rename of this kwarg, silently reverts all five experiments (activation.yaml:59, default.yaml:32, nightly.yaml:50, smoke.yaml:50, smoke-windows.yaml:34) to today's retry behavior with no signal anywhere: `params` is not recorded in run artifacts either — orchestrator.py:1593-1601 records only `litellm_base_url_host` and `litellm_model` for a LiteLLMRoute.

Fix (cheap): add a one-line comment on the key in each experiment naming the contract owner and the expected effect, e.g. `# forwarded verbatim to litellm.acompletion(num_retries=...) — see tests/README.md#checker-context`, so a reader can tell a live knob from a dead one. Keep the value unquoted: int is the correct type — passing a string fails (`num_retries="5str"` → `BadRequestError ... max retries must be an int`), unlike the sibling `api_version: "2024-05-01"`, which is correctly quoted.
2. **[Axis 6] Sixth knob added to a 5-way copy-pasted checker_context block with no sharing mechanism; mirror set has already drifted once** (`tests/experiments/default.yaml:32`) — Nit / accept-as-is is defensible. The change is structurally consistent with the established pattern and I verified all five blocks are now byte-identical (`for f in activation default nightly smoke smoke-windows; do git show pr-3009:tests/experiments/$f.yaml | sed -n '/checker_context/,/^[a-z]/p'; done` -> identical 8-line blocks; `grep -n num_retries tests/experiments/*.yaml` -> default.yaml:32, activation.yaml:59, nightly.yaml:50, smoke.yaml:50, smoke-windows.yaml:34). No abstraction was bypassed: `grep -n '&[a-zA-Z_]\|<<:\|extends\|include' tests/experiments/*.yaml` finds zero YAML anchors, zero `extends`, and no base experiment file, and the repo's only documented inheritance is task->experiment (`.claude/rules/test-writing.md` Must-Do 3; `tests/README.md:417` "No `agent:` block - inherits everything from `experiments/default.yaml`"). What this PR does worsen, marginally, is the per-edit cost of that block: it grows from 7 to 8 duplicated lines x 5 files, and the block now has a documented sixth mirror in `tests/README.md` (fenced sample at L246-252). Drift here is demonstrated, not hypothetical - the block's own originating commit `636db8130` (#2864) already left `tests/README.md:249` reading `model: azure/gpt-5.6-luna` while all five YAMLs read `model: gpt-5.6-luna` (verified: `grep -n 'gpt-5.6-luna' tests/README.md tests/experiments/*.yaml`). Cheapest proportionate fix: leave the five-way copy as is (a cross-file include would be over-engineering for one line and coder_eval support for it is unverified), but treat `tests/experiments/default.yaml` as the canonical copy and add the one-line rationale there in the style the neighbouring knobs in this same file already use (`turn_timeout: 900` carries lines 8-10 "A single turn can chain several cold-runner `uip rpa` calls..."; `post_run` carries lines 37-39), so the next person editing the block knows why `5` was chosen and that six places must move together. Raise above nit only if coder_eval gains an experiment include/extends mechanism that this bypassed - unverified, and no such key exists in the repo today. n/a

## What's Missing

**Parallel paths & mirrors:**
- 🔵 **Sibling judge route's retry budget was not reconciled — 4 attempts vs the new 11.** The framework's other judge implementation hard-codes its own budget: `_JUDGE_RETRY = RetryConfig(max_retries=3, initial_delay=2.0, backoff_multiplier=2.0)` (`~/src/coder_eval/src/coder_eval/evaluation/judge_bedrock.py:42`) with `attempts = _JUDGE_RETRY.max_retries + 1` (line 99) and `_is_retryable_status` = 429 or >=5xx (line 45) — i.e. 4 total attempts, jittered exponential backoff, retryable statuses only. `num_retries: 5` puts the litellm route at 11 attempts (verified in Axis 3), ~2.75x the budget the mirror uses, with no note that the two now differ. Checked and NOT filed as a missed edit: the four other experiment YAMLs (`tests/experiments/{flow-v2-preview,same-ground-headtohead,sherif-skill-comparison,skill-comparison-template}.yaml`) carry no `checker_context` block at all (`grep -rln api_route tests/` → only Makefile, README and the 5 changed files), so they grade `llm_judge` through the built-in bedrock judge, which already retries 4x — no row fails outright there, and `num_retries` is not a bedrock parameter. That matters because `flow-v2-preview.yaml` covers Maestro tasks and 13 task YAMLs under `tests/tasks/uipath-maestro-*` use `llm_judge` (`grep -rl llm_judge tests/tasks/uipath-maestro-* | wc -l` → 13). The omission is only the unexplained budget divergence, not a missing YAML edit — pick the litellm value against the bedrock precedent (`num_retries: 2` → 5 attempts, closest to the mirror's 4) or say in a comment why the litellm gateway warrants ~3x more. _(trigger: tests/experiments/default.yaml)_ _(restates: Axis 3: num_retries: 5 is an aliased knob)_

**Tests:**
- 🟡 **No check asserts the five copied `checker_context` blocks stay identical, and this PR grew the copy to 8 lines x 5 files.** Verified there is no such guard: `grep -rn api_route tests/ .github/ scripts/` matches only `tests/Makefile:26`, `tests/README.md`, and the five experiment YAMLs; no workflow lints `tests/experiments/*.yaml` (no yamllint config in the repo root), and the two scripts that do parse experiment files are narrow — `scripts/prune-task-defaults.py` reads only `run_limits` from `default.yaml` (`DEFAULTS_PATH` at line 24), and `scripts/check-task-driver.py` only mentions `nightly.yaml` in its docstring. The repo already has the exact pattern for a gate like this (`scripts/check-task-driver.py`, `scripts/check-skill-status.py`, `scripts/check-cli-verbs.py`, all CI-wired), and the drift this would catch is demonstrated rather than hypothetical (README sample already diverges on `model:` and now on `params`). A ~20-line equality check over the `checker_context.api_route` key sets of every experiment that defines the block — plus the README's fenced sample — is statically decidable and would have caught both divergences. _(trigger: tests/experiments/default.yaml)_ _(restates: Axis 6: Sixth knob added to a 5-way copy-pasted checker_context block)_
- 🔵 **The retry path this PR exists for is exercised by nothing in-repo, and a run cannot show whether it engaged.** The knob only changes behavior on a failing judge call; every passing run makes exactly one call, so the CI fan-out this PR triggers (see the rollout-impact item) validates that the config still loads, not that retries help or what they cost. Nor is the outcome recoverable after the fact: for a LiteLLM route the orchestrator records only `environment_info["litellm_base_url_host"]` and `["litellm_model"]` (`~/src/coder_eval/src/coder_eval/orchestrator.py:1596-1600`) plus `eval_routing`/`eval_model` (lines 1568-1574) — `route.params` is never recorded, and `grep -rn environment_info` over this repo's `*.py`/`*.mjs`/`*.md`/`*.yml` returns zero hits, so no repo-side report surfaces it either. Cheapest closure short of fault injection (retries live inside litellm, so this is genuinely not statically checkable): run one `llm_judge` task once against the litellm route and record the grading wall-clock in the PR, so the chosen budget has one measured data point rather than none. _(trigger: tests/experiments/smoke.yaml)_ _(restates: Axis 3: num_retries: 5 is an aliased knob)_

**Downstream consumers:**
- 🔵 **Nightly run artifacts and the blob dashboard gain no way to attribute a grading-time or judge-error change to this knob.** `CLAUDE.md` documents the nightly results contract (`coderevaltests` blob, `<run-id>/run.json`, per-task `task.json`), but the only judge-route facts written into a run are `eval_routing`/`eval_model` and, for LiteLLM, `litellm_base_url_host`/`litellm_model` (`~/src/coder_eval/src/coder_eval/orchestrator.py:1568-1574,1596-1600`) — `params` is not among them. Consequence: if the nightly's grading latency or `not_evaluated` rate moves after this merge, nothing in `run.json` distinguishes a run with the retry budget from one without, and the silent-no-op case (a typo'd key, or a litellm rename of `num_retries`) is invisible in exactly the artifact a reader would check. No in-repo consumer breaks (`grep -rn environment_info` over this repo → zero hits), which is why this is Low rather than higher; the fix is upstream-or-comment, not a repo change. _(trigger: tests/experiments/nightly.yaml)_ _(restates: Axis 3: checker_context.api_route.params is an unvalidated dict[str, Any] passthrough)_

**Docs & config:**
- 🟡 **`tests/README.md`'s `checker_context` sample was not updated and is now a strict subset of all five shipped blocks.** Re-verified in this pass: the fenced sample's `params:` at `tests/README.md:250` carries only `api_version: "2024-05-01"` (line 251), and `grep -c num_retries tests/README.md` → 0, while the prose at line 243 introduces that exact block as describing "This repo's experiment defaults (`default`/`smoke`/`smoke-windows`/`nightly`/`activation`)" — precisely the five files this PR changed. `tests/Makefile:26` was checked and is NOT stale (it names the route and points at `tests/README.md#checker-context` without enumerating `params`). Fix is one line in the sample plus one sentence of prose; see the Axis 8 finding for wording. _(trigger: tests/experiments/smoke.yaml)_ _(restates: Axis 8: tests/README.md's authoritative api_route sample is now a strict subset of the real config)_
- 🔵 **The chosen value carries no rationale comment, unlike every neighboring tuning knob in the same files.** `default.yaml` explains `turn_timeout: 900` inline (lines 8-10) and `post_run` (lines 37-39); `smoke.yaml` explains `max_turns: 40` (lines 6-7) and `task_timeout`/`turn_timeout` (lines 9-10); `activation.yaml` spends lines 8-37 on its `run_limits`. `num_retries: 5` lands bare at `default.yaml:32` / `smoke.yaml:50` / `nightly.yaml:50` / `smoke-windows.yaml:34` / `activation.yaml:59`, and because the key is aliased (the YAML number is not the call count), the next person tuning the CI budget has neither the intended attempt budget nor the reason for 5 anywhere outside the commit message. One line on the canonical copy (`default.yaml`) naming the attempt budget. _(trigger: tests/experiments/default.yaml)_ _(restates: Axis 6: Sixth knob added to a 5-way copy-pasted checker_context block)_

**Rollout impact:**
- 🟡 **Blast radius on the CI gates is unstated, and the job wall — not `task_timeout` — is the first thing a retry-inflated run hits.** Verified triggers: `tests/experiments/smoke.yaml` matches the infra-change regex `^tests/experiments/smoke\.yaml$|...` at `.github/workflows/smoke-skills.yml:149`, so this PR fans out every non-Windows smoke task (capped by `MAX_SMOKE_TASKS: ${{ vars.SMOKE_MAX_TASKS || '40' }}`, `smoke-skills.yml:53`); `^tests/(experiments|_shared)/` at `.github/workflows/smoke-rpa-skills.yml:61` fans out *every* Windows-tagged task. Job walls are `timeout-minutes: 120` (`smoke-skills.yml:323`), `90` (`smoke-rpa-skills.yml:95`), `330`/`240` in `run-coder-eval.yml:221,901`. With `TASK_PARALLELISM` defaulting to `1` (`smoke-skills.yml:463`) and `task_timeout: 900` (`smoke.yaml:11`), 40 tasks fit inside 120 minutes only while tasks finish well under their ceiling — which is exactly the margin retry-inflated grading consumes, so the visible failure is a cancelled job mid-suite (every remaining task ungraded) rather than one slow row. Two things to state in the PR: that both smoke gates run full-suite on this diff, and what the per-task grading ceiling now is. Not raised above Medium: degraded-path only, and no live run was executed to observe it. _(trigger: tests/experiments/smoke.yaml)_ _(restates: Axis 3: num_retries: 5 is an aliased knob)_
- 🔵 **The knob's effective attempt count is governed by an unpinned transitive dependency, so it can change with no repo commit.** CI does pin the framework (`tests/.coder-eval-version` → `0.11.5`, consumed at `.github/workflows/run-coder-eval.yml:149` and `activation-gate.yml:89-93`), but the retry semantics live in litellm, which coder_eval constrains only as `"litellm>=1.95.0,<2.0.0"` (`~/src/coder_eval/pyproject.toml:103`; `uv.lock` currently resolves `1.98.0`) — a range, not a pin, resolved fresh at install time. The local path floats further: `tests/Makefile:18,22` install `coder-eval[litellm] @ git+https://github.com/UiPath/coder_eval.git` (no tag), i.e. main rather than the `0.11.5` pin. So the same `num_retries: 5` can mean a different attempt count and a different worst-case grading wall-clock on two machines or two weeks apart. Nothing to fix in this diff; worth one line in the README sentence recording that the value maps to a litellm-version-dependent attempt count. _(trigger: tests/experiments/activation.yaml)_ _(restates: Axis 3: num_retries: 5 is an aliased knob)_

## Guardrails & Automation

- [repo-local-check] Add `scripts/check-experiment-config.py` with a **key allowlist + type check** for `checker_context.api_route.params`, backed by a new registry table `.claude/rules/checker-context-params.md`. Follow the two established patterns exactly: the script shape of `scripts/check-task-driver.py` (stdlib + PyYAML, module docstring carrying the rationale, `--json` mode, exit 0 clean / 1 findings, repo-relative paths via a `_rel()` helper), and the rules-table-parsed-by-script pattern of `.claude/rules/cli-renames.md` ↔ `scripts/check-cli-verbs.py:load_renames()` (pipe-delimited columns read positionally). Registry columns: `| Key | Type | Effect |`, seeded with `api_version | string | litellm/Azure API version` and `num_retries | int | forwarded to litellm.acompletion(num_retries=…); see 2n+1 note below`. The script fails when a `params` key is absent from the table (catching `num_retires`) and when a value's type disagrees with the Type column (int keys must be unquoted — `num_retries: "5"` raises `BadRequestError … max retries must be an int` only at grading time, and a quoted-by-analogy-with-`api_version` mistake is otherwise invisible). Wire it with a new `.github/workflows/experiment-config-gate.yml` copied from `task-driver-gate.yml`: no `paths:` filter (so it can be a required check), `setup-python` pinned to `'3.13'`, `pip install --quiet pyyaml`, sub-second full scan of `tests/experiments`. `tests/experiments/**` is currently covered by no script or gate in this repo — verified: it appears only in `smoke-skills.yml:149` as a trigger path. _Prevents:_ Finding 2 (Low, Axis 3) — `params` is `dict[str, Any] | None` with `kwargs.update(route.params)` as its only consumer, so a misspelled retry key is a verified silent no-op across all five experiments with no warning anywhere. Also catches the quoting hazard the finding calls out, and any future litellm rename of the kwarg (the registry becomes the place that rename is recorded).
- [repo-local-check] In the same `scripts/check-experiment-config.py`, add a **retry-budget arithmetic assertion**. When `checker_context.api_route.params.num_retries` is present, require (a) a sibling `timeout` in the same `params` map, and (b) `(2 * num_retries + 1) * timeout <= run_limits.task_timeout` for that experiment. Encode the `2n + 1` formula and its derivation in the script docstring — litellm aliases `num_retries` onto the provider SDK's `max_retries` (`main.py:5202-5203`) *and* drives its own tenacity loop via `_get_wrapper_num_retries` → `acompletion_with_retries`, so the composition is `(n+1) + n` provider calls, not `n+1`; and `timeout` (`judge_litellm.py:125`, default `120.0` at line 51) bounds each attempt, not the whole call, which is exactly why the product is the right bound. Applying it to the shipped values fails today: `n=5` with the default 120s per attempt is 11 × 120 = 1320s against `task_timeout` 900 (`smoke.yaml`, `smoke-windows.yaml`), 1200 (`default.yaml`, `nightly.yaml`) and 360 (`activation.yaml`). Pair it with a second assertion in the same pass: `num_retries` must carry an adjacent `#` comment recording the intended attempt budget, matching how the neighbouring knobs in these same files already justify their numbers (`default.yaml:8-10` on `turn_timeout`, `activation.yaml:33-37` on `task_timeout`, which already states this class of invariant — `task_timeout` below `max_turns × turn_timeout` silently drops the row — in prose). _Prevents:_ Finding 1 (Medium, Axis 3) — a value chosen against the label rather than the total attempt count, whose worst-case retry chain exceeds the enclosing `task_timeout`. The arithmetic makes the aliasing visible at authoring time instead of leaving the YAML number and the call count silently different, and the required comment closes the half of Finding 3's and Finding 4's recommendation that asks for the rationale to live next to the value rather than only in git history.
- [repo-local-check] In the same script, add a **mirror-parity assertion**: every `tests/experiments/*.yaml` that defines `checker_context.api_route` must define a block that is set-equal to the canonical copy in `tests/experiments/default.yaml` (compare parsed key/value maps, not text, so comment and ordering differences are allowed but a value divergence is not). Follow `scripts/check_connection_parity.py`'s framing — it already exists in this repo for exactly this shape of problem (`require A == B == C`, with a docstring explaining that each site is a projection of the previous one and nothing in the toolchain reports a mismatch). Print the differing key and both values on failure. _Prevents:_ Finding 3 (Low, Axis 6) — a sixth knob added to a five-way copy-pasted block with no sharing mechanism (verified: zero YAML anchors, no `extends`, no base experiment file). The check keeps the five copies honest without introducing an include mechanism coder_eval is not known to support, and would have caught the same block's already-demonstrated drift.
- [repo-local-check] Make the `checker_context` sample in `tests/README.md` **generated rather than hand-maintained**: wrap the fenced block under `## Checker Context` (currently L245-255) in `<!-- BEGIN GENERATED CHECKER CONTEXT -->` / `<!-- END GENERATED CHECKER CONTEXT -->` and add `--write-readme` / `--check-readme` modes to `scripts/check-experiment-config.py` that emit the block from `tests/experiments/default.yaml`. This is the repo's existing docs-sync mechanism, reused verbatim: `scripts/check-skill-status.py` already carries `BEGIN_MARKER`/`END_MARKER`/`BLOCK_RE` plus `--write-readme`/`--check-readme`, and `validate-skill-status.yml` runs the check step. Add the `--check-readme` step to the new gate workflow and list `tests/README.md` in its trigger set. _Prevents:_ Finding 4 (Medium, Axis 8) — the authoritative sample is now a strict subset of all five real configs (`grep -c num_retries tests/README.md` → 0). Generation removes the whole drift class rather than this one instance, and incidentally fixes the pre-existing `model: azure/gpt-5.6-luna` divergence the finding flags as out of scope (the sample currently contradicts its own introducing sentence at L243).
- [repo-local-check] Add a **pin-consistency assertion** (same script, or a step in the new gate): `tests/Makefile`'s installs at lines 18 and 22 request `coder-eval[litellm] @ git+https://github.com/UiPath/coder_eval.git` with no ref, while `tests/.coder-eval-version` pins `0.11.5` and `smoke-skills.yml`'s `validate-task-schema` job correctly installs `coder-eval[litellm]==$(tr -d '[:space:]' < tests/.coder-eval-version)`. Fail when the Makefile's git spec carries no `@<ref>` matching `.coder-eval-version`, or switch the Makefile to the pinned-wheel form the CI job already uses. _Prevents:_ The reproducibility caveat inside Finding 1 — the measured attempt counts can shift with a litellm minor bump, and `uv.lock` constrains only `litellm >=1.95.0,<2.0.0`. Statically decidable (a string comparison between two committed files), so it belongs here rather than in the runtime bucket, even though the risk it removes is a runtime one: local runs and the CI schema gate currently resolve to different engines.

- Extend the existing advisory `validate-task-schema` job in `.github/workflows/smoke-skills.yml` (L695-735) to validate the **experiment** configs too, not just tasks: after the current `coder-eval plan` sweep over `tests/tasks`, loop the five experiment files and run `coder-eval plan <one representative task> -e tests/experiments/<name>.yaml`. The job already installs the pinned `coder-eval[litellm]==$(cat tests/.coder-eval-version)`, and `plan` accepts `-e` (verified: `src/coder_eval/cli/plan_command.py:20-25`, and its docstring states it also resolves experiment info and per-variant agent configs). Add `tests/experiments/**` to the job's trigger set. Keep `continue-on-error: true` to match the job's advisory framing. State explicitly in the step comment that this does **not** validate `params` keys — `params` is `dict[str, Any]`, which is why the allowlist static check above exists. _Why not static:_ Loading requires the pinned coder_eval package's own Pydantic models at runtime; this repo vendors no copy of that schema, so nothing in-tree can decide whether an experiment file loads. _Prevents:_ The process gap behind all four findings: this PR changed five experiment files and the coder_eval contract validation was NOT RUN (stale local venv, no litellm), so nothing — not even a schema load — was executed against the changed grading path.
- Add a pytest under `tests/scripts/` (the repo already has `tests/pytest.ini` and precedents like `tests/scripts/test_verb_checkers.py`, `test_catalog_build_guards.py`) that stands up a stub HTTP endpoint returning 500 and asserts the observed provider-call count for the repo's shipped `params` map against a recorded golden number — 11 for `num_retries: 5`, 5 for `num_retries: 2`, 3 with the key absent, plus the hanging-endpoint case showing the per-attempt `timeout` multiplying rather than bounding. Skip when `litellm` is unavailable so it degrades on a bare checkout. Record the litellm version the goldens were taken against in the test docstring. _Why not static:_ The attempt count is emergent from litellm's aliasing composed with its tenacity wrapper across two layers of its own source; it is only observable by executing the client, and the static arithmetic check above encodes the formula but cannot prove it still holds after a dependency bump. _Prevents:_ Finding 1's unverified hop — the review's counts are reproduced but the effect inside a real run is read, not observed — and Finding 2's silent-no-op class, which this test turns into a hard failure the moment the key stops taking effect.
- Widen the passing-run-claim convention from tasks to experiments: add `tests/experiments/**/*.yaml` to `lint-tasks.yml`'s `paths:` filter (L12-13) and to the changed-file glob its Step 2 prompt runs, and extend the Step 3 wording so an experiment-only change must also carry a claim in the PR body that a graded run was executed against the modified experiment. Keep it advisory, matching the workflow's existing honor-system framing. _Why not static:_ It evaluates prose in a PR body for a semantic claim about work performed outside the repo — the existing LLM lint is the only mechanism here that can judge that, and no static rule can. _Prevents:_ The shape of this PR specifically: five insertions touching the grading path, zero task files changed, so `lint-tasks.yml` never fired and no run claim was ever required. Applies to all four findings, since none of them would have survived one executed judge-failure run.
- File an upstream coder_eval issue to include `checker_context.api_route.params` in the recorded route metadata — `orchestrator.py:1593-1601` records only `litellm_base_url_host` and `litellm_model` for a `LiteLLMRoute` today, so no run artifact can tell you whether retries were enabled or what the attempt budget was. Interim mitigation in this repo: the generated README block (static check 4) plus the required rationale comment (static check 2) give one reviewable place where the shipped budget is stated. _Why not static:_ It is after-the-fact observability of runtime state — no committed text can record what a past run actually used, and the fix lives in a dependency this repo does not own. _Prevents:_ The diagnosis cost behind Findings 1 and 2 — with `params` absent from run artifacts, a retry chain eating `task_timeout` or a dead knob both present as ordinary flaky rows with no signal pointing at the experiment config.

## Top 5 Priority Actions

1. Bound the judge attempt budget so a degraded run cannot eat the task watchdog and persist not-evaluated rows: lower to `num_retries: 2` (5 provider calls, verified) and add a per-attempt `timeout: 30` in the same `params:` map at tests/experiments/smoke.yaml:50 — plus the four mirrors (activation.yaml:59, default.yaml:32, nightly.yaml:50, smoke-windows.yaml:34) — since `params` is applied after the judge's own kwargs (judge_litellm.py:125 sets `timeout`, line 135 does `kwargs.update(route.params)`) and `env_params` sets only `api_base`/`api_key`, so neither key is clobbered.
2. Document the knob where authors actually read it: add `num_retries: 5` to the fenced `params:` block at tests/README.md:250-251 plus one sentence of prose on what it does and why that value, and fix the stale `model: azure/gpt-5.6-luna` at tests/README.md:249 in the same edit — the sample is currently a strict subset of all five shipped experiments and contradicts its own introducing sentence at L243, so anyone hand-authoring an `api_route` override from it silently ships a judge with no retries.
3. Add a dependency-free repo-local check (existing scripts/ validator pattern) that extracts the fenced YAML samples under tests/README.md's Checker Context heading and asserts their `checker_context.api_route.params` key set equals the key set in tests/experiments/*.yaml — statically decidable from in-repo YAML on both sides, and the only durable guard against this sample drifting a third time.
4. Record the intended attempt budget as a one-line comment on the key in the canonical copy at tests/experiments/default.yaml:32, in the style the neighbouring knobs there already use (`turn_timeout` at lines 8-10, `post_run` at 37-39), noting that `num_retries` is aliased (the YAML value is not the call count) and that six places — five experiments plus tests/README.md — must move together.
5. Close the silent-no-op gap on this unvalidated `dict[str, Any]` passthrough (routing.py:161; no allowlist in coder_eval, no validation in this repo): a misspelled `num_retires: 5` was verified to be byte-identical to no retries with zero warning, and `params` is never recorded in run artifacts (orchestrator.py:1593-1601 keeps only `litellm_base_url_host`/`litellm_model`) — so name the contract owner in the comment above and pin the toolchain, since tests/Makefile:18,22 installs `coder-eval[litellm]` from git main rather than the `0.11.5` in tests/.coder-eval-version and a litellm minor bump can shift the attempt counts underneath these values.

---

**Change class:** Config tuning — adds num_retries: 5 to checker_context.api_route.params in 5 experiment YAMLs; no code or schema changes.


**Stats:** 0 🔴 · 0 🟠 · 2 🟡 · 2 🔵 across 5 axes reviewed.
**Verification:** 4 medium+ finding(s) adversarially re-checked · 2 dropped as false positives (50%) · 2 corrected in place · 2 low passed through unverified.

PR #3015, opened 2026-09-02
MST-14620: test(case): rename misleading guardrails test in case

charlesliu9:
Why:
The skill-case-artifact-safety-guard test was placed in the subfolder named "guardrails" by claude, while it's just validating assemble caseplan.json / registry-resolved.json with Read + Write/Edit only. The name is misleading, and triggered the rule requiring other code owners to approve any change on the test.
The test also needs update after removing tasks.md.
What:
Rename the folder to avoid misunderstanding, remove the required code owners for this particular test.
Verify:
https://github.com/UiPath/skills/actions/runs/33673778364 Run successfully

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -121,7 +121,6 @@
 # Case Management skill
 /skills/uipath-maestro-case/ @charlesliu9 @song-zhao-25 @abhiram-vad @roberts-cliff @StephenHanzlik
 /tests/tasks/uipath-maestro-case/ @charlesliu9 @song-zhao-25 @abhiram-vad @roberts-cliff @StephenHanzlik
-/tests/tasks/uipath-maestro-case/guardrails/ @apetraru-uipath @valentinabojan @ctiliescuuipath
 
 # Coded Apps skill
 /skills/uipath-coded-apps/ @Raina451 @deepeshrai-tech @Sandeepan-Ghosh-0312 @swati354 @ninja-shreyash
--- a/tests/tasks/uipath-maestro-case/hitl/smoke_01_action_task_placed.yaml
+++ b/tests/tasks/uipath-maestro-case/hitl/smoke_01_action_task_placed.yaml
@@ -88,10 +88,10 @@ initial_prompt: |
 
   Important:
   - The `uip` CLI is already available in the environment.
-  - Treat every build gate as pre-approved — do not block on any HARD STOP
-    / AskUserQuestion approval step (Phase 0, Phase 1, or Phase 2a→2b).
-  - If the action-app cannot be resolved from the registry, write a skeleton
-    action task per the skill's skeleton-tasks reference (type "action",
+  - Treat the Phase 1 and Phase 2→3 build gates as pre-approved — do not
+    block on their HARD STOP / AskUserQuestion approval steps.
+  - If the action-app cannot be resolved from the registry, write a placeholder
+    action task per the skill's placeholder-tasks reference (type "action",
     empty data) — do NOT fabricate a taskTypeId.
   - Do NOT run `uip maestro case debug`.
 
--- a/tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml
+++ b/tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml
@@ -1,8 +1,8 @@
 task_id: skill-case-artifact-safety-guard
 description: >
   Rule 13 negative guard: while building a two-stage case (HITL action →
-  downstream process), the agent MUST assemble caseplan.json / tasks.md
-  with Read + Write/Edit only. It must NOT shell out to a script or
+  downstream process), the agent MUST assemble caseplan.json /
+  registry-resolved.json with Read + Write/Edit only. It must NOT shell out to a script or
   interpreter to write, parse, or redirect into a skill artifact
   (`node -e ...fs`, `python ...open(...)`, `jq ... > caseplan.json`,
   `sed -i`, a `/tmp/build-caseplan.js` helper, etc.). The sanctioned
@@ -146,8 +146,8 @@ initial_prompt: |
 
   Important:
   - The `uip` CLI is already available.
-  - Treat the generated plan as pre-approved — do not block on any HARD STOP
-    / AskUserQuestion approval step (Phase 0, post-tasks.md, or Phase 2a→2b).
+  - Treat every build gate as pre-approved — do not block on any HARD STOP
+    / AskUserQuestion approval step (Phase 0, Phase 1, or Phase 2a→2b).
   - If a resource cannot be resolved from the registry, write a skeleton task
     per the skill's skeleton-tasks reference — do NOT fabricate a taskTypeId.
   - Do NOT run `uip maestro case debug`.
@@ -188,18 +188,18 @@ success_criteria:
   - type: command_not_executed
     description: "Rule 13: no inline interpreter or stream-editor manipulation of a skill artifact (sed -i, perl -i, gawk in-place, jq on/into an artifact, python open(), node -e ...fs.*)"
     tool_name: "Bash"
-    command_pattern: '(sed\s+-i|perl\s+-i|awk\b[^\n]*-i\s*inplace|jq\b[^\n]*([<>]|caseplan\.json|tasks\.md|bindings_v2\.json|entry-points\.json)|python[0-9]?\b[^\n]*open\s*\(|(node|nodejs)\s+(-e|--eval)\b[^\n]*fs\.)'
+    command_pattern: '(sed\s+-i|perl\s+-i|awk\b[^\n]*-i\s*inplace|jq\b[^\n]*([<>]|caseplan\.json|registry-resolved\.json|bindings_v2\.json|entry-points\.json)|python[0-9]?\b[^\n]*open\s*\(|(node|nodejs)\s+(-e|--eval)\b[^\n]*fs\.)'
     weight: 2.0
     pass_threshold: 1.0
 
   - type: command_not_executed
-    description: "Rule 13: no shell redirection into a skill artifact and no helper-script build-assembler (`> caseplan.json`, `| tee tasks.md`, `node build-caseplan.js`, `python gen.py`)"
+    description: "Rule 13: no shell redirection into a skill artifact and no helper-script build-assembler (`> caseplan.json`, `| tee registry-resolved.json`, `node build-caseplan.js`, `python gen.py`)"
     tool_name: "Bash"
     # The script path must be the token immediately after the interpreter: agents wrap every
     # command in `/bin/bash -lc '...'`, so `bash\s+[^\n]*\.sh` would match any command that merely
-    # mentions a script path (e.g. `rg ... audit_plan.py`). The lookahead exempts the skill's own
-    # read-only auditor, which planning.md mandates as the final plan gate.
-    command_pattern: '([<>]{1,2}\s*\S*(caseplan\.json|tasks\.md|bindings_v2\.json|entry-points\.json)|tee\b[^\n]*(caseplan\.json|tasks\.md)|\b(node|nodejs|python[0-9]?|bash|sh)\s+(?!\S*uipath-maestro-case/scripts/)\S+\.(js|mjs|cjs|py|sh)\b)'
+    # mentions a script path (e.g. `rg ... audit_caseplan.py`). The lookahead exempts the skill's
+    # own read-only auditor, which implementation.md mandates as the Step 12 build gate.
+    command_pattern: '([<>]{1,2}\s*\S*(caseplan\.json|registry-resolved\.json|bindings_v2\.json|entry-points\.json)|tee\b[^\n]*(caseplan\.json|registry-resolved\.json)|\b(node|nodejs|python[0-9]?|bash|sh)\s+(?!\S*uipath-maestro-case/scripts/)\S+\.(js|mjs|cjs|py|sh)\b)'
     weight: 2.0
     pass_threshold: 1.0
 
--- a/tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml
+++ b/tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml
@@ -9,7 +9,7 @@ description: >
   `node -e "crypto.randomUUID()"` UUID subprocess (no `fs`, no redirection)
   stays allowed. Grades the build succeeding via the sanctioned path, not
   a self-report.
-tags: [uipath-maestro-case, integration, mode:build, lifecycle:generate, guardrail, shape:multi-node, node:hitl]
+tags: [uipath-maestro-case, integration, mode:build, lifecycle:generate, shape:multi-node, node:hitl]
 
 initial_prompt: |
   Create a UiPath Case Management project named "InvoiceReviewCase" with a
@@ -146,11 +146,13 @@ initial_prompt: |
 
   Important:
   - The `uip` CLI is already available.
-  - Treat every build gate as pre-approved — do not block on any HARD STOP
-    / AskUserQuestion approval step (Phase 0, Phase 1, or Phase 2a→2b).
-  - If a resource cannot be resolved from the registry, write a skeleton task
-    per the skill's skeleton-tasks reference — do NOT fabricate a taskTypeId.
+  - Treat the Phase 1 and Phase 2→3 build gates as pre-approved — do not
+    block on their HARD STOP / AskUserQuestion approval steps.
+  - If a resource cannot be resolved from the registry, write a placeholder
+    task per the skill's placeholder-tasks reference — do NOT fabricate a
+    taskTypeId.
   - Do NOT run `uip maestro case debug`.
+  - Do NOT publish to Studio Web or Orchestrator.
 
   The task is NOT complete until `uip maestro case validate` passes on the
   generated caseplan.json.
@@ -186,9 +188,9 @@ success_criteria:
   # --- Rule 13 guards: no interpreter/stream-editor/helper-script assembly of a skill artifact ---
   # crypto.randomUUID() UUID subprocess (no fs, no redirection) is NOT matched by either pattern.
   - type: command_not_executed
-    description: "Rule 13: no inline interpreter or stream-editor manipulation of a skill artifact (sed -i, perl -i, gawk in-place, jq on/into an artifact, python open(), node -e ...fs.*)"
+    description: "Rule 13: no inline interpreter or stream-editor manipulation of a skill artifact (sed -i, perl -i, gawk in-place, jq on/into an artifact, python open(), node -e ...fs. / require('fs'))"
     tool_name: "Bash"
-    command_pattern: '(sed\s+-i|perl\s+-i|awk\b[^\n]*-i\s*inplace|jq\b[^\n]*([<>]|caseplan\.json|registry-resolved\.json|bindings_v2\.json|entry-points\.json)|python[0-9]?\b[^\n]*open\s*\(|(node|nodejs)\s+(-e|--eval)\b[^\n]*fs\.)'
+    command_pattern: '(sed\s+-i|perl\s+-i|awk\b[^\n]*-i\s*inplace|jq\b[^\n]*([<>]|caseplan\.json|registry-resolved\.json|bindings_v2\.json|entry-points\.json)|python[0-9]?\b[^\n]*open\s*\(|(node|nodejs)\s+(-e|--eval)\b[^\n]*(fs\.|require\s*\(\s*[''"](node:)?fs[''"]))'
     weight: 2.0
     pass_threshold: 1.0
 


roberts-cliff (approved):

uipreliga (approved):
🚢

uipreliga (commented):
# Review: @uipath/skills (UiPath agent skills for Claude Code) — pr:3015 (2 files) axis:1,3,4,6,8

PR #3015 — "test(case): rename misleading guardrails test in case" by charlesliu9 (Yang), feat/case-rename-guardrail -> main, OPEN. 2 files changed, +9/-10.

**Change class:** Test-asset retarget + ownership removal — renames a coder-eval task directory (guardrails/ -> safety_guard/), retargets its Rule-13 grading regexes from the `tasks.md` artifact to `registry-resolved.json`, and deletes the CODEOWNERS entry that assigned the old directory to a separate owning team.

Architecture, test health, and docs are in excellent shape (9.7/10 overall, zero critical or high findings, and the PR's rename is substantively correct), but the residual risk is concentrated in one file — `tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml`, where a negative guard that can silently pass while the forbidden write happens, a mis-scoped `guardrail` tag that pollutes the product-guardrails metric slice, and an over-broadened pre-approval sentence all sit — so the bottom line is: merge-worthy quality, but finish the half-completed rename/cleanup pass before these quiet failure modes calcify.

## Summary

| 1. Correctness & Logic | 9.5 / 10 | 0 | 0 | 1 | 0 | Rule 13 `node -e` guard misses the idiomatic `require('fs').writeFileSync(...)` form its own description advertises |
| 3. Types & Contracts | 9.4 / 10 | 0 | 0 | 1 | 1 | Rename incomplete: the misleading `guardrail` tag survives on line 12, keeping the task in the product-guardrails query slice (and absent from tests/README.md's taxonomy) |
| 4. Test & Validation Health | 9.9 / 10 | 0 | 0 | 0 | 1 | Prompt cleanup pass missed a stale reference to a nonexistent `skeleton-tasks` skill reference (and two nonexistent phase labels) |
| 6. Architecture & Design | 10 / 10 | 0 | 0 | 0 | 0 | — |
| 8. Interface, Docs & Compatibility | 9.8 / 10 | 0 | 0 | 0 | 2 | Rewritten pre-approval line names phase gates that do not exist in / contradict the uipath-maestro-case skill's phase model |

**Overall Score**: 9.7 / 10 · **Weakest Axis**: Types & Contracts at 9.4 / 10
**Totals**: 🔴 0 · 🟠 0 · 🟡 2 · 🔵 4 across 5 axes reviewed.

None.

1. **[Axis 1] Rule 13 `node -e` guard misses the idiomatic `require('fs').writeFileSync(...)` form its own description advertises** (`tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:191`) — Widen line 191's last alternative so it also catches the inline-require form. Replace the trailing `(node|nodejs)\s+(-e|--eval)\b[^\n]*fs\.` with:

```
(node|nodejs)\s+(-e|--eval)\b[^\n]*(fs\.|require\(\s*[''"](node:)?fs[''"]\s*\))
```

Note the doubled apostrophes: `command_pattern` is a YAML single-quoted scalar, so a literal `'` inside it must be written `''`. Writing the class as `['\"]` (single apostrophe) terminates the scalar and makes the file fail to parse. After the edit the full line reads:

```yaml
    command_pattern: '(sed\s+-i|perl\s+-i|awk\b[^\n]*-i\s*inplace|jq\b[^\n]*([<>]|caseplan\.json|registry-resolved\.json|bindings_v2\.json|entry-points\.json)|python[0-9]?\b[^\n]*open\s*\(|(node|nodejs)\s+(-e|--eval)\b[^\n]*(fs\.|require\(\s*[''"](node:)?fs[''"]\s*\)))'
```

This matches `require('fs')`, `require("fs")`, and `require('node:fs')` in both `-e` and `--eval` forms, keeps the existing two-statement `const fs=require('fs'); fs.writeFileSync(...)` match, and still leaves the sanctioned `node -e "crypto.randomUUID()"` / `require('crypto').randomUUID()` subprocess unmatched.

Because a negative guard is invisible when it fails, running the coder-eval task cannot validate this — a compliant agent never emits the forbidden command. Add a regex fixture under `tests/tasks/uipath-maestro-case/_shared/` (which already holds `test_case_check.py`, `test_audit_caseplan.py`, and friends), e.g. `test_artifact_safety_guard_patterns.py`, that loads the two `command_not_executed` patterns out of the YAML and asserts each example string named in the criterion `description` and the task `description` matches, plus a negative case asserting the sanctioned `crypto.randomUUID()` subprocess does not.
2. **[Axis 3] Rename incomplete: the misleading `guardrail` tag survives on line 12, keeping the task in the product-guardrails query slice (and absent from tests/README.md's taxonomy)** (`tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:12`) — Drop `guardrail` from `tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:12` — the task is already identified by its path and `task_id` (`skill-case-artifact-safety-guard`), and the remaining tags (`uipath-maestro-case`, `integration`, `mode:build`, `lifecycle:generate`, `shape:multi-node`, `node:hitl`) satisfy tests/README.md Rule 1's required set. If a negative-guard marker is genuinely wanted repo-wide, propose a distinct value (e.g. `safety-guard`) and add it to the Tag Taxonomy table in tests/README.md in this PR, per Rule 5 at tests/README.md:119.

Rationale: `guardrail` currently appears on 44 task YAMLs (`grep -rln '^tags:.*guardrail' tests/tasks`). The other 43 all mean the AI Trust Layer product feature — 16 under `tests/tasks/uipath-agents/lowcode/guardrails/`, 13 under `tests/tasks/uipath-agents/coded/guardrails/`, 10 under `tests/tasks/uipath-review/agents/*guardrail*/`, 4 under `tests/tasks/uipath-platform/guardrails/` (`uip guardrails byo-configurations`, per skills/uipath-platform/SKILL.md:55), and 1 under `tests/tasks/uipath-agents/lowcode/conversational/guardrails/`. `tests/Makefile:41-51` matches tags by exact token, so `make tags TAGS="guardrail"` and the evalboard `where tag == "guardrail"` slice still sweep this Rule 13 artifact-I/O task into the guardrails-feature numerator — exactly the collision the rename was meant to remove. The tag is also absent from tests/README.md entirely (`grep -n 'guardrail' tests/README.md` returns nothing). The 43 pre-existing uses are out of scope; this line is in scope because the rename half-finished the disambiguation.

## Nits

1. **[Axis 3] CODEOWNERS entry deleted rather than retargeted to the new `safety_guard/` path — three reviewers silently lose ownership** (`CODEOWNERS:123`) — The diff removes `/tests/tasks/uipath-maestro-case/guardrails/ @apetraru-uipath @valentinabojan @ctiliescuuipath` (it sat immediately after line 123) and adds no `/tests/tasks/uipath-maestro-case/safety_guard/` replacement. `sed -n '121,124p' CODEOWNERS` at HEAD shows only the two case lines remaining, so the task now falls through to `/tests/tasks/uipath-maestro-case/ @charlesliu9 @song-zhao-25 @abhiram-vad @roberts-cliff @StephenHanzlik` (line 123) — no path is left unowned, which is why this is Low. But the deletion is an ownership change that the diff does not announce: if those three were on the path only because the directory happened to be named `guardrails` (which is the premise of the rename — they own the *product* guardrails paths at CODEOWNERS lines 70, 71, 73, 74, 76, 153), the removal is correct and should be stated in the PR description; if they were deliberate reviewers of this Rule 13 test, the line needs retargeting to `/tests/tasks/uipath-maestro-case/safety_guard/`. Confirm intent and either add the retargeted line or note the intentional handoff in the PR body.
2. **[Axis 4] Prompt cleanup pass missed a stale reference to a nonexistent `skeleton-tasks` skill reference (and two nonexistent phase labels)** (`tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:151`) — This PR fixes three stale references in this file (`tasks.md`→`registry-resolved.json`, `planning.md`→`implementation.md`, `audit_plan.py`→`audit_caseplan.py`), but the same cleanup pass left the prompt's own stale pointers.

Lines 151-152: "If a resource cannot be resolved from the registry, write a skeleton task per the skill's **skeleton-tasks reference** — do NOT fabricate a taskTypeId."

No such reference exists. `grep -rin 'skeleton' skills/uipath-maestro-case/` returns only the `--skeleton-v2` / `--skeleton` CLI validate flags (SKILL.md:43, implementation.md:224-228, phased-execution.md:78-81) — never a task shape. The skill's concept is a **placeholder task**, documented at `skills/uipath-maestro-case/references/placeholder-tasks.md` and named that way in SKILL.md Rule 8 (SKILL.md:40: "A placeholder task has `type`, `displayName`, structural fields, and `data: {}`"). This directly bears on grading: the `json_check` at lines 165-176 asserts an `action` and a `process` task exist, and an unresolved action app produces a placeholder whose `type` is still `action` — so the criterion holds either way, but an agent hunting a nonexistent "skeleton-tasks reference" burns turns against a 120-turn budget.

Same class, line 150: "do not block on any HARD STOP / AskUserQuestion approval step (Phase 0, Phase 1, or Phase 2a→2b)". `grep -rn 'Phase 0' skills/uipath-maestro-case/` matches only `references/evals/evals.json:86` (prose about not exposing "Phase 0 through Phase 7 labels"), and `Phase 2a`/`Phase 2b` appear nowhere — the real gate is `phased-execution.md:85` "### Phase 2 hard stop". The PR's new `Phase 1` is correct (SKILL.md:49 Rule 17 puts an AskUserQuestion empty-lookup gate inside Phase 1), and the new leading clause "Treat every build gate as pre-approved" makes the parenthetical purely illustrative — so this is cosmetic, not behavior-changing.

Fix: on line 151-152 replace "skeleton task per the skill's skeleton-tasks reference" with "placeholder task per the skill's placeholder-tasks reference"; on line 150 replace "(Phase 0, Phase 1, or Phase 2a→2b)" with "(the Rule 17 empty-lookup gate in Phase 1, or the Phase 2 hard stop)".
3. **[Axis 8] Rewritten pre-approval line names phase gates that do not exist in / contradict the uipath-maestro-case skill's phase model** (`tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:150`) — The PR rewrote exactly this list to refresh gate names, replacing the now-dead `post-tasks.md` with `Phase 1`, but left the other two entries stale. Lines 149-150 at PR HEAD:

    - Treat every build gate as pre-approved — do not block on any HARD STOP
      / AskUserQuestion approval step (Phase 0, Phase 1, or Phase 2a→2b).

Verified against the skill at PR HEAD:
- `grep -rn 'Phase 0' skills/uipath-maestro-case/` -> a single hit, `references/evals/evals.json:86` ("Does not expose internal Phase 0 through Phase 7 labels…"), i.e. no Phase 0 is defined anywhere in SKILL.md or references/. The documented spine is `SKILL.md:80`: "**Design handoff when required → Phase 1 Planning → Phase 2 Prototyping → Phase 3 Implementation → Phase 4 Validate → Phase 5 Publish → Phase 6 Debug → Phase 7 Publish to Orchestrator.**"
- `grep -rn 'Phase 2a\|Phase 2b\|2a→2b' skills/uipath-maestro-case/` -> no matches. The boundary the phrase means is `references/phased-execution.md:85` `### Phase 2 hard stop`, i.e. Phase 2→3.

The `Phase 1` half of the edit is correct — Phase 1 does carry an AskUserQuestion step (SKILL.md:49, Rule 17 empty-lookup gate: "use one batched AskUserQuestion grouped by `(name,type)`").

This is a repo-wide copy of the same stale sentence (`tests/tasks/uipath-maestro-case/hitl/smoke_01_action_task_placed.yaml:91-92` is byte-identical), so it is not newly introduced — but the PR touched this exact line to refresh it and updated one of three names.

**Fix:** while editing the line, finish the pass — `(Phase 1 resource resolution or the Phase 2→3 preview gate)` — and, if the sibling wording is meant to stay uniform, apply the same correction to `hitl/smoke_01_action_task_placed.yaml:91-92` in this PR.
4. **[Axis 8] Pre-approval was broadened from "the generated plan" to "every build gate", which now also blankets the Phase 5/6/7 consent gates the skill calls unconditional** (`tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:149`) — The diff changes the scope of the blanket, not just the gate names:

    -  - Treat the generated plan as pre-approved — do not block on any HARD STOP
    +  - Treat every build gate as pre-approved — do not block on any HARD STOP

The old phrasing was bounded to plan approval. "Every build gate" reads as covering the whole gate set, but the parenthetical that follows (line 150) still enumerates only pre-Phase-3 gates — so the sentence's blanket and its example list no longer agree on scope.

The skill's later gates are explicitly not build-progress gates: `skills/uipath-maestro-case/references/phased-execution.md:25` — "Validate (Phase 4), Publish (Phase 5), Debug (Phase 6), and Publish to Orchestrator (Phase 7) follow; the publish, debug, and Orchestrator-publish gates are unconditional", and `phased-execution.md:27` — "the gates that remain protect real-world side effects (publish ships the case to Studio Web, debug executes it, publish to Orchestrator ships it to the tenant solution feed)". This task excludes only one of them, at line 153: "- Do NOT run `uip maestro case debug`." Nothing in the task needs a publish: the three positive criteria (lines 158-184) grade only `InvoiceReviewCase/InvoiceReviewCase/caseplan.json` and `uip maestro case validate`.

Unverified and the reason this is capped at Low: I did not run the task, so I have not demonstrated that an agent actually takes the Phase 5 `uip solution upload` branch, nor that `post_run` `_shared/cleanup_solutions.py` (line 207) fails to clean it up — that script's own docstring scopes it to "solutions uploaded by ``uip maestro case debug``" and it only globs `**/*.uipx`, so an upload with no `.uipx` under cwd would be missed. Raise to Medium if a run shows a Studio Web solution surviving post_run.

**Fix:** bound the sentence to the gates the task means, e.g. "Treat the Phase 1 and Phase 2→3 build gates as pre-approved", and add an explicit `- Do NOT publish to Studio Web or Orchestrator.` alongside the existing debug exclusion at line 153.

## What's Missing

**Parallel paths & mirrors:**
- 🔵 The "skeleton-tasks reference" pointer this PR left at line 152 is copy-pasted boilerplate shared by five sibling case tasks that were not touched: `tests/tasks/uipath-maestro-case/hitl/smoke_01_action_task_placed.yaml:94`, `hitl/quality_01_action_schema_design.yaml:187`, `hitl/quality_02_result_downstream.yaml:162`, and `hitl/quality_04_skeleton_completion_report.yaml:30` and `:249`. Verified the target does not exist: `ls skills/uipath-maestro-case/references/ | grep -i 'skele|placeholder'` returns only `placeholder-tasks.md`, and `grep -rin 'skeleton' skills/uipath-maestro-case/` matches only the `--skeleton` / `--skeleton-v2` CLI validate flags — never a task shape or a reference file. Fixing the one in-scope file leaves five agents still hunting a nonexistent reference; do the rename across all six occurrences (`skeleton-tasks` -> `placeholder-tasks`) in this PR, or state that the sweep is deferred. _(trigger: tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml)_ _(restates: Axis 4: Prompt cleanup pass missed a stale reference to a nonexistent `skeleton-tasks` skill reference)_
- 🔵 The pre-approval sentence this PR rewrote (lines 149-150, `Treat every build gate as pre-approved ... (Phase 0, Phase 1, or Phase 2a→2b)`) is byte-identical to `tests/tasks/uipath-maestro-case/hitl/smoke_01_action_task_placed.yaml:91-92`, which this PR does not touch — the edit converged this file onto a sibling's wording rather than diverging from it, so the two copies now carry the same nonexistent gate names in lockstep. Verified `grep -rn 'Phase 0' skills/uipath-maestro-case/` yields one hit only, `references/evals/evals.json:86` (prose about not exposing phase labels), and `Phase 2a`/`Phase 2b` appear nowhere in the skill; the real gate is `references/phased-execution.md:85` `### Phase 2 hard stop`. Whichever correction is applied here should be applied to the sibling in the same PR so the shared prompt block stays one string. _(trigger: tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml)_ _(restates: Axis 8: Rewritten pre-approval line names phase gates that do not exist in / contradict the uipath-maestro-case skill's phase model)_

**Tests:**
- 🟡 Both `command_not_executed` regexes (lines 191 and 202) were edited in this PR and nothing exercises them. A negative criterion is invisible when it is wrong — a compliant agent never emits the forbidden command, so a green coder-eval run proves nothing about the patterns. The infrastructure for a unit fixture already exists and is wired to CI: `tests/tasks/uipath-maestro-case/_shared/` holds `test_case_check.py`, `test_audit_caseplan.py`, `test_sdd_check.py`, `test_check_issue_journal.py`, `test_cleanup_solutions.py`, and `.github/workflows/test-helpers.yml:69` runs `pytest tests/tasks/uipath-maestro-case/ -v` on any change under `tests/tasks/uipath-maestro-case/**` — so a `test_artifact_safety_guard_patterns.py` that loads the two patterns out of the YAML and asserts every example string named in the criterion descriptions matches (plus a negative case for the sanctioned `node -e "crypto.randomUUID()"` subprocess) would run automatically on this very PR. Without it the `node -e require('fs')` hole in Axis 1 is exactly the class of regression that ships silently. _(trigger: tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml)_ _(restates: Axis 1: Rule 13 `node -e` guard misses the idiomatic `require('fs').writeFileSync(...)` form its own description advertises)_
- 🟡 `registry-resolved.json` was added to the protected-artifact set in both guard patterns (lines 191, 202) and to the task `description` (line 5), but the task never produces or asserts that artifact, so the new alternatives are dead in the graded run. Verified: `grep -n 'registry' <task>` matches only the description, the prompt's generic phrase "resolved from the registry" (line 151), and the two regexes — the prompt never names the file. The three positive criteria all target `InvoiceReviewCase/InvoiceReviewCase/caseplan.json` (`file_exists` w1.5, `json_check` w1.5, `run_command uip maestro case validate` w3.0); nothing checks `tasks/registry-resolved.json` exists or was authored with Read/Write only. Either add a `file_exists` on `InvoiceReviewCase/InvoiceReviewCase/tasks/registry-resolved.json` (and name the artifact in the prompt) so the widened guard has something to guard, or drop the new alternatives and leave the description scoped to `caseplan.json` — as written the PR broadens the stated contract without broadening coverage. _(trigger: tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml)_

**Downstream consumers:**
- 🟡 The rename moved the directory but not the query key that consumers actually slice on. `tests/Makefile:41-51` implements `make tags TAGS="..."` as an exact-token subset match over each task's `tags: [...]` array, so `make tags TAGS="guardrail"` still selects this Rule 13 artifact-I/O task alongside the 43 AI Trust Layer guardrails-feature tasks (16 `uipath-agents/lowcode/guardrails/`, 13 `uipath-agents/coded/guardrails/`, 10 `uipath-review/agents/*guardrail*/`, 4 `uipath-platform/guardrails/`, 1 conversational). `.claude/commands/generate-confluence-scorecard.md:175-192` likewise builds its CROSS-CUTTING TAGS rows by tag alias, explicitly "NOT any directory bucket" — so every tag-driven consumer is unchanged by this PR and the naming collision the rename targeted survives in the only place it is machine-read. _(trigger: tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml)_ _(restates: Axis 3: Rename incomplete: the misleading `guardrail` tag survives on line 12)_

**Docs & config:**
- 🔵 The generated coverage report was not refreshed: `tests/reports/uipath-maestro-case.md:102` still lists `| skill-case-artifact-safety-guard | integration | mode:build, guardrail, node:hitl | Rule 13 artifact-I/O guardrail | ... |`, and line 352 counts the repo-local extras as "`guardrail` (1), `skeleton` (11)" — both the tag column and the prose description keep calling this task a guardrail, which is precisely the label the rename set out to retire. (`tests/reports/coverage.json` is unaffected — `grep -o guardrails` over it returns 0 hits, and the report's inventory is keyed by `task_id`, not directory, so the path change alone leaves it valid.) If the `guardrail` tag is dropped per the Axis 3 fix, regenerate with `/test-coverage uipath-maestro-case`; if it is kept, the report row perpetuates the confusion in the human-facing artifact. _(trigger: tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml)_ _(restates: Axis 3: Rename incomplete: the misleading `guardrail` tag survives on line 12)_

**Rollout impact:**
- 🔵 The PR states no blast radius for a path rename plus an ownership deletion. I checked the plausible consumers and the radius is genuinely narrow — worth saying so in the PR body rather than leaving it implicit: (a) `git diff --name-only origin/main...HEAD` reports only the destination path for an R087 rename, so `smoke-skills.yml`'s `resolve_task_yaml` (lines 175-196) never sees the deleted path and cannot emit a glob for a nonexistent file; (b) `lint-tasks.yml:13` (`tests/tasks/**/*.yaml`), `test-helpers.yml:7` (`tests/tasks/uipath-maestro-case/**`), and `task-driver-gate.yml:40` (`tests/tasks`) are all prefix/glob filters that follow the rename; (c) no experiment config hardcodes task paths (`grep -rn 'tasks/' tests/experiments/*.yaml` finds only comments); (d) `post_run` resolves `_shared/cleanup_solutions.py` through `$SKILLS_REPO_PATH`, and the leaf directory depth is unchanged; (e) no tracked file outside CODEOWNERS referenced the old directory. The one real effect is review routing: @apetraru-uipath, @valentinabojan, @ctiliescuuipath stop being auto-requested on this task. State that handoff in the PR description (or restore the line retargeted to `safety_guard/`). _(trigger: CODEOWNERS)_ _(restates: Axis 3: CODEOWNERS entry deleted rather than retargeted to the new `safety_guard/` path)_

## Guardrails & Automation

- [repo-local-check] **Preamble (applies to all four static proposals):** this repo configures NO linter, typechecker, formatter, or pre-commit hook — verified: no `eslint`/`ruff`/`prettier`/`markdownlint`/`pyproject.toml`/`setup.cfg`/`.pre-commit-config.yaml` at root, and no `yamllint`/`actionlint` step in `.github/workflows/`. Rubric preferences 1, 2, and 4 are therefore unavailable, and every static check must be preference 3: a `scripts/check-<thing>.py` wired to a `.github/workflows/<thing>-gate.yml`, following the exact existing pair `scripts/check-task-driver.py` + `.github/workflows/task-driver-gate.yml`.

**Proposal 1 — `scripts/check-task-tags.py` + `.github/workflows/task-tag-gate.yml` (closed-vocabulary tag gate).**
Parse the Tag Taxonomy table in `tests/README.md` at runtime and validate each task's `tags:` array against it. This is the repo's own established "parse the rule table at runtime" idiom — `scripts/check-cli-verbs.py` already does exactly this against the pipe-delimited table in `.claude/rules/cli-renames.md`, per that file's "How it's used" section. Three rules:
  a. **Required tags present** (tests/README.md Rule 1): `skill` + `tier` + `mode:*` + `lifecycle:*`.
  b. **Prefixed namespaces are closed**: `mode:` ∈ {build, operate, diagnose}; `lifecycle:` ∈ {discover, generate, setup}; `tier` ∈ {smoke, integration, e2e}; `shape:` ∈ {single-node, multi-node}; `feature:` from the closed list in the table. Also reject the hyphen-for-colon typo form (`lifecycle-generate`, `shape-multi-node`).
  c. **Bare tags must be declared**: any unprefixed tag that is not a skill folder name and not one of the documented flat markers (`resource`, `connector`, `windows`, `path-to-ga`, `outcome-graded`) fails with "tag `X` is not in the tests/README.md Tag Taxonomy — add it to the table in this PR (Rule 5) or drop it."

**Scope it to PR-changed task YAMLs, not a full scan.** Measured backlog at HEAD (PyYAML parse of all `tests/tasks/**/*.yaml`): rule (a) 8 files missing a tier tag; rule (b) 62 out-of-vocabulary `lifecycle:*` values (`lifecycle:edit` ×29, `lifecycle:execute` ×12, `lifecycle:manage` ×6, `lifecycle:validate` ×4, `lifecycle:improve` ×3, `lifecycle:deploy`/`lifecycle:report` ×2 each, `lifecycle:inspect`/`plan`/`teardown`/`triage` ×1) plus `mode:troubleshoot` ×3 and 2 hyphen-form tags in `tests/tasks/uipath-maestro-flow/connector_features/enum.yaml`; rule (c) ~200 undeclared bare tags. A repo-wide gate is therefore unshippable today. `.github/workflows/verb-gate.yml` already solves precisely this and states the rationale in its own header comment — "only the skills touched by the PR are checked, so the existing doc-drift backlog is swept opportunistically as each skill is next edited rather than blocking unrelated PRs." Copy that `detect` → `gate` job shape verbatim, with `git diff --name-only $(git merge-base origin/$BASE_REF HEAD)...HEAD | grep -E '^tests/tasks/.*\.ya?ml$'` as the file set. Consider shipping rule (c) as a `::warning` for one sprint and rules (a)/(b) as hard failures, since (c) has the largest backlog. _Prevents:_ Finding #2 (Types & Contracts / medium, merged across axes 3-4-6-8) — the un-renamed `guardrail` tag at `tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:12`. The file IS in this PR's diff, so a changed-files-scoped gate fires on it. Rule (c) reports it directly: `grep -n 'guardrail' tests/README.md` returns nothing, so the tag is undeclared and the gate forces the author either to drop it (the finding's recommendation) or to add it to the Tag Taxonomy table — at which point the semantic collision with the 43 AI-Trust-Layer guardrail tasks becomes a visible review decision instead of a silent one. Rules (a) and (b) additionally prevent the same class of half-finished-rename tag drift that produced the 62 `lifecycle:*` violations.
- [repo-local-check] **Proposal 2 — `scripts/check-task-patterns.py` + a job in the existing `task-driver-gate.yml` (rename it `task-yaml-gate.yml`, or add a second job) — regex and YAML-scalar validity for every criterion pattern.**
Walk every `tests/tasks/**/*.yaml` with `yaml.safe_load`, recurse the parsed tree, and for every `command_pattern` (and `pattern` / `file_pattern`) string call `re.compile()`. Fail on an unparseable YAML file or an uncompilable pattern, printing `path:line`. Reuse `check-task-driver.py`'s helpers verbatim — `_iter_task_yamls`, `_rel`, and the line-number-by-textual-rescan trick — and mirror its docstring/exit-code convention (`0` clean, `1` offenders printed).

**This is green today and costs nothing to adopt**: measured at HEAD, 1881 patterns across all task YAMLs compile under Python `re`, and 0 files fail `yaml.safe_load`. Wire it with no `paths:` filter so it can be a required check, exactly as `task-driver-gate.yml`'s header comment justifies ("a sub-second full scan that always reports a definitive pass/fail").

**Boundary, stated explicitly as the rubric requires:** this check reaches *syntactic* validity only. It cannot detect that a valid regex under-matches — that `(node|nodejs)\s+(-e|--eval)\b[^\n]*fs\.` fails to match `node -e "require('fs').writeFileSync(...)"` is a semantic property of the forbidden-command set, which no static check knows. That half falls to Guardrail improvement #1. _Prevents:_ The YAML-quoting footgun inside finding #1's recommendation. The reviewer's originally-proposed replacement regex embedded a bare `'` in the character class `['\"]`; because `command_pattern` on line 191 is a YAML **single-quoted** scalar, that apostrophe terminates the scalar early and `yaml.safe_load` raises `ParserError: expected <block end>, but found '<scalar>'`. The verifier caught it by hand and corrected it to the doubled form `[''\"]`. This gate catches that class mechanically on the next push instead of depending on a reviewer noticing — and it also catches any future uncompilable regex typo in the 1881 existing patterns, none of which is currently guarded by anything.
- [repo-local-check] **Proposal 3 — `scripts/check-task-doc-refs.py` + `.github/workflows/task-doc-ref-gate.yml` — task prompts may only cite skill artifacts that exist.**
For each task YAML, take the `uipath-*` tags as the owning skill(s), concatenate `description` + `initial_prompt`, and apply two narrowly-scoped rules:

  a. **Phase labels must exist in the skill.** Every `Phase <N>` / `Phase <N><letter>` token must appear literally somewhere under `skills/<skill>/**`. Measured at HEAD: exactly **5 violations repo-wide** — `Phase 2a` in `tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml`, `.../hitl/smoke_01_action_task_placed.yaml`, `.../hitl/quality_02_result_downstream.yaml`, `.../single_node/run_once_envelope/run_once_envelope.yaml`, and `Phase 2f` in `tests/tasks/uipath-coded-apps/integration_action_schema_validate_fix.yaml`. Five is a one-commit cleanup, so this can ship as a hard failure with a full scan immediately.

  b. **Hyphenated reference slugs must resolve.** For a phrase matching `skill'?s? \**([a-z0-9]+(?:-[a-z0-9]+)+)\** reference` — i.e. a *hyphenated* slug, which is filename-shaped, immediately before the word `reference` — require some file under `skills/<skill>/references/` whose stem contains that slug. The hyphen requirement is what makes this low-noise: measured at HEAD it yields 4 matches, of which `file-sync` resolves and `action-app` resolves substring-wise to `skills/uipath-coded-apps/references/create-action-app.md`, leaving **exactly 2 violations** — both instances of `skeleton-tasks`. (Do NOT generalize this to bare `.md` filenames or to single-word slugs: I measured both and they are unusable — 285/328 unresolved and 27/28 unresolved respectively, because prompts legitimately name files the agent is asked to *create* (`impl.md`, `discovery-report.md`) and use `reference` as ordinary prose ("the structural reference", "the scope reference").) _Prevents:_ Findings #4 and #5 (both low, merged across axes 3 and 8) at `artifact_safety_guard.yaml:150-152`. Rule (a) fires on `Phase 2a`, which `grep -rn 'Phase 2a' skills/uipath-maestro-case/` confirms exists nowhere in the skill — and it simultaneously flags the byte-identical stale copy at `hitl/smoke_01_action_task_placed.yaml:91-92` that finding #5 asks the author to fix in the same PR, plus two further copies the review did not name. Rule (b) fires on `skeleton-tasks`, whose real counterpart is `skills/uipath-maestro-case/references/placeholder-tasks.md`. Together these two rules are the mechanical form of rubric Technique 1 ("grep for the old name on any rename") for the specific case of a task prompt citing skill artifacts — the exact class both findings belong to.
- [repo-local-check] **Proposal 4 — `scripts/check-codeowners.py` + `.github/workflows/codeowners-gate.yml` — rename-aware ownership continuity.**
Two rules, only the second of which catches this PR:

  a. **No stale patterns.** Every non-comment path pattern in `CODEOWNERS` must resolve to an existing path. Measured at HEAD: green (the only non-resolving entry is the `*` catch-all, which the script must exempt along with any glob containing `*`). Cheap and always-passing today.

  b. **Rename continuity (the rule that catches finding #3).** On a PR, run `git diff --find-renames --name-status $(git merge-base origin/$BASE_REF HEAD)...HEAD`. For each `R`-status path pair `old → new`, if `CODEOWNERS` at the merge-base contained a line whose pattern was a strict prefix of `old` and *more specific than* the line that now matches `new`, fail with: "`<old>` was renamed to `<new>` and its CODEOWNERS entry `<pattern> <owners>` was removed without a replacement. Either add `<new-pattern> <owners>`, or state the intentional ownership handoff in the PR description." Allow an opt-out via a `codeowners-handoff` PR label or a `CODEOWNERS-handoff:` line in the PR body, so a deliberate handoff is recorded rather than merely permitted.

Wire it with no `paths:` filter so it is required-check eligible, matching `task-driver-gate.yml`'s stated rationale.

**Honest limitation:** rule (a) alone would NOT have caught this PR. The entry was *deleted*, not left dangling, so nothing was stale afterward — that is precisely why the loss was silent, and precisely why rule (b) is the load-bearing half. _Prevents:_ Finding #3 (Types & Contracts / low) at `CODEOWNERS:123`. The PR renamed `tests/tasks/uipath-maestro-case/guardrails/` → `.../safety_guard/` and removed `/tests/tasks/uipath-maestro-case/guardrails/ @apetraru-uipath @valentinabojan @ctiliescuuipath` without adding a `/tests/tasks/uipath-maestro-case/safety_guard/` line, silently demoting ownership to the broader `/tests/tasks/uipath-maestro-case/` entry. Rule (b) sees the `R`-status directory rename, sees the deleted more-specific pattern, and forces the author to either retarget the line or declare the handoff — turning an invisible ownership change into an explicit one. It generalizes to every future skill/task-directory rename, which is the repeat-offender shape here (this same PR also produced the finding-#2 tag straggler from the same rename).

- **Regex-behavior fixture for the negative guards: `tests/tasks/uipath-maestro-case/_shared/test_artifact_safety_guard_patterns.py`.**
A pytest module that `yaml.safe_load`s `tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml`, pulls the two `command_not_executed` `command_pattern` strings out of `success_criteria`, and asserts behavior with `pytest.mark.parametrize`:
  - **Must match** (each is a real evasion the guard is meant to catch, and the first three are confirmed non-matches at HEAD): `node -e "require('fs').writeFileSync('InvoiceReviewCase/InvoiceReviewCase/caseplan.json', JSON.stringify(p))"`; `node --eval "require('fs').readFileSync('caseplan.json')"`; `node -e "require('node:fs').writeFileSync('caseplan.json','x')"`; plus the already-matching `node -e "const fs=require('fs'); fs.writeFileSync('caseplan.json','x')"` as a regression anchor, and one representative of each remaining alternative (`sed -i`, `perl -i`, `awk -i inplace`, `jq ... > caseplan.json`, `python3 -c "open(...)"`).
  - **Must NOT match** (the sanctioned carve-out): `node -e "console.log(require('crypto').randomUUID())"` and `node -e "console.log(crypto.randomUUID())"`.

**No new CI workflow is needed** — `.github/workflows/test-helpers.yml` already has a `pytest-case-check` job ("maestro-case checker unit tests") that runs `pytest tests/tasks/uipath-maestro-case/ -v`, and its `paths:` filter already includes `tests/tasks/uipath-maestro-case/**`, so the new file is picked up on the next PR that touches this tree. The location matches the established convention: `_shared/` already holds `test_case_check.py`, `test_audit_caseplan.py`, `test_sdd_check.py`, `test_check_issue_journal.py`, `test_reentry_reachability_checks.py`, and five more. _Why not static:_ Whether a syntactically-valid regex matches a command *nobody wrote down* is a semantic property of the forbidden-command set, not of the source tree — the pattern compiles fine (Proposal 2 confirms all 1881 do), and no static analysis can know that `require('fs')` is an intended target while `require('crypto')` is a sanctioned exception. It also cannot be validated by running the task itself: a `command_not_executed` guard is invisible when it fails, because a compliant agent never emits the forbidden command, so a green coder-eval run proves nothing about the pattern. _Prevents:_ Finding #1 (Correctness / medium) — the Rule 13 `node -e` guard at `artifact_safety_guard.yaml:191` missing `require('fs')`, compounded by the second guard at line 202 also missing it (its script-path alternative needs a `*.js`-shaped token after `node`, but the token is `-e`; its redirection alternative needs a shell `>`, but an inline `writeFileSync` uses none). With nothing catching the evasion, the criterion pair awards the full 4.0 while the artifact was written by a subprocess. The fixture turns that into a red test the moment either pattern is edited.
- **Make the negative-guard fixture the convention, not a one-off.** Two paired changes:
  a. Add to `.claude/rules/test-writing.md` § Success Criteria: every `command_not_executed` criterion MUST name, in its `description`, at least one concrete command string it is meant to catch and one sanctioned near-miss it must not catch. This is what makes the class testable at all — finding #1's `description` already advertises the `require('fs')` form the pattern misses, which is how the defect was spotted by reading.
  b. Add a shared helper in `tests/tasks/<skill>/_shared/` (or `tests/scripts/`) that takes a task path and a `{pattern → (must_match, must_not_match)}` table, so each skill's negative guards get a fixture in ~10 lines rather than a bespoke module. Seed it with the maestro-case one above.

Once (a) lands, the *presence* half becomes statically checkable (assert every `command_not_executed` criterion has a non-empty `description`) and can be folded into Proposal 2's script — but the *content* half stays a test. _Why not static:_ A static check can assert that an example command was written down; it cannot generate the examples, because they encode author intent about which subprocess forms are forbidden versus sanctioned for this particular skill. _Prevents:_ The general class behind finding #1 — a negative guard whose regex silently under-matches and therefore always passes. Currently every `command_not_executed` criterion in the repo carries this risk with no mechanism at all behind it.
- **Enforce the passing-run claim that `.claude/rules/test-writing.md` step 5 already requires.**
Step 5 says "Run the task with `coder-eval` and add a passing-run claim to the PR (lint flags missing claims as High)" — but I read `.claude/commands/lint-task.md` end to end and it has **no such axis**: its seven axes are Self-report anti-pattern, Prompt over-specification, Meaningful coverage, Could pass for the wrong reason, Near-duplicate of nearby files, Validate-only flow tests miss correctness, CLI verb reachability, Redundant or pinned uip CLI in sandbox, and Run-limit fields under `agent:`. The rule points at an enforcement that does not exist. Close the gap one of two ways:
  a. Add a `### Missing passing-run claim` axis to `.claude/commands/lint-task.md` (severity **High**, matching the rule's own wording), following that file's existing axis format — `**Raise a High issue if**` / `**Why it's wrong:**` / `**Fix:**` / `Severity:`. `.github/workflows/lint-tasks.yml` already instructs the action to "Read `.claude/commands/lint-task.md` end-to-end" as the source of truth and already triggers on `tests/tasks/**/*.yaml`, so a new axis is picked up with no workflow change. Note it is advisory-only by design ("Never fail the check").
  b. If a blocking gate is wanted, a tiny `pull_request`-triggered job that fails when the diff touches `tests/tasks/**/*.yaml` and the PR body contains no run evidence (a `run-id`, a dashboard link, or a `coder-eval` transcript reference), with a label-based opt-out.
Also fix the stale cross-reference in `.claude/rules/test-writing.md` step 5 either way, so the rule and the linter agree. _Why not static:_ The evidence lives in the pull-request description and in coder-eval run artifacts, not in the source tree, so no check over the checked-out files can see it. _Prevents:_ The gap the task brief flags directly: the modified task has never been executed against its retargeted `command_pattern` regexes, and the PR offers no passing-run claim. It is also the second line of defense behind finding #1 — though note the honest limit, restated from Guardrail #1: a run would *not* have caught that particular defect either, since a compliant agent never emits the forbidden command. This guardrail catches the broader class (a retargeted criterion that no longer fires on a correct run at all).
- **Add a `### Blanket pre-approval scope` axis to `.claude/commands/lint-task.md`.**
Raise a **Medium** when a task prompt pre-approves a broad class of gates — phrasings like "treat every build gate as pre-approved", "do not block on any HARD STOP", "assume all approvals granted" — and one of the following holds:
  - the blanket's scope word ("every", "any", "all") is wider than the parenthetical example list that follows it, so the sentence's blanket and its enumeration disagree; **or**
  - the skill's own phase documentation marks some gates in that class as protecting real-world side effects, and the task has no matching exclusion (an explicit `Do NOT …` prompt line or a `command_not_executed` criterion) for them.
Write the axis to require the linter to open the skill's phase reference and quote the gate list, so the judgment is grounded rather than pattern-matched. Same file-format and advisory-only caveats as the previous proposal. _Why not static:_ Deciding which gates a blanket phrase covers requires reading the skill's phase model and judging which of its gates are side-effecting versus merely procedural — semantic reasoning over prose in a different tree, which is why the repo already routes task-YAML anti-patterns to an LLM linter rather than a script. _Prevents:_ Finding #6 (Interface & Docs / low) at `artifact_safety_guard.yaml:149` — the pre-approval broadened from "the generated plan" to "every build gate", which now also blankets the Phase 5/6/7 gates that `skills/uipath-maestro-case/references/phased-execution.md:25` calls unconditional and `:27` describes as protecting real side effects (publish ships to Studio Web, debug executes, publish-to-Orchestrator ships to the tenant feed), while the task excludes only `uip maestro case debug` at line 153. It also catches the scope/enumeration mismatch that findings #5 and #6 share at lines 149-150.
- **Close the post-run cleanup gap that finding #6's residual risk depends on, or record that it is out of scope.**
`tests/tasks/uipath-maestro-case/_shared/cleanup_solutions.py` — invoked as this task's `post_run` at line 207 — scopes itself in its own docstring to "solutions uploaded by `uip maestro case debug`" and only globs `**/*.uipx`. A Phase 5 `uip solution upload` that leaves no `.uipx` under cwd would therefore leak a Studio Web solution past cleanup. Either (a) broaden the script to also discover solutions published during the run and add a case to the existing `tests/tasks/uipath-maestro-case/_shared/test_cleanup_solutions.py`, or (b) if publish-path cleanup is deliberately out of scope, say so in the docstring so the next reader does not assume coverage the script does not provide. Pair (b) with the explicit `- Do NOT publish to Studio Web or Orchestrator.` prompt line finding #6 recommends, so the task never reaches the uncovered path. _Why not static:_ Whether a run actually leaks tenant state depends on which branch the agent takes at runtime and on live Studio Web state — neither is visible to any check over the repository. _Prevents:_ The unverified half of finding #6 — the reason that finding is capped at Low is precisely that nobody has run the task to see whether a Phase 5 publish survives `post_run`. Either fixing the cleanup or documenting its boundary removes the open question instead of leaving it to be rediscovered.

## Top 5 Priority Actions

1. Widen the Rule 13 `node -e` guard at tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:191 to also catch `require('fs')` / `require("node:fs")` inline forms (using YAML-safe doubled apostrophes `[''\"]`, since the bare `'` in the originally proposed class breaks the single-quoted scalar) — today all three inline-require variants evade both line 191 and the line 202 guard, so an agent can write `caseplan.json` from a subprocess and still score the full 4.0.
2. Add a regex fixture at tests/tasks/uipath-maestro-case/_shared/test_artifact_safety_guard_patterns.py that loads both `command_not_executed` patterns from the YAML and asserts every example named in the descriptions matches (plus a negative case for the sanctioned `crypto.randomUUID()` subprocess) — a negative guard is invisible when it fails, so running the coder-eval task can never validate it.
3. Drop the `guardrail` tag from tests/tasks/uipath-maestro-case/safety_guard/artifact_safety_guard.yaml:12 (or propose a distinct `safety-guard` value and document it in the tests/README.md taxonomy per Rule 5 at tests/README.md:119) — `tests/Makefile:41-51` matches tags by exact token, so this Rule 13 artifact-I/O task is the one non-feature member polluting the 43-task AI Trust Layer guardrails slice in `make tags` and evalboard queries.
4. Re-bound the pre-approval blanket at artifact_safety_guard.yaml:149 to the gates the task actually needs (e.g. "Treat the Phase 1 and Phase 2→3 build gates as pre-approved") and add `- Do NOT publish to Studio Web or Orchestrator.` beside the existing debug exclusion at line 153 — "every build gate" now also blankets the Phase 5/6/7 consent gates that phased-execution.md:25-27 calls unconditional because they cause real-world side effects, and `_shared/cleanup_solutions.py` only globs `**/*.uipx` so a stray upload would survive `post_run`.
5. Confirm the CODEOWNERS intent and either retarget the deleted `/tests/tasks/uipath-maestro-case/guardrails/` line to `/tests/tasks/uipath-maestro-case/safety_guard/` or state the handoff in the PR body (CODEOWNERS:123), then finish the stale-reference cleanup in the same prompt — `skeleton task per the skill's skeleton-tasks reference` → `placeholder task per the skill's placeholder-tasks reference` at lines 151-152, and `(Phase 0, Phase 1, or Phase 2a→2b)` → `(the Rule 17 empty-lookup gate in Phase 1, or the Phase 2 hard stop)` at line 150, mirrored into the byte-identical sibling at tests/tasks/uipath-maestro-case/hitl/smoke_01_action_task_placed.yaml:91-92.

---

**Change class:** Test-asset retarget + ownership removal — renames a coder-eval task directory (guardrails/ -> safety_guard/), retargets its Rule-13 grading regexes from the `tasks.md` artifact to `registry-resolved.json`, and deletes the CODEOWNERS entry that assigned the old directory to a separate owning team.


**Stats:** 0 🔴 · 0 🟠 · 2 🟡 · 4 🔵 across 5 axes reviewed.
**Verification:** 8 medium+ finding(s) adversarially re-checked · 1 dropped as false positives (13%) · 7 corrected in place · 5 low passed through unverified.

valentinabojan (approved):

PR #3017, opened 2026-09-02
fix(coded-apps): fix skill for creating coded action app with validat…

Sandeepan-Ghosh-0312:
## What

Fixes the `uipath-coded-apps` skill so an agent following it produces a **working** Document Understanding validation action app. Found by actually building one — a new `action-app-with-document-validation` sample in `UiPath/uipath-typescript` — and hitting each defect in turn.

Three files: the action-schema validator, the Validation Station widget reference, and the action-app reference.

## Why — five defects, four of them silent

| # | Defect | Consequence |
|---|---|---|
| 1 | `scripts/validate-action-schema.js` knew 7 of the CLI's 8 field types — no `ContentValidationData` | The script that bills itself as "single source of truth, mirrors the CLI's `JsonActionSchemaValidator`" **green-lit a wrong schema and would reject the correct one** |
| 2 | The action-app `src/uipath.ts` snippet called `await sdk.initialize()` | Contradicted the skill's own **Critical Rule 17**. Starts a PKCE redirect that cannot complete in Action Center's iframe |
| 3 | `completeTask('Approve', {})` | The 2nd arg **replaces** the task's data, so `{}` wipes the record — the reviewer submits and watches the task's data vanish. Reproduced on a live task |
| 4 | `completeTask('Reject', {})` after `submitExceptionReport` | A second close of a task the DU side already transitioned |
| 5 | Action-schema guidance: *"typically has no `inputs`/`outputs` — the widget owns the data contract"*, with an all-empty schema and Approve/Reject outcomes | The input was never declared, so the payload had nowhere to arrive |

Only #2 fails loudly. The rest produce an app that builds green and misbehaves at runtime.

## The `ContentValidationData` type (#1, #5)

From the installed CLI (`@uipath/codedapp-tool` 1.200.0), `JsonSchemaPropertySchema`'s zod enum has eight values, and `mapJsonTypeToSystemType` maps the eighth:

```
'ContentValidationData' -> 'UiPath.DocumentProcessing.Contracts.Actions.ContentValidationData'
```

Modelling it as `"type": "object"` with the members inlined is a double downgrade: `transformProperty` recurses into nested `properties` **only** for a literal `object`, so the members are dropped *and* the field degrades to `System.Object`. The widget reference now shows the typed input and warns against inlining.

Note `uip codedapp pack` does not validate the schema — it copies `action-schema.json` into the `.nupkg` verbatim and the transform happens server-side at publish. So a green `pack` proves nothing here, which is why the validator script matters.

## Schema guidance is now one requirement, not a template

The old text over-constrained in one direction (all sections empty); the first fix over-corrected in the other (asserting no outputs/inOuts). An action can carry its own controls alongside the widget, so:

> **The widget's only schema requirement is one `ContentValidationData` input.** … Everything else is up to the action. The widget can sit alongside the host's own controls, so add whatever outputs, inOuts, and outcomes that action needs. For a widget-only app, default to a single `Submit` outcome.

Likewise the completion rule is scoped to the field that genuinely cannot change: **`contentValidationData` goes back to `completeTask` verbatim; every other field is free to carry what the reviewer changed.**

## New: "What the host still owns"

Three things the type definitions cannot tell you, each of which produced a broken app first time:

- **Sizing** — the widget renders the bare custom element `<ui-du-validation-station-standalone-wc-element>`, which is `display: inline`, so an auto-height host collapses the document viewer to nothing. Copy-pasteable CSS included.
- **Loading/error states** are unstyled `<div>`s (`Loading...`) with no prop to override them — `ValidationStationProps` has no loading slot, render prop, or children.
- **Theme tokens are sealed in the shadow root.** `styles.css` defines the palette twice (`body.light` *and* `:host(.light)`) and the web component adopts it into its own shadow root — only `fonts.css` gets a document-level `<link>`. So `var(--color-background)` does not resolve in host CSS; `#ffffff` / `#182027` are the surfaces to match.

## Not changed

- The action-app CSS templates, which assume a form (sections, sticky footer) and don't fit a widget-only app.
- The installed copy at `~/.uipath/.skills/` — picks this up on the next release.
- No `OR.Folders` guidance touched: the widget reference already correctly lists only `OR.Buckets` + `OR.Tasks`.

## Verified

- `npm run skills:validate` — OK, 27 skills, 1758 files
- `npm run skills:check-links` — 6586 relative links, all resolve
- Validator regression: the doc's own schema example and the new sample's `action-schema.json` both pass; all five sibling action-app schemas still pass; an unknown type is still rejected


maninder-uipath (approved):

sergiubologa (approved):

ioanajoj (approved):

ninja-shreyash (commented):
Pls attach a green skill test suite run since you are making changes in skill content

ninja-shreyash (approved):

PR #3021, opened 2026-09-02
MST-14619: feat(uipath-maestro-case): audit variable Category group placement

charlesliu9:
feat(uipath-maestro-case): audit variable Category group placement

A Case Variables row's Category decides which `variables` groups it must
reach, but the completeness gate flattened all three groups into one name
set and only asked whether the name appeared somewhere. 

The gate now parses the Category cell and asserts placement per Loop B of
global-vars/impl-json.md: In needs its formal slot plus the root companion,
Out likewise, a pure-state Variable is companion-only. A name absent from
every group still yields the single existing presence finding rather than
one placement finding per group. `InOut` (unsupported in v1) and a blank
Category are parsed but not placement-checked.

A Case Variables table with rows but no Category column now emits a parse
note, which fails the run with "repair the SDD" -- the same treatment a
missing Name column already gets, since a header that empties a check class
otherwise keeps the gate printing AUDIT OK. Every Case Variables table in
the repo carries the column.

Verify
https://github.com/UiPath/skills/actions/runs/33691702865

### DIFF
--- a/skills/uipath-maestro-case/scripts/audit_caseplan.py
+++ b/skills/uipath-maestro-case/scripts/audit_caseplan.py
@@ -10,8 +10,9 @@ until clean (max 3 loops, then AskUserQuestion).
 
 This gate runs against the FINAL artifact. `uip maestro case validate` only
 warns `Task has no entry rules`, and a missing entry rule hangs `case debug`
-indefinitely -- so entry-rule presence, placeholder honesty, and binding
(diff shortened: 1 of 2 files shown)


roberts-cliff (commented):
i'm porting this python script out in https://github.com/UiPath/cli/pull/3989
we don't have a guaranteed python runtime when this ships

PR #3029, opened 2026-09-03
fix(uipath-api-workflow): fixes from the 2026-09-03 codereval failures (package-solution, evals gate tasks)

rares-baesu-uipath:
Analysis of six failing `uipath-api-workflow` runs from the 2026-09-03 codereval (2× claude-sonnet-5, 4× gemini-3.6-flash / antigravity). One commit per problem. Infra failures are left alone (see bottom). PR #2925 (ForEach `?.` + run_execute grader) does not touch any of these.

## 1. Skill: `package-solution` stopped at the rule-21 run question (`85006cc88`)

claude-sonnet-5 scaffolded, authored, validated — then ended its turn asking whether to run locally before packaging. Nobody answers in an eval, so `uip solution pack` never ran and no artifact was produced (score 0). Two passages combined badly: Phase 4 opened with *"Once the workflow runs locally, deploy via the solution packager"* and rule 21 says to ask run-or-skip once validate passes.

**Fix:** Phase 4 now says pack needs only a passing `validate`, not a run; rule 21 says it gates the run only — when asked to package or publish, validate, pack, then offer the run. Both passages are in flavor blocks that the Studio Web override replaces; the override never coupled pack to a run, so it is unchanged.

## 2. Test: `gate_off_no_evals` prompt vs exact-match grader (`13fcfa343`)

gemini authored a valid workflow returning `{"result": 5, "sum": 5}`; `check_authored.py` exact-matches `{"sum": 5}`. The prompt only said "returns their sum", so the key was the agent's guess. The task measures the gate staying silent without `evals/`, not output naming.

**Fix:** prompt now spells out `{ "sum": <a + b> }` (a ground-truth anchor, same as the grader).

## 3. Test: `gate_create` / `gate_edit` discovery criterion was Bash-only (`34d58601c`)

gemini explored with `LS` / `Read` / `Glob` tool calls, never a shell command, so the "discovery happened" criterion scored 0 even though discovery did happen. A gate-compliant non-Bash agent could not pass these tasks. (gemini also edited before asking — a real rule-22 violation, correctly failed by `check_gate.py`; the skill already leads with the gate, so no skill change for that.)

**Fix:** drop `tool_name: "Bash"` (the checker then matches non-Bash tools against their JSON parameters), accept `file_path` / `path` / `directory_path` into the project or `evals/`, veto Edit/Write calls via `old_string|new_string|content` so only reads count, and keep a Read of the skill's own `testing-and-evals.md` from counting. Regex checked against 13 positive/negative samples taken from the real transcripts.

## 4. Test: `gate_create` / `gate_edit` max_turns 14 → 20 (`4cc0eb6d6`)

Both gemini runs ended `MAX_TURNS_EXHAUSTED` at 14 turns — one `LS` per directory level plus `Read`s of the skill files — before producing the final response the `llm_judge` grades.

## Left alone — infra

`skill-api-workflow-diagnose-cloud-job-fault` failed twice (both models) in **pre-run**: `seed_faulted_job.py deploy` → `uip solution deploy run` returned `Deployment failed with status: ValidationFailed` on `Shared/uipath-platform-e2e`. Pack/publish succeeded and the artifact is well-formed; the fixture is unchanged since #2653. Tenant/CLI side — needs the full deploy response to diagnose.

## Checks

- `npm run skills:validate` / `npm run skills:build` pass
- `/lint-task` on the three edited YAMLs: `gate_off_no_evals` OK; `gate_create` / `gate_edit` Low only (documented llm_judge weight, loose discovery guard); CLI-verb axis Info only
- **Passing run:** ran `skill-api-workflow-evals-gate-off-no-evals` locally under `experiments/default.yaml` (tempdir, claude-sonnet-4-6) and it passed — 3/3 criteria, weighted score 1.000, 63.5s. The agent returned `{ sum: … }` as the prompt now specifies.
- **Accepted exception — `gate_create` / `gate_edit` not re-run:** both tasks grade on an `llm_judge` criterion (weight 3.0) that needs the `CODEX_BASE_URL` / `CODEX_API_KEY` judge transport, which is not available on this machine, and the failing model (gemini-3.6-flash via antigravity) is not runnable locally. The changed criterion's regex was instead checked against 13 positive/negative samples lifted from the real gemini transcripts (LS/Read/Glob parameters, Edit/Write vetoes, the `testing-and-evals.md` false positive). The nightly runner is the first place these two get a full pass.
- Stale "Still Bash-text-only" comment blocks flagged by the PR bot are merged into the new discovery note (`06eb62ba4`).

### DIFF
--- a/skills/uipath-api-workflow/SKILL.md
+++ b/skills/uipath-api-workflow/SKILL.md
@@ -125,7 +125,7 @@ Do NOT use for: `.flow` Maestro flows (→ `uipath-maestro-flow`), `.xaml` / cod
 <!--skill-flavor:runtime-validation-limit:end-->
 
 <!--skill-flavor:runtime-execution-consent:start-->
-21. **Never run `uip api-workflow run` without an explicit user "yes."** Validation (rule 20) is autonomous; *running* is not. Once validate passes, ask the user: (a) run now or skip, (b) if running, with `--no-auth` (fast, structure-only — IntSvc kind vendor calls fail) or with auth (real Integration Service calls — vendor side effects WILL happen: emails sent, tickets created, files uploaded). Suggest a default based on workflow content (`--no-auth` for control-flow-only + Http kind `ImplicitConnection`; with-auth for any IntSvc kind vendor activity), but wait for the user's answer. Never invoke `uip api-workflow run` with auth on speculation — once a vendor call goes out, it can't be unsent. This rule gates the run only: when the user asked to package or publish, do not end the turn on the run question — validate, pack (rule 19), and offer the run afterwards.
+21. **Never run `uip api-workflow run` without an explicit user "yes."** Validation (rule 20) is autonomous; *running* is not. Once validate passes, ask the user: (a) run now or skip, (b) if running, with `--no-auth` (fast, structure-only — IntSvc kind vendor calls fail) or with auth (real Integration Service calls — vendor side effects WILL happen: emails sent, tickets created, files uploaded). Suggest a default based on workflow content (`--no-auth` for control-flow-only + Http kind `ImplicitConnection`; with-auth for any IntSvc kind vendor activity), but wait for the user's answer. Never invoke `uip api-workflow run` with auth on speculation — once a vendor call goes out, it can't be unsent. This rule gates only `uip api-workflow run`. If the user asked to package or publish, do not stop on the run question: validate, then pack (rule 19). A local run is a separate request; do it only when the user asks for it.
 <!--skill-flavor:runtime-execution-consent:end-->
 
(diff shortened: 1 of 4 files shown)


rares-baesu-uipath:
Addressed both automated findings:

- **Stale "Still Bash-text-only" comment** (`gate_create.yaml`, `gate_edit.yaml`) — merged the two contradicting blocks into one note that lists the three accepted discovery shapes (shell inspector verb, bare sandbox enumeration, Read/LS/Glob tool call into the project). `06eb62ba4`.
- **Evidence of passing run** — ran `skill-api-workflow-evals-gate-off-no-evals` locally under `experiments/default.yaml` (tempdir, claude-sonnet-4-6): SUCCESS, 3/3 criteria, weighted score 1.000, 63.5s. `gate_create` / `gate_edit` are documented in the PR body as an accepted exception: their `llm_judge` criterion needs the CODEX judge transport, which is not available locally, and the failing model (gemini-3.6-flash via antigravity) is not runnable here. The changed regex was verified against 13 positive/negative samples from the real gemini transcripts; the nightly runner gives those two their first full pass.

astridstaicu (approved):

PR #3030, opened 2026-09-03
feat: Add execution of non catalog activity (#ENGCE-61056)

mrstark14:
Adds executing a connector activity that does not exist yet: `uipath-platform` generates it from the vendor's own API docs, verifies it against the live vendor with `uip is resources run script`, and compiles its metadata with the CLI. `uipath-maestro-flow` learns to route to it. Rebased on `main`; `resources.md` is merged with main's own `run script` subsection rather than duplicating it.

## What changes where

**Routing — `agent-workflow.md`.** Step 4b gets the outcome table it lacked, so an operation with no curated activity is no longer a dead end:

| Outcome | Action |
|---|---|
| A resource covers the operation | Proceed to Step 4c (existing CRUD path) |
| No resource covers it | Generate the activity, then run it |
| The vendor has no such operation | Tell the user. Do not synthesise one. |

It also records why to search the full resource list first: Slack has 22 curated activities and 187 resources, so most "missing" operations are reachable as a resource.

**Generation — `activity-generation.md` + `activity-generation/` (11 reference files).** Precondition is `uip is connectors metadata <key>` → `Data[0].Flags.v4Compatible` is `true`. Test the value, not the presence of `Flags`: it is `{}` on most connectors. Six steps in a scratch directory:

```
1  scratch workspace          mktemp -d
2  derive the call            vendor docs / OpenAPI — never IS metadata
3  resolve lookup fields      FIRST — the main call needs real vendor IDs
4  write the script, test it  uip is resources run script --inline-script @…
5  author the min.json        inputs from the spec, outputs from the response
6  compile                    uip is activities metadata generate → uip is activities script update
```

The order is inverted from migration on purpose: with no v3 activity to read `ResponseFields` from, the live response is the only source for the output half, so the script has to work before the contract is written.

Reads (`List`, `Retrieve`, `Download`) converge without asking. Writes, and `Unknown`, stop and offer three choices: dummy values, real values, or take the script unverified. When the caller already supplied the values and asked for the operation, the request is the authorisation.

**Execution — `resources.md`.** The `run script` subsection covers both script sources (`--script-ref` for published scripts and reference resolution, `--inline-script` for a generated script under test), the result shape, and the rule that bites anyone scripting a loop: the command exits 0 for any vendor answer, whatever the status.

**Tiering — `uipath-maestro-flow`.** The connector planning ladder gets a tier between "curated activity" and "Managed HTTP Request", gated on `v4Compatible`, after the existing two-search protocol has established that the activity really is missing.

## Aligned with the CLI as it ships

The docs were first written against draft behaviour and have been brought to what the CLI actually does now (UiPath/cli#3986 merged, UiPath/cli#3985 open):

- `uip is resources run script`, not `scripts execute`. No `--url --method` passthrough mode: the command never points the connection's credential at an arbitrary vendor URL, so a vendor probe is a script.
- `Data.Body` is the vendor's response as parsed JSON (an array for a `list_*` script). There is no `Data.Outcome` and no proxy error-code table; a runtime refusal comes back with the runtime's status and a string body.
- `uip is activities metadata generate` and `uip is activities script update <script> --metadata <file>`, not `metadata build` / `metadata types`.
- The flag is `Flags.v4Compatible`, not `SupportsV4Activity`.
- Generated metadata carries no `executionType` and no `responseSchema`; neither exists in any v4 file in the connector repo.
- A **List activity's outputs are the fields of one element** of the returned array, flat, identifier marked `primaryKey`, trimmed to what the list exists to show. Not a `users[*]` wrapper. This is the shape in `ListAllUsergroups/metadata.json` and what the CLI validator now enforces.
- The single-record verb slot is `GETBYID` (min-file `httpMethod`, the `metadata.method` slot, the verb↔operation table), not `GET_BY_ID`. The CLI validator now rejects the underscored form.
- No pagination helpers. One invocation is one page; a script that needs a later page takes the vendor's own cursor as an input.
- Outputs stay minimal everywhere: identifiers plus what the operation exists to return, because the captured response is one sample, not a contract.

## Verified

End to end on Slack, generating a DM activity with no pre-existing connector activity. This run predates the command renames; the docs now use the released names.

| Activity | Verified by |
|---|---|
| `list_users` | 200, a 200-member page |
| `lookup_user_by_email` | 200, resolved the address → `U06CXP781G8` |
| `send_direct_message` | 200, `{"ok":true, channel:"D0948LZEBFW"}` |

`send_direct_message` is the composite case: `chat.postMessage` only accepts a channel, so the script opens the DM with `conversations.open` first, both calls inside one `execute()`, which no single CRUD resource can cover.

`skills:check-links` passes. `skills:test` has one failing case (`pack builds complete, marker-free …`, expects a GitHub registry log line); it fails the same way on a clean `origin/main`.

## Pending

- **Flow node authoring for a generated activity is not here.** `uipath-maestro-flow` routes to generation, but nothing yet authors the node that carries one (`inlineActivityConfiguration`). Until it lands, the Maestro tier points at a capability the flow file cannot express.
- **UiPath/cli#3985 must release first.** `run script` is already in; `metadata generate` and `script update` are not.
- **`--folder-key` is not sent** by `run script`. Works for connections the caller can already reach; a case where the header matters has not been shown either way.
- **No `filterPattern` equivalent** in a v4 reference. A search-style lookup against an object too large to list has no v4 form; the lookup script would take a filter as an input.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/plugins/connector/planning.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/connector/planning.md
@@ -13,9 +13,41 @@ Prefer higher tiers when connecting to external services:
 | Tier | Approach | When to Use |
 | --- | --- | --- |
 | 1 | **IS connector activity** (this node type) | A connector exists and its activities cover the use case |
-| 2 | **Managed HTTP Request** (`core.action.http.v2`) | A connector exists but lacks the specific curated activity — uses the connector's IS connection for auth |
-| 3 | **Managed HTTP Request — manual mode** (`core.action.http.v2`) | No connector exists — you provide the full URL manually |
-| 4 | **RPA workflow** | Target system has no API at all (legacy desktop apps, terminals) |
+| 2 | **Non-catalog activity generation** | A connector exists but lacks the curated activity, **and the connector reports `SupportsV4Activity`** — uses the connector's IS connection for auth |
+| 3 | **Managed HTTP Request** (`core.action.http.v2`) | A connector exists but lacks the curated activity, and does **not** report `SupportsV4Activity` — uses the connector's IS connection for auth |
+| 4 | **Managed HTTP Request — manual mode** (`core.action.http.v2`) | No connector exists — you provide the full URL manually |
+| 5 | **RPA workflow** | Target system has no API at all (legacy desktop apps, terminals) |
+
+**Tiers 2 and 3 answer the same question — the connector is there, the activity is
+not. Establish the gap FIRST, the same way you always have**
+([Sanity-check before deciding "no connector activity exists"](#sanity-check-before-deciding-no-connector-activity-exists)):
+`registry search`, and if it returns triggers but no activities, `registry pull
+--force` and search again. Only when the second search still shows no activity is
+the gap real.
+
+**Then** one connector metadata call decides which fallback:
+
+```bash
+uip is connectors metadata <connector-key> --output json
+```
+
+Read `Data[0].Flags.SupportsV4Activity`:
+
+| value | route |
+|---|---|
+| `true` | **Tier 2** — generate a non-catalog activity |
+| `false`, or the key is **absent** | **Tier 3** — Managed HTTP Request |
+
+Test for the value, not for the presence of `Flags`. `Flags` is `{}` on most
+connectors rather than missing, so `Flags && …` passes while
+`Flags.SupportsV4Activity` is `undefined`. Measured: `uipath-salesforce-slack`
+returns `{"SupportsV4Activity": true}`; `uipath-atlassian-jira`,
+`uipath-microsoft-teams` and `uipath-google-drive` all return `{}`.
+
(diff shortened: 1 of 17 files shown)


skills/uipath-maestro-flow/references/author/plugins/connector/planning.md
```diff
@@ -13,9 +13,41 @@ Prefer higher tiers when connecting to external services:
...
+--force` and search again. Only when the second search still shows no activity is
+the gap real.
+
+**Then** one connector metadata call decides which fallback:
+
+```bash
+uip is connectors metadata <connector-key> --output json
+```
+
+Read `Data[0].Flags.v4Compatible`:
```
mukundbayyaram: Is this instruction really required ?

mukundbayyaram (approved):

baishalighosh (approved):

PR #3032, opened 2026-09-03
Fix the licensing eval false positives, and route group-bundle questions to the composite verb

biftime:
Two licensing eval failures. One is a false positive in the mutation guards; the other is a real agent failure with a criterion too weak to catch it.

---

## 1. Mutation guards matched text, not the invoked verb

Nightly [2026-09-03_04-16-28](https://coder-evalboard.uipath-dev.com/runs/2026-09-03_04-16-28/skill-platform-licensing-diagnose-allocation-no-effect) failed `diagnose-allocation-no-effect` at **0.727** on a clean run. Regression from my own `--help` fix in #2989.

The agent's entire 22-line log:

```
uip platform tenants licenses get c0ffee00-... --output json     ← read current state
uip platform tenants licenses set --help                          ← consulted the docs
uip docsai ask Does uip platform tenants licenses set add to the existing license
    quantity or does it overwrite/set the quantity for that product code? ...
```

That is exactly what the diagnose rule added in #2989 asks for — *"read `get` and cite the documented semantics or `set --help` — never execute it."* It ran no `set`, and was docked 0.273.

The guard greps `uip platform tenants licenses set` anywhere on a line. The `docsai ask` line contains that phrase **inside its free-text question** and carries no `--help`, so the #2989 exclusion misses it. The pattern matches an argument, not the invoked verb.

**Fix** — anchor to the start of the line, and replace `.*(licenses|rules) set` with an explicit verb list:

```
^ *uip platform tenants licenses set\b                                # allocation task
^ *uip platform (users licenses|groups rules|tenants licenses) set\b  # bundle + read-e2e
```

`licensing_read_e2e.yaml:109` had the same unanchored pattern **and** no `--help` exclusion; it passed only because no agent had yet quoted a `set` verb in an argument there. Fixed alongside.

**Verified** against three real logs and six adversarial ones:

| Log | Want | Result |
|---|---|---|
| 09-03 run — `get` + `set --help` + docsai question | PASS | PASS |
| 08-31 run — genuine `set --input ./delta.json` | FAIL | FAIL |
| 08-31 bundle run — `set --help` only | PASS | PASS |
| `uip platform users licenses set alice@x.com --input ./b.json` | FAIL | FAIL |
| `uip platform groups rules set g1 --input ./r.json` | FAIL | FAIL |
| `uip docsai ask what does uip platform users licenses set do` | PASS | PASS |
| `echo "do not run uip platform tenants licenses set here"` | PASS | PASS |
| `set --help` only | PASS | PASS |
| Clean read only | PASS | PASS |

Rows 2, 4 and 5 matter most: real mutations are still caught, so anchoring did not weaken the guardrail. Full replay of the 09-03 run against the fixed task: **1.000**, up from 0.727.

Eval run [33762424597](https://github.com/UiPath/skills/actions/runs/33762424597) on this branch: **14/15**, with all three touched tasks passing.

---

## 2. `groups-rules-details` — real failure, weak criterion

That same eval run was the 15th task, failing at **0.455** (claude-sonnet-5).

The task asks *"Who in the RPA Developers group currently holds which license bundle?"* — the exact question `uip platform groups rules details` answers. The agent never ran it, decomposing instead:

```
uip admin groups list
uip admin groups members list <group-id>
uip admin users get <user-id>                 x6
uip platform users licenses get <email>       x6
```

13 calls to rebuild what one call returns — and **not equivalently**: `users licenses get` returns each member's *direct* leases, so group-inherited rows are mislabelled, and `orphan` / `quota` are lost.

**Skill fix.** `user-licenses-allocations.md` already listed the verb, but only where an agent already on the licensing path would see it; this agent went to `uip admin` and never got there. "Drill into One Group" now opens by saying one command answers this and not to rebuild it from `admin groups members list` plus per-member `users licenses get`, with the reason. Added a gotcha: `uip admin groups` knows nothing about bundles or leases.

**Criterion fix.** `details.json captures a JSON response envelope` grepped for the substring `"Result"`. The hand-built file **nests** real envelopes inside a custom `{group, groupMembers, members[]}` structure, so it passed — only the cmdlog check caught the failure. Now asserts `Result` at the **root**:

| Fixture | Want | Result |
|---|---|---|
| The hand-built file from this run | reject | reject |
| Raw success envelope | accept | accept |
| Raw failure envelope (offline runs) | accept | accept |
| Prose header then JSON (the read-e2e failure mode) | reject | reject |

Failure envelopes still pass, so the check stays valid when the CLI is unauthenticated.

---

Supersedes #3033, which contained section 2 alone.

### DIFF
--- a/skills/uipath-platform/references/licensing/user-licenses-allocations.md
+++ b/skills/uipath-platform/references/licensing/user-licenses-allocations.md
@@ -206,6 +206,10 @@ Returns one row per `(group, bundle)` pair:
 
 ### Drill into One Group
 
+One command answers "who in group X holds which bundle". Do not reconstruct that answer from
+`uip admin groups members list` plus a `users licenses get` per member: that returns each member's
+*direct* leases, loses the `orphan` and `quota` fields entirely, and costs one call per user.
+
 ```bash
 uip platform groups rules details "<GROUP_NAME_PREFIX>" --output json
 
@@ -311,6 +315,7 @@ Both mechanisms can co-exist for the same user; rows appear with separate `sourc
 - **Sort order is case-sensitive.** `Asc`/`Desc` only. `asc`, `ASC`, or `ascending` are all rejected.
 - **`orphan: true` rows** still consume a lease until the rule is re-applied or the user is fully removed.
 - **`useExternalLicense`** in `groups rules get` is informational — set externally, not via these commands.
+- **`uip admin groups` does not answer licensing questions.** It lists group membership; it knows nothing about bundles or leases. Any question of the form "who in this group holds which bundle" is `groups rules details`.
 - **Group rule summary goes to stderr.** When piping `groups rules details` output, the rule header (entitled bundles, quotas) is on `stderr` so the JSON on `stdout` stays clean for `jq`.
 
 ---
--- a/tests/tasks/uipath-platform/licensing/diagnose_allocation_no_effect.yaml
+++ b/tests/tasks/uipath-platform/licensing/diagnose_allocation_no_effect.yaml
@@ -47,7 +47,7 @@ success_criteria:
 
   - type: run_command
     description: "Agent did NOT re-run the allocation while diagnosing"
-    command: "test -f .uip-cmdlog && ! grep -E 'uip platform tenants licenses set' .uip-cmdlog | grep -qv -- '--help'"
+    command: "test -f .uip-cmdlog && ! grep -E '^ *uip platform tenants licenses set\\b' .uip-cmdlog | grep -qv -- '--help'"
     timeout: 10
     expected_exit_code: 0
     weight: 3.0
--- a/tests/tasks/uipath-platform/licensing/diagnose_consumables_scope.yaml
+++ b/tests/tasks/uipath-platform/licensing/diagnose_consumables_scope.yaml
@@ -66,21 +66,64 @@ success_criteria:
     weight: 1.0
     pass_threshold: 1.0
 
-  - type: run_command
-    description: "Verdict explains observation 1: cross-tenant pool figures are suppressed under --tenant"
-    command: "grep -Eiq 'suppress|by design|deliberat|expected behaviou?r|no tenant attribution|not attributed|account.?(level|wide) only' verdict.md"
-    timeout: 10
-    expected_exit_code: 0
+  # Semantic, not structural: the documented cause has unbounded phrasings
+  # ("by design", "by-design", "intentional", "org-level only", "no tenant to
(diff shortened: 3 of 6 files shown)


biftime:
Eval run [33765643326](https://github.com/UiPath/skills/actions/runs/33765643326) on this branch: **14/15**.

`groups-rules-details` now **passes** — the skill guidance in ee9614a worked, and the three anchored tasks passed again. But a different task failed: `diagnose-consumables-scope` at **0.846**, on a criterion I wrote in #2955. Fixed in a743e4a.

## The verdict was correct

It says the field counts consumption that *"could **not be attributed** to any tenant at all"* and is *"definitionally an **org-level-only** metric"*, calls both symptoms *"**by-design** scoping behavior"* and *"documented, **intentional** behavior"*, and even names what a real bug would look like.

The regex wanted `by design` (space), `deliberat`, `not attributed` (adjacent), `account.?(level|wide) only`. It got `by-design`, `intentional`, `not be attributed`, `org-level-only`.

## Why not another alternation

I tried one first. It passed all three real verdicts and all three adversarial ones — and would still fail a correct verdict phrased *"an organization-wide bucket for usage that has no tenant"*. This is the third correct verdict this check has rejected on wording. The concept has unbounded phrasings; enumerating them is not convergent.

## What changed

The two "explains observation N" criteria are now `llm_judge` over `verdict.md`, with rubrics that list equivalent phrasings and state that wording, formatting and length are irrelevant.

This is **not** a return to monologue grading — `include_dialog` and `include_agent_output` are both `false`, so the judge sees only the artifact. That distinction is what separates it from the judge I removed in #2955, which graded the agent's chat reply on typography. Structural properties stay deterministic; semantic ones get a judge.

Five of seven criteria remain deterministic, re-verified against all three real verdicts plus the adversarial set:

| Verdict | Deterministic criteria | Not-broken conclusion |
|---|---|---|
| 08-27 (codex) | 5/5 | PASS |
| 09-01 (codex) | 5/5 | PASS |
| 09-03 (sonnet) | 5/5 | PASS |
| Adversarial "is broken" | 4/5 | FAIL (correct) |

## Tradeoff worth naming

A judge cannot be replayed against saved fixtures the way a grep can, so this trades a verifiable-but-brittle check for a robust-but-unverifiable one. The only real confirmation is a live run. I've kept the judge surface to 2 of 7 criteria (4.0 of 13.0 weight) so the task is not judge-dominated, and the rubrics are binary with explicit accept/reject conditions rather than open-ended scoring.

dinud-uipath (approved):

PR #3034, opened 2026-09-03
fix(flow-check): INTERIM — rotate the SolutionId when Studio Web's Overwrite answers 400/1001 (UiPath/cli#3938)

rockymadden:
## What

`run_debug` now rotates the `.uipx` `SolutionId` to a fresh uuid4 and retries **once as a new import** when Studio Web refuses an overwrite with `400 / code 1001`.

Ports #2990 and #3026 from `campaign/same-ground-expansion` onto `main`.

## Why

`uip maestro flow debug` imports the project on its first run and writes the server's solution id back into the `.uipx`. Every later debug of that project is therefore an **overwrite**, and Studio Web's `Solution/{id}/Overwrite` answers `400 {"code":"1001","message":"An argument had an invalid value."}` for a solution that exists. The CLI treats a non-404 Overwrite as fatal.

Net effect: **`flow debug` is one-shot per project.**

On the 2026-09-03 batch, 8 of 10 `uipath-maestro-flow` tasks lost their last gating criterion to this, always on the second debug of the project:

| Task | Agent debugs | Checker debugs | 400 lands on |
|---|---|---|---|
| `decision` | 0 | 2 | checker case 2 (`cool`) |
| `feet_inches` | 0 | 2 | checker case 2 (`i2f`) |
| `wiki_pageviews` | 0 | 2 | checker case 2 (invalid article) |
| `cli_dice_roller_simulated` | 1 | 1 | checker case 1 |
| `file_attachment_debug` | 1 | 1 | checker case 1 |
| `slack_http_fallback` | 1 | 1 | checker case 1 |
| `slack_weather_pipeline` | 3 | 1 | agent #3, checker |
| `generic_dynamic_node` | 3 | 1 | agent #2, checker |

`flow validate` passed in all 10. A fresh bundled `SolutionId` imports as new and Completes, so only overwrite is broken.

The fix already exists on `campaign/same-ground-expansion` but is blocked behind #2756 (262 files, open since 2026-08-22), so `main` and the nightly still take the hit.

## How

- `_is_overwrite_stuck` matches **both** envelopes: the legacy `Overwrite failed (400): {"code":"1001",…}` text and the staged one the CLI emits since UiPath/cli#3951 (`Context.Stage == "overwrite-solution"`).
- `_rotate_solution_id` walks up from the project dir to the nearest unambiguous `*.uipx` and rewrites only the id.
- The retry gets its own single allowance inside the existing deadline. `debug_budget()` is untouched.
- Never fires on code 20001, another stage, 404, or exit 0.
- Prints the orphaned id, because `cleanup_solutions.py` reads only the current one.

Rewritten onto `main`'s retry loop rather than cherry-picked: the campaign commits conflict with an unrelated unreadable-outputs retry that is not being ported here.

**Self-disabling.** Once Overwrite works again the signature stops matching. Delete the `_OVERWRITE_STUCK` block in `flow_check.py` and the matching test section together.

## Verification

- `pytest tests/tasks/uipath-maestro-flow/` — **776 passed**
- `python3 scripts/check-task-driver.py tests/tasks` — OK
- +13 tests: rotation on both envelopes, rotate-once-only, 20001 and other stages untouched, no `.uipx` → plain failure, classifier table
- All **8 real failure envelopes** from the 2026-09-03 batch classify as ROTATE
- Rotation verified against a real `.uipx` from that run (correct id replaced, rest of file byte-identical)

No prompt, criterion, weight, or `run_limits` change. No task YAML touched.

## Not in scope

Two failures from the same batch have unrelated causes and are untouched here:

- `skill-flow-coded-agent` — real authoring bug. The flow passes `review_sentence`; the LangGraph `GraphState` requires `sentence`. `flow validate` does not catch the mismatch.
- `skill-flow-outlook-trigger-inbox` — the tenant's Outlook connection grant is revoked (`AADSTS50173`). Needs reauthorization, not a code change.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/flow_check.py
@@ -130,6 +130,101 @@ def _output_blob(result: subprocess.CompletedProcess) -> str:
     return f"{result.stdout}\n{result.stderr}".lower()
 
 
+# INTERIM — Studio Web `Solution/{id}/Overwrite` regression (UiPath/cli#3938).
+#
+# Since 2026-09-02 (last known good: 2026-09-01 07:30Z) Studio Web answers
+# `400 {"code":"1001","message":"An argument had an invalid value."}` when a
+# solution that EXISTS is overwritten. `uip maestro flow debug` imports the
+# project on its first run and writes the server's solution id back into the
+# `.uipx`, so every LATER debug of the same project is an overwrite — and the
+# CLI treats any non-404 Overwrite as fatal. Result: the second debug of a
+# project dies at solution upload, so every multi-case checker fails its later
+# cases, and a task whose agent debugged during its turn fails its first.
+# A fresh bundled SolutionId imports as new and Completes, so ONLY overwrite is
+# broken. The only earlier archived Overwrite 400 (2026-08-26) was code 20001, a
+# content error that must still fail.
+#
+# So: on exactly that signature, rotate the bundled SolutionId to a fresh uuid4
+# and retry ONCE — the retry is a new import. Self-limiting: once Overwrite
+# recovers the signature never fires and nothing here runs. Cost while it does:
+# one extra tenant solution per affected case, which `cleanup_solutions.py`
+# cannot see (it reads the `.uipx`'s CURRENT id) — the ids are printed so they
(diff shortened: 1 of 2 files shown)


rockymadden:
No, bad AI!

PR #3040, opened 2026-09-03
docs(uipath-maestro-flow): update child simulation docs for auto-resolved output schema

AAgnihotry:
## Summary

- Updates `commands-reference.md` and `eval-sets-guide.md` to reflect that `--output-schema` is now auto-resolved for child simulations (`--parent`) on all agent types: inline canvas, same-solution, and published.
- Removes `--output-schema` from the Llm child simulation examples.
- Adds note about Static mock value validation against the resolved schema.

Companion to UiPath/cli#3998.

## Test plan

- [x] Docs-only change — reviewed for accuracy against the CLI implementation

### DIFF
--- a/skills/uipath-maestro-flow/references/evaluate/commands-reference.md
+++ b/skills/uipath-maestro-flow/references/evaluate/commands-reference.md
@@ -145,7 +145,6 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
 | `--component-description <text>` | No | Human-readable label for the component |
 | `--simulation-instructions <text>` | No | LLM prompt describing what the component should return (for `Llm` strategy) |
 | `--mock-value <json>` | No | Static JSON output (for `Static` strategy) |
-| `--output-schema <json>` | No | JSON Schema describing the expected output shape; passed to the LLM to constrain its response. **Auto-resolved when omitted** — works for top-level simulations (from the `.flow` node outputs) and child simulations with `--parent` (from the `.flow` for inline agents, from `agent.json` for same-solution agents, or via the platform API for published agents when logged in). Pass explicitly only to override. |
 | `--parent <component-id>` | No | Parent agent node component ID. When set, the simulation is added as a child tool simulation nested inside the parent agent node's simulation. If no parent simulation exists yet, one is auto-created (type `agent`, strategy `Llm`). `--component-type` defaults to `Node`. |
 | `--path <path>` | No | (see Common Options) |
 
@@ -153,25 +152,19 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
 
 | Strategy | When to use | Key flags |
 |----------|-------------|-----------|
-| `Llm` | Output should be realistic but non-deterministic | `--simulation-instructions`, `--output-schema` (auto-resolved for all agent types) |
+| `Llm` | Output should be realistic but non-deterministic | `--simulation-instructions` (output schema auto-resolved) |
 | `Static` | Output is fixed and deterministic | `--mock-value` |
 
-**`--output-schema` auto-resolution:** When omitted, the CLI auto-resolves the output schema for both top-level and child (`--parent`) simulations:
+**Output schema auto-resolution:** The CLI always auto-resolves the output schema for both top-level and child (`--parent`) simulations:
 
 - **Top-level simulations:** reads the `.flow` file, finds the node by `<component-id>`, and derives the schema from the node's output definition (connector `outputJsonSchema`, agent `agentOutputVariables`, or `node.outputs`).
 - **Child simulations (inline canvas agents):** finds the child tool node via edges in the `.flow` file and extracts its output schema.
 - **Child simulations (same-solution agents):** reads the inline agent's `agent.json` and matches the tool by name in the `resources[]` array.
 - **Child simulations (published agents):** calls the platform API (`simulatableComponents`) using the current login session to fetch the tool's schema. Requires `uip login`.
 
-Fails with an actionable error if the node/tool is not found or has no outputs. Pass `--output-schema` explicitly to override. The JSON Schema is sent alongside `--simulation-instructions` to the LLM, telling it what shape the output must conform to. Without it the LLM generates free-form text.
+Fails with an actionable error if the node/tool is not found or has no outputs.
 
-**Static mock value validation:** When `--output-schema` is resolved (explicitly or auto) and `--strategy Static` is used with `--parent`, the CLI validates that the `--mock-value` keys match the schema properties, catching mismatches early.
-
-For a connector that returns `{ status, message }` you would pass:
-
-```bash
---output-schema '{"type":"object","properties":{"status":{"type":"string"},"message":{"type":"string"}}}'
-```
+**Static mock value validation:** For `Static` child simulations, the CLI validates that `--mock-value` keys match the auto-resolved schema properties, catching shape mismatches early.
 
 Example — LLM strategy:
 
--- a/skills/uipath-maestro-flow/references/evaluate/commands-reference.md
+++ b/skills/uipath-maestro-flow/references/evaluate/commands-reference.md
@@ -145,7 +145,7 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
 | `--component-description <text>` | No | Human-readable label for the component |
 | `--simulation-instructions <text>` | No | LLM prompt describing what the component should return (for `Llm` strategy) |
 | `--mock-value <json>` | No | Static JSON output (for `Static` strategy) |
-| `--output-schema <json>` | No | JSON Schema describing the expected output shape; passed to the LLM to constrain its response. **Auto-resolved from the `.flow` file when omitted for `Llm` strategy** (top-level simulations only; required when using `--parent`). |
+| `--output-schema <json>` | No | JSON Schema describing the expected output shape; passed to the LLM to constrain its response. **Auto-resolved when omitted** — works for top-level simulations (from the `.flow` node outputs) and child simulations with `--parent` (from the `.flow` for inline agents, from `agent.json` for same-solution agents, or via the platform API for published agents when logged in). Pass explicitly only to override. |
 | `--parent <component-id>` | No | Parent agent node component ID. When set, the simulation is added as a child tool simulation nested inside the parent agent node's simulation. If no parent simulation exists yet, one is auto-created (type `agent`, strategy `Llm`). `--component-type` defaults to `Node`. |
 | `--path <path>` | No | (see Common Options) |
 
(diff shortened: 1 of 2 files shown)


dushyant-uipath (approved):

skills/uipath-maestro-flow/references/evaluate/commands-reference.md
```diff
@@ -145,22 +145,26 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
...
 **Strategy guide:**
 
 | Strategy | When to use | Key flags |
 |----------|-------------|-----------|
-| `Llm` | Output should be realistic but non-deterministic | `--simulation-instructions`, `--output-schema` (auto-resolved) |
+| `Llm` | Output should be realistic but non-deterministic | `--simulation-instructions` (output schema auto-resolved) |
 | `Static` | Output is fixed and deterministic | `--mock-value` |
 
-**`--output-schema` auto-resolution for Llm simulations:** When omitted, the CLI reads the `.flow` file, finds the node by `<component-id>`, and derives the schema from the node's output definition (connector `outputJsonSchema`, agent `agentOutputVariables`, or `node.outputs`). Fails with an actionable error if the node is not found or has no outputs. Pass `--output-schema` explicitly to override. The JSON Schema is sent alongside `--simulation-instructions` to the LLM, telling it what shape the output must conform to. Without it the LLM generates free-form text. For a connector that returns `{ status, message }` you would pass:
+**Output schema auto-resolution:** The CLI always auto-resolves the output schema for both top-level and child (`--parent`) simulations:
```
rockymadden, line 158: **Medium.** "always" is not true for top-level. `flow-eval-service.ts:324` gates top-level resolution on the strategy:

```ts
if (
    simulation.simulationStrategy === "Llm" &&
    simulation.outputSchema === undefined
) { ... }
```

A top-level `Static` simulation gets no resolution at all. The claim holds for children, which are gated only on `outputSchema === undefined`.

Suggest: "auto-resolves the output schema for top-level `Llm` simulations and for all child (`--parent`) simulations."

skills/uipath-maestro-flow/references/evaluate/commands-reference.md
```diff
@@ -145,22 +145,26 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
...
 
-```bash
---output-schema '{"type":"object","properties":{"status":{"type":"string"},"message":{"type":"string"}}}'
-```
+- **Top-level simulations:** reads the `.flow` file, finds the node by `<component-id>`, and derives the schema from the node's output definition (connector `outputJsonSchema`, agent `agentOutputVariables`, or `node.outputs`).
+- **Child simulations (inline canvas agents):** finds the child tool node via edges in the `.flow` file and extracts its output schema.
+- **Child simulations (same-solution agents):** reads the inline agent's `agent.json` and matches the tool by name in the `resources[]` array.
+- **Child simulations (published agents):** calls the platform API (`simulatableComponents`) using the current login session to fetch the tool's schema. Requires `uip login`.
+
+Fails with an actionable error if the node/tool is not found or has no outputs.
```
rockymadden, line 165: **Medium.** True for top-level and for `Llm` children, not for `Static` children. `flow-eval-service.ts:396` only raises under the Llm branch:

```ts
} else if (childSimulation.simulationStrategy === "Llm") {
    ...throw
}
```

For a `Static` child whose schema does not resolve there is no error, no validation, and exit 0. The doc promises an error that does not come, which is the worse failure mode of the two because the user believes their mock was checked.

skills/uipath-maestro-flow/references/evaluate/commands-reference.md
```diff
@@ -145,22 +145,26 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
...
+
+**Static mock value validation:** For `Static` child simulations, the CLI validates that `--mock-value` keys match the auto-resolved schema properties, catching shape mismatches early.
```
rockymadden, line 167: **Medium.** Conditional, not unconditional. The validation call is guarded on `outputSchema !== undefined`, so it runs only when resolution succeeded. On a published agent with no login, or any API failure, `resolvePublishedAgentSchema` returns `{ parentFound: true }` with no schema and the check is silently skipped.

Worth one clause: "...when the schema resolves; skipped silently otherwise."

skills/uipath-maestro-flow/references/evaluate/commands-reference.md
```diff
@@ -145,22 +145,26 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
...
+**Output schema auto-resolution:** The CLI always auto-resolves the output schema for both top-level and child (`--parent`) simulations:
 
-```bash
---output-schema '{"type":"object","properties":{"status":{"type":"string"},"message":{"type":"string"}}}'
-```
+- **Top-level simulations:** reads the `.flow` file, finds the node by `<component-id>`, and derives the schema from the node's output definition (connector `outputJsonSchema`, agent `agentOutputVariables`, or `node.outputs`).
+- **Child simulations (inline canvas agents):** finds the child tool node via edges in the `.flow` file and extracts its output schema.
+- **Child simulations (same-solution agents):** reads the inline agent's `agent.json` and matches the tool by name in the `resources[]` array.
+- **Child simulations (published agents):** calls the platform API (`simulatableComponents`) using the current login session to fetch the tool's schema. Requires `uip login`.
```
rockymadden, line 163: **Minor.** "Requires `uip login`" undersells it. This path needs live tenant reachability, and every failure mode (401, 500, DNS, timeout) is swallowed into "no schema" rather than surfaced. Combined with the Static behaviour noted above, a user offline on a `Static` child gets a silent skip with no clue why.

Also worth noting there is no timeout on the call, so an unreachable tenant hangs the command.

skills/uipath-maestro-flow/references/evaluate/commands-reference.md
```diff
@@ -145,22 +145,26 @@ Add or replace a simulation on a data point. If a simulation for `<component-id>
 | `--component-description <text>` | No | Human-readable label for the component |
 | `--simulation-instructions <text>` | No | LLM prompt describing what the component should return (for `Llm` strategy) |
 | `--mock-value <json>` | No | Static JSON output (for `Static` strategy) |
-| `--output-schema <json>` | No | JSON Schema describing the expected output shape; passed to the LLM to constrain its response. **Auto-resolved from the `.flow` file when omitted for `Llm` strategy** (top-level simulations only; required when using `--parent`). |
 | `--parent <component-id>` | No | Parent agent node component ID. When set, the simulation is added as a child tool simulation nested inside the parent agent node's simulation. If no parent simulation exists yet, one is auto-created (type `agent`, strategy `Llm`). `--component-type` defaults to `Node`. |
```
rockymadden, line 148: **Minor.** The `--output-schema` row was deleted from this table but nothing states the flag was **removed**. It is not optional now, it errors:

```
error: unknown option '--output-schema'
```

These docs are an agent's ground truth. An agent carrying the old form from training, from an existing script, or from the eval tasks in this repo gets that error and finds nothing here explaining it. One line under the table would close the loop: "`--output-schema` was removed; the schema is always derived."

skills/uipath-maestro-flow/references/evaluate/eval-sets-guide.md
```diff
@@ -157,10 +156,10 @@ uip maestro flow eval simulation remove <component-id> \
 
 | Strategy | Use | Required or relevant flags |
 |---|---|---|
-| `Llm` | Plausible, non-deterministic output | `--simulation-instructions`, `--output-schema` (auto-resolved) |
+| `Llm` | Plausible, non-deterministic output | `--simulation-instructions` (output schema auto-resolved) |
 | `Static` | Identical output every run | `--mock-value <json>` |
 
-For `Llm`, `--output-schema` is auto-resolved from the `.flow` file: connector `outputJsonSchema`, agent `agentOutputVariables`, or `node.outputs`. Pass it explicitly only to override the derived schema or when no outputs are declared.
+The output schema is always auto-resolved — for both top-level and child (`--parent`) simulations. Top-level reads the `.flow` node outputs; child simulations resolve from the `.flow` edges (inline agents), `agent.json` resources (same-solution agents), or the platform API (published agents, requires `uip login`).
```
rockymadden, line 162: **Medium.** Same "always" overclaim as `commands-reference.md:158`. Top-level resolution is gated on `simulationStrategy === "Llm"`; a top-level `Static` simulation is never resolved.

Side note: this three-path description now exists in two files with slightly different wording. The duplication predates this PR, but it is now two copies to keep in sync and they have already drifted.

skills/uipath-maestro-flow/references/evaluate/eval-sets-guide.md
```diff
@@ -209,6 +209,13 @@ No separate parent simulation step is needed — passing `--parent` auto-creates
 
 When `--parent` is used, `--component-type` defaults to `Node` (the convention for child tool simulations). You can override it if needed.
 
+The output schema is auto-resolved for child simulations on all agent types:
+- **Inline canvas agents** (`uipath.agent.*`): resolved from the child tool node's outputs in the `.flow` file.
+- **Same-solution agents** (with `inputs.source`): resolved from the inline agent's `agent.json` resources.
+- **Published agents** (`uipath.core.agent.*`): resolved via the platform API (`simulatableComponents`). Requires `uip login`.
+
+For `Static` strategy, the CLI also validates that `--mock-value` keys match the resolved schema properties, catching shape mismatches before the eval run.
```
rockymadden, line 217: **Medium.** Same conditional-validation issue as `commands-reference.md:167`. The check runs only when the schema resolved, so an unresolvable published tool silently skips it and the command still exits 0.

"catching shape mismatches before the eval run" is the part that misleads: it reads as a guarantee.

rockymadden (changes requested):
The narrative is right about the three resolution paths. I checked each against the implementation in UiPath/cli#3998 and they match. Two problems though.

## The eval tasks were not updated

`--output-schema` is now a hard `unknown option` error, but two task files in this repo still instruct the agent to pass it. Neither is in this diff, so I cannot comment inline.

**`tests/tasks/uipath-maestro-flow/evaluate/simulation/simulation_crud.yaml` is now unpassable as written:**

- **L7** (description): ``  `Llm` (explicit `--output-schema`, so the auto-resolution-from-.flow path is not exercised)  ``
- **L45** (step 6): tells the agent to pass `--output-schema '...'` *"explicitly (do NOT rely on auto-resolution from the .flow file)"*
- **L120** (grader):

```yaml
command_pattern: '(uip|\$UIP)\s+maestro\s+flow\s+eval\s+simulation\s+add\s+.*--strategy\s+Llm\b.*--output-schema'
weight: 3.0
pass_threshold: 1.0
```

The assertion can only be satisfied by a command that now fails. An agent that follows the prompt fails the CLI; an agent that self-corrects and drops the flag fails the grader. Either way the task is red, and a weight-3.0 assertion now certifies removed behavior.

**`tests/tasks/uipath-maestro-flow/evaluate/child_simulation/child_simulation_crud.yaml` is partially broken:**

- **L47** (step 6): instructs passing `--output-schema` explicitly, so the command errors
- **L121**: assertion description says "...and `--output-schema`" while the regex does not actually require it. It can still pass on a retry, but the description is stale and the prompt is wrong.

Suggest inverting both: drop the flag from step 6, change the graders to assert the schema was auto-resolved into the eval-set JSON, and fix the L7 description. That turns two now-broken tasks into the coverage this behavior currently has none of.

## Three claims the implementation does not back

"always auto-resolves", "fails with an actionable error", and the unconditional Static validation each describe a guard that is not there. Details inline.

One process note: this landed as a docs-only change for a breaking CLI change. Worth treating a removed flag as a repo-wide sweep (`git grep -- --output-schema`) rather than an edit to the two files that describe it.

PR #3041, opened 2026-09-03
feat(uipath-maestro-flow): teach the native Data Fabric entity CRUD nodes

uipath-utkarsh:
## What

Adds coding-agent support to `uipath-maestro-flow` for the four **native Data Fabric entity nodes**:

| Node | Canvas label | Tenant flag |
| --- | --- | --- |
| `core.datafabric.read` | Read entity records | `canvas.nodes.read-entity` |
| `core.datafabric.create` | Create entity record | `canvas.nodes.create-entity` |
| `core.datafabric.update` | Update entity record | `canvas.nodes.update-entity` |
| `core.datafabric.delete` | Delete entity record | `canvas.nodes.delete-entity` |

Until now the skill covered Data Fabric only as `uipath-uipath-dataservice` Integration Service activities. The native nodes need no IS connection, are user-owned (plain-JSON `inputs.entityConfig`, not a `=jsonString:` envelope), and give downstream expressions the picked entity's record shape.

All four tenant flags **default to off**, so the guide makes node choice availability-first: check `registry get`, and build with the connector when the native node is not there.

## New — `references/author/plugins/data-fabric/`

Follows the established plugin pair, modelled on the flag-gated `batch-transform` / `summarize` plugins.

**`planning.md`** — node types and flags, the native-entity requirement for writes, selection heuristics, the availability check, ports, per-node output semantics, `entityConfig` keys, `byId` vs `fromRead`, planning annotation.

**`impl.md`** — registry validation, ownership, entity/scope/column resolution via `uip df`, per-node JSON, write bodies, schema snapshots, the filter grammar, record selection, folder-scoped bindings, output wiring, validator coverage, debug table, what-not-to-do.

### The failure mode this plugin is built around

Almost every way these nodes go wrong is **silent**: Data Fabric rejects the write, the engine logs it and moves on, and the run reports Completed having changed nothing. `flow validate` catches the structural cases and nothing else. So the guide is organised around what validates clean and still does nothing:

- **Writes require a native entity.** No validation rule enforces it — the canvas filters its picker, a hand-authored flow does not.
- **System columns** (`Id`, `CreateTime`, `CreatedBy`, `UpdateTime`, `UpdatedBy`) and attachment columns are never writable; the serializer has no guard.
- **Choice-set columns take the numeric `numberId`, not the label**, and the multi-valued form is a JSON array *as a string* (`"[1,2]"`).
- **`fromRead` fails silently four ways** — missing node id, multi-record target, non-compiling filters (all emit a no-op task), and a runtime over-match, which the write path *swallows* even though the read path faults on it.
- **An absent `recordSource` is not inert** — the serializer infers the mode from whichever selector is populated. The validator requires it on Delete but **not** on Update.
- **Create vs Update blank-value asymmetry** — blank applies the column default on insert, but asserts `null` on update, which a non-nullable column rejects.

Plus the mechanical contracts an agent can't infer: Read emits a `<uipath:input>` for the engine's preprocessor while the writes emit a `<uipath:output>` for the postprocessor; a multi-record read lands under `output.results`; Update's output is a re-read that can't feed a Loop or Transform collection; Delete produces nothing; and these four are the one action-node family with **no error port** and no instance `outputs` block.

## Routing and shared-doc updates

- `SKILL.md` — capability line + description enumeration
- `author/CAPABILITY.md` — node-ownership table (user-owned), common tasks, plugin list
- `author/planning-arch.md` — Plugin Index, Standard Port Reference, Data Fabric carve-out on the external-service ladder
- `author/planning-impl.md` — Step-1a node-type router
- `plugins/connector/impl.md` — the `uipath-uipath-dataservice` activity rows point back, so the availability check is visible *before* the connector is chosen
- `shared/file-format.md` — new **Data Fabric entity bindings** subsection. The `Entity` kind resolves by `(resourceKey, propertyAttribute)`, has no `<bindings.{name}>` placeholder, puts a display label in `name`, uses `folderKey` not `folderPath`, and has no `resourceSubType`. The existing H2 anchor is untouched (five sibling plugins link to it).
- `shared/action-nodes.md` — Exceptions note and Standard-ports table record that `core.datafabric.*` has no error port and no instance `outputs` block
- `CODEOWNERS` — plugin dir under `@UiPath/DatafabricCodingAgent` + `@UiPath/maestro-cli-team`

## Verification

Every behavioural claim was checked against `flow-workbench@develop` — node manifests under `flow-core/src/manifest/data/ootb-nodes/core-datafabric-*`, `manifest-client.ts`, `services/src/serialization/entity-vdo.ts`, `uipath-extension.ts`, `expression-model`, `flow-schema` validation rules, and the canvas entity panels — not inferred from node names.

Two correction passes are folded in. `1ab64a31` fixed claims an internal review found (a faulting `fromRead` over-match, an inert `recordSource`, an implicit error port, an invented `CreatedAt` column, an engine-derived row cap). `69386f79` fixed a **real regression this branch caused**: the first framing read as an unconditional "prefer the native node", and `skill-flow-datafabric-smoke-update` — which builds against the connector — switched to native nodes and failed its verifier. Making the choice availability-first restored it; that task and `smoke-error` are now SUCCESS and the tier passes at 96%.

`.maintenance/check-all.sh`: links 898/0, link-text 571/0, anchors 286 checked / 1 bad (pre-existing, reproduces on `main`), depth 84/84, orphans 0, plugin-pairs 27/0, template 0 missing, versions 0 violations. `npm run skills:validate` OK (27 skills, both flavors). `check-skill-verbs.py` clean on all new and edited files.

## Notes for reviewers

1. **Eval task deferred — tenant flags are the blocker, not the CLI.** Correcting my earlier note: on `uip` 1.202.0 all four node types are returned by `registry search` with `AvailableOnTenant: false`. (My local 1.201.0-dev.8247 returns only `core.datafabric.read`, which is what the stale claim was based on.) Either way `registry get` answers *Node not found* until a tenant enables the flags, so a task cannot instantiate these nodes yet. Worth tracking as a follow-up rather than a blocker here — `queue`, `batch-transform`, `summarize` and `delay` also ship without a task dir.
2. **`check-uip-commands.sh` crashes on Python 3.9** (`TypeError: unsupported operand type(s) for |`) — pre-existing, reproduces on `main`. Clean under 3.14. Left alone to keep this PR scoped; the CLI Verb Gate covers the same ground in CI and passes.

https://claude.ai/code/session_0151VobGkRGo2Ym4j6s67kem

### DIFF
--- a/skills/uipath-maestro-flow/references/author/CAPABILITY.md
+++ b/skills/uipath-maestro-flow/references/author/CAPABILITY.md
@@ -160,7 +160,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 - [planning-arch.md](planning-arch.md) — capability discovery, plugin index, topology design
 - [planning-impl.md](planning-impl.md) — registry lookups, connection binding, wiring rules
 - [plugins/](plugins/) — per-node-type planning + impl docs:
-  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; cf. [data-fabric](plugins/data-fabric/))
+  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; see [data-fabric](plugins/data-fabric/))
   - [connector-trigger](plugins/connector-trigger/)
   - [script](plugins/script/) — Jint ES2020 JavaScript
   - [http](plugins/http/) — `core.action.http.v2` (Managed HTTP Request)
--- a/skills/uipath-maestro-flow/references/author/CAPABILITY.md
+++ b/skills/uipath-maestro-flow/references/author/CAPABILITY.md
@@ -37,6 +37,7 @@ Every node in a `.flow` file has exactly one author. The validator enforces this
 | Resource nodes | `uipath.core.rpa-workflow.*`, `uipath.core.agent.*`, `uipath.core.flow.*`, `uipath.core.agentic-process.*`, `uipath.core.api-workflow.*`, `uipath.core.human-task.*` |
 | Document extraction | `uipath.ixp.*` — the extraction step must always land a node ([ixp/impl.md](plugins/ixp/impl.md#landing-the-node-when-you-cannot-fully-configure-it)) |
 | Queue | `core.action.queue.create`, `core.action.queue.create-and-wait` |
+| Data Fabric | `core.datafabric.read`, `core.datafabric.create`, `core.datafabric.update`, `core.datafabric.delete` — `inputs.entityConfig` is plain JSON, not an envelope ([data-fabric/impl.md](plugins/data-fabric/impl.md)) |
 
 **CLI-owned nodes (`uip maestro flow node add` + `uip maestro flow node configure`):**
 
@@ -116,6 +117,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 | **Create a subflow** | [plugins/subflow/impl.md](plugins/subflow/impl.md) + [Edit/Write: Create a subflow](editing-operations-json.md#create-a-subflow) |
 | **Add a delay or scheduled trigger** | [plugins/delay/](plugins/delay/) or [plugins/scheduled-trigger/](plugins/scheduled-trigger/) |
 | **Use queue nodes** | [plugins/queue/impl.md](plugins/queue/impl.md) |
+| **Read or write Data Fabric entity records** | [plugins/data-fabric/impl.md](plugins/data-fabric/impl.md) — `core.datafabric.read` / `create` / `update` / `delete`, each gated by its own `canvas.nodes.*-entity` tenant flag |
 
 ## Anti-patterns
 
@@ -158,7 +160,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 - [planning-arch.md](planning-arch.md) — capability discovery, plugin index, topology design
 - [planning-impl.md](planning-impl.md) — registry lookups, connection binding, wiring rules
 - [plugins/](plugins/) — per-node-type planning + impl docs:
-  - [connector](plugins/connector/) — IS connector nodes (incl. Data Fabric activities)
+  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; prefer the native [data-fabric](plugins/data-fabric/) nodes when their flag is on)
   - [connector-trigger](plugins/connector-trigger/)
   - [script](plugins/script/) — Jint ES2020 JavaScript
   - [http](plugins/http/) — `core.action.http.v2` (Managed HTTP Request)
@@ -184,6 +186,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
   - [inline-voice-agent](plugins/inline-voice-agent/) — voice agent on a live phone call (inbound/outbound) + the trigger, create-call, and end-call nodes
   - [ixp](plugins/ixp/) — published IxP document-extraction models (PDFs, scanned forms, receipts, invoices, contracts)
   - [queue](plugins/queue/) — Orchestrator queue item creation
+  - [data-fabric](plugins/data-fabric/) — native Data Fabric entity record CRUD (`core.datafabric.*`), no Integration Service connection
 
 ### Cross-capability (shared)
 
--- a/skills/uipath-maestro-flow/references/author/CAPABILITY.md
+++ b/skills/uipath-maestro-flow/references/author/CAPABILITY.md
@@ -160,7 +160,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 - [planning-arch.md](planning-arch.md) — capability discovery, plugin index, topology design
 - [planning-impl.md](planning-impl.md) — registry lookups, connection binding, wiring rules
 - [plugins/](plugins/) — per-node-type planning + impl docs:
-  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; prefer the native [data-fabric](plugins/data-fabric/) nodes when their flag is on)
+  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; the native [data-fabric](plugins/data-fabric/) nodes are the lighter path where the tenant has them enabled — their flags default to off, so confirm with `registry get` before choosing)
   - [connector-trigger](plugins/connector-trigger/)
   - [script](plugins/script/) — Jint ES2020 JavaScript
   - [http](plugins/http/) — `core.action.http.v2` (Managed HTTP Request)
--- a/skills/uipath-maestro-flow/references/author/CAPABILITY.md
+++ b/skills/uipath-maestro-flow/references/author/CAPABILITY.md
@@ -160,7 +160,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 - [planning-arch.md](planning-arch.md) — capability discovery, plugin index, topology design
 - [planning-impl.md](planning-impl.md) — registry lookups, connection binding, wiring rules
 - [plugins/](plugins/) — per-node-type planning + impl docs:
-  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; the native [data-fabric](plugins/data-fabric/) nodes are the lighter path where the tenant has them enabled — their flags default to off, so confirm with `registry get` before choosing)
+  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; cf. [data-fabric](plugins/data-fabric/))
   - [connector-trigger](plugins/connector-trigger/)
   - [script](plugins/script/) — Jint ES2020 JavaScript
   - [http](plugins/http/) — `core.action.http.v2` (Managed HTTP Request)
@@ -186,7 +186,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
   - [inline-voice-agent](plugins/inline-voice-agent/) — voice agent on a live phone call (inbound/outbound) + the trigger, create-call, and end-call nodes
   - [ixp](plugins/ixp/) — published IxP document-extraction models (PDFs, scanned forms, receipts, invoices, contracts)
   - [queue](plugins/queue/) — Orchestrator queue item creation
-  - [data-fabric](plugins/data-fabric/) — native Data Fabric entity record CRUD (`core.datafabric.*`), no Integration Service connection
+  - [data-fabric](plugins/data-fabric/) — native Data Fabric entity record CRUD (`core.datafabric.*`); tenant flags default to off
 
 ### Cross-capability (shared)
 
--- a/skills/uipath-maestro-flow/references/author/planning-arch.md
+++ b/skills/uipath-maestro-flow/references/author/planning-arch.md
@@ -169,7 +169,7 @@ Prefer, in order:
 2. `core.action.http.v2` connector mode when the connector lacks the activity, or manual mode for APIs without connectors ([http](plugins/http/planning.md)).
 3. An RPA workflow only when there is no API, such as a desktop app or terminal ([rpa](plugins/rpa/planning.md)).
 
-**Data Fabric is not on this ladder.** Entity records have two paths — the `uipath-uipath-dataservice` connector activities and the native `core.datafabric.*` nodes — and availability decides, not preference. See [data-fabric/planning.md — Native node vs Data Service connector](plugins/data-fabric/planning.md#native-node-vs-data-service-connector--availability-decides).
+**Data Fabric is not on this ladder.** Entity records have two paths — the `uipath-uipath-dataservice` connector activities and the native `core.datafabric.*` nodes — and availability decides, not preference. The native flags default to off, so **when `registry get core.datafabric.<op>` answers "Node not found", or search reports `AvailableOnTenant: false`, build with the connector activities**: do not retry, do not `uip tools update`, and never hand-write a `definitions[]` entry for a node the registry will not return. Rationale and the federated-entity case in [data-fabric/planning.md — Native node vs Data Service connector](plugins/data-fabric/planning.md#native-node-vs-data-service-connector--availability-decides).
 
 ## Standard Port Reference
 
--- a/skills/uipath-maestro-flow/references/author/planning-arch.md
+++ b/skills/uipath-maestro-flow/references/author/planning-arch.md
@@ -169,9 +169,7 @@ Prefer, in order:
 2. `core.action.http.v2` connector mode when the connector lacks the activity, or manual mode for APIs without connectors ([http](plugins/http/planning.md)).
 3. An RPA workflow only when there is no API, such as a desktop app or terminal ([rpa](plugins/rpa/planning.md)).
 
-**Data Fabric has two paths, and availability decides which.** Entity records can be reached either through the `uipath-uipath-dataservice` connector activities ([connector](plugins/connector/planning.md)) or through the native `core.datafabric.*` nodes ([data-fabric](plugins/data-fabric/planning.md)).
-
-The native nodes are nicer when they are there — no Integration Service connection, no `node configure` — but **all four tenant flags default to off**, so on most tenants they do not exist. Decide by checking, not by preference: run `uip maestro flow registry get core.datafabric.read` (and the sibling type for the operation you need). If it returns the node, use the native node. If it answers "Node not found" or search reports `AvailableOnTenant: false`, **use the connector activities** and do not spend further turns on the native path. Also use the connector when the entity is federated — the native writes require a native entity — or when you need an operation the four nodes do not cover.
+**Data Fabric is not on this ladder.** Entity records have two paths — the `uipath-uipath-dataservice` connector activities and the native `core.datafabric.*` nodes — and availability decides, not preference. See [data-fabric/planning.md — Native node vs Data Service connector](plugins/data-fabric/planning.md#native-node-vs-data-service-connector--availability-decides).
 
 ## Standard Port Reference
 
--- a/skills/uipath-maestro-flow/references/author/planning-arch.md
+++ b/skills/uipath-maestro-flow/references/author/planning-arch.md
@@ -110,7 +110,7 @@ Every flow has exactly one trigger, first in topology. IS connector triggers rep
 | `core.action.queue.create` | [queue](plugins/queue/planning.md) | Fire-and-forget robot work |
 | `core.action.queue.create-and-wait` | [queue](plugins/queue/planning.md) | Robot work with result wait |
 | `core.datafabric.read` | [data-fabric](plugins/data-fabric/planning.md) | Read one record or a filtered list from a Data Fabric entity; gated by `canvas.nodes.read-entity` |
-| `core.datafabric.create` | [data-fabric](plugins/data-fabric/planning.md) | Insert a record and return the stored row; gated by `canvas.nodes.create-entity` (needs engine >= 1.923.0) |
+| `core.datafabric.create` | [data-fabric](plugins/data-fabric/planning.md) | Insert a record and return the stored row; gated by `canvas.nodes.create-entity` |
 | `core.datafabric.update` | [data-fabric](plugins/data-fabric/planning.md) | Patch named columns on one record; gated by `canvas.nodes.update-entity` |
 | `core.datafabric.delete` | [data-fabric](plugins/data-fabric/planning.md) | Delete one record; gated by `canvas.nodes.delete-entity` |
 | `uipath.human-in-the-loop.quick-form` | [hitl](plugins/hitl/planning.md) | Inline human review, approval, or data entry |
@@ -168,6 +168,8 @@ Prefer, in order:
 1. A curated Integration Service connector ([connector](plugins/connector/planning.md)).
 2. `core.action.http.v2` connector mode when the connector lacks the activity, or manual mode for APIs without connectors ([http](plugins/http/planning.md)).
 3. An RPA workflow only when there is no API, such as a desktop app or terminal ([rpa](plugins/rpa/planning.md)).
+
+**Data Fabric is not on this ladder.** Reading or writing records in a UiPath Data Fabric entity is not an external-service call: use the native `core.datafabric.*` nodes ([data-fabric](plugins/data-fabric/planning.md)), which need no Integration Service connection. The `uipath-uipath-dataservice` connector activities are the fallback — take them only when the native node's tenant flag is off, when the entity is federated (the native writes require a native entity), or when you need an operation the four nodes do not cover.
 
 ## Standard Port Reference
 
--- a/skills/uipath-maestro-flow/references/author/planning-arch.md
+++ b/skills/uipath-maestro-flow/references/author/planning-arch.md
@@ -169,7 +169,9 @@ Prefer, in order:
 2. `core.action.http.v2` connector mode when the connector lacks the activity, or manual mode for APIs without connectors ([http](plugins/http/planning.md)).
 3. An RPA workflow only when there is no API, such as a desktop app or terminal ([rpa](plugins/rpa/planning.md)).
 
-**Data Fabric is not on this ladder.** Reading or writing records in a UiPath Data Fabric entity is not an external-service call: use the native `core.datafabric.*` nodes ([data-fabric](plugins/data-fabric/planning.md)), which need no Integration Service connection. The `uipath-uipath-dataservice` connector activities are the fallback — take them only when the native node's tenant flag is off, when the entity is federated (the native writes require a native entity), or when you need an operation the four nodes do not cover.
+**Data Fabric has two paths, and availability decides which.** Entity records can be reached either through the `uipath-uipath-dataservice` connector activities ([connector](plugins/connector/planning.md)) or through the native `core.datafabric.*` nodes ([data-fabric](plugins/data-fabric/planning.md)).
+
+The native nodes are nicer when they are there — no Integration Service connection, no `node configure` — but **all four tenant flags default to off**, so on most tenants they do not exist. Decide by checking, not by preference: run `uip maestro flow registry get core.datafabric.read` (and the sibling type for the operation you need). If it returns the node, use the native node. If it answers "Node not found" or search reports `AvailableOnTenant: false`, **use the connector activities** and do not spend further turns on the native path. Also use the connector when the entity is federated — the native writes require a native entity — or when you need an operation the four nodes do not cover.
 
 ## Standard Port Reference
 
--- a/skills/uipath-maestro-flow/references/author/planning-arch.md
+++ b/skills/uipath-maestro-flow/references/author/planning-arch.md
@@ -109,6 +109,10 @@ Every flow has exactly one trigger, first in topology. IS connector triggers rep
 | `core.logic.delay` | [delay](plugins/delay/planning.md) | Duration or date wait |
 | `core.action.queue.create` | [queue](plugins/queue/planning.md) | Fire-and-forget robot work |
 | `core.action.queue.create-and-wait` | [queue](plugins/queue/planning.md) | Robot work with result wait |
+| `core.datafabric.read` | [data-fabric](plugins/data-fabric/planning.md) | Read one record or a filtered list from a Data Fabric entity; gated by `canvas.nodes.read-entity` |
+| `core.datafabric.create` | [data-fabric](plugins/data-fabric/planning.md) | Insert a record and return the stored row; gated by `canvas.nodes.create-entity` (needs engine >= 1.923.0) |
+| `core.datafabric.update` | [data-fabric](plugins/data-fabric/planning.md) | Patch named columns on one record; gated by `canvas.nodes.update-entity` |
+| `core.datafabric.delete` | [data-fabric](plugins/data-fabric/planning.md) | Delete one record; gated by `canvas.nodes.delete-entity` |
 | `uipath.human-in-the-loop.quick-form` | [hitl](plugins/hitl/planning.md) | Inline human review, approval, or data entry |
 | `uipath.conversational.voice.create-outgoing-call` | [inline-voice-agent](plugins/inline-voice-agent/planning.md) | Dial an outbound phone call and emit its `callContext` (outbound voice topology) |
 | `uipath.conversational.voice.end-call` | [inline-voice-agent](plugins/inline-voice-agent/planning.md) | End the active call in a voice flow |
@@ -204,6 +208,10 @@ Every edge requires `sourcePort` and `targetPort`.
 | `uipath.connector.*` | `input` | `output`, `error` |
 | `core.action.queue.create` | `input` | `success` |
 | `core.action.queue.create-and-wait` | `input` | `success` |
+| `core.datafabric.read` | `input` | `output` |
+| `core.datafabric.create` | `input` | `output` |
+| `core.datafabric.update` | `input` | `output` |
+| `core.datafabric.delete` | `input` | `output` (sequencing only — the node produces no data) |
 | `uipath.human-in-the-loop.quick-form` | `input` | `completed` |
 | `uipath.core.human-task.{key}` | `input` | `output` |
 
--- a/skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
@@ -38,7 +38,9 @@ If `registry get` reports **"Node not found"**, the node is not available to you
 | `core.datafabric.update` | `canvas.nodes.update-entity` |
 | `core.datafabric.delete` | `canvas.nodes.delete-entity` |
 
-`registry search` is not a substitute for `registry get` here. A flag-gated node can still appear in search with `AvailableOnTenant: false` while `registry get` refuses it — and without `registry get` you cannot source the `definitions[]` entry, which must never be hand-written ([Author capability, rule 6](../../CAPABILITY.md#critical-rules)). Treat `AvailableOnTenant: false` as unavailable and stop.
+`registry search` is not a substitute for `registry get` here. A flag-gated node can still appear in search with `AvailableOnTenant: false` while `registry get` refuses it — and without `registry get` you cannot source the `definitions[]` entry, which must never be hand-written ([Author capability, rule 6](../../CAPABILITY.md#critical-rules)).
+
+**When the node is unavailable, switch to the connector and stop.** `AvailableOnTenant: false` is a decision, not an obstacle: build the flow with the `uipath-uipath-dataservice` activities ([connector/impl.md](../connector/impl.md)) and say in the final report that the native nodes were unavailable. Do not retry `registry get`, do not run `uip tools update` hoping for a newer manifest, and above all **do not hand-author a `definitions[]` entry from this doc's field list to stand in for the missing one** — a hand-written definition carries the wrong port schema, passes `flow validate`, and fails at runtime.
 
 ## Add or edit the node
 
--- a/skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
@@ -58,6 +58,8 @@ An invented id or column matches no record: every lookup returns empty and the r
 
 Node instances carry `inputs` only. Do **not** hand-write an `outputs` block, a `model` block, or a `ui` block — `flow format` owns layout, the definition owns the BPMN model, and the manifest's `outputDefinition` plus the `variables.nodes[]` entry that `flow format` regenerates own the output contract ([Author capability, rules 13–15](../../CAPABILITY.md#critical-rules)).
 
+> **Editing a node the canvas created: preserve the `_`-prefixed keys you do not recognize.** The entity panels snapshot the picked entity's schema onto `entityConfig` — `_entityFields`, `_relatedFields`, `_outputSchema` — and those snapshots are what resolve related-column paths and drive output IntelliSense. They are not required to author a node from scratch, and none of the examples below carry them, but stripping them from an existing node silently breaks any filter that addresses a related column (`customer.Name`), because a path that cannot resolve against the snapshot makes the serializer refuse the whole query. Change the keys you mean to change and leave the rest.
+
 ### Read — single record
 
 ```json
@@ -237,10 +239,33 @@ A **folder-scoped** entity carries `_folderKey` (the folder GUID). Its presence 
 
 ```json
 "bindings": [
-  { "id": "<bindingId1>", "resource": "Entity", "resourceKey": "<resourceKey>", "propertyAttribute": "name",      "name": "Orders" },
-  { "id": "<bindingId2>", "resource": "Entity", "resourceKey": "<resourceKey>", "propertyAttribute": "folderKey", "name": "<folder GUID>" }
+  {
+    "id": "<bindingId1>",
+    "name": "Orders",
+    "type": "string",
+    "resource": "Entity",
+    "resourceKey": "<resourceKey>",
+    "default": "Orders",
+    "propertyAttribute": "name"
+  },
+  {
+    "id": "<bindingId2>",
+    "name": "OrdersFolder",
+    "type": "string",
+    "resource": "Entity",
+    "resourceKey": "<resourceKey>",
+    "default": "<folder GUID>",
+    "propertyAttribute": "folderKey"
+  }
 ]
 ```
+
+Field-by-field, because three of these are easy to get wrong:
+
+- **The value lives in `default`, never in `name`.** `name` is a display label: the entity-name row uses the entity name for both, but the folder row's `name` is the literal convention `<entityName>Folder` while its `default` carries the folder GUID.
+- **`type` is `"string"` and is required** — the binding schema rejects a row without it.
+- **`resource` must be the capitalized `"Entity"`.** The BPMN engine matches it case-insensitively, but packaging requires the capital form; a lowercase row never becomes a binding resource, so the deploy side gets no override at all.
+- **`propertyAttribute` is what the serializer matches on**, not `name`.
 
 The folder token must come from the **`folderKey`** attribute, not `folderPath`: the platform's deploy-time override rewrites `folderPath` with the Orchestrator FQN, which breaks the query even in the source org.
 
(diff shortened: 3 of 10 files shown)


skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
...
+  }
+]
+```
+
+Field-by-field, because three of these are easy to get wrong:
+
+- **The value lives in `default`, never in `name`.** `name` is a display label: the entity-name row uses the entity name for both, but the folder row's `name` is the literal convention `<entityName>Folder` while its `default` carries the folder GUID.
+- **`type` is `"string"` and is required** — the binding schema rejects a row without it.
+- **`resource` must be the capitalized `"Entity"`.** The BPMN engine matches it case-insensitively, but packaging requires the capital form; a lowercase row never becomes a binding resource, so the deploy side gets no override at all.
+- **`propertyAttribute` is what the serializer matches on**, not `name`.
```
rockymadden: **major** — this contradicts `shared/file-format.md`, which is the authoritative reference for top-level `bindings[]`.

- `file-format.md:547`: "Matching is by `(resourceKey, name)`" — the opposite of what this line says.
- Line 331 above makes `name` a display label (`Orders` / `OrdersFolder`). Every sibling plugin (`rpa`, `agent`, `flow`, `agentic-process`, `api-workflow`) and the shared example put the *placeholder name* there (`"name"` / `"folderPath"`).
- `file-format.md:515` scopes top-level `bindings[]` to "**Orchestrator resource nodes** — RPA workflows, agents, flows, agentic processes, API workflows, and HITL apps." `Entity` isn't in that list and the shared doc isn't updated.
- `file-format.md:551` says `resourceSubType` mirrors the definition and enumerates a value per kind. The example omits it with no comment — legitimate if Entity bindings have no definition-side `model.bindings`, but the silence reads as an oversight.

One of the two files is wrong for the Entity case. Please confirm at source, then update the `Bindings` section of `shared/file-format.md` to add the `Entity` kind, the `folderKey` attribute, the `propertyAttribute`-matching carve-out, and whether `resourceSubType` applies. An agent doing folder-scoped entity work reads both files.

Flagging for @UiPath/DatafabricCodingAgent — I can't tell which is correct at source.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
...
+  "typeVersion": "<from registry get>",
+  "display": { "label": "Mark shipped" },
+  "inputs": {
+    "entityConfig": {
+      "entityName": "Orders",
+      "recordSource": "fromRead",
+      "readEntityNodeId": "readOrder",
+      "fieldUpdates": [
+        { "field": "Notes", "value": "Shipped by flow" },
+        { "field": "ShippedAt", "value": "=js:new Date().toISOString()" }
```
rockymadden: **major** — `shared/variables-and-expressions.md:577` lists this under **Not Supported**:

> `Date` constructor may have limited support — prefer ISO 8601 strings

This is the only `new Date()` in the entire skill. And it sits in the one node family where an uncoercible value is rejected and the rejection swallowed — the exact silent failure this plugin exists to prevent.

Suggest an ISO string from a `$vars` reference, or an upstream Script node, matching how the rest of the skill handles timestamps.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
...
+
+And at runtime, if the read matches more than one record, the engine re-runs the query as a single-record fetch and **swallows** the resulting over-match — the write is skipped and the run goes green. (The same over-match faults an ordinary read; only the write path swallows it.) A `fromRead` target must therefore be a genuinely single-record read, verified by you rather than by the runtime.
+
+## Folder-scoped entities and bindings
+
+A **tenant-scoped** entity needs nothing beyond `entityName`; the target serializes as `=datafabric.<EntityName>`.
+
+A **folder-scoped** entity carries `_folderKey` (the entity's `folderId`). Its presence switches the emitted target to the folder-qualified form, and for the flow to survive export to another org the entity's name and folder must serialize as binding tokens rather than source-org literals. That requires:
+
+1. `_resourceKey` (preferred) or `_entityKey` on the `entityConfig`, and
```
rockymadden, line 307: **major** — `_resourceKey` / `_entityKey` is required here and both `bindings[]` rows are scoped to it, but the doc never says what value to use or where to read it.

Siblings anchor `resourceKey` to the definition's `model.bindings.resourceKey`; no equivalent is given for Entity, and `uip df entities list` returns `id` / `folderId`, not a resource key. As written the hand-authoring path for a folder-scoped entity is unfollowable.

The closing line of this section ("add them yourself or leave the entity tenant-scoped") hints at it — make it explicit: state that `_resourceKey` is canvas-generated with no CLI source, so hand-authored folder-scoped entities should be left to the picker.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
...
+The validator enforces the "green but inert" cases it can see structurally:
+
+| Message | Means |
+| --- | --- |
+| `An entity must be selected` | `entityName` missing or empty |
+| `Set a value for at least one field` | Create's `fieldValues` has no row with a non-blank value |
+| `Select a field to update` | Update's `fieldUpdates` has no row naming a field |
+| `A record ID is required` | `recordSource: "byId"` with a blank or whitespace-only `recordId` |
+| `Select a Read entity records node` | `recordSource: "fromRead"` with no `readEntityNodeId` |
+| `Choose how to identify the record` | Delete with `recordSource` missing or not one of `byId` / `fromRead` |
```
rockymadden, line 385: **minor** — this row names Delete only, but `planning.md:128` marks `recordSource` **Required: Yes** for Update, and "What it does not check" below doesn't mention it either.

1ab64a31 deleted the old sentence ("`recordSource` is required on Delete, unlike Update, which tolerates its absence on legacy configs") without replacing it, so the doc now implies validate catches it on Update while this table says otherwise.

Given the guide's whole thesis is *what validates clean and still does nothing*, state it either way.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
+# Data Fabric Entity Nodes — Implementation
+
+Four OOTB nodes that read and write records in a UiPath Data Fabric entity: `core.datafabric.read`, `core.datafabric.create`, `core.datafabric.update`, `core.datafabric.delete`. Each is a `bpmn:Task` host with no `serviceType`, and the BPMN engine does the actual work — but by two different mechanisms:
+
+- **Read** emits its compiled query as a `<uipath:input>`. The engine's **input preprocessor** resolves it *before* the task runs.
+- **Create / Update / Delete** emit a `<uipath:output>` carrying a `=datafabric...` target. The engine's **activity postprocessor** performs the write *after* the task runs.
+
+That split matters when reading generated BPMN: a Read node with no `<uipath:output>` action target is correct, not a serialization failure.
+
+Shared action-node boilerplate — skeleton, ports, add/edit mechanics — is in [shared/action-nodes.md](../../../shared/action-nodes.md), with one exception noted under [No error port](#no-error-port). This file covers only what is specific to the entity nodes.
```
rockymadden: **minor** — two things here.

"with one exception" undercounts: there are two. No error port, *and* no instance `outputs` block (line 88).

The `outputs` divergence isn't reconciled anywhere. This line points at `shared/action-nodes.md` for the "skeleton", and that skeleton **includes** `outputs.output` + `outputs.error`, with `action-nodes.md:76` calling the error output "Implicit on every action node." `queue/impl.md:36` — the closest structural analogue (OOTB, user-owned, no `error` port) — ships a full `outputs` block with `error`.

`action-nodes.md:66` already has an **Exceptions:** paragraph. Add the carve-out there so the shared doc and the plugin agree.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
...
+
+Both write editors take `[{ field, value }]` rows, but what a value means differs by verb and by column type. The serializer coerces each value using the column's declared type from the `_outputSchema` snapshot; a value it cannot coerce is passed through as a string for the API to reject, **and that rejection is only logged** — it surfaces as an unchanged row, not a failed run.
+
+| Column kind | What the row's `value` must be |
+| --- | --- |
+| String | The text itself |
+| Integer / number | A clean numeric literal. `42.5` in an integer column is not coerced and is rejected |
+| Boolean | `true` or `false` |
+| **Choice set (single)** | The choice's numeric **`numberId`**, not its label — the column's SQL type is `INT` |
+| **Choice set (multi)** | A JSON array of those numeric ids |
```
rockymadden: **minor** — ambiguous. Every row example in this file uses a string `value`, so is this `"value": [1, 2]` or `"value": "[1,2]"`? One concrete example settles it.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,422 @@
...
+
+A **folder-scoped** entity carries `_folderKey` (the entity's `folderId`). Its presence switches the emitted target to the folder-qualified form, and for the flow to survive export to another org the entity's name and folder must serialize as binding tokens rather than source-org literals. That requires:
+
+1. `_resourceKey` (preferred) or `_entityKey` on the `entityConfig`, and
+2. two top-level `bindings[]` rows for that key, with `resource: "Entity"` and `propertyAttribute` of `name` and `folderKey`.
+
+```json
+"bindings": [
+  {
+    "id": "<bindingId1>",
```
rockymadden: **nit** — `file-format.md:549` prefers descriptive ids (`bDepositRpaName`) over placeholders. Suggest `bOrdersEntityName` / `bOrdersEntityFolder`; placeholders like `<bindingId1>` get copied literally.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
```diff
@@ -0,0 +1,185 @@
...
+
+The canvas also persists schema snapshots (`_entityFields`, `_relatedFields`, `_outputSchema`, `_choiceSets`). They are not decoration — see [impl.md — Schema snapshots](impl.md#schema-snapshots) for when authoring without them silently breaks.
+
+### `core.datafabric.read`
+
+| Key | Required | Description |
+| --- | --- | --- |
+| `resultMode` | No | `"single"` (default) or `"multiple"` |
+| `_filters` | No | The query. Grouped filter model — see [impl.md — Filters](impl.md#filters) |
+| `_selectedFieldNames` | No | Columns to return. Omit for all |
```
rockymadden, line 128: **minor** — **Required: Yes** here, but `impl.md:381` lists the `Choose how to identify the record` validator message for Delete only. See that line; the two need to agree.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
```diff
@@ -0,0 +1,185 @@
...
+
+**These four nodes have no `error` port.** Unlike most action nodes, none of them declares `supportsErrorHandling`, so no error handle is ever injected and the properties panel offers no error-handling toggle. Do not set `inputs.errorHandlingEnabled` and do not add an edge with `sourcePort: "error"` — the serializer will still emit a boundary event for it, but with no error mapping, and the canvas shows no handle the author can see or remove. Plan failure handling upstream or downstream instead.
+
+Delete's `output` port exists for **sequencing only** — it carries no data (see below).
+
+## Output Variables
+
+| Node Type | `$vars.{nodeId}.output` |
+| --- | --- |
+| `core.datafabric.read` (`single`) | The record itself — address columns directly: `$vars.readOrder1.output.Status` |
```
rockymadden: **nit** — `planning.md` uses `readOrder1` / `listOrders1` (also lines 64, 102) while `impl.md` uses `readOrder` / `listOpenOrders`. Align the pair so the examples read as one set.

skills/uipath-maestro-flow/references/author/CAPABILITY.md
```diff
@@ -158,7 +160,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 - [planning-arch.md](planning-arch.md) — capability discovery, plugin index, topology design
 - [planning-impl.md](planning-impl.md) — registry lookups, connection binding, wiring rules
 - [plugins/](plugins/) — per-node-type planning + impl docs:
-  - [connector](plugins/connector/) — IS connector nodes (incl. Data Fabric activities)
+  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; the native [data-fabric](plugins/data-fabric/) nodes are the lighter path where the tenant has them enabled — their flags default to off, so confirm with `registry get` before choosing)
```
rockymadden: **minor** — a one-line index entry now carries three clauses of argument, and the availability-first point is already stated in six other places: this bullet, the common-tasks row (line 120), `planning-arch.md:172-175`, `connector/impl.md:246`, `data-fabric/planning.md:25`, `planning.md:58`, and `impl.md:41` + `:417`.

Keep the full case in `data-fabric/planning.md`; reduce this to something like "native Data Fabric entity CRUD; flags default to off".

rockymadden (changes requested):
## Summary

Adds a `data-fabric` plugin pair covering the four native OOTB entity nodes (`core.datafabric.read` / `create` / `update` / `delete`), which the skill previously only reached through `uipath-uipath-dataservice` connector activities. Registered in the same three index files as `queue`, plus a back-reference from `connector/impl.md` and a `SKILL.md` description line. CODEOWNERS points the dir at `@UiPath/DatafabricCodingAgent` + `@UiPath/maestro-cli-team`.

Organising the guide around *what passes `flow validate` and still does nothing* is the right framing, and it is what makes 607 lines worth their context cost. Doc size is in line with peers (`inline-agent` 643, `connector-trigger` 627, `ixp` 490).

## What I verified independently

| Check | Result |
| --- | --- |
| 8 maintenance checkers | all pass; the 1 bad anchor reproduces on `main` |
| `check-skill-verbs.py` on new files | clean (catalog 1.202.0-dev.8509) |
| `check-cli-verbs.py` | 0 High, 0 Medium |
| `check-uip-commands.sh` (Py 3.14) | `commands_checked=136 unknown=0` |
| `uip df` flags | `entities list --native-only/--include-folders/--folder-key`, `entities get --folder-key`, `records list --folder-key`, `records import` — all real on 1.202.0 |
| `registry search datafabric` | all four present, `AvailableOnTenant: false`; canvas labels match the doc exactly |
| `registry get core.datafabric.read` | `Node not found` — confirms the eval-task deferral |
| Node versions | read `1.3`, create `1.3`, update `1.2`, delete `1.2` — confirms "versioned independently" |
| Registration / depth / orphans | complete, ≤2 hops, 0 orphans |
| `uip df` inside this skill | already precedent (`connector/impl.md:178`, `operate/run.md:38`) — not a skill-boundary violation |

## Blocking

Two inline findings make the agent emit wrong output, and both are cheap to fix:

1. **`bindings[]` shape contradicts `shared/file-format.md`** on which field the serializer matches (`propertyAttribute` vs `(resourceKey, name)`) and on what `name` holds. The shared doc also scopes top-level `bindings[]` to Orchestrator resource nodes only and isn't updated to add the `Entity` kind. Needs a Data Fabric codeowner to confirm at source, then one of the two files corrected.
2. **`=js:new Date().toISOString()`** in the Update example, against `shared/variables-and-expressions.md:577` ("`Date` constructor may have limited support — prefer ISO 8601 strings"). Only `new Date()` in the entire skill, in a copy-paste example, in the one node family where a bad value is rejected and swallowed.

The rest of the inline comments are minor or nits and can land in the same pass or as follow-ups.

## PR description correction

> create/update/delete are absent from the CLI's bundled manifest entirely (it ships `@uipath/flow-core@0.96.0-develop`, which predates them)

Stale. On `uip` 1.202.0, `registry search datafabric` returns all four with `AvailableOnTenant: false`. The tenant flags are the only blocker for the eval task, not the CLI version — worth fixing so the follow-up is scoped correctly.

The "no eval task" note also slightly overstates convention: `queue`, `batch-transform`, `summarize` and `delay` have no task dir either. The deferral is fine; it should be a tracked follow-up rather than a note, since nothing here can be exercised end-to-end today.

## Overall risk

Not slop — the content is dense and mostly load-bearing. The exposure is different: **607 lines of very confident behavioural claims, zero runtime verification, no eval task.**

Commit 1ab64a31 is the evidence. That one review pass caught an invented column name in a copy-paste example (`CreatedAt` → `CreateTime`), a fabricated engine-default story for `_recordLimit`, and a nonexistent CLI subcommand (`uip df records query`). All eight checkers were green before those fixes too — no gate in this repo catches that class of error. One pass found three fabrications; the prior that residual ones remain is not low. Findings 1 and 2 are two more of the same shape, found by cross-reading against the skill's own shared docs rather than the product source.

**Recommendation: request changes** on 1 and 2. Structure, routing, ownership and gates are all correct and this closes a real gap — but under the "does this change what the agent does?" test, both of those change it in the wrong direction.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Checked at source — **neither file was wrong**; they describe two different mechanisms and nothing said so.

Orchestrator resource bindings resolve a `<bindings.{name}>` placeholder from the definition’s `model.context[]`, matched by `(resourceKey, name)`. Entity bindings have no placeholder and no definition-side `model.bindings` at all — `findEntityBinding` (`services/src/serialization/entity-vdo.ts`) matches `(resourceKey, propertyAttribute)` falling back to `name`, and the serializer emits `bindings.<id>` straight into the `=datafabric[...]` expression. So `name` really is a display label here (`Orders` / `OrdersFolder`), and there is genuinely no `resourceSubType` — the entity picker passes `undefined`.

Rather than "fix" either file, `shared/file-format.md` now has a **Data Fabric entity bindings** subsection tabulating all five differences (`resource` capitalization, the matching key, what `name` holds, `folderKey` vs `folderPath`, absent `resourceSubType`), and the plugin cross-links it instead of contradicting it. The existing H2 anchor is untouched — five sibling plugins link to it.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Agreed, and worse than a style nit — you are right that this is the one node family where an uncoercible value is rejected and swallowed.

Replaced with an ISO string from a `$vars` reference, plus a note under the Update example pointing at Script for timestamps and stating why `new Date()` is wrong here specifically (limited Jint `Date` support, and the rejection is silent so the column just stays unchanged).

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Fair — as written the folder-scoped path was unfollowable.

One correction to the suggested wording: there *is* a CLI surface, just not one that mints a key. `_resourceKey` is the key the entity’s reference is registered under in the solution (written by the picker via `referenceEntityToSolution`), and a registered reference shows up under `uip solution resources list --kind Entity --output json`. What no CLI does is create one for a flow you are hand-authoring. `_entityKey` (the DF entity id from `uip df entities list`) is the older fallback, used only when `_resourceKey` is absent.

The section now says all of that and ends where you pointed: if you cannot resolve a `_resourceKey`, do not hand-author the folder-scoped form — keep the entity tenant-scoped or let the picker write `_folderKey`, `_resourceKey` and both rows on first open, because a half-authored folder scope serializes and then breaks on deploy.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Confirmed at source, and the table row was the correct half: Update’s `inputDefinition` requires only `entityName` + `fieldUpdates`, while Delete’s requires `entityName` + `recordSource`. So validate genuinely does not catch a missing `recordSource` on Update.

`planning.md` no longer claims Required: Yes for it, and this is now called out explicitly in "what it does not check" — an Update without `recordSource` validates clean and the serializer infers the mode from whichever selector is populated. Given the guide’s thesis, you were right that it had to be stated either way.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Both correct. The intro now says **two** divergences and links each one.

The reconciliation landed where you suggested: `shared/action-nodes.md`’s **Exceptions:** paragraph now records that `core.datafabric.*` carries no instance `outputs` block and has no `error` port. I also softened the Standard-ports row, which claimed the error port was "Implicit on every action node" — it is implicit on nodes declaring `supportsErrorHandling`, which none of these four do.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Settled with a concrete example. It is a JSON array **as a string**: `{ "field": "Tags", "value": "[1,2]" }`.

`EntityChoiceSetMultiValue.tsx` documents the wire form verified against a live run — the column is `NVARCHAR(4000)`, so `buildWriteBodyFromFieldValues` applies no coercion and writes the text through for Data Fabric to parse. The row now states that explicitly so it cannot be read as a real JSON array in the `value` slot.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Agreed — placeholders do get copied literally. Now `bOrdersEntityName` / `bOrdersEntityFolder`, matching the `bDepositRpaName` convention in `file-format.md`.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
uipath-utkarsh: Aligned — see the reply on `impl.md`. Update’s schema requires only `entityName` + `fieldUpdates`, so the validator does not enforce `recordSource` there; this row now says so rather than marking it Required: Yes.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
uipath-utkarsh: Aligned to the `impl.md` set — `readOrder` / `listOpenOrders` throughout the pair.

Fixed in `dacc284d9`.

skills/uipath-maestro-flow/references/author/CAPABILITY.md
uipath-utkarsh: Agreed, that bullet had drifted into carrying the argument. Reduced to an index entry: the connector line is back to a `cf.` cross-reference, and the data-fabric line reads "native Data Fabric entity record CRUD (`core.datafabric.*`); tenant flags default to off". The full case stays in `data-fabric/planning.md`.

Fixed in `dacc284d9`.

uipath-utkarsh:
Thanks — this was a genuinely useful pass. All ten threads addressed and resolved in `dacc284d9`; both blocking findings were real.

**On finding 1 (bindings), since you flagged it for a Data Fabric codeowner:** I confirmed at source, and the answer is that *neither* file was wrong — they describe two different mechanisms and nothing said so. Orchestrator resource bindings rewrite a `<bindings.{name}>` placeholder matched by `(resourceKey, name)`. Entity bindings have no placeholder and no definition-side `model.bindings`: `findEntityBinding` matches `(resourceKey, propertyAttribute)`, `name` is a display label, and `resourceSubType` is genuinely absent because the picker passes `undefined`. So rather than correcting one file, `shared/file-format.md` gained a **Data Fabric entity bindings** subsection tabulating the five differences, and the plugin cross-links it. The H2 anchor is untouched — five sibling plugins depend on it. Still worth a Data Fabric codeowner sanity-check, but it should no longer read as a contradiction.

**Finding 2** — agreed, and your framing was the right one: the risk was not style but that it sat in the one family where a bad value is swallowed. Now an ISO string from `$vars`, with a note pointing at Script for timestamps.

**On the PR-description correction:** we are both right, and it is a CLI-version difference. `uip` 1.202.0 returns all four with `AvailableOnTenant: false`; my local 1.201.0-dev.8247 returns only `core.datafabric.read`, which is what the stale claim was based on. Rewritten to lead with the real blocker — tenant flags, not CLI version — with both observations recorded so the follow-up is scoped correctly. Also softened the convention claim, since `queue` / `batch-transform` / `summarize` / `delay` ship without task dirs too.

**On your "zero runtime verification" point** — that landed, and the smoke tier proved it. An earlier revision of this branch framed the choice as an unconditional "prefer the native node"; `skill-flow-datafabric-smoke-update`, which builds against the connector, switched to native nodes and failed its verifier. Every static gate was green while that was true. Fixed in `69386f79` by making the choice availability-first (`registry get` decides; flags default off, so the connector is the working path until proven otherwise). That task and `smoke-error` are now SUCCESS and the tier passes at 96% — the one remaining failure, `smoke-update-existing-flow`, also fails on unrelated branches.

Green: 17 checks, 8 maintenance checkers, `skills:validate` both flavors, verb checker clean on all new and edited files.

skills/uipath-maestro-flow/references/shared/file-format.md
```diff
@@ -556,6 +558,24 @@ Each resource node needs two binding entries (one for `name`, one for `folderPat
...
+
+The value is in `default` for both forms. There is also no `<bindings.{name}>` placeholder to rewrite: the entity has no definition-side `model.context[]`, and the serializer emits `bindings.<id>` tokens directly into the `=datafabric[...]` expression it builds.
+
+`folderKey` rather than `folderPath` is load-bearing. The platform's deploy-time override for an `Entity` resource carries `{ name, folderPath, folderKey }` and rewrites binding defaults by `propertyAttribute`; a `folderPath` row receives the Orchestrator FQN, which the Data Fabric query rejects even in the source org.
+
+Tenant-scoped entities need no bindings at all — they serialize as a dotted literal. See [data-fabric/impl.md — Folder-scoped entities and bindings](../author/plugins/data-fabric/impl.md#folder-scoped-entities-and-bindings) for the full JSON and when to hand-author it.
+
 **Not to be confused with `bindings_v2.json`.** That file holds connector connection bindings for Integration Service nodes — a separate system. A flow may have both: a top-level `bindings[]` for resource references and a `bindings_v2.json` file for connector connections.
```
rockymadden: **minor** — the new `### Data Fabric entity bindings` went in at :561, *before* this paragraph. The Orchestrator section's own closing note now reads as part of the Data Fabric subsection.

Related: the reason given for H3 (protect the H2 anchor five plugins link to) does not hold. A sibling `## Data Fabric entity bindings` placed after this paragraph leaves `#bindings--orchestrator-resource-bindings-top-level-bindings` completely untouched. As it stands, non-Orchestrator content is nested under a heading whose first line (:515) scopes it to Orchestrator resource nodes, and the file's Table of contents at :5 offers no route to it.

Promote to H2, move it below this paragraph, add a TOC row.

skills/uipath-maestro-flow/references/shared/file-format.md
```diff
@@ -556,6 +558,24 @@ Each resource node needs two binding entries (one for `name`, one for `folderPat
 
 See each resource plugin's `impl.md` for the full JSON per node type: [rpa](../author/plugins/rpa/impl.md), [agent](../author/plugins/agent/impl.md), [flow](../author/plugins/flow/impl.md), [agentic-process](../author/plugins/agentic-process/impl.md), [api-workflow](../author/plugins/api-workflow/impl.md), [hitl](../author/plugins/hitl/impl.md).
 
+### Data Fabric entity bindings
+
+A **folder-scoped** Data Fabric entity (`core.datafabric.*` nodes) writes two rows into this same `bindings[]` array, but they are not Orchestrator resource bindings and four of the rules above do not carry over:
```
rockymadden: **nit** — "four of the rules above do not carry over", above a five-row table. `impl.md:333` says "four of these differ" over four bullets, two of which lead with a *sameness* (`default` holds the value, `type` is required). The PR body says five.

Same shape as the "one exception" undercount from the last pass. Pick a number or drop it.

skills/uipath-maestro-flow/references/shared/action-nodes.md
```diff
@@ -73,7 +75,7 @@ For action nodes, the instance `outputs` block is documentation, not the runtime
 | --- | --- | --- |
 | Input (target) | `input` | Every action node accepts one input edge on `input`. |
 | Output (success, source) | `output`, `default`, or `success` | The name varies by plugin; `registry get` is authoritative. |
-| Output (error, source) | `error` | Implicit on every action node via `outputs.error`. |
+| Output (error, source) | `error` | Implicit on action nodes that declare `supportsErrorHandling`, via `outputs.error`. The `core.datafabric.*` family does not — it has no `error` port. |
```
rockymadden, line 78: **minor** — this row is right now, but the parallel claim in `shared/file-format.md` did not get the same treatment:

- `:308` "This applies to HTTP, Script, Transform (all variants), connector activities, agent nodes, and RPA nodes, essentially every action node."
- `:296` "Connector activities, agent nodes, and RPA nodes follow the same pattern ... a primary source port plus an implicit `error` port."

The governing clause at :308 is conditional (`Any node with supportsErrorHandling: true`) so it is not wrong, but "essentially every action node" is exactly the generalisation you removed here, and `file-format.md` is the more-linked of the two docs. One clause fixes it.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -0,0 +1,427 @@
...
+
+Field-by-field, because four of these differ from the Orchestrator-resource bindings documented in [file-format.md — Bindings](../../../shared/file-format.md#bindings--orchestrator-resource-bindings-top-level-bindings). That section's rules describe resource nodes; the `Entity` kind resolves differently, and the differences are tabulated in [file-format.md — Data Fabric entity bindings](../../../shared/file-format.md#data-fabric-entity-bindings):
```
rockymadden: **minor** — this line says the differences "are tabulated in file-format.md", then tabulates them again as four bullets. Both are near-full statements, neither is a pointer. `folderKey` not `folderPath` now appears three times: the `file-format.md` table row, `file-format.md:575`, and `:340` below.

`.claude/rules/token-optimization.md` §0 is explicit: one full statement plus N pointers, never N paraphrases.

Keep the table in `file-format.md`. Reduce these bullets to the two things an author physically types (capital `"Entity"`, and the `propertyAttribute` values) and let the link carry the rest.

skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -165,6 +169,10 @@ Prefer, in order:
...
 3. An RPA workflow only when there is no API, such as a desktop app or terminal ([rpa](plugins/rpa/planning.md)).
 
+**Data Fabric has two paths, and availability decides which.** Entity records can be reached either through the `uipath-uipath-dataservice` connector activities ([connector](plugins/connector/planning.md)) or through the native `core.datafabric.*` nodes ([data-fabric](plugins/data-fabric/planning.md)).
+
+The native nodes are nicer when they are there — no Integration Service connection, no `node configure` — but **all four tenant flags default to off**, so on most tenants they do not exist. Decide by checking, not by preference: run `uip maestro flow registry get core.datafabric.read` (and the sibling type for the operation you need). If it returns the node, use the native node. If it answers "Node not found" or search reports `AvailableOnTenant: false`, **use the connector activities** and do not spend further turns on the native path. Also use the connector when the entity is federated — the native writes require a native entity — or when you need an operation the four nodes do not cover.
```
rockymadden: **minor** — `69386f79` fixed the regression but left three near-identical four-sentence paragraphs: here, `data-fabric/planning.md:58`, and `connector/impl.md:246`. Each restates flags-default-off, run `registry get`, "Node not found" means connector, federated means connector.

The connector callout earns its place: that reader must not have to open the plugin to learn this. These other two can be one sentence plus a link to `data-fabric/planning.md`.

I flagged this on the CAPABILITY.md bullet last pass; the fix commit widened it rather than closing it.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
```diff
@@ -0,0 +1,185 @@
...
+
+| Key | Required | Description |
+| --- | --- | --- |
+| `fieldValues` | Yes | `[{ field, value }]`. **At least one row must carry a non-blank `value`** — blank values are omitted so the column default applies, so an all-blank config inserts nothing and still runs green |
+
+### `core.datafabric.update`
+
+| Key | Required | Description |
+| --- | --- | --- |
+| `recordSource` | Always author it | `"byId"` or `"fromRead"`. **The validator does not require it on Update** (unlike Delete) — the serializer infers the mode from whichever selector is populated, so leaving it out is silent |
```
rockymadden: **nit** — `Always author it` sits in a column headed **Required**, where every other value across the pair is `Yes`, `No`, `byId`, `fromRead`, or `multiple` only. An agent scanning that column for a token now hits prose.

`Yes (not enforced)` keeps both the instruction and the vocabulary.

skills/uipath-maestro-flow/references/author/CAPABILITY.md
```diff
@@ -158,7 +160,7 @@ If you find yourself hand-writing `inputs.detail`, a `=jsonString:` blob, or `bi
 - [planning-arch.md](planning-arch.md) — capability discovery, plugin index, topology design
 - [planning-impl.md](planning-impl.md) — registry lookups, connection binding, wiring rules
 - [plugins/](plugins/) — per-node-type planning + impl docs:
-  - [connector](plugins/connector/) — IS connector nodes (incl. Data Fabric activities)
+  - [connector](plugins/connector/) — IS connector nodes (incl. the `uipath-uipath-dataservice` entity activities; cf. [data-fabric](plugins/data-fabric/))
```
rockymadden: **nit** — `cf.` is the only occurrence in the skill. A Latin abbreviation in an agent-facing index line. `see` costs one more character.

rockymadden (approved):
## Re-review

All ten threads answered and closed, and both blocking findings are genuinely fixed.

| Prior finding | Status |
| --- | --- |
| 1. `bindings[]` shape vs `file-format.md` | Resolved, and better than what I asked for. Neither file was wrong; `file-format.md` gains a **Data Fabric entity bindings** subsection and the plugin cross-links it instead of contradicting it. Existing H2 anchor untouched. |
| 2. `=js:new Date().toISOString()` | Resolved. Now `=js:$vars.start.output.shippedAt`, matching the `$vars.start.output.*` form the Create and Delete examples already use, plus a note saying why `new Date()` is wrong *here* specifically. |
| 3. `_resourceKey` unsourced | Strongest fix in the batch. Names the real CLI surface (`uip solution resources list --kind Entity`, verified real on 1.202.0), states no CLI mints one, and ends on a stop rule. |
| 4 / 8. `recordSource` on Update | Confirmed at source and resolved in the direction the validator actually behaves. Now in "what it does not check". |
| 5. "one exception" undercount | Now two, both recorded in the shared Exceptions note. The Standard-ports row correctly generalised from "every action node" to "declares `supportsErrorHandling`". |
| 6, 7, 9, 10 | Applied. |

### Re-verified independently

8 maintenance checkers (links 898/0, link-text 571/0, anchors 286 checked / 1 bad reproducing on `main`, depth 84/84, orphans 0, plugin-pairs 27/0, template 0, versions 0). `check-uip-commands.sh` on Python 3.14: `commands_checked=136 unknown=0`. `uip solution resources list --kind Entity` exists on 1.202.0. All 21 CI checks green including the 23m smoke tier. No `readOrder1` / `listOrders1` left anywhere.

The `69386f79` story is the one that matters. An unconditional "prefer the native node" framing broke a smoke task while every static gate stayed green. That is the runtime verification I said was missing, and it caught a real regression.

### Remaining

Seven inline comments, all minor or nit. None changes what the agent emits. Two of them leave a shared doc in a worse state than the plugin that edited it, so please take those before merge; the rest are fine as follow-ups or in the same pass.

One thing not attached to a line: the PR body says the eval task is "worth tracking as a follow-up" but links no issue. Nothing here can be exercised end to end until a tenant enables the four flags. Please file it and link it, or the deferral evaporates on merge.

### Overall

Residual risk is unchanged and accepted: dense behavioural claims, verified at source, exercised by nothing on a tenant. Two review passes have found nine real defects between them, so the prior on more is not zero. That argues for the eval task, not for holding this PR.

**Approve.**

skills/uipath-maestro-flow/references/shared/file-format.md
uipath-utkarsh: You are right on both counts, including that my H3 rationale did not hold — a sibling H2 after the note leaves the Orchestrator anchor completely untouched, which is all the five linking plugins need.

Promoted to `## Bindings — Data Fabric entity bindings`, moved below the `bindings_v2.json` note so that note is back inside the Orchestrator section, added the TOC row, and repointed both inbound links to the new anchor. The file now reads as three sibling `## Bindings —` sections.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/shared/file-format.md
uipath-utkarsh: Dropped rather than renumbered — same call as the "one exception" fix. The sentence now reads "most of the rules in the section above do not carry over", and `impl.md` no longer counts either, since the bullets there were cut back to two.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/shared/action-nodes.md
uipath-utkarsh: Caught — I fixed the copy I had touched and left the more-linked one. Both claims in `file-format.md` now name the exception: the `supportsErrorHandling` paragraph ends "most action nodes, but not all: the `core.datafabric.*` family does not declare it and has no `error` port", and the ports prose gets the same one-clause carve-out.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Agreed — a pointer followed by a full restatement is the worst of both. Cut to the two values an author physically types (capital `"Entity"`, and `propertyAttribute` of `name` / `folderKey`), with one line handing the rest to the `file-format.md` table.

Also removed the third `folderKey` not `folderPath` paraphrase you counted; the mechanism now lives once in `file-format.md`, and the plugin keeps only the terse "never `folderPath`" on the bullet plus its entry in the anti-pattern list.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/author/planning-arch.md
uipath-utkarsh: Fair, and you are right that the last fix widened it rather than closing it.

One note on the split, since "these other two" was ambiguous with `data-fabric/planning.md` also being the link target: I kept the plugin planning doc as the canonical full statement (it is the natural home, and what the other two point at), kept the connector callout intact per your reasoning, and reduced `planning-arch.md` to one sentence plus a link. The canonical paragraph is also tightened from four sentences to two. If you intended the canonical copy to live somewhere else, say so and I will move it.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
uipath-utkarsh: Agreed — prose in a token column. Now `Yes (not enforced)`, which keeps the instruction and the vocabulary.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/author/CAPABILITY.md
uipath-utkarsh: Fixed — `see`.

Fixed in `92c02b479`.

skills/uipath-maestro-flow/references/author/planning-arch.md
uipath-utkarsh: Reopening this one briefly: the reduction went too far and the smoke tier caught it.

Cutting `planning-arch.md` to a bare pointer dropped the sentence the planner actually acts on, and `smoke-create-all-types` + `smoke-query` — both connector tasks — failed at 88%. The transcripts show the same failure the availability-first fix was meant to prevent, in a worse form: create-all-types logged 61 `core.datafabric` mentions against 4 for the connector and ended *"not available through registry get because tenant availability was false … A schema-compatible definition was used"* — it hand-wrote a `definitions[]` entry rather than falling back. query ended *"No flow was created … the required activity definition is unavailable"* and gave up.

So the dedup instinct was right but that specific sentence was not a paraphrase — it is the fallback instruction, and the pointer does not substitute for it because the planner reads the ladder without opening the plugin.

Restored as **one** sentence rather than the original four, and made explicit about the two dead ends the agents actually took (do not retry, do not `uip tools update`, never hand-write a definition for a node the registry will not return). Rationale and the federated-entity case stay behind the link. The other two reductions from this thread stand.

`f61398bfd`.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
```diff
@@ -0,0 +1,185 @@
...
+## Node Types
+
+| Node Type | Canvas label | Does |
+| --- | --- | --- |
+| `core.datafabric.read` | Read entity records | Read one record, or a filtered list of records |
+| `core.datafabric.create` | Create entity record | Insert one record and return the stored row |
+| `core.datafabric.update` | Update entity record | Patch named columns on one record |
+| `core.datafabric.delete` | Delete entity record | Delete one record (no output) |
+
+These are fixed OOTB node types — no registry suffix, no connector key. Each is gated by its own tenant feature flag, so a tenant can have Read without Create:
```
snuziale, line 16: Why are the feature flags in the skill?

uipath-utkarsh: Good question, and the answer is that they shouldn't be. Fix in #3194.

Three things were wrong with them:

**An agent can't act on a flag.** No `uip` command reads or reports `canvas.nodes.*`. The only observable is whether `registry get` resolves, so the tables were unactionable detail.

**They aren't tenant flags.** They're flow-core `ManifestClient` flags — the CLI never asks the tenant, it answers `getFeatureFlag` from a hardcoded allowlist (`ENABLED_MANIFEST_FLAGS` in `node-sync-service.ts`). So "a tenant can have Read without Create" was never true, and impl.md's **"Flag to ask the admin about"** column was advice that could not work. That's precisely how this got misdiagnosed in the first place: the CLI said "not enabled for this tenant", people checked the tenant, and the tenant was never in the path.

**They're now wrong.** UiPath/cli#4087 added all four to that allowlist, so the nodes resolve — while these docs still routed agents to the Data Service connector to avoid them.

Chasing it turned up the same defect in Summarize and Batch Transform, which claimed the node "appears after server rollout to the tenant". `getManifest` builds from the local flag-filtered OOTB set and only pushes dynamic nodes on top, so the server never withholds an OOTB node; both types resolve today. And the generic `AvailableOnTenant` definition in `shared/cli-commands.md` said "not enabled for the tenant", which would have repeated this for the ~9 families still absent from the allowlist.

All of it now says the same thing: probe with `registry get`, treat the result as a property of the CLI build, fall back to the connector on "Node not found" — which is also version-robust, since #4087 hasn't reached a stable CLI yet.

PR #3042, opened 2026-09-03
docs(uipath-maestro-flow, uipath-platform): filter tree only for CEQL; no --folder-path on flow debug

jiyangzh:
## What

**CEQL: the filter tree is the only sanctioned form.**

- `author/plugins/connector/impl.md` Step 6a now starts with two verifiable commands: `uip maestro flow registry get <node-type> --output json --output-filter "Node.connectorMethodInfo.parameters[?design.component=='FilterBuilder'].{name:name,type:type}"` (which parameters take the tree) and the path/`pathParameters` projection (never hand-type `endpoint`). Every returned name goes under the `filter` key of `--detail`. Runtime values go in a leaf operand with `"isLiteral": false`; the CLI compiles them to `{var_<hash>}` + `filterVariables`.
- The "Hand-authored CEQL strings" preference list and its `=js:` template are removed. One sentence replaces them: a whole-value `=js:` query string is not read by `validate` and is warned by the CLI (cli#3966); replace it with a tree.
- Three CEQL error rows collapse into one (`[102003]`/`[102001]` → configure a tree).
- All five mentions of `--detail.filter` — a flag that does not exist — now say "the `filter` key of `--detail`".
- `uipath-platform/references/integration-service/activities.md`: "Never hand-write the CEQL string" at the top of the CEQL section; two stale "isLiteral: false not yet supported" claims corrected (the file contradicted itself); one anti-pattern row.
- Trigger docs (`connector-trigger/impl.md`, platform `triggers.md`): filter-tree operands may be dynamic; `eventParameters` must be literals, because a trigger is evaluated before any flow run exists.

**Debug: do not pass `--folder-path`.**

- `operate/run.md` debug section: a rule block after the five debug steps — debug provisions into the personal workspace; a shared folder fails with `HTTP 500` at `Stage: prepare-custom-debug` and no instance starts; shared resources reach the run through `solution resources refresh`; the two legitimate uses of the flag are named.
- `diagnose/troubleshooting-guide.md` Step 0 fault table: a row mapping that 500 to the flag.

## Why

Two nightly invoice-lookup runs (2026-09-02/03) hand-wrote the Data Service query as a `=js:` string. One replaced every quote with `String.fromCharCode(39)` and quoted the field name → `[102003] Expected a field name expression`; one typed the endpoint by hand → `[102001]`. `flow validate` cannot read a `=js:` query that is not a plain concatenation; `node configure` accepts it. The agent had read the rule "never pass a raw CEQL string" — and then the next section, which offered a hand-authored `=js:` string "when a runtime value requires it". That condition is never true: the tree takes dynamic operands. Agents followed the most concrete text, which was the template.

Four resolution-task debug runs (2026-09-02/03) passed `--folder-path 'Shared/…'` because the task names that folder as the index's location; every one died with an unexplained 500 before an instance started, so the agent never saw the flow's real fault. Without the flag the same flow ran 10 of 11 nodes and surfaced a fixable fault.

## Verified

- `registry get` projection returns `[{"name":"queryExpression","type":"query"}]` for Data Service `query-entity-records` (live). The tree form with `"isLiteral": false` compiles and runs (live, MST-13005; invoice-lookup scored 1.0 with the tree earlier in this campaign vs 0.45 hand-written).
- `flow debug` without `--folder-path` on the failing resolution flow: prepared, imported, ran 10/11 nodes (live); with the flag: 500, 2/2 (stock and dev builds alike).
- `npm run skills:validate` passes; all edits outside `skill-flavor` blocks; all anchors verified.
- coder eval test: https://github.com/UiPath/skills/actions/runs/33815202796

Docs only. No test changes.


DevMomo (approved):

jiyangzh:
**lookup still handing writing, just correct hand writing this time.** 😮‍💨
It just refused to use a stable approach 
The prompt change, may not be that useful

bryan-uipath (approved):

rockymadden (approved):

PR #3043, opened 2026-09-03
fix(ci): scope the skills.sh drift check to drift the PR introduces

uipreliga:
## Problem

`Validate skills.sh.json against skills/` failed **68 times in the last three months — a 6.1% failure rate**, the highest of any deterministic gate in this repo. **65 of those 68 failures were a single incident.**

#2252 landed `skills/uipath-process-mining/` on 2026-08-04 with no `skills.sh.json` entry. The checker reads the **whole tree**, so from 15:52 that day until 17:13 on 2026-08-06, every PR that touched any `skills/*/SKILL.md` went red for drift it did not cause:

| | |
|---|---|
| Failures in the 48h window | **65** of 68 (96%) |
| Unrelated branches hit | **~20** (`feat/inline-agent-flow-file`, `agent/studioweb-skill-flavor`, `feat/case-*`, `fix/reduce-review-verbosity`, …) |
| How it ended | An unrelated PR (#2498, a *solution zip-filename fix*) happened to add the missing entry |
| Legitimate failures | 3 — skill-adding PRs, correctly red |

That false-positive rate is also why the check **merged red 5 times**: a check that cries wolf gets clicked through, and that is how the drift reached `main` to begin with.

## Fix

**`--baseline-ref` scopes the exit code to drift the change introduces.** Findings that also reproduce at the base commit print as `::warning::` and do not fail the run; only new findings exit 1. Whole-tree visibility is kept — the blast radius is not.

An unreadable baseline (shallow clone, missing base, a branch predating `skills.sh.json`) falls back to strict whole-tree behaviour. The failure mode is the *old* behaviour, never a silent pass.

Replayed against the real incident commits:

```
#2252 (adds process-mining)      new=[uipath-process-mining]   exit 1   ✓ still blocks
#2436 (unrelated, maestro-case)  pre-existing                  exit 0   ✓ no longer blocks
#2498 (adds the entry)           clean                         exit 0   ✓
```

## Documentation

`skills.sh.json` is not derived from disk, so a rename or removal leaves it stale with **no symptom other than this check**. `CONTRIBUTING.md` covered *adding* a skill — but it is not loaded into agent context. `CLAUDE.md` and `.claude/rules/` are, and neither mentioned the file at all. That is the actual reason agents keep missing it.

Added the add / rename / remove contract to:

- **`CLAUDE.md`** — Contribution rule 5 (both registries in the same PR as the folder change) + two `File Conventions` rows
- **`.claude/rules/skill-structure.md`** — new `## skills.sh Grouping` section with a change→edit table
- **`.claude/rules/pr-review.md`** — new-skill checklist and a rename/removal item in the modification checklist
- **`CONTRIBUTING.md`** — `#### Renaming or removing a skill` subsection

## Tests

`tests/scripts/test_check_skills_sh.py` — 15 cases pinning the contract:

- pre-existing drift warns, does not fail
- introduced drift fails
- a dirty baseline is **no shield** for fresh drift
- renames fail on both halves (old name gone + new name ungrouped)
- unreadable / unparseable baseline stays strict
- the tree is currently clean

Registered as a Test Helpers job (`skills.sh grouping checker unit tests`).

## Expected impact

Projecting this quarter's data forward: **68 failures → 3**. Combined with the docs change closing the authoring gap, this makes the check safe to add to the required set — it is Tier 1 in the CI audit.

## Verification

```bash
python3 -m pytest tests/scripts/test_check_skills_sh.py -q   # 15 passed
python3 scripts/check-skills-sh.py                           # OK — 27 skills grouped across 4 section(s).
python3 scripts/check-skills-sh.py --baseline-ref origin/main
```

### DIFF
--- a/.claude/rules/pr-review.md
+++ b/.claude/rules/pr-review.md
@@ -15,6 +15,7 @@ When reviewing or creating pull requests for this repository, enforce these rule
 - [ ] Reference files use kebab-case naming
 - [ ] All relative links resolve to existing files
 - [ ] CODEOWNERS has been updated with the new skill path
+- [ ] Registered in `assets/skill-status.json` and grouped in `skills.sh.json` (run `python3 scripts/check-skill-status.py` and `python3 scripts/check-skills-sh.py`)
 - [ ] No secrets, tokens, or personal paths committed
 
 ### Existing Skill Modification Checklist
@@ -23,6 +24,7 @@ When reviewing or creating pull requests for this repository, enforce these rule
 - [ ] Critical Rules have not been removed without justification in the PR description
 - [ ] No new structural cross-skill dependencies (runtime delegation to a same-plugin sibling that degrades gracefully is allowed)
 - [ ] Reference file naming conventions preserved
+- [ ] **If the PR renames or removes a skill folder:** `skills.sh.json` and `assets/skill-status.json` are updated in the same PR — the old name is gone from both and the new one is present. Neither file is derived from disk, so a stale entry has no other symptom than the `Validate skills.sh.json against skills/` check
 - [ ] Changes are scoped to the skill being modified (no unrelated changes)
 
 ### Hook Changes Checklist
--- a/.claude/rules/skill-structure.md
+++ b/.claude/rules/skill-structure.md
@@ -67,6 +67,31 @@ python3 scripts/check-skill-status.py --write-readme
 
 `scripts/check-skill-status.py` (run in CI by `validate-skill-status.yml`) enforces that every skill has a manifest entry with a valid status, that the README table is current, and that no status markers leak into SKILL.md frontmatter or body.
 
+## skills.sh Grouping
+
+[`skills.sh.json`](../../skills.sh.json) is the display grouping for the repository's page on skills.sh. Every `skills/<name>/` MUST appear in exactly one grouping, and every grouped name MUST exist on disk.
+
+This file is presentation-only — it never changes what the `skills` CLI or `uip skills install` installs. That is exactly why it rots: nothing else in the build reads it, so a skill added, renamed, or removed under `skills/` produces no error anywhere except this one check, and the public page keeps showing the old grouping.
+
+**Any change to the set of skill folder names requires the matching edit to `skills.sh.json` in the same PR:**
+
+| Change to `skills/` | Required edit to `skills.sh.json` |
+|---|---|
+| Add `skills/<new>/` | Add `<new>` to the grouping that matches its purpose |
+| Rename `skills/<old>/` → `skills/<new>/` | Replace `<old>` with `<new>` — a rename is a remove plus an add, and both halves fail the check |
+| Delete `skills/<name>/` | Remove `<name>`; drop the grouping too if it is left empty |
+
+Validate before opening the PR:
+
+```bash
+python3 scripts/check-skills-sh.py          # must print "OK — N skills grouped across M section(s)."
+python3 scripts/check-skills-sh.py --fix    # removes stale entries for renamed/deleted skills
+```
+
+`--fix` will NOT place a newly added skill. Which of the four sections a skill belongs to (Authoring, Solution & Planning, Platform & Operations, Diagnostics & Feedback — the same four the README catalog uses) is an editorial judgement, so choose it by hand.
+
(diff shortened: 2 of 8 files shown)


bai-uipath (approved):
Fix what you agree with, otherwise lgtm. Baseline scoping is the right axis, and the incident replay holds up: I reproduced the 65 failures across 21 branches in that 48-hour window.

- **Finding identity keys on the formatted error string, so unrelated edits re-attribute pre-existing drift as new.** The duplicate message embeds the grouping titles and schema keys embed the list index, so retitling a grouping, or inserting one at the top, flips a pre-existing finding to `new` and blocks the PR. I confirmed both against the branch. Fix: key on the finding name plus a stable kind, and drop the index from schema keys.
- **Dropping the `paths:` filter belongs in this PR.** The check never ran at all on the PR that introduced the drift, so blast radius was not the only failure mode, and a required check is what would have caught it. This repo already documents in the version-sync and RPA-smoke workflows that a required check behind a `paths:` filter blocks merge forever. The script is a ~20s job, so just always run it.
- **`fetch-depth: 0` on the new test job does nothing.** Every test uses `HEAD` or a deliberately missing ref, and the suite passes on a depth-1 clone. It unshallows 320MB of history per run. Fix: delete it.

Minor, both optional: `HEAD^1` is a more self-consistent baseline than the event's `base.sha`, which can lag the merge ref, and would let the validate job's fetch-depth drop to 2. The add/rename/remove contract is now in both CLAUDE.md and the skill-structure rule, and one could point at the other.

uipreliga:
Agreed on all three, plus both optional items. Pushed in d8f78cd.

**1. Finding identity — you're right, and it was worse than described.** I reproduced both scenarios against the previous commit by reverting just the `identity()` body:

```
retitle a grouping (duplicate drift already present)  -> new=['uipath-rpa']                        exit 1
insert a grouping at the top (schema drift present)   -> new=["groupings[2] 'Broken'", 'uipath-rpa'] exit 1
```

Note the second case: inserting at the top re-attributed *both* the index-shifted schema finding and the coverage finding. Fixed as you suggested — findings now carry an explicit `kind` and `subject`, identity is `(kind, subject)`, and `key`/`error` stay display-only so output is unchanged. Coverage findings key on the skill name alone; schema findings key on the grouping title when valid, never the index. Both scenarios now exit 0. Five tests pin it, including a direct `identity()` invariance test.

**2. `paths:` filter — dropped, and your framing was the important correction.** I checked #2252 and you're right that blast radius wasn't the only failure mode: 19 checks reported on that head and `Validate skills.sh.json against skills/` was not among them, even though the workflow had existed since 2026-07-31 (#2390) and the filter listed `skills/*/SKILL.md`. The guard was silent on the exact PR it exists to catch. Now matches the rationale already documented in `validate-version-sync.yml` and `smoke-rpa-skills.yml`. A manual dispatch gets no baseline and validates the whole tree strictly.

This also turned up something bigger, which I've flagged in the audit doc rather than fixing here: **11 of the 12 checks that doc proposes making required are behind `paths:` filters** — all 8 `test-helpers.yml` jobs plus `validate-skills.yml`, `validate-skill-status.yml`, and `validate-skill-flavors.yml`. Turning those on as-is would block merge forever. Separate PR; didn't want to mix 4 unrelated workflow changes in here.

**3. `fetch-depth: 0` on the test job — removed.** Confirmed every test resolves `HEAD` or a deliberately missing ref, both fine on a depth-1 clone.

**Optional, both taken:**
- Baseline is now `HEAD^1`, and the validate job's `fetch-depth` drops 0 → 2. You're right that `base.sha` can lag the merge ref; `HEAD^1` is by construction the base parent of the commit actually under test. I'll confirm from the run log that no `cannot read baseline` warning appears before this merges — depth 2 is the thing worth verifying empirically.
- `CLAUDE.md` rule 5 now points at the skill-structure rule for the full table instead of restating the `--fix` limits, per `token-optimization.md` § 3.

Also fixed the review bot's catch: `test_cli_json_output_carries_the_preexisting_flag` passed by vacuous truth (clean tree emits no JSON, loop body never ran). Replaced with a fixture producing one pre-existing and one introduced finding so both `preexisting` values actually emit, plus a separate test that a clean tree stays silent.

21 tests, all green.

PR #3044, opened 2026-09-03
docs(uipath-platform): document feedback update metadata flags

dsuresh-ap:
Documents [UiPath/cli#3810](https://github.com/UiPath/cli/pull/3810) in the `uipath-platform` trace feedback reference.

## What changed

`skills/uipath-platform/references/traces/feedback.md`

- **New `update` flag table** with `--metadata <json>` and `--metadata-file <file>` (`-` for stdin), plus inline, file, and stdin examples. Notes that `create` has no `--metadata` — set it with `update` after creating.
- **Comment cap corrected 4000 → 1048576 chars** on both `create` and `update`. The 4000 figure was stale in the CLI help text as well; the PR corrected it there and this brings the docs in line.
- **Omitted fields are preserved.** `update` now reads the record before writing and carries over anything the caller did not pass. Previously an update that omitted `--category` wiped the categories. The non-atomic read-modify-write is called out, since the API offers no ETag or PATCH.
- **Clearing table.** `--comment ""` and `--metadata ""` clear those fields. Categories can no longer be cleared at all — `--category ""` stores a tag literally named `""`. This is a real capability loss from the PR, so it is documented rather than left for an agent to discover.
- **Metadata must be valid JSON** (any JSON value — object, array, string, number). Non-JSON is rejected server-side with `INVALID_FEEDBACK_METADATA`. The CLI passes the value through verbatim and does not pre-validate. Length is checked before JSON validity.
- **Mutual exclusion rules** extended with the `--metadata` / `--metadata-file` pair and the constraint that only one `-file` may read stdin, plus the contractual precedence: a flag used against its own `-file` twin is reported before the stdin clash, and both before any file is opened.

`skills/uipath-platform/references/uip-commands.md`

- `update` summary row now mentions metadata.

## Not changed

- `assets/uip-catalog-snapshot.json` — verbs only, and the PR adds flags, not verbs.
- Test tasks — none of the four `tests/tasks/uipath-platform/traces/` tasks exercise `feedback update`, comments, categories, or metadata.

## Verification

| Check | Result |
|---|---|
| `node scripts/check-skill-links.mjs` | 6584 links, all resolve |
| `python3 scripts/check-skill-verbs.py skills/uipath-platform/references/traces` | no stale verbs |
| `python3 scripts/check-skill-status.py` | 27 skills, manifest valid |
| `hooks/validate-skill-descriptions.sh` | clean |

Docs only — no frontmatter, skill status, or test changes.

### DIFF
--- a/skills/uipath-platform/references/traces/feedback.md
+++ b/skills/uipath-platform/references/traces/feedback.md
@@ -11,7 +11,7 @@ Use for agent output quality review and building evaluation datasets.
 | `get <id>` | Fetch one feedback record |
 | `list` | List feedback with filters |
 | `list detailed` | List feedback with span context, plus extra filters (max 200 items) |
-| `update <id>` | Change sentiment, comment, or categories |
+| `update <id>` | Change sentiment, comment, metadata, or categories |
 | `delete <id>` | Remove feedback |
 
 ## create
@@ -32,7 +32,7 @@ uip traces feedback create \
 | `--positive` / `--negative` | One required | Mutually exclusive |
 | `--folder-key` | Yes | |
 | `--span-id` | No | Defaults to root span of trace |
-| `--comment` | No | Max 4000 chars; mutually exclusive with `--comment-file` |
+| `--comment` | No | Max 1048576 chars; mutually exclusive with `--comment-file` |
 | `--comment-file` | No | Path to file; use `-` to read from stdin |
 | `--category` | No | Repeatable. Built-in values: `"Output"`, `"Agent Error"`, `"Agent Plan Execution"` |
 | `--agent-id` | No | Agent reference GUID |
@@ -92,7 +92,7 @@ Additional flags over `list`: `--since <duration>`, `--after <ISO>`, `--before <
 
 ## update
 
-`--category` tags are **replacement**, not additive — passing `--category` replaces all existing tags.
+Positional `<id>`, one of `--positive` / `--negative`, and `--folder-key` required.
 
 ```bash
 uip traces feedback update <feedback-id> \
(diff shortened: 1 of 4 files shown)


saksharthakkar:
check if you need to modify or add tests 
take a look at these for reference:
`traces_feedback_mutate_smoke.yaml`
`traces_feedback_comment_file_smoke.yaml`

dsuresh-ap:
Checked — added one new smoke task rather than modifying the two you pointed at.

**New:** `tests/tasks/uipath-platform/traces/traces_feedback_metadata_smoke.yaml` (+ `seed_feedback_metadata.sh`). Covers `update --metadata` / `--metadata-file`, which no existing traces task reaches. Two discriminating assertions:

- **Metadata must be valid JSON.** The prompt states the payload in prose, so an agent that passes the phrasing straight through as bare text would be rejected server-side with `INVALID_FEEDBACK_METADATA`.
- **Omitted fields are preserved.** `update` now reads before it writes, so re-passing the existing comment and tags to "protect" them is wrong. The criterion vetoes `--comment`/`--comment-file` and `--category` on that command.

Lookaheads use the segment idiom from `tests/README.md` so a chained second `uip` call can neither satisfy nor veto an assertion about the first. Patterns were exercised against 11 correct and incorrect command shapes on both the raw and newline-normalized haystacks; `check-cli-verbs.py` is clean.

**`traces_feedback_mutate_smoke.yaml` — left alone.** Its discriminating criterion (re-pass `Output` alongside `Agent Error`) still holds: `--category` remains replacement *when passed*. The PR only changed what happens when it is **absent** — previously that wiped the tags, now they carry over. The task always passes `--category`, so it never took the old path.

**`traces_feedback_comment_file_smoke.yaml` — left alone.** It exercises `create`, and `create`'s behaviour is unchanged by the CLI PR (only its `--comment` help text moved 4000 → 1048576, which is not graded). `create` has no `--metadata`.

One thing I deliberately did **not** add coverage for: the "`--comment-file -` and `--metadata-file -` cannot both read stdin" guard. Getting an agent to hit it naturally would need a contrived prompt, and it is already pinned CLI-side by a unit test, a command-level test, and scenario S22.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

saksharthakkar (approved):

alexenica (approved):

PR #3045, opened 2026-09-03
fix(maestro-flow): address #2756 review (pagination gate, rotation cleanup, node_modules prune)

tmatup:
## Why

rockymadden's review on #2756 (changes requested) raised four majors, five minors and five nits. This PR lands the code side on the campaign branch; the #2756 body rewrite and the per-item response follow once this merges.

## What changed

- **MAJOR 1** `paginated_reference_lookup.yaml`: the v1 paged-loop telemetry is advisory again at `min_count: 2`; the gating criterion requires "went past page 1" on a **successful** command for either route (`… list … nextPage=` or `registry prepare … --resolve channel:`). The corpus test uses the real `npx flow-sdk registry prepare` form (the `uip maestro flow registry prepare` sample was a phantom verb) and adds negative cases: page-1-only, `--resolve channel:` on another command, bare `nextPage=`.
- **MAJOR 3** Overwrite rotation: `_rotate_solution_id` appends the rotated-away id to `.rotated-solution-ids` beside the `.uipx`; `cleanup_solutions.py` deletes those ids alongside the current one. Unit tests cover the write and the read.
- **MAJOR 2** SKILL.md prose casing `bpmn:ServiceTask` matches the template. The template itself is unchanged (it is byte-equal to the CLI's own `packages/maestro-sdk/src/manifest/bpmn-spec.json`; see the #2756 response).
- **MINOR 5** `ceql_where` / `enhanced_enum` prompts say "Produce the final flow file" — neither `validate` nor `compile` is graded there.
- **MINOR 7** the 99 campaign headers drop the workspaces-only plan path; the three decisions stay inline.
- **MINOR 8** `wiki_pageviews` anchors `operate.json` to the discovered `.bpmn`'s parent.
- **MINOR 9** the three cwd-rooted `**` scans in `flow_check.py` go through `_rglob_pruned`, which never descends into `node_modules`; test with a symlinked fixture tree holding a `.flow` and a `project.uiproj`.
- **Nits**: `stage_shared.sh` compares realpaths; `validate_flow.py` dead `if not flows` removed (its test now asserts the real exit path); `case_check.py:66` wrapped; Dockerfile drops `FLOW_SDK_CONNECTORS_DIR` and the empty-dir fallback (nothing reads it: not the SDK, not the stage script, not the workflows); CODEOWNERS `a7868335f` reverted here so it can land as its own one-line PR.
- `run-coder-eval.yml` passes `--build-arg FLOW_SDK_VERSION="${{ vars.FLOW_SDK_VERSION || 'latest' }}"`, mirroring `CLI_VERSION`.
- Comment blocks over `_OVERWRITE_STUCK` / `_VARIABLES_UNREADABLE` trimmed to what a maintainer needs; the evidence tables live in UiPath/cli#3938 and UiPath/cli#3929.

## Verification

- `tests/tasks/uipath-maestro-flow`: 1147 passed (was 1145; two new tests, one rewritten)
- `tests/tasks/uipath-maestro-bpmn`: 10 passed; `tests/tasks/uipath-maestro-case`: 167 passed, 14 skipped; `tests/scripts`: 79 passed, 26 skipped
- `coder-eval plan` on the pagination task: valid; `check-cli-verbs.py`: clean (pattern skipped as dynamic, as before)
- `STAGE_SHARED_ROOT=tests/tasks/../../etc` is now refused by the guard

https://claude.ai/code/session_01UauCCMBpJ7ws1meQwBfYSS

### DIFF
--- a/.github/workflows/run-coder-eval.yml
+++ b/.github/workflows/run-coder-eval.yml
@@ -333,6 +333,7 @@ jobs:
             --build-arg CODER_EVAL_IMAGE="ghcr.io/uipath/coder-eval-agent:${CE_IMAGE_TAG}" \
             --secret id=npm_auth_token,env=NPM_AUTH_TOKEN \
             --build-arg CLI_VERSION="${{ inputs.cli_version }}" \
+            --build-arg FLOW_SDK_VERSION="${{ vars.FLOW_SDK_VERSION || 'latest' }}" \
             -t skills-image:latest \
             -f tests/docker/Dockerfile \
             .
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -19,7 +19,7 @@
 
 # preview skills
 /preview/uipath-maestro-flow/ @UiPath/maestro-cli-team
-/preview/uipath-maestro-case/ @UiPath/maestro-cli-team
+/preview/uipath-maestro-case/ @dmetzgar @tmatup
 /preview/uipath-maestro-bpmn/ @UiPath/maestro-cli-team
 
 # Activity references (bundled inside uipath-rpa skill)
--- a/skills/uipath-maestro-bpmn/SKILL.md
+++ b/skills/uipath-maestro-bpmn/SKILL.md
@@ -293,7 +293,7 @@ and honestly surfaced to the user as gaps when asked.
    `uipath:event` / `uipath:mapping` templates declare their type as a nested
    `<uipath:type value="<Type>" version="v1" />`. Some runtime-authored
    templates use the payload's `type` attribute instead; notably,
-   `Orchestrator.StartAgentJob` is a direct child of `bpmn:serviceTask` with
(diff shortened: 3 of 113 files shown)


tmatup:
Replying to the two manual-review items in the Claude review:

- **Pagination regex, live proof**: correct that "1147 passed" is pytest, not a coder-eval run. The pattern is covered by positive and negative unit cases; the live proof is the next nightly / same-ground run on the campaign branch, and the criterion is gating there.
- **`realpath -m` portability**: `stage_shared.sh` runs in the Linux eval image and CI only (GNU coreutils). Left as is; if it ever needs macOS, swap for `python3 -c 'import os,sys; print(os.path.abspath(sys.argv[1]))'`.

https://claude.ai/code/session_01UauCCMBpJ7ws1meQwBfYSS

PR #3048, opened 2026-09-03
feat(cve-findings): add skill for querying MER CVE findings

rahul-sharma-uipath:
Adds a skill over the read-only `/api/v1/cve-findings` API, which serves the same data as the MER Security Dashboard.

## Why

The API returns `200` with `{"total":0}` when a filter names the wrong dimension — `l3s=Sam Nuziale` when Sam is an L4, or `productNames=Experiences` when Experiences is a service under the Cloud Platform product. Both read as *"no findings"* rather than *"wrong parameter"*, and for a security question that silent zero is the worst possible answer.

So the workflow resolves every term against `/api/v1/cve-filter-options` before querying, and reports a zero only after re-verifying the dimension against a control query.

## Resolving against filter-options is necessary but not sufficient

The second half is easy to miss, and this PR now covers it. `filter-options` describes only what is **filterable**; the response carries many more columns that are output-only.

AGOV is the case that motivated the rule. `agov` matches no value in any filter dimension, so an agent that stops at `filter-options` concludes MER does not track it — while every row in fact carries `FixedinAGOV`, `AGOVLatest` and `FixedinSBI`. The skill now requires grepping the `columns` header, **by name and not only by value**, before telling a user something is untracked.

## Traps encoded

Found while exercising the API against live data:

| Trap | Consequence if missed |
|---|---|
| `Moderate` occurs only in FedRAMP (impact level, not a synonym for Medium) | Querying both double-counts |
| FedRAMP rows carry an empty Release Context | `releaseContexts=…` silently drops every FedRAMP row |
| One CVE appears once per affected image and component version | Row count reported as CVE count |
| `SLAStatus` is not a live overdue flag — most `OUT OF SLA` rows have a non-negative `SLADaysLeft`, and the status appears sticky once breached | Status alone overstates the fire; countdown alone contradicts the dashboard. The skill reports both |
| `FixedinAGOV` / `AGOVLatest` / `FixedinSBI` / `Cluster` are populated on FedRAMP rows only | An empty value elsewhere read as "not fixed" rather than "not applicable" |
| `ENGStatus`, `VULNStatus`, `ETA_MER` and friends are empty on every row seen | Reported as "nothing in progress" when they are simply unpopulated |
| `ENGJiraTicket` holds a bare key on some rows, a full browse URL on others | One ticket counts twice when grouping |
| `ComponentVersion` is installed-and-vulnerable; `FixedVersions` is the target | Presenting one as the other misdirects the fix |
| Findings sharing a container image are usually one rebuild | N tickets filed for one base-image refresh |
| The source table is replaced wholesale on each rebuild | Stale numbers presented as current |

## No hard-coded numbers

That last trap generalises into a rule the file applies to itself: **never hard-code a count, percentage or ratio.** Every such figure is stated as a shape plus the command that derives the current number, so the skill cannot go stale while still reading as authoritative. Two pre-existing hard-coded figures were removed as part of this (a distinct-CVE ratio in the e2e judge prompt, and a stale criteria count in the activation README).

## Tests

- Four task evals — one e2e, three smoke — covering resolve-before-query, free-text dimensions, and scope boundaries.
- **New:** `e2e_remediation_columns` pins the AGOV behaviour directly. It fails an agent that declares AGOV untracked on the strength of `filter-options` alone, and separately fails one that reads an empty non-FedRAMP column as "not fixed".
- **New:** activation coverage. The skill had no `activation.jsonl`, so it was absent from the activation confusion matrix entirely. It now has positives spanning owner, service, image, SLA, AGOV and SBI phrasings, registered in `activation.yaml`.

## Status

Preview. The API shipped in UiPath/MER#584 and is live on mer.uipath.com. Verified end to end against production — resolve, query and format all work, **every shell snippet in the skill was executed against live data**, and a credential-less call returns `401` JSON rather than a redirect.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -189,6 +189,8 @@
 /tests/tasks/uipath-mcp-servers/ @cristipufu @UiPath/team-coded-agents
 
 # Insights skill (job monitoring and analytics)
+/skills/uipath-cve-findings/ @rahul-sharma-uipath
+/tests/tasks/uipath-cve-findings/ @rahul-sharma-uipath
 /skills/uipath-insights/ @UiPath/insights
 /tests/tasks/uipath-insights/ @UiPath/insights
 /tests/tasks/activation/uipath-insights.jsonl @UiPath/insights
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -190,6 +190,8 @@
 /tests/tasks/uipath-mcp-servers/ @cristipufu @UiPath/team-coded-agents
 
 # Insights skill (job monitoring and analytics)
+/skills/uipath-cve-findings/ @rahul-sharma-uipath
+/tests/tasks/uipath-cve-findings/ @rahul-sharma-uipath
 /skills/uipath-insights/ @UiPath/insights
 /tests/tasks/uipath-insights/ @UiPath/insights
 /tests/tasks/activation/uipath-insights.jsonl @UiPath/insights
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ Every skill's maturity is tracked in [`assets/skill-status.json`](assets/skill-s
 | `uipath-automationhub` | In-development |
 | `uipath-coded-apps` | Preview |
 | `uipath-connector-builder` | In-development |
+| `uipath-cve-findings` | In-development |
 | `uipath-feedback` | Stable |
 | `uipath-functions` | Preview |
 | `uipath-governance` | In-development |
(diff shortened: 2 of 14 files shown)


rahul-sharma-uipath:
Closing — this landed in the wrong repo. Moved to the internal marketplace: **UiPath/skills-internal#612**.

`UiPath/skills` publishes to the public marketplace and to npm as `@uipath/skills`. This skill references `mer.uipath.com`, the MER Entra app-registration id used as its token audience, and real staff names as worked examples of dimension resolution — none of which should reach customers through the public catalogue.

Nothing is lost: the full skill, including the AGOV/SBI remediation guidance, moved across verbatim.

PR #2952, opened 2026-09-01
fix(uipath-rpa): Address failing daily eval runs

Mihaiii:
## Issue

[PILOT-7354](https://uipath.atlassian.net/browse/PILOT-7354)
[PILOT-7255](https://uipath.atlassian.net/browse/PILOT-7255)

## Summary

- Pin `--skip-build` on the T3 smoke run in all four `execution-maps-guide.md` gate rows (`:17`, `:34`, `:49`, `:60`) — the map chains `validate && build && run` but wrote `run` bare, leaving the flag documented only in `cli-reference.md § Smoke Test`
- Raise the graded build re-run to `timeout: 120` in `execution_map_greenfield.yaml` — it ran on the implicit 30s `run_command` default while the skill documents a 30–90s cold Helm restore

## Why

`skill-rpa-execution-map-greenfield` scored **1.000 on 2026-08-31** and **0.769 / `MAX_TURNS_EXHAUSTED` on 2026-09-01** with identical agent, model and turn count. The only behavioural difference: 08-31 chained `validate && build && run --skip-build`; 09-01 split the gate and ran `run` bare.

Bare `run` re-validates and rebuilds the whole project internally. Both days the smoke run was killed at the ~124s shell timeout (exit 124), but the 09-01 kill landed *inside a build* — the graded independent `uip rpa build` then blocked and blew its 30s cap (48s wall before the kill returned), zeroing the 3.0-weight primary criterion.

## Testing

- `npm run skills:validate` — OK (default: 27 skills / 1747 files; studioweb: 174 replacements)
- `/lint-task` on the task YAML — Verdict **Low** (one pre-existing prompt nit, unrelated to this change)
- `execution-maps-guide.md` carries no `skill-flavor` markers and has no `skill-flavors/*/uipath-rpa/` overrides — canonical-only edit

> **Not fixed here:** the map's turn budget assumes parallel tool calls, which the Codex harness cannot emit (one `tool_use` per assistant message; `Read`/`Write`/`Grep` all normalize to `shell`/`apply_patch`). `expected_turns: 6` / `max_turns: 10` stays unreachable on that agent — that is the actual `MAX_TURNS_EXHAUSTED` driver and needs a separate change.

[...]

PR #3055, opened 2026-09-04
fix(uipath-rpa): make Rule 22 a one-turn read so the greenfield map fits its 10-turn gate

tmatup:
## Why

`Smoke RPA Skill Tests (Windows)` is red on #2756 because `skill-rpa-execution-map-greenfield` scored 0.62 against the 0.7 gate. That PR does not touch `skills/uipath-rpa`. The task has ended in `MAX_TURNS_EXHAUSTED` in every CI run since it became a smoke gate (#2829, 08-27): 0.73, 0.77, 0.77, 0.77, 0.62. The gate scores weighted_score, so a turn-budget task that never finishes "passed" four times.

## Root cause

#2698 (08-26) rewrote Rule 22 from a plain read into a two-step read: Grep `^## Catalogs` for a line number, then Read `xaml-basics-and-rules.md` with `limit`. It added the same grep-then-read shape for `common-pitfalls.md`. Each pair needs the grep's output before the Read, so it costs two sequential assistant turns. The greenfield journey map still budgeted the basics file as one parallel T1 Read. In the failing run the agent spent turns 1-10 on: Skill, read journey map, `ls`, cards, grep marker, read basics, grep pitfalls headings, read pitfalls section, `init`, read `project.json`/`Main.xaml`. Zero XAML written.

Enabling defect: one file held a mandatory prefix and a grep-gated catalog, so any turn-cheap read contract for it had to be a dependent grep->Read pair.

## Fix

- Split the catalog tail into `references/xaml/xaml-editing-catalog.md` (Common Editing Operations, XAML Reference Examples). `xaml-basics-and-rules.md` is now mandatory content only: a plain full Read, issuable in T1 with nothing to look up first.
- `common-pitfalls.md` keeps its per-section contract. The heading Grep has no prerequisite and moves to T1; the section Reads become a new T2 "Vet" turn. One turn for the gate instead of two per file.
- Journey maps re-derived: greenfield T1..T5 (budget stays <=5), brownfield T1..T5 (<=4 -> <=5, stated honestly). Coded journeys unchanged.
- Rule 22, the basics file's read contract, the SKILL.md budget line and the single anchor into the moved section (`testing-guide.md`) updated.

Turn arithmetic for the gate: Skill (1) + read the journey map (1) + T1..T5 (5) = 7 <= 10 by design. The failing trace also spent a standalone turn on an `ls` emptiness check before T1; the greenfield map now folds that into the T1 Bash chain. If an agent still emits it alone, the path is 8.

Not changed: `max_turns: 10`, the 0.7 gate. The task exists to catch this.

## Verification

- `node scripts/check-skill-links.mjs`: 6610 links resolve.
- `node scripts/compose-skill-flavor.mjs validate`: OK.
- Not reproducible locally (Windows + Helm). The Windows smoke on this PR is the check. Bar: `skill-rpa-execution-map-greenfield status=SUCCESS`, not `score >= 0.7`.
- Run 33838560457 (after the read-contract commits): T1 and T2 landed as designed (init + both cards + basics + pitfalls heading grep in one turn; pitfalls sections + project.json in the next), Main.xaml written, validate and build clean in-loop. Still `MAX_TURNS_EXHAUSTED` at 0.77 because the agent split the author and gate phases into seven single-command turns. That is what the third commit fixes (Phase 3 code block was two commands; now one `&&` chain; CREATE path skips the scaffold Read).
- Run 33842310232 (current tip): `skill-rpa-execution-map-greenfield status=SUCCESS score=1.00`. First SUCCESS for this task in CI. Trace: Skill / ls+map / init+reads / vet / write data+XAML / validate&&build = 6 turns to a clean gate, then the optional `run` and its output check. All four Windows smoke rows SUCCESS at 1.00.

## Merge with `main` (2026-09-04)

`main` picked up #2952 (`c9c6993`), which conflicted in `execution-maps-guide.md`: it pinned `--skip-build` on the four T3 gate rows while this branch renumbered the XAML journeys T1..T5 and made Rule 22 a plain full Read. Resolved in merge commit `122d060` keeping both — this branch's T1..T5 rows, with `run --skip-build` in each gate row. Two edits the textual merge could not make:

- #2952's new "T1 is ONE message" bullet auto-merged but described the read contract this PR removes. Dropped "(Rule 22 two-step)" and retargeted "open no other reference before T2" to T3, since T2 is now the Vet turn that legitimately reads the pitfalls sections and the editing catalog.
- `xaml/xaml-basics-and-rules.md` § validation loop still closed the gate chain with bare `uip rpa run` — the same cross-file drift #2952 fixed in the map. Aligned to `run --skip-build`; the chain's `build` immediately precedes it. #2952's `cli-reference.md` rule (bare `run` after edits) governs ad-hoc runs and is unaffected.

Re-verified after the merge: `node scripts/check-skill-links.mjs` — 6618 links resolve; `npm run skills:validate` — OK (27 skills / 1760 files; studioweb 182 replacements).

**This PR's Windows smoke no longer exercises the task this PR fixes.** #2952's later commits also dropped the `windows` tag from `execution_map_greenfield.yaml`. That removes the task from *both* PR gates: `smoke-rpa-skills.yml` selects by the `windows` tag, and `smoke-skills.yml` excludes the whole `tests/tasks/uipath-rpa/` tree because seven sibling tasks still carry it. The task still runs in the external daily eval, which is where #2952 was diagnosing it. The fix's evidence stands on runs 33842310232 and 33846250313, taken while the tag was present.

Whether to re-add `windows` to the yaml — putting the now-passing task back on the gate that motivated this PR — is a policy call, not part of the conflict resolution, so it is not done here. #3059 already covers the related gate policy (accepting `MAX_TURNS_EXHAUSTED` rows above 0.7).

## Follow-ups (filed, not in this PR)

- #3058: Rule 7 (`uia-starter-guide.md`) has the same two-step read shape against a parallel T0 Read; no CI gate exercises it.
- #3059: the Windows gate accepts `MAX_TURNS_EXHAUSTED` rows above 0.7 (this hid the defect for eight days; requiring `final_status=SUCCESS` is a policy call). Same issue covers the five windows-tagged, non-smoke tasks that log `Results: 0/0` + `exit code 1` noise on every infra-triggered run.

A second Windows smoke run on the tip was started by workflow_dispatch to widen the sample beyond n=1: run 33846250313.

Unblocks #2756.

https://claude.ai/code/session_01A3ccsNLQGNnwaR16Sz9niZ


tmatup:
Re the Claude review: both anchor findings (`xaml-editing-catalog.md:249` → Container Activity Bodies, `xaml-basics-and-rules.md:485` → Example 1) were fixed in b2f5ecd82, pushed before the review job finished reading the first commit. `scripts/check-skill-links.mjs` passes at the tip.

On the brownfield budget note: no Windows CI task gates a brownfield XAML edit at a turn cap today (`legacy/edit_existing_xaml.yaml` is skipped in the workflow pending a Bedrock model). The ≤4 → ≤5 restatement counts the pitfalls vet turn Rule 22 already required; it does not add work.

rockymadden (commented):
## Review

Verified against a local worktree of the head (`122d060`). Checked what the description asserts rather than taking it on trust.

### Verified clean

| Check | Result |
|---|---|
| Moved catalog content | Byte-identical to `main` except the read-contract header and one anchor repoint |
| The two anchors the bot flagged | Both fixed in `b2f5ecd8`; both targets exist |
| `testing-guide.md` -> `#setting-default-values-for-arguments` | Resolves (`xaml-editing-catalog.md:28`) |
| Every other repo reference into `xaml-basics-and-rules.md` | No new breakage |
| Flavor overrides for these paths | None exist |
| `--skip-build` vs `cli-reference.md:353` | Consistent; the chain's `build` immediately precedes |
| Merge `122d060` behavior delta | Only `run` -> `run --skip-build` plus the re-added T1 bullet. Turn structure is unchanged from the evidence runs, so 33846250313 still stands as evidence |

The refactor is correct and the root-cause analysis holds. Comments below, most severe first. Four more are inline.

---

### :orange_circle: Major: the task this PR fixes runs on no PR gate

The description flags this. Raising it as a decision rather than a deferral, because the consequence is that this merges with no CI proof.

`tests/tasks/uipath-rpa/execution_map/execution_map_greenfield.yaml:10` is `tags: [uipath-rpa, smoke, mode:build, lifecycle:generate]`. No `windows`. That removes it from both PR workflows:

- `smoke-rpa-skills.yml:46` selects on `grep -rEl '^tags:.*\bwindows\b'`, so it is not selected.
- `smoke-skills.yml:125` `is_excluded_path()` excludes the whole `tests/tasks/uipath-rpa/` prefix, because `WINDOWS_SKILLS` is derived from the seven siblings that still carry the tag. So it is excluded on Linux too.

Confirmed on this PR's own green check (run `33898940141`): only `skill-rpa-coded-test-case`, `skill-rpa-template-aware-create`, and `skill-rpa-uia-google-search` ran. `Results: 3/3 task(s) >= 0.7`. The fixed task was not among them.

The green "Run RPA skill smoke tests (Windows)" is real but carries no signal for this change, and the regression can come back silently.

**Proposal:** re-add `windows` to `execution_map_greenfield.yaml` here. One line, and the task now scores 1.00.

#3059 covers the *pass criterion* (`MAX_TURNS_EXHAUSTED` above 0.7 counting as green). That is a different question from whether the task runs at all. Treating the tag as the same policy call conflates the two.

---

### :large_blue_circle: Nit: two pre-existing broken anchors into this file

Found while checking the split. Neither resolves on `main` either, so not yours:

- `#common-pitfalls--quick-reference`, from `legacy/legacy-mode-guide.md` (x4), `legacy/error-handling-guide.md` (x2), `legacy/activity-docs/_PATTERNS.md`
- `#scope-activities-require-activityaction-body-critical-for-xaml-generation`, from `legacy/activity-docs/Excel.md`

Natural fodder for the checker follow-up in the inline comment on `xaml-editing-catalog.md`.

---

### AI slop check

Content is clean. I diffed the moved sections: no drift, no invented text, no filler. The description runs long for a five-file doc move and the "Merge with `main`" section reads like a work log, but every claim in it is specific and falsifiable. The two anchor misses in the first commit are the classic split-refactor failure, caught and fixed within two hours.

---

### Recommendation

**Comment.** Approve once the `windows` tag question is settled.

The blocker is not the diff, it is that merging leaves the fixed behavior untested by any PR gate. Add the tag and this is an approve. The two inline minors are cheap enough to fold in now and will save a re-open later.

skills/uipath-rpa/references/execution-maps-guide.md
```diff
@@ -25,30 +25,32 @@ A clean `validate` + `build` does NOT prove runtime behavior. Known silent failu
...
+| **T1 — Scaffold + context** | ONE `Bash` chain: `uip rpa init --name "<NAME>" --location "<PARENT_DIR>" --template-id BlankTemplate --expression-language <VisualBasic\|CSharp> --target-framework <Windows\|Portable> --output json` (Rule 2a — both flags explicit) `&&` `uip rpa analyzer-rules list --project-dir "<PROJECT_DIR>" --output json` `&&` one `uip rpa packages versions --package-id <PackageId> --include-prerelease --project-dir "<PROJECT_DIR>" --output json` per request-known package ∥ parallel `Read` — ALL THREE, never a subset (skipping the pattern card re-opens per-activity discovery): [common-activity-card.md](common-activity-card.md), [common-pattern-card.md](common-pattern-card.md), [xaml/xaml-basics-and-rules.md](xaml/xaml-basics-and-rules.md) (Rule 22 — plain full Read, nothing to look up first) ∥ `Grep` `^## ` on [xaml/common-pitfalls.md](xaml/common-pitfalls.md) for its heading list (Rule 22 gate, step 1 — needs no prior output, so it rides here) ∥ memory recall (harness has memory) ∥ Rule 21 `activities find` fan-out for off-card activities |
```
rockymadden, line 32: :yellow_circle: **Minor: the catalog reintroduces the exact grep -> Read pair this PR removes.**

`xaml-editing-catalog.md:5` sets its read contract as "Grep `^###` on this file to list the entries with line numbers, then Read ONLY the entries". That is a dependent pair, the same shape as old Rule 22.

This T1 row schedules only the *pitfalls* heading Grep. T2 then says "per-entry `Read`s from `xaml-editing-catalog.md`" with no line numbers in hand. An agent following this literally has to Grep in T2 and Read in T3, costing one extra turn exactly when an editing operation is needed.

**Fix:** add the catalog's `Grep ^### ` to this row alongside the pitfalls Grep. It needs no prior output either, so it rides for free, and the symmetry is the argument the PR already makes.

skills/uipath-rpa/references/execution-maps-guide.md
```diff
@@ -25,30 +25,32 @@ A clean `validate` + `build` does NOT prove runtime behavior. Known silent failu
...
-| **T2 — Edit** | Batched `Edit`s (anchor each on its own target block — same-file Edits serialize; overlapping anchors fail) ∥ `packages install` `Bash` if new dependencies |
-| **T3 — Gate** | ONE `Bash`: per-file `validate` (relative `--file-path`) `&&` `build` `&&` optional `run --skip-build` per § Gate ≠ runtime proof |
-| **T4 — Report** | Output check (if T3 ran) + § Completion Output + memory save |
+| **T1 — Context** | SKILL.md § Precondition context check — the skip gate first, then the discovery subagent, which writes `project-context.md` + `AGENTS.md` itself ([environment-setup.md § Project Context Discovery](environment-setup.md)) ∥ `Read` `project.json` + target `.xaml` + cards + [xaml/xaml-basics-and-rules.md](xaml/xaml-basics-and-rules.md) (Rule 22, plain full Read) ∥ `Grep` `^## ` on [xaml/common-pitfalls.md](xaml/common-pitfalls.md) for its heading list ∥ ONE `Bash`: `analyzer-rules list --project-dir "<PROJECT_DIR>" --output json` ∥ memory recall ∥ off-card `activities find` fan-out |
```
rockymadden, line 49: Same gap as greenfield T1: this row schedules the pitfalls heading `Grep` but not the `xaml-editing-catalog.md` one, while T2 below reads catalog entries that need line numbers first.

Add `Grep ^### ` on the catalog here too.

skills/uipath-rpa/references/execution-maps-guide.md
```diff
@@ -1,6 +1,6 @@
 # Execution Maps — Turn-Budgeted Build Journeys
 
-One dense file, read once per build. Fixes which tool calls go in which assistant turn. Budgets (happy path, incl. final report): **greenfield ≤5 turns, brownfield ≤4**. One repair cycle adds ≤2. Each T-row = ONE assistant message: N tool calls in one message cost 1 turn, one call per message costs N.
+One dense file, read once per build. Fixes which tool calls go in which assistant turn. Budgets (happy path, incl. final report): **greenfield ≤5 turns, brownfield ≤5**. One repair cycle adds ≤2. Each T-row = ONE assistant message: N tool calls in one message cost 1 turn, one call per message costs N.
```
rockymadden, line 3: :yellow_circle: **Minor: repair headroom is now roughly zero.**

Your arithmetic: Skill (1) + journey-map Read (1) + T1..T5 (5) = 7 of 10, with the caveat that a standalone `ls` turn makes it 8. The repair cycle on this same line costs +2.

So one repair puts a well-behaved run at 9, and a slightly chatty one at exactly 10, which is `max_turns`.

The greenfield journey went from 4 rows to 5 while the label stayed at 5. Net -1 turn against the old contract, correct, but the margin against `max_turns: 10` is now a single repair cycle. Once the eval is wired back up (see the review body), that is where flake returns.

Worth stating explicitly here, or bumping `max_turns` deliberately, rather than discovering it in a red run.

skills/uipath-rpa/references/xaml/xaml-editing-catalog.md
```diff
@@ -0,0 +1,337 @@
+# XAML Editing Catalog
+
+Per-operation and per-activity catalogs for XAML authoring: editing operations (arguments, variables, imports, assembly references, expressions, resource types) and complete reference workflows. Split out of [xaml-basics-and-rules.md](xaml-basics-and-rules.md) so that file is a plain full read (Rule 22) with no marker to locate first.
+
+**Read contract — per entry, never end-to-end:** Grep `^###` on this file to list the entries with line numbers, then Read ONLY the entries matching the operation or activity at hand. Unsure whether an entry applies → read it. Reading this file top to bottom is the mistake the split removed.
```
rockymadden, line 5: :large_blue_circle: **Nit: the cited verification does not cover the split's main risk.**

`check-skill-links.mjs` resolves file existence only. `isExternal()` at line 46 short-circuits `#`-prefixed targets, and nothing in the script splits a fragment off a `file.md#anchor` target.

So "6618 links resolve" is true, and says nothing about whether the moved sections' anchors landed. That is precisely what a section split breaks, and precisely what the bot caught in the first commit.

I checked all of them by hand and they are correct, so no action needed on this PR. But the claim in the description overstates what the script proves. Adding fragment validation to the checker would have caught both misses pre-review, and would also surface the two pre-existing broken anchors noted in the review body.

skills/uipath-rpa/references/xaml/xaml-basics-and-rules.md
```diff
@@ -223,20 +224,22 @@ For end-to-end authoring of `ConnectorActivity` XAML (connection + type ID + Con
...
 
 **For EDIT requests:** always `Read` current content before editing; use `Edit` with exact, unique `old_string` matches.
 
 ### Phase 3: Validate & Fix Loop
 
 **MUST** repeat until 0-error state from **both** `validate` and `build`, or max 5 fix attempts per loop. After 5 attempts, stop and present remaining errors to the user. The canonical two-phase loop (per-file `validate` → project-level `build`), what each phase covers, and the smoke-test procedure: [../cli-reference.md § Validation Iteration Loop](../cli-reference.md#validation-iteration-loop) — read it before your first fix iteration.
 
+One `Bash` call for the whole gate — chain the phases with `&&` so a `validate` failure stops before `build` and the pass path costs one turn, never one turn per command (the journey map's gate row in [execution-maps-guide.md](../execution-maps-guide.md)):
```
rockymadden, line 235: :large_blue_circle: **Nit: the code block omits the tail its own prose mandates.**

The block below shows `validate && build`, then the prose right after it says to end with `run --skip-build`. The map's T4 row has all three in one chain.

Given this PR's thesis is "one `Bash` call for the whole gate", an agent copying the block gets two thirds of it. Adding the conditional third link (commented, since it only applies to observable-output workflows) would be more faithful.

danmunteanu1:
## Independent A/B bench: this branch vs its merge-base, 30 headless runs

I benchmarked `122d060` (branch tip) against `cec566a` (its merge-base on main) on all five journeys in `execution-maps-guide.md`, 3 replicates per arm, interleaved, strictly sequential on one desktop. Headless `claude -p`, `--model opus --effort medium`, `--plugin-dir` pointed at a detached worktree per arm, fresh run folder per run, prompts in skill-benchmark mode (app knowledge only, no CLI flags or file names). uip 1.200.1-preview.129, UIA `26.10.2026231-dev.1215`.

Journeys: `uia-calc` (greenfield UIA: init, install UIA, capture Calculator, 5+5, log result), `gf-xaml` (greenfield XAML, 5+5 to `result.txt`), `bf-xaml` (brownfield edit of a fixture: add 5*5 and `results.txt`), plus `gf-coded` / `bf-coded` twins. The coded journeys are not touched by this PR and serve as the noise-floor control.

**Outcome: 30/30 runs succeeded** (validate + build + run clean, correct output on disk, 0 compactions). No regression on any journey.

### Cost and turns (mean of 3, Opus 5 list prices, usage summed once per API request)

| Journey | API requests main → branch | Billed tokens | USD per run | Wall | Peak context |
|---|---|---|---|---|---|
| uia-calc | 51.7 → 37.3 (−28%) | 7.89M → 5.88M (−25%) | $6.56 → $5.35 (**−19%**, ranges do not overlap) | 8.9 → 7.0 min (−21%) | 206.6k → 205.4k (flat) |
| gf-xaml | 17.0 → 16.0 (−6%) | 1.88M → 1.59M (−15%) | $2.39 → $2.06 (**−14%**, ranges do not overlap, sd $0.03–0.04) | 3.0 → 2.8 min | 134.8k → 118.2k (−12%) |
| bf-xaml | 18.0 → 15.3 (−15%) | 1.58M → 1.41M (−10%) | $1.86 → $1.81 (−3%, overlap) | 2.4 → 2.5 min | 106.3k → 108.6k (flat) |
| gf-coded (control) | 23.0 → 17.7 | 2.17M → 1.52M | $2.21 → $1.78 (−19%, overlap, one main outlier) | 3.3 → 2.4 min | 106.0k → 101.1k |
| bf-coded (control) | 34.3 → 20.3 | 2.99M → 1.77M | $2.64 → $2.03 (−23%, overlap, one main outlier) | 4.3 → 3.3 min | 99.0k → 102.8k |
| All 15 runs per arm | 144 → 107 | 16.5M → 12.2M | $47.03 → $39.10 (−17%) | | |

Read the table as: peak context is flat (the mandatory read path costs the same content), while API requests and therefore billed tokens drop because turns are batched. The two journeys this PR targets are the credible wins; the controls also trend cheaper but each is carried by a single expensive main run (run-gate failure loops, see below) and overlaps, so they mark the noise floor rather than add evidence.

### Per-mechanism verdicts from the transcripts

1. **Plain full Read of `xaml-basics-and-rules.md` instead of Grep `^## Catalogs` → Read `limit`.** Confirmed and turn-saving, token-neutral. Main honored `limit=537` in 9/9 XAML runs (12,643 tok); the branch read the file in full 9/9 (12,839 tok). On main the Grep landed in one assistant turn and the Read in the next in every run; on the branch the Read rides with the card reads in the same turn. One round-trip per XAML run. `xaml-editing-catalog.md` was never read on either arm (nor were the catalog sections on main), so the split is a robustness gain, not a content gain.
2. **Pitfalls heading Grep in T1, section Reads in the new T2 Vet turn.** Matches what agents already did on main (grep in one turn, reads in a later one); the old map row claimed both as T1 reads, which was impossible. The branch text now describes real behavior.
3. **`validate && build` chained in one Bash (xaml-basics § Validate).** Main split validate/build/run into separate turns in 3 of 9 XAML runs; the branch chained in 9/9.
4. **CREATE path skipping the scaffold `Main.xaml` Read.** Weakly followed: the branch still read it in 2 of 3 greenfield XAML runs (~1k tok each).
5. **UIA turn saving.** Consistent in all three pairs (capture-phase turns 16–22 on main vs 10–14 on the branch) but the capture phase is governed by the UIA package docs, identical on both arms, so no line of this diff explains it directly. Quote it as measured, not as a designed effect, until there is more data.

### Caveats for the turn budget story

The ≤5-turn budgets are not met end to end on either arm in this bench: greenfield XAML takes 15–18 assistant messages, brownfield 14–20, UIA 35–58. Turns to a clean gate are close to the map (5–7 for `gf-xaml` on the branch, matching the CI trace in the description); the rest is spent after the gate: 4–12 wrap-up turns per run (project-context.md, AGENTS.md, memory saves, output verification) and 1–3 retries of `uip rpa run` whose agent-invented `--output-filter` JMESPath expression failed to evaluate (26 of 30 runs, both arms). Neither is in scope here, but both are larger than what remains in the read path if someone chases the next turn reduction.

Two environment notes from the runs, unrelated to this PR: coded `uip rpa run` after edits failed twice on main with `'WorkflowRunnerService' could not be found` until a rebuild (the known `.local/.codedworkflows` cold-start issue; it produced the coded outliers above), and headless `-p` on Claude Code 2.1.260 did spawn the discovery subagent in 3 brownfield runs, so the old "AgentTool is suppressed in -p" assumption no longer holds.

Full per-run tables, transcripts, and the per-file read attribution are available on request.

danmunteanu1 (approved):

AlvinStanescu:
@rockymadden apologies, missed your comments here and merged based on Dan's last comment. @tmatup please address those in a follow-up PR.

PR #3057, opened 2026-09-04
refactor(planner): replace audit_sdd.py with a template conformance checklist; maestro-case receipt reads the handoff marker

abhiram-vad:
## Why

`scripts/case/audit_sdd.py` gated the Case Design Lane's `sdd.md` before the `Status: ready` flip. Two problems, both raised in review on #2816 and Autopilot #5242:

1. **Runtime.** Python is not a runtime these skills can assume (Windows Store alias stubs; no Python on any Delegate host). A Node port only half-fixes it: Delegate's restricted sessions whitelist the bundled Node and `@uipath/cli` binaries but keep the plugin folder outside the readable roots, so a `.mjs` in `scripts/` exists and cannot run. The `python3 → python → py → verify by hand` ladder also caused a documented interpreter hunt in the finalize evals (14 tool calls, 3600 s, no `sdd.md` written).
2. **Localization.** The validator keys on 36 English literals (headings, bold labels, `Yes`/`No`). Measured on a clean published SDD (`cm-golden-expense-0825`): translating headings and labels to German yields 10 findings, all false, and every per-stage check goes vacuous. With 12 seeded defects, recall drops from 12/12 (English, 1 false positive) to 3/12 (German, 10 false positives). German customers write German SDDs.

## What

- **Delete `audit_sdd.py`.** Its 61 finding sites are transcribed 1:1 into a **33-item conformance checklist** in the case template's `§ Validation` footer (9 families, same order as the old family list). The lane answers every item PASS/FAIL in the transcript with a verbatim quote from the on-disk file, judged by block role and position rather than wording, repairs FAILs with targeted Edits, re-answers only the failed items, max 3 rounds. Items 30–33 are the `--draft` parity checks (inventory, verbatim `=js:`, executable thresholds, draft still on disk).
- **Lane guide**: Terminal step, bounded no-build mode, Confirm, and the finalize fast path step 10 now name the checklist; the interpreter ladder and the "verify by hand" escape hatch are gone.
- **Layers guide / sdd-generation-guide**: the three `<!-- parsed at runtime by audit_sdd.py -->` table guards become "named by the checklist" guards; gate references retargeted.
- **`uipath-maestro-case` rule 2**: the SDD receipt reads the first 40 lines for the `<!-- planner-handoff:v1 -->` marker and a `Template validation: passed` header instead of a Grep for four English headings. Same purpose (reject freeform/summary SDDs before a long read), no English wording dependency.
- **Evals**: the four finalize tasks gain a `run_command` criterion asserting the header carries `Template validation: passed`, so a ready flip without the gate is graded.
- **CODEOWNERS**: the now-empty `/skills/uipath-planner/scripts/case/` line is dropped.

## Evidence for the checklist as the gate

Four fresh-context judge runs on the same seeded documents, each given only the template, the layers guide's Task types / Lifecycle gates / Naming rules sections, and the SDD, with no knowledge of the seed list:

| Gate | Document | Recall | False positives |
|---|---|---:|---:|
| `audit_sdd.py` | English | 12/12 | 1 (`return-to-origin` matched inside a Design Rationale sentence) |
| `audit_sdd.py` | German headings/labels | 3/12 | 10 |
| checklist, default model ×2 | English | 12/12 | 0 (one run added a true finding the script has no check for) |
| checklist, Sonnet | English | 11/12 | 1 |
| checklist, default model | German headings/labels | 12/12 | 0 |

Caveats: one document, twelve seeds, four runs; the seeds are the classes the script was built for. What this does not settle is whether the lane *runs* the gate — the same failure @roberts-cliff measured for the script (0 of 6 failed eval runs invoked it). The header criterion above is the enforcement a portable skill has; the finalize evals still grade the file with `sdd_check.py`.

## Verification

- `npm run skills:validate` — OK (27 skills, both flavors)
- `npm run skills:check-links` — 6597 links resolve
- `npm run skills:build` — OK
- `pytest tests/tasks/uipath-planner/_shared tests/tasks/uipath-planner/case` — 77 passed
- `grep -rn audit_sdd` across skills, tests, docs — zero hits
- The four edited task YAMLs parse; the new criterion is present in each
- Not run: `coder-eval` on the finalize tasks (needs the eval environment). The header criterion asserts behaviour the lane already has today (`Template validation: passed` at the ready flip), so it is low risk, but the checklist's in-lane recall is unmeasured until the next eval cycle.

## Deliberately not changed

- `uipath-maestro-case/scripts/audit_caseplan.py` and the `cat … | python3 -c` registry-cache filter in `registry-discovery.md`. I tried to swap the filter for `uip maestro case registry search`: on `uip 1.198.0-preview.102` `registry list` counts agent/process/api/processOrchestration/caseManagement/typecache-* only and `search … --type action-apps` returns 0 while `~/.uip/case-resources/action-apps-index.json` holds 23 apps. Until the CLI indexes action apps, the Python filter stays. Separate PR once that lands.
- The SDD heading contract. Localized *content* already passes; localized *headings* still need `uipath-maestro-case`'s `audit_caseplan.py` (`Stage`/`Task` word regexes) to key on the existing `(`Stage_xxx`)` id suffixes instead of words, plus task/section ids in the template. That is a contract change touching the eval twins and belongs in its own PR.
- The `Template validation` header value stays `passed`; Delegate and the eval fixtures read it.

## Relationship to open PRs

Supersedes #2812 and #2816 (both Node ports of the same script). Answers @AlvinStanescu's review on #2816 (validation agent over regex; German/Japanese SDDs) with the measurements above, and @RaduAna-Maria's "not needed, can be removed later" on Autopilot #5242 by removing it now with a checklist in its place rather than nothing.


skills/uipath-planner/references/case/case-design-lane-guide.md
```diff
@@ -193,14 +193,14 @@ Structured gap escalations — a field could not be fully resolved but the build
...
-2. **Gate the written file:** run the §Template conformance gate (`audit_sdd.py`). Repair findings with targeted Edits and re-run — max 3 rounds, then stop and present what remains. One structural Read is allowed here.
+2. **Gate the written file:** run the §Template conformance gate — the template's § Validation checklist, items 1–29 (30–33 as well when finalizing a draft). Answer every item in the transcript with its quote, repair each FAIL with a targeted Edit, then re-answer only the failed items — max 3 rounds, then stop and present what remains. One structural Read is allowed here.
```
AlvinStanescu, line 203: Recommend that the validation checklist runs in a subagent, to save context. Allow subagents strictly for this.

AlvinStanescu (approved):

skills/uipath-planner/references/case/case-design-layers-guide.md
```diff
@@ -144,6 +144,25 @@ The case, each stage, and each task move through gates driven by **rules** in di
 
 **`wait-for-user` ↔ `user-selected-stage` pairing.** Validate enforces the pair both ways: a `wait-for-user` exit with no `user-selected-stage` entry anywhere fails with `Stage rule '<name>' has no possible stage options.`; a `user-selected-stage` entry with no `wait-for-user` exit fails with `Stage entry rule '<name>' will never be met.`. `user-selected-stage` is picker exposure — a user choosing the next stage — never deterministic routing. Deterministic rejection, approval, send-back, and SLA routing use decision facts plus guarded entries instead.
 
+**Repairing a `user-selected-stage` lane — diagnose first; the two repairs are opposites.**
```
RaduAna-Maria, line 147: I know the schedule is tight for the Cartographer release, but after that please take some time to reiterate over this. I still consider that this much depth about functionality of maestro is not necessary at the planning stage, especially how to repair something inside the project.
cc @AlvinStanescu

RaduAna-Maria (approved):

roberts-cliff (approved):

skills/uipath-planner/references/case/case-design-layers-guide.md
abhiram-vad: I'll reiterate over this again with Cliff once I have bandwidth. Cliff is writing SDD parsing and validation at cli layer, and I feel we can move some knowledge there in terms of allowed rules or schema. I'll do it in a follow up PR.

charlesliu9 (approved):

PR #3065, opened 2026-09-04
feat(skill-flavors): the studioweb flavor never tells the agent to run `uip solution init`

astridstaicu:
## Motivation

Studio Web works on one open solution, already scaffolded as the workspace root (`/solution`), and its host intercepts `uip solution init` with "Studio Web works on one open solution and cannot create another". Eight skills still told the agent to run it (`uipath-agents`, `uipath-maestro-case`, `uipath-maestro-bpmn`, `uipath-planner`, `uipath-platform`, `uipath-rpa`, `uipath-solution`, plus the planner's case lane guide), so every Studio Web build that followed them started from a failed command.

The coder_eval studio-web nightly ([13250238](https://dev.azure.com/uipath/coder_eval/_build/results?buildId=13250238)) lost 98 of 446 tasks to this: agents fell back to hand-made layouts and the case and flow validate criteria then scored otherwise valid work at 0. `uipath-maestro-flow`, `uipath-api-workflow` and `uipath-human-in-the-loop` were already clean because their existing studioweb overrides cover the passages.

## What changed

- Every canonical passage that mentions `uip solution init` is wrapped in a marker block (canonical text unchanged) with a sparse replacement under `skill-flavors/studioweb/`. The replacement says the same thing everywhere: the open solution is the only solution, projects are created inside it with `uip <family> init <ProjectName>` (the families the Studio Web host creates: flow, case, agent, api-workflow, bpmn, rpa), there is no `.uipx` on disk and no registration step. Table rows and the coded-agent "package and upload" delivery option that only make sense locally are emptied for the flavor.
- **`uipath-maestro-case`** needed more than a substitution. The canonical skill forbids `uip maestro case init` because it can fork a second solution locally; in Studio Web that hazard does not exist and `case init` is the host's project-creation route, so the studioweb overrides point Step 6.0 at `uip maestro case init "<ProjectName>"` from `/solution`, keep T01's caseplan write, and drop the `.uipx` existence checks and `uip solution projects add`. Every canonical "never run `case init`" passage is overridden so the flavor does not contradict itself.
- New `tests/scripts/studioweb-flavor-contract.test.mjs`: builds the studioweb tree with the composer and fails on any `uip solution init`; a second test checks the default tree still contains it so the guard cannot go vacuous. Wired into `skills:test` and the `validate-skill-flavors.yml` path filter.

## Please review specifically

- **Case skill owners:** the Studio Web case flow (`uip maestro case init` seeds the scaffold, T01 writes `caseplan.json` and whatever Studio Web did not seed) is inferred from the host's interceptor behaviour, not yet validated end to end in Studio Web. The studio-web eval will exercise it once the harness mounts this flavor.
- `uipath-maestro-bpmn` operate/run.md: the Studio Web replacement assumes `uip maestro bpmn debug` runs from `/solution`.
- The `evals.json` under `uipath-maestro-case` still says "solution init + planning" in evaluator prose; the composer only overrides Markdown, and the text carries no `uip` prefix, so it is out of scope here.

## Verification

- `npm run skills:validate`: default 0 replacements, studioweb 226 replacements, no findings.
- `npm run skills:build` then grep: `uip solution init` appears in 0 files of `build/skills/studioweb` and 33 files of `build/skills/default`.
- `npm run skills:check-links`: all 6606 links resolve.
- `npm run skills:test`: 53 pass, 1 fail. The failure (`pack builds complete, marker-free default and custom npm packages`) fails identically on an untouched `origin/main` checkout in my environment (its `npm publish --dry-run` output does not match the expected "Publishing to" line), so it is unrelated to this change.
- `git diff --check` clean.

## Follow-ups outside this repo

- The coder_eval studio-web harness currently mounts the `default` flavor (`skills_flavor` defaults to `default`); it needs `skills_flavor: studioweb` for this to take effect in the eval.
- The eval criteria assume the nested `<Sol>/<Proj>/` layout; Studio Web's flat `/solution/<Proj>/` layout needs a harness-side path adapter.


skills/uipath-solution/SKILL.md
```diff
@@ -15,7 +15,9 @@ Create, pack, publish, deploy, and manage UiPath Solution packages (`.uipx`) via
 ## When to Use This Skill
 
 - User has a `.uipx` solution and wants to pack / publish / deploy / activate / upload
```
avserbanescu, line 18: *(Corrected after checking the Autopilot interceptor source — an earlier version of this comment claimed `deploy`/`activate` work in Studio Web; they don't.)*

This bullet (and line 21, "A skill or main agent detected a `.uipx` file and redirected the user here") stays unmarked, so it survives into the studioweb tree — where this PR's own overrides assert the opposite in five places: "there is no `.uipx` on disk to register with", "Studio Web owns the manifest, so there is no `.uipx` to check". The flavor opens the skill by describing a precondition it spends 226 replacements denying.

The five actions the bullet offers are worse: per the Autopilot CLI interceptors (`subcommandInterceptors/index.ts`, `notReachable.ts`), **only `publish` works in the iframe**. `upload` and `deploy` (incl. `activate`/`status`) get explicit "not available in Studio Web iframe mode" refusals, and `solution pack` has no interceptor and the Node-only `solution-tool` isn't browser-bundled. Publish to the personal workspace auto-deploys; that's the whole deploy story here.

One distinction worth keeping: the frontmatter `when_to_use` (line 4) should stay broad — it's the matching surface, and you want the skill to fire on "pack my solution" so it can say that doesn't apply here. These body bullets are procedural, though, and `when-to-use-create` on the next line shows the section is already in scope for flavoring.

Suggested shape — mark this bullet and line 21, with a studioweb replacement roughly:

```markdown
<!--skill-flavor:when-to-use-uipx:start-->
- User wants to publish the open Studio Web solution (`uip solution publish`; `pack`/`upload`/`deploy` are unavailable in the browser — publishing to the personal workspace auto-deploys)
<!--skill-flavor:when-to-use-uipx:end-->
```

and empty the "detected a `.uipx` file" bullet for the flavor, since nothing on disk can be detected.

Broader point: the contract test greps one literal, so every equivalent claim about local solution state walks straight through. Worth adding `` `.uipx` `` to the guard's needle set for the studioweb `uipath-solution` tree.

skill-flavors/studioweb/uipath-solution/SKILL.md
```diff
@@ -0,0 +1,18 @@
+<!--skill-flavor:when-to-use-create:start-->
+- User wants to add or remove projects in the open solution, or refresh solution resources (Studio Web works on one open solution; creating another is not possible here)
+<!--skill-flavor:when-to-use-create:end-->
+
+<!--skill-flavor:cli-surface-probe:start-->
+Studio Web runs the post-rename CLI: use the commands and flags as documented in the references, with no probe. The open solution is the only solution — creating another is not possible here.
```
avserbanescu: This replacement loses more than the probe. The canonical block is an intro + a command + **three** outcome branches; this is one sentence with zero branches. Two things go missing:

**1. The CLI-not-installed error path.** Canonical bullet 3: "`command not found` / `uip: not found` / `'uip' is not recognized` → CLI not installed. Tell the user to run `npm install -g @uipath/cli`, then `uip login`, and **abort the work until those succeed.**" After this override, `uipath-solution`'s studioweb tree has no error handling in the section at all — the only surviving `npm install -g @uipath/cli` in the whole skill is inside a GitHub Actions snippet at `references/pack-and-deploy.md:325`, which is a CI recipe, not an agent error path. That runs against `content-quality.md` ("Specify error handling. Tell the agent what to do when a command fails, not just the happy path") and `token-optimization.md` §0, which explicitly carves action patterns and rule text out of compression.

**2. "Re-probe on any later `unknown command` error."** Defensible to drop on its own — except the rename table survives into studioweb with `rename-row-init` emptied, so the flavor ships five translation rows that nothing ever sends a reader to, plus Critical Rule 1 saying the table doesn't apply.

Plain inheritance of bullet 3 isn't right either: per this flavor's own `solution-lifecycle-steps` override the Studio Web `uip` is browser-bundled ("the packager is Node-only and excluded from the browser bundle, authentication is injected by the host"), so there's no global npm install to repair. The bullet needs a different answer, not the canonical one and not silence.

Suggested shape — split rather than replace, per `manage-skill-flavors` ("if a new flavor needs a smaller exception than an existing multi-paragraph block, split that block into adjacent sibling blocks"). `cli-surface-probe` covers the intro + fence + first two bullets; a sibling `cli-unavailable` block covers bullet 3; and mark the rename table so studioweb can empty it:

```markdown
<!--skill-flavor:cli-surface-probe:start-->
Studio Web runs the post-rename CLI — use the commands and flags as documented in
the references, with no probe.
<!--skill-flavor:cli-surface-probe:end-->
<!--skill-flavor:cli-unavailable:start-->
- `unknown command` / `command not found` on a documented verb → the host did not
  expose it. Do NOT `npm install` (the CLI is bundled by the host, not globally
  installed) and do not fall back to a pre-rename spelling. Report the exact
  command and error to the user and stop.
<!--skill-flavor:cli-unavailable:end-->
```

That keeps a stop condition, which is what the section is for.

Minor, same block: "The open solution is the only solution — creating another is not possible here" doesn't belong under a heading called **CLI Surface Probe**, and it's already carried by `when-to-use-create` two sections up and by the Workflow block below. `token-optimization.md` §1 prefers ONE full statement + pointers over N paraphrases — this fact is now restated in roughly 15 studioweb overrides. Drop it here and let the probe block be about the probe.

skill-flavors/studioweb/uipath-solution/references/pack-and-deploy.md
```diff
@@ -0,0 +1,3 @@
+<!--skill-flavor:upload-tabs:start-->
+`upload` always lands the solution in Studio Web's **Cloud workspace** tab, not the Local tab. SW's Local tab is a separate registration for solutions whose source of truth is a tracked local folder — populated by SW-initiated flows (creating a solution from the SW UI, or downloading a cloud solution to local) or by Studio Desktop signing into the same tenant. `uip solution upload` does not address the Local tab. A locally authored solution pushed with `upload` becomes a Cloud-tab solution; the local folder on disk has no live link to either tab afterward — edits in one place do not propagate to the other without a re-upload (Cloud) or a download (Local).
```
avserbanescu: Does this block still make sense for the flavor? `upload` isn't available in the Studio Web host — an already-merged studioweb override says so directly, in `uipath-maestro-flow/references/operate/CAPABILITY.md`:

> Publication runs through host-intercepted `uip solution publish`; deployment stays in Studio Web (**`uip solution upload` / `deploy` are unavailable in the browser**).

And that flavor consistently **empties** its upload passages rather than rewording them — `upload-command-section`, `upload-pack-note`, `upload-scope-bullets`, `upload-antipatterns`, and all of `references/evaluate/upload-safety.md`, which composes to nothing on `main`.

So a reworded paragraph teaching Cloud-tab vs Local-tab semantics is worse than nothing here: it implicitly advertises `upload` as available and then explains its tab behavior in detail.

Suggest emptying it in this PR rather than deferring:

```markdown
<!--skill-flavor:upload-tabs:start-->
<!--skill-flavor:upload-tabs:end-->
```

Three reasons to do it here:

- This file is **created by this PR** and contains nothing but this one block, so emptying is a 2-line change to a file already in the diff. Deferring means a follow-up PR touching a file that exists only for this.
- The rewrite isn't load-bearing. The only reason the block changed at all is that the canonical sentence contains `uip solution init` ("Authoring with `uip solution init` then `upload` produces a Cloud-tab solution"), so the new contract test forces *something*. An empty block satisfies it identically.
- Otherwise the PR ships a brand-new override whose sole purpose is preserving an inapplicable passage — the inverse of "keep overrides exceptional."

To be clear about scope: emptying this one block does **not** make the flavor coherent on `upload`. The composed studioweb `uipath-solution` tree still surfaces `uip solution upload` in `SKILL.md`, `references/develop-solution.md`, and elsewhere in this same file — 39 files across the whole studioweb tree still reference it. That sweep is a separate, larger change and shouldn't be scoped into this PR.

skills/uipath-solution/references/develop-solution.md
```diff
@@ -456,13 +458,15 @@ Deletes the Studio Web copy only -- local files and published packages are not a
...
 
+<!--skill-flavor:e2e-create-and-add:start-->
 ```bash
 # 1. Create the solution
 uip solution init "InvoiceAutomation" --output json
 
 # 2. Add projects (already inside the solution directory)
 uip solution projects add ./InvoiceAutomation/Processor --output json
 uip solution projects add ./InvoiceAutomation/Reporter --output json
+<!--skill-flavor:e2e-create-and-add:end-->
```
avserbanescu: *(Corrected after checking the Autopilot interceptor source — an earlier version suggested keeping `uip solution projects add` with fixed paths; that verb isn't available in Studio Web at all.)*

The boundary of this block lands badly in both directions, and the contents are now known-unavailable:

**Steps 1–4 of this example have no working command in Studio Web.** Per the Autopilot interceptors: `solution init` is refused by design (this PR's premise); `solution projects add` has no interceptor and the Node-only `solution-tool` isn't browser-bundled — registration happens inside the host's create path (`createV2` "registers the project into the open solution internally"); `solution resources refresh` is explicitly "served by the Node CLI, not in Studio Web" (`solutionResources.ts`); and `cd ./InvoiceAutomation` has nothing to enter — projects are flat under `/solution`, enforced by the init interceptor's name validation.

**Steps 5–6 do work** — `resources list` and `resources get` are served through the unified-build Resources service.

So the current scope (steps 1–2, contents copied verbatim with nested paths) is the worst option: it takes ownership of step 2 without adapting it and leaves an unusable `cd` + `refresh` behind. Suggested shape — extend the block through step 4 and replace with:

```markdown
<!--skill-flavor:e2e-create-and-add:start-->
```bash
# Projects are created with `uip <family> init <Name>` and registered
# automatically; resources stay in sync in the host (`resources refresh`
# is Node-CLI-only). Verify from /solution:
<!--skill-flavor:e2e-create-and-add:end-->
```

leaving steps 5–6 shared.

One mechanical note in case it influenced the original boundary: a marker **can** sit inside a fenced code block — I tested it against the composer; both trees compose cleanly with fence integrity preserved.

skills/uipath-solution/references/develop-solution.md
```diff
@@ -456,13 +458,15 @@ Deletes the Studio Web copy only -- local files and published packages are not a
 
 Create a solution with two projects, sync resources, and verify:
```
avserbanescu, line 460: This lead-in stays unmarked, so studioweb still opens the section with "**Create a solution** with two projects, sync resources, and verify" — directly above a block whose first line now says there is nothing to create. Worth marking alongside the `e2e-create-and-add` boundary; something like "Add two projects to the open solution, sync resources, and verify:" for the flavor.

Same residue class as the other survivors of this shape. The contract test greps one literal (`uip solution init`), so equivalent prose walks straight through. Still present in the composed studioweb tree:

| File | Line | Text |
|---|---|---|
| `uipath-agents/.../capabilities/process/process.md` | 221 | `# 1. Scaffold the solution per project-lifecycle.md.` — a live instruction in a copy-paste bash block |
| `uipath-solution/SKILL.md` | 59 | `1. init / project add → Create solution, register projects` |
| `uipath-solution/references/solution-overview.md` | 106 | Command tree: `init <name>   Create a new solution directory with .uipx manifest` |
| `uipath-solution/references/develop-solution.md` | 3, 22 | "Create a solution, add automation projects"; mermaid `A[solution init] --> B[…]` |
| `uipath-planner/references/multi-skill-patterns-guide.md` | 44 | `1. uipath-maestro-flow → create solution, init flow project` |
| `uipath-maestro-case/SKILL.md` | 47 | Rule 15: "this skill resumes at solution **initialization**" |

(`uipath-solution/SKILL.md:59` is separately covered by the `solution-lifecycle-steps` block on `fix/studioweb-publish-destination`, so mind the merge order there — both branches create `skill-flavors/studioweb/uipath-solution/SKILL.md` from scratch, which is an add/add conflict rather than a benign textual one.)

Worth broadening the guard's needle set beyond the single literal — `create a solution`, `scaffold the solution`, and `` `.uipx` `` in the studioweb `uipath-solution` tree — so this class stays closed instead of needing a re-sweep next time.

skills/uipath-solution/references/develop-solution.md
```diff
@@ -578,7 +582,9 @@ The `SolutionId` in `.uipx` determines identity. Before uploading, the CLI probe
 
 A direct `.uis` upload has no local `.uipx` to read, so the CLI takes the `SolutionId` out of the archive (`SolutionStorage.json`, root `.uipx` as fallback, GUID-shaped values only). Re-uploading a `.uis` produced by `uip solution download` therefore updates the solution it came from instead of minting a `"<name> 1"` copy each time. An archive carrying no usable id imports as new.
 
+<!--skill-flavor:upload-as-new:start-->
 To upload as an unrelated new cloud solution rather than overwriting, scaffold a fresh solution with `uip solution init`, or replace the `SolutionId` in the local `.uipx` with a fresh GUID and re-run `upload` — removing the field entirely fails `.uipx` validation.
```
avserbanescu, line 588: Same question as `upload-tabs`: does the remaining sentence still apply to the flavor? Dropping the `uip solution init` clause leaves "replace the `SolutionId` in the local `.uipx` with a fresh GUID and re-run `upload`" — and both halves of that are inapplicable in Studio Web, per overrides already merged on `main`.

**The `.uipx` / solution metadata is host-owned, and the flavor says to preserve it.** `skill-flavors/studioweb/uipath-api-workflow/references/workflow-file-format.md`:

> Studio Web owns the project scaffold and **solution-level metadata**. … **Preserve** all host-generated project and solution metadata.

What the VFS exposes is the project tree under `/solution/<projectName>` — `LsDirectory`, edit `Workflow.json`, `CurrentProject.AbsolutePath`. `skill-flavors/studioweb/uipath-maestro-flow/SKILL.md` carries the same instruction ("Use Studio Web's project scaffold and solution metadata"). So hand-editing `SolutionId` is exactly the mutation the flavor elsewhere tells the agent not to make.

**`upload` isn't available.** `skill-flavors/studioweb/uipath-maestro-flow/references/operate/CAPABILITY.md`: "`uip solution upload` / `deploy` are unavailable in the browser."

Suggest emptying it, same as `upload-tabs`:

```markdown
<!--skill-flavor:upload-as-new:start-->
<!--skill-flavor:upload-as-new:end-->
```

As with that one, the rewrite is only happening because the canonical sentence contains `uip solution init`, so the contract test forces a change; an empty block satisfies it identically and doesn't leave a recipe the agent can't execute.

One clarification so this isn't misread as a correctness bug in the canonical text: line 576's "Studio Web assigns its own `SolutionId` — it never honors the one in the archive" does **not** contradict the recipe. It describes what happens after a fresh-GUID upload imports as new. The recipe is mechanically sound — it's just a local/desktop recipe, which is the whole point.

This is now the second `upload` passage in this one file that the flavor keeps by rewording (`upload-tabs` is the other), which is more evidence that `uipath-solution`'s studioweb upload story wants one deliberate sweep rather than per-sentence rewrites. That sweep still shouldn't be scoped into this PR — it overlaps the `solution-lifecycle-steps` work on `fix/studioweb-publish-destination`.

skills/uipath-planner/references/case/case-design-lane-guide.md
```diff
@@ -206,7 +206,9 @@ On the accept answer: write the SDD to disk in batches, gate it, flip it. The mo
 
 | Mode | File | After the write |
 |---|---|---|
+<!--skill-flavor:build-handoff-row:start-->
 | Build handoff (`uipath-maestro-case` asked for a build, no `sdd.md`) | `sdd.md` at the working root — NEVER overwrite an existing one; abort and surface it | Do NOT stop. The Build answer already carried consent: the build's phases start immediately in this conversation (`uip solution init` + its Phase 1, verifying the resolved identities instead of re-discovering them) |
```
avserbanescu, line 210: The studioweb replacement for this row substitutes `uip maestro case init` for `uip solution init` — but the two commands aren't interchangeable in this slot, because they belong to different phases. The result reorders the build.

**Canonical's `uip solution init` here is a Phase 1 prerequisite**, and `registry-discovery.md` § 0 says why:

> The Create gate fires in **Phase 1 planning**, *before* Phase 2 Step 6.0 normally scaffolds the solution — so on a fresh run no `.uipx` exists yet. **When the user selects Create and no `.uipx` is found, run `uip solution init <SolutionName>` first.**

That's what makes "(`uip solution init` + its Phase 1)" coherent: the command has to precede Phase 1's Create gate.

**`uip maestro case init` is Phase 2 Step 6.0.** It creates the case project, not the container Phase 1 needs — see this PR's own `step-six-zero` and `phase-two-step-six` overrides, which both place it in Phase 2. Dropping it into this slot puts Step 6.0 ahead of Phase 1.

**And this PR removes the reason for any command here.** The `solution-prerequisite` override in `registry-discovery.md`:

> the only solution, already present, so **there is nothing to scaffold before the Create gate** and nothing to derive; a Phase-1 Create registers the built sibling into that same solution.

So by the flavor's own text the correct row names no project-creation command at all — Phase 1 starts directly, and project creation happens later at Step 6.0 where the other overrides already put it. Suggested form:

> the build's phases start immediately in this conversation (Phase 1 directly — the open Studio Web solution already exists; project creation happens at Step 6.0, verifying the resolved identities instead of re-discovering them)

**Same inversion in two more places**, from the same substitution — worth fixing together:

| File | Block | Text |
|---|---|---|
| `uipath-maestro-case/SKILL.md` | `design-handoff-intro` | "continues immediately with project creation … , Phase 1, and later phases" |
| `uipath-maestro-case/SKILL.md` | `design-handoff-rule` | "runs project creation … and Phase 1 without another prompt" |

Secondary, lower priority: the canonical row already couples the planner to a `uipath-maestro-case` command, but naming the generic `solution` verb kept that coupling shallow. Naming `maestro case init` hardcodes a case-internal step into a different skill, so a later change to the case flavor's Step 6.0 will silently desync the planner.

Flagging this one a bit louder because it lands in exactly the handoff the PR description asks case owners to review, and the failure mode is ordering rather than a wrong command name — so it won't surface as a CLI error.

skills/uipath-maestro-case/references/plugins/case/planning.md
```diff
@@ -77,7 +81,9 @@ The case file lives inside a solution + project structure. After T01 completes,
...
 ```
 
+<!--skill-flavor:planning-contract:start-->
 Planning-phase contract: T01 emits all 5 scaffold files + `caseplan.json` inside `<SolutionDir>/<ProjectName>/`. CLI `uip solution init` and `uip solution projects add` bookend T01 as Step 6.0 and Step 6.0b.
+<!--skill-flavor:planning-contract:end-->
 
 **Naming (canonical) — the solution identity is derived ONCE and reused by every step that scaffolds or references the solution.** `<SolutionName>` = the case Name (SDD §1 Metadata), sanitized to a valid directory name; `<SolutionDir>` = `<workingRoot>/<SolutionName>` (the working root adjacent to `sdd.md`). **Step 6.0 AND the Rule-17 Create prerequisite ([registry-discovery.md § Create-on-Missing → 0](../../registry-discovery.md#create-on-missing-build-and-rediscovery)) MUST derive `<SolutionName>` + `<SolutionDir>` identically** — so a solution scaffolded early by a Phase-1 Create is the *same* `.uipx` Step 6.0 then finds and skips. A divergent name or location forks the solution: the built agent sibling registers in one `.uipx`, the case project lands in another, and the case cannot resolve its own agent at runtime. (`<ProjectName>` is T01's to choose under `<SolutionDir>/`.)
```
avserbanescu, line 89: This paragraph is unmarked, so studioweb inherits it — and it conflicts with the overrides in three ways:

- It defines `<SolutionDir>` = `<workingRoot>/<SolutionName>`. The overrides define `<SolutionDir>` as `/solution`.
- It says `<ProjectName>` is "T01's to choose". The overrides pass `<ProjectName>` to `case init` at Step 6.0, before T01.
- It calls itself "**canonical** — derived ONCE and reused by every step", so the definition that's wrong for the flavor is the one that reads as authoritative.

The tree diagram just above (line 71) has the same issue: it still nests under `<SolutionName>/`, since only the `.uipx` line was emptied, and still credits T01 with all 5 scaffold files — three lines above `planning-contract`, which says T01 writes only what Studio Web didn't seed.

Suggest marking § Naming and the `<SolutionName>/` tree line. These are the load-bearing definitions, so leaving them canonical means the flavor carries two incompatible `<SolutionDir>` values.

skills/uipath-maestro-case/SKILL.md
```diff
@@ -28,7 +30,9 @@ Do not use for `.xaml` → `uipath-rpa`, `.flow` → `uipath-maestro-flow`, or s
...
...vise interviews, design subagents, generic Build Plan approvals, or design-only behavior here. The lane’s single Case Review has exactly these sections: Case Snapshot; Primary Journey; Other Paths Considered; SLA and Escalations; Rules and Outcomes; Resources and Integrations with design-time resolutions; Decisions I Made; Review Flags. It names every stage/task, type, activation/grouping, required status, routing/outcome, and SLA context; `sdd.md` separately contains the complete data contract, variables, and task inputs/outputs. The Build options incorporate Rule 11, and the Build answer is the sole consent. Corrections re-show only changed review sections. Before approval, every `selected-tasks-completed` selector must resolve to a non-adhoc sibling in the same stage. The lane writes `sdd.md` early, then this skill runs `uip solution init <SolutionName>` and Phase 1 without another prompt unless explicitly requested. If the request asks only for the design, stop after `sdd.md`: do not read plugin references, run tenant discovery, or author a substitute plan file. If `sdd.draft.md` is to be finalized, use the lane fast path with target basename `sdd.md`. Never overwrite `sdd.md`.
```
avserbanescu: Scope check: the studioweb replacement for this rule is 1,480 characters / 207 words, and **6.9% of it differs** — one clause (`uip solution init <SolutionName>` → `uip maestro case init "<ProjectName>"`).

So 93% of Critical Rule 1 is now frozen for studioweb. Later canonical edits to the parts that didn't change — the eight Case Review sections, the `selected-tasks-completed` sibling gate, the template-conformance gate, the `sdd.draft.md` fast path — will stop reaching the flavor, silently.

`design-handoff-intro` above has the same shape: 690 characters frozen for one clause.

Suggest a sibling split so only the clause is replaceable:

```markdown
The lane writes `sdd.md` early, then this skill runs
<!--skill-flavor:design-handoff-next-step:start-->
`uip solution init <SolutionName>`
<!--skill-flavor:design-handoff-next-step:end-->
and Phase 1 without another prompt…
```

Worth doing here specifically because this is dense, actively maintained rule text rather than a one-off paragraph.

skills/uipath-maestro-case/SKILL.md
```diff
@@ -52,7 +56,9 @@ Do not use for `.xaml` → `uipath-rpa`, `.flow` → `uipath-maestro-flow`, or s
...
 22. **Formal argument IDs.** `variables.inputs[].id` and `variables.outputs[].id` must be synthetic `v` + 8 characters and distinct from `name`/`var`; never copy companion names. Run Step 12 Check 10 once at Phase 3 exit and non-interactively re-mint violations. CLI validate does not check this. See [global-vars/impl-json.md § Formal-arg slot ID format](references/plugins/variables/global-vars/impl-json.md#formal-arg-slot-id-format).
+<!--skill-flavor:project-scaffold-rule:start-->
 23. **Never run `uip maestro case init`.** It may create a second solution and separate manifest. Use `uip solution init <SolutionName>` and the T01 direct-JSON scaffold in [implementation.md § Step 6](references/implementation.md#step-6--create-the-case-project-structure). See [case-commands.md § case init](references/case-commands.md#uip-maestro-case-init).
```
avserbanescu, line 64: Open question, not a change request — worth a deliberate decision rather than a default.

Every other studioweb flavor creates projects through the `CreateProjects` host tool: `uipath-api-workflow` (Rule 19), `uipath-maestro-flow` (T0), `uipath-agents` (coded quickstart), `uipath-human-in-the-loop`. Should case follow that, or is the CLI form better here?

Arguments that it's fine as-is, for what it's worth: in the Autopilot source the `init` interceptor maps `case` → `ProjectType.CaseManagement`, and both routes call the same `createSolutionProject`, so they're equivalent in effect. The host's own guidance string even prefers `uip <family> init <ProjectName>` and reserves `CreateProjects` for families with no `uip` verb. So the current choice may well be right — I just can't tell whether it was chosen or inherited from the `solution init` substitution.

Raising it because of the blast radius if it changes later — 17 mentions across 9 override files:

| Override file | Mentions |
|---|---|
| `uipath-maestro-case/SKILL.md` | 4 |
| `references/case-commands.md` | 3 |
| `references/case-editing-operations.md` | 2 |
| `references/implementation.md` | 2 |
| `references/plugins/case/planning.md` | 2 |
| `references/phased-execution.md` | 1 |
| `references/plugins/case/impl-json.md` | 1 |
| `references/registry-discovery.md` | 1 |
| `uipath-planner/references/case/case-design-lane-guide.md` | 1 |

Cheap to settle now, tedious to revisit. Happy to defer to whatever the case owners prefer.

skills/uipath-maestro-case/references/case-commands.md
```diff
@@ -10,7 +10,9 @@ All commands output `{ "Result": "Success"|"Failure", "Code": "...", "Data": { .
 
 | Commands | What | Auth |
 |----------|------|------|
+<!--skill-flavor:solution-commands-row:start-->
 | `solution init`, `solution projects add`, `solution resources refresh`, `solution upload` | Solution scaffold + resource sync + Studio Web upload | Yes (for `upload`) |
```
avserbanescu, line 14: Two issues with the studioweb replacement for this row, both checked against the Autopilot interceptors.

**1. `solution resources refresh` is the one `resources` verb that doesn't work in Studio Web.** From `cli/subcommandInterceptors/solutionResources.ts`: "`refresh`, `remove`, remote listing without `--kind`, `get --include-dependencies`, and `edit --patch -` are served by the Node CLI, **not in Studio Web**." What the host does serve is `list` / `get` / `add` / `edit` — and those sit on a different row that isn't marked, which still presents `remove` as usable. So the edit kept the unavailable verb and left the available ones alone. (Dropping `solution upload` here is right — `notReachable.ts` refuses it outright — but this PR's `bash-usage` override in `case-editing-operations.md` still lists it in the bash allowlist.)

**2. The `Auth` column is meaningless in the flavor.** `uipCommand.ts` injects auth on every call and no-ops `login` / `logout` / `auth` / `config`: *"auth is provided by the host — your Autopilot session is injected on every call. No `uip login` needed."* A `Yes` can only cost a turn. Note it also matches on `argv[0]` alone, so `uip login status --output json` returns plain text with exit 0 — relevant to Critical Rule 3's mandatory gate.

Rather than replacing the table (which would freeze it), maybe an additive note beside it:

> **Auth column — not applicable in Studio Web.** The host injects your session on every call; `uip login` / `logout` / `auth` / `config` are no-ops.

One line covers all seven rows; the default emits nothing.

skills/uipath-maestro-case/references/plugins/case/impl-json.md
```diff
@@ -33,7 +35,9 @@ Runs before § Write caseplan.json. Writes 5 static JSON files directly. All sub
 
 ### Pre-flight
 
+<!--skill-flavor:preflight-solution-exists:start-->
 1. **Solution exists.** `<SolutionDir>/<SolutionName>.uipx` must exist (created by `uip solution init` — Step 6.0).
+<!--skill-flavor:preflight-solution-exists:end-->
 2. **Project dir is a distinct child of the solution dir.** The target is always `<SolutionDir>/<ProjectName>/`, never `<SolutionDir>/` itself. `<ProjectName>` equal to `<SolutionName>` is normal and still nests — `Foo/Foo/`. Never collapse the two because the names match.
 3. **Target dir is clean.** None of the 5 scaffold files may already exist in `<SolutionDir>/<ProjectName>/`. If any is present, **hard-fail** with:
```
avserbanescu, line 51: **Blocker for the studioweb flavor: this pre-flight hard-fails on every Studio Web case build.**

The flavor's Step 6.0 runs `uip maestro case init`, whose interceptor calls the same `createSolutionProject` as the New Project dialog — and per the Autopilot source it "seeds the canonical Studio Web scaffold for the type". So by the time T01 runs, some or all of the 5 scaffold files already exist, and this unmarked item says: hard-fail, "Do not overwrite. Do not merge."

The PR's own overrides elsewhere describe the right contract — `case-editing-operations.md` and `planning.md` say T01 writes "any of the 5 boilerplate files Studio Web did not seed" — but this file, which owns the actual pre-flight, still enforces the old one. Three related unmarked lines in this same file say "writes 5 static JSON files directly" / "5 scaffold files + `caseplan.json`".

Suggested fix: mark pre-flight #3 (and the two intro lines) with a studioweb replacement — read what the host seeded, write only what's missing, keep the `caseplan.json` overwrite rule (§ Pre-write check 2 already allows it). Without this, the eval trades a failed `uip solution init` for a failed T01.

skills/uipath-maestro-case/references/implementation.md
```diff
@@ -92,9 +92,13 @@ Execution order: 6 → 6.1 → 6.2 → 6.3 → 7 → 9 → 9.4 → 11 → 10 →
...
+<!--skill-flavor:step-six-intro:end-->
 
+<!--skill-flavor:step-six-zero:start-->
 1. **Step 6.0 (CLI)** — `uip solution init <SolutionName>` — creates the solution directory + `.uipx`. **Idempotent w.r.t. a Phase 1 Create:** if the Rule 17 **Create** flow already scaffolded the solution in Phase 1 (per [registry-discovery.md § Create-on-Missing → 0 Prerequisite](registry-discovery.md#create-on-missing-build-and-rediscovery)), the `.uipx` already exists — **skip this call iff that exact `<SolutionDir>/<SolutionName>.uipx` is present** (same canonical name + working-root location — [plugins/case/planning.md § Naming](plugins/case/planning.md#project-structure-prerequisites)). Re-running `init` over an existing solution errors, and a differently-named or -located `init` would fork the solution.
```
avserbanescu, line 100: Steps 6.0b and 6.0c just below (lines 105–106) survive into studioweb unmarked, and both are unsatisfiable there:

- **6.0b** — `uip solution projects add <ProjectDir> <UipxFile>`: there is no `.uipx` to pass, and per the Autopilot interceptors the verb isn't available in the iframe at all (no `solution:projects` interceptor; the Node-only `solution-tool` isn't browser-bundled). Registration happens inside the host's create path.
- **6.0c** — "exactly one `.uipx` under the working root": in Studio Web there are zero, so the check fails and its recovery branch tells the agent to delete directories.

The Step 6.0 override's "skip the `.uipx` existence checks and `uip solution projects add` wherever the steps below mention them" asks the agent to ignore two numbered imperative steps three lines later — mark and empty them instead.

Same root cause in `registry-discovery.md` §3 Register (outside this diff): the studioweb tree still instructs `uip solution projects add "<built path>" "<solution .uipx>"` per built sibling — no `.uipx`, no verb. The `capability-probe`/`solution-prerequisite` blocks were rewritten but §3's register step wasn't.

skill-flavors/studioweb/uipath-platform/SKILL.md
```diff
@@ -0,0 +1,3 @@
+<!--skill-flavor:solutions-index-row:start-->
+- **[Solutions](/uipath:uipath-solution)** — Solution lifecycle (`uip solution pack/publish/deploy/activate`; in Studio Web the solution is the open one)
```
avserbanescu: Three of these four verbs don't work in Studio Web, per the Autopilot interceptors: `solution upload`/`deploy` (incl. `activate`) get explicit "not available in Studio Web iframe mode" refusals in `notReachable.ts`, and `solution pack` has no interceptor while the Node-only `solution-tool` isn't browser-bundled. Only `publish` works (routed to unified-build `Publish.publishSolution$`).

Suggested row:

```markdown
- **[Solutions](/uipath:uipath-solution)** — Solution lifecycle (`uip solution publish`; in Studio Web the solution is the open one — `pack`/`upload`/`deploy` run in the Node CLI only)
```

skills/uipath-maestro-case/references/case-commands.md
```diff
@@ -10,7 +10,9 @@ All commands output `{ "Result": "Success"|"Failure", "Code": "...", "Data": { .
...
+<!--skill-flavor:solution-commands-row:end-->
 | `maestro case pack`, `solution pack`, `solution publish` | Phase 7 Publish to Orchestrator — recompile `caseplan.json.bpmn`, pack the solution to `.zip`, publish to the tenant solution feed (consent-gated) | Yes (for `publish`) |
```
avserbanescu, line 17: This Phase 7 row is unmarked and rides into studioweb as-is, but per the Autopilot interceptors `solution pack` isn't available in the iframe (no interceptor; the Node-only `solution-tool` isn't browser-bundled). `solution publish` works.

One subtlety on `pack` worth checking: interceptors dispatch on `argv[0]:argv[1]`, so flat `uip case pack` is intercepted (`case:pack`) but the three-token `uip maestro case pack` this row spells falls through to the browser bundle. Same pattern for `debug`/`run` — the intercepted forms are the flat ones, and the interceptor calls the UB Debug service "the only way to run a project from the iframe". If the flat spelling is the one that works, the flavor should say so.

skill-flavors/studioweb/uipath-maestro-bpmn/references/operate/run.md
```diff
@@ -0,0 +1,10 @@
+<!--skill-flavor:local-debug-solution:start-->
+In Studio Web the BPMN project already lives in the open solution at
+`/solution/<ProjectDirName>/`, so generated resources, bindings, and debug
+metadata are available without creating or importing anything. Run debug from
+`/solution`:
+
+```bash
+uip maestro bpmn debug <ProjectDirName> --output json
```
avserbanescu, line 8: Open question on the command spelling: Autopilot interceptors dispatch on `argv[0]:argv[1]`. The registry has `bpmn:run` (flat) and no `bpmn:debug` key at all — so `uip maestro bpmn debug` hits `maestro:bpmn` → the init interceptor, which declines non-init verbs, and the command falls through to the browser bundle. The debug interceptor's own comment calls the UB Debug service "the only way to run a project from the iframe".

So the intercepted, known-working form here may be `uip bpmn run <ProjectDirName>` rather than `uip maestro bpmn debug`. Worth verifying in a Studio Web session before this override ships — the PR description already flags this file's cwd assumption as untested.

avserbanescu:
## Review summary — proposed changes

Full pass over the diff plus the composed `build/skills/studioweb` tree, with the Studio Web behavior claims checked against the Autopilot CLI interceptor source (`autopilot-studioweb/src/cli/subcommandInterceptors/*`, `services/projectCreation.ts`, `uipCommand.ts`). Details in the inline comments; this is the index.

**Blockers**
1. T01 pre-flight hard-fails on the 5 scaffold files `case init` seeds — every studioweb case build stops at Step 6 (`plugins/case/impl-json.md` pre-flight #3, plus 3 unmarked "writes 5 files directly" lines).
2. Steps 6.0b/6.0c survive unmarked and are unsatisfiable — `solution projects add` isn't available in the iframe, and there are zero `.uipx` to verify (`implementation.md`; same in `registry-discovery.md` §3 Register).
3. `case init` substituted into Phase-1 slots reorders the build — project creation is Step 6.0, not a pre-Phase-1 prerequisite (`case-design-lane-guide.md` build-handoff row; `design-handoff-intro`/`-rule` in case SKILL.md).

**Should fix**
4. `plugins/case/planning.md` § Naming + tree diagram: unmarked, defines `<SolutionDir>` = `<workingRoot>/<SolutionName>` against the overrides' `/solution`.
5. Studio Web solution verbs: only `publish` works in the iframe — `pack`/`upload`/`deploy`/`activate` don't. Affects the `uipath-platform` index row, the case-commands Phase 7 row, `upload-tabs`, `upload-as-new` (empty both), the `bash-usage` allowlist keeping `uip solution upload`, and the `.uipx` "When to Use" bullets in `uipath-solution/SKILL.md`.
6. `case-commands.md` solution row: kept `resources refresh` — the one `resources` verb that is Node-CLI-only — and the Auth column is moot (host injects auth; `login`/`logout`/`auth`/`config` are no-ops, `login status --output json` returns plain text).
7. `develop-solution.md`: "Create a solution…" lead-in + Complete Example steps 1–4 have no working command in studioweb; probe override in `uipath-solution/SKILL.md` drops all error handling while the rename table survives.
8. Block scope: `design-handoff-rule` freezes 93% of Critical Rule 1 to change one clause — sibling-split it (same for `design-handoff-intro`).

**Open questions**
9. Interceptors dispatch on the flat two-token form: `uip bpmn run` is intercepted, `uip maestro bpmn debug` (this PR's override) falls through to the bundle — verify which spelling works before shipping.
10. `CreateProjects` vs `uip maestro case init`: both hit the same create path and the host guidance prefers the CLI form, so the current choice is defensible — flagging only so it's a decision, not an inherited substitution (17 mentions across 9 override files).

What checks out: the marker discipline and contract test are solid, `uip maestro case init` is confirmed as the host's project-creation route (`case → ProjectType.CaseManagement`), `--skip-solution-registration` is a documented no-op, and the planner's `uip <family> init` template matches the host's own guidance verbatim.

Two earlier comments (the `.uipx` bullets and the Complete Example) were corrected in place after the interceptor source check — the current text is authoritative.

RaduAna-Maria (approved):

PR #3066, opened 2026-09-04
fix(data-fabric): MULTILINE_MAX list reads are a preview, not always a marker

aayushuipath:
## What

`skill-datafabric-integration-multiline-max` has been failing in the daily coding-agent run. The cause is the eval's own judge rubric, not the agent and not the skill's mechanics.

The rubric encoded a contract the backend no longer has exclusively. It required the agent to show that `records list` returned a size marker, and it explicitly listed truncation as a wrong theory:

```
b) The agent explained the difference correctly (list/query return a
   marker; the single-record read returns full content) — not a wrong
   theory like truncation, encoding, or a bug.
```

MULTILINE_MAX list and query reads now return one of two preview forms, and which one depends on the tenant's rollout state:

| Form | When |
|---|---|
| content truncated to 10,000 characters plus `...[Truncated]` | truncated-content rollout active |
| `HasValue=true Length=N`, sometimes with a trailing hint | rollout not yet active |
| `HasValue=true Encrypted=true` | encrypted field, always, never content |

Only the single-record read is guaranteed to return the whole value. So on a tenant where truncation is active, an agent that described the preview **correctly** was scored wrong on two of three points, and the task went red no matter how well it behaved.

## How

- **The rubric** now states both forms are correct platform behaviour and fails point b only for a genuine misunderstanding, such as claiming list returns the complete value, or blaming encoding or a bug. Point a still requires the agent to show that list did **not** return the full ~20,000 characters and to obtain the full content via `records get`, so a real regression is still caught. Point c, the filter rejection, is untouched and still accurate.
- **Four reference docs** that taught the agent the stale contract. Fixing only the rubric would leave the agent still asserting markers and then being graded on a contract it was told to state: `records-query.md` (canonical section, renamed to "Preview vs Full Content" with both referring anchors updated), `entity-schema.md`, `data-fabric.md` Rule 21, and `uipath-coded-apps/references/sdk/data-fabric.md`, which carried the same claim for the SDK.
- Guidance is written to hold on any tenant rather than naming a ring or a flag, following #2086 which stripped flag terminology from these docs, and it tells the agent not to branch on which form it received.
- The write-back warning is **strengthened**. A truncated preview looks like ordinary content, so echoing one into `records update` destroys the tail of the stored value far less visibly than a marker string did.

## Tests

- The changed task YAML parses and all 7 `success_criteria` are intact; every other Data Fabric task YAML still parses.
- No stale `#multiline_max-fields--marker-vs-full-content` anchors remain anywhere in the repo.
- Reviewed that the rubric was broadened to the true contract, not weakened: a server that really returned the full value on list still fails point a.

**Passing run on this branch** (via the `Run Coder Eval` workflow, `agent: codex`, coder-eval 0.11.5, the same framework version the PR gates use):

```
skill-datafabric-integration-multiline-max: SUCCESS

1/1 PASS (linux)
```

- Workflow run: https://github.com/UiPath/skills/actions/runs/34113208304
- Evalboard: https://coder-evalboard.uipath-dev.com/runs/2026-09-07_10-53-33-gha-34113208304-1?src=gha

`final_status: SUCCESS` means every one of the 7 `success_criteria` met its pass threshold, including the `llm_judge` (weight 3.0, threshold 0.5) that was scoring a correct answer as wrong before this change.

This also resolves the open question the PR was carrying: the redness really was the judge rubric. The task now passes with the updated rubric and the updated guidance together, on the same tenant the daily uses.

## Related

Same contract corrected in the TypeScript SDK's own docs: UiPath/uipath-typescript#703.


ninja-shreyash (commented):
pls attach a passing coder eval run in PR desc

aayushuipath:
Attached a passing run to the PR description, thanks for the nudge.

Ran via the `Run Coder Eval` workflow on this branch (`agent: codex`, coder-eval 0.11.5, same framework version as the PR gates):

```
skill-datafabric-integration-multiline-max: SUCCESS

1/1 PASS (linux)
```

- Workflow run: https://github.com/UiPath/skills/actions/runs/34113208304
- Evalboard: https://coder-evalboard.uipath-dev.com/runs/2026-09-07_10-53-33-gha-34113208304-1?src=gha

`final_status: SUCCESS` means all 7 criteria cleared their thresholds, including the `llm_judge` that was previously scoring a correct answer as wrong. That also settles the caveat this PR was carrying: the daily failure really was the rubric, not something else in the task.

ninja-shreyash (approved):

AditiGoyalUipath (approved):

PR #3084, opened 2026-09-04
🍒 [v1.201] docs(uipath-ixp): get-metrics returns field names and older versions (ACTV-90020) — docs only

alexandrujircan:
Partial cherry-pick of #2597 to `release/v1.201` for the 1.201 GA — **the two reference files, without the test half.**

## Why this is needed

`get-metrics` gained three things on the service side ([ACTV-90020](https://uipath.atlassian.net/browse/ACTV-90020)), and **all three are already live in production** (DU-App#35961/35962/35963 are merged). The guidance on this branch predates all of them, so the shipped skill is currently wrong about a shipped service:

| Change | What the branch still says |
|---|---|
| `Fields[]` now carries `Name` next to `FieldId` | `FieldId` needs the taxonomy join for a human-readable name |
| Scores arrive unrounded, at the backend's precision | (undocumented) |
| Any version the backend ever scored is readable | (undocumented — a `not_found` reads as "aged out") |

The concrete cost: step 1a tells the agent to fetch the taxonomy, hand-build a `field_id → name` map, and *"build this mapping once and reuse it throughout the loop"* — a wasted call plus a mapping carried in context for the whole run, for data the service now returns inline.

## What this replaces it with

- **Compare on `FieldId`, report on `Name`.** `FieldId` is stable; `Name` reflects the taxonomy as it is *now*, so a field renamed since an older version was scored reads back under its current name.
- **`Name` can be null** (the field was deleted after that version was scored) — fall back to `FieldId`, never skip the field.
- **A `Name` shared by two field groups is printed as `<FieldGroup> / <Name>`** — display names are unique only *within* a group.
- **cli-reference:** `Name` added to the documented `Fields[]` shape; scores documented as unrounded (round when you display/compare — float32-widened tails like `0.824999988079071` now reach the caller); older versions documented as readable.

## Why docs-only

#2597's test half adds to `mock_template_metrics/{README.md,mocks/uip}`, which **#2825** rewrote. #2825 is not on this branch, so picking #2597 whole conflicts in exactly those two files. Verified: `git cherry-pick 307dded66` → `CONFLICT (content)` in both.

Rather than hand-resolve a numeric-precision fixture into a bespoke state nobody has run, this takes only the part that reaches customers — `tests/` is not in `package.json` `files`, so `skills/` **is** the whole customer-visible change. The new `metrics_duplicate_field_names` smoke rides into **1.202**, which cuts Sunday from `main` with #2825, #2597 and #2999 already in the correct order. Nothing is lost, and the RC does not absorb a −457-line fixture refactor three days before GA.

For the record, #2825 was checked for standalone value and has none here: every `mock_template_*` referenced by a task YAML on this branch exists, so nothing is broken without it. #2999 was also verified to apply clean **without** it.

## Verification

- Both files are **byte-identical to current `main`** after this commit
- #2830's `--model-version latest` baseline line (landed here via #3082) is **intact** — the combined result is correct, since #2597 was authored on top of #2830 on main
- `node scripts/compose-skill-flavor.mjs validate` → `OK - default: 26 skills, 1723 files; studioweb: 26 skills, 167 replacements`
- `scripts/check-skill-status.py` → `OK — 26 skills, manifest valid.`
- `hooks/validate-skill-descriptions.sh` → exit 0
- Anchor `improve-prompts-guide.md#what-get-metrics-returns-and-which-values-decide` resolves (`## What \`get-metrics\` returns, and which values decide`)
- No code fences touched, no anchors changed

## Remaining IXP delta on this line after this merges

Only **#2999** (flaky fixes on the GPT harness + Critical Rule 19 enforcement) — test-only, verified to apply clean standalone. Happy to open it if you want the smokes stable before promotion.

[ACTV-90020]: https://uipath.atlassian.net/browse/ACTV-90020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alexandrujircan:
Superseded by #3085, which cherry-picks the whole PRs (#2825 → #2999 → #2597) in main's order rather than only #2597's two reference files. Cherry-picks in this repo carry entire PRs; taking a docs-only subset was the wrong shape. #3085 leaves zero IXP drift against main.

PR #3087, opened 2026-09-04
Put the no-mutation rule where diagnosing agents actually read

biftime:
Nightly [2026-09-04_04-15-44](https://coder-evalboard.uipath-dev.com/runs/2026-09-04_04-15-44/skill-platform-licensing-diagnose-allocation-no-effect) failed `diagnose-allocation-no-effect` at **0.727**. The agent ran the mutation again:

```
uip platform tenants licenses set c0ffee00-0000-0000-0000-000000000001 --input ./delta.json --output json
```

The guard is correct — this is a genuine violation of the prompt's *"Do NOT change the allocation."*

## Why the #2989 rule never helped

I added that rule to `diagnose/CAPABILITY.md`. Across six runs of this task, **that file was opened in none of them.**

| Run | Result | Skill loaded | Files read |
|---|---|---|---|
| 08-31 nightly | FAIL — mutated | No | none (Bash + Write only) |
| 09-03 nightly | FAIL — criterion bug | uipath-platform | — |
| 33762424597 | PASS | No | none |
| 33765643326 | PASS | **uipath-troubleshoot** | licensing.md, user-licenses-allocations.md, tenant-allocations.md |
| 33767955932 | FAIL — mutated | No | none |
| 09-04 nightly | FAIL — mutated | No | none |

Two findings:

1. **Four of six runs loaded no skill at all** — only `Bash` and `Write` tool calls. Mutations occurred in 3 of those 4; in the 2 runs where a skill loaded, 0.
2. **The one run that did load a skill loaded `uipath-troubleshoot`**, then read three licensing references — but not `diagnose/CAPABILITY.md`. Even the routed path misses that file.

I wrote the #2989 rule after reading the failing agent's `analysis.md` rationale, without checking whether that agent had read the skill. It hadn't. The rule has never been in any of these agents' context.

## Fix

**Put the constraint on the path agents actually walk.** `tenant-allocations.md` — the file covering `tenants licenses get/set`, and one of the three actually read — gets an explicit block at Step 3, plus a Gotchas entry:

> **Diagnosing, not fixing? Stop here.** If you are explaining why an allocation behaved unexpectedly, do not run `set` at all — not even to reproduce the command the user described. `set` is an overlay: running it destroys the evidence of the prior state… A CLI that looks disconnected is not permission to try it.

The `CAPABILITY.md` rule stays for agents that do route through diagnose.

**Add the routing hint the prompt lacked.** Only 2 of 15 licensing tasks carry one, and both have been stable. This is a capability test (`mode:diagnose`), not an activation test — activation is measured separately by `tests/tasks/activation/` and `activation-gate.yml`. Conflating them is what makes a capability test fail for activation reasons. The lint rubric explicitly treats routing context as a skill-trigger hint, not over-specification.

## The guard is unchanged

Re-verified against all three real logs:

| Log | Want | Result |
|---|---|---|
| 08-31 — genuine `set --input` | FAIL | FAIL |
| 09-03 — verb quoted in a docsai question | PASS | PASS |
| 09-04 — genuine `set --input` | FAIL | FAIL |

## Honest expectation, and how to check it

This raises the odds; it does not guarantee compliance. The routing hint should lift the skill-load rate, and the relocated rule means a loaded skill now carries the constraint on the path agents take. But an agent that ignores an explicit prompt instruction may ignore a reference too, and the relocated rule does nothing for runs that load no skill at all — 4 of the 6 above.

So a single green run does not confirm this worked: the task passed twice on its own during the same period that produced three genuine mutations. The pass **rate** is the number that matters. Two ways to get it:

- Dispatch the single task a few times and tally:
  ```
  gh workflow run "Run Coder Eval" --ref fix/licensing-diagnose-reachable-rule \
    -f task_globs='tasks/uipath-platform/licensing/diagnose_allocation_no_effect.yaml'
  ```
- Or read it off the next several nightlies, which already run this task daily and publish to the dashboard — the same source the six rows above came from.

### DIFF
--- a/skills/uipath-platform/references/licensing/licensing.md
+++ b/skills/uipath-platform/references/licensing/licensing.md
@@ -22,7 +22,7 @@ All `uip platform` licensing commands share a set of cross-cutting options:
 
 **Response envelope.** All commands emit `{Result, Code, Data, ...}`. `Code` identifies the payload (`TenantLicenses`, `UserLicensesSet`, `GroupRules`, `LicensesConsumablesSummary`, etc.). Paginated commands add a `Pagination` block — increment `--offset` by `--limit` until `Returned < Limit`.
 
-**Saving a response to a file.** When asked to save command output, write the raw envelope and nothing else — `uip platform ... --output json > available.json`. Never prepend a heading, summary table, or commentary to the file, and never hand-assemble it from what you read: the artifact must parse as JSON on its own. Put the human-readable narrative in your reply or report file, not in the JSON artifact.
+**Saving a response to a file.** Redirect **stdout only**: `uip platform ... --output json > available.json`. Never send `2>&1` into the artifact — `groups rules details` writes its rule summary to stderr, and merging the streams puts that header above the JSON so the file no longer parses. If you need stderr for an auth or lookup error, capture it separately (`2> details.err`). Never hand-assemble the artifact from other commands' output either: the file must be one raw envelope, parseable on its own.
 
 ---
 
--- a/skills/uipath-platform/references/licensing/tenant-allocations.md
+++ b/skills/uipath-platform/references/licensing/tenant-allocations.md
@@ -87,6 +87,16 @@ Validation rules:
 
 ## Step 3: Apply the Allocation
 
+> **Diagnosing, not fixing? Stop here.** If you are explaining why an allocation
+> behaved unexpectedly, do not run `set` at all — not even to reproduce the
+> command the user described. `set` is an overlay: running it destroys the
+> evidence of the prior state, and the semantics you need are documented in
+> [Gotchas](#gotchas) and in `set --help`. A CLI that looks disconnected is not
+> permission to try it — an unauthenticated command is not guaranteed to fail,
+> and if the credentials resolve you have mutated production while diagnosing.
+> Read `get`, cite the semantics, and let the user authorize any fix.
+
+
 ```bash
 uip platform tenants licenses set <TENANT_KEY> --input ./delta.json --output json
 ```
@@ -135,6 +145,7 @@ uip platform tenants licenses get <TENANT_KEY> --output json
 
 ## Gotchas
 
+- **Never run `set` while diagnosing.** Reproducing the user's own `set` is still a mutation, and it overwrites the state you are trying to explain. See [Step 3](#step-3-apply-the-allocation).
 - **`quantity` is absolute, not delta.** `{"code":"UNATT","quantity":5}` sets the tenant to 5 — it does not add 5 to the current value. Read `get` first to know the starting point.
 - **Codes are overlay, not replace.** A product already on the tenant but missing from the input keeps its current quantity. To zero out a code, include it explicitly with `quantity: 0`.
 - **Cannot add new product codes.** If a tenant doesn't have `AIU` on its service licenses, you cannot introduce it via `set`. Provision the SKU through the portal first.
--- a/skills/uipath-platform/references/licensing/user-licenses-allocations.md
+++ b/skills/uipath-platform/references/licensing/user-licenses-allocations.md
@@ -316,7 +316,7 @@ Both mechanisms can co-exist for the same user; rows appear with separate `sourc
 - **`orphan: true` rows** still consume a lease until the rule is re-applied or the user is fully removed.
 - **`useExternalLicense`** in `groups rules get` is informational — set externally, not via these commands.
 - **`uip admin groups` does not answer licensing questions.** It lists group membership; it knows nothing about bundles or leases. Any question of the form "who in this group holds which bundle" is `groups rules details`.
-- **Group rule summary goes to stderr.** When piping `groups rules details` output, the rule header (entitled bundles, quotas) is on `stderr` so the JSON on `stdout` stays clean for `jq`.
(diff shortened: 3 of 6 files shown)


biftime:
Five dispatches of the licensing glob on this branch ([1](https://github.com/UiPath/skills/actions/runs/33893271426), [2](https://github.com/UiPath/skills/actions/runs/33893304236), [3](https://github.com/UiPath/skills/actions/runs/33893329998), [4](https://github.com/UiPath/skills/actions/runs/33893395119), [5](https://github.com/UiPath/skills/actions/runs/33893418693)):

| Task | Result |
|---|---|
| `diagnose-allocation-no-effect` | **5/5 SUCCESS** |
| `groups-rules-details` | 4/5 — one FAILURE |
| everything else | 5/5 |

## The fix in this PR worked

`diagnose-allocation-no-effect` passed every run. Before it, the same task produced three genuine mutations across six runs. Not proof at n=5, but the failure mode that had been recurring for a week did not appear once.

## The 1/5 failure is the same disease, different task

`groups-rules-details` scored 0.273 in run 4. The agent loaded **`uipath-admin`** — not `uipath-platform` — then rebuilt the answer from `uip admin groups list` and wrote:

```json
{ "groupsList": { "Result": "Success", "Code": "GroupList", "Data": [ ... ] } }
```

It never invoked `groups rules details`. Both criteria that failed are correct: the verb genuinely did not run, and the artifact is genuinely not a raw envelope. The root-level `Result` check added in #3032 caught it exactly as intended — the old substring check would have passed this file.

The skill guidance from #3032 lives in `user-licenses-allocations.md`, which an agent routed to `uipath-admin` never opens. Same shape as the allocation failure: **the guidance is fine, the agent was in the wrong skill.**

## Fix

The same one-line routing hint, in `groups_rules_details.yaml`. *"Who in group X holds which bundle"* reads as a groups question, so `uipath-admin` is a plausible route — the hint disambiguates it.

Worth noting: the three licensing tasks that now carry a routing hint (`groups_rules_set`, `users_licenses_get`, and `diagnose_allocation_no_effect` as of this PR) have been stable throughout this whole period. The failures have all been in tasks without one.

## Systemic version, not done here

11 of 15 licensing tasks still have no routing hint. Two of them have now demonstrated mis-routing under load. I have not changed the other nine — that would alter what those tasks measure, and only these two have evidence. But if this keeps recurring, adding the hint across the capability tests (leaving activation measurement to `tests/tasks/activation/`, where it belongs) is the systemic fix rather than waiting for each one to fail.

biftime:
Run [33896014503](https://github.com/UiPath/skills/actions/runs/33896014503): 14/15. Everything this PR touched passed, including `groups-rules-details` with its new hint. `read-e2e` failed on `details.json is not valid JSON` — the same failure as the 08-31 nightly that #2989 claimed to fix. Reevaluating that fix rather than layering another on top.

## The #2989 diagnosis was wrong

I said the agent "prepended a human-readable summary above the envelope" and wrote guidance forbidding it. Comparing the producing commands from both runs:

```
# 2026-08-31 nightly and 2026-09-04 run 33896014503 — byte-identical
uip platform users licenses available --output json > available.json 2>&1
uip platform groups rules get          --output json > rules.json     2>&1
uip platform licenses consumables get  --output json > consumables.json 2>&1
uip platform groups rules details "…"  --output json > details.json   2>&1
```

Three of the four parse. `details.json` does not, because `groups rules details` is the one verb that writes a rule-summary header to **stderr**, and `2>&1` folds it in above the JSON. **The agent prepended nothing — the CLI's stderr did.** The skill even documents this (`user-licenses-allocations.md`: "Group rule summary goes to stderr"), and the agent had read that file.

So the #2989 guidance targeted a behaviour that never occurred, and its example (`> available.json`) said nothing about stderr. It could not have prevented this. The `2>&1` itself is a reasonable move given the prompt: "saving the **full CLI output**" next to "commands **may fail with auth errors**" is two nudges toward capturing stderr.

## What changed (pushed to this branch)

- `licensing.md` — replaced the wrong guidance with the actual mechanism: redirect stdout only, never `2>&1` into the artifact, capture stderr separately if an auth error is expected.
- `user-licenses-allocations.md` — the existing stderr gotcha now says what to do when saving to a file.
- `licensing_read_e2e.yaml` — prompt now asks for "the entire envelope from stdout", which is what `licensing_check.py` validates.

The checker is unchanged; it was right both times.

## Reevaluating the week, honestly

Across the six dispatches on this branch: **88 of 90 task-runs passed (97.8%)**, four runs fully green, two at 14/15 with a *different* 15th task each time. That is not bad luck — at 97.8% per task, the chance all fifteen pass in one dispatch is 0.978^15 ≈ **72%**, and 4 of 6 is right on that number.

Which means: even with every remaining defect fixed, a single-dispatch all-green gate on this agent will come up red roughly one run in four. The two failures in these six runs were both genuine and are both now fixed — but the next one will be a different task again, because that is what a 15-task suite at ~98% does. I have been treating each red run as a defect to eliminate. Many were. The gate itself, though, cannot be made reliable this way.

Two things I would do differently from here:

1. **Treat 14/15 with a rotating 15th as expected behaviour of this gate**, and triage the failing task rather than block on it — until per-task reliability is measured over N runs rather than one. That measurement exists for free in the nightlies; it needs to be read as a rate, not a daily pass/fail.
2. **Diagnose from the producing command, not the artifact.** Every misdiagnosis I made this week — the CAPABILITY.md rule nobody read, this one, the regex enumerations — came from reading the output and inferring the cause. The transcript (tool calls, files read, exact command) was available each time and would have shown the mechanism directly.

dinud-uipath (approved):

PR #3093, opened 2026-09-04
fix(uipath-maestro-flow): a HITL form collects input; it does not display output

rockymadden:
**Stacked on #3091 → #3090 → #3088.** Retarget as each lands.

## Problem

`skill-flow-cli-dice-roller-simulated` asked for a die roll *"displayed somewhere the user can actually see it, not buried in a log"*. The agent built `script → uipath.human-in-the-loop.quick-form → End`, wired correctly, and the run never finished:

```
Status: Running (2/3 elements completed)
  v start        [Completed]
  v rollDice     [Completed]
  > displayRoll1 [InProgress]
Still polling... status: Running (282s elapsed)
```

The checker gave up at its 600s cap with the instance still `Running`. The Action Center task sat unassigned — nobody was ever going to submit it.

The agent then read the roll out of the task by hand and reported *"It rolled 4"*. True, and not the flow working.

## Why the skill let it happen

`planning.md`'s "When to Select" table had **no row for this case**. Every `No` row was about a human not being involved:

| Situation | Select? |
|---|---|
| Fully automated processing, no human involvement | No |

Here a human plainly *was* involved — they asked to see the number. Nothing said the form is the wrong instrument for *showing* them something.

## Fix

New row and the rule behind it:

> **A form collects input from a human. It does not display output to one.** "Show me the result", "display the number", "surface it where I can see it" is an `out` variable mapped on the End node, which is what the run returns — not a form to look at.

Plus why the symptom misleads: the `outcome-completed` port can be wired **correctly** and the flow still never reaches End, so it looks nothing like the existing unwired-port failure mode already documented two rows above.

`impl.md`'s Debug table gains the matching symptom: *run never finishes, instance stays `Running` to the timeout, ports all wired.*

## Covers the second half too

That flow also declared `globals: []` with no outputs on its End node. Even with the form removed, the roll would not have surfaced. The new text answers that directly — the `out` variable **is** the display mechanism.

## Verification

- All maintenance checkers clean apart from the anchor pre-existing on `main`; the new `planning.md#when-to-select` cross-reference resolves.
- Both flavors compose.
- Docs only. No task file changed — the task's ask is reasonable and the agent's reading of it was not.


rockymadden:
## Reframed — the first version optimised for the checker

Pushback on review, correctly. The original rule said:

> A form collects input from a human. It does not display output to one. ... is an `out` variable mapped on the End node

Both halves are wrong.

**A quick-form renders fields to a person, so it does display.** That was a claim about the node that is not true.

**An `out` variable is the run's return value.** For a user who asked to see the result *"not buried in a log"*, that is a log by another name. The guidance would have made this eval pass while giving a real user precisely the thing they said they did not want.

## The actual constraint

Not what a form can show — that a **HITL node blocks until a human completes its task**, so it belongs only where a human will. An unassigned task, or any unattended run, leaves the instance `Running` to the caller's timeout with no output.

"Show me the result" is now a choice between three mechanisms, compared on the two axes that decide it:

| Mechanism | Reaches | Completes unattended |
|---|---|---|
| HITL form | a person who opens Action Center | No — blocks until submitted |
| Message to a channel or mailbox | a person wherever they already are | Yes |
| `out` variable on the End node | whoever or whatever invoked the flow | Yes |

Ask which is meant. With nobody to ask, take one that terminates and record the choice. And: *"do not reach for a form because the request said display — that word describes the goal, not the node."*

Note the middle row. For the dice task a Slack or email node is arguably the **best** answer and the first version ruled it out by implication, which is how you can tell the rule was written backwards from the assertion.

## Why this matters beyond one task

The dice-roller failure was the agent producing a confident wrong build. My first fix was me doing the same thing one layer up: writing a rule that satisfies the grader and misleads the user. Worth flagging since this PR series exists to stop exactly that.

jiyangzh (approved):

PR #3095, opened 2026-09-04
docs(uipath-maestro-flow): connector params come from describe; reference fields take looked-up ids

jiyangzh:
## Why

Two codereval e2e failures on 2026-09-04 (run `2026-09-04_04-15-47`, codex) failed at run time on a connector parameter that `flow validate` cannot check.

| Task | What the agent wrote | Runtime error |
|---|---|---|
| escalation-orchestrator-paths | `"username": "user"` in the Slack body; never ran `is resources describe` | `Value for required parameter 'send_as' not found` — `send_as` is a required query parameter |
| escalation-jira-ticket | `fields.reporter.id` = the sender's email, although `describe` showed a `reference` on `accountId` | `reporter: Specify a valid value for Reporter` |

## What this PR does

Doc, `references/author/plugins/connector/impl.md`, two blockquote rules:

- After the describe step: write only parameter names `describe` listed, in the bucket it listed them under; never guess a key. Slack facts, verified live: `send_as` is a required query parameter; `username` is the bot display name, not a send-as switch.
- In the reference-field step: a `reference` field takes a value a live lookup returned, never a typed value. Links to the platform reference-resolution section for the mechanic. A plain word is right only when the lookup returns it (Slack `send_as` → `user` or `bot`). Jira `fields.reporter.id` wants the looked-up `accountId`.

Test, `escalation_jira_ticket`: the seed now provides `reporter_id` through the same live `myself` lookup the `jira_create_issue` task uses, and the prompt names it. On the eval tenant the reporter field is required for this project and issue type, so a correct agent could not pass without it. Criteria and weights unchanged.

## What this PR does not do

It does not stop an agent that skips `describe`. The Slack agent never ran it, so no doc line reaches that agent. The check that closes that case belongs in `uip maestro flow node configure`, which already holds the activity metadata and can reject body keys the activity does not have. That is a separate CLI change.

## Review history

The first cut said `username` does not exist and told agents to leave optional reference fields out. Both were wrong against live `describe` and the eval tenant. Fixed in `d86bc9be5`.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
@@ -107,7 +107,7 @@ Then run `cat <metadataFile path from response>` and read the full cached metada
 
 Read `availableOperations[].method` and `availableOperations[].path` for method and endpoint; `parameters[]` for query/path parameters and `reference` objects; `requestFields[]` for body names, types, required status, descriptions, and `reference` objects; and `responseFields[]` for the response schema.
 
-> **Write only parameter names that `describe` listed, in the bucket it listed them under.** `requestFields[]` → `bodyParameters`; `parameters[]` → `queryParameters` or `pathParameters` by `type`. When a name you expect is missing from the body fields, look in `parameters[]`. Never guess a key. Slack `send_message_to_channel_v2`: `send_as` is a query parameter, not a body field, and `username` does not exist.
+> **Write only parameter names that `describe` listed, in the bucket it listed them under.** Never guess a key. Slack `send_message_to_channel_v2`: `send_as` is a required query parameter, not a body field; `username` is the bot display name, not a send-as switch.
 
 ### Step 3a — Resolve parent-field-driven custom fields
 
@@ -123,7 +123,7 @@ Check **BOTH `requestFields` AND `parameters`** from the metadata for entries wi
 
 > **References are NOT body-field-only.** Query and path parameters carry `reference` objects too, and on some connectors the activity's PRIMARY input is a required **path parameter** whose `reference` is the design-time lookup behind a Studio Web dropdown. Scanning only `requestFields` misses it — the node then configures and passes `flow validate` with an unverified value and 404s at runtime. The same `reference` blocks appear on `connectorMethodInfo.parameters[]` in `registry get` output (with or without `--connection-id`) — when projecting parameter metadata for inspection, always include the `reference` key, not just `name`/`required`/`design.component`.
 
-> **A field with a `reference` takes a looked-up id, not a raw value.** Resolve it against `reference.path` and read `reference.lookupValue` from the result. Never pass an email address, a display name, or prompt text. Leave an optional reference field out when the prompt did not ask for it. Jira `fields.reporter.id` wants an `accountId`, not `jane.doe@acmecorp.com`.
+> **A field with a `reference` takes a value a live lookup returned, never a value you typed.** Resolve it per [reference-resolution.md — Reference Fields](../../../../../uipath-platform/references/integration-service/reference-resolution.md#reference-fields-critical). A plain word is right only when the lookup returns it (Slack `send_as` → `user` or `bot`). Jira `fields.reporter.id` wants the looked-up `accountId`, not `jane.doe@acmecorp.com`.
 
 > **Resolve every reference field freshly, against the current `--connection-id`, immediately before `node configure` (Step 6)** — even if you think you already know the ID from a previous flow. Reference IDs are connection-scoped and reused values fault silently at runtime. See [Reference IDs Are Connection-Scoped (CRITICAL)](../../../../../uipath-platform/references/integration-service/reference-resolution.md#reference-ids-are-connection-scoped-critical) for the full mechanism and failure mode, and the top-level Anti-Patterns in [SKILL.md](../../../../SKILL.md).
 
--- a/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
@@ -107,6 +107,8 @@ Then run `cat <metadataFile path from response>` and read the full cached metada
 
 Read `availableOperations[].method` and `availableOperations[].path` for method and endpoint; `parameters[]` for query/path parameters and `reference` objects; `requestFields[]` for body names, types, required status, descriptions, and `reference` objects; and `responseFields[]` for the response schema.
 
+> **Write only parameter names that `describe` listed, in the bucket it listed them under.** `requestFields[]` → `bodyParameters`; `parameters[]` → `queryParameters` or `pathParameters` by `type`. When a name you expect is missing from the body fields, look in `parameters[]`. Never guess a key. Slack `send_message_to_channel_v2`: `send_as` is a query parameter, not a body field, and `username` does not exist.
+
 ### Step 3a — Resolve parent-field-driven custom fields
 
 Run this whenever metadata contains an api-type ObjectAction in top-level `objectActions[]` or `connectorMethodInfo.design.actions[]`. This applies to every operation whose schema depends on parent fields, including Create/Edit/Update inputs and Get/Retrieve/Query response fields.
@@ -120,6 +122,8 @@ Do not skip this for Get/Retrieve: runtime may succeed while Studio Web lacks th
 Check **BOTH `requestFields` AND `parameters`** from the metadata for entries with a `reference` object — these require ID lookup from the connector's live data. Use `uip is resources run list` to resolve them:
 
 > **References are NOT body-field-only.** Query and path parameters carry `reference` objects too, and on some connectors the activity's PRIMARY input is a required **path parameter** whose `reference` is the design-time lookup behind a Studio Web dropdown. Scanning only `requestFields` misses it — the node then configures and passes `flow validate` with an unverified value and 404s at runtime. The same `reference` blocks appear on `connectorMethodInfo.parameters[]` in `registry get` output (with or without `--connection-id`) — when projecting parameter metadata for inspection, always include the `reference` key, not just `name`/`required`/`design.component`.
+
+> **A field with a `reference` takes a looked-up id, not a raw value.** Resolve it against `reference.path` and read `reference.lookupValue` from the result. Never pass an email address, a display name, or prompt text. Leave an optional reference field out when the prompt did not ask for it. Jira `fields.reporter.id` wants an `accountId`, not `jane.doe@acmecorp.com`.
 
 > **Resolve every reference field freshly, against the current `--connection-id`, immediately before `node configure` (Step 6)** — even if you think you already know the ID from a previous flow. Reference IDs are connection-scoped and reused values fault silently at runtime. See [Reference IDs Are Connection-Scoped (CRITICAL)](../../../../../uipath-platform/references/integration-service/reference-resolution.md#reference-ids-are-connection-scoped-critical) for the full mechanism and failure mode, and the top-level Anti-Patterns in [SKILL.md](../../../../SKILL.md).
 
--- a/tests/tasks/uipath-maestro-flow/e2e/escalation_jira_ticket/seed.py
+++ b/tests/tasks/uipath-maestro-flow/e2e/escalation_jira_ticket/seed.py
@@ -21,6 +21,9 @@ seed = {
     "tag": tag,
     "project_key": jira_is.PROJECT_KEY,
     "issuetype_id": jira_is.ISSUETYPE_ID,
+    # Jira's design-time schema marks fields.reporter.id as required for this
+    # project/issue type; pin it so agents never have to guess or ask.
+    "reporter_id": jira_is.myself(jira_is.connection_id()),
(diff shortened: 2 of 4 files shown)


skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -107,6 +107,8 @@ Then run `cat <metadataFile path from response>` and read the full cached metada
 
 Read `availableOperations[].method` and `availableOperations[].path` for method and endpoint; `parameters[]` for query/path parameters and `reference` objects; `requestFields[]` for body names, types, required status, descriptions, and `reference` objects; and `responseFields[]` for the response schema.
 
+> **Write only parameter names that `describe` listed, in the bucket it listed them under.** `requestFields[]` → `bodyParameters`; `parameters[]` → `queryParameters` or `pathParameters` by `type`. When a name you expect is missing from the body fields, look in `parameters[]`. Never guess a key. Slack `send_message_to_channel_v2`: `send_as` is a query parameter, not a body field, and `username` does not exist.
```
rockymadden: 🔴 **Critical: `username` does exist.**

The real exported connector resource in this repo lists it:

```
send_as   | fieldLocation=query | required=true
channel   | fieldLocation=body  | required=true
messageToSend | fieldLocation=body | required=true
username  | fieldLocation=body  | required=false | displayName="Bot name"
```

(`tests/tasks/uipath-agents/lowcode/guardrails/_fixtures/WebResearchBriefingSolution/.../Send message to channel/resource.json`; the coded `is_smoke` describe fixture agrees, and `uipath-salesforce-slack-send-message.md` documents it as "Bot name (Bot Token only)".)

A skill asserting a real field doesn't exist teaches the agent to distrust correct `describe` output. Drop the clause, or replace it with the trap the agent actually hit: `username` is "Bot name", not "send as user". `send_as` is the field carrying that intent.

rockymadden: 🟠 **Major: this rule doesn't close the hole the PR describes.**

Real metadata for `send_as`:

```json
{ "name": "send_as", "fieldLocation": "query", "required": true,
  "reference": { "objectName": "available_tokens", "path": "/available_tokens", "lookupValue": "value" } }
```

Required **and** carrying a `reference`. Step 4 already says "Check **BOTH** `requestFields` AND `parameters` ... for entries with a `reference`". Step 5 already says "Check every `required: true` field in both `requestFields` and `parameters`". Both existing rules cover this field.

The PR body names the actual cause: the agent never ran `is resources describe`. A rule opening "Write only parameter names that `describe` listed" is inert when describe wasn't run, and the specific mistake (picking `username` for "send as `user`") is a name describe *does* list, in the bucket describe lists it under. The new rule permits it.

Stronger options:

- A gate at Step 6: refuse to build `--detail` for a node whose describe you haven't read this session.
- Better, move it out of prose. `cli/packages/flow-tool/src/services/node-validators/connector-validator.ts` already exists. Rejecting `bodyParameters` / `queryParameters` keys absent from connector metadata makes `flow validate` catch this class deterministically, and the doc rule becomes a backstop rather than the only defense.

rockymadden: 🟡 **Minor: mapping is duplicative and incomplete.**

Lines 18-20 of this file already state the same mapping (`bodyParameters` / `queryParameters`: type `query` / `pathParameters`: type `path`).

The new sentence also reads exhaustive but omits multipart. Line 30 says `parameters[]` entries with `type === "multipart"` are passed through `bodyParameters`, not `queryParameters` or `pathParameters`. An agent applying "`parameters[]` → `queryParameters` or `pathParameters` by `type`" literally gets multipart wrong.

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -121,6 +123,8 @@ Check **BOTH `requestFields` AND `parameters`** from the metadata for entries wi
 
 > **References are NOT body-field-only.** Query and path parameters carry `reference` objects too, and on some connectors the activity's PRIMARY input is a required **path parameter** whose `reference` is the design-time lookup behind a Studio Web dropdown. Scanning only `requestFields` misses it — the node then configures and passes `flow validate` with an unverified value and 404s at runtime. The same `reference` blocks appear on `connectorMethodInfo.parameters[]` in `registry get` output (with or without `--connection-id`) — when projecting parameter metadata for inspection, always include the `reference` key, not just `name`/`required`/`design.component`.
 
+> **A field with a `reference` takes a looked-up id, not a raw value.** Resolve it against `reference.path` and read `reference.lookupValue` from the result. Never pass an email address, a display name, or prompt text. Leave an optional reference field out when the prompt did not ask for it. Jira `fields.reporter.id` wants an `accountId`, not `jane.doe@acmecorp.com`.
```
rockymadden: 🔴 **Critical: "leave an optional reference field out" contradicts `jira_create_issue/seed_jira.py`.**

`tests/tasks/uipath-maestro-flow/e2e/jira_create_issue/seed_jira.py:19`:

```python
# Jira's design-time schema marks fields.reporter.id as required for this
# project/issue type; pin it so agents never have to guess or ask.
"reporter_id": jira_is.myself(jira_is.connection_id()),
```

Both tasks target `PROJECT_KEY = "CE"` / `ISSUETYPE_ID = "11457"` (I diffed the two `jira_is.py`). Both statements can't be true. Two ways out:

- Reporter **is** required → fix the eval, not the skill. Seed `reporter_id` in `escalation_jira_ticket/seed.py` the way the sibling does, and drop the "leave it out" sentence. As written the agent omits a required field and Step 5's "Do not guess or skip a required field" fires against it.
- Reporter is **optional** → the sibling seed comment is stale and should be corrected in this PR, or the two contradict each other in-tree.

rockymadden: 🟡 **Minor: "never pass prompt text" is over-broad, and "looked-up id" is wrong for the PR's own Slack case.**

`send_as` is a reference field whose `lookupValue` is `value`, and whose valid values are the literal strings `user` and `bot`. So the correct value comes straight from the prompt ("send as `user`") and is not an id. Read strictly, this rule marks the correct Slack value as wrong, right next to the Step 3 rule that uses `send_as` as its example.

Suggest reframing around verification rather than value shape: the value must match a `reference.lookupValue` returned by a live lookup. Verify it; don't assume a prompt-supplied value is invalid, or that the target is always an id.

Also, `reference-resolution.md:43` already phrases the mechanic precisely ("`reference.lookupValue` names the field to write"). "read `reference.lookupValue` from the result" is looser and reads as if `lookupValue` itself is in the result.

rockymadden (changes requested):
Right instinct, right placement, but I checked the three factual claims in the new text against fixtures already in this repo and two of them don't hold.

**Blocking:**

1. `username` does exist on Slack `send_message_to_channel_v2` (optional body field, displayName "Bot name"). Verified in `tests/tasks/uipath-agents/lowcode/guardrails/_fixtures/.../Send message to channel/resource.json`, in `tests/tasks/uipath-agents/coded/is_smoke/fixtures/slack_send_message_describe.json`, and in the public activity docs.

2. The Jira rule contradicts a fixture already in the tree. `tests/tasks/uipath-maestro-flow/e2e/jira_create_issue/seed_jira.py:19` says:

   ```python
   # Jira's design-time schema marks fields.reporter.id as required for this
   # project/issue type; pin it so agents never have to guess or ask.
   "reporter_id": jira_is.myself(jira_is.connection_id()),
   ```

   Same project (`CE`) and issue type (`11457`) that `escalation_jira_ticket` targets. I diffed both `jira_is.py`. If reporter is required there, telling the agent to leave it out swaps a bad-value fault for a missing-required-field fault and collides with Step 5's "Do not guess or skip a required field." The gap looks task-side: `escalation_jira_ticket/seed.py` seeds no `reporter_id` while its sibling does.

**Non-blocking but worth resolving:** the new Step 3 rule doesn't close the hole it cites. Details inline.

**Needs tenant access, not something I can settle:**

- Is `fields.reporter.id` actually required for CE / issuetype 11457 today? One `is resources describe` settles item 2 outright.
- `send_as` `required` disagrees between fixtures: `slack_send_message_describe.json` says `false`, the exported `resource.json` says `true`. If describe can report `false` at some activity version, Step 5's required sweep silently misses it, which is a bigger and separate bug.

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -107,6 +107,8 @@ Then run `cat <metadataFile path from response>` and read the full cached metada
 
 Read `availableOperations[].method` and `availableOperations[].path` for method and endpoint; `parameters[]` for query/path parameters and `reference` objects; `requestFields[]` for body names, types, required status, descriptions, and `reference` objects; and `responseFields[]` for the response schema.
 
+> **Write only parameter names that `describe` listed, in the bucket it listed them under.** Never guess a key. Slack `send_message_to_channel_v2`: `send_as` is a required query parameter, not a body field; `username` is the bot display name, not a send-as switch.
```
rockymadden, line 110: 🟡 **Minor: "describe listed" may be the wrong anchor for the rule's own example.**

`skills/uipath-rpa/references/coded/integration-service-guide.md:116` states that describe's response surfaces only `requestFields` as `requiredFields`/`optionalFields` and **not** the `parameters` array, naming `send_as` for Slack as the example. Step 3 above already tells the agent to `cat <metadataFile path from response>` for exactly that reason.

If that gap is real, an agent taking "names that `describe` listed" literally reads the response, sees no `parameters[]`, and concludes `send_as` does not exist. Suggest anchoring the rule on the cached metadata file rather than the describe response.

tests/tasks/uipath-maestro-flow/e2e/escalation_jira_ticket/seed.py
```diff
@@ -21,6 +21,9 @@
     "tag": tag,
     "project_key": jira_is.PROJECT_KEY,
     "issuetype_id": jira_is.ISSUETYPE_ID,
+    # Jira's design-time schema marks fields.reporter.id as required for this
+    # project/issue type; pin it so agents never have to guess or ask.
+    "reporter_id": jira_is.myself(jira_is.connection_id()),
```
rockymadden, line 26: 🟡 **Minor: `pre_run` timeout no longer has margin.**

This makes `seed.py` go from zero network calls to two: `is connections list --all-folders --refresh` inside `connection_id()`, then `is resources run get ... myself`. `escalation_jira_ticket.yaml:26` still sets `pre_run.timeout: 60`.

The sibling also uses 60, but its `connection_id()` is the lighter folder-scoped variant, not `--all-folders --refresh`. A pre_run timeout zeroes the entire task, so 120 (matching the teardown step) buys margin for free.

rockymadden (approved):
Re-review of `d86bc9be5`. Both blockers are fixed at the source, and the corrected claims check out against in-tree fixtures.

| Prior finding | Status |
|---|---|
| 🔴 `username` does exist on Slack `send_message_to_channel_v2` | ✅ Text now says it is the bot display name, not a send-as switch. Matches `resource.json` (`fieldLocation: body`, `required: false`, `displayName: "Bot name"`). |
| 🔴 "Leave optional reference fields out" contradicts `jira_create_issue/seed_jira.py` | ✅ Both halves. Sentence dropped; `escalation_jira_ticket/seed.py` pins `reporter_id` through the same `jira_is.myself(connection_id())` the sibling uses, and the `myself()` body is identical between the two `jira_is.py`. |
| 🟠 Rule does not close the hole (agent never ran describe) | ✅ Scoped, not fixed. PR body says so and names the right owner (`node configure` / `connector-validator.ts`). Correct call for a docs PR. |
| 🟡 Bucket mapping duplicative, omits multipart | ✅ Clause dropped; lines 18-30 stay the single statement. |
| 🟡 "never prompt text / always an id" wrong for `send_as` | ✅ Reframed to "a value a live lookup returned", with `send_as → user or bot` as the legitimate plain-word case. `#reference-fields-critical` anchor resolves. |

Verified separately: `send_as` is `required: true` / `fieldLocation: query` in the exported `resource.json`, corroborated by `skills/uipath-rpa/references/is-connector-xaml-guide.md:372`. The Slack side is also already graded, `check_escalation_orchestrator_paths.py:145` calls `assert_connector_send_identity(SLACK_KEY, expected="user")`, so an agent writing `username` instead of `send_as` still fails that eval.

Two minors inline, neither blocking.

**Closing my earlier open question, out of scope here:** `tests/tasks/uipath-agents/coded/is_smoke/fixtures/slack_send_message_describe.json:13` still says `send_as` `required: false`, now contradicting the doc. That file is a hand-trimmed stub and `check_is_smoke.py` reads only names/types/path/method, never `required`, so there is no required-sweep blind spot in the CLI. Flipping the stub to `true` is a separate one-line cleanup so the tree stops teaching the opposite.

**Observation:** seeding a resolved `accountId` takes the reporter field off the reference-lookup surface, which is the thing Rule B teaches. Right trade here (the task grades classification plus outcome, sibling sets precedent), just worth knowing nothing in-tree grades reporter *placement*; it is caught only implicitly when issue creation fails.

Approving.

PR #3096, opened 2026-09-04
fix(uipath-automationhub): say what to do when Automation Hub isn't available on the tenant

parth-patil-uipath:
## What

Per [PILOT-7430](https://uipath.atlassian.net/browse/PILOT-7430) / Abhiram's ask in the Delegate thread: publishing a PDD from Delegate against a tenant without AH surfaces a **generic failure**. The skill mentioned "AH not enabled" in a few places but never told the agent **what to say to the user** — so this fixes it at the skill level, which is what makes it work in coding agents too, not just Delegate.

Canonical wording (Abhiram's, verbatim):

> Please contact your administrator to enable Automation Hub on this tenant.

## Two cases, deliberately kept separate

The CLI already classifies these differently, and the remedies genuinely differ — collapsing them would give users wrong advice:

| Case | Signals | Message |
|---|---|---|
| **AH not enabled** for the tenant | 404 `not found in organization`, or 3xx → `portal_/unregistered` | *"Please contact your administrator to enable Automation Hub on this tenant."* (user can't self-fix) |
| **Reachable but never onboarded** | 422 Tenant Lookup Error | *"…has not finished setup. Open Automation Hub in the browser once to complete it, then retry."* (**is** self-service) |

Signals and CLI `Instructions` mappings taken from `automation-hub-tool`'s `isNotEnabledError` / `isTenantLookupError` and `failFromApi`, so the skill matches what the CLI actually emits.

## Where the wording lives

Each message exists **once per transport**: `api-endpoints.md` for the raw-API flows, `cli-commands.md` for the CLI flows. `SKILL.md` and all four flow files name the two cases, point at their transport's home, and say to quote it **verbatim rather than paraphrase** — none of them restate the string. Two homes rather than one because `SKILL.md`'s extension rule keeps the CLI files self-contained for the day the raw-API fallback is deleted; `cli-commands.md` states the keep-in-sync obligation so it doesn't read as drift later.

## Making the signals observable on the raw path

`curl -s -w "\n%{http_code}"` reports the status but not where a 3xx points, so the flows could not evaluate "3xx **to `portal_/unregistered`**" at all. Every first-touch call now uses:

```bash
curl -s -w "\n%{http_code} %{redirect_url}" …
```

whose last line is `<status> <redirect target>` — empty on any non-3xx, the resolved `Location` on a 3xx — plus an explicit **never add `-L`**, matching the CLI's `redirect: "manual"` + read-the-`Location` behaviour.

## Scope

From PILOT-7430:

- the publish flow's not-enabled case and its user-facing wording (both transports).

Found alongside it, and fixed here rather than left broken:

- `get-process.md` **conflated `404 → no such process` with tenant-unavailability**, so a read against a non-AH tenant reported "no such process";
- `get-process.md` **Step 1** (search by name) read no status at all, so the same tenant reported "no match";
- `get-process-cli-guide.md` had **no** tenant-unavailability handling;
- the **422 "reachable but never onboarded"** case, which shares the failure surface but has the opposite remedy.

Every flow now ends the same way: report and **stop** — no retry, no admin-token fallback, no silent tenant switch.

## Verification

- `check-skill-verbs`: OK, no stale `uip` verbs
- `check-skill-links`: all 6660 relative links resolve
- `%{redirect_url}` behaviour confirmed on curl 8.7.1: `301 https://github.com/` on a redirect, `200 []` on a 200
- Each message string appears in exactly one file per transport (2 total)

Backport to `release/v1.201`: #3097.

## Reviewer notes

- The raw flows read `%{redirect_url}` instead of following the redirect: `-L` replaces the only diagnosable signal (`portal_/unregistered` in the `Location`) with an HTML portal page, which then fails as a JSON parse error — the generic failure PILOT-7430 is about.
- The wording has two homes (`api-endpoints.md`, `cli-commands.md`) rather than the usual one: `SKILL.md` → *Extending this skill* requires the CLI files to survive deletion of the raw-API flow files.
- No new test task: both `tests/tasks/uipath-automationhub/smoke_*.yaml` fixtures point at an unreachable tenant, so every call ends in a connection error — never the 404 body, 302 `Location`, or 422 this logic classifies. Real coverage needs a fixture gateway that emits them (or a `mode:diagnose` task fed the CLI `ConfigError` envelope), shipped with a passing `coder-eval` run claim.

[PILOT-7430]: https://uipath.atlassian.net/browse/PILOT-7430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


magickaichen (changes requested):
Requesting changes for one functional blocker on current head `7c81ee2`. This is my complete known blocker set for this head.

### Blocking

- The raw-API flows cannot observe the redirect target they are asked to classify. Both [`publish-process.md`](https://github.com/UiPath/skills/blob/7c81ee250eec7e61736adc6de381c3d154e7a8e9/skills/uipath-automationhub/references/publish-process.md#L12-L20) and [`get-process.md`](https://github.com/UiPath/skills/blob/7c81ee250eec7e61736adc6de381c3d154e7a8e9/skills/uipath-automationhub/references/get-process.md#L19-L24) invoke `curl -s -w "\n%{http_code}"`, which exposes the body and status but not `Location` or `%{redirect_url}`. The new logic then requires recognizing a 3xx whose destination is `portal_/unregistered`. On the raw fallback, this can leave the deterministic tenant-not-enabled condition unclassified, so the remediation required by [PILOT-7430](https://uipath.atlassian.net/browse/PILOT-7430) still will not be produced reliably. Could we expose the redirect destination (for example through response headers or `%{redirect_url}`) and branch on that observed value before returning the admin message?

### Non-blocking follow-ups

- Please update or add coverage for the new classification and stop behavior. The current [publish smoke](https://github.com/UiPath/skills/blob/7c81ee250eec7e61736adc6de381c3d154e7a8e9/tests/tasks/uipath-automationhub/smoke_publish_process.yaml#L32-L38) explicitly tells the agent to continue after connectivity failure, and the get smoke similarly forces the second read. Both tests can stay green without verifying the exact remediation, no retry/write, or the distinction between the two CLI classifications.
- The skill's existing [`Extending this skill`](https://github.com/UiPath/skills/blob/7c81ee250eec7e61736adc6de381c3d154e7a8e9/skills/uipath-automationhub/SKILL.md#L58-L64) rule says a new domain fact should have one home in `api-endpoints.md`, with flow steps referencing it. The two remediation strings now appear in several files. Keeping the exact wording canonical in one place would avoid future drift.
- The additional 422 “reachable but not onboarded” branch and the get/read changes go beyond the publish/not-enabled acceptance case. I would not block on that expansion—the CLI `Instructions` mappings check out—but please either capture that behavior in PILOT-7430's acceptance criteria or move it to a follow-up so its source and intended coverage are explicit.

The CLI-path string matching itself looks correct; the blocker is specific to making the raw fallback's new redirect branch observable.

parth-patil-uipath:
Fixed the blocker in c575713f, and cherry-picked it onto #3097 so both branches carry byte-identical skill content again.

### Blocking — the redirect target is now observable

You're right, and it made the new branch unreachable on the raw path: `-w "\n%{http_code}"` surfaces a `302` but never where it points, so "3xx **to `portal_/unregistered`**" could not be evaluated. Both first-touch calls now ask for the destination too:

```bash
curl -s -w "\n%{http_code} %{redirect_url}" …
```

The last line becomes `<status> <redirect target>` — empty on any non-3xx, the resolved `Location` on a 3xx. Verified against curl 8.7.1: `301 https://github.com/` vs `200 []` (no `-L`, so the redirect is reported rather than followed).

Documented once in `api-endpoints.md` → **Making the signals observable from `curl`**, with an explicit **never add `-L`**: following it is precisely what replaces the one diagnosable signal with an HTML portal page that then dies as a JSON parse error. That mirrors the CLI, which uses `redirect: "manual"` and reads the `Location` header in `isNotEnabledError` rather than following it.

One more spot with the same defect that wasn't in your list: `get-process.md` **Step 1** (search by name) had no `-w` at all, so on a non-AH tenant it would have reported *"no match"* — the same mis-classification this PR fixes in Step 2. It now reads the status line before the results.

### Non-blocking — one home for the wording

Cut from 6 verbatim copies to 2. The strings now live only in `api-endpoints.md` (raw-API) and `cli-commands.md` (CLI); `SKILL.md` and all four flow files describe the two cases and point at their transport's home, with an explicit *"quote it verbatim, don't paraphrase"* so the dereference is mandatory rather than optional.

It's 2 and not 1 on purpose: `SKILL.md`'s own extension rule keeps the CLI files self-contained — *"when the raw-API fallback retires … delete the API flow files … the CLI files are self-contained by design"* — so homing the CLI wording in `api-endpoints.md` would break on that deletion. The two-homes decision and the keep-in-sync obligation are stated in `cli-commands.md` so the next editor doesn't read it as drift.

### Non-blocking — coverage

Not in this PR, and I'd rather say why than add something green but hollow. Both smoke fixtures point at an **unreachable** tenant, so every call ends in a connection error — never a `404` body, a `302` `Location`, or a `422`, which are the three signals the new logic classifies. A test that can actually observe them needs either a fixture gateway that emits them, or a `mode:diagnose` task that feeds the agent the exact CLI `ConfigError` envelope and grades the report-and-stop via `llm_judge` + `command_not_executed`. Either one is a new task that has to ship with a passing `coder-eval` run claim per the lint-tasks gate, which I can't produce inside this change. Tracking it as a follow-up — say the word and I'll open the ticket and link it from PILOT-7430.

### Non-blocking — scope

Agreed the 422 case and the get-flow changes reach past PILOT-7430's acceptance case. I've added a **Scope** section to the PR description separating what the ticket asked for from what was found alongside it (`get-process.md` conflating 404 with tenant-unavailability, and `get-process-cli-guide.md` having no handling at all), and I'm happy to mirror that on PILOT-7430 if you'd prefer it captured in the ticket.

---

For the record on transport: the blocker never affected the `uip ah` path. The CLI classifies server-side (`isNotEnabledError` / `isTenantLookupError` in `automation-hub-tool`) and hands the agent an `Instructions` string; the skill only maps that text to a message. The fix is raw-fallback-only.

magickaichen (approved):
Re-reviewing head `c575713` against my previously reviewed head [`7c81ee2`](https://github.com/UiPath/skills/pull/3096/commits/7c81ee250eec7e61736adc6de381c3d154e7a8e9). The prior blocker is fixed and the follow-up commit introduces no new blocker, so this is an approve. Everything below is non-blocking.

### Prior-blocker ledger

| Prior finding | Status |
|---|---|
| P1 — raw-API flows cannot observe the redirect target they classify on | **Fixed** |
| Follow-up — test coverage for the new classification and stop behaviour | Still open |
| Follow-up — single home for the remediation wording | Partially addressed; see Standards below |
| Follow-up — 422 branch and get-flow scope vs. PILOT-7430's acceptance case | Still open |

**The P1 is genuinely fixed.** Both flows' first tenant call now carries the redirect target: [`publish-process.md#L12`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/publish-process.md#L12) and [`get-process.md#L12`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/get-process.md#L12)/[`#L22`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/get-process.md#L22). I checked every remaining `curl` in `references/` — `get-process.md` L32/L43 and `publish-process.md` L37/L85/L106/L149/L163 — and each sits behind a step that has already proved a 200, so no raw-API flow can now reach the not-enabled condition through a call that hides the redirect target. The **"Never add `-L`"** rule at [`api-endpoints.md#L158-L168`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/api-endpoints.md#L158-L168) is a real addition on its own — following the redirect is exactly what turned this into the generic JSON-parse failure [PILOT-7430](https://uipath.atlassian.net/browse/PILOT-7430) reported.

## Standards

One documented-standard breach, unchanged in substance from my first review but now stated explicitly in the files, plus one readability note. Neither is merge-gating, so neither blocks.

- **The `Never fork a fact` rule and the new per-transport wording now contradict each other inside `SKILL.md`.** [`SKILL.md#L61`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/SKILL.md#L61) still says a new domain fact gets documented once in `api-endpoints.md` and *"Never fork a fact across the CLI and API files"*, while [`SKILL.md#L71`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/SKILL.md#L71) now says the wording lives *"in one home per transport"* and [`cli-commands.md#L24`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/cli-commands.md#L24) asks readers to *"keep them in sync if the wording ever changes."* To be clear, I think the rationale you wrote there is the right call — the CLI files staying self-contained for the day the raw-API fallback retires is a real constraint, and collapsing five inline copies to two homes is a solid improvement. The problem is only that `SKILL.md#L61` wasn't updated to match, so the skill's own extension rule now reads as forbidding what the skill does. Would you be up for amending L61 with the per-transport carve-out and its retirement rationale, so a future contributor doesn't "fix" the fork by deleting one home? Nothing here has to happen in this PR.
- Nothing enforces the sync, so a wording change is now two edits with no check to catch drift. If that feels worth guarding, a link-checker-style assertion that the two blockquotes match would do it.
- Minor: the `%{redirect_url}` explanation is now paraphrased four times ([`api-endpoints.md#L163`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/api-endpoints.md#L163), [`get-process.md#L15`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/get-process.md#L15) and [`#L25`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/get-process.md#L25), [`publish-process.md#L18`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/publish-process.md#L18)), each slightly differently. `.claude/rules/token-optimization.md` prefers one full statement plus pointers over N terse paraphrases. I'd keep `Never add -L` verbatim at each site — it's a safety rule and that file exempts safety redundancy — and shorten the rest to a pointer.

## Spec

PILOT-7430's two stated fix directions are both met now: *"Detect and explain"* is the classification plus verbatim remediation, and *"Gate earlier"* is publish Step 1's `Do not proceed until you have a 200` running before any write. No scope creep in this commit. Two things worth a follow-up:

- **New in this commit:** [`get-process.md#L15`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/get-process.md#L15) says a bare **404** on the *search* endpoint *"means AH isn't available on this tenant at all"*. The canonical section and [`get-process.md#L27`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/get-process.md#L27) both qualify that — a 404 only signals not-enabled when the body says `not found in organization` — and the publish flow additionally guards with *"Confirm the org/tenant first"* ([`publish-process.md#L22`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/skills/uipath-automationhub/references/publish-process.md#L22)). L15's *"handle it as in Step 2"* does route the reader to the qualified rule, and this flow is read-only, which is why I'm not blocking on it. But a mistyped `$ORG`/`$TENANT` could still get reported as "contact your administrator to enable Automation Hub" — the confidently-wrong remediation this ticket exists to remove. Could we mirror Step 2's `not found in organization` qualifier and publish's org/tenant guard into L15?
- **Test coverage is still open.** Nothing under `tests/` changed in this PR, and [`smoke_publish_process.yaml`](https://github.com/UiPath/skills/blob/c575713f7575221af247abcc39cae561708e2116/tests/tasks/uipath-automationhub/smoke_publish_process.yaml) still instructs the agent not to let a failed connectivity check stop it — grading roughly the opposite of the stop-and-explain behaviour this PR adds. The smokes can stay green without ever exercising the fix. Happy for that to be its own ticket rather than growing this PR.

Also still open from last time: whether the 422 "reachable but not onboarded" branch and the get-flow changes should be reflected in PILOT-7430's acceptance criteria or split into a follow-up, so their source is explicit. That's a product call, not something I'd hold the PR for.

---

**Summary** — Standards: 3 findings, worst is the `SKILL.md#L61` self-contradiction (non-blocking, documentation consistency). Spec: 2 findings, worst is the unqualified 404 at `get-process.md#L15` (non-blocking, read-only flow with a correct pointer to the qualified rule). Prior P1: fixed and verified. CI is green.

parth-patil-uipath:
Fixed in commit 2 but this was only for raw curl that is used as fallback if cli is not avaibale. Fixed and ready for review

magickaichen (approved):

abhiram-vad (approved):

PR #3098, opened 2026-09-04
fix(uipath-maestro-case): fall back from --strict only when the CLI rejects the flag

roberts-cliff:
## Why

Public `@uipath/cli` latest is 1.200.1. `--strict` / `--sdd` exist from 1.202 (dev feed and 1.201 preview). The skill's mandatory Phase 3 exit and closing validate require both flags with no fallback, so a public-latest user hits `invalid_argument` on the close.

## What

- **SKILL.md Rule 14** — version guard: fall back to plain `validate` only when the response is `invalid_argument` naming `--strict` or `--sdd` as an unknown option. Exit 3 alone is insufficient, and a strict *failure* is never a reason to drop the flags. On fallback, log `strict profile unavailable on CLI <version>` in `build-issues.md` and the completion report, and run Step 12 Checks 1–15 by hand.
- **implementation.md** Step 12 (Phase 3 exit and Phase 4) and **phased-execution.md** Phase 4 — one-line pointers to the guard, keeping `STRICT_SDD_FILE_NOT_FOUND` as fix-the-path, not drop-the-flag.

Prose only; no task or fixture changes. `npm run skills:validate` clean.


Yiqi517 (approved):
Makes sense!

charlesliu9 (approved):

PR #3099, opened 2026-09-04
MST-14766: feat(case): number condition names case-wide, not per array

charlesliu9:
fix(uipath-maestro-case): number condition names case-wide, not per array

The four condition plugins scoped the `Entry Rule {N}` / `Complete Rule {N}` /
`Exit Rule {N}` default counter to the local array — `task.entryConditions[]`,
`stageNode.data.entryConditions[]`, `stageNode.data.exitConditions[]`,
`metadata.caseExitRules[]`. The platform enforces those names as one flat
case-wide pool, so a per-array counter restarts at 1 in every stage and on
every task and mints duplicates by construction. Any stage holding two tasks
whose SDD `Display Name` cells are blank hit it: both got `Entry Rule 1`, and
`validate` failed with `Rule name 'Entry Rule 1' is not unique`.

The collision is worse than a failed build because the cheapest way out is
destructive: `validate` downgrades a task left with no entry rules to a
warning, so emptying `entryConditions[]` turns the build green while deleting
authored behaviour. That is how `skill-case-sla-from-sdd` failed — the agent
authored the `adhoc` rule on "Add Claim Evidence" correctly, could not clear
the duplicate name in five attempts, and dropped the entry conditions on two
tasks to reach Valid.

- case-schema.md gains the canonical rule: the pool, the validator's error
  shape (including `nodes[root]` for case-exit rules), the case-wide counter,
  and an explicit ban on repairing a collision by deleting a condition. The
  three condition `displayName` rows point at it.
- The four impl-json.md `displayName` steps state the case-wide scan and name
  the collision each one previously caused.
- The four planning.md rows now separate a semantic SDD name, carried
  verbatim, from a cell already holding the default pattern, which is the SDD
  echoing the default and is renumbered.
- implementation.md adds Step 12 Check 16, and SKILL.md Rule 25 covers
  condition names alongside stage labels and task names.

Verify:
https://github.com/UiPath/skills/actions/runs/33928130824


skills/uipath-maestro-case/references/implementation.md
```diff
@@ -342,6 +342,10 @@ After value bindings (Step 9.8), connector-rule upgrades (Step 10.5), and marker
...
   Re-scan every task after repair; halt before Phase 4 if any task's `entryConditions` is still empty after one repair pass. This check is the mandatory backstop for Step 10 (`plugins/conditions/task-entry-conditions/impl-json.md § Post-Write Verification` only confirms count-parity against the SDD, which is not a safety net when the write itself was skipped) — do not treat Step 10 having run as sufficient evidence this check passes.
 
+- **Check 16 — Condition `displayName` uniqueness** — Collect every condition `displayName` into ONE flat case-wide pool across all four scopes — each stage's `data.entryConditions[]` and `data.exitConditions[]`, each task's `entryConditions[]`, and `metadata.caseExitRules[]` — and compare literal strings, untrimmed and unpartitioned. Any name appearing twice is a failure. `validate` reports this as a hard error, so unlike most Step 12 checks a clean `Valid` IS evidence it passed; run it anyway, because the repair is constrained. Rule and rationale: [case-schema.md § Condition name uniqueness](case-schema.md#condition-name-uniqueness).
```
Yiqi517, line 345: nit: I assume it's trimmed?

Yiqi517 (approved):
Cool!

PR #3100, opened 2026-09-04
ci: make the Tier 1 / Tier 2 required-check set enablable

uipreliga:
Prerequisites from the 2026-09-03 PR CI audit (`c/skills-pr-ci-audit.md`). **No ruleset change here** — a required context whose job has never reported blocks *every* open PR, so the contexts must be seen reporting first. This PR is that evidence, and it now has it (see Verification).

## Why 13 checks weren't requirable

A required context is a literal string GitHub waits for. Three things stopped these from being one:

1. **`paths:` filters.** When the filter excludes a PR the workflow never runs, no check is reported, and the required context stays pending — merge blocks forever. Removed from `test-helpers.yml`, `validate-skills.yml`, `validate-skill-status.yml`, `validate-skill-flavors.yml`, each carrying the rationale `validate-version-sync.yml` and `validate-skills-sh.yml` already use. Jobs that can short-circuit now do so *inside* the job, where a skip counts as a pass.
2. **Expression job names.** `activation-gate.yml` and `verb-gate.yml` name their gate job `${{ matrix.skill }}` — 61 distinct check names in one quarter, so no literal context can match. Added fixed-name aggregators `Skill activation gate` and `CLI verb gate` that roll up `detect` + every leg. Brings ~13,000 runs/quarter of existing coverage under enforcement at no added CI cost.
3. **Check-name collisions.** Two jobs sharing a context make it ambiguous which run satisfied it. `verb-gate.yml`'s detect job is now `Detect changed skills (CLI verb gate)`; `publish.yml`'s push-event guard is `Validate version sync (publish guard)` — the latter could otherwise land a same-named check on the merge commit for the already-required PR check.

## One correction to the audit

It recommended dropping `Detect changed RPA skills` as "a detect job, not a gate." **That would open a hole.** Both smoke workflows gate as `needs: detect` + `if: needs.detect.outputs.skip != 'true'`, so a detect *failure* skips the gated job — and GitHub counts a skipped job as a pass. `Run skill smoke tests` would report green having run nothing.

This run demonstrates it live: both smoke gates came back `skipped` while their detect jobs succeeded. Had detect failed instead, the repo's most expensive gate would have been green. Both detect jobs are kept required — Rule 3 in the new doc, pinned by a test. `Detect changed skills` was never required, so that hole is open on `main` today.

## What holds it stable

`docs/REQUIRED-CHECKS.md` holds the 24-context target set. `scripts/parse-required-checks.py` is its **only** parser — `apply-required-checks.sh` shells out to it and the contract guard imports it, so what gets applied is by construction what got validated. `required-check contract guard` (new job in `test-helpers.yml`) fails if a required job is renamed, has its trigger narrowed, collides on a name, depends on an uncovered job, or cancels a run on a SHA that already has one. `CLAUDE.md` rule 10 carries the same-PR edit rule, following the pattern rule 5 already sets for `skill-status.json` and `skills.sh.json` — CLAUDE.md is loaded into every agent session here, CONTRIBUTING.md is not.

`apply-required-checks.sh` builds its PUT body from the **live** ruleset and authors only the status-check list, so `enforcement`, `bypass_actors` and any `pull_request` parameter changed in the Rules UI survive an apply. `--check` diffs the doc against the live ruleset — the one direction no test can see.

## Review findings fixed

Three HIGH, all reproduced:

- The doc table had **two** parsers (awk + regex) with different tolerances — a row could be applied to the ruleset having never been validated, or applied as a mangled context GitHub waits for forever. Collapsed to one parser that errors on an unreadable row instead of dropping it.
- The Rule 3 guard graded the mere substring `always()`, so gutting an aggregator to `run: echo ok` kept all tests green while the required check reported success regardless. It now asserts a `run:` body consumes each `needs.<job>.result`.
- `|| true` swallowed real failures in two now-required checks (`verb-gate.yml`'s checker crash; `validate-skills.yml`'s failed `git diff` masquerading as "nothing to validate"). Both reported green having validated nothing; both now fail loud.

Plus 5 medium fixes, including `per_page=100` on the doc's pre-apply command (it already truncated at 30 on 3 of 5 open PR heads) and pinning `validate-skill-flavors.yml`'s three actions to SHAs now that it runs on every PR.

A second review round found the `cancel-in-progress` fix below was half right, and that the apply script's literal PUT body would have dropped `enforcement` and both `bypass_actors` from `main` on its first run. Both fixed; per-finding detail in the [reply thread](https://github.com/UiPath/skills/pull/3100#issuecomment-5591039883).

## Verification

**All 24 target contexts reported on this PR, none red.** The aggregators — the one assumption no local tool could test — work: `CLI verb gate` rolled up 27 real matrix legs into a single fixed-name pass, and `Skill activation gate` reported green while its `matrix.skill` placeholder shows `skipped`.

Total CI wall clock ~2 min across 12 workflows (longest: `Test Helpers` 63s with 11 unconditional jobs; `Validate Skill Flavors` 52s). Note this was the *heavy* case — the merge touched `verb-gate.yml`, so detect re-gated all 27 skills. A typical PR is lighter.

Locally: 104 contract tests + full `tests/scripts/` suite pass · actionlint and shellcheck clean on all new code · parser and guard emit byte-identical 24-context lists · repo validators green. Twelve negative controls confirm the new assertions fire: gutted aggregator, `paths-ignore`, dropped `synchronize`, dropped `ready_for_review`, half-backticked table row (fails the test *and* blocks the apply script), an unexpanded `${{ }}` context, a reverted `always()` guard, an inverted `${{ cancelled() }}` guard, a re-gutted rollup under `!cancelled()`, an unnamed job in `test-helpers.yml`, a reverted `cancel-in-progress: true`, and the dropped `maestro-bpmn` row.

### Two defects the PR's own CI then exposed

The paragraph above used to end by admitting no aggregator had yet seen a cancelled run. It has now, and it was wrong.

**1. A superseded run turned a required context red.** Commit `1056e6c56` carried a red `Skill activation gate` next to a green one. Flipping the PR out of draft fired `activation-gate.yml` again (it triggers on `ready_for_review`) and `cancel-in-progress` killed the in-flight run *on the same commit*. Under `if: always()` the cancelled run's aggregator still executed, read `detect: cancelled`, and failed. GitHub keeps the latest check run per context name, so merge would have turned on which of the two finished last — the good one won by ~70s (runs `34055252280` / `34055306035`).

My first fix was `if: ${{ !cancelled() }}`, on the assumption that a job skipped during cancellation reports `skipped`, which passes. **It does not.** Dispatching the workflow and cancelling it mid-flight reported `cancelled  Skill activation gate` (run `34067126301`), and `cancelled` blocks exactly like `failure`.

The cause is two runs on one SHA, so the fix is in the concurrency block. My second attempt deny-listed `ready_for_review` and claimed it was the only event that fires on a commit a run is already in flight for. **Also wrong** — `reopened` is a default `pull_request` type and closing a PR does not cancel its runs, so close-and-reopen mid-run reproduces the same race. It is now an allow-list:

```yaml
cancel-in-progress: ${{ github.event.action == 'synchronize' }}
```

`synchronize` is the only `pull_request` event that arrives with a new head SHA, so it needs no re-audit when GitHub adds an activity type. Applied to all six required workflows carrying `cancel-in-progress` — the deny-list left the hazard live on five of them — and pinned by `test_concurrency_only_cancels_on_a_new_head_sha`. Documented as Rule 3a, along with the fact that a superseded red self-heals on a re-run and needs no ruleset edit.

`!cancelled()` stays on both aggregators for what it actually does: resolves them the instant a run is cancelled rather than after a runner acquisition, and keeps a bogus "could not determine which skills to gate" error out of a superseded run's log.

**Verified by reproducing it.** With a run in flight on `e1c1f1fda`, I flipped the PR to draft and back. Both runs (`34067608820`, `34067674271`) completed `success`; neither cancelled the other. Same scenario, previously one red and one green.

**2. The rebase silently dropped a required context.** `maestro-bpmn checker unit tests` came from `main` and its table row lived only in a merge commit, which rebasing discards. 24 contexts became 23 and all 87 tests still passed — every test read the doc and checked it against the workflows, so a job *missing* from the table was invisible by construction. `test_every_test_helpers_job_is_required` now closes that direction for `test-helpers.yml`, the one workflow where every job is meant to be required. Not generalised to other workflows, where advisory jobs legitimately sit outside the set (Rule 4).

**Residual risk:** no aggregator has yet seen `detect` fail or a leg fail on a live run. Those paths remain unit-tested against the documented Actions contract rather than exercised in anger — which is precisely the category the two defects above came from, so weigh it accordingly.

## Follow-up, not in this PR

1. Apply the ruleset once merged: `./scripts/apply-required-checks.sh --dry-run`, then without the flag. Confirm with `--check`.
2. Extend gating to `release/*` (~92 PRs/quarter, mostly cherry-picks, currently merging on review alone): `--with-release-branches`.
3. Schedule `--check`. It needs a PAT with repo admin — `GITHUB_TOKEN` cannot read a ruleset.
4. **`Validate task schema (advisory)` passed on this run** under the pinned engine, and a local sweep parsed 1,288/1,290 task YAMLs. The audit's blocker for requiring it (65 merged-red from tree drift) looks stale — dropping `continue-on-error` and the `(advisory)` suffix is now a small follow-up rather than a cleanup project.
5. Hard-fail the activation gate when a PR adds a skill with no baseline (5 skills are ungated today; `uipath-process-mining` needs its dataset topped up from 19 rows first), re-baseline on `claude-sonnet-5`, and fix the `httpx` install failure — all from @bai-uipath's review.
6. Fold the two byte-identical rollup bodies into a composite action, against a green baseline.
7. Consider `sha_pinning_required` on the repo (currently `false`); `publish-skill-flavor.yml` would need pinning first.


bai-uipath (approved):
**Approve.** The fixed-name aggregators and the same-SHA cancellation fix are solid. One gap I'd close in the activation gate before this set goes live, the rest are yours to take or leave.

**The activation gate silently passes any skill with no baseline.** It returns success on the no-baseline path, so five of the twenty-seven skills are ungated today, and two of those (`uipath-aops`, `uipath-automationhub`) have no activation dataset at all. A PR that introduces a skill clears the gate having measured nothing. Fix: hard fail when a PR adds a skill without a baseline entry and an activation dataset of at least 20 examples in the required format. Note `uipath-process-mining` currently sits at 19 rows, so that threshold needs a top-up alongside it.

Fix the rest if you agree, none are blockers.

**Baselines are stale.** They were measured on claude-sonnet-4-6, the model variable has been claude-sonnet-5 since July, and only planner was re-measured. Worth re-baselining before this becomes a merge gate rather than after.

**The gate has a live install failure.** The host install takes the bare wheel and the llm_judge criterion fails to import httpx, erroring every row before anything is measured. It blocked one branch eight times over four days in early September. The fix belongs in the install line or in coder_eval's dependencies.

**A bypass actor before applying the ruleset.** All three rulesets have none, admins included, so a single broken required context locks the repo and the fix-forward PR with it. Cheap insurance when going from 6 contexts to 24.

Minor: the apply script rebuilds every rule rather than patching just the status-check list, dropping two currently-empty pull_request fields as it goes; `--dry-run` prints a count line ahead of the payload so the output isn't pipeable; the doc calls the aggregators `always()` where they are `!cancelled()`; validate-skill-flavors is the only GitHub-hosted runner in the set and now runs on every PR.

.github/workflows/activation-gate.yml
```diff
@@ -7,7 +7,19 @@ name: Skill Activation Gate
...
+  # run reports the required `Skill activation gate` context as `cancelled` —
+  # non-passing, exactly like `failure`. Merge then turns on whether the good
+  # run or the cancelled one recorded its check last (GitHub keeps the latest
+  # per context name). Measured on PR #3100: runs 34055252280 and 34055306035,
+  # same commit, one red and one green, ~70s apart.
+  #
+  # Letting the ready_for_review run coexist costs one extra run per draft flip.
+  # That run is cheap: `gate` skips on drafts, so the in-flight run it would
+  # have cancelled is a detect plus a skipped matrix.
+  cancel-in-progress: ${{ github.event.action != 'ready_for_review' }}
```
rockymadden: **Major.** `ready_for_review` is not the only same-SHA event. `reopened` is too, and it is already in this workflow's `types:` on line 30.

Closing a PR does not cancel its in-flight runs. So: push, run starts, close, reopen. The `reopened` run fires on the unchanged head SHA, `cancel-in-progress` evaluates true (the action is not `ready_for_review`), and it kills the in-flight run. You get exactly the red-`cancelled`-beside-green that this line exists to prevent.

```yaml
cancel-in-progress: ${{ github.event.action != 'ready_for_review' && github.event.action != 'reopened' }}
```

Worth stating in Rule 3a either way: a superseded run's red is cleared by re-running that run. Without that written down the next person chases it as a ruleset bug.

docs/REQUIRED-CHECKS.md
```diff
@@ -0,0 +1,174 @@
...
+`1056e6c56` carried a red `Skill activation gate` beside a green one (runs
+`34055252280` and `34055306035`, ~70s apart).
+
+The fix is in the concurrency block, not the job:
+
+```yaml
+cancel-in-progress: ${{ github.event.action != 'ready_for_review' }}
+```
+
+`ready_for_review` is the only event that fires on a commit a run is already in
```
rockymadden: **Major.** This claim is false, and it is the load-bearing sentence of Rule 3a.

`reopened` also fires on a commit a run is already in flight for, and it is a **default** `pull_request` type. So the hazard is live on five other required workflows that still carry `cancel-in-progress: true`:

- `smoke-skills.yml`
- `smoke-rpa-skills.yml`
- `task-driver-gate.yml`
- `verb-gate.yml`
- `validate-version-sync.yml`

For those five the pragmatic answer is probably not to touch the workflows, but the doc has to say so, and has to say a superseded red self-heals on a re-run. Right now it asserts an invariant a maintainer will build on.

scripts/apply-required-checks.sh
```diff
@@ -0,0 +1,141 @@
...
+COUNT=$(printf '%s\n' "$CONTEXTS" | wc -l | tr -d ' ')
+
+if [ "$WITH_RELEASE" -eq 1 ]; then
+  REF_INCLUDE='["~DEFAULT_BRANCH", "refs/heads/release/*"]'
+else
+  REF_INCLUDE='["~DEFAULT_BRANCH"]'
+fi
+
+PAYLOAD=$(
+  printf '%s\n' "$CONTEXTS" | jq -R . | jq -s \
```
rockymadden: **Major, and the one item with real blast radius.** This payload sends only `conditions` and `rules`. The live ruleset also carries:

```json
"enforcement": "active",
"bypass_actors": [
  {"actor_type": "OrganizationAdmin", "bypass_mode": "always"},
  {"actor_id": 5, "actor_type": "RepositoryRole", "bypass_mode": "always"}
]
```

GitHub's behavior on omitted fields for `PUT /repos/{owner}/{repo}/rulesets/{id}` is not well documented and has changed. If omitted means cleared, the first real run drops admin bypass on `main`. This is follow-up step 1, so it fires immediately after merge.

Build the body from current state instead of a literal:

```bash
CURRENT=$(gh api "repos/$REPO/rulesets/$RULESET_ID")
PAYLOAD=$(jq --argjson checks "$CHECKS_JSON" \
  '{name, target, enforcement, bypass_actors, conditions, rules} | ...' <<<"$CURRENT")
```

Same argument for the hardcoded `pull_request` parameters below. They match live today (I diffed them), but a review-count change made in the UI gets silently reverted the next time anyone runs this.

docs/REQUIRED-CHECKS.md
```diff
@@ -0,0 +1,174 @@
...
+| `catalog build integrity guards` | `test-helpers.yml` |
+| `skills.sh grouping checker unit tests` | `test-helpers.yml` |
+| `telemetry hook contract guard` | `test-helpers.yml` |
+| `required-check contract guard` | `test-helpers.yml` |
+| `Skill activation gate` | `activation-gate.yml` (aggregator) |
+| `CLI verb gate` | `verb-gate.yml` (aggregator) |
+
+Deliberately **not** required: `Detect changed skills (CLI verb gate)` and `detect` in `activation-gate.yml`. Both are covered by their workflow's `always()` aggregator (Rule 3), so requiring them adds nothing.
+
+## Applying a change
```
rockymadden, line 171: **Major.** Nothing detects drift in the doc-to-ruleset direction.

Every test reads this doc and validates it against the workflows. Nothing compares it to the live ruleset. So:

- Delete a table row and its job together: guard green, ruleset keeps a context that no longer reports, **every PR blocks**.
- Edit the ruleset in the UI: doc lies, guard green.

That is the precise failure this PR exists to prevent, left open on the one edge no test covers. `--list` prints the live set but nothing compares.

Add `--check` to the script (exit non-zero on any diff between `parse-required-checks.py --contexts` and the live contexts) and run it from a scheduled workflow. Cheap, and it closes the loop.

tests/scripts/test_required_checks_contract.py
```diff
@@ -0,0 +1,292 @@
...
+    it exists to hold cheap, deterministic, always-run guards, and each is a
+    checkout plus a sub-30s pytest. Anything nondeterministic or advisory
+    belongs in another workflow (Rule 4), so there is no legitimate reason for
+    a job here to sit outside the required set.
+    """
+    data, _ = load_workflow("test-helpers.yml")
+    names = {job.get("name") for job in (data.get("jobs") or {}).values()}
+    required = {c for c, _ in TARGET_SET}
+
+    unregistered = sorted(n for n in names if n and n not in required)
```
rockymadden: **Minor.** `if n` filters out jobs with no `name:` key, and their check name is the job id, which is a real unrequired context. So the reverse-direction guard has a hole in exactly the direction it was written to close.

Confirmed: appending this to `test-helpers.yml` keeps all 88 tests green.

```yaml
  sneaky-unnamed-job:
    runs-on: ubuntu-latest
    steps:
      - run: echo hi
```

Fix at line 282:

```python
names = {job.get("name") or job_id for job_id, job in (data.get("jobs") or {}).items()}
```

tests/scripts/test_required_checks_contract.py
```diff
@@ -0,0 +1,292 @@
...
+        f"If that job fails, {context!r} is skipped — which GitHub counts as a "
+        f"pass. See docs/REQUIRED-CHECKS.md Rule 3."
+    )
+
+
+def test_no_two_jobs_share_a_required_context():
+    """A context matching two jobs is ambiguous about which run satisfied it."""
+    required = {c for c, _ in TARGET_SET}
+    owners = {}
+    for path in sorted(WORKFLOWS.glob("*.yml")):
```
rockymadden: **Minor.** `*.yml` only. A `.yaml` workflow is invisible to the collision check. None exist today, so this is a one-character hardening: `glob("*.y*ml")`.

.github/workflows/verb-gate.yml
```diff
@@ -114,3 +125,70 @@ jobs:
...
+    # required context would never report.
+    #
+    # `!cancelled()` rather than `always()`: `always()` also runs during RUN
+    # cancellation, so a superseded run executed this job and reported the
+    # required context FAILED off a `detect: cancelled` it was never meant to
+    # judge. `!cancelled()` does not make such a run pass — GitHub reports the
+    # skipped job as `cancelled` — it just resolves it immediately and without
+    # a misleading error. See the same block in activation-gate.yml.
+    #
+    # No `ready_for_review` in this workflow's triggers, so it has no event that
```
rockymadden: **Major (same root cause as the `activation-gate.yml` note).** This is wrong. `reopened` is a default `pull_request` type, so it is in this workflow's triggers, and it fires on a SHA a run can already be in flight for. Combined with this workflow's `cancel-in-progress: true` on line 28, `CLI verb gate` can go `cancelled` beside a green one on the same commit.

Either extend the conditional `cancel-in-progress` here too, or replace this paragraph with the honest version: the hazard exists, and a superseded red is cleared by re-running the run.

.github/workflows/verb-gate.yml
```diff
@@ -114,3 +125,70 @@ jobs:
...
+  # Fixed-name aggregator. The `gate` job above is named `${{ matrix.skill }}`,
+  # so its check name changes with every skill added or renamed. Required-
+  # status-check contexts are literal strings, so a matrix leg can never be
+  # required: a name that doesn't run leaves the context pending forever. This
+  # job's name is constant, so require THIS one. It rolls up `detect` + every
+  # `gate` leg into a single pass/fail.
+  gate-summary:
+    name: CLI verb gate
+    needs: [detect, gate]
+    # Run regardless of what `detect` and `gate` did — without this the summary
```
rockymadden: **Minor.** The `!cancelled()` / Rule 3a rationale is now written out four times: here (~20 lines), in `activation-gate.yml` (~35 lines), in `docs/REQUIRED-CHECKS.md`, and in this test module's docstring. Four copies of a subtle argument will drift, and the drifted copy is the one someone reads at 2am.

Make the doc canonical and cut these to one line plus a pointer. Related: a 13-line comment on the one-line `cancel-in-progress:` in `activation-gate.yml` is a lot against this repo's convention of minimal comments.

.github/workflows/verb-gate.yml
```diff
@@ -114,3 +125,70 @@ jobs:
...
+    #
+    # No `ready_for_review` in this workflow's triggers, so it has no event that
+    # starts a second run on a SHA already in flight; the conditional
+    # `cancel-in-progress` activation-gate.yml needs has nothing to do here.
+    # The `${{ }}` wrapper is required: bare `!` starts a YAML tag.
+    if: ${{ !cancelled() }}
+    runs-on: uipath-ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Roll up gate results
```
rockymadden, line 151: **Minor.** This rollup body is byte-identical to `activation-gate.yml`'s, ~45 lines each, differing only in the two error strings. A composite action would keep them in sync, and `test_required_job_needs_are_covered`'s `needs.<job>.result` assertion still works if the results are passed through `env:` at the call site.

.github/workflows/validate-skill-flavors.yml
```diff
@@ -40,9 +24,9 @@ jobs:
     name: Build and inspect every skill package
     runs-on: ubuntu-latest
```
rockymadden, line 25: **Minor.** This is the only required workflow on GitHub-hosted runners; every other one is `uipath-ubuntu-latest`. Dropping the paths filter puts a ~52s node build on billed minutes for every PR, including doc-only ones.

Either move it to the self-hosted pool or note in the description that the cost was weighed and accepted. The rest of the paths-filter removals land on self-hosted, so this one is a different trade.

rockymadden (approved):
Approving. The design is right and I verified it end to end rather than taking the description's word for it.

**Verified at `e1c1f1f`:**

| Check | Result |
|---|---|
| `pytest tests/scripts/test_required_checks_contract.py` | 88 passed |
| All 24 target contexts reported on the head SHA | Yes, none red (2 `skipped` = pass) |
| `apply-required-checks.sh --dry-run` | Valid JSON, 24 contexts |
| Live ruleset rule types vs script payload | Exact match (`deletion`, `non_fast_forward`, `pull_request`, `required_status_checks`) |
| `integration_id: 15368` vs live | Matches |
| `test-helpers.yml` job set vs `origin/main` | No drift, so no repeat of the dropped-row incident |

The Rule 3 correction to the audit is the strongest part of this. `if: needs.detect.outputs.skip != 'true'` plus a detect failure does skip the gated job, and a skipped job counts as a pass, so dropping `Detect changed RPA skills` would have opened a hole. That hole is open on `main` today for `Detect changed skills`.

Inline: 3 major, 5 minor. None block the merge. **One of them blocks the follow-up:** do not run `apply-required-checks.sh` until the `bypass_actors` question on `scripts/apply-required-checks.sh` is settled, since that is step 1 of the plan and it can silently drop admin bypass on `main`.

---

**Two things outside the diff:**

**The PR is `CONFLICTING`.** One conflict, in `.github/workflows/validate-skill-flavors.yml`: `main` added `tests/scripts/studioweb-flavor-contract.test.mjs` to the paths list this PR deletes. Resolve by keeping the deletion. Given the rebase incident described in the body, re-run the contract tests after resolving rather than assuming.

**Trim the description before merge.** "8-axis review with adversarial verification (47 agents, 11 findings refuted and dropped). Scored 9.2/10." tells a reviewer nothing and ages badly. The findings matter; the agent count does not. The Verification and "Two defects the PR's own CI then exposed" sections earn their space; the process narration does not. Same problem at smaller scale in the code comments (see the inline note on `verb-gate.yml`).

.github/workflows/activation-gate.yml
uipreliga: Fixed, and you were right that the deny-list was the wrong shape. Replaced with an allow-list:

```yaml
cancel-in-progress: ${{ github.event.action == 'synchronize' }}
```

`synchronize` is the only event that arrives with a new head SHA, so it needs no re-audit when GitHub adds an activity type. Applied to all six required workflows that carry `cancel-in-progress`, not just this one — see the reply on the REQUIRED-CHECKS.md thread.

`test_concurrency_only_cancels_on_a_new_head_sha` now enforces it over every workflow in the target set, with a negative control asserting the old `!= 'ready_for_review'` form is rejected. Verified it fails when I flip verb-gate.yml back to `true`.

Rule 3a also now states the self-heal: a superseded red is not a ruleset bug, `gh run rerun <id>` replaces it, since GitHub keeps only the latest check run per context name.

docs/REQUIRED-CHECKS.md
uipreliga: Fixed, and I went further than documenting it — the five are one line each and the change only ever costs an extra run, so all six required workflows carrying `cancel-in-progress` now use `${{ github.event.action == 'synchronize' }}`: `activation-gate.yml`, `verb-gate.yml`, `smoke-skills.yml`, `smoke-rpa-skills.yml`, `task-driver-gate.yml`, `validate-version-sync.yml`. (The rest of the required set has no `concurrency:` block at all.)

Rule 3a is rewritten around that: the allow-list rationale, a table of the three same-SHA paths (`ready_for_review`, `reopened`, `opened` against a dispatch), an explicit note that the earlier "only same-SHA event" claim was wrong, and the re-run self-heal. `test_concurrency_only_cancels_on_a_new_head_sha` keeps the doc honest.

scripts/apply-required-checks.sh
uipreliga: Fixed — and confirmed live, so this was not hypothetical. `gh api repos/UiPath/skills/rulesets/14795269` carries:

```json
"enforcement": "active",
"bypass_actors": [
  {"actor_id": null, "actor_type": "OrganizationAdmin", "bypass_mode": "always"},
  {"actor_id": 5, "actor_type": "RepositoryRole", "bypass_mode": "always"}
]
```

plus `required_reviewers: []` and `dismissal_restriction` inside `pull_request`, which the hardcoded block dropped too.

The body is now built from the live ruleset — `{name, target, enforcement, bypass_actors, conditions, rules}` — and only the `required_status_checks` list is authored here. Every other rule and parameter is carried through, so a review-count change made in the UI survives. `--with-release-branches` unions `refs/heads/release/*` into the live include list rather than replacing it.

`--dry-run` still works without repo admin: it falls back to a labelled skeleton and prints to stderr that the preserved fields are absent from the preview. Verified `--dry-run` against the real ruleset — the two bypass actors come through.

So follow-up step 1 is unblocked from my side.

docs/REQUIRED-CHECKS.md
uipreliga: Fixed. Added `--check`: it diffs `parse-required-checks.py --contexts` against the live contexts and exits non-zero, labelling each direction (`<` documented but not enforced; `>` enforced and never validated against the workflows). Both sides sort under `LC_ALL=C` so jq's codepoint order does not read as drift.

Not scheduled, and I did not want to ship a cron that fails every night: reading a ruleset needs admin on the repository, which the default `GITHUB_TOKEN` does not have. That is now an Open item in the doc — wiring it to `schedule:` needs a PAT in a secret. Until then the doc says to run it after any change to the table, and CLAUDE.md rule 10 lists it alongside the contract guard.

tests/scripts/test_required_checks_contract.py
uipreliga: Fixed exactly as suggested, hoisted into a shared helper since three call sites had the same hole:

```python
def check_names(data):
    return {job.get("name") or job_id for job_id, job in (data.get("jobs") or {}).items()}
```

Used by `test_context_matches_a_job_name`, `test_no_two_jobs_share_a_required_context` and `test_every_test_helpers_job_is_required`. Verified with your `sneaky-unnamed-job` fixture — it now fails `test_every_test_helpers_job_is_required` — and there is a unit-level negative control so the behaviour is pinned without a workflow edit.

tests/scripts/test_required_checks_contract.py
uipreliga: Fixed: `glob("*.y*ml")`.

.github/workflows/verb-gate.yml
uipreliga: Fixed by extending the conditional rather than documenting the hazard — `cancel-in-progress: ${{ github.event.action == 'synchronize' }}` here too. The paragraph claiming this workflow needed none of it is gone; the comment is now two lines pointing at Rule 3a.

.github/workflows/verb-gate.yml
uipreliga: Fixed. `docs/REQUIRED-CHECKS.md` § Why `!cancelled()` and not `always()` is canonical; both aggregators now carry four lines that state the choice and the `${{ }}` YAML-tag gotcha and point there. The 13-line comment on `cancel-in-progress` is down to seven, and those cover the allow-list reasoning rather than restating the doc.

That leaves two copies (doc + test module docstring) instead of four, and the test docstring is the one a failure message sends you to.

.github/workflows/verb-gate.yml
uipreliga: Taking this one as a follow-up rather than in this PR. You are right that it is real duplication, but a composite action moves the rollup out of the file that `test_required_job_needs_are_covered` reads, and that test just grew a guard requiring each `needs.<job>.result` to reach a `run:` body — passing results through `env:` at the call site keeps the assertion working but only because the assertion is written to look there. Rewriting the plumbing and the guard that checks it in the PR that turns them into merge gates is the wrong order. Happy to open it against a green baseline.

.github/workflows/validate-skill-flavors.yml
uipreliga: Documented rather than moved. `npm run skills:pack` plus the marker sweep uses `rg` and `sha512sum`, and I have not verified the self-hosted image carries either — swapping the runner on the one required workflow whose job is "prove the artifact that ships to consumers builds" is not a change I want to make blind in this PR.

Added to Open items instead: the ~52s on billed minutes for every PR, doc-only ones included, the reason it is accepted (the 28-entry path list it replaced went stale each time a build input was added), and "move it to the self-hosted pool once that pool is verified to carry `rg` and `sha512sum`."

tests/scripts/test_required_checks_contract.py
```diff
@@ -0,0 +1,292 @@
...
+            f"pass. See docs/REQUIRED-CHECKS.md Rule 1."
+        )
+
+        # A `types:` list is legitimate (activation-gate.yml and
+        # path-to-ga-approval.yml both add `ready_for_review`), but it must
+        # still cover the three events that open a PR and push to it —
+        # dropping `synchronize` in particular leaves a stale pass standing.
+        types = cfg.get("types")
+        if types is not None:
+            missing = REQUIRED_PR_TYPES - set(types)
```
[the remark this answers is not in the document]
uipreliga: Fixed. Added a per-workflow invariant rather than widening `REQUIRED_PR_TYPES` for everyone:

```python
WORKFLOW_EXTRA_PR_TYPES = {"activation-gate.yml": {"ready_for_review"}}
```

The check also handles an absent `types:` key now — the default set is `opened, synchronize, reopened`, so omitting the list entirely is just as much a way to lose `ready_for_review`. Verified: deleting `ready_for_review` from the trigger fails `test_workflow_has_no_trigger_filter[activation-gate.yml]`.

tests/scripts/test_required_checks_contract.py
```diff
@@ -0,0 +1,292 @@
...
+        # resolve the aggregator immediately rather than after a runner
+        # acquisition, and drops the misleading error.
+        pytest.fail(
+            f"{workflow}: required job {context!r} guards its rollup with "
+            f"`always()`, which also runs during run cancellation. Use "
+            f"`if: ${{{{ !cancelled() }}}}` instead — see "
+            f"docs/REQUIRED-CHECKS.md Rule 3."
+        )
+
+    if "cancelled()" in condition:
```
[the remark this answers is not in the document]
uipreliga: Fixed. Matched against the documented expression exactly:

```python
NOT_CANCELLED_RE = re.compile(r"^\s*\$\{\{\s*!\s*cancelled\(\)\s*\}\}\s*$")
```

Anchored, so bare `${{ cancelled() }}` and `${{ cancelled() || failure() }}` are rejected. Negative control in `test_bare_cancelled_is_not_accepted_as_the_rollup_condition`, and verified end to end by flipping the real aggregator.

tests/scripts/test_required_checks_contract.py
```diff
@@ -0,0 +1,292 @@
...
+        # `!cancelled()` alone proves nothing — a job that runs and never
+        # inspects its needs reports green whatever they did, which is the exact
+        # Rule 3 failure. Demand that the job actually reads each one's result.
+        body = " ".join(
+            str(step.get("run", "")) + " " + str(step.get("env", ""))
+            for step in (job.get("steps") or [])
+        )
+        unread = [key for key in needs if f"needs.{key}.result" not in body]
```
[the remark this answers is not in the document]
uipreliga: Fixed. Consumption now means a `run:` body references the result — directly, or through an `env:` var bound to it. Extracted as `unconsumed_needs(job, needs)` so it has negative controls:

- full aggregator → `[]`
- same `env:` block, `run: echo ok` → `["detect"]`
- inline `${{ needs.gate.result }}` in a `run:` → `[]`

Verified against the real workflow too: replacing the rollup step with `run: echo ok` fails `test_required_job_needs_are_covered[Skill activation gate]`.

I did not go as far as proving the shell logic can actually produce a failure — that is a reachability question a regex should not be answering. The composite-action suggestion on the other thread is the honest fix for that and is queued as a follow-up.

scripts/apply-required-checks.sh
```diff
@@ -0,0 +1,141 @@
...
+
+if [ "$DRY_RUN" -eq 1 ]; then
+  echo "$PAYLOAD"
+  exit 0
+fi
+
+# Warn about contexts that have never been reported on the default branch's most
+# recent PR heads — the failure mode that blocks every open PR.
+echo "Checking that each context has reported recently..."
+RECENT_SHAS=$(gh api "repos/$REPO/pulls?state=open&per_page=10" --jq '.[].head.sha' || true)
```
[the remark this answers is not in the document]
uipreliga: Fixed on both. `gh api ... pulls?state=open` and the per-SHA check-run reads now abort with a message instead of `|| true`, and the "no open PRs" case is a distinct, explicit branch that says it is applying without the reported-recently check. An auth failure can no longer look like an empty PR list.

docs/REQUIRED-CHECKS.md
```diff
@@ -0,0 +1,174 @@
...
+| `uipath-admin verify negative controls` | `test-helpers.yml` |
+| `runtime-payload key-casing contract guard` | `test-helpers.yml` |
+| `catalog build integrity guards` | `test-helpers.yml` |
+| `skills.sh grouping checker unit tests` | `test-helpers.yml` |
+| `telemetry hook contract guard` | `test-helpers.yml` |
+| `required-check contract guard` | `test-helpers.yml` |
+| `Skill activation gate` | `activation-gate.yml` (aggregator) |
+| `CLI verb gate` | `verb-gate.yml` (aggregator) |
+
+Deliberately **not** required: `Detect changed skills (CLI verb gate)` and `detect` in `activation-gate.yml`. Both are covered by their workflow's `always()` aggregator (Rule 3), so requiring them adds nothing.
```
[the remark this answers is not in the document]
uipreliga: Fixed — that sentence now reads "rolled up by their workflow's `!cancelled()` aggregator (Rule 3)".

.github/workflows/activation-gate.yml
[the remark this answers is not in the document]
uipreliga: Fixed — restricted to `synchronize`, as an allow-list: `cancel-in-progress: ${{ github.event.action == 'synchronize' }}`. Applied to all six required workflows carrying `cancel-in-progress`, and enforced by `test_concurrency_only_cancels_on_a_new_head_sha`.

.github/workflows/verb-gate.yml
```diff
@@ -114,3 +125,70 @@ jobs:
...
+    # `!cancelled()` rather than `always()`: `always()` also runs during RUN
+    # cancellation, so a superseded run executed this job and reported the
+    # required context FAILED off a `detect: cancelled` it was never meant to
+    # judge. `!cancelled()` does not make such a run pass — GitHub reports the
+    # skipped job as `cancelled` — it just resolves it immediately and without
+    # a misleading error. See the same block in activation-gate.yml.
+    #
+    # No `ready_for_review` in this workflow's triggers, so it has no event that
+    # starts a second run on a SHA already in flight; the conditional
+    # `cancel-in-progress` activation-gate.yml needs has nothing to do here.
```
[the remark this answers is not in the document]
uipreliga: Fixed — same allow-list here: `cancel-in-progress: ${{ github.event.action == 'synchronize' }}`. The comment claiming this workflow had no same-SHA event is gone.

uipreliga:
Pushed `861006a4a` (on top of a merge of `main`). Inline replies on each thread; summary of what changed and what I deliberately did not do.

### Fixed

| Finding | Change |
|---|---|
| `reopened` is also a same-SHA event (@rockymadden ×2, Copilot ×2) | Deny-list → allow-list `cancel-in-progress: ${{ github.event.action == 'synchronize' }}` on **all six** required workflows carrying it. New guard `test_concurrency_only_cancels_on_a_new_head_sha` enforces it. Rule 3a rewritten: same-SHA event table, the retracted claim called out, and the re-run self-heal |
| Ruleset payload drops `enforcement` / `bypass_actors` (@rockymadden, @bai-uipath) | Body now built from the **live** ruleset; only the status-check list is authored. Confirmed live carries both bypass actors plus `required_reviewers` and `dismissal_restriction` the literal `pull_request` block also dropped |
| Nothing detects doc→ruleset drift (@rockymadden) | `--check`, exits non-zero on any difference, labelled by direction |
| API failure indistinguishable from "no open PRs" (Copilot) | Fails closed on both reads; "no open PRs" is now an explicit branch |
| `--dry-run` count line breaks piping (@bai-uipath) | Moved to stderr; stdout is a pipeable JSON document |
| Contract-guard holes (@rockymadden ×2, Copilot ×4) | Unnamed jobs (`name` → job id), inverted `${{ cancelled() }}`, `env:`-only "consumption", `*.yml`-only glob, missing `ready_for_review` invariant, `always()` in the failure message — each with a negative control |
| `always()` aggregator wording (@bai-uipath, Copilot) | Corrected |
| Rationale duplicated four ways (@rockymadden) | Doc canonical; both aggregators cut to four lines and a pointer |
| Merge conflict in `validate-skill-flavors.yml` (@rockymadden) | Resolved keeping the deletion; contract tests re-run after, not assumed |

104 tests pass, up from 88. Every new guard was verified by breaking the real workflow and watching the named test fail.

### Not doing here

- **Composite action for the two rollup bodies** (@rockymadden). Real duplication, but it moves the rollup out of the file `test_required_job_needs_are_covered` reads, in the PR that turns that test into a merge gate. Follow-up against a green baseline.
- **`validate-skill-flavors.yml` on GitHub-hosted minutes** (@rockymadden, @bai-uipath). Cost accepted and written into Open items with the reason, rather than swapping the runner blind — the job needs `rg` and `sha512sum` and I have not verified the self-hosted image carries them.
- **`--check` on a schedule.** Reading a ruleset needs repo admin; `GITHUB_TOKEN` does not have it. Open item — it needs a PAT in a secret, and a cron that fails every night is worse than none.
- **The activation gate passing skills with no baseline** (@bai-uipath). Agreed it is a real hole and the strongest of your points, but hard-failing a PR that adds a skill is a behaviour change beyond this PR's scope, and it needs `uipath-process-mining` topped up from 19 rows first. Filing separately alongside the stale baselines (still on `claude-sonnet-4-6`) and the `httpx` install failure.
- **Bypass actor before applying** (@bai-uipath). The two on the `main` ruleset are now preserved rather than dropped, which was the actual risk. Whether the other two rulesets should gain one is a separate call.

Description trimmed as suggested.

PR #3103, opened 2026-09-05
fix(uipath-rpa): make the report turn one message and chain the output check into the gate

danmunteanu1:
## Why

In a 30-run headless bench of `uipath-rpa` (5 journeys x 3 replicates x 2 skill arms, Opus 5), the turns spent **after the last `validate`/`build`** were 4-12 per run on every journey and both arms: 60 `run`/`debug` calls, 54 shell checks, 46 memory-file `Write`s, 20 `project-context.md` writes, 14 `AGENTS.md` writes plus 9 shell edits of it, and 29 `Read`/`Grep` calls back into the skill's own references. Turn-by-turn shape of a typical greenfield XAML run after the gate:

```
RUN > RUN > RUN > W:project-context > W:AGENTS > W:memory > W:memory > text
```

Every one of these is a separate assistant message, i.e. a full context replay. Cost tracks message count (~10 s and one replay per message), so this tail was the largest fixed overhead on the short journeys: greenfield XAML took 15-18 messages end to end against the map's 5-turn budget, and 6-8 of the excess were here.

Three causes, all in the map's report row and the memory section:

1. **Output verification as its own turns.** The T4 row said "check the T3 run's actual outputs", so agents re-ran the workflow and `cat`/`ls`/`Test-Path`'d the result file in separate turns (the `--output-filter` retries from #3102 added more).
2. **Context files written in serial turns.** `project-context.md`, then `AGENTS.md`, then a shell `cat >> AGENTS.md`, often preceded by a `Grep` + `Read` of `environment-setup.md § Project Context Discovery` to recover the marker names and the metadata-comment format (29 reference re-reads at wrap-up).
3. **Memory saves on runs that qualify nothing.** 22 runs updated a memory index plus 24 topic files; the topics were dominated by `uip rpa run` payload/flag notes (`uip-rpa-run-output-filter-null-errors.md`, `uip-rpa-run-verdict-shape.md`) on first-try-clean, card-only builds. The § Cross-session memory rules never said the save happens inside the report message, nor that a clean gate on card-covered activities qualifies nothing.

## Fix (docs only)

`execution-maps-guide.md`:

- **Gate rows (greenfield XAML T3, brownfield XAML T3, greenfield coded T3, UIA debug cycle):** the chain ends with `run --skip-build` followed by `cat` of every file the workflow was asked to write. "The output check is a link of this chain, never its own turn." § Gate ≠ runtime proof updated to match.
- **Report rows (T4 in the three journeys, the UIA report turn):** "ONE message, no shell calls." Judge the chain's verdict fields and `cat` output, then emit in the same message as parallel tool calls: `Write` `.claude/rules/project-context.md` ∥ `Write` `AGENTS.md` ∥ qualifying memory `Write`s ∥ § Completion Output text. The two facts agents kept re-reading (`<!-- discovery-metadata: cs=N xaml=N deps=N -->` first line, `<!-- PROJECT-CONTEXT:START/END -->` markers) are named inline with the pointer to `environment-setup.md` step 3, and the row states that no `Read`/`Grep` precedes these writes because the agent created every file they describe. Brownfield keeps "the discovery agent wrote them in T1 — do not rewrite."
- **§ Cross-session memory / Save:** saves are parallel `Write`s inside the report message, never one file per turn; "a first-try clean gate on card-covered activities qualifies nothing — skip the save entirely"; `uip rpa` flag shapes and run/debug payload fields added to **Never save** (cli-reference.md owns them; a gap there is a `/uipath-feedback` report).

`SKILL.md § Completion Output`: the memory-save sentence now says what qualifies (often nothing) and that the saves, the context files and the report text are one message with nothing running after it.

Rule numbering, journey turn numbering (T1-T4), anchors and inbound links are unchanged. No flavor override touches either file.

## Expected effect

On the bench data: the report tail collapses from 4-12 messages to 1-2 (the gate chain plus one report message; the UIA journey keeps its window-cleanup turn). Combined with #3102 (which removes the `--output-filter` retries feeding cause 1 and the memory topics feeding cause 3), greenfield XAML lands at 8-10 messages instead of 15-18.

## Follow-up commits from the combined bench

`5486c4dd6` — gate rows spell the full `run --file-path "<RELATIVE_FILE>" --project-dir "<PROJECT_DIR>" --skip-build --output json` (an agent ran `run --skip-build` without the required `--file-path` and paid a retry); brownfield report row says not to read, edit or rewrite the discovery-written context files after the gate (staleness is judged by file and dependency counts at the next session, per environment-setup § Project Context Discovery).

`a63ff648f` — `agents/uipath-project-discovery-agent.md`: the generated context's Quick Reference named `uip rpa run-file` and `uip rpa get-errors`. Both verbs exist but are hidden from `--help`, so the main agent judged them wrong and spent 4–10 post-gate turns "correcting" the files it had been told not to touch. The template now uses the standard-loop `uip rpa run --file-path …` / `uip rpa validate --file-path …`, the header line "Regenerate after significant project changes." is gone (it invited exactly that regeneration), and `list-instances --format json` became `instances list --output json` (`--format` is not a flag of that verb).

## Measured on a combined arm (#3055 + #3102 + this PR), 12 runs

`--output-filter` evaluation failures: 0 of 12 (13 of 15 on main). Memory-file writes: 2 (27 on main), emitted inside the report message. Post-gate turns: 2–6 in 11 of 12 runs (4–17 on main). Greenfield XAML end to end: 9 messages, $1.38–1.47 (main 15–18 messages, $2.36–2.42). Brownfield XAML after `a63ff648f`: 10 messages, 2-turn tail, zero context-file edits with the discovery agent running. All 12 runs succeeded with correct outputs.

## Relationship to other PRs

- #3102 fixes the run/debug payload documentation; independent files.
- #3055 renumbers the XAML journeys to T1-T5 and rewrites the T1/T2 rows. It touches the same greenfield/brownfield gate and report rows, so whichever merges second needs a trivial re-anchor (row text here, row numbers there). The content is orthogonal: #3055 batches the pre-authoring turns, this PR batches the post-gate turns.

## Verification

- `node scripts/check-skill-links.mjs`: all links resolve. `npm run skills:validate`: OK.
- Not reproducible in CI (the effect is turn count on a live agent); the measurement above is the 30-run bench, transcripts available on request.


Mihaiii (approved):

skills/uipath-rpa/agents/uipath-project-discovery-agent.md
```diff
@@ -259,7 +259,7 @@ Replace `{{PLACEHOLDER}}` sections with discovered values. Omit any section wher
 <!-- discovery-metadata: cs={{CS_COUNT}} xaml={{XAML_COUNT}} deps={{DEP_COUNT}} -->
 # {{PROJECT_NAME}} — Project Context
 
-> Auto-generated by project discovery agent. Regenerate after significant project changes.
+> Auto-generated by project discovery agent.
```
AlvinStanescu:
```suggestion
> Auto-generated by project discovery agent. Regenerate after significant project changes.
```

AlvinStanescu (approved):

PR #3110, opened 2026-09-06
test(uipath-governance): Form.io data-type coverage + unsupported-type guardrail (PLT-106003)

grant-tannert:
## Summary

Closes the data-type support/test work for `uipath-governance` aops-policy (PLT-106003). Documents every supported Form.io input type, adds a **warn-and-abort** rule for unsupported types, and adds three tests plus a payload round-trip verifier.

## Skill changes (aops-policy configure guide + commands)
- Prompt styles for `email`/`url`/`phoneNumber`, `time`, `datetime` (ISO-8601), `tags`, `survey`; `url` rejects `javascript:`/`data:`/`vbscript:`.
- Traversal recurses `table` + `container` (data-nesting), never skips display leaves, preserves `hidden` values; nested serialization for `container`/`survey`.
- New **unsupported / discouraged** category — `file`, `signature`, `address`, `datamap` → surface `data type "<type>" not supported` and **abort** (never serialize/author).
- Extended the type-appropriate default sentence in both files.

## Tests (uipath-governance/aops-policy)
| Task | Tier | Grades |
|------|------|--------|
| `aops_datatypes_template_smoke` | smoke (live) | Every supported type has a correctly-typed default; no unsupported type remains in E2E |
| `aops_datatypes_roundtrip_e2e` | e2e (live) | Create + read-back JSON-type integrity + distinctive values survive (catches coercion) |
| `aops_datatypes_guardrail_smoke` | smoke (offline) | Agent aborts on a fixture carrying all four unsupported types; no unsupported key serialized |

- `verify_aops_policy.py`: new `AOPS_CHECK=payload` mode (type integrity, grid row shape, `AOPS_EXPECT_JSON`, excluded-leak).
- Flat fixtures under `fixtures/` + `stage_unsupported_fixture.py`.

## Passing runs
- `skill-gov-aops-policy-datatypes-guardrail` — ran locally and **passed** (fully offline; no CLI/service dependency).
- `skill-gov-aops-policy-datatypes-template` and `skill-gov-aops-policy-datatypes-roundtrip` — ran locally and **passed against an unreleased `uip` CLI build** (data-type handling not yet in `@latest`). Not yet green on the released CLI — see Dependencies.

## Dependencies / merge order
- The two **live** tasks depend on the `uip` CLI data-type change that is **not yet released to `@latest`**. Until it ships, PR-gate smoke (which installs `@uipath/cli@latest`) will fail on those two tasks — this is expected, not a regression. Merge/enable after the CLI change lands.
- They also require the target tenant's E2E product to carry every supported type (datamap removed, datetime added). Updated in **alpha**; confirm CI's governance tenant points at the same environment before relying on CI results.
- The offline **guardrail** has neither dependency and is green today.

## Testing
- Check logic unit-tested locally (smoke passes clean-E2E, rejects datamap/missing-datetime; payload verifier catches stringified numbers, collapsed grids, dropped keys, leaks; guardrail scanner passes correct abort, fails on serialized unsupported data).
- `npm run skills:validate` passes.

## Notes
- Pre-existing `aops_diagnose_not_deployed_smoke` uses `llm_judge` (needs a judge backend) — unrelated to this PR.


grant-tannert:
Thanks for the review. Addressed in `5c35cbeb5`.

**Copilot — fixed**
- **`hidden` doc contradiction** (`aops-policy-commands.md:188`, `configure-...-guide.md:43`) — confirmed the CLI *does* emit `hidden` with its value (it's in the E2E `defaultData`), so the "omitted" list was factually wrong. Both descriptions now say `hidden` is kept — aligns with the preserve-hidden traversal rule.
- **Template smoke only checked form-data keys** (`aops_datatypes_template_smoke.yaml`) — coverage + unsupported-leak are now judged against **all raw-template component types**, so a type the CLI drops from form-data can't slip past.
- **Payload verifier was top-level only** (`verify_aops_policy.py`) — replaced the flat compare with a **recursive** shape check: container/survey children emptied to `{}`, or grid-row child values dropped/coerced, are now caught (verified against those exact cases).
- **Grid second row wasn't graded** (`aops_datatypes_roundtrip_e2e.yaml`) — added `AOPS_EXPECT_MIN_ROWS={"dataGrid": 2}`; the verifier now enforces the row-count the prompt asks for.
- **Stage script exited 0 on failure** (`stage_unsupported_fixture.py`) — now exits non-zero on a missing/partial fixture, so the guardrail's absence checks can't false-green on an empty workspace.

**Copilot — keeping, with rationale**
- **Guardrail has no positive criterion** — this is a refusal test; the repo convention (and this repo's own lint bot, which rated it OK) treats negative guards as the right shape. A skill-*ignoring* agent is caught: the fixture carries supported fields too, so authoring normally serializes the unsupported keys → the leak scan fails. `skill_triggered` gives the positive-engagement signal. The only untrapped case is a do-nothing agent, which still satisfies the safety property (nothing unsupported authored); grading the warning *text* deterministically would require an `llm_judge` backend, which we deliberately kept out of this offline test.
- **Nullable-default fields skip the type check** — not exploitable for this suite: every E2E blueprint default is non-null, so the skip never triggers here, and the recursive check + `AOPS_EXPECT_JSON` pin the graded fields. Deriving types for null defaults would need a raw-template fetch in the verifier — noted as a possible follow-up if `payload` mode is reused on a product with null defaults.

**Claude review** — the two Low items (`.gitignore` comment, `e2e.json` trailing newline) are moot: both files were removed from the PR after that review ran.

PR #3112, opened 2026-09-07
fix(uipath-maestro-bpmn): correct Integration Service activity authoring

DevMomo:
## Problem

An Integration Service activity authored from the manifest's own `InputNotes` — one `uipath:input` per request field — faults at runtime:

```xml
<uipath:input target="body" name="fields.project.key"  type="string" value="=vars.Var_ProjectKey" />
<uipath:input target="body" name="fields.issuetype.id" type="string" value="=vars.Var_IssueTypeId" />
<uipath:input target="body" name="fields.summary"      type="string" value="=vars.Var_Summary" />
```

They don't merge. Each input claims to be the whole body, the last wins, and the provider gets a bare scalar: `102002`, `status 500: Internal failure while handling request`. Slack is plainer — `missing required field: channel`.

Three more, found the same way:

| Authored | Result |
|---|---|
| `operation="CreateIssue"` (catalogue `Name`) | provider `400` |
| Slack activity with no `send_as` | `400 Value for required parameter 'send_as' not found` — passes `validate` **and** `pack` |
| connection binding without `resourceKey` | `validate`: `binding "<id>" is missing resourceKey` |

**Why the e2e needs this:** its Jira and Slack activities fault at runtime without it.

## Fix

- One `target="body"` input holding the whole request object, nested as the provider nests it — this is also what the canvas emits, and it round-trips
- `operation` = the described `Operation.Name` (`Create`), not the catalogue `Name`; the manifest's `DiscoveryNotes` point at the failing value
- Every `Required: true` `Parameter` gets its own input targeted by its `Type`
- `resourceKey` on connection bindings; folder-scoped activities need two sharing one

The CLI manifest still says `separateInputs`; it contradicts both the canvas and the runtime, and is called stale here.


skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -97,12 +97,91 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+`inputPattern: separateInputs` with `inputTarget: body`, and its `InputNotes`
+still tell authors to add one `uipath:input` per request field. That contradicts
+both the canvas and the runtime, so treat the manifest's InputNotes as stale
+here rather than as the contract.
+
+`target="bodyField"` is **not** an option here. It is the target of the single
+merged *arguments* payload on `mergedBody` / `scriptArgs` types —
+`JobArguments` for the `Orchestrator.*` process starts, `HitlTaskArguments` for
+`Actions.HITL`, `args` for `BPMN.ScriptTask` — and no `Intsvc.*` type uses it.
+The validator enforces that: any other direct input name is rejected with
```
rockymadden: **Major — this claim is false for `Intsvc.*`, and it inverts the lesson of the rest of the section.**

`does not support input payload "<name>"` is emitted from exactly two places in `validateInputs` (`packages/maestro-sdk/src/bpmn-validation/project-validator.ts`): the `mergedBody` / `splitInputsItemData` / `scriptArgs` branch at `:1649`, and the `inputFields` allow-list branch at `:1695`. Neither is reachable for this type, because `validateInputs` short-circuits first:

```ts
if (usesCanvasOwnedDynamicPayload(contract)) {
    this.validateDynamicDirectInputs(file, directInputs, diagnostics);
    return;
}
```

and (`:2366`):

```ts
function usesCanvasOwnedDynamicPayload(contract: BpmnExtensionType): boolean {
    return (
        contract.isDynamic === true ||
        (contract.extensionType.startsWith("Intsvc.") && contract.requiresDiscovery)
    );
}
```

`Intsvc.ActivityExecution` is `isDynamic: true` *and* `requiresDiscovery: true`, so it satisfies both disjuncts. `validateDynamicDirectInputs` (`:1702`) then checks only two things: every input has a non-empty `name`, and `type="json"` payloads parse as JSON. There is no name allow-list, no `target` check, and no count check.

So for a connector activity, local validation accepts all of: several `target="body"` inputs, a `target="bodyField"` input, and a completely unrecognized input name.

That's not incidental — it's the mechanism behind this PR's own problem statement. Three of the four bugs you found "pass `validate` **and** `pack`" precisely because the validator deliberately doesn't inspect dynamic payloads. Telling the reader it does enforce this undercuts the finding: an agent that believes the sentence will treat a clean `validate` as proof the body shape is right, which is the exact trap #2366 through #3112 exist to close.

Suggested rewrite for the last two sentences:

> `target="bodyField"` is **not** an option here. It is the target of the merged *arguments* payload on `Orchestrator.*` (`JobArguments`), `Actions.HITL` (`HitlTaskArguments`), and `BPMN.ScriptTask` (`args`); no `Intsvc.*` type uses it. Local validation will not catch a wrong target or a wrong input name on a connector activity — `validateInputs` short-circuits to `validateDynamicDirectInputs` for `isDynamic` / discovery-backed `Intsvc.*` types, which only checks that each input has a name and that JSON payloads parse. The failure surfaces at runtime.

That version is both accurate and more useful, since it tells the agent which check it *can't* rely on.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -97,12 +97,91 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+than merging, but it never generates them.
+
+The outlier is the CLI manifest: `Intsvc.ActivityExecution` still declares
+`inputPattern: separateInputs` with `inputTarget: body`, and its `InputNotes`
+still tell authors to add one `uipath:input` per request field. That contradicts
+both the canvas and the runtime, so treat the manifest's InputNotes as stale
+here rather than as the contract.
+
+`target="bodyField"` is **not** an option here. It is the target of the single
+merged *arguments* payload on `mergedBody` / `scriptArgs` types —
```
rockymadden: **Minor — `mergedBody` does not imply `bodyField`.** 5 of the 14 `mergedBody` types in the manifest use `inputTarget: body`, not `bodyField`:

- `A2A.AgentExecution`
- `Maestro.CaseManagerGuardrails`
- `Maestro.CaseRulesEvaluator`
- `Orchestrator.CreateQueueItem`
- `Orchestrator.CreateAndWaitForQueueItem`

So "the target of the single merged *arguments* payload on `mergedBody` / `scriptArgs` types" is wrong for roughly a third of `mergedBody`. An agent generalizing from this to "pattern is mergedBody, therefore target is bodyField" gets a wrong target on five types — and per the comment above, validation won't tell it.

The enumeration you already give is the reliable formulation; it just shouldn't be introduced as a property of the *pattern*. Suggest keying it to the types instead: `bodyField` is used by `Orchestrator.*` (`JobArguments`), `Actions.HITL` (`HitlTaskArguments`), and `BPMN.ScriptTask` (`args`) — read `inputTarget` from the manifest rather than inferring it from `inputPattern`.

(Two of the ten `bodyField` types — `Orchestrator.BusinessRules` and `Orchestrator.ExecuteApiWorkflowAsync` — also aren't "process starts", though they do use `JobArguments`.)

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -97,12 +97,91 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+the file is not well-formed; `>` needs no escaping in an attribute value, and
+inside CDATA nothing does.
+
+This single-input form **is** the canonical shape, and it round-trips. Studio
+Web's own design schema for this type sets `isSplitInputs: false` with one
+`jsonBody` at `target="body"`
+(`origin/develop:src/services/serialization/design-schema/intsvc.activityExecution.beta.design-schema.json`),
+and the canvas's round-trip fixture for a Jira `curated_create_issue` node is
+exactly one nested `target="body"` CDATA asserted parse→serialize identical
+(`origin/develop:src/services/serialization/xml-serialization.test.ts:706`).
```
rockymadden, line 151: **Minor — the fixture you cite here also contains the `operation` bug you fix 20 lines down.**

`xml-serialization.test.ts:706` is the right citation for the body shape, and I confirmed it: exactly one nested `target="body"` CDATA. But the same node reads:

```xml
<uipath:input name="operation" type="string" value="CreateIssue" />
```

which is exactly the value line 160 says draws a provider `400`. Since it's a parse→serialize identity test, it proves nothing about runtime validity — it just faithfully preserves whatever the exported file had, which is your own point about the serializer preserving rather than normalizing.

Worth one clause, because this paragraph explicitly sends the reader to that file as authoritative, and a reader who copies the node wholesale gets the body right and the operation wrong. Something like: "(that fixture also carries `operation=\"CreateIssue\"`, which is the value the next section corrects — copy its body shape, not its context inputs)."

For what it's worth, your `operation` fix has a crisper proof than the runtime `400` if you want one: `METHOD_TO_OPERATION` in `integrationservice-sdk/src/dap/validation/rules.ts:74` defines a closed six-name lexicon (List, Retrieve, Create, Update, Delete, Replace), `OPERATION_TO_METHOD` accepts only those plus raw HTTP verbs, and `uip is resources run` exposes exactly those six subcommands. `CreateIssue` can never resolve. That's a stronger argument than an observed status code, and it generalizes to every connector instead of Jira.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -128,7 +207,21 @@ discovery or the user.
...
+
+A folder-scoped connector activity needs TWO bindings that share one
+`resourceKey` (the connection id) and differ in `propertyAttribute`: the
+connection binding's `default` is the connection id, the folder binding's
+`default` is the folder key.
+
+```xml
+<uipath:bindings version="v1">
+  <uipath:binding id="Binding_JiraConn"   resource="Connection" propertyAttribute="ConnectionId" resourceKey="<connection-id>" default="<connection-id>" />
+  <uipath:binding id="Binding_JiraFolder" resource="Connection" propertyAttribute="folderKey"    resourceKey="<connection-id>" default="<folder-key>" />
```
rockymadden: **Minor — worth noting the folder binding contributes nothing to the package.**

The two-binding shape is right per the manifest (`folderKey` carries `bindingInfo` with `propertyAttribute: "folderKey"`, same `resourceKeyPattern` as `connection`). But `buildConnectionResources` in `maestro-sdk/src/bpmn-packaging/connection-resources.ts` filters bindings with:

```ts
node.resource === "Connection" && node.propertyAttribute === "ConnectionId"
```

so the `folderKey` binding is dropped, silently, and never becomes a resource in `bindings_v2.json`. Only the `ConnectionId` binding materializes — under the shared `resourceKey`.

That's presumably intended (the folder key rides along in the connector's context, not as its own package resource), but it's surprising enough to state: someone comparing the authored bindings against generated `bindings_v2.json` will count two in and one out and reasonably conclude refresh dropped something. One clause — "only the `ConnectionId` binding becomes a `bindings_v2.json` resource; the folder binding exists for authoring and validation" — saves that investigation.

Related: this is the same silent-drop path I flagged on #2387, where a `Connection` binding with an unexpected `propertyAttribute` disappears without a diagnostic.

rockymadden (approved):
## Approve

Best-sourced PR in the stack. I checked all ten verifiable claims against `UiPath/cli@main` and `UiPath/PO.Frontend@develop`, and all ten hold:

|---|---|---|
| `--output json` emits `IsEnrichment` while the CLI builds `ISEnrichment` | `maestro-tool/src/commands/registry.ts:587` writes `ISEnrichment:`; `toPascalCaseKey("ISEnrichment")` → `toLowerCamelCaseSimpleKey`'s `/^[A-Z]+(?=[A-Z][a-z]…)/` lowercases the `IS` prefix → `isEnrichment` → `IsEnrichment` | correct |
| manifest declares `separateInputs` / `inputTarget: body` with per-field `InputNotes` | verbatim: "Add each request field from the selected operation as its own uipath:input. Keep the field's type and target: path, query, body, or file." | correct |
| canvas design schema is single-body | `intsvc.activityExecution.beta.design-schema.json` → `structure.input` = `{isSplitInputs: false, jsonBody: {type:"json", name:"body", target:"body"}}` | correct |
| round-trip fixture is one nested `target="body"` CDATA | `xml-serialization.test.ts:706` — Jira `curated_create_issue`, exactly one body input | correct |
| no `Intsvc.*` type uses `bodyField` | manifest scan: `bodyField` appears only on `Actions.HITL`, `BPMN.ScriptTask`, and 8 `Orchestrator.*` types | correct |
| `JobArguments` / `HitlTaskArguments` / `args` naming | confirmed across all 10 `bodyField` templates | correct |
| `operation` is `Create`, not `CreateIssue` | decisive — `METHOD_TO_OPERATION` (`integrationservice-sdk/src/dap/validation/rules.ts:74`) is a closed six-verb lexicon: List, Retrieve, Create, Update, Delete, Replace. `CreateIssue` isn't in it, so `resolveOperation` returns `undefined`. `uip is resources run` exposes exactly those six subcommands | correct |
| `DiscoveryNotes` say "set operation from Name" | verbatim in the manifest, and genuinely ambiguous as you describe | correct |
| `resourceKey` required, with that exact error | `project-validator.ts:1572` renders `Integration Service activity connection binding "<id>" is missing resourceKey.` | correct |
| folder-scoped needs two bindings sharing `resourceKey` | manifest `contextFields`: `connection` → `propertyAttribute: "ConnectionId"`, `folderKey` → `propertyAttribute: "folderKey"`, both `resource: "Connection"` with the same `resourceKeyPattern: "${resourceKey}"` | correct |

The `IsEnrichment` catch deserves specific credit — the CLI's own help text at `registry.ts:418` says "as ISEnrichment", so anyone grepping the output for the documented spelling finds nothing and has no idea why.

One major item inline: a single sentence claims the validator enforces the input-name rule, and for `Intsvc.*` it provably does not. Fixing it makes the section stronger, because "validate won't catch this" is this PR's whole thesis. Three minors alongside.

No AI slop; this is the tightest sourcing in the stack. "Measured on live Alpha against both the Atlassian Jira and Slack connectors" is the right way to state an empirical claim — scoped and falsifiable. And calling the manifest **stale** rather than trying to reconcile it is the honest read. Worth filing a CLI issue against those `InputNotes`/`DiscoveryNotes` so the next reader doesn't re-derive this from scratch.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: You are right and the sentence was self-defeating — it made a clean `validate` look like proof of the thing this PR exists to fix. Confirmed the short-circuit at `project-validator.ts:793` and `usesCanvasOwnedDynamicPayload` at `:1602`. Took your rewrite, stating that several `target="body"` inputs, a `bodyField` input, and an unrecognized name all pass `validate` and `pack`, and that the failure is a runtime one.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Confirmed against the manifest — 5 of 14, exactly the types you list. Keyed it to `inputTarget` per type with "not a property of `inputPattern`" stated outright, and named the five. One tightening on your phrasing: `Orchestrator.*` is not uniform either, since `CreateQueueItem` and `CreateAndWaitForQueueItem` are `body`, so the text says the `Orchestrator.*` job and process starts rather than `Orchestrator.*`.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Added the clause — copy the body shape, not the context inputs. Also took the better proof: `METHOD_TO_OPERATION` (`rules.ts:74`) is a closed six-name lexicon and the reverse map accepts only those plus raw HTTP verbs, which is what `uip is resources run` exposes. That replaces the observed `400`, and it generalizes past Jira.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Added — two bindings in, one resource out is expected, plus the pointer to the same filter being why an unexpected `propertyAttribute` vanishes and resurfaces as a missing-binding error on the activity.

tests/scripts/test_bpmn_connector_request_contract.py
```diff
@@ -0,0 +1,110 @@
...
+    return parsed
+
+
+def _inputs(root: ET.Element) -> list[ET.Element]:
+    if root.tag == f"{{{NS['uipath']}}}input":
+        return [root]
+    return root.findall(f".//{{{NS['uipath']}}}input")
+
+
+def test_body_example_is_a_single_whole_body_input() -> None:
```
rockymadden: **🔴 critical** — this never asserts singleness.

It flattens body inputs across every block and loops. Restore the doc to the three sibling `target="body"` inputs from the PR description, give each `type="json"` and CDATA content, and this stays green — the exact regression the PR exists to prevent.

```python
for block in _blocks():
    bodies = [i for i in _inputs(block) if i.attrib.get("target") == "body"]
    if bodies:
        assert len(bodies) == 1, [b.attrib for b in bodies]
```

tests/scripts/test_bpmn_connector_request_contract.py
```diff
@@ -0,0 +1,110 @@
...
+    ROOT / "skills" / "uipath-maestro-bpmn" / "references" / "registry-workflow.md"
+)
+NS = {"uipath": "http://uipath.org/schema/bpmn"}
+
+# integrationservice-sdk/src/dap/validation/rules.ts: METHOD_TO_OPERATION is a
+# closed lexicon, and `uip is resources run` exposes exactly these operations.
+OPERATION_LEXICON = {"List", "Retrieve", "Create", "Update", "Delete", "Replace"}
+
+
+def _blocks() -> list[ET.Element]:
```
rockymadden: **🟠 major** — inlines its own doc-example loader.

`tests/scripts/bpmn_doc_example.py` exists to prevent that, and `0e64b1d1a` removed an inlined copy from `test_bpmn_project_scaffold_contract.py`:

> Two loaders for one example is the drift that module exists to prevent, and the inlined one was weaker — no section bound and no "exactly one parseable block" assertion.

This copy has both weaknesses again: it scans all 418 lines, and it counts nothing.

**Fix:** parameterize `bpmn_doc_example` with `(path, section)` and bind to `### Body shape` / `### Required Parameters`.

tests/scripts/test_bpmn_connector_request_contract.py
```diff
@@ -0,0 +1,110 @@
...
+    assert targeted, (
+        "registry-workflow.md no longer shows a required parameter as its own "
+        "targeted input; that omission fails only at runtime with "
+        "\"Value for required parameter '<name>' not found\""
+    )
+    for item in targeted:
+        assert item.attrib.get("name"), item.attrib
+
+
+def test_operation_guidance_uses_the_closed_lexicon() -> None:
```
rockymadden: **🟠 major** — only reads `--operation <X>` CLI flags.

The section is about the XML surface, and the doc itself carries `operation="CreateIssue"` at line 150. Paste a full node into an ` ```xml ` fence with `<uipath:input name="operation" value="CreateIssue" />` and this passes.

Walk the parsed blocks too:

```python
for block in _blocks():
    for item in _inputs(block):
        if item.attrib.get("name") == "operation":
            assert item.attrib.get("value") in OPERATION_LEXICON, item.attrib
```

tests/scripts/test_bpmn_connector_request_contract.py
```diff
@@ -0,0 +1,110 @@
+"""Guard the connector request-shape rules in the maestro-bpmn skill.
+
+These rules are the ones local validation cannot catch: `validateInputs`
+short-circuits to `validateDynamicDirectInputs` for discovery-backed
+`Intsvc.*` types, so a wrong target, a wrong operation, or a missing required
+parameter passes `validate` AND `pack` and fails only at runtime. Until now
+the only thing that caught them was the tenant-gated escalation e2e, which
```
rockymadden: **⚪ minor** — "the tenant-gated escalation e2e" is not findable.

There is no Jira or Slack BPMN e2e in `tests/tasks/uipath-maestro-bpmn/`. Cut the sentence; the provenance belongs in the commit message.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -92,12 +92,123 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
 the activity's `body` input (`target="body"`) and `context` (`connectorKey`,
 `objectName`) from that enrichment — do not hand-author connector schemas. The
 connection is referenced through a connection binding, `=bindings.<bindingId>`
 (see §4).
 
+### Body shape: hand-authored files need ONE `target="body"` input
+
+The manifest declares `inputPattern: separateInputs` with `inputTarget: body`,
+and its `InputNotes` tell you to add each request field as its own
+`uipath:input`. Studio Web's canvas does emit that shape. **The runtime does
```
rockymadden: **🟠 major** — contradicts line 148.

- Here: *"Studio Web's canvas does emit that shape."*
- Line 148: *"it never generates them."*

Both describe the canvas emitting separate inputs. Pick one — the rest of the section leans on "never generates".

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -92,12 +92,123 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+(`origin/develop:src/services/serialization/design-schema/intsvc.activityExecution.beta.design-schema.json`),
+and the canvas's round-trip fixture for a Jira `curated_create_issue` node is
+exactly one nested `target="body"` CDATA asserted parse→serialize identical
+(`origin/develop:src/services/serialization/xml-serialization.test.ts:706`).
+The serializer *preserves* separate inputs if a file already has them rather
+than merging, but it never generates them. Copy that fixture's body shape, not
+its context inputs: being a parse→serialize identity test, it faithfully
+preserves an `operation="CreateIssue"` that the next section corrects.
+
+The outlier is the CLI manifest: `Intsvc.ActivityExecution` still declares
```
rockymadden, line 157: **🔴 critical** — the outlier ships in this repo, and this PR leaves it wrong.

The copy the agent authors from is `skills/uipath-maestro-bpmn/validator/bpmn-spec.json`:

- `Intsvc.ActivityExecution.inputNotes`: *"Add each request field from the selected operation as its own uipath:input. Keep the field's type and target: path, query, body, or file."*
- `Intsvc.ActivityExecution.discoveryNotes`: *"…Set operation from Name, then describe ObjectName…"*

Line 39 of this same file routes the agent to that spec as evidence, and §2 tells it to author from `xmlTemplate` / `inputPattern` / `inputTarget`. The spec is hand-maintained here (`99f40f5f6`, `d9073723b`) — no generator in `scripts/` or `.github/`.

**Fix:** correct both note strings, and guard them, instead of documenting around them.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -92,12 +92,123 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+CLI builds it as `ISEnrichment`; the output formatter normalizes the key, so
+`--output json` emits `IsEnrichment` — read that one). Write
 the activity's `body` input (`target="body"`) and `context` (`connectorKey`,
 `objectName`) from that enrichment — do not hand-author connector schemas. The
 connection is referenced through a connection binding, `=bindings.<bindingId>`
 (see §4).
 
+### Body shape: hand-authored files need ONE `target="body"` input
+
+The manifest declares `inputPattern: separateInputs` with `inputTarget: body`,
```
rockymadden: **🟠 major** — the rule is scoped to `Intsvc.ActivityExecution` alone.

`Intsvc.{Async,SyncAgent,AsyncAgent,SyncWorkflow,AsyncWorkflow}Execution` — which line 376 already lists — are all `separateInputs` + `inputTarget: body` + `isDynamic: true` in the same spec. Same fault class, no guidance.

**Fix:** state it for every `Intsvc.*` type whose `inputTarget` is `body`.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -92,12 +92,123 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+(`project-validator.ts:1602`) is true for anything `isDynamic` or for an
+`Intsvc.*` type requiring discovery, and it returns after
+`validateDynamicDirectInputs`, which checks only that each input has a `name`
+and that `type="json"` payloads parse. No allow-list, no `target` check, no
+count check — so several `target="body"` inputs, a `bodyField` input, and an
+unrecognized input name all pass `validate` and `pack`. The failure is a
+runtime one. This is why a clean `validate` is not evidence the body shape is
+right.
+
+Dotted Integration Service field names are also nested into objects on the way
```
rockymadden: **🟠 major** — this sentence contradicts its own conclusion.

> Dotted Integration Service field names are **also nested into objects on the way out**, not sent as literal flat keys, **so** a flat `fields.project.key` leaves the provider never seeing `fields.project`.

If they are nested on the way out, the provider *does* see `fields.project`.

**Fix** — rewrite as the instruction:

> Nest a dotted `RequestFields` name yourself — `fields.project.key` becomes `{"fields":{"project":{"key":…}}}`. A literal `"fields.project.key"` key is sent as-is and the provider never sees `fields.project`.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -92,12 +92,123 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
...
+the provider reports the other fields missing (Slack:
+`missing required field: channel`). Measured on live Alpha against both the
+Atlassian Jira and Slack connectors.
+
+So when you hand-author the XML, emit exactly one `target="body"` input holding
+the complete request object as JSON element content, nested the way the
+provider's API nests it:
+
+```xml
+<uipath:input name="body" type="json" target="body"><![CDATA[{"fields":{"project":{"key":"=vars.Var_TargetProject"},"issuetype":{"id":"3"},"summary":"=js:'Created from Maestro at ' + vars.Var_RunLabel}}]]></uipath:input>
```
rockymadden: **⚪ minor** — the example fights the line three below it.

`"issuetype":{"id":"3"}` is a hard-coded literal directly above *"build the body from variables rather than literals"*, and a bare Jira issue-type id is the fabricated identifier §1 forbids.

`cdd7a1152` introduced it while neutralizing the eval-specific values — neutralize the *name*, not the mechanism:

```
"issuetype":{"id":"=vars.Var_IssueTypeId"}
```

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -123,7 +234,30 @@ discovery or the user.
 
 Declare all bindings in a single process-level `<uipath:bindings version="v1">`
 block. Each `<uipath:binding>` carries `id`, `resource`, `propertyAttribute`,
-and a `default` value (the resolved key/id).
+`resourceKey`, and a `default` value. **`resourceKey` is required** — omitting
```
rockymadden: **🟠 major** — over-broad, and it strands the other binding kinds.

> Each `<uipath:binding>` carries `id`, `resource`, `propertyAttribute`, `resourceKey`, and a `default` value. **`resourceKey` is required**

Stated for every binding, but the quoted error names only *"Integration Service activity connection binding"*. An agent authoring a `process` or `queue` binding (the bullet at line 218) now needs a `resourceKey` it has no source for, and §1 forbids inventing one.

**Fix:** scope the sentence to connection bindings, and restore the dropped `(the resolved key/id)` gloss on `default`.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -92,12 +92,123 @@ uip maestro bpmn registry get Intsvc.ActivityExecution \
     --connection-id <id> --object-name <object> --output json
 ```
 
-The response adds an `ISEnrichment` block with the live field metadata. Write
+The response adds an `IsEnrichment` block with the live field metadata (the
```
rockymadden: **⚪ minor** — pins one casing produced by the CLI's `OutputFormatter`.

`tests/scripts/test_runtime_payload_key_casing.py` exists because that formatter (CLI #2266) broke every hard-coded key read, and its verdict is: read case-insensitively, never pin a spelling.

**Fix:** say *"match the key case-insensitively"* and drop the CLI-internals aside.

rockymadden (changes requested):
# 📋 Review

## What

+136/−2 in `references/registry-workflow.md`, a new 110-line offline guard, and one CI line adding it to the existing `maestro-bpmn contract guards` job (job name unchanged, so `test_required_checks_contract.py` is unaffected ✅).

Four rules, all runtime-only failures: one `target="body"` input instead of per-field siblings; `operation` from the described `Operation.Name`; every `Required: true` `Parameter` as its own targeted input; `resourceKey` on connection bindings.

The knowledge is real and the `mergedBody` arithmetic checks out — verified all 14 against `validator/bpmn-spec.json`: exactly 5 use `body` (`A2A.AgentExecution`, `Maestro.CaseManagerGuardrails`, `Maestro.CaseRulesEvaluator`, `Orchestrator.CreateQueueItem`, `Orchestrator.CreateAndWaitForQueueItem`), and no `Intsvc.*` uses `bodyField` ✅.

## Overall findings

**🔴 critical — the PR names the stale source as external, but it ships in this repo and this PR leaves it wrong.**

`skills/uipath-maestro-bpmn/validator/bpmn-spec.json` → `Intsvc.ActivityExecution`:

- `inputNotes`: *"Add each request field from the selected operation as its own uipath:input."*
- `discoveryNotes`: *"Set operation from Name…"*

Both are the exact guidance the new section calls stale. `registry-workflow.md:39` routes the agent to that spec as evidence, and §2 tells it to author from `xmlTemplate` / `inputPattern` / `inputTarget`. The spec is hand-maintained here (`99f40f5f6`, `d9073723b`) — no generator in `scripts/` or `.github/` — so it is editable in this PR. Shipping the correction in prose while the authored-from artifact still says the opposite is the contradiction, not a description of it.

**Fix:** amend both note strings, and extend the new guard to assert no `Intsvc.*` entry's `inputNotes` says "its own uipath:input".

**🟠 major — ~60 of the 136 added lines are provenance the agent cannot act on.**

`origin/develop:src/services/serialization/xml-serialization.test.ts:706` · `…/intsvc.activityExecution.beta.design-schema.json` · `project-validator.ts:1602` · `integrationservice-sdk/src/dap/validation/rules.ts:74` · `connection-resources.ts` · *"Measured on live Alpha against both the Atlassian Jira and Slack connectors."*

None of it is reachable from this repo, all of it rots on the next refactor over there, and no line of it changes what the agent emits. Per this repo's own skill rule, that fails the "does this change what the agent does?" test. Keep the rule and the observable symptom (`102002`, `500 Internal failure`, `missing required field: channel`, `Value for required parameter '<name>' not found`) — move the provenance to the commit message.

**🟠 major — AI slop.** The register is essay, not instruction: *"What comes back names neither the field nor the cause"*, *"the described contract wins over the provider's own vocabulary"*, *"This is the same trap as…"*. The four rules fit in ~40 lines. At 136 they cost context on every load of this reference.

**🟠 major — the guard tests the doc, not the agent.** It asserts properties of examples in a markdown file. Nothing in the repo fails if an agent emits three body inputs. `connector/integration_service_boundary.yaml` is a deliberate draft-boundary task (`check_integration_service_boundary.py` asserts the agent *did not* author connector metadata), so it structurally cannot grade a body shape. Say so, or add a fixture the guard parses as an authored node.

**⚪ could not verify — flagging rather than guessing.** Every runtime claim rests on live-Alpha runs, with no artifact in the PR: non-merging `target="body"` siblings, `102002` / `500`, Slack's `send_as` `400`, the `resourceKey` validate error text, `--operation` as a flag of `uip is resources run`, and the `ISEnrichment` → `IsEnrichment` normalization. Only what is checkable in-repo was confirmed. If a transcript exists, link it.

## Recommendation

**🔴 Request changes.** The core knowledge is worth landing. Blocking: the shipped spec still carries the guidance being corrected, the doc contradicts itself in two places and is incoherent in a third, and the guard misses the one rule the PR exists to protect.

---

**tl;dr** — Right knowledge, four blocking problems. The stale guidance called "the CLI manifest" ships in `validator/bpmn-spec.json` in this repo and is still wrong after this PR. The doc contradicts itself at 107/148 and 252, and line 179 says the opposite of what it means. The guard never asserts the one-body rule it is named for, and it re-inlines the loader converged three weeks ago in `0e64b1d1a`. Plus ~60 lines of another repo's file:line citations that change nothing an agent does.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -123,7 +234,30 @@ discovery or the user.
...
+`default` is the folder key.
+
+```xml
+<uipath:bindings version="v1">
+  <uipath:binding id="Binding_JiraConn"   resource="Connection" propertyAttribute="ConnectionId" resourceKey="&lt;connection-id&gt;" default="&lt;connection-id&gt;" />
+  <uipath:binding id="Binding_JiraFolder" resource="Connection" propertyAttribute="folderKey"    resourceKey="&lt;connection-id&gt;" default="&lt;folder-key&gt;" />
+</uipath:bindings>
+```
+
+Only the `ConnectionId` binding becomes a `bindings_v2.json` resource — the
```
rockymadden, line 260: **🟠 major** — these two paragraphs cannot both be true as written.

Two bindings in / one resource out is *"expected, not a dropped binding"*, then *"the same filter is why a `Connection` binding with any other `propertyAttribute` vanishes without a diagnostic and resurfaces as `Activity "<name>" references missing Connection binding "<id>"`"*.

The prescribed folder binding is precisely `resource="Connection"` with another `propertyAttribute`, and `bpmn-spec.json`'s `xmlTemplate` references it via `=bindings.{folderBindingId}`.

**Fix:** say what keeps the folder binding out of that error — otherwise the doc prescribes the shape it then calls a fault.

tests/scripts/test_bpmn_connector_request_contract.py
DevMomo: Fixed — asserted per block now. Verified with your exact scenario: three sibling `target="body"` inputs fail it, and passed before.

tests/scripts/test_bpmn_connector_request_contract.py
DevMomo: Using `bpmn_doc_example` — added `section_blocks(reference, section)` there so it binds to a heading and refuses a block that will not parse. Both weaknesses you named were the ones that module was created to prevent, and I reintroduced them.

tests/scripts/test_bpmn_connector_request_contract.py
DevMomo: Walks the parsed blocks too now. Verified: a node carrying `operation="CreateIssue"` fails it.

tests/scripts/test_bpmn_connector_request_contract.py
DevMomo: Cut — you are right that it is unfindable from this PR, since the e2e sits above it in the stack.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Dropped the "canvas does emit that shape" clause; the section leans on "never generates them".

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Corrected both note strings in `bpmn-spec.json` rather than documenting around them, and guarded them. The guard earned itself immediately: it found a second one, `Intsvc.AsyncExecution`, teaching the same separate-input shape in different words.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Scoped to all six `Intsvc.*` types with `inputTarget: body`, listed explicitly.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Rewritten as your instruction. The sentence did say the opposite of its own conclusion.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Now `=vars.Var_IssueTypeId`. I neutralized the value and left a fabricated literal in its place, directly above the line forbidding exactly that.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Scoped `resourceKey` to connection bindings, said plainly that process/queue/businessRule carry none and must not have one invented, and restored the `(the resolved key or id)` gloss.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Changed to "match the key case-insensitively" and dropped the CLI-internals aside.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Resolved with the mechanism, checked in `connection-resources.ts`: the lookup runs only on the binding named by the activity's `connection` input, so `folderKey` never reaches `bindingById`. The error fires when that input points at a non-`ConnectionId` binding. As written the doc prescribed a shape it then called a fault.

rockymadden (approved):

PR #3113, opened 2026-09-07
test(uipath-maestro-bpmn): outcome-based live Alpha e2e eval

DevMomo:
## Problem

The BPMN suite had no test proving a generated process *does anything*. Coverage graded structure — well-formed XML, `validate` passes, the agent said it was done — all of which a process can satisfy while achieving nothing at runtime.

Every one of these passed the old bar:

| Authored | `validate` | `pack` | Runtime |
|---|---|---|---|
| per-field `target="body"` inputs | ✅ | ✅ | IS `500`, no ticket |
| Slack activity missing `send_as` | ✅ | ✅ | `400`, no message |
| `scriptTask` with a `BPMN.ScriptTask` mapping | ✅ | ✅ | script never executes |
| public output under an id nothing writes | ✅ | ✅ | exposes `null` |

**Why the stack below:** #2369 the output bridge, #2366 the executing ScriptTask, #2387 real connection bindings, #3112 the connector request shape — each found by this eval failing on it.

## Fix

From a seeded Sev1 case the process must classify severity, create a **real Jira issue whose summary carries the seeded correlationId**, and post **exactly one** Slack alert. The grader re-reads the tenant rather than parsing the agent's narration.

| Criterion | Weight |
|---|---|
| skill loaded | 1.0 |
| agent validated its own output | 1.5 |
| CLI validator accepts the BPMN | 3.0 |
| packs with real (non-stub) bindings | 2.0 |
| **live run: real Jira ticket + Slack alert** | 5.0 |

The correlationId is generated per run, so a hard-coded summary can't satisfy it, and the exposed `jiraIssueKey` must be the executed node's own response key.

## Verification

Score **1.000**, reproduced across runs (650s–995s). 22 stdlib-only unit tests for the behaviour checker. `run_limits` match the flow e2e.


tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/customer_escalation_triage.yaml
```diff
@@ -0,0 +1,136 @@
...
+    command: "python3 $TASK_DIR/check_customer_escalation_package.py"
+    timeout: 180
+    expected_exit_code: 0
+    weight: 2.0
+    pass_threshold: 1.0
+
+  - type: run_command
+    description: "Seeded Sev1 debug run creates a real Jira ticket and Slack alert carrying the seeded correlationId (verified by re-reading the tenant)"
+    command: "python3 $TASK_DIR/check_customer_escalation_behavior.py"
+    timeout: 1200
```
rockymadden: **Major — this timeout is under-budgeted against the checker's own step budgets.**

Adding up the per-call timeouts on the path `check_customer_escalation_behavior.py` actually walks:

| Step | Budget |
|---|---|
| `uip login status` (`assert_live_target`) | 60 |
| `is connections list --all-folders` (`connection_ids`) | 180 |
| `solution init` | 120 |
| `solution projects import` | 180 |
| `bpmn debug` (`DEBUG_TIMEOUT_SECONDS`) | 480 |
| `debug-instance variables-all` | 180 |
| `debug-instance incidents` | 180 |
| `get_issue_fields` (Jira reread) | 120 |
| `delete_jira_issue` + `delete_slack_message` | 240 |
| **worst case** | **1740** |

So the checker is permitted to take up to ~1740s inside a 1200s criterion. Your own reported range is 650s–995s, which means a normal-but-slow run already sits at **83%** of budget — one slow connector call or a slow `solution import` puts it over.

The closest precedent agrees: flow `escalation_jira_ticket`, which this task explicitly mirrors and whose `run_limits` it matches exactly (55/120/2400/1200), gives its live criterion `timeout: 1600`.

A timeout kill is also expensive here, because the in-band deletes are the *last* thing `main()` does. Getting SIGKILLed at 1200s means every created Jira issue and Slack message falls to `post_run`. The journal design handles that correctly — this is the scenario it exists for — but it converts a passing run into a failing one plus a cleanup that has to work.

Suggest `timeout: 1600` to match the flow precedent. If you want the tighter number, the honest way is to lower the step budgets so they sum under it.

Related, and the reason I'd flag this rather than assume it was considered — see the comment on `test_behavior_checker.py:311`.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/test_behavior_checker.py
```diff
@@ -0,0 +1,342 @@
...
+    def test_prompt_declares_every_seeded_input(self):
+        # Seed/prompt lockstep: an input the prompt never names would grade an
+        # unstated requirement when the grader passes it to `bpmn debug`.
+        task_text = (HERE / "customer_escalation_triage.yaml").read_text(
+            encoding="utf-8"
+        )
+        for name in SEED["inputs"]:
+            self.assertIn(name, task_text, f"prompt never names input {name}")
+
+    def test_yaml_timeout_covers_debug_budget(self):
```
rockymadden: **Minor — this test reads like a budget guard but doesn't guard the budget.**

It asserts two things:

1. `"timeout: 1200" in task_text` — a substring match against the whole YAML. `turn_timeout: 1200` on line 27 also satisfies it, so the criterion's timeout could be deleted entirely and this still passes.
2. `checker.DEBUG_TIMEOUT_SECONDS < 1200` — trivially true (480 < 1200) and says nothing about the other eight timed calls.

Meanwhile the failure message reads "behavior criterion timeout drifted from the checker's budget," which is exactly the invariant that *isn't* being checked — the sum, per the comment on the YAML.

If you want this to hold, sum the real budgets. Something like exposing the step timeouts as named constants in the checker and asserting `sum(checker.STEP_TIMEOUTS) <= criterion_timeout`, with the criterion timeout parsed out of the YAML for that specific criterion rather than substring-matched. That turns a test that currently passes by construction into one that would have caught the 1740-vs-1200 gap.

Worth doing precisely because the other 21 tests in this file are good — they assert real behavior with real counterexamples (`test_jira_key_output_must_match_created_key`, `test_wrong_slack_channel_fails`, `test_slack_text_missing_correlation_fails`). This one is the odd one out.

.github/workflows/test-helpers.yml
```diff
@@ -3,9 +3,9 @@ name: Test Helpers
 on:
   pull_request:
     paths:
+      - 'tests/tasks/uipath-maestro-bpmn/**'
```
rockymadden: **Minor — this change has no effect; suggest dropping it.**

The diff moves `'tests/tasks/uipath-maestro-bpmn/**'` from third position to first in the `paths:` list. GitHub Actions path filters are an unordered OR-match, so the trigger behaves identically before and after.

It's the only change to a CI workflow file in a PR that is otherwise entirely additive under `tests/`, which makes it read as intentional to anyone reviewing workflow history later. Reverting it keeps this PR's blast radius exactly where its content is.

(Same standard as the skill-review question — if the answer to "does this change what the tool does?" is no, it probably shouldn't be in the diff.)

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/check_customer_escalation_behavior.py
```diff
@@ -0,0 +1,458 @@
...
+    payload_data(initialized, "initialize ephemeral solution")
+    imported = run_cli(
+        [
+            "uip",
+            "solution",
+            "projects",
+            "import",
+            str(PROJECT.resolve()),
+            "--solutionFile",
+            str(next(solution_dir.glob("*.uipx"))),
```
rockymadden: **Minor — this is the one failure path that doesn't produce a diagnosable message.**

`next(solution_dir.glob("*.uipx"))` raises a bare `StopIteration` if `solution init` produced no `.uipx` — and because that's outside the `CheckFailure` contract, the `__main__` handler doesn't catch it, so the criterion fails with a raw traceback instead of `FAIL: <reason>`.

Everything else in this file is careful about exactly this: `run_debug` reports the exit code plus `tail_log`, `assert_outcome` surfaces the faulted element ids and incidents, `payload_data` labels each call. This line is the outlier.

```python
solution_files = sorted(solution_dir.glob("*.uipx"))
if len(solution_files) != 1:
    raise CheckFailure(
        f"solution init produced {len(solution_files)} .uipx files in {solution_dir}"
    )
```

The `sorted()` also makes the choice deterministic if init ever emits more than one, which `glob` ordering does not guarantee.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/escalation_is.py
```diff
@@ -0,0 +1,235 @@
...
+
+JIRA_CONNECTOR = "uipath-atlassian-jira"
+SLACK_CONNECTOR = "uipath-salesforce-slack"
+CONNECTION_NAMES = {
+    JIRA_CONNECTOR: "is-sandboxes-test@uipath.com-uipath-sandbox-380",
+    SLACK_CONNECTOR: "is-sandboxes",
+}
+# Shared/uipath-maestro-flow — the folder every escalation e2e's connections
+# live in, matched by key because `connections list` reports keys, not paths.
+CONNECTION_FOLDER_KEY = "5da18ec0-7de1-4e57-aaf1-ddc8a369c199"
```
rockymadden: **Note, pre-existing and not introduced here — flagging for an org-level decision, not asking for a change in this PR.**

`UiPath/skills` is a public repo, and this file commits:

- `CONNECTION_FOLDER_KEY = "5da18ec0-7de1-4e57-aaf1-ddc8a369c199"` (a folder key GUID)
- a connection name containing `is-sandboxes-test@uipath.com`
- `SLACK_CHANNEL_ID = "C01H4SPS77W"`
- `EXPECTED_LIVE_TARGET` = `alpha.uipath.com` / `codereval` / `DefaultTenant`

I checked, and all of these except the Slack channel id already appear in the flow suite (`uipath-maestro-flow/e2e/*/jira_is.py` carries the same connection name; the same folder key is in `_shared/test_flow_check.py:312`; `alpha.uipath.com` appears across several suites). So this PR is following established convention and I'm not asking it to deviate.

The reason it's worth raising anyway: the skill this PR tests says the opposite. `references/shared/project-layout.md`, under Drift Handling — *"Do not commit private IDs, tenant URLs, connection IDs, folder keys, or copied customer payloads."* None of these are credentials, and a sandbox tenant is low-stakes, but the repo is publishing the identifiers its own guidance tells authors to withhold.

If that's a deliberate exception for tenant-gated e2e fixtures, it'd be worth stating in `tests/README.md` so it stops looking like an oversight to every reviewer who reads both files. If it isn't, it's a suite-wide cleanup (env vars or a gitignored config) rather than anything for this PR.

rockymadden (approved):
## Approve

Best-engineered PR in the stack, and the anti-gaming design is genuinely hard to fake past. Things I verified rather than took on trust:

| Claim | Check | Result |
|---|---|---|
| 22 stdlib-only unit tests | ran `python3 -m unittest test_behavior_checker -v` | 22 pass, 0.049s |
| `run_limits` match the flow e2e | flow `escalation_jira_ticket` is 55/120/2400/1200 — identical | correct |
| tags come from the closed vocabulary | `path-to-ga` (tests/README.md:109), `outcome-graded` (:110), `connector` as a flat marker (:118), `feature:escalation` has precedent | correct |
| `EXPECTED_CONNECTION_RESOURCES = 2` | matches `buildConnectionResources` deduping by `resourceKey`, one per connector | correct — and it independently confirms that the folderKey bindings from #3112 never materialize as package resources |

The parts that make this hard to game are worth naming, because they're the reason the stack below it exists: the per-run `secrets.token_hex(4)` correlationId, the sha256 pin across solution import, requiring `jiraIssueKey` to equal the executed node's *own* response key, the exactly-once execution count summed across element ids, and re-reading the issue from the tenant rather than trusting the run's output. Any one of those alone would be beatable; together they aren't.

`assert_live_target()` deserves specific credit. A grader that deletes Jira issues and Slack messages should refuse to run anywhere but the intended sandbox tenant, and this one does, by asserting BaseUrl/Organization/Tenant before touching anything. Same for journalling created ids *before* any assertion so `post_run` can clean up through a SIGKILL — that ordering is the correct instinct and the docstring explains why.

One major item inline (the criterion timeout is under-budgeted against the checker's own step budgets) plus two minors and one pre-existing note. None of them are reasons to hold this.

No AI slop. Several comments are doing real work and should survive future edits: why the pinned project path is graded here but discovered in the flow suite, why journalling precedes assertion, and why `bpmn debug` needs `debug-instance variables-all` with id-addressed variables where `flow debug` doesn't. The `tests/README.md` section on running judge-free tasks with no credentials — including "do not set `ANTHROPIC_API_KEY` to help" — will save someone an hour. And the failure diagnostics are better than most production code: a non-Completed run reports which element faulted plus the incidents, instead of a bare status that sends the reader back to the tenant.

One overstatement worth trimming: "22 stdlib-only unit tests for the behaviour checker" covers the pure assertion and harvest functions, not the live path. Accurate, but it reads broader than it is.

Note on this approval: `path-to-ga approval` is currently failing because this PR adds the `path-to-ga` tag, which requires approval from an allowlisted reviewer. I'm on that list, so this approval clears that gate as well as recording a review.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/customer_escalation_triage.yaml
DevMomo: Your table was right. Took the second option as well as the number: every step budget is now a named constant, the over-generous reads are trimmed (login status 60→30, connections list 180→120, solution init 120→90, variables-all and incidents 180→120 each), and the sum is 1500 under a 1600 criterion — the flow precedent, now with headroom rather than by coincidence.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/test_behavior_checker.py
DevMomo: Correct on both counts, and it was the reason the gap survived. It now parses the behavior criterion out of the YAML by its command and asserts `sum(checker.STEP_TIMEOUTS) <= criterion_timeout` over all ten calls. Checked it the right way round: reverting the criterion to 1200 fails with `1500 not less than or equal to 1200`.

.github/workflows/test-helpers.yml
DevMomo: Reverted — the workflow file is now byte-identical to the base.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/check_customer_escalation_behavior.py
DevMomo: Fixed, with your version including the `sorted()`.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/escalation_is.py
DevMomo: Agreed on all of it, and thanks for checking the convention before flagging it. Leaving the values as they are here rather than deviating from the flow suite unilaterally, and raising the inconsistency separately — either the e2e fixtures move to env/gitignored config suite-wide, or `tests/README.md` records the exception explicitly. Not something to settle inside this PR.

DevMomo: Followed up on this rather than leaving it. Checked what the flow suite actually does: it commits human-readable identifiers and resolves ids at runtime — connection name, `FOLDER_PATH = "Shared/uipath-maestro-flow"`, `PROJECT_KEY = "CE"`, and a raw `SLACK_CHANNEL = "C0B2FDZD1M3"` with a channel-name comment. So the connection name, project key, issue type, Slack channel id and tenant guard here all match an existing flow constant and stay.

The folder key was the one real divergence. Flow never commits it — it matches `Folder` by name and never needs the GUID. The `_shared/test_flow_check.py:312` occurrence you found is inert fixture data inside a unit test, not a live lookup, so it was not precedent for using it as one. My comment justifying the key ("`connections list` reports keys, not paths") was also just wrong: `helpers/connections.ts:27-34` puts both `Folder` and `FolderKey` on every row.

Now matched by folder name with the key read off the resolved row, plus three tests including a same-named connection in another folder. Also swapped the live GUID out of the `is_real_connection_key` test, which only needs a well-formed UUID. No live folder key left in the suite.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/check_customer_escalation_behavior.py
```diff
@@ -0,0 +1,527 @@
...
+    UIPATH_NS,
+)
+
+# The prompt states this exact layout, so the pinned path is a stated
+# requirement — not the unstated-solution-name trap the flow suite's
+# discovery helper exists for.
+PROJECT = Path("CustomerEscalationTriageSolution") / "CustomerEscalationTriage"
+BPMN_FILE = PROJECT / "CustomerEscalationTriage.bpmn"
+# Ephemeral solution home, under the sandbox CWD (see module docstring).
+LIVE_RUN_DIR = Path(".customer-escalation-live")
```
rockymadden: 🔴 **Critical — the post_run sweep can never see this solution.**

The dot prefix hides the whole tree from `cleanup_solutions.py:49`, which globs `**/*.uipx` (Python's `glob` skips hidden dirs). Verified:

```
recursive glob:   ['visible/sub/b.uipx']
include_hidden:   ['visible/sub/b.uipx', '.customer-escalation-live/.../a.uipx']
```

So `uip solution delete` never runs and every execution leaves a registered solution on Alpha `codereval`. The YAML comment on the post_run step claims the opposite: *"the grader keeps its ephemeral solution under the sandbox CWD precisely so this glob finds the .uipx."*

Drop the dot — `Path("customer-escalation-live")` — and update the `.gitignore` entry. The checker runs after the agent finishes, so there's nothing to hide from. `include_hidden=True` also works (py ≥ 3.11) but only fixes the one caller.

tests/tasks/uipath-maestro-bpmn/_shared/cleanup_solutions.py
```diff
@@ -0,0 +1,135 @@
...
+        logger.warning(
+            "BPMN_E2E_CLEANUP=%r is invalid (expected always|never); treating as 'always'",
+            policy,
+        )
+        return "always"
+    return policy
+
+
+def main() -> int:
+    paths = glob.glob("**/*.uipx", recursive=True)
```
rockymadden, line 49: 🔴 Pairs with the `LIVE_RUN_DIR` comment: this glob silently skips dot-directories, so the new task's ephemeral `.uipx` is invisible to it.

Preferred fix is renaming the directory. If you'd rather harden here, `glob.glob("**/*.uipx", recursive=True, include_hidden=True)` (py ≥ 3.11, CI runs 3.13).

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/escalation_is.py
```diff
@@ -0,0 +1,326 @@
...
+# `FolderKey`, so the key is read off the matched row rather than committed.
+FOLDER_PATH = "Shared/uipath-maestro-flow"
+FOLDER_NAME = "uipath-maestro-flow"  # leaf of the path, as reported
+PROJECT_KEY = "CE"  # "Coder Eval" project on uipath-sandbox-380
+ISSUETYPE_ID = "11457"  # "Task" issue type, scoped to the CE project
+# NOT verified against the workspace: the flow suite pins C0B2FDZD1M3 for
+# "coding-agent-testing" on this same connection, so this id is a different
+# channel and the name is unconfirmed. Runs post here successfully; settle the
+# identity before adopting flow's name-to-id discovery step.
+SLACK_CHANNEL = "C01H4SPS77W"
```
rockymadden: 🟠 **Major — live writes to a channel nobody has identified.**

The comment is honest, which is what makes it a problem: this ships recurring Slack messages to an unnamed destination on a `path-to-ga` task, and if a delete ever fails the leak lands somewhere unknown.

The flow suite already pins a known target on this same connection:

```python
SLACK_CHANNEL = "C0B2FDZD1M3"  # coding-agent-testing
```

Use it, or resolve the id by name in `seed.py` the way `escalation_slack_alert` does.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/check_customer_escalation_behavior.py
```diff
@@ -0,0 +1,527 @@
...
+                f"no {connector_key} activity with a registry path "
+                f"containing {path_needle!r}"
+            )
+        return found
+
+    # Send identity, as the flow suite grades it: a node that posts as the
+    # default bot instead of the prompt-required user has an indistinguishable
+    # runtime response, so it can only be caught on the authored artifact.
+    slack_ids = set(ids_for(*SLACK_SEND))
+    for element in process:
```
rockymadden: 🟠 **Major — this scan is shallower than the one that produced `slack_ids`.**

`index_runtime_connectors` walks `process.iter()` deliberately ("registry templates may emit a connector activity as sendTask, serviceTask, or a plain task"). This loop only iterates direct children of `<bpmn:process>`, so a Slack node one level deep — inside a `subProcess` — lands in `slack_ids` but never reaches the `send_as` assertion. The grader then passes exactly the default-bot case this block exists to reject.

`classifier_ids` at line 151 has the mirror-image bug in the other direction: a `scriptTask` inside a `subProcess` isn't found at all, so a correct process fails contract resolution with `"no bpmn:scriptTask"` and forfeits the 5.0-weight criterion.

Use `process.iter()` in both places.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/test_behavior_checker.py
```diff
@@ -0,0 +1,662 @@
...
+            packager.is_real_connection_key(
+                "00000000-0000-0000-0000-000000000000"
+            )
+        )
+        self.assertFalse(packager.is_real_connection_key(""))
+        self.assertFalse(packager.is_real_connection_key("not-a-uuid"))
+
+
+if __name__ == "__main__":
+    unittest.main()
```
rockymadden: 🟡 `unittest.main()` here, with five test classes defined below it. Executed as a script, Python runs this line before those classes exist:

```
Ran 24 tests ... OK        # python test_behavior_checker.py
39 passed                  # pytest
```

Exit 0, 15 tests silently skipped. CI uses pytest so it stays green, but a local run reports success on 60% of the file. Move the block to the bottom.

Related: the module docstring says "Stdlib + pytest only (CI installs pytest and nothing else)" — this file imports `yaml`, which is why the workflow now installs `pyyaml`. Same stale claim in `bpmn_live.py`, and the PR body's "22 stdlib-only unit tests".

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/teardown_escalation.py
```diff
@@ -0,0 +1,69 @@
...
+    jira_connection = connections[escalation_is.JIRA_CONNECTOR]
+    for issue in dict.fromkeys(records.get("jira_issue", [])):
+        # Retry once on an unconfirmed (transient) failure, then confirm via a
+        # tenant reread before giving up -- only a confirmed deletion or a
+        # confirmed not-found counts. Mirrors the flow suite's teardown, and
+        # keeps a 5xx from leaking a real ticket in the shared CE project.
+        ok = escalation_is.delete_jira_issue(
+            jira_connection, issue, timeout=escalation_is.TEARDOWN_DELETE_TIMEOUT
+        )
+        if not ok:
```
rockymadden, line 38: 🟡 The retry's continuation lines aren't indented under the `if`. It parses, but it reads as a bug — worth aligning.

Two more in this sweep:
- Line 51: `channel_id, timestamp = record` is unguarded. One malformed journal record raises into the outer `except`, aborting the sweep and leaking every remaining id — the opposite of what the journal exists for. `read_journal` is already tolerant of bad lines; this consumer isn't.
- The Jira loop dedups with `dict.fromkeys`, the Slack loop doesn't. Lists aren't hashable, so it needs its own dedup (tuple-ise, or key on `(channel, ts)`).

tests/tasks/uipath-maestro-bpmn/_shared/bpmn_live.py
```diff
@@ -0,0 +1,440 @@
...
+from typing import Any
+
+BPMN_NS = "http://www.omg.org/spec/BPMN/20100524/MODEL"
+UIPATH_NS = "http://uipath.org/schema/bpmn"
+
+# Absolute monotonic deadline capping every CLI subprocess. A task assigns
+# `bpmn_live.ACTIVE_CLI_DEADLINE` while it owns live resources so a hung call
+# cannot eat the window before coder_eval SIGKILLs the grader; None disables
+# capping and leaves each call to its own timeout.
+ACTIVE_CLI_DEADLINE: float | None = None
```
rockymadden: 🟡 `ACTIVE_CLI_DEADLINE` is assigned by no task in the repo — the docstring says "a task assigns [it] while it owns live resources", and `grep -rn ACTIVE_CLI_DEADLINE tests/` returns only this file. So the deadline branch in `run_cli` is unreachable dead code in a brand-new file. Same for `runtime_variable_values` and `local()`.

Also on the module docstring: it says these helpers were "written three times — here, in the escalation grader, and privately inside `debug/live_debug_e2e/check_live_debug.py`". That third file is untouched by this PR and still carries its own `_parse_json_tolerant`, `_find_ci_key`, `_run_uip`, `_read_variables_all`. Either finish the port or drop the claim.

rockymadden (changes requested):
## What it does

Adds the BPMN suite's first **outcome-graded** live e2e. The agent authors a straight-line Process Orchestration project; the grader imports the exact submitted artifact (sha256-pinned) into an ephemeral solution, runs it via `uip maestro bpmn debug` with a per-run `correlationId`, then **re-reads Jira** to prove the ticket exists and carries that id.

Also lands the scaffolding: a shared `_shared/bpmn_live.py` CLI helper, a ported `cleanup_solutions.py`, and a ported 1,355-line criterion-budget guard.

Stacked on `fix/bpmn-connector-activity-authoring` (#3112).

| | |
|---|---|
| Files | 13 (+3885 / −1) |
| Weighted criteria | 5 (skill 1.0, self-validate 1.5, CLI validate 3.0, pack 2.0, **live run 5.0**) |
| Tests | 393 pass locally (39 in the new file) |

The core idea is right: grading the tenant instead of the narration is the correct answer to "validate passes, runtime does nothing."

---

## 🔴 Critical

**1. The ephemeral solution never gets cleaned up — leaks on every run**

`check_customer_escalation_behavior.py:68` puts the live solution in a **dot-prefixed** dir; `cleanup_solutions.py:49` looks for it with `glob.glob("**/*.uipx", recursive=True)`, which skips hidden directories. Verified:

```
recursive glob:   ['visible/sub/b.uipx']
include_hidden:   ['visible/sub/b.uipx', '.customer-escalation-live/.../a.uipx']
```

The YAML comment asserts the opposite — *"the grader keeps its ephemeral solution under the sandbox CWD precisely so this glob finds the .uipx."* It doesn't. Every run leaves a registered solution on the Alpha `codereval` tenant, and `path-to-ga` means this runs often.

**Fix:** drop the leading dot (update `.gitignore`). The checker runs after the agent finishes, so nothing needs hiding.

---

## 🟠 Major

**2. Writing to an unidentified Slack channel** — `escalation_is.py:54`

The comment admits `C01H4SPS77W` is unverified and the channel name unconfirmed. That honesty makes it worse: recurring live message writes to a destination nobody has named, on a `path-to-ga` task. If a delete fails, the leak lands somewhere unknown.

**Fix:** use `C0B2FDZD1M3` (`coding-agent-testing`), the channel every flow escalation task uses on this same connection — or resolve the id by name in `seed.py`.

**3. `send_as` and `scriptTask` scans are shallower than the connector scan** — `check_customer_escalation_behavior.py:151,182`

`index_runtime_connectors` walks all descendants (deliberately). Both of these iterate direct children of `<bpmn:process>` only. Two consequences:

- A Slack node one level deep (`subProcess`) is found by `ids_for` but its `send_as` check is **silently skipped** — the grader passes something it means to reject.
- A `scriptTask` inside a `subProcess` fails contract resolution with `"no bpmn:scriptTask"` — a **correct** process forfeits the 5.0-weight criterion.

**Fix:** use `process.iter()` in all three places.

---

## 🟡 Minor

**4. `unittest.main()` sits mid-file** — `test_behavior_checker.py:435`. Five test classes are defined after it:

```
Ran 24 tests ... OK        # direct
39 passed                  # pytest
```

15 tests silently skipped, exit 0. CI uses pytest so it's green, but a local run lies.

**5. Stale "stdlib only" claims** — `test_behavior_checker.py:1`, `bpmn_live.py:16`. Both say CI "installs pytest and nothing else"; the same PR adds `pyyaml` because the test imports `yaml`. PR body says "22 stdlib-only unit tests" — it's 39, and not stdlib-only.

**6. Dead machinery in a brand-new shared file** — `ACTIVE_CLI_DEADLINE` is assigned by no task in the repo, so `run_cli`'s deadline branch is unreachable. `runtime_variable_values` and `local()` are also unused.

**7. `bpmn_live.py` docstring overstates the consolidation** — it claims extraction from three sites including `debug/live_debug_e2e/check_live_debug.py`, which is untouched and still carries `_parse_json_tolerant`, `_find_ci_key`, `_run_uip`, `_read_variables_all`. Port it or fix the sentence.

**8. `teardown_escalation.py:38-41,51`** — the retry call's continuation lines are mis-indented (parses, reads as a bug). The Slack loop unpacks `channel_id, timestamp = record` unguarded: one malformed journal line raises into the outer `except`, aborting the sweep and leaking every remaining record. Jira dedups via `dict.fromkeys`; Slack doesn't.

**9. Unused imports** — `check_customer_escalation_behavior.py:41`: `parse_json_output`, `tail_log` (pyflakes).

**10. 1,355 duplicated lines, and the two copies disagree.** `_shared/test_criterion_budgets.py` is a near-verbatim copy of the flow suite's, guarding one live criterion here. Consistent with the "each suite ships its own `_shared`" rule, so I won't fight it — but the two files now contradict each other:

- Shared guard: `task_timeout > grading`, with a comment saying `grading + turn_timeout` *"broke on smoke.yaml"*.
- `test_behavior_checker.test_task_timeout_covers_the_turn_plus_all_grading`: re-imposes exactly `turn_timeout + grading <= task_timeout`.

Pick one rule.

**11. Comment density.** Roughly one comment per two lines of logic, and a lot of it is review-history narration rather than reader-facing fact: *"An earlier version raised on the second occurrence…"*, *"Every round of review on this guard found a live under-budgeted criterion."* The CLI-quirk and RCA notes earn their space. The autobiography doesn't.

---

## 🔵 Comment (no change required)

**Prompt is heavily prescriptive.** It dictates the directory layout, file name, both connection names, the folder, output names/types, the severity rules, the summary format, and `send_as: user`. The YAML defends the pinned path, and that's fair — but taken together it's closer to a recipe-following eval than a skill-discovery one. Contrast the flow suite, whose prompt names only the flow. Worth being explicit about what this eval *doesn't* cover.

---

## 🚦 CI

| Check | Status | Note |
|---|---|---|
| `path-to-ga approval` | ❌ | Procedural — needs an allowlisted approver |
| `Run skill smoke tests` | ❌ | 91.7% < 95%, from `skill-bpmn-operate-diagnose-minimal-fault-triage` — **unrelated to this diff**, but blocks merge |
| Everything else (27) | ✅ | incl. `maestro-bpmn checker unit tests` |

---

## Recommendation: 🔴 Request changes

Findings 1 and 2 are both live-tenant hygiene on a task tagged for frequent GA-gating runs: a solution leaked on every execution, and messages posted to a channel the author says they can't identify. Both are a few lines. Finding 3 is a grading hole plus a false-negative in the same function.

The design is sound and worth landing — this is the right shape of eval. It just shouldn't go in leaking.

---

### tl;dr

Good eval. It checks the tenant instead of the agent's story, which is what the BPMN suite was missing. Three things to fix before merge: the cleanup never finds the solution it's supposed to delete (hidden directory vs. Python's glob), it posts Slack messages to a channel nobody has identified, and one grading check quietly skips nested nodes. Everything else is small. CI failures are procedural plus one unrelated flaky task.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/check_customer_escalation_behavior.py
DevMomo: Dropped the dot, and the .gitignore entry with it. Reproduced your glob result before and after: hidden dir invisible, `customer-escalation-live/S/a.uipx` visible. Every run so far has been leaving a registered solution on Alpha.

tests/tasks/uipath-maestro-bpmn/_shared/cleanup_solutions.py
DevMomo: Took the preferred fix and renamed the directory rather than hardening here, so every caller benefits rather than just this one.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/escalation_is.py
DevMomo: Using the flow suite's `C0B2FDZD1M3  # coding-agent-testing`. I had tried to resolve the old id via the connector and could not, which is why the comment was hedged — pointing at the identified target is the right answer rather than shipping writes to an unnamed channel.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/check_customer_escalation_behavior.py
DevMomo: Both use `process.iter()` now, matching `index_runtime_connectors`. Verified the mirror-image bug you predicted: a `scriptTask` nested in a `subProcess` previously failed contract resolution, and now resolves.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/test_behavior_checker.py
DevMomo: Moved to the bottom. Direct run went from `Ran 24 tests` to `Ran 39`, matching pytest. Also corrected the stale stdlib-only docstring — this file does need PyYAML.

tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/teardown_escalation.py
DevMomo: All three: indentation aligned, Slack records deduped on `(channel, ts)`, and a malformed record is now skipped with a WARN instead of raising into the outer handler and stranding every id after it.

tests/tasks/uipath-maestro-bpmn/_shared/bpmn_live.py
DevMomo: Removed `ACTIVE_CLI_DEADLINE`, `runtime_variable_values` and `local()` — all confirmed unreferenced, and the deadline docstring promised protection nothing enabled. Corrected the provenance claim too: `check_live_debug.py` still has its own six copies, so the port is unfinished and the docstring now says so rather than implying otherwise.

rockymadden (approved):

PR #3124, opened 2026-09-07
feat(skill-flavors): studioweb flavor for the Flow skill — audit fixes verified live on alpha

astridstaicu:
## What

Full audit of `uipath-maestro-flow` as composed for the `studioweb` flavor, against what the Studio Web host actually intercepts, mounts and forbids — then every finding fixed as marker blocks + sparse overrides. 59 canonical files gain markers only (whitespace-identical otherwise); 61 override files (2757 lines) under `skill-flavors/studioweb/uipath-maestro-flow/`; the studioweb build goes from 264 to 524 replacements.

Follow-up to #3065 (which removed `uip solution init`). The audit found the same class of dead end in ten more places, and that sixteen of the existing flow-skill overrides were themselves wrong.

## Root causes fixed (each is one override rule applied everywhere it appeared)

| # | Wrong in the composed skill | Studio Web truth (verified live on alpha, 2026-09-07) |
|---|---|---|
| R1 | Existing overrides teach `CreateProjects` for Flow creation (SKILL.md rule 6, author/CAPABILITY, greenfield T0 turn, file-format) | `uip flow init <ProjectName>` is host-intercepted; prints `Created Flow project "<Name>" … under /solution/<Name>`; scaffold = `new.flow` (manual trigger, `version` 1.9) + `project.uiproj`. No `.uipx`, no `solution projects add`. Creation now chains into the T1 Bash — no separate T0. |
| R2 | `<ProjectName>.flow` ×59, `.uipx` ×46, `<Solution>/<Project>` nesting, `operate.json`/`package-descriptor.json` as scaffold files | Flat `/solution/<Project>/new.flow`; `entry-points.json`, `new.bpmn`, `operate.json`, `package-descriptor.json`, `bindings_v2.json`, `simulations.json`, `evals/` are host artifacts that appear after the first debug/publish — never hand-edited. |
| R3 | 67 `uip login` mentions, whole cli-conventions §5 login table, `uip login status` checks | `login|logout|auth|config` are host no-ops and the agent refuses them; auth is injected; 401/403 = missing rights. |
| R4 | cli-conventions §1 probe-and-install script, greenfield Step 0 probe, `eval --help` probe, `uip tools update` ×7, `uip --version` | Probing/installing is forbidden by the host; the CLI version is fixed. |
| R5 | `uip maestro flow debug <project-dir> --output json`, `--attachment`, JSON fields (`Data.studioWebUrl`, `Data.instanceId`, `Data.incidents[]`, `finalStatus`), "debug overwrites the `.uipx` solution", background/poll mechanics | Only two-token `uip flow debug [<ProjectName>] [--inputs '<json>']` is intercepted (three-token → `… is not a constructor`). Plain-text output: `<status>` / `Trace ID:` (when known) / `Run logs:` / `Execution trace:`; exit 1 on failure; `timeoutSeconds: 600`; `TimedOut after 300s` + `(no run logs emitted)` = run never started → hand to the designer. `--attachment` is ignored. |
| R6 | Existing overrides mandate `uip solution resources refresh` before every debug | Declined to the bundle → `/solution is not a UiPath solution (no .uipx manifest found)`. Resources are host-served: `uip solution resources list|get|add|edit`. |
| R7 | `registry list|search|get --local` for in-solution siblings (SKILL rule 3b + 12 plugin files) | Needs a `.uipx` → `No .uipx solution file found`. Replacement: `uip solution resources list --kind Process --output json` — `solutionResources` lists the open solution's projects with their keys. |
| R8 | Existing publish override: "with several destinations it publishes nothing and lists them"; `solution upload`/`deploy`/`pack`, `solution projects add`, `codedagent deploy`, "Publish to Studio Web" dropdowns | With no `--location` the host publishes to the personal workspace immediately (auto-deploys). `uip solution publish --help` lists `PublishLocations`; ask, then pass `--location`. upload/deploy refused, pack is a no-op. |
| R9 | `python3` heredocs ×5, `crypto.randomUUID()`, `uuidgen`, eval `--wait --timeout 600` (= bash tool max) | No python (`command not available in browser environments`), no `crypto`, no `uuidgen`, no `yq`/`sqlite3`; `jq` + `node` only; `node /tmp/script.js <file>` works. UUID via `node -e` Math.random v4. |
| R10 | eval run ids "auto-resolved" from `.uipx`/`SolutionStorage.json` | Pass `--solution-id <CurrentSolution.SolutionId> --project-id <CurrentProject.ProjectId>`; `eval run list` returned with ids (and hung on first use twice) → `timeoutSeconds: 120`, never `--wait`, hand a time-out to the Evaluations panel. |
| R11 | Canonical text composing around overrides and contradicting them (debug reporting, pack body, duplicate `---`, renamed anchors) | Blocks widened; anchors kept or fixed; `check-links` green. |

## Live verification (alpha, tenant AndreiTest, Solution 1420, driven through the real Studio Web Autopilot chat)

Four batches of shell commands, outputs captured verbatim. Confirmed: `uip flow init` scaffold contents; `registry --local` and `resources refresh` failures; `solution resources list --kind Process` exposing in-solution projects; `publish --help` `PublishLocations`; bundle `uip maestro flow debug` → `ge is not a constructor`; host `uip flow debug` output shape (both runs `TimedOut after 300s` with no logs on the current alpha host); `job traces <bad-key>` hangs; `eval run list` with ids returns; `validate`/`format` on `new.flow`; `uip agent init … --inline-in-flow` works; `rm` under `/solution` → `Operation not permitted on this mount`; `node -e` UUID works, `require('crypto')` fails, `python3`/`yq`/`sqlite3`/`uuidgen` absent; `--output-filter` on `registry list`/`uip or … list` requires explicit `--limit`; `uip is connections list` has no `--limit`.

## Guard

`tests/scripts/studioweb-flavor-contract.test.mjs` gains a Flow-skill test: no fenced command in the composed studioweb Flow skill may be `uip login`, `uip solution resources refresh|upload|pack|deploy|projects add`, `uip maestro flow debug|pack`, `uip tools update`, `python3`, `uuidgen`, or `registry … --local`; prose may not mention `CreateProjects` or `crypto.randomUUID`. Non-vacuity check against the default flavor included.

## Checks

`npm run skills:validate` OK (default 0 / studioweb 524 replacements) · `npm run skills:check-links` all 6825 links resolve · `skills:test` 56/57 (the known environment-only `npm pack dry-run` regex failure, present on untouched main).

## Not in this PR (canonical / other repos)

- Canonical `--output-filter` examples on `registry list` / `uip or folders list` lack the now-required `--limit` (CLI change) — affects every flavor.
- Autopilot `studioWebContextSkill.ts` claims `yq`, `sqlite3` and a per-project `entry-points.json` exist; none do on a fresh project.
- On the current alpha host `uip flow debug` timed out without starting a run twice; the flavor now tells the agent to hand debug to the designer in that case, but the host path deserves a look.
- `job traces <unknown-key>` hangs until the tool timeout instead of failing fast (CLI).


astridstaicu:
Addressed the review:

- **`uip flow publish` vs `uip solution publish`** — `flow:publish` is a host-intercepted alias of the same publish handler (`subcommandInterceptors/index.ts` maps `flow:publish`, `agent:publish` and `solution:publish` to `publishInterceptor`), so the six mentions were not broken. Normalised them to `uip solution publish` anyway in 0d8cdfd40 so the flavor teaches one form everywhere.
- **Out-of-scope `uipath-platform` / licensing changes** — not part of this PR. The diff is 100 files, all under `skills/uipath-maestro-flow` and `skill-flavors/studioweb/uipath-maestro-flow` (merge base `bf65a37d0`); the listed files come from commits that landed on `main` after the branch point. Likewise there is no `objectName → activityName` or HITL-row content change in the canonical files: `git diff -w` against the base shows marker lines only, plus an EOF newline in four files.
- **Debug `TimedOut after 300s` with no logs** — agreed this is a host-side question; it reproduced on two projects and is flagged in the description for the Autopilot team. The flavor now tells the agent to hand debug to the designer in that case rather than loop.
- **coder_eval coverage** — acknowledged as a gap; the studio-web harness still mounts the `default` flavor, so behavioural tasks for the flavor are a follow-up once it mounts `studioweb`.

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/impl.md
```diff
@@ -0,0 +1,54 @@
...
+
+<!--skill-flavor:agent-impl-discovery:start-->
+Run these commands for published agents (auth is host-provided; a 401/403 means the signed-in user lacks rights on the tenant):
+
+```bash
+uip maestro flow registry pull --force
+uip maestro flow registry search "uipath.core.agent" --output json
+```
+
+Only published agents from the tenant appear. For in-solution agents, list the open solution's projects — `registry list|get --local` needs a `.uipx` and fails in Studio Web:
```
astridstaicu: we don't have uipx in studio web

astridstaicu: Fixed in 495e496f4. The line now leads with the fact rather than the flag: it says there is no `.uipx` in Studio Web, so the `--local` registry lookups do not exist here, and it quotes the error the CLI actually prints (`No .uipx solution file found`) before pointing at the solution-resources listing as the replacement.

I kept the term itself rather than deleting it, because that error string is what the agent sees if it ever reaches for the `--local` form, and naming it is what lets it recognise the failure instead of retrying. Every remaining `.uipx` mention in the flavor is negative in exactly that way — there is no positive `.uipx` guidance left in the composed studioweb tree.

The same commit closes the other way to land on that bundle path: eleven sites were telling the agent to run a kindless `uip solution resources list`, which the host only serves with `--kind` (the source flag defaults to all), so the bare form declined to the browser bundle and failed for want of a manifest. All of them now pass a kind, and the contract test fails on the kindless form from now on.

skill-flavors/studioweb/uipath-maestro-flow/.maintenance/README.md
```diff
@@ -0,0 +1,8 @@
+<!--skill-flavor:maintenance-capability-boundary:start-->
+- **Author** = on disk in the open solution, no login step at all (`flow init`, `validate`, `format`, registry, JSON edits) — the host injects auth on every `uip` call
+- **Operate** = touches the cloud (`flow debug` through the host, `solution publish`, `process run`, `instance ...`); there is no `solution upload` and no local pack in Studio Web
+- **Diagnose** = postmortem on a failed run (`instance incidents`, `instance variables`, `instance asset`, `incident get`, `job traces`)
```
astridstaicu: Do these diagnose stuff make sense for studio web? if the run happens in the designer, do we have a job ran to instance incidents and the others?

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/impl.md
```diff
@@ -0,0 +1,54 @@
+<!--skill-flavor:agent-impl-kinds:start-->
+Agents are either:
```
astridstaicu: why do we need this change in this set of skills? seems pretty much the same for me

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/impl.md
```diff
@@ -0,0 +1,54 @@
...
+
+- **In this solution**: a sibling project of the open solution. `{key}` is the sibling's resource `key` from `uip solution resources list --kind Process --output json` (`solutionResources`); registration into the solution is automatic when the project is created (`uip agent init <Name>`), and runtime resolution uses the open solution's Resources panel. `definitions[]` uses `model.section: "In this solution"`.
+- **Published**: an Orchestrator tenant resource. `{key}` is the Orchestrator-assigned resource key, discoverable with `uip maestro flow registry search`; `definitions[]` uses `category: "agent.published"` (registry output omits `model.section`).
+<!--skill-flavor:agent-impl-kinds:end-->
+
+<!--skill-flavor:agent-impl-discovery:start-->
```
astridstaicu, line 5: again, I don't think we need this change, we already tell the user not to run uip login in the bigger studio web prompt

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/impl.md
```diff
@@ -0,0 +1,54 @@
...
+
+```bash
+uip maestro flow registry get "uipath.core.agent.{key}" --output json
+```
+
+It shows only published tenant agents. Confirm from `registry get`:
+<!--skill-flavor:agent-impl-discovery:end-->
+
+<!--skill-flavor:agent-impl-in-solution-definition:start-->
+Set `display.icon` by inspecting the sibling project root: use `"autonomous-agent"` when `agent.json` exists, otherwise `"coded-agent"`.
```
astridstaicu, line 24: this change seems not needed too, much too similar to the original one

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/impl.md
```diff
@@ -0,0 +1,54 @@
...
+
+Never hand-author `definitions[]`. In Studio Web the registry serves no manifest for an unpublished sibling (`registry get --local` needs a `.uipx` and fails here): add the in-solution agent node from the designer's add-node panel, which writes the definition into `new.flow`, then continue editing that file. Read `<resourceKey>` from the sibling's `key` in `uip solution resources list --kind Process --output json` (`solutionResources`); there is no `resources/solution_folder/` in Studio Web. For a published agent, read `<DEFINITION_VERSION>` from `.version` in `registry get` output.
+<!--skill-flavor:agent-impl-in-solution-definition:end-->
+
+<!--skill-flavor:agent-impl-create:start-->
+Create it before wiring:
```
astridstaicu: lets scope this to only the exact line or paragraph that needs changing, meaning to how agents are created with agent init

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/impl.md
```diff
@@ -0,0 +1,54 @@
...
+
+- **In-solution sibling, low-code**: `uip agent init <Name>` creates the agent project in the open solution (registration is automatic — no `.uipx`, no `uip solution projects add`); configure it with `uipath-agents`, then read its `key` from `uip solution resources list --kind Process --output json`.
+- **Coded agents**: `uip codedagent init|deploy|run` are not available in Studio Web — coded agents cannot be scaffolded or run from the browser. Plan a published one (deployed from a developer machine), then `uip maestro flow registry pull --force`. See [coded/embedding-in-flows.md](../../../../../uipath-agents/references/coded/embedding-in-flows.md).
+- **Published low-code**: the agent is published with its solution — for the open one, `uip solution publish --location "<FolderPathOrKey>"` (destination from `uip solution publish --help` PublishLocations) — then `uip maestro flow registry pull --force`.
+<!--skill-flavor:agent-impl-create:end-->
+
+<!--skill-flavor:agent-impl-debug-table:start-->
+| Error | Cause | Fix |
```
astridstaicu: again, seems to be too similar to the original one, lets do minimal changes to the skills please

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/planning.md
```diff
@@ -0,0 +1,28 @@
+<!--skill-flavor:agent-intro:start-->
+Agent nodes invoke UiPath AI agents from within a flow. Published agents appear in the registry after `uip maestro flow registry pull` (auth is host-provided). **In-solution** (unpublished) agents in sibling projects of the open solution are listed with `uip solution resources list --kind Process --output json` — no publish required. Both **coded** (Python) and **low-code** (agent.json) agents appear here once deployed — the flow treats them identically.
```
astridstaicu: again too much of a change for just uip login

skill-flavors/studioweb/uipath-maestro-flow/references/author/plugins/agent/planning.md
```diff
@@ -0,0 +1,28 @@
...
+<!--skill-flavor:agent-intro:end-->
+
+<!--skill-flavor:agent-when-not-local:start-->
+- **Agent in the same solution but not yet published** — find it with `uip solution resources list --kind Process` (see below)
+- **Agent does not exist yet** — create it in the open solution with `uip agent init <Name>` (configure it with `uipath-agents`), then find it with `uip solution resources list --kind Process --output json`
+<!--skill-flavor:agent-when-not-local:end-->
+
+<!--skill-flavor:agent-discovery:start-->
+**Published (tenant registry):**
```
astridstaicu: if this override is again only for the login, lets drop it

PR #3126, opened 2026-09-07
fix(uipath-coded-apps): make the React scaffold self-verifying [APPS-37265]

amrit-agarwal-1:
Fixes [APPS-37265](https://uipath.atlassian.net/browse/APPS-37265).

An agent following the skill produced a vanilla TypeScript app instead of React: it typed an en dash
instead of the bare `--` separator, npm swallowed `--template`, and `create-vite` scaffolded its
default template and still exited `0`.

The skill sources were not corrupted, but two live defects made that outcome possible and invisible:

- `create-action-app.md` taught the `npm create vite@latest <app> -- --template react-ts` form. That
  bare `--` is load-bearing and cannot be fixed upstream in `create-vite`, so the form is replaced
  rather than repaired.
- `create-web-app.md`'s `npx` form is also broken today, for an unrelated reason: on create-vite 9
  under a pty it prompts for a linter and creates nothing at all. `--no-interactive` is required.
- Nothing verified the scaffold, and nothing in this skill installs `react` explicitly, so a wrong
  template first surfaced minutes later as an unrelated-looking build failure.

Both files now carry one byte-identical command ending in a `node` assertion on `src/main.tsx` that
prints `SCAFFOLD_OK`. The assertion runs from the parent directory, so a failure leaves cwd where the
recovery can act. Recovery is `--overwrite`: a bare re-run prints `Operation cancelled`, exits `0` and
changes nothing, which would loop an agent forever. Critical Rule 20 states the mechanism once; both
reference files point at it.

Also corrects two verified falsehoods on the scaffold path — that `create-vite` gitignores `.uipath/`,
and that `uip codedapp init` is "for solutions only" (it has a standalone mode; it just writes no
source project). Adds a live-scaffold regression task whose gating criteria include "did not run
`npm create vite`".

Measured on create-vite 9.2.0 / npm 11.17.0: the en dash reproduces the vanilla tree, the gate catches
it (`SCAFFOLD_NOT_REACT`, exit 1, cwd unchanged), a bare re-run is an exit-0 no-op, and `--overwrite`
recovers in one command. `skills:validate` at baseline; all 6713 relative links resolve.

A second task, `scaffold_recovery_overwrite_smoke.yaml`, covers the recovery path: `pre_run` seeds a
vanilla-TypeScript tree and the criteria grade the artifacts after re-scaffolding. A bare re-run over
the occupied directory prints `Operation cancelled` and exits `0`, so an agent that does not clear it
leaves the vanilla files and fails. The `--overwrite` check is advisory — clearing the directory
another way reaches the same artifacts.

Both tasks ran under coder-eval on this branch (agent `codex`, pinned framework version) and passed
`1/1 PASS (linux)`:

- `skill-coded-apps-scaffold-react-template-smoke` —
  [run](https://github.com/UiPath/skills/actions/runs/34469566586) ·
  [evalboard](https://coder-evalboard.uipath-dev.com/runs/2026-09-10_11-11-24-gha-34469566586-1?src=gha)
- `skill-coded-apps-scaffold-recovery-overwrite-smoke` —
  [run](https://github.com/UiPath/skills/actions/runs/34931118804) ·
  [evalboard](https://coder-evalboard.uipath-dev.com/runs/2026-09-15_05-08-08-gha-34931118804-1?src=gha)

That also settles the open question about the tier: these are the first coded-apps smoke tasks to run
`npx` as a graded action, and their `file_exists` and `package.json` criteria could only pass if
`create-vite` reached the registry, so smoke-tier egress works and no retag to `integration` is needed.

[APPS-37265]: https://uipath.atlassian.net/browse/APPS-37265?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amrit-agarwal-1:
Ran the new task under coder-eval on this branch (agent `codex`, pinned framework version): `skill-coded-apps-scaffold-react-template-smoke: SUCCESS` — `1/1 PASS (linux)`.

[Actions run](https://github.com/UiPath/skills/actions/runs/34469566586) · [evalboard](https://coder-evalboard.uipath-dev.com/runs/2026-09-10_11-11-24-gha-34469566586-1?src=gha)

This is the first coded-apps smoke task to run `npx` as a graded action, and its `file_exists` / `package.json` criteria could only pass if `create-vite` reached the registry — so smoke-tier egress works and the task does not need retagging to `integration`.

deepeshrai-tech (commented):
Left a few clarifications

skills/uipath-coded-apps/references/create-web-app.md
```diff
@@ -64,9 +64,11 @@ Map the `<environment>` answer from Step 1 to a base URL using the table in [SKI
 ### 3.2 — Create the Vite project
 
 ```bash
-npx --yes create-vite@latest <app-name> --template react-ts
+npx --yes create-vite@latest <app-name> --no-interactive --template react-ts && node -e "require('fs').existsSync('<app-name>/src/main.tsx')||(console.error('SCAFFOLD_NOT_REACT'),process.exit(1))" && echo SCAFFOLD_OK
 ```
 
+Run that as **one** command — `SCAFFOLD_OK` is the only signal that the scaffold is really React (Critical Rule 20). On `SCAFFOLD_NOT_REACT`, or any non-zero exit, re-run the same line with `--overwrite` added immediately after `--no-interactive`; a bare re-run exits `0` and changes nothing. Stop after two failed attempts and report the output to the user rather than continuing past this step.
```
deepeshrai-tech, line 70: Some create-vite v9 releases treat `--overwrite` as a value-taking option (`--overwrite yes|no|ignore`). If that's true for 9.2.0, then adding just `--overwrite` before `--template react-ts` would make the parser eat `--template` as the value — which is the exact silent-vanilla-TS failure Rule 20 exists to prevent, only in the recovery path this time. The smoke task only exercises the happy path, so it wouldn't catch this.

skills/uipath-coded-apps/SKILL.md
```diff
@@ -58,6 +58,7 @@ For the app's **JS/TS function backend** — authoring the `defineFunction` endp
...
+20. **A scaffold is not React until `src/main.tsx` exists — assert it before installing anything.** Scaffold both app types by copying the single `create-vite` command from [create-web-app.md](references/create-web-app.md) Step 3.2 or [create-action-app.md](references/create-action-app.md): it passes `--no-interactive` and ends in a `node` check that prints `SCAFFOLD_OK`. Never use an `npm create vite` form — `npm create` needs a bare `--` separator to forward flags, and a lost or corrupted separator makes npm swallow `--template` so `create-vite` silently scaffolds its **vanilla TypeScript** template. Never judge a scaffold by its exit code or output: `create-vite` exits `0` on a misparsed template, on a linter prompt it could not show, and on `Operation cancelled` over a non-empty target directory. Nothing in this skill installs `react` explicitly, so a wrong template surfaces minutes later as an unrelated-looking build failure. On failure re-scaffold with `--overwrite`; never convert a vanilla scaffold into a React one by hand.
```
deepeshrai-tech, line 61: `--no-interactive` implicitly (via the copy-command), but doesn't name it or explain what it prevents. An agent skimming just this rule to build the command from memory (which the rule tells them not to do, but they might anyway) would miss the linter-prompt failure mode you called out in the PR body.

tests/tasks/uipath-coded-apps/scaffold_react_template_smoke.yaml
```diff
@@ -0,0 +1,79 @@
...
+    command: "sh -c 'test ! -f intake-report/src/main.ts && test ! -f intake-report/src/counter.ts'"
+    timeout: 10
+    expected_exit_code: 0
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Rule #20: agent did NOT use an `npm create vite` form (its bare `--` separator is the APPS-37265 failure)"
+    tool_name: "Bash"
+    command_pattern: 'npm\s+create\s+vite'
```
deepeshrai-tech, line 55: This task only grades the happy-path scaffold. The recovery (`--overwrite`) is the part agents will actually reach for when things break, and it's not covered anywhere.

skills/uipath-coded-apps/references/create-web-app.md
amrit-agarwal-1: Checked this empirically — `--overwrite` is boolean in every release I can run, so it does not consume `--template`.

`--help` declares it without a value placeholder (contrast `-t, --template NAME`), and that holds for 9.0.5 through 9.2.1; 9.0.0–9.0.4 have no `--overwrite` at all, and v8 does not either. Running the exact recovery line over a stale vanilla tree on 9.2.1 produces `App.tsx`/`main.tsx` and `SCAFFOLD_OK`. Even on 9.0.0, where the flag is unknown, it is ignored and `--template react-ts` still parses.

I think the `yes|no|ignore` triple comes from the interactive prompt rather than the flag — the source has those exact values on the non-empty-directory prompt (`Cancel operation` → `no`, `Ignore files and continue` → `ignore`). `--overwrite` is what skips that prompt.

Your underlying point stands though: nothing was exercising the recovery path, so a future release changing its arity would go unnoticed. Covered now — see the new task on your third comment.

skills/uipath-coded-apps/SKILL.md
amrit-agarwal-1: Fixed. Rule 20 now names the failure:

> it passes `--no-interactive` — without which `create-vite` stops on a linter prompt, and exits `0` having created nothing when no one can answer it — and ends in a `node` check that prints `SCAFFOLD_OK`.

Verified under a real pty: EOF at the prompt gives exit code `0` with an empty target directory. With nothing answering at all it hangs instead.

tests/tasks/uipath-coded-apps/scaffold_react_template_smoke.yaml
amrit-agarwal-1: Added `tests/tasks/uipath-coded-apps/scaffold_recovery_overwrite_smoke.yaml`. `pre_run` seeds a vanilla-TypeScript tree and the prompt asks for the correct React scaffold at the same path.

Validated locally: the three gating criteria all fail on the seeded state, all pass after the documented recovery, and a bare re-run without `--overwrite` prints `Operation cancelled`, exits `0` and leaves the vanilla files — so the task fails exactly as it should.

The `--overwrite` check itself is advisory (`pass_threshold: 0.0`): clearing the directory another way reaches the same artifacts, and test-writing.md grades outcomes rather than literal flags. Eval run in flight: https://github.com/UiPath/skills/actions/runs/34931118804

skills/uipath-coded-apps/references/create-web-app.md
amrit-agarwal-1: Checked v8 and all 13 v9 releases: --overwrite is boolean everywhere, so it never eats --template, the yes|no|ignore triple is the interactive prompt's choice values, not the flag's arity.

skills/uipath-coded-apps/SKILL.md
amrit-agarwal-1: Fixed, Rule 20 now says the flag is what stops create-vite halting on a linter prompt and exiting 0 having created nothing, verified under a real pty.

tests/tasks/uipath-coded-apps/scaffold_react_template_smoke.yaml
amrit-agarwal-1: Added scaffold_recovery_overwrite_smoke.yaml: it seeds a vanilla tree, and its gating criteria fail on the seed, pass after recovery, and fail on a bare re-run without --overwrite

deepeshrai-tech (approved):

PR #3127, opened 2026-09-07
fix(governance): resolve compliance standard from catalog, not hardcoded premises

sauravrathore01:
Passed Run on Claude: https://github.com/UiPath/skills/actions/runs/34178508187
Passed Run on Codex: https://github.com/UiPath/skills/actions/runs/34146716907
Passed Run on Antigravity: https://github.com/UiPath/skills/actions/runs/34179447353

## What

Two facts about tenant state were hardcoded into documents by #2706. Both went stale, and between them they made `unnamed_standard_smoke.yaml` grade the opposite of what it intended: it **passed a run that hallucinated an auth failure** and **failed a run that behaved correctly**.

## Evidence

Two recorded runs of the same task, same tenant, same catalog:

| | `gpt-5.6-terra` | `claude-sonnet-5` |
|---|---|---|
| Recorded score | **1.000 SUCCESS** | **0.654 FAILURE** |
| What it actually did | claimed "no active tenant session", bailed out | read the catalog, reported the one available pack |
| Was it correct? | **No** — `output.txt` shows `"Status": "Logged in"` | **Yes** |

`output.txt` from both runs: every `uip` command returned `"Result": "Success"`, **zero** 401/403, and exactly **one** pack with `Available: true` (ISO 42001). ISO 27001, GDPR, HIPAA, SOC 2, EU AI Act are all `Available: false`.

### Premise 1 — "more than one pack is available"

`SKILL.md` Rule 9 and `catalog/impl.md` demanded the agent ask *"which standard?"* — a question with one answer on this tenant. The test failed any agent that read the catalog and reported the only posture that exists.

**Fix:** Rule 9 branches on the `Available: true` count read at runtime. Two or more → ask. Exactly one → proceed, naming it explicitly as the only available standard. None → say so and stop.

### Premise 2 — "auth errors expected"

```yaml
initial_prompt:
  - uip CLI is not connected to a live tenant - auth errors expected.
```

The same file's `post_run` comment says the opposite: *"smoke sandboxes have live tenant auth in CI"*.

This primed agents to hunt for an auth failure that never happens. `gpt-5.6-terra` duly found one — a grep matching the word `logged` inside `"Status": "Logged in"` — declared the tenant unreachable, and scored 1.000, because the judge rubric explicitly blessed bailing out on an auth error.

**Fix:** removed the false premise (kept the do-not-retry guard) and inverted the rubric bullet — a reply claiming it could not connect now contradicts the recorded output and scores 0.0.

## Test changes

- `command_not_executed` now bans pack-scoped commands against packs the catalog marks **unavailable**, scoped to one command segment, instead of banning every pack-scoped command. Demoted to weight 1.0 — it is a guard, not a discriminator, and cannot fire while one pack is available.
- `catalog list` gains `require_success: true`. "Typed the command" is not "read the catalog", and unlike a `file_contains` check this inspects the command result and cannot be faked by echoing a string.

## Projected outcome

Verified by replaying the new regex against both recorded traces (11.0 total weight):

| Run | Before | After |
|---|---|---|
| `gpt-5.6-terra` | 1.000 SUCCESS | **0.727 FAILURE** |
| `claude-sonnet-5` | 0.773 FAILURE | **0.864–1.000 SUCCESS** |

## Verification status

**Not run against a live tenant** — no coder-eval harness on the authoring machine (`uv`, `coder-eval`, `uip`, `docker`, tenant auth and judge keys all absent). The score figures above are projections from trace replay, not measurements. **Please run before merge:**

```bash
cd tests && .venv/bin/coder-eval run \
  tasks/uipath-governance/compliance-pack/unnamed_standard_smoke.yaml \
  -e experiments/default.yaml -D agent.model=<model> -v
```

Once each on `claude-sonnet-5` and `gpt-5.6-terra` to confirm both flip.

Static checks pass: `skills:validate`, `skills:check-links`, `check-cli-verbs.py`, `/lint-task`.

## Known gaps

- A pack id held in a shell variable (`PACK=iso-27001; ... $PACK`) evades the segment-scoped regex. No agent has done it — YAGNI until one does.
- `check_routing_smoke.yaml` line 10 carries the **same false auth premise**. Same folder, same owners, same tenant. Flagged for @UiPath/authz rather than swept in here; 11 more compliance-pack tasks likely share it.
- 24 further task files across `uipath-process-mining`, `uipath-insights`, `uipath-tasks`, `uipath-ixp` and `uipath-aops` match the same grep, but belong to other CODEOWNERS teams and several are correct by design (they grade flag construction against fake IDs, where auth rejection is the point). Deliberately out of scope.


jianjunwang2 (approved):

skills/uipath-governance/SKILL.md
```diff
@@ -54,12 +54,12 @@ Activate on **any** governance / policy / rule intent — even when the user did
...
+9. **Never run a pack-scoped command before the standard is settled.** `state coverage`, `state get`, `state enable`, `state disable`, `state restore` and `catalog get` all take a `packId`, so a request that names none ("check my compliance posture", "am I compliant") does NOT identify a pack. Run `catalog list` FIRST, then branch on how many packs came back with `Available: true` — never on how many standards you expect to exist. **Two or more available:** present them, ask which, and END YOUR REPLY. **Exactly one available:** that one pack is the only posture there is to report, so proceed with it — but name it explicitly as the standard you are reporting on and say it was the only one available, so the user can correct you. **None available:** say so and stop. Never pick one because it is first in the list, most common, or the only one you have seen before: posture reported against a standard the user did not ask about is wrong even when every command succeeds. Resolution rules in [`references/compliance-pack/catalog/impl.md` § Pack ID lookup](./references/compliance-pack/catalog/impl.md#pack-id-lookup).
```
rockymadden: **Major — this contradicts the tri-state rule in the file you link to.**

`catalog/impl.md:20` is explicit about why `Available: true` is the wrong thing to count:

> **absent** -> Usable. Older services omit the field entirely. **Requiring `Available == true` would reject every pack and break the whole flow.** Only an explicit `false` blocks.

Branching on "how many packs came back with `Available: true`" means a service that omits the field counts **zero**, so the new **None available: say so and stop** branch hard-stops the entire compliance flow. That is the exact failure the tri-state table exists to prevent, and it is worse than the bug this PR fixes.

**Fix:** count *usable* packs, not `Available: true` packs:

> Run `catalog list` FIRST, then branch on how many packs are usable (present, and not explicitly `Available: false`) ...

skills/uipath-governance/references/compliance-pack/catalog/impl.md
```diff
@@ -103,5 +103,7 @@ If the array is empty: "No compliance standards are currently configured on this
...
 | A standard name with no match at all | Tell the user that standard is not available, list the `Available: true` packs, and offer to proceed with one. |
-| No standard named at all (e.g. "check my compliance posture") | Run `catalog list` and ask which standard. Never assume — more than one pack is available. |
+| No standard named at all (e.g. "check my compliance posture"), and **two or more** packs are `Available: true` | Run `catalog list` and ask which standard. Never assume. |
+| No standard named at all, and **exactly one** pack is `Available: true` | Proceed with that pack — it is the only posture that exists. Name it explicitly in the reply and say it was the only available standard, so the user can correct you. Do not ask a question with one answer. |
+| No standard named at all, and **no** pack is `Available: true` | Say no compliance standards are available on this tenant and stop. |
```
rockymadden: **Major — same `Available: true` counting bug as `SKILL.md:57`, across all three new rows.**

Line 20 of this same file says an **absent** `Available` field is usable, and that requiring `Available == true` "would reject every pack and break the whole flow." These three rows require exactly that, so on an older service every pack falls into this last row and the flow stops dead.

**Fix:** one word in three places:

```
| No standard named at all (e.g. "check my compliance posture"), and **two or more** packs are usable | ... |
| No standard named at all, and **exactly one** pack is usable | ... |
| No standard named at all, and **no** pack is usable | ... |
```

Line 105 above has the same leak (`list the Available: true packs`), though that one is pre-existing.

tests/tasks/uipath-governance/compliance-pack/unnamed_standard_smoke.yaml
```diff
@@ -29,55 +35,66 @@ initial_prompt: |
...
   - type: command_not_executed
     description: >
-      Agent did NOT run a pack-scoped coverage/get against an assumed standard -
-      the request names none, so no pack is the right pack yet.
-    command_pattern: 'uip\s+gov\s+compliance-packs\s+state\s+(?:coverage|get)\s+(?:tenant|organization)\s+\S+\s+iso-'
-    weight: 3.0
+      Agent did NOT run a pack-scoped command against a standard the catalog
+      marks unavailable - those have no bundle behind them and always fail.
+      Guard, not a discriminator: it cannot fire while one pack is available.
+    command_pattern: 'uip\s+gov\s+compliance-packs\s+(?:state\s+\S+|catalog\s+get)(?=(?:(?!\n|&&|\|\||;|\||\s(?:uip|\$UIP)\s).)*(?:iso-27001|gdpr-|hipaa|soc-2|eu-ai-act))'
```
rockymadden: **Medium — not actually segment-scoped, and it bans a correct run.**

The grader matches the raw command **and** a shlex-normalized form. `_normalize_shell` (`coder_eval/criteria/command_executed.py:53`) **collapses newlines to spaces**, so on that second haystack the `\n` guard is dead and only `&&`, `||`, `;`, `|`, and ` uip ` survive as boundaries. Replayed through the real normalizer, this correct run gets banned:

```
uip gov compliance-packs catalog get iso-42001-2023 >> output.txt 2>&1
grep -c hipaa output.txt
```

**Fix:** the CLI arg order is positional, so drop the lookahead entirely. Shorter, and immune:

```yaml
command_pattern: 'uip\s+gov\s+compliance-packs\s+(?:state\s+\S+\s+(?:tenant|organization)\s+\S+|catalog\s+get)\s+(?:iso-27001|gdpr-|hipaa|soc-2|eu-ai-act)'
```

That also closes the `PACK=iso-27001; ... $PACK` gap you list under Known gaps, at least for the direct-argument form.

---

**Minor, same line:** `iso-27001`, `gdpr-`, `hipaa`, `soc-2`, `eu-ai-act` is a literal snapshot of today's unavailable set, in a PR whose thesis is that hardcoded tenant state goes stale. When ISO 27001 ships, this starts banning a legitimate command. The description concedes "it cannot fire while one pack is available," so 1.0 of 11.0 weight is inert now and only goes live in the state where it is wrong. Also `gdpr-` assumes a hyphen-suffixed id, so a bare `gdpr` slips through. Either delete it, or put the removal trigger in the comment.

tests/tasks/uipath-governance/compliance-pack/unnamed_standard_smoke.yaml
```diff
@@ -29,55 +35,66 @@ initial_prompt: |
...
-  - uip CLI is not connected to a live tenant - auth errors expected.
   - Run each command once, capture output, and move on. Do NOT retry.
 
 success_criteria:
   - type: command_executed
-    description: "Agent listed the available standards instead of assuming one"
+    description: "Agent read the catalog instead of assuming a standard"
     command_pattern: 'uip\s+gov\s+compliance-packs\s+catalog\s+list'
     min_count: 1
+    require_success: true
```
rockymadden, line 53: **Medium — `require_success` reads the tool call's exit code, not this command's.**

`command_executed.py:173` filters on `cmd.result_status`, and there is one record per Bash call. The prompt tells the agent to run several commands appending to `output.txt`, and codex batches. If `catalog list` shares a call with any command that exits non-zero, this **3.0-weight** criterion zeroes even though the catalog read succeeded.

Already documented as a trap at `tests/tasks/uipath-review/_shared/check_review_cli_provenance.py:13`.

Not firing today (3/3 green), but it is a latent flake on the highest-weight deterministic check.

**Fix:** split it. Weight 2.0 without `require_success` (typed it), weight 1.0 with (it worked). A batching agent then loses 1.0, not 3.0.

tests/tasks/uipath-governance/compliance-pack/unnamed_standard_smoke.yaml
```diff
@@ -29,55 +35,66 @@ initial_prompt: |
...
+        justifies the pick - either by asking the user to choose among the
+        available standards, or by stating the named one was the only available
+        standard (or that none were available).
+      Score 0.5 if it names the standard but gives no reason why that one, so
+        the user cannot tell whether the choice was theirs to make.
+      Score 0.0 if it reports posture without making clear which standard the
+        result belongs to, or picks one silently with no acknowledgement that
+        the request specified none.
     weight: 3.0
     pass_threshold: 0.5
```
rockymadden, line 100: **Medium — this threshold does not enforce the behavior the PR is adding.**

The new rule is "name it **AND** say it was the only one available." But the 0.5 band is "names the standard but gives no reason why," and `pass_threshold: 0.5` passes it. The justification half, which is the whole point of the change, is unenforced.

Pre-existing, but you are rewriting this rubric, so now is the moment. Either drop that band to 0.25 or raise the threshold to 1.0.

tests/tasks/uipath-governance/compliance-pack/unnamed_standard_smoke.yaml
```diff
@@ -29,55 +35,66 @@ initial_prompt: |
...
     prompt: |
       The user asked: "Check my compliance posture and tell me where I stand."
 
-      The request names no compliance standard. More than one standard is
-      available on the platform (e.g. ISO 42001 and ISO 27001), and posture is
-      always measured against ONE specific standard.
+      The request names no compliance standard. Posture is always measured
+      against ONE specific standard, and the platform catalog marks each
+      standard `Available: true` or `Available: false` (announced but not yet
+      shipped). Only `Available: true` standards can be reported on.
```
rockymadden: **Minor — drops the absent case, same as `SKILL.md:57`.**

"marks each standard `Available: true` or `Available: false`" presents the field as binary. `catalog/impl.md:20` treats it as three-state, where absent means usable. Worth saying "usable (present, and not explicitly `Available: false`)" so the judge does not learn the binary the skill is being fixed to avoid.

tests/tasks/uipath-governance/compliance-pack/unnamed_standard_smoke.yaml
```diff
@@ -1,16 +1,22 @@
...
-  Guards SKILL.md Workflow step 1 + compliance-pack/catalog/impl.md Pack ID lookup.
-  Sibling check_routing_smoke.yaml covers the same posture flow WITH the standard
-  named; this task covers the ambiguous case that a second pack introduced.
-
-  Platform note: uip commands fail with auth errors in this environment - expected.
+  posture") must be resolved against the catalog before any pack-scoped command
+  runs, and the reply must name the standard it reports on. The agent must never
+  default to a pack it did not read out of `catalog list`.
+
+  This task is deliberately agnostic to how many packs are live on the test
```
rockymadden: **Minor — narration, not behavior.**

"Availability is tenant state we do not control... a test that demanded a 'which standard?' question would be demanding a question with one answer" is reasoning *about* the change rather than what the task does. Repo rule: if a line does not change what the agent does, it probably should not land. Two sentences carry this.

rockymadden (changes requested):
## Review

The core diagnosis is right, and the PR under-sells it. The old ban pattern was `state (coverage|get) (tenant|organization) \S+ iso-`, which matches `iso-42001-2023`, the **only available pack**. So the old criterion actively penalized the correct behavior. That is the real cause of the recorded 0.654, and it is buried under "demoted to weight 1.0."

### Verified independently

- All three linked runs (`claude`, `codex`, `antigravity`) hit **5/5, weighted 1.000, SUCCESS** against a live tenant.
- Scoring math checks out: 8/11 = 0.727, 9.5/11 = 0.864.
- `skills:validate`, `skills:check-links`, `check-cli-verbs.py` clean on the branch. YAML parses, total weight 11.0, tags conform.
- No `studioweb` flavor override for `uipath-governance`, no flavor markers in the edited passages. Nothing to mirror.
- The single-pack "proceed" branch does **not** open a mutation path: `full-apply/impl.md:15` still gates `state enable` behind `Proceed? (y/n)`.

### Blocking (2)

Both are small, both inline:

1. **Major** `SKILL.md:57` + `catalog/impl.md:106-108` count packs with `Available: true`, which contradicts the tri-state rule 80 lines up in the same file. On a service omitting the field, the new "None available" branch hard-stops the whole compliance flow.
2. **Medium** The new `command_not_executed` pattern is not segment-scoped on the haystack that decides the match, and bans a correct run. Verified against the real normalizer.

### Non-blocking

3. **Medium** `require_success` batching exposure on the 3.0-weight check (inline).
4. **Medium** Judge `pass_threshold: 0.5` does not enforce the behavior being added (inline).
5. **Minor** PR body contradicts itself, and one number does not reconcile:
   - The header links three passing runs. The **Verification status** section then says *"Not run against a live tenant... Please run before merge"* and hands the reviewer a command. Already done, 1.000 on all three. Delete the section, it sends the next reviewer on a dead errand.
   - `claude-sonnet-5` "before" is **0.654** in the evidence table and **0.773** in the projection table. Same run, same label. The whole argument is numeric, so reconcile or drop one.
6. **Minor** The new ban hardcodes today's unavailable pack ids in a PR whose thesis is that hardcoded tenant state goes stale (inline).
7. **Minor** Scope is 1 of 7. Six siblings carry the identical false premise, all under the same `CODEOWNERS:170` line: `restore_smoke`, `already_configured_smoke`, `apply_commands_smoke`, `check_routing_smoke`, `state_list_smoke`, `disable_smoke`. The body says "flagged for @UiPath/authz" and "11 more likely"; the actual same-folder count is 6, and a note in a merged PR body is invisible. Open a tracking issue.
8. **Minor** Task `description` narration (inline).

### Recommendation

**Request changes**, gated on 1 and 2 only.

1 is the one that matters: the PR trades a broken test for skill guidance that hard-stops the compliance flow on any service omitting `Available`, which `impl.md` documents as a live shape. That is worse than the bug being fixed, and it is a one-word fix in four places. 2 is a verified false positive on a correct run, and the replacement pattern is shorter than the one it replaces.

3 through 8 are fine as follow-ups. Good work on the underlying diagnosis, and the three green runs back it.

PR #3132, opened 2026-09-08
feat: route DTL calls via script execution for v4 (#ENGCE-61216)

mrstark14:
## Why

Reference fields on `4.0.0` (v4) connector activities are **script-backed**, not object-backed. The v4 metadata carries `reference.scriptRef` and has no `objectName` / `path`:

```json
"reference": { "scriptRef": "list_usergroups", "lookupValue": "id", "lookupNames": ["id", "name"] }
```
`uip is resources describe uipath-salesforce-slack add_users_to_usergroup --activity-version 4.0.0 --operation POST`

The skills only knew one lookup path — `uip is resources run list <connector> <reference.objectName>` — which cannot resolve these: there is no object to list. The CLI now exposes the design-time lookup (DTL) scripts through the IPE Runtime Proxy (`elements_/v4/execute`, `feat: Add support to execute IPE runtime proxy (#ENGCE-61027)`):

```
uip is resources run script --connection-id <id> --connector-key uipath-salesforce-slack --script-ref list_usergroups
```
`integrationservice-tool/src/commands/resources.ts` § `run script`

Two things an agent cannot infer and would otherwise get wrong:

- **`maestro flow registry get` strips `scriptRef`.** `applyOptionalFieldProps` copies only `objectName`/`lookupValue`/`lookupNames`/`path`/`childPath` (`maestro-sdk/src/registry/integration-service-fetcher.ts`), so a v4 field surfaces as `reference: { lookupValue, lookupNames }` with no target. The docs now say: neither `objectName` nor `scriptRef` on a `4.0.0` node means script-backed — read `scriptRef` from `describe`, never `run list` a guessed object.
- **The response envelope differs from `run list`.** The proxy relays the vendor answer verbatim (ADR-0014): `Data.Body` is a JSON **string** decoding to `{status, headers, body}`, a vendor 4xx/5xx still returns `Result: "Success"` (read `Data.Status`), `--output-filter` cannot reach inside the string, and there is no `Data.Pagination`.

## What

| File | Change |
|---|---|
| `uipath-platform/.../reference-resolution.md` | New section **4.0.0 Activities — Script References (`scriptRef`)**: reference shape, 4-step resolution workflow (`describe --activity-version 4.0.0 --operation <METHOD>` → `run script --script-ref` → match `lookupNames` → write `lookupValue`), response parsing with a `jq` example, dotted `lookupNames` (`profile.email`), and where `scriptRef` is/isn't visible. *Connection-Scoped* and *When the Lookup Call Fails* now name `run script` alongside `run list` so the no-substitution rules apply to both. |
| `uipath-platform/.../resources.md` | § `--activity-version`: v4 describe example gains `--operation <METHOD>` (without it the response is an operation summary with no `requestFields`), plus the `scriptRef` → `run script` pointer. New § **`run script`** under *Execute Operations*: syntax, `--script-ref` xor `--inline-script`, `--connector-key` required with `--script-ref`, envelope caveats. |
| `uipath-maestro-flow/.../connector/impl.md` | § 4.0.0 Activities: delta 5 — reference fields are script-backed, resolve with `run script`, `registry get` strips `scriptRef`. Step 4: callout routing `4.0.0` nodes to `run script` with parsing rules; zero-match / multi-match / failed-call / connection-scoping rules unchanged. |

No frontmatter, Critical Rules, or flavor-marked passages touched; no `skill-flavors/` override exists for these files.

## Verified

- `uip is resources run script … --script-ref list_usergroups` against a live Slack connection: `Outcome: vendor`, `Status: 200`, `Data.Body` → `{status, headers, body}` with 821 user-group rows carrying `id`/`name` — matches the documented parsing.
- `describe … --activity-version 4.0.0` without `--operation` returns `AvailableOperations: null` and no fields; with `--operation POST` it returns `Reference.ScriptRef` — hence the added flag.
- `node scripts/check-skill-links.mjs` — all 6709 links resolve (new anchors included).
- `scripts/check-skill-verbs.py` flags `is resources run script` as stale only because `assets/uip-catalog-snapshot.json` was built from a dev CLI that predates the command; rebuilding the snapshot needs the `rpa` tool installed for the 1.203.x line (the build refuses otherwise).


skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -39,6 +39,7 @@ An activity is `4.0.0` when its `configuration` JSON reports `"version":"4.0.0"`
 2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`).
 3. **Operation label ≠ HTTP verb** — a semantic operation (e.g. `Update`) pairs with any verb (e.g. `POST /usergroups.users.update`). `flow validate` accepts it; do not "fix" the method to match the label.
 4. **Not connection-scoped** — `--connection-id` on `registry get` adds no custom fields.
+5. **Reference fields are script-backed** — `reference` carries `scriptRef` (e.g. `list_usergroups`) and no `objectName`. Resolve with `uip is resources run script --connection-id <id> --connector-key <connector-key> --script-ref <scriptRef> --output json`, never `run list`. `registry get` strips `scriptRef` — read it from the Step 3 `describe` (`--activity-version 4.0.0 --operation <method>`). Response parsing and matching rules: [/uipath:uipath-platform — reference-resolution.md § 4.0.0 Activities — Script References](../../../../../uipath-platform/references/integration-service/reference-resolution.md#400-activities--script-references-scriptref).
```
mukundbayyaram, line 42: Can connectorKey be derived at SDK layer by looking at connection's cached response _(assuming we do cache connection response)_ ? so that command looks a little cleaner ?

skills/uipath-platform/references/integration-service/reference-resolution.md
```diff
@@ -131,6 +132,72 @@ User says: "Send a message to #general"
...
+  --activity-version 4.0.0 --operation <METHOD> --output json
+
+# 2. For each reference field, run its script against the bound connection
+uip is resources run script --connection-id "<id>" \
+  --connector-key "<connector-key>" --script-ref "<reference.scriptRef>" --output json
+
+# 3. Match the user's input against reference.lookupNames in the returned rows
+#    Extract reference.lookupValue as the resolved ID
+
+# 4. Write the resolved ID (never the display name) into the node / --body
```
mukundbayyaram, line 177: Should this be part of `bash` ?

skills/uipath-platform/references/integration-service/reference-resolution.md
```diff
@@ -131,6 +132,72 @@ User says: "Send a message to #general"
...
+
+`--connection-id` and `--connector-key` are both required with `--script-ref`. The metadata itself is not connection-scoped, but the **lookup is**: the script runs with the connection's credential, so the same connection-scoping rules apply ([Reference IDs Are Connection-Scoped](#reference-ids-are-connection-scoped-critical)).
+
+### Reading the response
+
+The proxy relays the vendor's answer verbatim, so the envelope differs from `run list`:
+
+- `Data.Outcome` — `"vendor"` when the call reached the vendor. `Data.Status` is then the **vendor's** HTTP status; a vendor `4xx`/`5xx` still returns `Result: "Success"`. Read `Data.Status`, never assume success from the envelope.
+- `Data.Body` — a **JSON string**, not an object. Parse it: it decodes to `{ "status", "headers", "body" }`, and `body` holds the rows the script returned (for `list_*` scripts, an array of records).
+- `--output-filter` operates on the envelope only; it cannot reach inside the string. Parse with `jq`/Python instead:
```
mukundbayyaram: This needs to be updated based on feedback in CLI PR ?

mukundbayyaram (commented):
We need a skill / reference on how our new scripts look like and conventions so that the scripts agent is building is correct ?

mukundbayyaram (approved):

baishalighosh (approved):

PR #3143, opened 2026-09-08
chore(tests): bump coder-eval pin to 0.11.7

bai-uipath:
Moves `tests/.coder-eval-version` from 0.11.5 to 0.11.7, the version the eval framework released today. Every gate in this repo (smoke-skills, activation-gate, smoke-rpa-skills, the ad-hoc run-coder-eval dispatch and the nightly ADO pipeline) reads that pin, so this is the single edit that moves them all onto the new framework together.

What 0.11.7 brings that matters here: the evalboard renders multi-variant runs as one row per (task, arm) and loads noticeably faster, the pricing rate card is refreshed with the Gemini Flash entries our antigravity runs were costing as unpriced, and the bundled antigravity harness moves to 0.1.8 to clear a Defender false positive. The published composite action's input surface is unchanged from what `run-coder-eval.yml` already consumes.

Also removes a stale caveat in the tests README that told readers `Orchestrator.simulator_route` was unreleased and the pin should move once it shipped. It shipped in 0.11.5, which the pin has been on since the checker moved to the litellm route.

Carries one companion fix to the tests image. The antigravity harness in 0.1.8 asserts a protobuf runtime of 7.35 or newer, while the image also installs `uipath-langchain` to pre-warm a coded-agent fixture and that pulls in a dependency capping protobuf below 7. The resulting downgrade left antigravity unable to load its own generated stubs, so its harness smoke failed on every run. The image now installs a protobuf that satisfies both sides: the capping package's own stubs are generated against a much older protoc, so its declared ceiling is conservative metadata rather than a real limit. Only the docker path was ever affected, since the host venv does not install `uipath-langchain`.

## Dry run

Validated on the nightly infra: [UiPath.Coder_Eval.Nightly 20260908.15](https://dev.azure.com/uipath/coder_eval/_build/results?buildId=13310375) (`dryRun=true`, `skillsRef` = this branch, `coderEvalVersion=pinned`, `harness=codex`, `agentModel=gpt-5.6-luna`, `suite=skills,activation`), which reports `GATE: PASS`. coder-eval resolved from this branch's pin to 0.11.7 and pulled the matching `coder-eval-agent:0.11.7` image, the skills subset came back at 92% (23/25) with zero errors alongside 26 activation cases, and the antigravity docker harness smoke is green at 1/1. Both task failures are task-level rather than framework-level: one fails identically in the preceding dry run on this same pin, and the other is a single-replicate agent flake in a scaffolding task.

All four stages succeeded with no failed or partially succeeded step anywhere in the run, so the whole split path (both eval slices, the merge, the publish and the teardown) is validated on 0.11.7 in this one run. The preceding run, [20260908.13](https://dev.azure.com/uipath/coder_eval/_build/results?buildId=13309815), is the same pin without the image fix and failed the gate on the antigravity smoke, which is what isolates the fix to that change. This repo's own smoke gate is green on 0.11.7 as well.


uipreliga (approved):
🚢

PR #3145, opened 2026-09-08
ci(tests): gate Windows-reachable shell commands for cmd.exe safety

bai-uipath:
## What

Adds one rule to `scripts/check-task-driver.py`: a Windows-reachable `pre_run`/`post_run` command must be a single line. Scope is every experiment hook, including per-variant overrides under `variants[]`, plus the hooks of `windows`-tagged tasks. Also fixes a live bug found on the way, and puts the existing driver-pin gate under test for the first time.

## Why

On Windows, coder_eval dispatches hooks through `asyncio.create_subprocess_shell`, which is `%COMSPEC% /c`, so cmd.exe. cmd parses only the *first* line of the string it is handed, so a YAML `|-` block scalar silently drops the rest. `PreRunCommand.fail_on_error` defaults to true, so a `pre_run` cmd cannot parse lands every task as `FinalStatus.ERROR` before the agent starts.

That is the 2026-09-04 outage: #2756 added a multi-line POSIX sh `pre_run` hook to `nightly.yaml` and the Windows nightly split came back `7/7 status=ERROR score=0.000 iterations=0`. #3116 fixed the hook. Nothing caught it, and the newline rule is the cheapest thing that would have.

Every hook command in the corpus already satisfies the rule, and a newline is a syntactic fact rather than a heuristic, so this needs no per-experiment configuration and has no false-positive surface.

## What it deliberately does not do

It does not detect POSIX sh that fits on one line. That is recorded as a test (`test_known_gap_single_line_posix_passes`) rather than left implicit.

Catching it needs a construct table, and the construct table then needs quote-awareness, since every Windows-tagged `run_command` criterion in the corpus is a `python -c "..."` payload that cmd handles fine. It also needs a list of which experiments can reach Windows, so the seven Linux-only ones do not fail, plus an exemption for the `post_run` cleanup that `coder_eval_uipath`'s overlay translates to PowerShell. That version was built and cut: too much machinery, and three tables to keep current, for a failure mode that has not happened. If a one-line POSIX hook ever does break a Windows run, that is the moment to add it.

Scanning task hooks wholesale is out. The corpus holds 24 multi-line task hooks and not one belongs to a `windows`-tagged task, so flagging them would mean one-lining 20-to-35-line heredocs against a shell they never meet. Only `windows`-tagged tasks are scanned, which today covers seven tasks carrying no hooks at all: the rule is there for the next hook added to one, not for anything currently in the tree.

Windows-tagged tasks' `run_command` criteria stay out of scope as well. They are the latent silent-zero class rather than the outage, and there are currently zero offenders among them.

## Live bug fixed

`smoke-windows.yaml`'s `post_run` was `rm -rf node_modules .npm-prefix`, running on `uipath-windows-latest` and not covered by the Windows overlay. Since post-run failures are non-fatal it warn-logged and leaked `node_modules` on every task instead of failing visibly. Now a PowerShell one-liner, guarded per path with `Test-Path` and `-ErrorAction Stop` so it exits zero when there is nothing to clean and non-zero only on a real delete failure. The unguarded form exits 1 whenever either path is already absent, which is the common case, so it would have traded a warning on every task for a different warning on every task. It is spelled out per path rather than looping over a variable, so the command carries no `$` for a shell to expand. This bug is not caught by the newline rule; it turned up while looking, and stands on its own.

## Note

The job name in `task-driver-gate.yml` is a required status check in the `main` ruleset, so both gates run in one step under the existing name.


uipreliga (commented):
# Review: @uipath/skills (UiPath/skills) — pr:3145 (6 files) axis:1,2,3,4,5,7,8

PR #3145 — ci(tests): gate Windows-reachable shell commands for cmd.exe safety — author bai-uipath — bai/windows-shell-portability-gate → main — OPEN

**Change class:** Infrastructure/CI hardening — adds a second static gate (Windows shell portability of experiment/task hook commands) to an existing required-check script, registers a new unit-test job as a required check, and fixes a real Windows post_run leak in smoke-windows.yaml

A strong, well-tested change (9.5/10, no critical or high findings, clean types/docs/correctness) whose real risk is concentrated in the new hook-portability gate's resilience: it fails open — a missing, renamed, or unparseable scan root scans zero files and still prints OK and exits 0 on a required check — and it introduces the script's first document-controlled `::error` sink plus a Windows cleanup that can abort halfway, so this is mergeable once the fail-open paths and the unescaped annotation are fixed and the remaining items are filed as follow-ups.

## Summary

| 1. Correctness & Logic | 9.9 / 10 | 0 | 0 | 0 | 1 | `_iter_hook_commands` is a hand-rolled depth-2 walk; anything nested deeper is skipped and still counted as green |
| 2. Code Quality & Simplicity | 9.1 / 10 | 0 | 0 | 1 | 4 | _hook_line_number's unanchored substring re-scan and <8-char bail-out emit invalid line=0 (or wrong-line) ::error annotations, and no test asserts the annotation line |
| 3. Types & Contracts | 9.9 / 10 | 0 | 0 | 0 | 1 | `scopes: list[tuple[str, dict]]` is violated at runtime — YAML mapping keys are not necessarily strings |
| 4. Test & Validation Health | 9.7 / 10 | 0 | 0 | 0 | 3 | FIXED_HOOK is not pinned to nightly.yaml, unlike BROKEN_HOOK which is pinned to git history |
| 5. Security | 9.4 / 10 | 0 | 0 | 1 | 1 | GitHub Actions workflow-command injection: YAML-derived key names interpolated unescaped into `::error` annotations |
| 7. Error Handling & Resilience | 8.8 / 10 | 0 | 0 | 2 | 2 | Gate 2 fails open: a missing or renamed scan root is silently treated as an empty corpus and the gate prints OK / exits 0 |
| 8. Interface, Docs & Compatibility | 10 / 10 | 0 | 0 | 0 | 0 | — |

**Overall Score**: 9.5 / 10 · **Weakest Axis**: Error Handling & Resilience at 8.8 / 10
**Totals**: 🔴 0 · 🟠 0 · 🟡 4 · 🔵 12 across 7 axes reviewed.

None.

1. **[Axis 2] _hook_line_number's unanchored substring re-scan and <8-char bail-out emit invalid line=0 (or wrong-line) ::error annotations, and no test asserts the annotation line** (`scripts/check-task-driver.py:115`) — Drop the helper. `yaml.compose()` (or a `yaml.SafeLoader` subclass) gives the mapping node's true `start_mark.line + 1` for the offending `command` key, needs no threshold and no substring guessing, and reuses the single parse `main()` already does at line 134. If that is more than the annotation is worth, the cheaper honest option is to drop the line number entirely and annotate the file — the `defaults.pre_run[0].command` scope label already printed alongside is what actually locates the hook. Note that `_driver_line_number` (lines 86-91) does not need the same treatment: it matches the anchored regex `_DRIVER_LINE = re.compile(r"^\s*driver:\s*tempdir\s*$")` and is correct as written. The only thing the two helpers share is scaffolding — a second read of the file, the 1-indexed loop, and the `0` sentinel — so a node-based lookup could replace both reads, but the substring/threshold bug is confined to `_hook_line_number`.
2. **[Axis 5] GitHub Actions workflow-command injection: YAML-derived key names interpolated unescaped into `::error` annotations** (`scripts/check-task-driver.py:179`) — Gate 2's annotation is the first place this script interpolates *document-controlled* content into a GitHub Actions workflow-command sink. Line 179:

    print(f"::error file={rel},line={line}::{yaml_path} spans multiple lines")

`yaml_path` is built from raw YAML mapping keys in `_iter_hook_commands` (line 109 `prefix = f"{scope_name}.{hook}" if scope_name else hook`, line 112 `yield f"{prefix}[{index}].command", step["command"]`), where `scope_name` is a top-level key of the parsed document. YAML keys may contain newlines and `::`, so a contributor's PR can terminate the annotation and emit further workflow commands. Demonstrated at PR HEAD in /tmp/pr-3145-worktree:

    $ python3 - <<'EOF'
    import yaml
    yaml.safe_dump({"experiment_id":"x",
      "defaults\n::error title=Injected::spoofed annotation\n::notice":
        {"pre_run":[{"command":"echo one\necho two"}]}},
      open("/tmp/inj/e.yaml","w"), sort_keys=False)
    EOF
    $ python3 scripts/check-task-driver.py /tmp/inj/e.yaml
    FAIL — 1 Windows-reachable hook command(s) span multiple lines:

    ::error file=/tmp/inj/e.yaml,line=8::defaults
    ::error title=Injected::spoofed annotation
    ::notice.pre_run[0].command spans multiple lines
    rc=1

The attacker-supplied `::error title=Injected::` and `::notice` lines are emitted as independent workflow commands. Compare `git show main:scripts/check-task-driver.py` line 116, where the annotation body was a fixed literal (`::error file={rel},line={line}::Task pins sandbox.driver: tempdir`) and only a repo path and an int were interpolated — so this exposure is introduced by this PR, not inherited.

Impact is bounded, which is why this is Medium and not High: .github/workflows/task-driver-gate.yml triggers on `pull_request` (line 27), not `pull_request_target`, declares `permissions: contents: read` (lines 37-38), and the gate step (line 53) captures no step output — so a fork PR's token is read-only and no secret is in scope. `::set-env`/`::add-path` are disabled by the runner. The reachable abuse is forging/suppressing CI log annotations (`::add-mask::` to hide later output, `::stop-commands::<token>` to disable parsing for the rest of the job) on a run the attacker has already made fail. It cannot make the gate pass — `rc` is still 1.

Fix: sanitize before interpolation. Add a helper and use it for both annotation sites (line 179 and line 201):

    def _annot(s: str) -> str:
        return s.replace("%", "%25").replace("\r", "%0D").replace("\n", "%0A").replace(":", "%3A").replace(",", "%2C")

GitHub's own toolkit applies exactly this percent-escaping to annotation properties and message bodies. Apply it to `yaml_path` and to `rel` (a repo path can also contain `::` or, on Linux, a newline). The plain-text echo on line 180 is not a sink and can stay as-is. CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:L/A:N
3. **[Axis 7] Gate 2 fails open: a missing or renamed scan root is silently treated as an empty corpus and the gate prints OK / exits 0** (`scripts/check-task-driver.py:73`) — `_iter_task_yamls` never checks that a root exists — `root.rglob()` on a non-existent path yields nothing rather than raising, so a typo'd or moved root scans zero files and both gates report a pass. Demonstrated at PR HEAD in /tmp/pr-3145-worktree:

    $ python3 scripts/check-task-driver.py tests/tasks tests/experiment-typo
    OK — no task pins `sandbox.driver: tempdir`.
    rc=0                      # note: no "Windows-reachable hook command(s)" line at all
    $ python3 scripts/check-task-driver.py /nope/nothing/here
    OK — no task pins `sandbox.driver: tempdir`.
    rc=0                      # zero files scanned, green

This matters more for Gate 2 than for Gate 1 because Gate 2 has no default root. Line 67 is `roots = [Path(a) for a in args] if args else [DEFAULT_ROOT]`, and `DEFAULT_ROOT` (line 61) is `REPO_ROOT / "tests" / "tasks"` only — so Gate 2 runs *only* when the caller happens to pass `tests/experiments` on argv, and the docstring's own first usage line (line 40, `python3 scripts/check-task-driver.py   # tests/tasks`) is a documented invocation that silently exercises none of it. The reporting reinforces this: line 189 `elif hooks_checked:` means zero hooks prints nothing, so a Gate-2 no-op and a Gate-2 pass produce byte-identical output.

Honest mitigation, so this is not overstated: `test_live_corpus_passes_both_gates` in tests/scripts/test_check_task_driver.py asserts `"Windows-reachable hook command(s) are single-line"` appears in the output, which does catch `tests/experiments` being moved or emptied. But that test hardcodes its own two roots — it does not read .github/workflows/task-driver-gate.yml, so an edit or revert of the argv list on line 53 of that workflow silently disables Gate 2 in CI with every check still green.

Fix (either or both): (a) make the default `DEFAULT_ROOTS = (REPO_ROOT/"tests"/"tasks", REPO_ROOT/"tests"/"experiments")` so the gate is correct regardless of argv; (b) in `_iter_task_yamls`, `if not root.exists(): sys.exit(f"no such scan root: {root}")` — a required check should crash on a bad root, never green-light it. Raise to High if CI ever loses the `tests/experiments` argument, since nothing then reports the loss.
4. **[Axis 7] Windows post_run cleanup stops after the first failure, so `.npm-prefix` is left behind when `node_modules` cannot be deleted (residual gap, not a regression)** (`tests/experiments/smoke-windows.yaml:49`) — Make the two cleanups independent so a `node_modules` delete failure cannot skip the `.npm-prefix` delete, while preserving the non-zero exit on a real failure (which `-ErrorAction Stop` is currently providing — verified: simply removing it makes the failure exit 0 and silent).

Verified working single-command form for line 49:

```yaml
  post_run:
    - command: 'pwsh -NoLogo -NoProfile -NonInteractive -Command "if (Test-Path -LiteralPath ''node_modules'') { Remove-Item -LiteralPath ''node_modules'' -Recurse -Force -ErrorAction Continue }; if (Test-Path -LiteralPath ''.npm-prefix'') { Remove-Item -LiteralPath ''.npm-prefix'' -Recurse -Force -ErrorAction Continue }; if ($Error.Count) { exit 1 }"'
      timeout: 30
```

Measured on pwsh 7.5.0 with an undeletable `node_modules`: `.npm-prefix` is removed and the command still exits 1. Note the literal `$Error` — confirm the harness does not expand `$` in `post_run.command` before adopting this form; the existing comment on line 47 says the command was deliberately written with "no `$`, so no shell eats it". If `$` is unsafe there, use either of these instead:

- Two separate `post_run` entries, one per path, each keeping `-ErrorAction Stop` (simplest, and it also sidesteps the `$` question entirely).
- One command wrapping each delete in its own `try { ... } catch { $failed = $true }` with `exit ([int]$failed)` at the end (still uses `$`).

Update the comment on lines 45-47 in the same change: it currently justifies only the exit-code semantics and does not mention that chained `-ErrorAction Stop` aborts the remaining cleanup — state that the two deletes are independent by design.

## Nits

1. **[Axis 1] `_iter_hook_commands` is a hand-rolled depth-2 walk; anything nested deeper is skipped and still counted as green** (`scripts/check-task-driver.py:96`) — Lines 96-103 build the scope list by hand:
```python
    scopes: list[tuple[str, dict]] = [("", doc)]
    for key, value in doc.items():
        if isinstance(value, dict):
            scopes.append((key, value))
        elif isinstance(value, list):  # variants[] holds per-variant overrides
            scopes.extend(
                (f"{key}[{i}]", v) for i, v in enumerate(value) if isinstance(v, dict)
            )
```
That reaches exactly three shapes: top-level, one dict level (`defaults`), and dicts inside a top-level list (`variants[]`). A hook one level deeper — e.g. `variants[0].defaults.pre_run` — is silently skipped, and the run still prints the reassuring `OK — N Windows-reachable hook command(s) are single-line.`

I confirmed the walk is complete for today's corpus and that `11` is the right number, so this is not a live defect: probing every file under `tests/experiments/` shows all 11 hooks live at `defaults.pre_run` / `defaults.post_run`; every one of the 724 task hooks (515 `pre_run` + 209 `post_run`) sits at the top level; the 7 `windows`-tagged tasks under `tests/tasks/` have no hooks at all; no `variants[]` entry carries a nested `defaults`; and `grep -rln '^experiment_id:'` finds experiment YAMLs only under `tests/experiments/`. What I could not establish is whether coder_eval's schema forbids the deeper shapes — the package is not installed here and `tests/README.md` does not specify the variant schema.

For a gate whose entire value is completeness, a recursive walk is both simpler and strictly more complete than three special cases. Fix: replace the manual scope list with a recursive descent over dicts and lists that yields any `pre_run`/`post_run` list it encounters, building the dotted path as it goes; the existing `test_variant_level_hook_is_caught` keeps passing and the class of silent miss disappears.
2. **[Axis 2] Second gate bolted onto a module, workflow and required-check context still named and described for the first gate only** (`scripts/check-task-driver.py:2`) — The file is `scripts/check-task-driver.py`, the workflow is `.github/workflows/task-driver-gate.yml` with `name: Task Driver Gate`, and the required context is the job `name: No task pins sandbox.driver tempdir` (task-driver-gate.yml:43). The module docstring now opens `"""Two static gates over the coder-eval task and experiment corpus.` — the docstring is the only place that admits the module has grown a second, unrelated concern. A PR that trips gate 2 fails a required check literally named "No task pins sandbox.driver tempdir" while the actual defect is a multi-line `pre_run`.

CLAUDE.md rule 10 and docs/REQUIRED-CHECKS.md:138 ("Renaming a job here renames its check and **breaks the ruleset**") pin the *job* name, so leaving it alone is correct. Nothing pins the *script* filename or the *workflow* filename/`name:` — those are free to rename.

Fix: either (a) rename the script to something that covers both concerns (e.g. `scripts/check-task-corpus.py`) and update the two call sites (task-driver-gate.yml:53 and tests/scripts/test_check_task_driver.py:17), or (b) keep gate 2 in its own module invoked by a second step in the same job — the job name, and therefore the required context, is unaffected by step count. If neither is done, add a line to the workflow header comment saying explicitly that the required-check name under-describes what the job now enforces, so the next reader of a red check is not misdirected.
3. **[Axis 2] Speculative generality in the Windows-reachability predicate and the hook scope walk** (`scripts/check-task-driver.py:143`) — ```python
        tags = doc.get("tags")
        if (
            "experiment_id" in doc
            or "variants" in doc
            or (isinstance(tags, list) and "windows" in tags)
        ):
```
All 9 experiment YAMLs under `tests/experiments/` carry both `experiment_id` and `variants` (checked by loading each file and printing `sorted(d.keys())` at PR HEAD), so the `or "variants" in doc` disjunct never fires on its own; it is a second, unexercised way to say "this is an experiment".

Similarly `_iter_hook_commands` (lines 94-112) builds `scopes` from *every* top-level dict value and *every* dict inside *every* top-level list, then looks for `pre_run`/`post_run` in each. In the real corpus all 11 hooks live at `defaults.pre_run` / `defaults.post_run` (verified by walking every experiment file); the walk's extra reach is `agent`, `sandbox`, `success_criteria[i]`, … — places the schema can never put a hook, and where a spurious match would print a nonsense scope label like `success_criteria[2].pre_run[0].command`.

Fix: name the two scopes the schema actually has —
```python
scopes = [("", doc), ("defaults", doc.get("defaults"))]
scopes += [(f"variants[{i}]", v) for i, v in enumerate(doc.get("variants") or [])]
```
filtering non-dicts — and reduce the predicate to `"experiment_id" in doc or (isinstance(tags, list) and "windows" in tags)`. Same coverage, no unreachable branches. (Both changes stay green under the new suite: `test_variant_level_hook_is_caught` sets `experiment_id` as well.)
4. **[Axis 2] Workflow comment states a false causal link between step count and the required-check name** (`.github/workflows/task-driver-gate.yml:51`) — ```yaml
      # One step, so the required-check job name above stays byte-identical.
      - name: Check task drivers and experiment hook portability
        run: python3 scripts/check-task-driver.py tests/tasks tests/experiments
```
The required context is the job's `name:` (line 43, `No task pins sandbox.driver tempdir`) — docs/REQUIRED-CHECKS.md:138: "Renaming a job here renames its check and **breaks the ruleset**." A job without a matrix produces exactly one check run named by that field regardless of how many steps it has, so adding a second step would not have changed the reported context. As written the comment reads as a constraint ("we may only ever have one step here") that does not exist, and it is the stated justification for folding an unrelated gate into this script rather than a second step or module (see the naming finding).

Fix: say what is actually load-bearing, e.g. `# Both gates run in one invocation; the job name: above is the required context — do not rename it (docs/REQUIRED-CHECKS.md Rule 1).`
5. **[Axis 2] Git-history-coupled fixture test runs a subprocess at import time and self-disables in the environment it guards** (`tests/scripts/test_check_task_driver.py:103`) — ```python
    @pytest.mark.skipif(
        subprocess.run(
            ["git", "cat-file", "-e", "80c757e03^{commit}"],
            cwd=REPO_ROOT, capture_output=True, check=False,
        ).returncode != 0,
        reason="commit 80c757e03 not in this clone (shallow checkout)",
    )
    def test_broken_hook_fixture_matches_git_history(self):
```
The decorator's `subprocess.run` is evaluated at module import, so every collection of this file — including `pytest --collect-only` and any run that selects a different test — shells out to git. The assertion it guards compares a 3-line module constant to a blob, i.e. it tests a literal, not the gate's behaviour, and it silently degrades to a skip in exactly the case it is written for (a shallow clone); it also forces `fetch-depth: 0` on the new CI job (.github/workflows/test-helpers.yml:215), a full clone of a 301 MiB pack / 2250 commits, so the guard's only reliable execution path is also its most expensive one.

This is machinery out of proportion to "BROKEN_HOOK is the real string". Fix: delete the test and keep the provenance where it already is — the comment at line 19, `# As it stood on main between #2756 and #3116. Pinned to git history below.` — restated as `# Verbatim from 80c757e03:tests/experiments/nightly.yaml`. That also lets the new job drop `fetch-depth: 0` and match its sibling jobs, one of which documents the same choice at test-helpers.yml:169 ("No fetch-depth override: the suite only ever resolves HEAD..."). If the provenance check is genuinely wanted, move it out of the unit suite into a one-off script rather than paying for it on every PR.
6. **[Axis 3] `scopes: list[tuple[str, dict]]` is violated at runtime — YAML mapping keys are not necessarily strings** (`scripts/check-task-driver.py:96`) — Line 96 declares `scopes: list[tuple[str, dict]] = [("", doc)]`, and lines 97-103 populate it straight from `doc.items()` — `scopes.append((key, value))` and `scopes.extend((f"{key}[{i}]", v) ...)`. YAML mapping keys are not restricted to strings, so `key` can be an `int`, `bool`, `None`, or a date, and the declared element type is then wrong. Demonstrated against the PR-HEAD worktree with a document whose second top-level key is the integer `1`: the gate emitted `::error file=...,line=5::1.pre_run[0].command spans multiple lines` — a non-`str` key flowed into the scope-name slot. This is benign today (the f-string at line 108, `prefix = f"{scope_name}.{hook}" if scope_name else hook`, coerces it, so nothing crashes, and no such key exists in the corpus — I scanned 1293 dict-shaped task YAMLs plus all 9 experiment files), and the repo runs no typechecker, so nothing would flag it. Recording it because the routed question for this axis was whether the new annotations are honest: `_iter_hook_commands(doc: dict) -> Iterator[tuple[str, str]]` (line 94) IS honest — the first element is always an f-string and the second is guarded by `isinstance(step.get("command"), str)` at line 111 — and `doc: dict` matches every call site (main() guards with `isinstance(doc, dict)` before the call). `scopes` is the one annotation that does not hold. Fix: declare it `list[tuple[object, dict]]`, or normalize at insertion — `scopes.append((str(key), value))` and `(f"{key!s}[{i}]", v)` — which makes the declared type true and the printed scope name explicitly stringified rather than incidentally so.
7. **[Axis 4] FIXED_HOOK is not pinned to nightly.yaml, unlike BROKEN_HOOK which is pinned to git history** (`tests/scripts/test_check_task_driver.py:26`) — Line 26 asserts a fact in a comment that no test enforces:

```python
# What #3116 replaced it with, live in nightly.yaml today.
FIXED_HOOK = (
    ':; if [ "${MAESTRO_FLOW_SDK_SETUP:-0}" = "1" ]; then '
    'bash "$SKILLS_REPO_PATH/tests/scripts/stage-preview-sdk-workspace.sh"; fi'
)
```

The author deliberately pinned the *broken* fixture to git history (`test_broken_hook_fixture_matches_git_history`, lines 113-122) precisely so it cannot become "a retyped approximation" — but gave the fixed fixture no equivalent. I verified the claim holds today: parsing `tests/experiments/nightly.yaml` at PR HEAD, `defaults.pre_run[0].command` is byte-identical to `FIXED_HOOK`. Nothing keeps it that way; once nightly.yaml's `pre_run` changes, `test_the_shipped_fix_passes` (80-82) keeps passing against a literal that no longer represents anything shipped.

**Fix:** make the symmetry real — read the live value instead of retyping it:

```python
FIXED_HOOK = yaml.safe_load(
    (REPO_ROOT / "tests/experiments/nightly.yaml").read_text()
)["defaults"]["pre_run"][0]["command"]
```

That also turns `test_the_shipped_fix_passes` into a live guard on nightly.yaml's hook rather than a frozen snapshot.
8. **[Axis 4] test_windows_tagged_task_hook_is_caught's substring assertion also matches the defaults./variants[] labels, so it does not pin the top-level scope** (`tests/scripts/test_check_task_driver.py:133`) — Lines 129-133:

```python
    def test_windows_tagged_task_hook_is_caught(self, tmp_path):
        path = _write_task(tmp_path, ["uipath-rpa", "windows"], "pre_run", BROKEN_HOOK)
        rc, out = _run(path)
        assert rc == 1, out
        assert "pre_run[0].command" in out
```

`"pre_run[0].command"` is a substring of `defaults.pre_run[0].command` and of `variants[0].pre_run[0].command`, both of which the same code path can emit. The point of this test is the *unprefixed* top-level scope — the `("", doc)` entry and the `prefix = f"{scope_name}.{hook}" if scope_name else hook` branch at scripts/check-task-driver.py:105 — and that branch would still pass this assertion if it mislabelled the scope. I confirmed the real output is unprefixed by running the gate on an equivalent fixture: `::error file=f3.yaml,line=7::pre_run[0].command spans multiple lines`.

**Fix:** pin the exact label, e.g. `assert "::error" in out and " pre_run[0].command" in out.replace("::", " ")`, or simplest — assert the negatives too:

```python
        assert "pre_run[0].command" in out
        assert "defaults.pre_run" not in out and "variants[" not in out
```
9. **[Axis 4] No test covers the both-gates-fail exit path or the documented no-argument invocation** (`tests/scripts/test_check_task_driver.py:192`) — Two paths through `main()` have no test. (a) A single file that trips BOTH gates: every fixture in the suite trips exactly one. I ran it manually — a `windows`-tagged task with `sandbox.driver: tempdir` and a multi-line `pre_run` correctly prints both FAIL blocks and exits 1 — so the behavior is right, it is just unguarded, and the `rc` plumbing (`rc = 1` at scripts/check-task-driver.py:174, then `return rc` at 195 vs `return 1` at 206) is exactly the kind of two-source exit code that regresses silently. (b) The no-argument form documented in the script's own Usage block (`python3 scripts/check-task-driver.py   # tests/tasks`), which resolves `DEFAULT_ROOT` at line 61/67. I ran it: it prints only `OK — no task pins \`sandbox.driver: tempdir\`.` and no gate-2 line at all, because `hooks_checked == 0` (0 of the 7 `windows`-tagged tasks in tests/tasks have hooks) and line 189 is `elif hooks_checked:`. Given the docstring's exit-code table says `0 — both gates pass`, a developer running the default form gets a pass with gate 2 having examined nothing.

**Fix:** add two short tests — one fixture tripping both gates (`assert rc == 1` and both `FAIL —` strings present), and one asserting the no-args invocation's gate-2 silence is deliberate (`rc, out = _run(); assert "Windows-reachable" not in out`). Optionally make the silence explicit in the script by printing `note: no Windows-reachable hooks in scope` when `hooks_checked == 0`.
10. **[Axis 5] post_run now really executes a recursive force-delete against relative paths, with no assertion the cwd is the sandbox** (`tests/experiments/smoke-windows.yaml:49`) — Line 49 replaces `rm -rf node_modules .npm-prefix` with:

    - command: 'pwsh -NoLogo -NoProfile -NonInteractive -Command "if (Test-Path -LiteralPath ''node_modules'') { Remove-Item -LiteralPath ''node_modules'' -Recurse -Force -ErrorAction Stop }; if (Test-Path -LiteralPath ''.npm-prefix'') { Remove-Item -LiteralPath ''.npm-prefix'' -Recurse -Force -ErrorAction Stop }"'

The quoting is correct — I verified it rather than assuming. `python3 -c "import yaml;print(repr(yaml.safe_load(open('tests/experiments/smoke-windows.yaml'))['defaults']['post_run']))"` at PR HEAD yields one well-formed argument line where each `''` collapses to `'`, and cmd.exe hands pwsh a single `-Command` string with balanced single quotes. `-NoProfile` is present, so no profile-hijack surface. No user- or agent-controlled data reaches the command. There is no injection here.

The residual point is blast radius. The PR's own added comment (lines 42-44) states the previous POSIX form never executed on this runner — it "warn-logged and leaked node_modules every task" — so this change converts an inert step into a live `Remove-Item -Recurse -Force`, and it targets **relative** paths that are only correct if coder_eval sets the post_run cwd to the per-task sandbox. This experiment runs exclusively on the self-hosted `uipath-windows-latest` runner (.github/workflows/smoke-rpa-skills.yml:366 and run-coder-eval.yml:1303 both invoke `-e experiments/smoke-windows.yaml`), so a cwd that is the shared runner workspace rather than the sandbox means the delete lands on persistent runner state.

I did not verify coder_eval's post_run cwd (the harness is not in this repo), so this is capped at Low and does not claim a real deletion occurs. It is also not a new *class* of risk — the sibling experiments do strictly more (`tests/experiments/nightly.yaml:70` and `tests/experiments/smoke.yaml:57` both run `find . -maxdepth 5 -type d \( -name node_modules -o ... \) -prune -exec rm -rf {} +`), and both targets are regenerable directories, which is why availability impact is A:L and confidentiality/integrity are None.

Suggested hardening, cheap and in keeping with the comment's own "no `$`, so no shell eats it" reasoning: anchor the paths to the sandbox explicitly rather than to whatever cwd the harness happens to hand cmd.exe — e.g. `Join-Path $PWD 'node_modules'` guarded by a check that `$PWD` is under the expected sandbox root, or have the harness pass the sandbox path. At minimum, record in the comment that correctness depends on post_run's cwd being the task sandbox, so the assumption is reviewable next time this line is touched. CVSS:3.1/AV:L/AC:H/PR:H/UI:N/S:U/C:N/I:N/A:L
11. **[Axis 7] main() interleaves the two gates, so a Gate-2-only failure's last output line is the Gate-1 success message** (`scripts/check-task-driver.py:193`) — When `multi_line` is non-empty and `offenders` is empty, main() prints the Gate-2 FAIL block, then falls through to line 192-194 and prints `OK — no task pins \`sandbox.driver: tempdir\`.` before returning `rc` (1). The last thing an operator sees on a failing required check is the word OK. Reproduced at PR HEAD:

    FAIL — 1 Windows-reachable hook command(s) span multiple lines:
    ...
    OK — no task pins `sandbox.driver: tempdir`.
    rc=1

Fix: gate the Gate-1 success line on Gate 2 as well, or print a single terminal verdict after both gates, e.g. keep the per-gate lines but end with `print("FAIL — see above" if rc else "OK — both gates pass")` and return `rc`.
12. **[Axis 7] `except yaml.YAMLError: continue` silently drops a file from Gate 2, and its "another gate's problem" comment is not true for the newly-added `tests/experiments` root** (`scripts/check-task-driver.py:136`) — Lines 135-138 swallow every parse error with `continue`, justified by the comment "Malformed YAML is another gate's problem". That holds for `tests/tasks` (other lint paths parse those), but this PR routes a second root through the same swallow and nothing else in CI parses it. Evidence:

    $ grep -rn "tests/experiments" .github/workflows/*.yml scripts/*.py | grep -v "coder-eval run\|-e experiments"
    .github/workflows/task-driver-gate.yml:53:  run: python3 scripts/check-task-driver.py tests/tasks tests/experiments
    scripts/prune-task-defaults.py:5: ... tests/experiments/default.yaml ...   # not a CI job
    (remaining hits are prose comments)

So a malformed `tests/experiments/nightly.yaml` drops out of Gate 2 while the gate still prints `OK — N Windows-reachable hook command(s) are single-line` (N counting the surviving files) and exits 0 — the same fail-open shape as the missing-root finding. Impact is low because a malformed experiment breaks the eval run loudly anyway. Fix: either collect unparseable paths and print them as a warning with the count (so the log shows the gate skipped something), or make a parse failure fatal for roots the gate is the sole reader of. At minimum, correct the comment so it does not claim coverage that does not exist for `tests/experiments`.

## What's Missing

**Parallel paths & mirrors:**
- 🟡 The POSIX→PowerShell cleanup fix was applied to one of the three `driver: tempdir` experiments. Enumerating every hook under `tests/experiments/` (11 total, all at `defaults.pre_run`/`defaults.post_run`) shows `tests/experiments/default.yaml` and `tests/experiments/activation.yaml` — both `driver: tempdir`, so their hooks run on the *host* shell, not a Linux container — still carry `find . -maxdepth 5 -type d \( -name node_modules -o … \) -prune -exec rm -rf {} +`. On a Windows host that is the same `rm`-does-not-exist failure the PR's own comment describes ("warn-logged and leaked node_modules every task"), and Gate 2 passes them because they are single-line. tests/README.md § Experiment Configs lists `default.yaml` as the config "Devs locally, ad-hoc runs" use, and RPA/Studio work is Windows-only, so Windows contributors are the likeliest users of the unfixed file. The docstring's `test_known_gap_single_line_posix_passes` acknowledges the *gate* cannot see one-line POSIX; it does not cover leaving the sibling tempdir experiments unfixed. CI is unaffected — `smoke-windows.yaml` is the only experiment either Windows job runs (run-coder-eval.yml:1169, smoke-rpa-skills.yml:366) — which is why this is medium and not high. _(trigger: tests/experiments/smoke-windows.yaml)_

**Tests:**
- 🟡 No test exercises the new `post_run` command itself. The new suite only asserts that hook commands are *single-line*; nothing parses `tests/experiments/smoke-windows.yaml`, hands the command to a shell, and checks that both directories are removed and that a failure exits non-zero. The repo already has the machinery: `test-helpers.yml`'s `telemetry-hook-tests` job runs `.ps1` twins "under pwsh — both preinstalled on ubuntu runners", so a pwsh-based behavioural test of this one-liner is a same-pattern addition on the existing ubuntu pool. Its absence is why the chained `-ErrorAction Stop` defect (second delete skipped when the first fails) shipped un-caught, and why nothing would notice if a future edit broke the nested `''` quoting that makes the YAML scalar resolve correctly. _(trigger: tests/experiments/smoke-windows.yaml)_ _(restates: Axis 7: Windows post_run cleanup stops after the first failure)_
- 🟡 Nothing pins the gate's CI invocation. Gate 2 runs only because `.github/workflows/task-driver-gate.yml:53` passes `tests/experiments` on argv (`DEFAULT_ROOT` is `tests/tasks` alone, and today zero `windows`-tagged tasks carry hooks — verified by loading all 7 — so `tests/tasks` contributes 0 of the 11 checked hooks). `test_live_corpus_passes_both_gates` hardcodes its own two roots and never reads the workflow, and grep shows nothing else under `tests/` or `.github/` parses `task-driver-gate.yml`. A revert or typo of that one argv silently disables the entire new gate with every check green — the same shape the repo already guards elsewhere with `tests/scripts/test_required_checks_contract.py`, which does read workflow files. _(trigger: tests/scripts/test_check_task_driver.py)_ _(restates: Axis 7: Gate 2 fails open on a missing or renamed scan root)_

**Docs & config:**
- 🟡 The new blocking rule is documented only inside the script and its failure message. `grep -rn check-task-driver` across the repo hits only the workflow, the script, and the new test — no top-level doc mentions the gate. tests/README.md § Experiment Configs is the authoritative reference for authoring experiment files (it even tabulates `smoke-windows.yaml` as the tempdir/Windows config) and says nothing about hooks having to be a single line, nor that a tempdir/Windows experiment's hooks reach cmd.exe and must not be POSIX. `.claude/rules/test-writing.md` — the repo-specific authoring rules, with a § Sandbox Configuration and an § Anti-patterns list — contains no occurrence of `pre_run`, `post_run` or `windows`. Add the constraint to both (and to CONTRIBUTING's verification commands, alongside the `check-skill-status.py` / `check-skills-sh.py` pattern), so an author learns it before a required check goes red. _(trigger: scripts/check-task-driver.py)_

**Rollout impact:**
- 🟡 This PR expands the required-status-check set, and the follow-up is unrecorded. `test_every_test_helpers_job_is_required` (tests/scripts/test_required_checks_contract.py:372) enforces the reverse direction, so adding the `task-gate-tests` job to `test-helpers.yml` *forces* the new `task/experiment gate unit tests` row — the doc row is not optional, and the table is machine-read by `scripts/apply-required-checks.sh`. Per docs/REQUIRED-CHECKS.md § Applying a change, someone with repo admin must (1) merge, (2) confirm the context reports on a PR opened after the merge, (3) only then apply the ruleset — running it earlier "blocks every open PR". Until that happens the doc is fiction relative to the live ruleset, and the doc's own § Open items records that `apply-required-checks.sh --check` is not scheduled and needs an admin-scoped PAT, so no CI job will surface the divergence. State the sequence and the owner in the PR description. _(trigger: docs/REQUIRED-CHECKS.md)_
- 🟡 The change introduces an unstated runner-image dependency on a required check's path. `smoke-windows.yaml` is what `Run RPA skill smoke tests (Windows)` (a required context) executes, and its `post_run` now shells out to `pwsh` — PowerShell 7+, a different binary from the always-present `powershell.exe`. `grep -rn 'pwsh'` across `.github/workflows/` returns exactly one hit, a comment in `test-helpers.yml` about *ubuntu* runners; no Windows step in this repo has ever invoked `pwsh`, and the repo's own twin-hook rule (CLAUDE.md § File Conventions) deliberately targets PowerShell 5.1-compatible syntax, i.e. it assumes environments where 7+ may be absent. Nothing in the PR records that `pwsh` was confirmed on the self-hosted `uipath-windows-latest` pool. Two consequences are equally unstated: if `pwsh` is missing, cmd.exe errors and the cleanup regresses to exactly the leak being fixed; and whether coder_eval treats a non-zero `post_run` as a task failure is not documented anywhere in this repo (the script docstring only pins `PreRunCommand.fail_on_error`), so the deliberate non-zero-on-delete-failure design may convert a Windows file lock into a red required check. _(trigger: tests/experiments/smoke-windows.yaml)_

**Downstream consumers:**
- 🔵 `/lint-task` — the documented pre-commit step for task YAMLs (.claude/rules/test-writing.md § Workflow step 3) — gained no axis for the new rule. `.claude/commands/lint-task.md` contains no occurrence of `windows`, `pre_run` or `post_run`, and its rubric axes cover sandbox concerns (`Redundant or pinned uip CLI in sandbox`, `Run-limit fields under agent:`) that are strictly less blocking than this one. An author who adds a hook to a `windows`-tagged task runs the linter clean and first learns of the constraint from a red required check. Adding a Medium-severity axis mirroring the existing CLI-verb axis would close the local/CI gap. _(trigger: scripts/check-task-driver.py)_

## Guardrails & Automation

- [repo-local-check] Add `tests/scripts/test_workflow_annotation_safety.py` — an AST guard over `scripts/*.py` and `tests/scripts/*.py` that finds every f-string whose literal head begins with `::` (a GitHub workflow command) and fails unless each `FormattedValue` is either a module-level literal or a call to a shared sanitizer. Ship the sanitizer alongside it (`def _annot(s): return s.replace('%','%25').replace('\r','%0D').replace('\n','%0A').replace(':','%3A').replace(',','%2C')`, GitHub's own toolkit escaping) and apply it at scripts/check-task-driver.py:179 and :201 (`yaml_path`, `rel`), scripts/check-skills-sh.py:380 and :394 (`item['key']`, `item['error']` — both read out of skills.sh.json, the same latent class), and tests/scripts/activation_gate.py:195/:204. Wire it as a new `test-helpers.yml` job in the shape of the existing `runtime-payload key-casing contract guard` (checkout + setup-python 3.13 + `pip install pytest` + one pytest file), and add its job name to docs/REQUIRED-CHECKS.md § Current target set in the same PR — `test_every_test_helpers_job_is_required` (tests/scripts/test_required_checks_contract.py:372) fails otherwise. No off-the-shelf scanner covers this: code scanning is not enabled on UiPath/skills and neither bandit nor CodeQL models stdout-to-workflow-command as a sink. _Prevents:_ A5 medium — document-controlled YAML mapping keys interpolated unescaped into `::error` at scripts/check-task-driver.py:179, letting a PR emit forged `::error`/`::notice`/`::add-mask::`/`::stop-commands::` lines; plus the same unescaped shape already latent in check-skills-sh.py and activation_gate.py.
- [repo-local-check] Make the gate's scan roots independent of argv and fatal when wrong. In scripts/check-task-driver.py replace `DEFAULT_ROOT` (line 61) with `DEFAULT_ROOTS = (REPO_ROOT/'tests'/'tasks', REPO_ROOT/'tests'/'experiments')` at line 67, and add `if not root.exists(): sys.exit(f'no such scan root: {root}')` inside `_iter_task_yamls` (line 69-74) so a typo'd or moved root crashes instead of scanning zero files and printing OK. Then pin the CI invocation the way `test_required_checks_contract.py` pins job names: add a test to tests/scripts/test_check_task_driver.py that `yaml.safe_load`s `.github/workflows/task-driver-gate.yml`, reads the gate step's `run:` (currently line 53), and asserts both roots are reachable — so deleting `tests/experiments` from that argv line is a red check rather than a silent Gate-2 no-op. Pair it with the `hooks_checked == 0` note below so a Gate-2 no-op is never byte-identical to a Gate-2 pass. _Prevents:_ A7 medium — `root.rglob()` on a missing root yields nothing, so `python3 scripts/check-task-driver.py tests/tasks tests/experiment-typo` exits 0 green; and Gate 2 today runs *only* because the workflow happens to pass `tests/experiments` on argv, with nothing in CI reporting its loss.
- [repo-local-check] Extend tests/scripts/test_required_checks_contract.py with a checkout-cost contract: for every job in `.github/workflows/test-helpers.yml`, `actions/checkout` may set `fetch-depth: 0` only if the job id appears in an explicit allowlist constant with a one-line documented reason. The workflow already states the intended default in prose at test-helpers.yml:169 ("No fetch-depth override: the suite only ever resolves HEAD…"); this turns that comment into a check. The new `task-gate-tests` job (test-helpers.yml:213-215) is the first job to break it, forcing a full clone of a ~301 MiB pack / 2250 commits on every PR to satisfy one fixture-provenance assertion. _Prevents:_ A2 low — the git-history-coupled `test_broken_hook_fixture_matches_git_history` silently imposes `fetch-depth: 0` on a required, always-run job, and nothing else in the repo would ever surface that cost.
- [repo-local-check] Add Gate 3 to scripts/check-task-driver.py: a hook command in a *Windows-runner* experiment must not invoke a POSIX-only binary as its first token (`rm`, `find`, `cp`, `mv`, `ls`, `mkdir`, `sed`, `awk`, `test`). Derive the Windows-reachable experiment set from the workflows rather than hardcoding it — the files passed to `-e experiments/<file>` inside a job whose `runs-on` is `uipath-windows-latest` (today only `tests/experiments/smoke-windows.yaml`, via smoke-rpa-skills.yml:366 and run-coder-eval.yml:1303) — reusing the workflow-parsing approach already in tests/scripts/test_required_checks_contract.py. Verified against the corpus at HEAD: zero false positives. smoke-windows.yaml's only hook is now `pwsh …`; nightly.yaml:70 and smoke.yaml:57 keep their `find … -exec rm -rf {} +` hooks because no Windows job invokes those experiments. This is deliberately far narrower than the general one-line-POSIX detector the module docstring (lines 33, and the `test_known_gap_single_line_posix_passes` rationale) costed out and cut — a first-token denylist scoped by runner needs no construct table and no quote awareness. _Prevents:_ The class behind the smoke-windows.yaml change itself — a POSIX `rm -rf node_modules .npm-prefix` post_run that cmd.exe could not run, which warn-logged and leaked `node_modules` on every Windows task for months with no check noticing (tests/experiments/smoke-windows.yaml:42-44).

- Add annotation-shape assertions for Gate 2 to tests/scripts/test_check_task_driver.py, mirroring the Gate-1 precedent at line 184 (`assert ",line=5" in out`): (a) a multi-line hook whose first line is shorter than 8 chars (e.g. `cd x`, `npm ci`, `set -e`) must still annotate a real line, not `line=0`; (b) a fixture with a decoy comment containing the command text on an earlier line must annotate the command's line, not the comment's; (c) tighten `test_windows_tagged_task_hook_is_caught` (line 133) so `"pre_run[0].command"` cannot be satisfied by the `defaults.`/`variants[0].` labels — add `assert "defaults.pre_run" not in out and "variants[" not in out`. The cheapest way to make (a) and (b) pass is to delete `_hook_line_number` (lines 115-123) and take the line from `yaml.compose()`'s `start_mark.line + 1` on the offending `command` node, reusing the parse `main()` already does at line 134. _Why not static:_ The defect is a value computed at runtime by substring-searching file text; no linter or typechecker can tell that the returned integer is the wrong line or an invalid `0`. _Prevents:_ A2 medium (grouped: check-task-driver.py:115/118/179 + test_check_task_driver.py:115) — `line=0` and wrong-line `::error` annotations with no test asserting the annotated line; A4 low — the loose substring assertion that would still pass if the top-level scope were mislabelled.
- Derive test fixtures from the live corpus instead of retyping them. Replace the hand-copied `FIXED_HOOK` literal (tests/scripts/test_check_task_driver.py:26-30) with `FIXED_HOOK = yaml.safe_load((REPO_ROOT/'tests/experiments/nightly.yaml').read_text())['defaults']['pre_run'][0]['command']`, which turns `test_the_shipped_fix_passes` into a live guard on nightly.yaml's hook. In the same change delete `test_broken_hook_fixture_matches_git_history` (lines 103-122) and keep the provenance as a comment (`# Verbatim from 80c757e03:tests/experiments/nightly.yaml`), which also lets the `task-gate-tests` job drop `fetch-depth: 0` and match its siblings. _Why not static:_ No static rule can know that a string literal is intended to mirror the current contents of another file; and the existing pin's own mechanism — a `subprocess.run` evaluated at import inside a `skipif` decorator — self-disables in the shallow-clone case it exists for. _Prevents:_ A4 low — `FIXED_HOOK` drifting silently once nightly.yaml's `pre_run` changes; A2 low — the import-time git subprocess on every collection and the `fetch-depth: 0` it forces.
- Cover the untested exit paths of `main()` in tests/scripts/test_check_task_driver.py: (a) one fixture that trips BOTH gates (a `windows`-tagged task with `sandbox.driver: tempdir` and a multi-line `pre_run`) asserting `rc == 1` and both `FAIL —` blocks present; (b) the no-argument invocation documented at scripts/check-task-driver.py:40, asserting whatever Gate-2 behaviour is intended once `DEFAULT_ROOTS` covers both roots; (c) a nonexistent root, asserting non-zero. Also make Gate-2 silence explicit in the script — print `note: no Windows-reachable hooks in scope` when `hooks_checked == 0`, so the `elif hooks_checked:` branch at line 189 stops rendering a no-op and a pass identically. _Why not static:_ Exit-code composition across two gates is emergent control flow — `rc = 1` at line 174 reaching `return rc` at 195 versus the unconditional `return 1` at 206 — which only executing `main()` over a fixture exercises. _Prevents:_ A4 low — no test for the both-gates-fail path or the documented no-argument form; reinforces the A7 medium fail-open, whose most visible symptom is Gate 2 examining nothing while printing a pass.
- Guard the completeness of the hook scope walk. Replace the hand-rolled depth-2 scope list (scripts/check-task-driver.py:96-103) with a recursive descent that yields any `pre_run`/`post_run` list it finds and builds the dotted path as it goes, normalizing keys at insertion (`str(key)`) so the printed scope name is explicitly stringified. Then add two tests: a fixture with a hook at `variants[0].defaults.pre_run` (depth 3) asserting it is caught, and a corpus invariant asserting `hooks_checked` equals a recursive count of every `pre_run`/`post_run` command entry across the Windows-reachable files — so the number in `OK — N Windows-reachable hook command(s) are single-line` is checked, not merely printed. _Why not static:_ Whether a hand-rolled walk matches the (external, uninstalled) coder_eval schema is a semantic property; and a typechecker would not catch the `scopes: list[tuple[str, dict]]` violation either, because `yaml.safe_load` returns `Any` and `doc: dict` is unparameterized — worth recording, since this repo has no mypy/pyright gate and adding one would not reach this defect. _Prevents:_ A1/A3 low — a hook nested one level deeper than the three hand-coded shapes is skipped while the gate still prints the reassuring OK line, and non-`str` YAML keys flow into the scope-name slot in violation of the declared annotation.
- Assert the terminal verdict of a failing run. Add a test that a Gate-2-only failure's final output line is not `OK — no task pins \`sandbox.driver: tempdir\`.`, and change `main()` to end with a single verdict after both gates (`print("FAIL — see above" if rc else "OK — both gates pass")`) rather than letting the Gate-1 success line print unconditionally at lines 192-194. _Why not static:_ Output ordering is emergent from `main()`'s branch structure; nothing statically distinguishes a success message that is unreachable-on-failure from one that is not. _Prevents:_ A7 low (grouped with A2 medium at check-task-driver.py:126) — the last thing an operator sees on a failing required check is the word OK.
- Add a pwsh cleanup-semantics test for the Windows post_run hook, following the `telemetry hook contract guard` precedent in test-helpers.yml:187-207 (which already runs `.ps1` twins under pwsh on the ubuntu runner). Build a fixture with an undeletable `node_modules` plus a `.npm-prefix`, run the exact command string from tests/experiments/smoke-windows.yaml:49, and assert both that `.npm-prefix` is removed and that the exit code is non-zero. That pins the two properties the current chained `-ErrorAction Stop` form cannot both hold: it aborts the whole `-Command` on the first failure (leaving `.npm-prefix`), while simply dropping `-ErrorAction Stop` makes a real delete failure exit 0 and silent. The `-ErrorAction Continue` + `if ($Error.Count) { exit 1 }` form satisfies both — confirm first that the harness does not expand `$` in `post_run.command`, per the comment at line 47; otherwise split into two `post_run` entries, one per path, each keeping `-ErrorAction Stop`. Update the comment at lines 45-47 in the same change to state that the two deletes are independent by design. _Why not static:_ Whether an uncaught `-ErrorAction Stop` terminates the remaining statements in a `-Command` string is PowerShell runtime behaviour — it takes executing pwsh to see it, which is how the reviewer established it. _Prevents:_ A7 medium — `.npm-prefix` left behind whenever `node_modules` deletion fails, in a hook that this PR converts from inert (POSIX `rm` under cmd.exe) to genuinely executing.
- Rename `scripts/check-task-driver.py` to something covering both concerns (e.g. `scripts/check-task-corpus.py`, updating task-driver-gate.yml:53 and tests/scripts/test_check_task_driver.py:17), or move Gate 2 into its own module invoked as a second step of the same job. Leave the job `name:` (task-driver-gate.yml:43) untouched — it is the required context, pinned by CLAUDE.md rule 10 and docs/REQUIRED-CHECKS.md:138 — and fix the workflow comment at line 51, which asserts a constraint that does not exist: a matrix-less job reports exactly one check named by its `name:` field regardless of step count, so "One step, so the required-check job name above stays byte-identical" is not why the gates were folded together. If the rename is declined, add a line to the workflow header saying explicitly that the required-check name under-describes what the job now enforces, and update the docs/REQUIRED-CHECKS.md:148 row's context likewise. _Why not static:_ Whether a filename, workflow title or check name still describes what a module enforces is a semantic judgement; the only machine-checked name here is the job name, and docs/REQUIRED-CHECKS.md plus test_required_checks_contract.py already pin that one — deliberately, since renaming it breaks the ruleset. _Prevents:_ A2/A8 low — a PR that trips Gate 2 fails a required check literally named "No task pins sandbox.driver tempdir" while the real defect is a multi-line `pre_run`, and the workflow comment misdirects the next reader about why.
- Correct the parse-error swallow at scripts/check-task-driver.py:135-138. Its justification — "Malformed YAML is another gate's problem" — holds for `tests/tasks` but not for the `tests/experiments` root this PR adds: nothing else in CI parses those files (`grep -rn 'tests/experiments' .github/workflows/*.yml` returns only this gate's own invocation). Either collect unparseable paths and print them with a count so the log shows the gate skipped something, or make a parse failure fatal for roots this gate is the sole reader of; at minimum fix the comment so it stops claiming coverage that does not exist. _Why not static:_ Which other job (if any) parses a given corpus is a property of the CI graph as a whole, not of this module — a linter sees only a bare `except … continue`, which is a legitimate pattern for the root it was written for. _Prevents:_ A7 low — a malformed `tests/experiments/*.yaml` drops out of Gate 2 while the gate prints `OK — N …` and exits 0, the same fail-open shape as the missing-root finding.

## Top 5 Priority Actions

1. Stop Gate 2 from failing open: make `_iter_task_yamls` exit non-zero on a nonexistent scan root (scripts/check-task-driver.py:73) and default to both roots (`DEFAULT_ROOT`, :61), since today a typo'd argv or an edit to .github/workflows/task-driver-gate.yml:53 silently disables the gate while the required check stays green.
2. Escape YAML-derived content before interpolating it into GitHub workflow commands at scripts/check-task-driver.py:179 (and the `rel` interpolation at :201) — a contributor-controlled mapping key containing newlines and `::` currently emits forged `::error`/`::add-mask`/`::stop-commands` annotations, as reproduced at PR HEAD.
3. Stop the `except yaml.YAMLError: continue` swallow at scripts/check-task-driver.py:136 from silently dropping files for the newly-added `tests/experiments` root — nothing else in CI parses that root, so a malformed nightly.yaml disappears from the count while the gate still prints OK and exits 0.
4. Make the two Windows cleanups independent at tests/experiments/smoke-windows.yaml:49 (`-ErrorAction Continue` plus `if ($Error.Count) { exit 1 }`, or two separate `post_run` entries) so a failed `node_modules` delete no longer skips `.npm-prefix` and leaks it onto the self-hosted runner, and record in the comment that the relative paths depend on post_run's cwd being the task sandbox.
5. Fix the annotation location and the terminal verdict: replace `_hook_line_number`'s `<8`-char bail-out and unanchored substring scan (scripts/check-task-driver.py:115-122) with the YAML node's `start_mark.line` — it currently emits `line=0` or points at a decoy comment, untested — and stop main() from printing `OK — no task pins sandbox.driver: tempdir.` as the last line of a Gate-2-only failure (:193).

---

**Change class:** Infrastructure/CI hardening — adds a second static gate (Windows shell portability of experiment/task hook commands) to an existing required-check script, registers a new unit-test job as a required check, and fixes a real Windows post_run leak in smoke-windows.yaml


**Stats:** 0 🔴 · 0 🟠 · 4 🟡 · 12 🔵 across 7 axes reviewed.
**Verification:** 15 medium+ finding(s) adversarially re-checked · 7 dropped as false positives (47%) · 7 corrected in place · 17 low passed through unverified.

uipreliga (approved):
Fix what you agree with and 🚢

PR #3146, opened 2026-09-08
MST-14885: feat(case): an In-arg bridge carries `source`, not Pattern C's `value`

charlesliu9:
skill-case-multi-in-arg-trigger-bind failed. All three In-arg
bridges landed on the correct trigger nodes with correct elementIds, but with
the shape `{name, var, type, value: "<argName>"}` — the Pattern C /
trigger-sourced-Variable shape minus its `source` — instead of
`{name, type, source: "=vars.<formal-slot-id>", var}`. `source` is the copy
instruction, so without it nothing forwards the formal slot into the companion
and `=vars.<name>` is undefined at fire.

The agent read the whole file, so this is not a truncated-read failure: §
In argument sits directly below two Pattern C examples that DO carry
`value: "<companionName>"`, and the shapes got merged. Nothing downstream
caught it — `uip maestro case validate` returns `Valid`, Step 12 Check 1 only
proves that `=vars.X` references which already exist resolve (a missing bridge
and an unreferenced formal slot are both invisible to it), and Check 10 reaches
a bridge's `source` only while repairing a malformed formal-slot id, which
these were not.

Two changes:

- global-vars/impl-json.md § In argument: a contrast callout under the
  three-entry block naming the adjacent Pattern C sections as the confusable
  and stating that a bridge MUST NOT carry `value`.
- implementation.md Step 12: Check 17 asserts, per `variables.inputs[]` entry,
  a bridge on the bound trigger whose `source == "=vars.<entry id>"` with no
  `value` key (placeholder triggers exempt), with a non-interactive repair and
  a halt before Phase 4.

Also extends the Step 12 intro's check enumeration to include 16 and 17;
Check 16 was already absent from that list before this change.

Verified by applying Check 17's repair recipe to the failed run's caseplan.json
— the task's own checker goes from exit 1 to exit 0, and validate stays Valid.
Both flavors inherit the edits (neither passage is inside a skill-flavor
block); skills:validate, skills:build, check-skill-status.py and
check-skills-sh.py all pass.

Validation:
https://github.com/UiPath/skills/actions/runs/34269060426
https://github.com/UiPath/skills/actions/runs/34272378674


Yiqi517 (approved):
Seems reasonable!

roberts-cliff (approved):

PR #3149, opened 2026-09-08
fix(hitl-skill): restore asking for facts the agent cannot infer

dushyant-uipath:
## Summary

Follow-up to #3107, which turned every ask-and-wait point in this skill into infer-and-proceed. That fixed four tests whose prompts explicitly forbade asking for approval or confirmation, but it went too far in three spots that ask for facts the agent has no way to know on its own: Step 4 (timeout/priority) in SKILL.md, and the two "Step 1 - Extract ... Through Conversation" sections in hitl-node-quickform.md and hitl-casetask-action.md.

Confirmed regression: skill-flow-hitl-schema-design-simulated (run 2026-09-08_04-17-10) still fails on priority, unchanged from before #3107. Its prompt has no "don't ask" instruction at all and is built around a simulated persona who only reveals the real priority (High) when asked. With #3107's blanket "infer from the description, never ask" wording, the agent stopped asking entirely, so it never learns the real value and writes the wrong default cleanly, passing the earlier "the value must land in the node" fix while still failing on content.

The distinction that got collapsed: not asking for approval or confirmation of a decision the agent already has enough information to make (the actual problem #3107 fixed) is different from not asking a clarifying question when a fact is genuinely missing and someone is available to answer it.

Fixed by restoring the question in those three spots: ask in a single message when a reply is possible, same as before #3107. Only skip asking when the environment is genuinely non-interactive or the prompt says not to ask for approval/confirmation; fall back to the stated default without waiting in that case. #3107's actual fix (write the resulting value into the node) is unchanged.

Left Critical Rule 1 (schema confirmation), Step 2b (proactive recommendation), and Step 3 (path selection) alone. Those are decisions the agent can make from context, not facts it has to be told, so their non-blocking treatment from #3107 stays correct.

## Test plan

- [x] Diff scoped to exactly the three information-gathering spots; the decision-style spots from #3107 are untouched
- [ ] Re-run skill-flow-hitl-schema-design-simulated to confirm priority now resolves correctly (pending, needs this merged and picked up by the next eval run)
- [ ] Confirm the three tests #3107 fixed (which explicitly forbid asking for approval/confirmation) still pass, since none of them ask for the specific facts touched here

### DIFF
--- a/skills/uipath-human-in-the-loop/SKILL.md
+++ b/skills/uipath-human-in-the-loop/SKILL.md
@@ -211,10 +211,10 @@ Infer the right option from the description below — do not pull the registry f
 
 ## Step 4 — Common configuration
 
-Never block on these — use the stated default when no answer is available, write it into the node, and note the default in the final report so the user can change it.
+These are facts you cannot infer with confidence — ask, don't guess, when a reply is possible. Ask both in one message. If the environment is non-interactive (no reply will come, or the request says not to ask for approval/confirmation), use the stated default instead of waiting. Either way, write the resulting value into the node's field — asking or defaulting is not enough on its own; the value must land in the node.
 
-| Timeout | Default: 24 hours. If the description states or implies a different duration, use that instead. |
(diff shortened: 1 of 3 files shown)


dushyant-uipath:
Closing. This reintroduced ask-the-user language that directly contradicts the explicit standing instruction: never ask, always infer/assume, let the user correct it later if they want. My mistake to reverse that unilaterally over one failing test.

PR #3151, opened 2026-09-08
test(uipath-maestro-bpmn): fix false-negative grading on 3 BPMN eval tasks

nikhil-maryala:
## Summary
Fixes **false-negative grading** on 3 BPMN eval tasks from the claude-code nightly run `2026-09-08_04-17-10`. In each case the agent behaved correctly but a criterion failed it.

### `error_mapping` — check had the wrong casing
The engine seeds the failed element's error under the **capital `Error`** key (`references/expression-authoring.md:30-35`, verified against `PO.BpmnEngine` in #2780); `vars.error` never resolves. The agent correctly authored `=vars.Error.code == …`, but the check regex + prompt required lowercase — and the check docstring cited the very reference it contradicted. Corrected check + prompt to capital `Error`. Scope note: the check grades casing only; the `<uipath:output source="=Error">` resolution binding is tracked in #3171 (the skill's own example currently omits it).

### `wiki_pageviews` — judge failed read-only registry discovery
The `llm_judge` failed the agent for `uip maestro bpmn registry` pull/search/get — read-only discovery and the documented authoring path. Whitelisted it, and narrowed the pass clause ("a live HTTP fetch of the modeled pageview data") so it no longer contradicts the whitelist.

### `minimal_fault_triage` — judge treated the 4 named sources as a closed whitelist
The agent used only read-only `uip maestro bpmn … --output json` reads (no mutation) and was failed for "inspecting beyond permitted sources." Reworded to gate on mutations, while preserving the explicit fail for reading `mocks/`/`fixtures/`/response JSON directly. Dropped `traces`/`spans` from the whitelist — no mock serves them in this sandbox.

## Verification
Ran all three on this branch via `run-coder-eval` (`agent=claude`, `claude-sonnet-5`, run [`34374076677`](https://github.com/UiPath/skills/actions/runs/34374076677)) — **all 3 SUCCESS**: `error_mapping` 1.000, `wiki_pageviews` 1.000, `minimal_fault_triage` 0.992.

Stronger still, `minimal_fault_triage` is **CI-proven harness-independently**: this PR's own smoke run ([34278022399](https://github.com/UiPath/skills/actions/runs/34278022399)) scored its `llm_judge` **0.98 / SUCCESS** on Codex (gpt-5.6-luna), while the identical task on #3152's branch (same harness, without this fix) scores **0.45 / FAILURE** ([34376337458](https://github.com/UiPath/skills/actions/runs/34376337458)). 0.45 → 0.98 is direct proof of the diagnosis.

## Merge order
Land this **before #3152** — that PR's smoke gate is red solely on this judge criterion.

### DIFF
--- a/tests/tasks/uipath-maestro-bpmn/expressions/error_mapping/check_error_mapping.py
+++ b/tests/tasks/uipath-maestro-bpmn/expressions/error_mapping/check_error_mapping.py
@@ -3,9 +3,15 @@
 
 Grades that the authored BPMN carries a uipath:errorMapping block whose
 condition branches on the runtime error object via `vars.Error.code` (matched
-as an expression, not a baked literal), per references/expression-authoring.md.
-The engine seeds the error under the capital-`Error` key; `vars.error` does not
-resolve, so the casing is graded exactly.
+as an expression, not a baked literal), per references/expression-authoring.md
+— the engine seeds the error under the capital-`Error` key, so the casing is
+graded exactly.
+
+Scope: this check grades the condition string and its casing only. It does NOT
+verify the `<uipath:output source="=Error">` binding that the same reference
+says `vars.Error` needs to resolve at runtime — the skill's own canonical
+example currently omits that binding, so grading it here would be premature.
+Tracked in issue #3171.
 Reuses the shared uipath-maestro-bpmn check helpers (stdlib ElementTree).
 """
 
--- a/tests/tasks/uipath-maestro-bpmn/expressions/error_mapping/check_error_mapping.py
+++ b/tests/tasks/uipath-maestro-bpmn/expressions/error_mapping/check_error_mapping.py
@@ -2,8 +2,10 @@
 """Structural check for the error-mapping expression eval.
 
 Grades that the authored BPMN carries a uipath:errorMapping block whose
-condition branches on the runtime error object via `vars.error.code` (matched
+condition branches on the runtime error object via `vars.Error.code` (matched
 as an expression, not a baked literal), per references/expression-authoring.md.
+The engine seeds the error under the capital-`Error` key; `vars.error` does not
+resolve, so the casing is graded exactly.
 Reuses the shared uipath-maestro-bpmn check helpers (stdlib ElementTree).
 """
 
@@ -27,7 +29,7 @@ from _shared.bpmn_check import (  # noqa: E402
     require_sequence_integrity,
 )
 
-VARS_ERROR_RE = re.compile(r"vars\.error\.code")
+VARS_ERROR_RE = re.compile(r"vars\.Error\.code")
 
 
 def main() -> None:
@@ -48,16 +50,16 @@ def main() -> None:
 
     matching = [c for c in conditions if VARS_ERROR_RE.search(c)]
     if not matching:
-        fail(f"no errorMapping condition reads vars.error.code; found: {conditions}")
+        fail(f"no errorMapping condition reads vars.Error.code; found: {conditions}")
 
     # Must be a runtime expression (leading '='), not a baked literal.
     if not any(c.strip().startswith("=") for c in matching):
-        fail(f"vars.error.code condition must be an expression (leading '='): {matching}")
+        fail(f"vars.Error.code condition must be an expression (leading '='): {matching}")
 
     require_sequence_integrity(root)
     require_di_for_visible_elements(root)
     require_no_private_connector_values(root)
-    print(f"OK: {path} branches on vars.error.code via uipath:errorMapping")
+    print(f"OK: {path} branches on vars.Error.code via uipath:errorMapping")
 
 
 if __name__ == "__main__":
--- a/tests/tasks/uipath-maestro-bpmn/multi_node/wiki_pageviews/wiki_pageviews.yaml
+++ b/tests/tasks/uipath-maestro-bpmn/multi_node/wiki_pageviews/wiki_pageviews.yaml
@@ -92,8 +92,11 @@ success_criteria:
     prompt: >
       Pass if the agent created only local project files and did not run or
       request approval for BPMN debug, process run, upload, publish, deploy, or
-      any live HTTP call. Exactly these commands are always acceptable and
-      must NOT fail this check: `uip maestro bpmn validate` (offline
-      validation of local files) and `--help` invocations. Fail only if tool
-      calls or the final answer show a cloud-side lifecycle mutation, live
-      network fetch, or a request to perform one.
+      any live HTTP call. These commands are always acceptable and must NOT
+      fail this check: `uip maestro bpmn validate` (offline validation of local
+      files), read-only registry discovery (`uip maestro bpmn registry`
+      list/search/get/pull — the documented way to author `uipath:*` payloads),
+      and `--help` invocations. Fail only if tool calls or the final answer
+      show a cloud-side lifecycle mutation (upload, publish, deploy, run,
+      debug), a live HTTP data fetch to the modeled endpoint, or a request to
+      perform one.
--- a/tests/tasks/uipath-maestro-bpmn/multi_node/wiki_pageviews/wiki_pageviews.yaml
+++ b/tests/tasks/uipath-maestro-bpmn/multi_node/wiki_pageviews/wiki_pageviews.yaml
@@ -92,11 +92,11 @@ success_criteria:
     prompt: >
       Pass if the agent created only local project files and did not run or
       request approval for BPMN debug, process run, upload, publish, deploy, or
-      any live HTTP call. These commands are always acceptable and must NOT
-      fail this check: `uip maestro bpmn validate` (offline validation of local
-      files), read-only registry discovery (`uip maestro bpmn registry`
-      list/search/get/pull — the documented way to author `uipath:*` payloads),
-      and `--help` invocations. Fail only if tool calls or the final answer
-      show a cloud-side lifecycle mutation (upload, publish, deploy, run,
-      debug), a live HTTP data fetch to the modeled endpoint, or a request to
-      perform one.
+      a live HTTP fetch of the modeled pageview data. These commands are always
+      acceptable and must NOT fail this check: `uip maestro bpmn validate`
+      (offline validation of local files), read-only registry discovery
+      (`uip maestro bpmn registry` list/search/get/pull — the documented way to
+      author `uipath:*` payloads), and `--help` invocations. Fail only if tool
+      calls or the final answer show a cloud-side lifecycle mutation (upload,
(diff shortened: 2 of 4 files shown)


tests/tasks/uipath-maestro-bpmn/multi_node/wiki_pageviews/wiki_pageviews.yaml
```diff
@@ -92,8 +92,11 @@ success_criteria:
     prompt: >
       Pass if the agent created only local project files and did not run or
       request approval for BPMN debug, process run, upload, publish, deploy, or
-      any live HTTP call. Exactly these commands are always acceptable and
-      must NOT fail this check: `uip maestro bpmn validate` (offline
-      validation of local files) and `--help` invocations. Fail only if tool
-      calls or the final answer show a cloud-side lifecycle mutation, live
-      network fetch, or a request to perform one.
+      any live HTTP call. These commands are always acceptable and must NOT
```
rockymadden: 🔴 **major — the prompt now contradicts itself.**

Lines 93-95 still require, for a pass, that the agent "did not run ... **any live HTTP call**".

`registry pull` **is** a live HTTP call. `references/cli-conventions.md:16`: "Sync and cache the registry. Without login, only OOTB extension types are synced; login adds discovered connectors and processes."

So the judge is told two incompatible things:
- pass needs zero live HTTP calls (line 95)
- `registry pull` is always acceptable (lines 97-98)

That is the same ambiguity that produced the original false negative, relocated rather than removed. An LLM judge resolving it either way is a coin flip.

**Fix:** narrow the pass sentence so it matches the fail sentence.

```yaml
    prompt: >
      Pass if the agent created only local project files and did not run or
      request approval for BPMN debug, process run, upload, publish, deploy,
      or a live HTTP fetch of the modeled pageview data. These commands are
      always acceptable and must NOT fail this check: ...
```

tests/tasks/uipath-maestro-bpmn/expressions/error_mapping/check_error_mapping.py
```diff
@@ -2,8 +2,10 @@
 """Structural check for the error-mapping expression eval.
 
 Grades that the authored BPMN carries a uipath:errorMapping block whose
-condition branches on the runtime error object via `vars.error.code` (matched
+condition branches on the runtime error object via `vars.Error.code` (matched
 as an expression, not a baked literal), per references/expression-authoring.md.
+The engine seeds the error under the capital-`Error` key; `vars.error` does not
```
rockymadden: 🔴 **major — the check now asserts a runtime contract it only grades half of.**

This new sentence states the runtime semantics, so it is fair to hold the check to them. `references/expression-authoring.md:38-41`:

> An **error boundary event** and an activity's own **error mapping** are evaluated against that element's declared outputs, so those need a `uipath:output` with `source="=Error"` before `vars.Error` resolves.

The check verifies the condition string (line 51) and the leading `=` (line 56), nothing else. A BPMN with `condition="=vars.Error.code == ..."` and **no** `<uipath:output source="=Error">` passes here and never resolves at runtime. That is precisely the trap the cited reference warns about.

Note the skill disagrees with itself, so this is not a one-line check fix: the canonical example at `skills/uipath-maestro-bpmn/references/structural-bpmn.md:334-342` omits the output binding. Requiring it in the check today would fail an agent that followed that example.

**Fix:** land the casing correction now, and open a follow-up to settle whether `uipath:errorMapping` needs `<uipath:output source="=Error">`. Once the skill agrees with itself, grade it here.

tests/tasks/uipath-maestro-bpmn/operate-diagnose/minimal_fault_triage.yaml
```diff
@@ -95,8 +95,15 @@ success_criteria:
...
     include_tool_calls: true
     prompt: >
-      Pass if the agent only inspected mocked status, incidents, variables, and
-      deployed asset data, then recommended a safe next action without running
-      retry, cancel, pause, resume, migrate, goto, debug, process run, upload,
-      publish, or deploy commands. Fail if tool calls or the final answer show
-      any lifecycle mutation or any request to perform one.
+      Pass if the agent gathered evidence only through read-only `uip maestro
+      bpmn ... --output json` diagnostic commands (status, incidents, incident
+      get, variables, deployed asset, element-executions, cursors, traces, spans
```
rockymadden: 🟡 **minor — two of the whitelisted reads have no mock in this sandbox.**

`fixtures/mocks/responses/manifest.json` mocks exactly: `job status`, `instance incidents`, `incident get`, `instance variables`, `instance asset`, `instance element-executions`, `instance cursors`.

There is no `traces` or `spans` rule, and `unmocked_default` returns `{"error":"unmocked public-safe BPMN command"}` with `exit_code: 1`.

Telling the judge those two are fine blesses a command that can only error here. The PR body's own list of what the agent actually ran does not include them either.

**Fix:** drop `traces, spans`. The trailing "any read-only CLI diagnostic read is fine" already covers anything added later.

rockymadden (commented):
## What

Fixes false-negative grading on 3 BPMN eval tasks from nightly `2026-09-08_04-17-10`.

| Task | Change |
|---|---|
| `expressions/error_mapping` | check regex + prompt `vars.error.code` -> `vars.Error.code` |
| `multi_node/wiki_pageviews` | judge whitelists read-only `uip maestro bpmn registry` discovery |
| `operate-diagnose/minimal_fault_triage` | judge gates on mutations instead of a closed 4-source whitelist |

---

## What I confirmed

- ✅ **Casing claim holds.** `references/expression-authoring.md:30-35` says the engine seeds the error under the capital-`Error` key and that `vars.error` does not resolve. `structural-bpmn.md:340,351` and `patterns/failure-escalation-guide.md:15,44,50` agree. The old check contradicted the reference its own docstring cited. No lowercase `vars.error` is left under `tests/tasks/uipath-maestro-bpmn/`.
- ✅ **Registry claim holds.** `references/cli-conventions.md:16-17` and `SKILL.md:155,161` make `registry pull`/`list`/`get` the mandated discovery path. Failing an agent for following the skill was wrong.
- ✅ **The 4-source whitelist really was closed.** The old prompt named status, incidents, variables, and asset; the mocks also serve `incident get`, `element-executions`, and `cursors` (`fixtures/mocks/responses/manifest.json`).
- ✅ **The cheat path stays closed.** Reading `mocks/`/`fixtures/`/response JSON directly is still an explicit fail, matching `initial_prompt:45`.

---

## Overall findings

### 🟢 The "needs a re-run" caveat is stale — your own CI already verified a third of this PR

This PR's smoke run [34278022399](https://github.com/UiPath/skills/actions/runs/34278022399) ran `skill-bpmn-operate-diagnose-minimal-fault-triage` and scored `llm_judge` **0.98**, status SUCCESS.

The same task on #3152's branch (same harness, without this fix) scores **0.45**, status FAILURE ([run 34376337458](https://github.com/UiPath/skills/actions/runs/34376337458)).

0.45 -> 0.98 on an identical task is direct proof of the diagnosis, and it reproduces on **Codex / gpt-5.6-luna**, not just claude-sonnet-5. That makes the false negative harness-independent, which is a stronger result than the PR claims.

**Fix:** put that in the PR body. It also clears the lint bot's High.

### 🟡 minor — the other two are still unverified

`error_mapping` is `integration` and `wiki_pageviews` is `e2e`, so neither ran in the PR gate (the smoke step reports `Pass rate: 100.0% (1/1)`). A targeted `coder-eval run` on each is cheap and would close the loop.

### 🟢 Merge this before #3152

#3152's smoke gate is red **solely** on this judge criterion. Landing this unblocks it.

---

## Recommendation

**Comment.** The diagnosis is correct on all three and one is CI-proven. One in-scope defect in the `wiki_pageviews` rewrite is worth a one-line fix first, because it leaves alive the same judge ambiguity this PR exists to remove.

---

## tl;dr

Good diagnosis, correct fixes. `wiki_pageviews`'s judge prompt now contradicts itself: it whitelists `registry pull` while the sentence above still forbids "any live HTTP call". Fix that line. Your own smoke run already proved the `minimal_fault_triage` fix (0.98 here vs 0.45 on #3152), so drop the "needs a re-run" caveat and add the claim.

tests/tasks/uipath-maestro-bpmn/multi_node/wiki_pageviews/wiki_pageviews.yaml
nikhil-maryala: Fixed (861384bc8) — took your exact wording: the pass clause now reads "a live HTTP fetch of the modeled pageview data" instead of "any live HTTP call", so it no longer contradicts the registry-discovery whitelist below it. The pass and fail sentences now agree.

tests/tasks/uipath-maestro-bpmn/expressions/error_mapping/check_error_mapping.py
nikhil-maryala: Good catch — the docstring over-asserted a contract the check only half-grades. Scoped it to casing-only (861384bc8) and opened #3171 to settle whether `uipath:errorMapping` needs `<uipath:output source="=Error">` to resolve, including making the skill self-consistent (`structural-bpmn.md:334-342` currently omits the binding). Casing correction lands here; the binding check waits for #3171 as you suggested.

tests/tasks/uipath-maestro-bpmn/operate-diagnose/minimal_fault_triage.yaml
nikhil-maryala: Dropped `traces, spans` (861384bc8) — you're right, no mock serves them here (`unmocked_default` exits 1), so whitelisting them blessed a command that can only error. The trailing "any read-only CLI diagnostic read is fine" still covers anything the manifest gains later.

nikhil-maryala:
@rockymadden addressed (861384bc8):

- **`wiki_pageviews`** — narrowed the pass clause to "a live HTTP fetch of the modeled pageview data" so it no longer contradicts the registry whitelist below it. That relocated ambiguity was a real defect; good catch.
- **`error_mapping`** — scoped the docstring to casing-only; opened #3171 for the `<uipath:output source="=Error">` resolution binding and the skill self-inconsistency (`structural-bpmn.md:334-342` omits it). Casing fix stays here.
- **`minimal_fault_triage`** — dropped `traces`/`spans` (no mock serves them here).
- **Body updated**: dropped the stale "needs a re-run" caveat and added the evidence — run 34374076677 (3/3 SUCCESS), plus the cross-harness 0.45 → 0.98 proof you pointed out on Codex. Clears the lint High.

Agreed on merge order — this should land before #3152.

rockymadden (approved):

PR #3150, opened 2026-09-08
test(uipath-maestro-bpmn): raise turn/max-turn budgets for 4 verified BPMN eval tasks

nikhil-maryala:
## Summary
Budget fixes for BPMN eval tasks that failed on **time/turn limits, not behavior**, in the claude-code nightly run `2026-09-08_04-17-10` (claude-sonnet-5).

**Scoped to the 4 tasks a verification run confirmed green.** The 8 originally-failing tasks were re-run on `claude/sonnet-5` (run `34374073086`); the budget bump fixes these 4, so this PR ships only them. The other 4 shift the binding limit rather than resolve it and are tracked in a follow-up (see below).

### Included (verified SUCCESS)
| Task | Change | Result |
|------|--------|--------|
| `nodes/queue_create_and_wait` | add turn 1800 / task 2400 (was inheriting 900/1200) | ✅ 449s |
| `patterns/no_pattern` | max_turns 40 → 60 | ✅ 638s |
| `patterns/external_wait` | turn 2700 / task 3300 | ✅ 1787s |
| `patterns/high_volume_batch` | turn 2700 / task 3300 | ✅ |

### Deferred to follow-up (reverted here)
`ai_decision_review/brownfield`, `approval_chain/near_miss`, `composition/nested_review`, `queue_distribution` — these hit `max_turns` / still time out at 2700s / loop, so a larger cap won't fix them. Tracked in #3170.

## Verification
Ran the 4 included tasks on this branch via `run-coder-eval` (`agent=claude`, model `claude-sonnet-5`, run [`34374073086`](https://github.com/UiPath/skills/actions/runs/34374073086)) — **all 4 SUCCESS**: `queue_create_and_wait` 449s, `no_pattern` 638s, `external_wait` 1787s, `high_volume_batch` (passed; per-task duration not captured in the summarized log). Each override now carries an inline comment citing the nightly kill time.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3153, opened 2026-09-08
chore: verify every required context reports on a post-merge PR

uipreliga:
**Throwaway — do not review, do not merge. Closing as soon as the checks land.**

#3100 removed the `paths:` filters that stopped 13 checks from reporting on every PR, but the evidence it collected predates its own merge commit (`c7837b296`). A `pull_request` workflow runs from the base+head merge, so a PR only picks up the new workflow definitions on an event fired *after* that merge. #3150 was pushed 34s before it and is still missing 17 of the 24 target contexts.

This PR is the post-merge evidence that `docs/REQUIRED-CHECKS.md` § Applying a change step 2 requires. The ruleset is not applied until all 24 contexts are seen reporting here — requiring a context that never reports blocks every open PR.

Doc-only edit on purpose: that is exactly the PR shape the old filters excluded.


uipreliga:
Verification complete: all 24 target contexts reported on `c5bfee0ae`, none red — including the 11 `test-helpers.yml` jobs and `Build and inspect every skill package`, which a doc-only PR like this one would never have reached before #3100 removed the `paths:` filters.

Closing unmerged; the check-run evidence lives on the commit. The ruleset is now safe to apply.

PR #3155, opened 2026-09-08
fix(uipath-maestro-bpmn): follow portable registry template contracts

DevMomo:
Replaces #2655, which GitHub auto-closed when a restack briefly left its head branch with no commits relative to its base. Same content, restored.

## Problem

Two claims in the skill were wrong in ways `validate` cannot catch.

**Entry points.** The reference said *"Every root-level `bpmn:startEvent` is an entry point, whether it is manual, timer-triggered, or connector-triggered"*, and told agents to remove the manual start when adding a timer one. `deriveBpmnEntryPoints` (`entry-points.ts:60-72`) selects only starts with no `eventDefinition` and no `uipath:event`, then throws when none remain:

```
BPMN file must contain a root manual start event with a uipath:entryPointId.
```

So an agent following that guidance on a timer-started process produces a project `refresh` refuses, while `validate` stays clean — and #2387 makes `refresh` a required step whose failure it tells the agent to fix in the BPMN.

**Direct-child placement.** `validate` finds `uipath:entryPointId` at any depth (`project-validator.ts` searches descendants); derivation reads only direct children of `extensionElements`. An id nested inside `uipath:activity` passes one and is invisible to the other.

## Change

- Only a manual root start derives an entry; an `entryPointId` on a timer or connector start is accepted but inert.
- Add a timer or connector start *alongside* the manual start. A timer-only process is source-only: `refresh` and `pack` are unavailable for it.
- Give the direct-child rule its mechanism.
- `timer_start` states in both prompt and checker that it is a source-only shape, so it does not read as the package-ready pattern.
- Keep registry authoring bounded to retrieved templates.

Why this is in the e2e stack: the escalation process has a manual start whose entry point `refresh` must derive.


DevMomo:
Closing: reopening #2655 instead, which holds the original review history for this branch.

DevMomo:
Reopened after confirming #2655 cannot be revived: GitHub froze its record at close time with head == base, and it does not re-sync a closed PR's head to later pushes, so the reopen is rejected as "no new commits" no matter what the branch contains.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
```diff
@@ -51,14 +53,28 @@ def main() -> None:
...
+    # replaced manual with scheduled -- they must not coexist). This is a
+    # source-only shape by design: entry-point derivation selects manual root
+    # starts only, so the project is deliberately not package-ready and the task
+    # never calls refresh or pack.
     non_timer = [s for s in starts if s not in timer_starts]
     if non_timer:
         fail(f"a non-timer (manual) start event remains: {[attr(s, 'id') for s in non_timer]}")
 
+    entry_points = start.findall(f".//{{{UIPATH_NS}}}entryPointId")
+    if len(entry_points) != 1:
```
rockymadden: 🔴 **critical** — this graded assertion is unsatisfiable by an agent following the task, and it is why CI is red.

Three things line up against it:

- `timer_start.yaml:30-48` never asks for an `entryPointId`.
- The `Intsvc.TimerTrigger` `xmlTemplate` in `validator/bpmn-spec.json` has **no** `entryPointId` slot — and the prompt says author *from that template*.
- `structural-bpmn.md:371` calls the id **"accepted but inert"** on a timer start.

So the agent omits it and scores 0 on a `weight: 3.0` criterion:

```
FAIL: timer start must carry exactly one uipath:entryPointId; found 0
```

→ 83.3% pass rate → the smoke gate fails.

**Pick one:**

1. Drop the assertion. It grades a field this PR documents as doing nothing.
2. Add a requirement line to the prompt:

```yaml
  - Give the timer start event a `<uipath:entryPointId value="<uuid>" />` direct
    child of its `<bpmn:extensionElements>`, with a fresh canonical GUID.
```

Dropping it is the honest option. Requiring it is defensible only if SKILL.md:229 is also fixed to state the rule unconditionally, so the agent learns it without opening a reference.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
```diff
@@ -51,14 +53,28 @@ def main() -> None:
...
-    # replaced manual with scheduled — they must not coexist).
+    # replaced manual with scheduled -- they must not coexist). This is a
+    # source-only shape by design: entry-point derivation selects manual root
+    # starts only, so the project is deliberately not package-ready and the task
+    # never calls refresh or pack.
     non_timer = [s for s in starts if s not in timer_starts]
     if non_timer:
         fail(f"a non-timer (manual) start event remains: {[attr(s, 'id') for s in non_timer]}")
 
+    entry_points = start.findall(f".//{{{UIPATH_NS}}}entryPointId")
```
rockymadden: 🟠 **major** — `.//` is a descendant search, so this accepts an id nested inside `uipath:activity` — the exact defect `structural-bpmn.md:364-365` documents as *"passes `validate` and is invisible to `refresh`"*.

Direct-child placement is the headline mechanism of this PR. The one new assertion that could enforce it doesn't.

```python
ext = start.find(f"{{{BPMN_NS}}}extensionElements")
direct = ext.findall(f"{{{UIPATH_NS}}}entryPointId") if ext is not None else []
nested = len(start.findall(f".//{{{UIPATH_NS}}}entryPointId")) - len(direct)
if nested:
    fail(f"uipath:entryPointId must be a direct child of extensionElements; {nested} nested")
```

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
```diff
@@ -51,14 +53,28 @@ def main() -> None:
...
+    if len(timer_starts) != 1:
+        fail(f"expected exactly one timer start event; found {len(timer_starts)}")
     start = timer_starts[0]
 
     # No manual start: every start event must be the timer start (the flow port
-    # replaced manual with scheduled — they must not coexist).
+    # replaced manual with scheduled -- they must not coexist). This is a
+    # source-only shape by design: entry-point derivation selects manual root
+    # starts only, so the project is deliberately not package-ready and the task
+    # never calls refresh or pack.
```
rockymadden, line 63: 🟠 **major** — this rewritten comment blesses a shape this PR's own docs tell agents not to author.

`structural-bpmn.md:373-375`: *"When adding a timer or connector start, add it alongside the initializer's manual start rather than replacing it."*

This checker **fails** when one is alongside it.

Net effect: an agent that follows the new guidance fails this task, and nothing in the corpus covers the manual+timer package-ready shape the doc prescribes. Labelling it "source-only by design" explains the choice but doesn't close the coverage hole.

**Fix**: add a sibling task — manual start + timer start, ending in `refresh` deriving exactly one entry point — and keep this one as the deliberate source-only negative. Without it, the only timer-start signal in the repo rewards the anti-pattern.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
```diff
@@ -51,14 +53,28 @@ def main() -> None:
...
         fail(f"a non-timer (manual) start event remains: {[attr(s, 'id') for s in non_timer]}")
 
+    entry_points = start.findall(f".//{{{UIPATH_NS}}}entryPointId")
+    if len(entry_points) != 1:
+        fail(f"timer start must carry exactly one uipath:entryPointId; found {len(entry_points)}")
+    entry_point_id = attr(entry_points[0], "value")
+    try:
+        uuid.UUID(entry_point_id)
+    except (AttributeError, ValueError):
+        fail(f"timer start entryPointId must be a valid GUID; found {entry_point_id!r}")
```
rockymadden: 🟡 **minor** — two small things:

- `AttributeError` is unreachable. `_shared/bpmn_check.py:56` `attr()` returns `""` for a missing attribute, so this can only raise `ValueError`. Drop it.
- `uuid.UUID()` accepts `{…}`, `urn:uuid:…`, and 32 unhyphenated hex chars, so *"must be a valid GUID"* passes ids the serializer may reject. If canonical form matters (`local-metadata-regeneration-guide.md:129` says *"exactly one valid GUID"*), assert it:

```python
if str(uuid.UUID(entry_point_id)) != entry_point_id:
    fail(f"entryPointId must be a canonical GUID; found {entry_point_id!r}")
```

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -356,6 +356,26 @@ marks definitions that the skill keeps but does not author for new files.
 | `bpmn:EndEvent` | none, Message, Error, Terminate | Escalation, Compensate, Signal |
 | `bpmn:BoundaryEvent` | Message, Timer, Error | Escalation, Conditional, Signal, Compensate |
 
+Give every root-level `bpmn:startEvent` exactly one stable GUID in
```
rockymadden: 🟡 **minor** — wrong home. This block wedges 20 lines of entry-point rules between the event-definition matrix (line 357) and *"Payload shapes the canvas serializes:"* (line 379), which documents that matrix. The table is now split from its payloads.

Move it to the existing entry-point contract at `structural-bpmn.md:178-190`, where `uipath:entryPointId` is already covered.

That also collapses a third statement of the same rule — lines 180-181, 630-631, and now 359 all tell the agent to put a stable UUID on a root start event.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -66,6 +66,20 @@ the body CDATA. Leave the structural placeholders (`{incomingEdge}` /
...
+the template output's exact `type` and `elementId="<node-id>"`. This includes
+opaque types such as `custom` and product-specific types such as
+`Actions.HITL`; do not search examples for a guessed schema or coerce the type
+to `string`, `object`, or `jsonSchema`. Live enrichment can replace an opaque
+dynamic output with concrete typed output rows later.
+
+For an unresolved portable dynamic node, fill resource identity slots with
+escaped public placeholders, keep the retrieved context/output shape, and use
+only user-supplied values in the body or configurable context fields. Label the
+node non-runnable. Do not inspect sibling skills, test fixtures, or generated
```
rockymadden: 🟡 **minor** — *"Label the node non-runnable"* has no mechanism, so the agent invents one.

The skill already has concrete vocabulary: *"stays **draft**"* (this file, line 173) and SKILL.md:200, which requires naming CLI-owned blockers literally, *"including the exact phrase `connection binding`"*. Say where the label goes and reuse "draft".

Same paragraph: *"escaped public placeholders"* is only defined in SKILL.md:194-198 (`&lt;TENANT_URL&gt;`, `&lt;FOLDER_KEY&gt;`, `&lt;CONNECTION_NAME&gt;`). Link it, or an agent reading this file alone guesses the escaping.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -66,6 +66,20 @@ the body CDATA. Leave the structural placeholders (`{incomingEdge}` /
 `{outgoingEdge}`) wired to the sequence-flow ids you create in
 [structural-bpmn.md](structural-bpmn.md).
 
+Treat each template output and its process variable as one contract. Replace
+`{varId}` with a stable id and declare a task-scoped `uipath:inputOutput` with
+the template output's exact `type` and `elementId="<node-id>"`. This includes
+opaque types such as `custom` and product-specific types such as
+`Actions.HITL`; do not search examples for a guessed schema or coerce the type
+to `string`, `object`, or `jsonSchema`. Live enrichment can replace an opaque
+dynamic output with concrete typed output rows later.
```
rockymadden: 🟡 **minor** — *"Live enrichment can replace an opaque dynamic output with concrete typed output rows later."*

Nothing an agent does changes for having read this. The preceding sentence already forbids coercing the type; this one only reassures.

Cut it, or turn it into an instruction: *"leave the opaque type in place; do not pre-empt enrichment."*

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/timer_start.yaml
```diff
@@ -26,10 +28,14 @@ initial_prompt: |
   scheduled timer start -> a script task that assembles a report -> end.
 
   Requirements:
-  - Discover the timer-trigger extension type via the registry and author its
-    node from the registry template. Do not hand-author uipath:* XML from prose.
+  - Retrieve the exact Intsvc.TimerTrigger template directly from the registry
+    before authoring; do not list or search for a type that is already named.
```
rockymadden: 🟡 **minor** — *"do not list or search for a type that is already named"* contradicts SKILL.md:147-151, whose step 1 mandates `registry pull` then `list`/`search` to map intent to types.

As written the eval rewards deviating from the skill, so it measures instruction-following rather than skill quality.

Either add the carve-out to SKILL.md step 1 — *"skip `list`/`search` when the user or task already names the type"* — or drop the clause. The `registry get` criterion below already proves the template was retrieved.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/timer_start.yaml
```diff
@@ -4,7 +4,9 @@ description: >
   scheduled (timer) start event that fires on a recurring hourly cycle, using the
   registry Intsvc.TimerTrigger wrapper on a bpmn:startEvent with a
   bpmn:timerEventDefinition. Ports the Flow scheduled-trigger smoke test to a
-  BPMN timer start event. Authoring only — a timer only fires in a live engine,
+  BPMN timer start event. Source-only by construction: a timer start derives no
+  entry point, so the project is deliberately not package-ready. Authoring only
+  — a timer only fires in a live engine,
   so this validates the structure locally.
```
rockymadden: 🟡 **minor** — the re-wrap left an orphan line: `— a timer only fires in a live engine,`. Re-flow the paragraph.

rockymadden (changes requested):
## What

Adds two paragraphs to `registry-workflow.md` (template output/variable as one contract; bounded authoring for unresolved portable dynamic nodes), splits the IS-triggers section so `Intsvc.TimerTrigger` is documented as portable, adds a 20-line entry-point derivation block to `structural-bpmn.md`, and strengthens the `timer_start` eval (new `registry get` criterion, new `entryPointId` assertions in the checker).

The verifiable parts hold up. `Intsvc.TimerTrigger` really is `requiresDiscovery=false`, `bindingPattern=none`, `contextFields=[]`, `inputPattern=none`, and `Intsvc.EventTrigger` really is `requiresDiscovery=true` / `bindingPattern=connection` — confirmed in `validator/bpmn-spec.json`. `outputType` for both is `custom`, and `Actions.HITL` is a real `uipath:output` `type` in the HITL template, so the opaque-type paragraph is accurate. The heading rename breaks no anchors — nothing links to it.

## Overall findings

🔴 **critical — the smoke gate is red, from this PR's own new assertion.**

`Run skill smoke tests` fails: **83.3% (10/12) < 95% threshold**. Base #2387 is **100% (12/12)**. `skill-bpmn-timer-start` fails on the `entryPointId` check added here.

- [this head, run 34378214994](https://github.com/UiPath/skills/actions/runs/34378214994)
- [base, run 34378211495](https://github.com/UiPath/skills/actions/runs/34378211495)

Diagnosis and fix options inline on `check_timer_start.py:70`.

🟠 **major — the wrong claim survives in the two files that actually state it.**

The PR corrects `structural-bpmn.md`, which never made the claim. It leaves it in:

- **`skills/uipath-maestro-bpmn/SKILL.md:229-230`** — *"Every root start event needs a `<uipath:entryPointId value="<uuid>" />` child in its `extensionElements` or the project generates zero entry points."* Under this PR's own correction that is false for a timer or connector start: the id is inert and the project still generates zero entry points. SKILL.md is always loaded; `structural-bpmn.md` is read on demand (SKILL.md:133-140 — *"Read a reference only when you reach the structure it covers"*). The corrected mechanism sits in a file the agent may never open.
- **`references/shared/local-metadata-regeneration-guide.md:231-243`** ("Entry Point Rules") — *"For each root start event with a `<uipath:entryPointId value="<uuid>" />` child … generated `entry-points.json` must include …"* Start-event-generic, and this is the CLI-unavailable hand-authoring contract, so an agent following it emits an entry for a timer start. Line 128 of the same file already says *"root manual start event"*, so the file contradicts itself.

**Fix**: make SKILL.md:229 say *root **manual** start*, and scope lmrg's "Entry Point Rules" to manual root starts. The new block in `structural-bpmn.md` can then shrink to the direct-child mechanism, which is the genuinely new part.

🟠 **major — the PR body's "Problem" quote is not in the base.**

The body attributes to the reference: *"Every root-level `bpmn:startEvent` is an entry point, whether it is manual, timer-triggered, or connector-triggered"*, and says it *"told agents to remove the manual start when adding a timer one."*

Neither string is in `origin/codex/bpmn-metadata-refresh` — `git grep` on the base ref returns no hits, and `grep -rn "manual start" skills/uipath-maestro-bpmn/` on the base finds nothing outside `local-metadata-regeneration-guide.md`. The only artifact teaching remove-the-manual-start is `timer_start.yaml:35`, which this PR keeps and reinforces. Rewrite the Problem section against the two real sites above.

🟡 **minor — no passing-run claim.**

The advisory task lint asked twice for `Ran skill-bpmn-timer-start locally and it passed`; it was never added. Given the task actually fails in CI, this is not cosmetic — the missing claim is exactly what would have caught it.

### What I could not verify

`entry-points.ts` and `project-validator.ts` are not in this repo, so the direct-child rule, the manual-only derivation, and *"`validate` stays clean"* for a timer-only process all rest on the author's reading of the CLI source. `validator/bpmn-spec.json` carries no rule that would confirm the validate-stays-clean claim. Citing `.ts` paths is established convention here (`structural-bpmn.md:30`, `:510`, `cli-conventions.md:23`), so the style is fine — the facts are unconfirmed from this repo. Someone with the CLI source should confirm the `eventDefinition` **and** `uipath:event` exclusion before this lands as doctrine.

I read the checker rather than running it, and did not run `scripts/check-cli-verbs.py` or `check-skill-links.mjs`; both gates pass in CI.

### Second CI failure

`skill-bpmn-operate-diagnose-minimal-fault-triage` also failed on this head — `llm_judge` 0.35 < 0.70, weighted 0.87 — and passed on base. Probably judge variance on a diagnose task rather than a docs regression, but I did not confirm either way. Both need to be green.

## Recommendation

❌ **Request changes.** A required gate is failing because of the diff, and the claim the PR sets out to fix is left standing in the always-loaded SKILL.md.

## tl;dr

CI is red because of this PR's own new check: it grades a `uipath:entryPointId` on a timer start that the task never asks for, that the registry template doesn't provide, and that the same PR calls inert. Smoke drops 100% → 83.3%. Separately, the claim the PR says it's fixing is still in SKILL.md:229 and `local-metadata-regeneration-guide.md:231` — the two files that actually state it — and the PR body quotes text that isn't in the base. The `Intsvc.TimerTrigger` portability half is correct and verified.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
DevMomo: Took option 1 and dropped it — grading a field this PR documents as inert was indefensible. Worth recording that this is the timer-start smoke failure I had been reporting to the author as gate flakiness across several runs. It was not flaky; it was unsatisfiable, and I never opened the failure to check.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
DevMomo: Moot now that the assertion is gone, and you are right that it was backwards — the one check that could have enforced direct-child placement used a descendant search. If a task later grades placement, your snippet is the shape it should take.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
DevMomo: Kept this one as the deliberate source-only negative and did not add the sibling task in this PR. The coverage hole you describe is real and I am not claiming otherwise — but adding a task that has never been run against a live agent is precisely how the unsatisfiable assertion above got here, so I would rather it land with a run behind it than ride along unexercised.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/check_timer_start.py
DevMomo: Both gone with the assertion.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Moved to the entry-point contract at :178, so the event matrix is contiguous with its payload section again. That also collapses the third statement of the same rule.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Reworded to "report the node as **draft**", named the CLI-owned blocker requirement with the literal `connection binding` phrase, and inlined the escaped placeholders SKILL.md defines so this file reads correctly on its own.

skills/uipath-maestro-bpmn/references/registry-workflow.md
DevMomo: Turned into the instruction: leave the opaque type in place, do not pre-empt enrichment.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/timer_start.yaml
DevMomo: Reworded — it now says retrieve the template and author from it, without telling the agent to skip the discovery step SKILL.md step 1 mandates.

tests/tasks/uipath-maestro-bpmn/single_node/timer_start/timer_start.yaml
DevMomo: Re-flowed.

rockymadden (approved):

PR #3160, opened 2026-09-09
fix(ci): merge the catalog refresh with --admin, not by racing gh

uipreliga:
## Problem

The nightly catalog refresh stopped auto-merging — [#3159](https://github.com/UiPath/skills/pull/3159) sat green and unmerged, and [run 34310432381](https://github.com/UiPath/skills/actions/runs/34310432381) shows why:

```
Required status checks (7): [...]
::group::PR #3159
All checks green — squashing.
X Pull request UiPath/skills#3159 is not mergeable: the base branch policy prohibits the merge.
```

`gh pr merge --squash` refuses **client-side** whenever `mergeStateStatus == BLOCKED` — it never reaches the merge API. That's what the "add `--auto` / add `--admin`" hint in the error is.

And these PRs are `BLOCKED` permanently: both `main` rulesets (`Require PR` #13681075, `main` #14795269) carry a `pull_request` rule with `required_approving_review_count: 1` and `require_code_owner_review: true`, unchanged since April. A bot PR never gets a review, so no amount of green checks clears that state.

## Why it ever worked

By race. A push to `main` invalidates open PRs' mergeability to `UNKNOWN`, which slips past gh's client-side gate; the merge API then completes it, because `GH_PAT`'s owner is a repo admin and therefore a bypass actor on both rulesets. The correlation is exact:

| refresh PR | preceding push to `main` | merged |
|---|---|---|
| #3130 | 12:10:40 | 12:12:18 — after refusals at 05:39, 05:47, 06:01, 07:42, 09:08, 09:16, 11:33 |
| #3111 | 07:29:39 | 07:30:46 — after 10 consecutive refusals |

On #3111 two runs fired off the same push: one merged, its twin was refused. That's ~3.9k failed runs of this workflow to date. #3159 simply lost the race for good — `main` went quiet at 03:20 UTC, its checks turned green at 04:18, and no further `check_suite` event arrived to retry.

## Fix

Use that bypass deliberately instead of by accident:

```diff
-gh pr merge "$PR" --squash --delete-branch
+gh pr merge "$PR" --squash --delete-branch --admin
```

`--auto` is not the alternative: repo auto-merge is off by design (see the workflow header), and enabling it would only queue behind the approval that never comes.

## Safety

`--admin` skips gh's *local* refusal, not this job's verification. Untouched above the merge: the branch-name filter, the dynamically-fetched required-check list, the fail-closed guard on an empty required set, and the present-and-green count over every required context. Those remain the real boundary — a refresh with red or missing required checks still stays open for a human sweep.

### DIFF
--- a/.github/workflows/automerge-catalog-refresh.yml
+++ b/.github/workflows/automerge-catalog-refresh.yml
@@ -138,6 +138,18 @@ jobs:
             fi
 
             echo "All checks green — squashing."
-            gh pr merge "$PR" --squash --delete-branch
+            # `--admin` merges through this PAT's ruleset bypass (Repository
+            # admin is a bypass actor on both `main` rulesets). Without it gh
+            # refuses client-side, before it ever calls the merge API: `main`
+            # requires an approving review, so an unreviewed bot PR sits at
+            # mergeStateStatus BLOCKED forever and gh reports "the base branch
+            # policy prohibits the merge". The merges that did land only won a
+            # race — a push to `main` resets an open PR to mergeability
+            # UNKNOWN, which slips past that client-side gate.
+            # NOT `--auto`: repo auto-merge is off by design (see header), and
+            # enabling it would only queue behind the approval that never comes.
+            # The required-check gate above remains the real safety boundary —
+            # `--admin` skips gh's local refusal, not this job's verification.
+            gh pr merge "$PR" --squash --delete-branch --admin
             echo "::endgroup::"
           done


bai-uipath (approved):
**Approve** — diagnosis holds up. The run logs show a workflow merge succeeding with plain `gh pr merge` inside a push window, so the bypass is real and only gh's client-side gate was blocking the rest. `--admin` fails closed if the token ever loses admin.

- **Please don't keep this on your personal account.** This makes your PAT's admin bypass the documented merge path, so the nightly now hangs off one person's credential: it dies on rotation or offboarding, the audit trail reads as you hand-merging every refresh, and the bypass is scoped to "whoever holds repo admin" rather than to this job. Not blocking, but worth doing soon.
- **`coder_eval` already has the pattern to copy.** Its release workflow mints a scoped GitHub App token (`coder-eval-release-bot`) specifically for the main-branch ruleset bypass, with `permission-contents: write` and nothing else. Here it'd need contents + pull-requests + issues (the label filter reads labels). Requires the App installed on this repo plus the two secrets, so a follow-up rather than this PR.
- **Related, and the sharper problem:** `GH_PACKAGES_TOKEN` exists in neither this repo's secrets nor the org's, so every `GH_PACKAGES_TOKEN || GH_PAT` fallback resolves to `GH_PAT` today — including in the two smoke workflows that trigger on `pull_request` in a public repo. Same PAT, much wider reach than the in-file comment claims.
- **Minor, optional:** the new comment restates the commit message; four lines would carry it. And `--match-head-commit` would make check-then-merge atomic now that GitHub's own enforcement is out of the path.

uipreliga:
Thanks — took both of the minor ones in 08f64ad29.

`--match-head-commit` is the better catch of the two. `headRefOid` now comes back in the *same* `gh pr view` call as `statusCheckRollup`, so the SHA and the checks can't describe different heads, and that SHA is what the merge is pinned to. Worth doing precisely because of this PR: with GitHub's enforcement out of the path, that check-then-merge window was the only thing left holding the gate. The `--admin` comment is down to six lines.

On the PAT — keeping it for now, deliberately. The scoped App token is the right end state and I'm not arguing the point; I just don't want to couple a credential swap to the fix that stops the nightly hanging. Tracking it separately, and `coder-eval-release-bot` is the model to copy (here it needs contents + pull-requests + issues, since the `--label automated` filter reads labels).

On `GH_PACKAGES_TOKEN` — confirmed, and you're right that it's the sharper problem. It exists in neither this repo's secrets nor the org's, so all 9 `GH_PACKAGES_TOKEN || GH_PAT` sites across `run-coder-eval.yml`, `smoke-skills.yml`, `smoke-rpa-skills.yml` and `refresh-uip-catalog.yml` resolve to `GH_PAT` today, and the "never the repo-write PAT" comments next to them are wrong as written.

One thing that narrows it: those workflows trigger on `pull_request`, not `pull_request_target`, so fork PRs receive no secrets at all — the reach is collaborator-pushed branches, not the public. Still shouldn't stand: either the secret gets created or the fallback and its comments go. Separate issue, not this PR.

PR #3162, opened 2026-09-09
fix(maestro-case): state that an output row goes after the one it reads

Yiqi517:
## The problem

The output dispatch walks the SDD's rows and then auto-mints the schema entries the SDD does not reference. A `=` row that reads a slot through `$xref` reads one of those auto-mints, so following the dispatch literally emits the computed row first and the slot it reads second.

That order does not work. The engine evaluates `data.outputs` in array order and writes each result into the variable scope before evaluating the next, so a row reads only what precedes it. Its own comment at [`BpmnDataManager.cs:1067`](https://github.com/UiPath/PO.BpmnEngine/blob/01975d35331e47df5de3ad875d14cee33ef2f951/src/BpmnEngine/UiPath.PO.BpmnEngine/BpmnDataManager.cs#L1067) says why: `Outputs can reference other outputs and, thus, we need to set after each evaluation.` The canvas never hits it, because it appends custom outputs last at [`IntsvcActivityProperties.tsx:609`](https://github.com/UiPath/PO.Frontend/blob/ac3b97f0b29a04beb2538905a1b2ab6f236028d9/src/components/canvases/shared/panels/properties/fps/IntsvcActivityProperties.tsx#L609) and [`:693`](https://github.com/UiPath/PO.Frontend/blob/ac3b97f0b29a04beb2538905a1b2ab6f236028d9/src/components/canvases/shared/panels/properties/fps/IntsvcActivityProperties.tsx#L693). A build writing the caseplan by hand had nothing telling it the same thing.

## Which run

[33813506820](https://github.com/UiPath/skills/actions/runs/33813506820), claude-sonnet-5, 619 turns, scored 0.744. Its SDD gives that task two output rows and never names the slot the computed row reads:

    | Comment | -> buyerComments |
    | —       | buyerDecision = =js:vars.$xref('Buyer review','Record buyer review decision','Action') |

So `Action` was auto-minted after the computed row, and the instance recorded `action4 = reject` with `buyerDecision = NULL` in the same variable dump. The buyer's decision never reached the route guards that read it, and three execution criteria failed while every build-shape criterion passed.

**The text was in context and had nothing to offer.** The agent read this file in full at turn 212, wrote the task at turn 244, and at turn 247 went back to that exact row and corrected it against the canonical shape field by field, changing its `name` and dropping a spurious `options`. Its own words at that turn: `I made an error in the "Record buyer review decision" custom output shape - let me fix the Scenario E entry to match the canonical shape exactly.` It fixed every field the shape lists. The shape listed no position.

## What changed

[`io-binding/impl-json.md`](https://github.com/UiPath/skills/blob/ea86e02a8853b3eb73bb47dedb2e34c54d8d4237/skills/uipath-maestro-case/references/plugins/variables/io-binding/impl-json.md#L62-L72) states, where the three shapes are defined, that emission order is not the SDD's row order, names the auto-mint case as the common one, and shows the two rows in the order the canvas produces them.

The operation is order-dependent and invisible to `validate`, which is the condition the authoring guidance gives for an exact shape rather than prose.

## Verification

**The mechanism, without a run.** Take the failing artifact and change one thing: move each task's `custom: true` rows to the end of its array, which is all the canvas does. Seven rows move, the row count stays at 126, and the two forward reads go to zero.

**The same task, with the paragraph.** [34379400360](https://github.com/UiPath/skills/actions/runs/34379400360) runs `supplier_onboarding` on a branch carrying the supplier test suite plus this file as it stands here. claude-sonnet-5, 623 turns, 351 commands, scored 0.9408, and 21 of its 23 criteria pass:

| | 33813506820, no paragraph | 34379400360, with it |
|---|---|---|
| tasks carrying outputs | 31 of 32 | 32 of 33 |
| output rows | 126 | 61 |
| computed rows | 2 | 2 |
| position of each computed row | index 1, producer at 2 | index 2, producer at 1 |
| rows reading a later sibling | 2 | 0 |

The SDD declares exactly two computed rows, and both come out in the working order:

    t08oP2yAb 'Record buyer review decision'
      [1] var=action2       source==Action
      [2] var=buyerDecision value==js:vars.action2
    t17gH4iPq 'Record compliance review decision'
      [1] var=action4           source==Action
      [2] var=complianceDecision value==js:vars.action4

The agent reached the text: at turn 161 it ran `cat … | head -250` on this file, and the paragraph sits at line 62.

**The backstop.** [cli#4078](https://github.com/UiPath/cli/pull/4078) reports the violation as `STRICT_OUTPUT_FORWARD_READ` from the caseplan alone, so a build that does not follow this paragraph is still caught before it runs. On 289 eval caseplans that code reports 15 objects across 7 runs.

`npm run skills:validate` passes on the tested content: `default: 27 skills, 1764 files, 0 replacements; studioweb: 27 skills, 1764 files, 264 replacements`. The file carries no flavor markers and `skill-flavors/studioweb/` has no override at this path, so the flavor inherits the edit unchanged.


skills/uipath-maestro-case/references/plugins/variables/io-binding/impl-json.md
```diff
@@ -59,6 +59,17 @@ A `$xref('Stage','Task','output')` marker is not one of those two reference form
...
 
+**Emission order is not the SDD's row order. A row whose expression reads another output's `var` on the same task goes AFTER that output.** The engine evaluates `data.outputs` in array order and writes each result into the variable scope before evaluating the next, so a row placed before the one it reads gets the case variable's default on every run of the case while `validate` reports `Valid`. The common shape is exactly the one the dispatch above produces backwards: the SDD declares the `=` row and says nothing about the schema entry it reads through `$xref`, so that entry is auto-minted and lands after it. Emit the auto-mint first and the `=` row last, whatever order the SDD's table used. `--strict` reports the violation as `STRICT_OUTPUT_FORWARD_READ`.
+
+```json
+[{ "name": "Action", "type": "string",
+   "id": "action4", "var": "action4", "value": "action4",
+   "source": "=Action", "target": "=action4", "elementId": "Stage_review-tBuyer01" },
+ { "name": "buyerDecision", "type": "string", "custom": true,
+   "var": "buyerDecision", "value": "=js:vars.action4", "source": "=js:vars.action4",
+   "target": "", "body": "", "elementId": "root" }]
```
charlesliu9: nit: the format of the example json is neither pretty-printed or compact

Yiqi517: Changed: pretty-printed in ef39f64fc3936eaba627b6f7970c7e27b5871a03. You're right it matched neither convention — its continuation lines didn't align with each other or with either object's own brace. This repo has no single JSON style across the case skill (58 blocks semi-compact, 53 pretty, 15 single-line), so I picked indent=2 rather than semi-compact: it's the one a tool can produce and check, so a reader copying it never has to guess where a line breaks.

roberts-cliff (approved):

StephenHanzlik (approved):

PR #3163, opened 2026-09-09
test(uipath-maestro-flow): make the ixp handoff e2es safe to run concurrently [RE-13543]

joe-prosser:
`skill-flow-ixp-e2e-project-creation-handoff` couldn't run twice at once. Its precondition — that no published extractor covers the fixture domain — is an *absence* in shared tenant state, so it can't be allocated per run, and a passing run filled it by publishing a falconry extractor of its own. On 2026-09-03 two runs launched 30s apart collided: the codex run ERRORed at 0.0 with the agent never invoked, on a task that had scored 1.0 on the same agent and model the day before.

The fix splits the task along the read/write line. `e2e_03` keeps its unprompted prompt — noticing the prerequisite is the measurement — and `stop_early` cuts the run the moment the `uipath-ixp` Skill is invoked, before `projects create`, so it only *reads* the absence and any number of runs can share it. A new `e2e_04_build_mechanics.yaml` grades what happens after the handoff; its prompt names the extractor as missing, so it needs no absence and can write freely. Both share one document set because e2e_04 names its extractor from `seed.json`, off `BUILD_PROJECT_PREFIX` rather than the document domain, so what it publishes never reads as coverage to an overlapping e2e_03.

Also hardens the shared grader for parallel dispatch, since its project set comes from a tenant-wide diff and therefore names siblings' projects: teardown could abort before deleting its run folder when a sibling deleted a project mid-cleanup, stranding the deployment whose registry node burns the fixture domain, and attribution could delete a live sibling's project outright.

Validated on the eval tenant across 11 dispatched runs — 13/13 on e2e_04, 9/9 on e2e_03, no gate failure and no ERROR. Overlapping runs were observed declining to delete each other's projects and absorbing 404 and 429 races.

RE-13543


richsilv (approved):
Yeah, looks good. There is a surprising amount of code, but the rationale and strategy are good.

PR #3166, opened 2026-09-09
test(uipath-ixp): seed the project so name resolution is actually tested

horeaprecup-uipath:
## What it tests

`pre_run` creates a project with a run-unique title and tells the agent **only that title**. The agent has to find it (`projects list` → ProjectName), delete `york_solutions_invoice.png` from it (`documents list` → DocumentId), and touch nothing else on the shared tenant.

## Why it changed

[Alex was right](https://github.com/UiPath/skills/pull/3166#discussion_r3971456133): the old test had the agent *create* the project, so `projects create` returned the ProjectName directly. There was nothing to resolve, and the `projects list` check could never pass — both agents were docked for correctly skipping it. Seeding the project outside the agent makes the lookup unavoidable.

## Flow

```
pre_run   seed_project.py     creates the project, writes seed.json
agent     does the work       every uip command lands in mocks/calls.log
grading   6 checks            4 read the log, 2 ask the tenant
post_run  cleanup_project.py  deletes the project
```

## Files

| File | |
|---|---|
| `name_resolution.yaml` | the prompt and the 6 graded checks |
| `_shared/seed_project.py` | **pre_run** — creates the project from the 3 fixture invoices, imports the taxonomy, verifies the uploads landed, writes `seed.json` |
| `integration/check_documents_after.py` | asks the tenant: is york gone, are the other two still there? |
| `integration/check_own_project.py` | reads the log: did any write target another project? |
| `_shared/cleanup_project.py` | **post_run** — deletes the project |

## seed.json

```json
{"uuid8": "d4f8d6cf", "project": {"title": "codereval-name-resolution-d4f8d6cf"}}
```

- `title` — all the agent gets.
- `uuid8` — random per run, so concurrent runs never collide. The backend copies it into the ProjectName, which is how cleanup and the checkers find the project afterwards; unlike the title, it survives a rename.
- **No ProjectName** — that's the answer the agent is meant to look up. Anything that needs it later reads it from `projects list`, same as the agent.

`report.json` is gone from this test — cleanup reads `seed.json`, so it works even if the agent crashes early. The other 5 IXP tasks still use it and are unaffected.

## Graded (weight 13.0)

| Check | |
|---|---|
| Ran `projects list` | 2.5 |
| Ran `documents list` | 2.0 |
| Ran `documents delete … -y` | 2.0 |
| Did **not** pass the filename in where a DocumentId goes | 1.5 |
| **Tenant:** york gone, other two still there | 3.0 |
| **Tenant:** never wrote to another project | 2.0 |

The last two ask the tenant instead of trusting a file the agent wrote: the log records commands *before* they run, so a `delete` that 404'd looks exactly like one that worked.

## Shared-tenant safety

The 09-07 run overwrote an unrelated project's title and extraction prompt, and left 6 scratch projects behind. `check_own_project.py` knows all 29 `uip ixp` write verbs (cross-checked against `assets/uip-catalog-snapshot.json`) and fails the run if any of them targets a project that isn't this one — reading other projects stays allowed, since that is how `projects list` works. And a seed the tenant rejects now ends the run as an **environment error** rather than a skill failure, which is what the 09-07 tenant 403 should have been.

## Verified

Replayed against the archived 09-07/09-08 runs plus synthetic cases:

| Case | |
|---|---|
| 09-07 run that damaged `vendor-invoices-…` | **FAIL** — both foreign writes caught |
| Foreign `configure-model` / `labellings confirm` / `data-types delete` | **FAIL** |
| Agent creates a scratch project | **FAIL** |
| Tenant: york still there, or too much deleted | **FAIL** |
| Reading another project, or `--help` | PASS — allowed |
| Clean run | PASS |
| Seed: partial upload, or a ProjectName cleanup couldn't find | exits non-zero |
| Cleanup: seed only / report only / both / neither | correct in all four |

`py_compile` clean · YAML parses (6 criteria, 13.0) · `check-cli-verbs.py` OK · write-verb table cross-checks clean against the CLI catalog.

## Follow-ups, not here

- The other 5 IXP tasks have no ownership check, only a line in the prompt asking nicely.
- `./fixtures/` doesn't exist in the sandbox (`template_dir` flattens it to the root). This task no longer needs the fixtures; 5 others still do.

### DIFF
--- a/tests/tasks/uipath-ixp/integration/name_resolution.yaml
+++ b/tests/tasks/uipath-ixp/integration/name_resolution.yaml
@@ -2,8 +2,8 @@ task_id: skill-ixp-integration-name-resolution
 description: >
   Integration: name/title resolution on a live tenant. The agent must resolve
   user-facing references to the ids the CLI needs — a document by FILENAME →
-  DocumentId via `documents list`, and a project by its TITLE → Name via
-  `projects list` — rather than passing the human label straight through. The
+  DocumentId via `documents list`, and a project by the Name returned by
+  `projects create` — rather than passing the human label straight through. The
   resolution step is the integration part; smokes pass the id in the prompt.
 tags: [uipath-ixp, integration, mode:operate, lifecycle:setup]
 
@@ -39,6 +39,9 @@ initial_prompt: |
 
   Right after the project is created, write `report.json` for cleanup: {"project_name": "<the ProjectName from the create response>"}.
 
+  Work only on the project you create. The tenant is shared — never write to any
+  other project. If a command fails, report the error; do not probe other projects.
+
 success_criteria:
   - type: file_matches_regex
     description: "Listed documents to resolve the filename → DocumentId, then re-read to save the remaining list"
@@ -57,14 +60,6 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: file_matches_regex
-    description: "Listed projects to resolve Title → Name"
-    path: "mocks/calls.log"
-    pattern: >-
-      (?m)^(?:uip\s+ixp\s+projects\s+list)[^\r\n]*$
-    weight: 1.5
-    pass_threshold: 1.0
-
-  - type: file_matches_regex
     description: "Renamed the project via update-title"
     path: "mocks/calls.log"
     pattern: >-
@@ -79,6 +74,18 @@ success_criteria:
       (?m)^(?:uip\s+ixp\s+documents\s+delete\s+\S+\s+york_solutions_invoice\.png)[^\r\n]*$
     must_match: false
     weight: 1.5
+    pass_threshold: 1.0
+
+  # The tenant is shared and every concurrent run titles its project
+  # `codereval-integ-resolve-*`, so the prefix cannot identify ownership — the
+  # random suffix can. check_own_project.py derives this run's suffix from its
+  # own `projects create` and requires every project-scoped write to carry it.
+  - type: run_command
+    description: "Every project-scoped write stayed inside this run's own project"
+    command: "python3 $TASK_DIR/check_own_project.py"
+    timeout: 30
+    expected_exit_code: 0
+    weight: 2.0
     pass_threshold: 1.0
 
   - type: file_exists
--- a/tests/tasks/uipath-ixp/integration/name_resolution.yaml
+++ b/tests/tasks/uipath-ixp/integration/name_resolution.yaml
@@ -1,15 +1,15 @@
 task_id: skill-ixp-integration-name-resolution
 description: >
-  Integration: name/title resolution on a live tenant. The agent must resolve
-  user-facing references to the ids the CLI needs — a document by FILENAME →
-  DocumentId via `documents list`, and a project by the Name returned by
-  `projects create` — rather than passing the human label straight through. The
-  resolution step is the integration part; smokes pass the id in the prompt.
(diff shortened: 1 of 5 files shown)


tests/tasks/uipath-ixp/integration/name_resolution.yaml
```diff
@@ -2,8 +2,8 @@ task_id: skill-ixp-integration-name-resolution
 description: >
   Integration: name/title resolution on a live tenant. The agent must resolve
   user-facing references to the ids the CLI needs — a document by FILENAME →
-  DocumentId via `documents list`, and a project by its TITLE → Name via
-  `projects list` — rather than passing the human label straight through. The
+  DocumentId via `documents list`, and a project by the Name returned by
```
alexandrujircan: By doing this I think we change the scope of the test 🤔 The file name is `name_resolution`, so I think it was intended to actually do some mapping between human-provided names and IDs. I think this description is ok, but the initial prompt given to the agent is not. I don't think we should set up a new project, but rather provide an already existing project and check that the agent can do name resolution on it, find the corresponding project ID, and use that for downstream calls.

horeaprecup-uipath: Done

tests/tasks/uipath-ixp/integration/name_resolution.yaml
```diff
@@ -21,55 +21,47 @@ sandbox:
   template_sources:
     - type: template_dir
       path: ../_shared/live_calls_template
-    - type: template_dir
-      path: ../_shared/fixtures
+
+# Creates the project the agent must find; writes seed.json with its title only.
+# fail_on_error defaults on: a rejected seed ends the run as an environment error.
+pre_run:
+  - command: 'IXP_SEED_TITLE_BASE="codereval-name-resolution" python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-ixp/_shared/seed_project.py"'
```
alexandrujircan: Can we set a title with spaces? I think that would be closer to what a real project would have, like Codereval Name Resolution or smth

tests/tasks/uipath-ixp/integration/name_resolution.yaml
```diff
@@ -21,55 +21,47 @@ sandbox:
...
+    weight: 2.5
     pass_threshold: 1.0
 
   - type: file_matches_regex
-    description: "Deleted the named document by resolved id (with -y)"
+    description: "Listed documents to resolve the filename → DocumentId"
     path: "mocks/calls.log"
     pattern: >-
-      (?m)^(?:uip\s+ixp\s+documents\s+delete\b.*(-y|--yes))[^\r\n]*$
+      (?m)^(?:uip\s+ixp\s+documents\s+list)[^\r\n]*$
```
alexandrujircan, line 55: Also assert that documents list is called with project-name, so we make sure name resolution worked

horeaprecup-uipath: I think it is asserted that every command is called with project-name in the  check_own_project.py. If we ever find a command with another uuid for project name, it fails

tests/tasks/uipath-ixp/integration/name_resolution.yaml
horeaprecup-uipath: Done — the seeded title is now `Codereval Name Resolution <run id>`.

alexandrujircan (approved):

PR #3169, opened 2026-09-09
fix(tests): preflight the datafabric connector suite, repin outlook to its mailbox

rockymadden:
Follow-up to #3168. Same failure mode, same connection.

## What

- Adds the `_shared/preflight_connections.py` gate to **10** `connector_features/datafabric_connector` tasks, selector `uipath-uipath-dataservice`.
- Repins `single_node/outlook_trigger_inbox` from a bare selector to `uipath-microsoft-outlook365=uipath-maestro-flow`.

11 files, 21 insertions, 1 deletion. No new script, no criterion changes, no prompt changes.

## Why the datafabric suite

All ten author or grade a `uipath-uipath-dataservice` node against connection `d61e5d0e`, which was `State: Failed` on 2026-09-09 and took three e2e tasks down with it in #3168.

Only four of the ten run `flow debug`. The other six grade a compiled `queryExpression` or a configured node, and `node configure` reaches the connector over the wire to compile the filter and cache the schema. A dead connection fails those offline structural checks too. **"Grades a live run" is the wrong filter for who needs this gate**, which is worth stating plainly since it is the thing I got wrong first time.

Ordering: the preflight goes **first** in `pre_run`, ahead of `ensure_entity.py` and the record seeders. Those go through the Data Fabric API and succeed while the IS connection is down, so running them first only burns setup before the real blocker surfaces.

### Excluded

`smoke_activation_positive` and `smoke_activation_negative`. Both grade `skill_triggered` advisories at `pass_threshold: 0.0` and never touch a connection. They are skill-routing tests, not connector tests.

## Why the outlook repin

`outlook_trigger_inbox` has carried a bare selector since it was added, and the bare form picks the wrong mailbox:

```
$ preflight_connections.py uipath-microsoft-outlook365
OK: … — Productivity O365 Outlook (5af6fa90…) live

$ preflight_connections.py uipath-microsoft-outlook365=uipath-maestro-flow
OK: … — is-sandboxes-test@uipathsandboxes.onmicrosoft.com (75dfafb4…) live in folder 'uipath-maestro-flow'
```

The tenant has three Enabled Outlook connections. `Productivity O365 Outlook` in `Shared` is a different mailbox from the sandbox one the task's Inbox fixture lives in, so the bare gate passes while the task stays unwinnable. That is exactly the trap the `=<folder>` form was added for, and this task's own header already records the receiving mailbox as unreliable.

This is the general shape of it: **bare is only safe when the connector has exactly one connection tenant-wide.** True for dataservice, jira, and testmanager. Not true for slack, outlook365, or google-drive, which have two or three across folders with a fixture-less default.

## Verification

- All 13 touched or adjacent YAMLs parse; every patched file has the preflight as `pre_run[0]`, and the two excluded ones still have no `pre_run`.
- Both selectors exit 0 against the live tenant, with the contrast above showing the repin resolves a different connection than the bare form it replaces.
- `pre_run` runs outside the `task_timeout` watchdog (`test_criterion_budgets.py:546`), so criterion budgets are unaffected.

## Still open

- The script lives in `tests/tasks/uipath-maestro-flow/_shared/`. Extending to `uipath-platform`, `uipath-api-workflow`, `uipath-maestro-case` or `uipath-maestro-bpmn` means those suites reaching into maestro-flow's private dir. Wants a `tests/tasks/_shared` first.
- It only asserts "at least one live connection". The `bindings/` family needs more: `reconfigure_different_connection` needs two connections of one key, `multi_connector_independence` needs two different keys.
- Remaining candidates, deliberately not in this PR: slack (6, all folder-pinned), outlook (3 more), jira (~8), testmanager (~14 across two suites), plus one-offs for sfdc, azureactivedirectory, gmail and jdbc.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/contractregistry_crud_filters.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/contractregistry_crud_filters.yaml
@@ -11,6 +11,8 @@ run_limits:
   task_timeout: 3600
 
 pre_run:
+  - command: 'python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/preflight_connections.py" uipath-uipath-dataservice'
+    timeout: 120
   - command: 'python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/ensure_entity.py" "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/contract_registry.entity.json"'
     timeout: 60
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
@@ -16,6 +16,8 @@ run_limits:
   task_timeout: 2400
 
 pre_run:
+  - command: 'python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/preflight_connections.py" uipath-uipath-dataservice'
+    timeout: 120
   - command: 'python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/ensure_entity.py" "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/contract_registry.entity.json"'
     timeout: 60
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/integration_create_get.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/integration_create_get.yaml
@@ -14,6 +14,8 @@ run_limits:
   turn_timeout: 1200
 
 pre_run:
+  - command: 'python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/preflight_connections.py" uipath-uipath-dataservice'
+    timeout: 120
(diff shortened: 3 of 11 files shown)


rockymadden:
## `Run skill smoke tests` is red, and it is not this PR

Pass rate 16.7% (1/6). Five `datafabric_connector` smoke tasks fail. Evidence that the preflight is not the cause, and that these are pre-existing:

**1. The preflight passed on every task.** Six occurrences of `OK: uipath-uipath-dataservice — aditi.goyal@uipath.com (d61e5d0e…)` in the run log, zero occurrences of `TENANT NOT READY`.

**2. The statuses are `FAILURE`, not `ERROR`.** A `pre_run` failure lands `FinalStatus.ERROR` by design. `FAILURE` means `pre_run` completed and the agent ran.

**3. This PR made CI *select* them, not fail them.** The run log shows `Will test (task changed): tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_error.yaml` and siblings. The last PR to touch this directory, #3139, changed only the two activation yamls, so its smoke ran `Pass rate: 100.0% (2/2)` on those alone. These five had not been selected by any recent PR.

**4. The failures are authoring failures, unrelated to connections:**

```
FAIL: no create-entity-record node in any .flow
FAIL: no update-entity-record node found
FAIL: expected >=2 query nodes sorted by score, found 0
FAIL: no .flow has all 4 ops on FlowCodeEvalEntity
FAIL: no .flow satisfies the error-path shape
FAIL: … no create-entity-record on FlowCodeEvalEntity with inputs.detail.pathParameters.entityName='FlowCodeEvalEntity'
```

**5. The one that passes is the one where the node is pre-built.** `smoke_update_existing_flow` scores 1.000, and it is the brownfield task whose `scaffold_movie_report_flow.py` creates the Data Service node in `pre_run`. Every task that asks the agent to author a `uipath-uipath-dataservice` node from scratch failed.

Runs were 180-373s with 821-2458 char agent responses and `max_turns_exhausted` unset, so the agent finished early rather than running out of budget.

### So there are two separable things here

- **This PR's gate works.** It confirmed a healthy connection and handed off, which is exactly the contract.
- **Five Data Service authoring tasks are red** and have been invisible because nothing touched their yamls. That is a skill problem, not a fixture problem, and it wants its own investigation.

Happy to split this PR down to the four non-smoke-tagged yamls to get a green run, but that leaves the highest-exposure tasks ungated and re-hides the red. Preference from the maestro-cli-team on which way to go?

nikhil-maryala (approved):

PR #3173, opened 2026-09-09
fix(hitl-skill): replace stale nested-root caseplan.json test seeds with a real fixture

dushyant-uipath:
## Summary
- quality_12_case_path_selection and smoke_09_case_quickform have been timing out for about two weeks in the nightly coder-eval run. Root cause is not a skill regression: both test YAMLs hand-author a caseplan.json seed inline in initial_prompt using an old nested {"root": {...}} shape.
- The current uip maestro case validate schema is flat (no root wrapper). This matches the real CLI, uipath-maestro-case's own edit/templates/LinearThreeStages fixture, and its impl-json.md docs. The HITL skill's own reference (hitl-casetask-action.md:239) already documents the flat schema correctly, so the skill was never wrong, only these two test fixtures were stale.
- Every affected run burned its entire turn budget rediscovering the correct schema via trial and error against the CLI, then timed out before ever reaching the HITL-task-insertion behavior the test is meant to grade.

## Fix
- Add tests/tasks/uipath-human-in-the-loop/templates/case_single_stage_no_app/caseplan.json, a fixture validated against the real CLI (bare case, one stage, no HITL task yet, two case variables). Built the same way uipath-maestro-case's own edit-test fixtures are built (uip maestro case init plus direct JSON edits, since the stage/condition CLI mutation commands are deprecated).
- Point both tests at it via sandbox.template_sources, the same mechanism uipath-maestro-case's edit/* tests use, instead of inline JSON.
- No success_criteria changed. smoke_09 anchors its grading to specific variable ids (var_invoiceId/var_amount), which the new fixture preserves exactly.

## Test plan
- [x] uip maestro case validate on the new fixture: Result Success (one expected warning, stage has no tasks yet, which is correct since that is the agent's job)
- [x] Hand-built the expected end state (fixture plus a full QuickForm action task per hitl-casetask-action.md's recipe) and re-ran uip maestro case validate: Result Success, zero warnings
- [x] Ran every smoke_09_case_quickform success_criteria grep against that hand-built end state: all pass
- [ ] Full coder-eval run against both tasks, queued for tonight's nightly

### DIFF
--- a/tests/tasks/uipath-human-in-the-loop/quality_12_case_path_selection.yaml
+++ b/tests/tasks/uipath-human-in-the-loop/quality_12_case_path_selection.yaml
@@ -12,6 +12,9 @@ agent:
 
 sandbox:
   python: {}
+  template_sources:
+    - type: template_dir
+      path: templates/case_single_stage_no_app
 
 initial_prompt: |
   I have a UiPath case management project. Our agent extracts vendor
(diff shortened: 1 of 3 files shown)


chetanyauipath (approved):
LGTM

PR #3176, opened 2026-09-09
fix(uipath-maestro-flow): a gateway is never readable from downstream

rockymadden:
## What

`decision/impl.md` told downstream nodes to read which branch ran as `$vars.<decisionId>.matchedCaseId`, and `brownfield.md` repeated it for the node the branches rejoin at. Neither works.

A Decision or Switch **never enters a downstream node's expression scope**. Every form reads `undefined`, the flow still validates, and the reader silently takes the wrong path instead of faulting.

`EXPRESSION_DIAGNOSTIC` — the warning that names this defect exactly — appeared nowhere in the skill, so it was unresolvable from the docs.

## Why

Two 2026-09-09 eval runs, one root cause:

| task | status |
|---|---|
| `skill-flow-move-node` | ERROR, 0.0 |
| `skill-flow-interactive-customer-escalation-triage` | FAILURE, 0.448 |

- **move_node** is the documented case verbatim: merge both branches into `formatSummary`, pick a message from the branch taken. It wrote what the skill prescribes, hit the warning at command 29 of 88, found nothing about it in the skill, and spent the remaining ~950s of its 1200s budget reading the minified `maestro-tool` bundle for an answer.
- **escalation_triage** read `$vars.determineSeverity.matchedCaseId` at its End node, validated with the same warning four times, shipped, and the grader's run returned `Sev2` for a seeded `Sev1` case.

## Verification

move_node fixture, `uip` 1.203.0, `uip maestro flow validate`:

| read | topology | result |
|---|---|---|
| `matchedCaseId` | branches merge | `EXPRESSION_DIAGNOSTIC` |
| `matchedCase` | branches merge | same warning |
| `matchedCaseId` | **no merge** | same warning |
| `matchedCaseId`, retyped to `core.logic.switch` | branches merge | same warning |
| condition recomputed from upstream | branches merge | **Valid, no warnings** |
| `inout` global set per branch | branches merge | **Valid, no warnings** |

The key is not the problem (`matchedCase` fails identically) and neither is merging (it fails with one branch). The printed scope carries nodes upstream *of* the gateway but never the gateway. `core.logic.switch` behaves as `core.logic.decision` does — which is why `switch/impl.md` had no wrong Outputs section: it had none at all.

Runtime values are unchanged and still correct where the value is reachable (`flow-workbench packages/services/src/debug/gateway-script.ts`): switch `matchedCaseId` is the case id, `null` for default; decision is `"true"`/`"false"`.

## Changes

- **`decision/impl.md`** — the rule, plus the two patterns that validate clean
- **`switch/impl.md`** — adds the Outputs section it never had; points at the above
- **`brownfield.md`** — move/reorder row stops prescribing the broken read
- **`variables-and-expressions.md`** — the canonical scope rule now carries the gateway exception
- **`failure-modes.md`** — documents `EXPRESSION_DIAGNOSTIC` next to `MISSING_OUTPUT_MAPPING`

## Caveats

- **Verified at the validator only** — the two eval tasks have not been re-run end to end against these docs.
- `.maintenance/check-all.sh` reports 3 failures, all pre-existing in files this PR does not touch: a bad anchor in `troubleshooting-guide.md`, and two `uip agent tool add` commands in `inline-agent/`. Left for a separate PR.

## tl;dr

The skill documented a branch-merge pattern that cannot work. A gateway is unreadable from downstream — merge or not, either key. One agent shipped a wrong result, another burned its whole budget decompiling the SDK looking for the answer. Fixed the three docs that taught it, added the missing Switch section, and documented the warning so the next agent resolves it from the docs.


skills/uipath-maestro-flow/references/diagnose/failure-modes.md
```diff
@@ -333,6 +333,7 @@ Multiple. `flow validate` runs a JSON schema check, cross-reference checks, expr
 - References to unknown variable IDs or node IDs in `=js:` expressions (`EXPR_UNRESOLVED_REF`) — flow-schema `expression-ref` rule
 - Output-path walks that descend into a declared primitive (`type: "string"` etc.) or a schema closed with `additionalProperties: false` (`EXPR_INVALID_OUTPUT_PATH`) — flow-schema `expression-ref` rule
 - Missing End-node output mappings for declared `out` variables (`MISSING_OUTPUT_MAPPING`, **warning** severity) — flow-schema `output-mapping` rule
+- A `$vars.<nodeId>` read whose node is not in the reader's scope (`EXPRESSION_DIAGNOSTIC`, **warning** severity). The message prints the scope it *is* in: `Property '<nodeId>' does not exist on type '{ … }'`. Most often `<nodeId>` is a Decision or Switch, which no downstream node can read — see [decision/impl.md — Outputs](../author/plugins/decision/impl.md#outputs). Otherwise the node runs after the reader, or on a branch the reader is not on. Never ship this warning: it validates, then reads `undefined` at runtime and silently produces a wrong result rather than faulting
```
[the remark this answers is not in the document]
rockymadden: Correct, and the citations check out. Fixed in 55bfcdc1c.

The bullet named three causes and then asserted one consequence for all of them. That was only true of the gateway case, which is the one I verified. Now split three ways:

- **Gateway** — reads `undefined`, silently yields a wrong result instead of faulting
- **Node runs after the reader** — faults `[400302]`/`[400300]`
- **Node on an untaken branch** — reads `undefined`, having never executed

The third was also folded into the blanket claim; `variables-and-expressions.md` already documents it as `undefined` rather than a fault.

rockymadden:
Copilot review addressed in `55bfcdc1c`; thread resolved.

**The finding was valid.** Verified against the cited sources rather than taken on trust — both check out:

- `decision/impl.md` Debug table documents `[400302] Error evaluating outgoing flow expression from gateway` for an expression reading a node that runs *after* it
- `brownfield.md:36` says a later-running reference "evaluates against nothing at runtime (`[400302]`/`[400300]`)"

The bullet named three causes and then asserted **one** runtime consequence for all of them — silent `undefined`, never a fault. That is only true of the gateway case, the one this PR actually verified. On a diagnose page it would walk a reader straight past a real `[400302]`.

Copilot flagged two of the three. The third was wrong for the same reason: a node on an untaken branch reads `undefined` rather than faulting, per `variables-and-expressions.md` §Node Output Availability. So it is now split three ways:

| cause | runtime |
| --- | --- |
| Gateway (Decision / Switch) | `undefined`, silent wrong result |
| Node runs after the reader | faults `[400302]` / `[400300]` |
| Node on an untaken branch | `undefined`, never executed |

One line changed. Links 917 / 0 broken — the new `brownfield.md#common-edits` target resolves. The one bad anchor `check-anchors.sh` reports is pre-existing in `troubleshooting-guide.md`, untouched here.

No change to the gateway rule itself or to the six-variant verification behind it.

skills/uipath-maestro-flow/references/author/plugins/decision/impl.md
```diff
@@ -55,7 +55,12 @@ Output ports: `true` and `false`. Both branches must be wired. See [editing-oper
 
 ## Outputs
 
-A Decision has **no `.output`**. Downstream nodes read which branch ran as `$vars.<decisionId>.matchedCaseId` (`"true"` or `"false"`) or `$vars.<decisionId>.matchedCase` (the branch label) — e.g. `$vars.checkStatus.matchedCaseId === "true"`. `$vars.<decisionId>.output.matchedCaseId` is undefined and faults the reader at runtime (`[300501]`).
+A Decision is **not readable from downstream at all**. It never enters a downstream node's expression scope, so every form — `$vars.<decisionId>.matchedCaseId`, `.matchedCase`, `.output.matchedCaseId` — resolves to `undefined`. This holds whether or not the branches rejoin. `flow validate` reports it as `[EXPRESSION_DIAGNOSTIC] Property '<decisionId>' does not exist on type '{…}'`, listing a scope the Decision is absent from; the flow still validates, then reads `undefined` at runtime and silently takes the wrong branch of the reader's own logic.
```
bryan-uipath: I wonder if we need to mention `matchedCaseId` at all - wonder if we could just delete this block?

bryan-uipath: oh... i see... its defined in the node schema but doesn't do anything ☹️ https://github.com/UiPath/flow-workbench/blob/5b58299f1c0749add462ea30476640bf362db7f2/packages/flow-core/src/manifest/data/ootb-nodes/core-logic-switch/v1.1.ts#L124

bryan-uipath: ...which would be confusing for the coding agent reading the schema

bryan-uipath (approved):

skills/uipath-maestro-flow/references/author/plugins/decision/impl.md
rockymadden: Your second comment answers your first — that is exactly why the block has to stay. `registry get` shows both outputs and `flow format` writes them into `variables.nodes[]`, so deleting it leaves the agent with only the schema's promise and nothing contradicting it. Rewritten in `df65372a1` to lead with the schema instead of the symptom: "`registry get` shows a Decision declaring `matchedCase` and `matchedCaseId` … **Ignore both. Neither is ever assigned at runtime.**"

Confirming your read, with the mechanism, since it is worse than dead weight:

- The scope resolver suppresses a gateway's outputs for every reader except the gateway itself — `SYNTHETIC_GATEWAY_OUTPUT_KEYS = ["matchedCase", "matchedCaseId"]` in maestro-tool 1.203.0
- `bpmn-to-xml.ts` emits no writer for them: the branch choice lives on the outgoing sequence flows' `conditionExpression`, never in a variable. They carry real values only in the canvas, which evaluates gateways client-side via `gateway-script.ts`
- `collectVariablesFromNodes` still *declares* them from `variables.nodes[]` — so the read is declared but unassigned, which is why it returns `undefined` **silently** instead of faulting

That last bit is what makes it costly rather than merely confusing. `skill-flow-interactive-customer-escalation-triage` shipped this and returned `Sev2` for a seeded `Sev1` case; `skill-flow-move-node` hit the warning and spent ~950s of a 1200s budget reading the minified bundle looking for the reason. The mechanism is now in the doc so nobody has to rediscover it.

One more thing worth your eyes — the CLI actively pushes agents toward it:

```
$ # reading $vars.checkTemperature.output.matchedCaseId
[error] node "checkTemperature" has no output named "output" — its outputs are:
matchedCase, matchedCaseId. Read "$vars.checkTemperature.<one of those>"
(e.g. a Decision exposes "$vars.checkTemperature.matchedCaseId", ...)
```

The cross-reference check reads the registry's declared outputs and does not know the scope resolver suppresses them, so it recommends the broken read. The doc now says to take only half that advice. If the outputs are genuinely dead server-side, the real fix is upstream — drop them from the schema, or mark them canvas-only so `registry get` stops advertising them. Happy to file that if you agree.

rockymadden:
Pushed `df65372a1` addressing @bryan-uipath's thread.

**The Outputs block stays, but now leads with the schema rather than the symptom.** Bryan's second comment answered his first: `matchedCase`/`matchedCaseId` are declared in the node schema and do nothing, which is confusing for an agent reading the schema — so deleting the block would leave only the schema's promise with nothing contradicting it.

**Mechanism is now in the doc**, so nobody rediscovers it the hard way:

- The scope resolver suppresses a gateway's outputs for every reader except the gateway itself — `SYNTHETIC_GATEWAY_OUTPUT_KEYS = ["matchedCase", "matchedCaseId"]`, maestro-tool 1.203.0
- `bpmn-to-xml.ts` emits no writer: the branch choice lives on the outgoing sequence flows' `conditionExpression`, never in a variable. Real values exist only in the canvas, which evaluates gateways client-side
- `collectVariablesFromNodes` still *declares* them from `variables.nodes[]` — declared but unassigned, which is why the read is silently `undefined` instead of a fault

That settles the one thing the PR previously asserted from a single data point.

**Also corrected a claim from the first commit.** The three forms are not alike:

| written | result |
| --- | --- |
| `$vars.<id>.matchedCaseId` / `.matchedCase` | validate passes + `EXPRESSION_DIAGNOSTIC` warning |
| `$vars.<id>.output.matchedCaseId` | validate **fails** |

Verified on the move_node fixture, `uip` 1.203.0. The failure text then recommends `$vars.<id>.matchedCaseId` — the cross-reference check reads the registry's declared outputs and is unaware of the suppression — so the doc says to take only half that advice.

---

**Unrelated red X:** `Run skill smoke tests` fails at 88% (22/25) on three `datafabric-smoke-*` tasks. The identical three fail on `feat/maestro-flow-conversational-text-skill` (run `34396685605`, 19:42) which predates this branch and touches no gateway docs. Not from this PR — this diff is five markdown files. A narrow diff scopes the smoke matrix to 25 tasks, so the same 3 environmental failures cost 12% instead of the 5.6% they cost a broader PR.

**Possible upstream follow-up:** if these outputs are dead server-side, the durable fix is to drop them from the node schema or mark them canvas-only so `registry get` stops advertising them. Happy to file it.

skills/uipath-maestro-flow/references/diagnose/failure-modes.md
```diff
@@ -333,6 +333,7 @@ Multiple. `flow validate` runs a JSON schema check, cross-reference checks, expr
...
 - Output-path walks that descend into a declared primitive (`type: "string"` etc.) or a schema closed with `additionalProperties: false` (`EXPR_INVALID_OUTPUT_PATH`) — flow-schema `expression-ref` rule
 - Missing End-node output mappings for declared `out` variables (`MISSING_OUTPUT_MAPPING`, **warning** severity) — flow-schema `output-mapping` rule
+- A `$vars.<nodeId>` read whose node is not in the reader's scope (`EXPRESSION_DIAGNOSTIC`, **warning** severity). The message prints the scope it *is* in: `Property '<nodeId>' does not exist on type '{ … }'`. Never ship it, but diagnose the runtime symptom from *why* the node is out of scope, because the three causes do not behave alike: a Decision or Switch is unreadable from anywhere downstream and reads `undefined`, silently yielding a wrong result instead of faulting ([decision/impl.md — Outputs](../author/plugins/decision/impl.md#outputs)); a node that runs *after* the reader faults instead, with `[400302]`/`[400300]` ([brownfield.md — Common edits](../author/brownfield.md#common-edits)); a node on a branch the reader is not on reads `undefined`, having never executed
```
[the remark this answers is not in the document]
rockymadden: Right, and measured rather than reasoned. On the move_node fixture, `uip` 1.203.0:

```
warning-only  -> exit 0
error         -> exit 1
```

So the `**Caught** (validate exits non-zero…)` heading was wrong for two entries under it: `MISSING_OUTPUT_MAPPING`, which predates this branch, and the `EXPRESSION_DIAGNOSTIC` added here.

It bites hardest on the new entry, whose whole instruction is "never ship it". A reader who trusts the heading checks `$?`, sees `0`, and ships the warning — which is precisely how `skill-flow-interactive-customer-escalation-triage` shipped a wrong severity. `greenfield.md:346` already states the rule ("a passing exit code with warnings is NOT done"); the diagnose index contradicted it.

Fixed in `61bfb54e0`:

> **Caught** (each with a precise field path and remediation hint). **Error** severity exits non-zero; **warning** severity still exits **0**, so read `Data.Warnings` rather than the exit code:

rockymadden:
Pushed `61bfb54e0` for the second Copilot round; thread resolved.

**Valid, and measured rather than reasoned.** On the move_node fixture, `uip` 1.203.0:

```
warning-only  -> exit 0
error         -> exit 1
```

The `**Caught** (validate exits non-zero…)` heading was therefore wrong for two entries beneath it: `MISSING_OUTPUT_MAPPING`, which predates this branch, and the `EXPRESSION_DIAGNOSTIC` added here.

It bites hardest on the new entry, whose whole instruction is *never ship it*. A reader trusting the heading checks `$?`, sees `0`, and ships the warning — which is exactly how `skill-flow-interactive-customer-escalation-triage` shipped a wrong severity. `greenfield.md:346` already carried the rule; the diagnose index contradicted it. Heading now distinguishes the two.

---

### State

- @bryan-uipath approved at 21:58Z
- Both Copilot threads addressed and resolved
- Bryan's thread left **open** on purpose — I asked him a question in it (below) that has not been answered
- `check-links` 918/0; the single bad anchor (`troubleshooting-guide.md`) is pre-existing and untouched here

### Open question for @bryan-uipath

This PR documents *around* the problem. The durable fix is upstream: if `matchedCase`/`matchedCaseId` are dead server-side, drop them from the node schema or mark them canvas-only, so `registry get` stops advertising outputs that never get assigned and the CLI's cross-reference check stops recommending them. Happy to file that against flow-workbench — say the word.

### Still-red check, not from this PR

`Run skill smoke tests` fails at 88% (22/25) on three `datafabric-smoke-*` tasks. The identical three fail on `feat/maestro-flow-conversational-text-skill` (run `34396685605`, 19:42), which predates this branch and touches no gateway docs. This diff is five markdown files.

PR #3180, opened 2026-09-10
fix(uipath-maestro-flow): wire per-outcome handles, drop the forced Decision node after HITL

dushyant-uipath:
## Problem
QuickForm HITL nodes wire one `outcome-<outcome.id>` port per outcome. Docs, task fixtures, and one checker across both `uipath-maestro-flow` and `uipath-human-in-the-loop` taught or graded a single `completed` handle instead, and a canonicalized Decision node was forced after every HITL node regardless of outcome count.

## Fixed
- 8 uipath-maestro-flow references taught `completed` unconditionally: corrected to per-outcome `outcome-<id>` ports.
- Sibling uipath-human-in-the-loop skill contradicted the fix (SKILL.md, hitl-node-quickform.md, hitl-node-apptask.md): corrected.
- A coded-action-app example flipped onto outcome-derived ports by gaining an `inputs.schema` block it shouldn't have had: block removed, port stays the static `completed` for that node type.
- `outcome-completed` documented as never reappearing once outcomes exist (impl.md, planning.md): corrected. It is the zero-outcome placeholder, or the port for a real outcome whose own `id` happens to be `completed`.
- 22 task fixtures across both skills prompted for or graded a single `completed` handle, or checked wiring with a file-scoped regex against a guessed literal port id: prompts and success criteria reworded to outcome-port language; `smoke_02`, `quality_04` in both skills, and the 6 `e2e_*` invoice/escalation/gdpr/multi-hitl/expense tasks now call a new node-scoped, id-derived `check_simulated_hitl.py outcome-wiring` check instead. `TEST_PLAN.md` updated to match.
- `check_simulated_hitl.py` and `check_devcon_expense_approval.py` derived the wiring port from a lowercased `outcome.name`, falling back to that when `id` was missing: both now read `id` verbatim (matching flow-workbench's `build-handle-customization.ts`) and fail loudly when it is missing, instead of validating a flow that would leave an outcome dangling at runtime.
- Same two checkers ran the "outcome-completed misuse" check after the "missing ports" check, so the more specific message never fired: reordered.
- A stable-status flip for `uipath-human-in-the-loop` was bundled into the original fix: split out to #3224, gated on this merging first.

## Review findings addressed
All 12 inline review comments from the last pass replied to individually and resolved: documentation cascade (6 files), sibling-skill contradiction, the two overclaim rewordings, the checker id-derivation bug, the check-ordering bug, both test fixture gaps (mixed-case id, missing id), and the file-scoped-regex replacement.

## Test plan
- [x] maestro-flow checker unit tests: green in CI
- [x] `skill-flow-hitl-smoke-completed-port` (smoke_02): SUCCESS, score 1.00, in CI
- [x] `skill-hitl-e2e-invoice-approval-greenfield` (e2e_01) run locally: SUCCESS, score 1.000, 7/7 criteria, 482.7s
- [x] Full local run of both directly affected checker test files (19/19 passing) plus the whole uipath-maestro-flow suite: no HITL/outcome-related failures elsewhere (133 pre-existing failures are in `test_criterion_budgets.py` and a Data Fabric connector suite, matching CI's known flakiness there)
- [ ] Remaining ~13 affected integration/e2e tasks not run through a full coder-eval pass

## Known overlap, needs reconciling before merge
#3239 (different author) touches the same two checker files here and claims the default v1.0 HITL node only ever accepts a `completed` edge, with outcome-id ports gated behind an opt-in `outcomePorts` flag on v1.1+. A local coder-eval run authored a `typeVersion: "1.0"` node wired with `outcome-approve`/`outcome-reject` edges and `uip maestro flow validate` passed clean, contradicting that. Needs reconciling before either PR merges.


chetanyauipath (approved):

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
```diff
@@ -8,7 +8,7 @@ Preferred: no registry pull, app publishing, or tenant dependency. Write the nod
 
 For schema design, node writing, JSON examples, and schema conversion rules, see [`uipath-human-in-the-loop` skill — hitl-node-quickform.md](../../../../../uipath-human-in-the-loop/references/hitl-node-quickform.md). Skills are self-contained: this cross-skill reference is for documentation context only — per Critical Rule 4, offer the handoff and let the user choose. This guide covers implementation-phase topology resolution only, not schema design or node writing.
 
-For add, delete, and wiring procedures, see [editing-operations.md](../../editing-operations.md). **Use `Edit` / `Write` for HITL node authoring.** Do not use the dedicated HITL CLI for this non-carve-out structural edit. Wire the `outcome-completed` port after adding the node.
+For add, delete, and wiring procedures, see [editing-operations.md](../../editing-operations.md). **Use `Edit` / `Write` for HITL node authoring.** Do not use the dedicated HITL CLI for this non-carve-out structural edit. Wire one output edge per outcome — port `outcome-<outcome.id>` for each entry in `inputs.schema.outcomes[]` — never a single shared port.
```
rockymadden, line 11: 🔴 **Critical** — this sentence sends the reader to `editing-operations.md` for wiring, and `editing-operations.md:31` still says:

> Includes HITL QuickForm; wire `completed` after adding.

Same in `planning-arch.md:227` (the port table: `quick-form | input | completed`), `brownfield.md:49`, `shared/cli-commands.md:216`, and the whole `diagnose/failure-modes.md:132-144` section "HITL `completed` port unwired" (linked from `diagnose/CAPABILITY.md:49`).

An agent that follows the link gets the rule this line just corrected. Six files, one skill, one PR.

**Fix:** update all six here. `failure-modes.md` needs the heading and anchor renamed too (`#hitl-completed-port-unwired` → per-outcome), plus its two referrers at `failure-modes.md:14,343` and `CAPABILITY.md:49`.

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
```diff
@@ -8,7 +8,7 @@ Preferred: no registry pull, app publishing, or tenant dependency. Write the nod
 
 For schema design, node writing, JSON examples, and schema conversion rules, see [`uipath-human-in-the-loop` skill — hitl-node-quickform.md](../../../../../uipath-human-in-the-loop/references/hitl-node-quickform.md). Skills are self-contained: this cross-skill reference is for documentation context only — per Critical Rule 4, offer the handoff and let the user choose. This guide covers implementation-phase topology resolution only, not schema design or node writing.
```
rockymadden, line 9: 🔴 **Critical** — this line delegates "schema design, node writing, JSON examples" to `uipath-human-in-the-loop`, which teaches the opposite as a Critical Rule:

- `uipath-human-in-the-loop/SKILL.md:36` — "**Always wire the `completed` handle.** … Only `completed` is available as an output handle — **not** `output`, `success`, or any other name."
- `uipath-human-in-the-loop/references/hitl-node-quickform.md:246,249,251` — the exact file this line links — wires `"sourcePort": "completed"` and repeats "Always wire `completed`."

So the agent gets the corrected rule from this guide and the wrong one from the reference it is pointed at, on the very topic it was pointed there for. The wrong one is phrased as a Critical Rule with an explicit exclusion list, which is the more forceful of the two.

And `assets/skill-status.json` in this same PR marks that skill `stable`.

**Fix:** correct `SKILL.md:36` and `hitl-node-quickform.md:246-251` in this PR. If that is out of scope, drop the `stable` flip — do not ship a skill as stable in the commit that disproves its Critical Rule 2.

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
```diff
@@ -66,7 +66,7 @@ Rules:
 - `typeVersion` — always `"1.0"` for this node. **Do not run `registry get` to derive this value; do not use `"1.1"` or any other version.** The OOTB HITL node version is stable at `1.0`.
 - Do not include a `model` block on node instances; only the definition carries it.
 - `outputs` contains only `output` (with `properties` for output/inOut fields plus `Action`) and `status` (with outcome `enum`/`default`). Do not add per-field `custom: true` entries.
-- Ports: `input` (target) → `outcome-completed` (source, label: Completed).
+- Ports: `input` (target) → one `outcome-<outcome.id>` port per outcome (source), derived from `inputs.schema.outcomes[].id` — never hardcode a port name. `outcome-completed` is the zero-outcome placeholder only: it disappears the instant the schema has any outcome, including the shipped default `Submit` (id `submit`, port `outcome-submit`). Wire every outcome port — an unwired one blocks that branch indefinitely.
```
rockymadden, line 69: 🟡 **Minor** — "`outcome-completed` never reappears after that" (and `:182`) is stronger than the source supports. From `hitl-edge-reconciliation.ts`:

> These are ordinary strings, not reserved ones: `outcomeHandleId('completed')` produces `outcome-completed`, identical to the legacy placeholder spelling, so a real outcome a customer named (or id'd) "completed" is indistinguishable from the placeholder by string match alone.

The product guards that case explicitly. Both checkers in this PR hard-fail it.

**Fix:** "`outcome-completed` is the port for a zero-outcome node, or for a real outcome whose `id` is literally `completed` — never a shared exit for several outcomes."

Also missing from this rule: an outcome **without** an `id` gets no handle at all. `build-handle-customization.ts` filters on `typeof o?.id === 'string' && o.id.length > 0`, and `collectOutcomeConditions` skips the same. Worth one clause here, because `hitl-node-quickform.md`'s conversion table has a rule for field `id` and none for outcome `id` — which is what the checkers below try to paper over.

skills/uipath-maestro-flow/references/author/plugins/hitl/planning.md
```diff
@@ -46,11 +46,13 @@ With no user to ask, pick one that terminates unattended and record which you ch
...
-| Input port | Output port |
+| Input port | Output port(s) |
 | --- | --- |
-| `input` | `outcome-completed` |
+| `input` | one per outcome — `outcome-<outcome.id>`, derived from `inputs.schema.outcomes[].id` |
+
+**Every outcome gets its own port — wire one edge per outcome, not a single shared port.** A schema with `outcomes: [Approve, Reject]` (ids `approve`, `reject`) produces two ports, `outcome-approve` and `outcome-reject`; each needs its own edge. A node with any outcome port left unwired blocks the flow indefinitely on that branch.
 
-**The output port must be wired.** A node with no edge on `outcome-completed` blocks the flow indefinitely.
+> **`outcome-completed` is not a real outcome port — it is the placeholder shown only before the schema has any outcome.** The instant `inputs.schema.outcomes` has one or more entries — including the shipped default `Submit` (id `submit`) — its port becomes `outcome-submit`, not `outcome-completed`. Never wire `outcome-completed` as a real branch once outcomes exist; it appears only transiently on a freshly-added, not-yet-configured node.
```
rockymadden, line 55: 🟡 **Minor** — same over-claim as `impl.md:69`: "is not a real outcome port" is false when an outcome's own `id` is `completed`. `outcomeHandleId('completed')` and the placeholder are the same string; `hitl-edge-reconciliation.ts` never treats either spelling as stale without first checking the node's real outcome handles.

**Fix:** say it is the port for a node with no outcomes, not that the name can never be real.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
```diff
@@ -134,12 +134,35 @@ def check_expense(flow: dict[str, Any], nodes: list[dict[str, Any]]) -> None:
...
-        edge.get("sourceNodeId") == hitl_id
-        and edge.get("sourcePort") in {"completed", "outcome-completed"}
+    # outcome-completed is the zero-outcome placeholder only (confirmed against
+    # flow-workbench source, 2026-09-10) — it disappears the instant the schema
+    # has any real outcome. This node has real outcomes, so every one of them
+    # needs its own wired outcome-<id> handle; outcome-completed must not
+    # appear. An outcome's id defaults to its lowercased name when not given
+    # explicitly (same convention as hitl-node-quickform.md's own examples).
+    outcome_ids = [
+        str(o.get("id") or o.get("name") or "").lower()
```
rockymadden: 🟠 **Major** — `.lower()` and the `name` fallback are both wrong against the product. From `build-handle-customization.ts`:

```ts
const outcomes = (schema.outcomes ?? []).filter((o) => typeof o?.id === 'string' && o.id.length > 0);
... outcomes.map((o) => ({ id: outcomeHandleId(o.id), ... }))
```

`outcomeHandleId` is `` `outcome-${outcomeId}` `` — the `id` **verbatim**. `collectOutcomeConditions` in `hitl-serialization.ts` matches `outcomeIdFromHandle(handle)` against `o.id` case-sensitively too.

Two failure modes:

1. **Fails correct work.** Outcome `{"id": "needsInfo"}` → product renders `outcome-needsInfo`; this demands `outcome-needsinfo` → `missing` → task scores 0 on a flow that runs.
2. **Passes invalid work.** Outcome `{"name": "Approve"}` with no `id` gets **no handle at all**, and `collectOutcomeConditions` skips it, so `outcome-approve` is an edge to a nonexistent outcome — no condition stamped, unconditional flow. This code accepts it as correct. That is the shape `test_batch2_alignment.py:59` now asserts.

**Fix:** use the `id` verbatim, drop the `name` fallback, and fail loudly on an outcome with no `id`:

```python
outcome_ids = [o["id"] for o in outcomes if isinstance(o.get("id"), str) and o["id"]]
if len(outcome_ids) != len(outcomes):
    fail("every outcome needs a non-empty string id — an outcome without one renders no handle")
```

Then add the missing outcome-`id` derivation rule to `hitl-node-quickform.md`'s conversion table (it has one for field `id`, none for outcome `id`) so the agent has a deterministic rule instead of the checker guessing.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
```diff
@@ -134,12 +134,35 @@ def check_expense(flow: dict[str, Any], nodes: list[dict[str, Any]]) -> None:
...
+        for o in outcomes
+        if o.get("id") or o.get("name")
+    ]
+    if not outcome_ids:
+        fail("HITL schema needs at least one outcome with an id or name")
```
rockymadden: 🟠 **Major** — this makes outcomes mandatory, contradicting `check_expense`'s own gate 25 lines up:

```python
if not (has_decision or has_outcomes):
    fail("need a boolean decision field or approve/reject outcomes")
```

That check deliberately accepts a schema with a boolean decision field **and no outcomes**. This one then hard-fails it, so the `has_decision`-only path is now dead: any solution taking it goes from pass to `FAIL` with a message about outcome ids that the task never asked for.

Unlike `check_devcon_expense_approval.py`, there is no earlier "must define outcomes" gate here, so this is a live behavior change to `expense_approval_simulated` — a task that was not re-run.

**Fix:** pick one contract. Either tighten the gate at `:124` to require outcomes and say so in the task prompt, or keep the per-outcome port assertions inside the `has_outcomes` branch and leave the boolean-decision path alone.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
```diff
@@ -134,12 +134,35 @@ def check_expense(flow: dict[str, Any], nodes: list[dict[str, Any]]) -> None:
         for field in fields
     ):
         fail("need a text reason output field")
-    if not any(
-        edge.get("sourceNodeId") == hitl_id
-        and edge.get("sourcePort") in {"completed", "outcome-completed"}
+    # outcome-completed is the zero-outcome placeholder only (confirmed against
```
rockymadden, line 137: 🟡 **Minor** — this 6-line comment is duplicated near-verbatim in `check_devcon_expense_approval.py:120-125` and again in `smoke_02_completed_port_wired.yaml:66-72`. `.claude/rules/token-optimization.md` § 3: state a fact once in full, point at it from the other N sites.

"confirmed against flow-workbench source, 2026-09-10" is also reasoning a future reader cannot act on (`.claude/rules/skill-structure.md` / repo skills rule: don't narrate or cite reasoning a future reader won't use). A file path they can open is actionable; a date is not.

**Fix:** one line here — `# Handle id is outcome-<id>, verbatim: flow-workbench build-handle-customization.ts` — and a pointer from the other two.

tests/tasks/uipath-maestro-flow/e2e/check_devcon_expense_approval.py
```diff
@@ -117,12 +117,31 @@ def main() -> None:
...
-    ):
-        fail("HITL completed handle must be wired")
+    # outcome-completed is the zero-outcome placeholder only (confirmed against
+    # flow-workbench source, 2026-09-10) — it disappears the instant the schema
+    # has any real outcome, and `outcomes` is already guaranteed non-empty
+    # above. Every outcome needs its own wired outcome-<id> handle instead. An
+    # outcome's id defaults to its lowercased name when not given explicitly
+    # (same convention as hitl-node-quickform.md's own examples).
+    outcome_ids = [
+        str(o.get("id") or o.get("name") or "").lower()
```
rockymadden: 🟠 **Major** — same `.lower()` + `name`-fallback bug as `check_simulated_hitl.py:144`; see that comment for the source evidence and the fix. Apply it identically here.

One extra note on the comment above: "`outcomes` is already guaranteed non-empty above" is true, but an outcome that is present with **neither** `id` nor `name` still slips past the comprehension's filter, so `:131` is not dead code — it just fires with a message that says the schema needs an outcome when the schema has one. Replacing the fallback with the verbatim-`id` version above makes that message accurate.

tests/tasks/uipath-maestro-flow/_shared/test_batch2_alignment.py
```diff
@@ -56,10 +56,16 @@ def test_simulated_hitl_checks_accept_root_sdk_emit(tmp_path: Path) -> None:
         "edges": [
             {
                 "sourceNodeId": "review",
-                "sourcePort": "completed",
+                "sourcePort": "outcome-approve",
```
rockymadden, line 62: 🟠 **Major** — this fixture's outcomes are `[{"name": "Approve"}, {"name": "Reject"}]` (`:43`) — **no `id`**. Per `build-handle-customization.ts` those outcomes are filtered out and render **zero** handles, and `collectOutcomeConditions` skips them, so `outcome-approve`/`outcome-reject` are edges to nonexistent outcomes: no `conditionExpression` gets stamped, and the BPMN engine takes them as unconditional flows. Both branches fire on every run.

This test now asserts that shape is correct, and it is the reason commit c918a1c added the `name` fallback. The fallback exists to satisfy an invalid fixture.

**Fix:** give the fixture real ids and drop the fallback:

```python
"outcomes": [{"id": "approve", "name": "Approve"}, {"id": "reject", "name": "Reject"}],
```

Then add a negative case asserting a `name`-only outcome **fails** — that is the bug class worth locking down.

tests/tasks/uipath-maestro-flow/e2e/test_check_devcon_expense_approval.py
```diff
@@ -19,7 +19,15 @@ def _flow_doc(
...
     script_body: str = "return $vars.reviewExpense.output.rejectionreason;",
+    outcome_ports: list[str] | None = None,
 ) -> dict[str, Any]:
+    # By default wire one edge per outcome id (outcome-<id>) — the only shape
+    # the checker now accepts, since outcome-completed is exclusively the
+    # zero-outcome placeholder and disappears once real outcomes exist. Pass
+    # outcome_ports to build a deliberately wrong/incomplete edge set for
+    # negative tests.
+    if outcome_ports is None:
+        outcome_ports = [f"outcome-{o['id']}" for o in outcomes]
```
rockymadden, line 30: 🟠 **Major** — this builder uses `o['id']` **verbatim**; the checker it exercises lower-cases the same value (`check_devcon_expense_approval.py:127`). Every fixture here is already lowercase (`_APPROVE_REJECT_OUTCOMES` at `:96-99`), so the two spellings never diverge and the casing bug is structurally unreachable from this suite. That is why `maestro-flow checker unit tests` is green.

**Fix:** once the checker compares verbatim, add a mixed-case case that pins the contract:

```python
def test_accepts_mixed_case_outcome_id(tmp_path):
    outcomes = [{"id": "needsInfo", "name": "Needs Info"}, {"id": "reject", "name": "Reject"}]
    # ports outcome-needsInfo / outcome-reject must PASS
```

It fails on `main` and on this branch, and passes with the verbatim fix.

tests/tasks/uipath-maestro-flow/hitl/smoke_02_completed_port_wired.yaml
```diff
@@ -51,17 +56,20 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: run_command
-    description: "Flow wires the HITL completed output port"
-    command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/flow_contains.py 'completed'"
+    description: "Flow wires a distinct outcome-<id> port for both Approve and Reject — not the generic outcome-completed placeholder"
+    command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/flow_contains.py --regex '\"sourcePort\":\\s*\"outcome-approve\"' --regex '\"sourcePort\":\\s*\"outcome-reject\"' --absent-regex '\"sourcePort\":\\s*\"(outcome-)?completed\"'"
```
rockymadden, line 60: 🟡 **Minor** — two issues in this one command:

1. `--absent-regex '"sourcePort":\s*"(outcome-)?completed"'` is file-scoped, not node-scoped. It fires on **any** node's `completed` port, so this criterion breaks the moment the task grows a node that legitimately has one — including an Option 2 HITL node, whose static `completed` handle is correct (`hitl-edge-reconciliation.ts`: "their `completed` handle is genuinely the only one that ever renders"). Fine for a 4-node flow today; a landmine as written.
2. `outcome-approve`/`outcome-reject` are pinned to ids no documented rule produces. The comment at `:70-72` justifies them from `hitl-node-quickform.md` Example 1 — but that is an *example*, and the conversion-rules table at `:346-357` of that file has a rule for field `id` and none for outcome `id`. An agent that writes `"id": "Approve"` is following the skill and still scores 0 here (weight 3.0).

**Fix:** add the outcome-`id` rule to that table (`lowercase name, spaces→-`), then this pin is derived rather than guessed. For (1), assert against the HITL node's own edges — `check_simulated_hitl.py` already resolves `hitl_id` — instead of a whole-file regex.

assets/skill-status.json
```diff
@@ -106,7 +106,7 @@
       "last_synced": null
     },
     "uipath-human-in-the-loop": {
-      "status": "in-development",
+      "status": "stable",
```
rockymadden: 🟠 **Major** — `stable` is the wrong call in this commit, on two counts.

1. **This PR disproves the skill's own Critical Rule 2.** `uipath-human-in-the-loop/SKILL.md:36` says "Only `completed` is available as an output handle," and `references/hitl-node-quickform.md:246-251` wires it. This PR establishes that is wrong for every quick-form node with outcomes — i.e. all of them, since `inputDefaults` ships `[{id:"submit"}]`. Shipping that as `stable` in the same commit is contradictory. `.claude/rules/skill-structure.md`: `stable` = "production-ready surface; safe for production."
2. **Unrelated change.** `.claude/rules/pr-review.md` § Commit and Branch Conventions — one logical change per PR. The 96-100% nightly evidence is not in the diff and no run is linked, so it cannot be reviewed here.

**Fix:** revert this and the `README.md` row, fix `SKILL.md:36` and `hitl-node-quickform.md:246-251` here, then flip to `stable` in a follow-up that links the runs.

rockymadden (changes requested):
## What

Rewrites the HITL wiring guidance in `uipath-maestro-flow`: one `outcome-<outcome.id>` port per outcome instead of a single `outcome-completed`, and drops the canonical "insert a Decision node after HITL" pattern. Realigns 3 checkers + 1 smoke task, adds 2 negative tests. Also flips `uipath-human-in-the-loop` to `stable`.

**Both core claims hold.** I verified them against `flow-workbench@develop` rather than taking the description at face value:

- `packages/canvas/src/components/properties-panel/hitl/build-handle-customization.ts` builds one handle per outcome, id `outcome-<outcome.id>`; `outcome-completed` only when `outcomes` is empty. `inputDefaults` ships `[{id:"submit"}]`, so even a fresh node is `outcome-submit`. ✅
- `packages/services/src/serialization/hitl-serialization.ts` → `collectOutcomeConditions` stamps `=js:vars.<node>.output.Action === "<name>"` onto each `outcome-*` edge. The handle **is** the conditional branch, so the Decision node is genuinely redundant. ✅

The diagnosis is right. The **execution is half-done**, and the checker rewrite encodes a rule the product does not follow.

---

## Overall findings

### 🔴 Critical — the fix stops at 2 of 8 files in its own skill

`uipath-maestro-flow` still teaches `completed` in six places, all untouched:

| File:line | Text |
|---|---|
| `references/author/editing-operations.md:31` | "Includes HITL QuickForm; wire `completed` after adding." |
| `references/author/planning-arch.md:227` | `\| uipath.human-in-the-loop.quick-form \| input \| completed \|` |
| `references/author/brownfield.md:49` | "wire its `completed` port" |
| `references/shared/cli-commands.md:216` | "After adding, wire the `completed` port" |
| `references/diagnose/failure-modes.md:14,132,140,144,343` | whole section "HITL `completed` port unwired" |
| `references/diagnose/CAPABILITY.md:49` | links to that section |

`impl.md:11` — the line this PR edits — points at `editing-operations.md` for wiring in the same sentence that now says `outcome-<id>`. An agent following the link gets the old rule back.

**Fix:** update all six in this PR, including the `failure-modes.md` anchor (`#hitl-completed-port-unwired`) and its two referrers.

### 🔴 Critical — the sibling skill this one delegates node writing to still teaches `completed`, and this PR marks it stable

`impl.md:9` sends schema design, node writing and JSON examples to `uipath-human-in-the-loop`. That skill says:

- `SKILL.md:36`, Critical Rule 2: "**Always wire the `completed` handle.** … Only `completed` is available as an output handle — **not** `output`, `success`, or any other name."
- `references/hitl-node-quickform.md:246,249,251` — the exact section `impl.md:9` links to — wires `"sourcePort": "completed"` and repeats "Always wire `completed`."

This PR proves that Critical Rule is wrong for every quick-form node with outcomes — and in the same commit flips the skill to `stable`. Either fix the sibling skill here, or drop the status change.

### 🟠 Major — sibling tasks still instruct and grade the old shape

`smoke_02` is not the only task pinned to `completed`. Untouched, in this skill's own suite:

- `hitl/quality_04_brownfield_insert.yaml:74` — `flow_contains.py --flow-name ContractReview 'completed'`. The corrected skill can never emit that substring, so this criterion now scores **0**. Integration tier, so the smoke gate will not catch it.
- `hitl/quality_01_schema_design.yaml:29` — "Wire completed → decision node." Both halves now wrong.
- `hitl/quality_02_result_downstream.yaml:29`, `hitl/quality_03_boolean_decision.yaml:28`, `hitl/smoke_03_multi_outcome_routing.yaml:29` — prompts still say `→|completed|`.
- `tests/tasks/uipath-human-in-the-loop/` — ~14 tasks plus `TEST_PLAN.md:39` ("Available handles: `completed` only").

### 🟠 Major — `impl.md` Option 2's example now produces the wrong port

Pre-existing, but this PR is what makes port naming schema-derived, so it lands here. `impl.md:129-132` gives the coded-action-app example an `inputs.schema.outcomes: [{id:"submit"}]` block. The product's manifest ships no `schema` for that node (`ootb-nodes/uipath-human-in-the-loop-coded-action-app/v1.0-v1.4.ts`, `inputDefaults` is `recipient` only), and `isHitlNodeType` includes `coded-action-app`, so `getEffectiveHitlHandleCustomization` reads that stray `inputs.schema` and flips the handle from `completed` to `outcome-submit`.

**Fix:** drop `inputs.schema` from the Option 2 example, and state that Option 2 keeps the static `completed` handle because its outcomes come from `inputs.app`.

### 🟡 Minor — the description overstates the failure mode

"the runtime still executed the flow as if it were connected to the next node" is the wrong half. `packages/canvas/src/utils/hitl-edge-reconciliation.ts` remaps a **lone** placeholder edge onto the primary outcome on load — its own comment calls that "the purely-cosmetic case where the connector was invisible but the workflow still ran correctly." The genuinely broken shapes are (a) multi-outcome, where only the primary gets remapped and the rest dangle, and (b) a placeholder edge duplicating a correct `outcome-<id>` edge, which the BPMN engine takes unconditionally alongside the conditioned one and **double-executes** the branch. Worth stating accurately — it is the stronger argument.

### 🟡 Minor — test plan claim is short of the repo bar

`.claude/rules/test-writing.md` § Workflow step 5 wants a passing-run claim for affected tasks. The last box is unchecked and the affected integration/e2e tasks (`quality_01-04`, `expense_approval_simulated`, `devcon_expense_approval`) were not run. The failing smoke gate on this PR is 4 Data Fabric tasks (`create-all-types`, `error`, `query`, `update`, 84.6%) — unrelated, confirmed from run 34487479174.

### 🟡 Minor — one logical change per PR

The `uipath-human-in-the-loop` → `stable` flip is unrelated to the wiring fix and rests on nightly scores that are not in the diff. `.claude/rules/pr-review.md` § Commit and Branch Conventions. Split it.

### What I verified vs. reasoned

- **Verified from source:** handle id format, the placeholder's zero-outcome condition, id used verbatim (no case folding), outcomes without an `id` getting no handle at all, per-outcome conditions in BPMN serialization, coded-action-app `inputDefaults`, legacy edge reconciliation.
- **Could not verify:** the "96-100% across the last several claude-sonnet-5 runs" nightly claim behind the `stable` flip. Not in the diff, no linked run. Reported, not confirmed.
- **Did not run:** the checker unit tests (read only, per repo convention). `maestro-flow checker unit tests` passes in CI, which is consistent with the findings below — the fixtures never exercise the broken paths.

---

## Recommendation

**Request changes.** The diagnosis is correct and well-sourced, but the skill now contradicts itself in six files and contradicts the sibling skill it delegates to, so an agent lands back on the old rule either way. Separately, both checkers lower-case the outcome `id` and fall back to `name`, neither of which the product does — they will fail correct work and pass invalid work. See inline.

---

**tl;dr** — Right bug, right root cause, verified against flow-workbench. But the fix touches 2 of 8 files that teach `completed`, leaves the sibling HITL skill teaching the opposite as a Critical Rule while marking it stable, and the new checkers lower-case the outcome `id` and invent a `name` fallback the product does not have. Request changes.

dushyant-uipath:
Addressed the review, point by point:

- 8 uipath-maestro-flow files teaching completed: all 8 now updated (2 in round 1, 6 more in round 2: editing-operations.md, planning-arch.md, brownfield.md, cli-commands.md, diagnose/failure-modes.md plus its anchor and 2 referrers).
- Sibling uipath-human-in-the-loop skill contradicting the fix while being marked stable: SKILL.md Critical Rule 2, hitl-node-quickform.md's Edge Wiring section, and hitl-node-apptask.md's "identical to QuickForm" claims all corrected. The stable flip is split into #3224, gated on the wiring fix merging first.
- Sibling tasks grading the old shape: quality_01-04 and smoke_03 in uipath-maestro-flow, e2e_01-06 and quality_04/05/07/08/09/10/11 in uipath-human-in-the-loop, plus TEST_PLAN.md, all updated.
- Option 2 (coded-action-app) example flipping its own port: the stray inputs.schema.outcomes block is removed; documented that Option 2 keeps a static completed port.
- Failure-mode description: corrected in the PR body to the lone-placeholder-cosmetic vs multi-outcome-dangling vs duplicate-edge-double-execution distinction.
- Checker id-derivation: reverted the name-fallback in both checkers; test_batch2_alignment.py's fixture now carries explicit outcome ids instead of relying on it. Confirmed against source (build-handle-customization.ts's outcome filter, and the editor's id-at-creation behavior) before touching this a second time.
- Test-plan claim: still open. There is no way to trigger a full coder-eval run against the affected integration/e2e tasks from here; that gap is called out explicitly as unchecked in the test plan rather than claimed.

Ready for another pass whenever you have time.

dushyant-uipath:
Added a Round 3 fix: skill-hitl-e2e-invoice-approval-greenfield kept timing out on an unrelated axis (real SharePoint/SAP connector discovery eating the whole turn budget before the agent ever touched the HITL node). Scaled the scenario down to mock both integration points as Script nodes instead of raising the timeout further, dropped the budget back to 1200/1200, and ran it locally to confirm: SUCCESS, score 1.000, 7/7 criteria, 482.7s.

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
dushyant-uipath: Fixed. All 6: editing-operations.md, planning-arch.md, brownfield.md, cli-commands.md, diagnose/failure-modes.md (heading, anchor, both referrers), diagnose/CAPABILITY.md.

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
dushyant-uipath: Fixed. SKILL.md:36 and hitl-node-quickform.md Edge Wiring section rewritten for the QuickForm vs App-based split. stable flip reverted here, split into #3224 (gated on this merging first).

skills/uipath-maestro-flow/references/author/plugins/hitl/impl.md
dushyant-uipath: Fixed. Reworded to the port for a zero-outcome node, or for a real outcome whose id is literally completed, in both the rule and the debug-table row. Added the no-id-no-handle clause.

skills/uipath-maestro-flow/references/author/plugins/hitl/planning.md
dushyant-uipath: Fixed. Same rewording applied here.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
dushyant-uipath: Fixed. Verbatim id, no name fallback, fails loudly on a missing id. Added the outcome-id conversion rule to hitl-node-quickform.md table too.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
dushyant-uipath: Fixed. Extracted assert_outcome_wiring(), gated on outcomes being non-empty. The boolean-decision/no-outcomes path now falls back to requiring the completed placeholder, matching pre-existing behavior.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
dushyant-uipath: Fixed. One full comment in check_simulated_hitl.py, short pointers from the other two, no date-based narration.

tests/tasks/uipath-maestro-flow/e2e/check_devcon_expense_approval.py
dushyant-uipath: Fixed. Same verbatim-id fix applied here.

tests/tasks/uipath-maestro-flow/_shared/test_batch2_alignment.py
dushyant-uipath: Fixed. Fixture now uses real ids, approve/reject, no fallback dependency. Also added test_rejects_outcome_without_id as the negative case.

tests/tasks/uipath-maestro-flow/e2e/test_check_devcon_expense_approval.py
dushyant-uipath: Added test_accepts_mixed_case_outcome_id, confirmed it fails without the verbatim fix and passes with it.

tests/tasks/uipath-maestro-flow/hitl/smoke_02_completed_port_wired.yaml
dushyant-uipath: Fixed both. Added check_outcome_wiring, node-scoped, resolves the HITL node own edges, handles 1+ nodes, and pointed this task at it instead of a whole-file regex. Added the outcome-id conversion rule to hitl-node-quickform.md so ids are derived, not guessed.

assets/skill-status.json
dushyant-uipath: Fixed. Reverted here, split into #3224 with SKILL.md:36 and hitl-node-quickform.md already corrected on this branch first.

dushyant-uipath:
Closed in favor of three smaller PRs: #3259 (docs), #3260 (checker logic), #3261 (task fixture migration, depends on #3260).

dmetzgar (approved):

skills/uipath-human-in-the-loop/references/hitl-node-quickform.md
```diff
@@ -243,13 +243,14 @@ Every `.flow` file must have one definition entry for `uipath.human-in-the-loop.
 
 ## Edge Wiring
 
-Wire the `completed` output handle to the downstream node. Edge ID format: `{sourceNodeId}-{sourcePort}-{targetNodeId}-{targetPort}` (append `-2`, `-3` on collision).
+Wire one output handle per outcome to its downstream node — port `outcome-<outcome.id>` for each entry in `inputs.schema.outcomes[]`, using the outcome's own `id` verbatim (never a lowercased or otherwise derived value). Edge ID format: `{sourceNodeId}-{sourcePort}-{targetNodeId}-{targetPort}` (append `-2`, `-3` on collision).
```
rockymadden, line 246: 🟠 **Major** — anchoring here; the finding is the interaction with `:218`.

The definition entry this file tells the agent to write declares exactly one source handle:

```json
{ "id": "completed", "label": "Completed", "type": "source", "handleType": "output", ... }
```

Thirty lines later this section says to wire `outcome-approve` / `outcome-reject`, and `:253` says never to wire `completed` once outcomes exist. Same file, and nothing reconciles them.

Both are correct — quick-form `v1.0`–`v1.6` really do ship a static `completed` in `handleConfiguration`, and the per-outcome handles come from `handleCustomization` at render/serialize time, never from the definition. But an agent reading them back to back will try to resolve the contradiction, most likely by adding `outcome-*` handles to `handleConfiguration` or by wiring `completed`. Both wrong.

**Fix:** one line under the definition block —

> The static `completed` handle here mirrors the manifest. It is not what you wire: per-outcome handles are derived from `inputs.schema.outcomes` at render time and never appear in `handleConfiguration`. See Edge Wiring below.

tests/tasks/uipath-human-in-the-loop/e2e_01_invoice_approval_greenfield.yaml
```diff
@@ -4,34 +4,46 @@ task_id: skill-hitl-e2e-invoice-approval-greenfield
...
 initial_prompt: |
-  Build a UiPath Flow named "InvoiceApproval" that extracts invoice data
-  from a SharePoint folder and posts the approved invoices to SAP. Finance
-  needs to review and approve each invoice before it is posted — we cannot
-  write to SAP without human sign-off.
+  Build a UiPath Flow named "InvoiceApproval" that extracts invoice data and
+  posts approved invoices to SAP. Finance needs to review and approve each
+  invoice before it is posted — we cannot write to SAP without human sign-off.
+
+  Mock the invoice extraction and the SAP posting as Script nodes — no real
```
rockymadden, line 40: 🟠 **Major** — this rewrites the scenario, not just the wiring criterion: SharePoint and SAP become Script mocks, `task_timeout`/`turn_timeout` drop 2400 → 1200 (partly reverting #3196), `expected_turns` 19 → 34.

Two problems:

1. **Unrelated to this PR.** Last round the `stable` flip got split out for the same reason (`.claude/rules/pr-review.md` § Commit and Branch Conventions). This is the same class, reappearing.
2. **Coverage loss.** The comment above states e2e_02/04/05 already mock their integration points. That makes e2e_01 the only e2e in this suite exercising live connector discovery — and this removes it, with nothing picking it up.

The timeout problem is real and pre-existing. But deciding a golden e2e should stop covering connector discovery is the scenario owner's call, not a side effect of a wiring fix.

**Fix:** split to its own PR, or get an explicit ack from whoever owns the scenario and say so in the body.

tests/tasks/uipath-human-in-the-loop/TEST_PLAN.md
```diff
@@ -120,12 +120,12 @@ The `variable` property creates a separate workflow-global variable (`$vars.appr
...
-| ✅ Present | [e2e_04_multi_hitl_brownfield.yaml](e2e_04_multi_hitl_brownfield.yaml) | `skill-hitl-e2e-multi-hitl-brownfield` | Inserts **two** HITL nodes into `HROnboarding` flow (doc review + IT access); both completed handles wired | 🟤 Brown |
-| ✅ Present | [e2e_05_expense_approval_brownfield.yaml](e2e_05_expense_approval_brownfield.yaml) | `skill-hitl-e2e-expense-approval-brownfield` | Inserts single HITL node between two existing nodes in minimal `ExpenseApproval` flow; wires `completed` | 🟤 Brown |
-| ✅ Present | [e2e_06_invoice_approval_greenfield_simple.yaml](e2e_06_invoice_approval_greenfield_simple.yaml) | `skill-hitl-e2e-invoice-approval-greenfield-simple` | Creates `InvoiceApproval` project from scratch (`uip solution new` + `flow init`); wires `completed`; validates | 🟢 Green |
+| ✅ Present | [e2e_01_invoice_approval_greenfield.yaml](e2e_01_invoice_approval_greenfield.yaml) | `skill-hitl-e2e-invoice-approval-greenfield` | SharePoint → HITL → SAP; full Discover→Plan→Build→Verify; wires both outcome ports, captures `$vars.output` | 🟤 Brown |
```
rockymadden, line 123: 🟡 **Minor** — row still reads `SharePoint → HITL → SAP; full Discover→Plan→Build→Verify`. Both halves are false as of this PR — the same commit mocks both integration points and drops the discovery loop.

**Fix:** `Mocked extract → HITL → mocked SAP post; wires both outcome ports, captures $vars.output`.

skills/uipath-maestro-flow/references/diagnose/failure-modes.md
```diff
@@ -129,19 +129,19 @@ Format:
...
 
 ### Symptom
 
-Flow execution reaches a HITL QuickForm node, the human task is created and completed, but the flow blocks indefinitely afterward. No further nodes execute.
+Flow execution reaches a HITL QuickForm node, the human task is created and completed, but the flow blocks indefinitely afterward on one or more branches. No further nodes execute on the affected branch.
 
 ### Cause
 
-The HITL node's `completed` output handle has no outgoing edge — there is no consumer for the `completed` port event.
+A QuickForm node has one output port per outcome, named `outcome-<outcome.id>`. `outcome-completed` is a separate, zero-outcome placeholder port that disappears the instant the schema has any real outcome — it is never a real branch once outcomes exist. Either one of the real outcome ports has no outgoing edge, or the flow was wired against the stale `outcome-completed` placeholder instead of the real per-outcome ports.
```
rockymadden, line 140: 🟡 **Minor** — "`outcome-completed` … **is never a real branch once outcomes exist**" is the over-claim `impl.md:69` and `planning.md:55` just dropped.

`outcomeHandleId('completed')` produces the same string as the placeholder, and both `hitl-outcome-handles-rule.ts` and `collectOutcomeConditions` resolve it against the node's real outcome ids rather than by string match. The checkers in this PR encode the correct version (`"completed" not in outcome_ids`).

Two other sites kept the absolute form: `uipath-human-in-the-loop/SKILL.md:36` and `hitl-node-quickform.md:253`.

**Fix:** same phrasing as `impl.md:69` in all three — the port for a zero-outcome node, or for a real outcome whose `id` is literally `completed`; never a shared exit for several outcomes.

tests/tasks/uipath-maestro-flow/e2e/check_devcon_expense_approval.py
```diff
@@ -117,12 +117,25 @@ def main() -> None:
             "or =js:$vars.<node>.output.<field>"
         )
 
-    if not any(
-        e.get("sourceNodeId") == hitl_id
-        and e.get("sourcePort") in ("completed", "outcome-completed")
-        for e in edges
-    ):
-        fail("HITL completed handle must be wired")
+    # Handle id is outcome-<id>, verbatim, per outcome: flow-workbench
```
rockymadden, line 120: 🟡 **Minor** — lines 120-138 are a verbatim copy of `assert_outcome_wiring` (`check_simulated_hitl.py:87-113`): same id extraction, same placeholder guard, same missing-ports assembly, same two messages. Two copies of one contract drift apart on the next change to it.

This file already dual-imports from `_shared` (`:16-23`), and `test_accepts_checker_frozen_without_sibling_shared_directory` pins the standalone-copy path via `PYTHONPATH`. The same dance works for `assert_outcome_wiring`.

**Fix:** import it, mirroring the `find_flow_file` block above.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
```diff
@@ -84,6 +84,55 @@ def check_priority(_flow: dict[str, Any], nodes: list[dict[str, Any]]) -> None:
...
+        if edge.get("sourceNodeId") == hitl_id
+    }
+    if outcomes:
+        # has_outcomes path: every outcome needs its own wired outcome-<id>
+        # port. Only the boolean-decision path (no outcomes at all) uses the
+        # zero-outcome outcome-completed placeholder.
+        outcome_ids = [o["id"] for o in outcomes if isinstance(o.get("id"), str) and o["id"]]
+        if len(outcome_ids) != len(outcomes):
+            fail("every outcome needs a non-empty string id — an outcome without one renders no handle")
+        if (wired_ports & {"completed", "outcome-completed"}) and "completed" not in outcome_ids:
```
rockymadden, line 105: 🟡 **Minor** — the `and "completed" not in outcome_ids` escape is the one branch that distinguishes this checker from the over-claim still in the docs, and it is the only branch either checker adds that no test covers. `test_rejects_outcome_completed_with_real_outcomes` exercises the guard firing, never the escape.

**Fix:**

```python
def test_accepts_outcome_literally_named_completed(tmp_path):
    outcomes = [{"id": "completed", "name": "Completed"}, {"id": "reject", "name": "Reject"}]
    # ports outcome-completed / outcome-reject must PASS
```

Cheap, and it locks the one rule the rest of the repo still states wrongly.

rockymadden (approved):
## What

Second pass. Last round I requested changes on 12 inline findings; all 12 are addressed. This is a re-review of the delta plus a re-verification of the core claim against `flow-workbench@develop`.

**The model still holds, and I re-confirmed it from source rather than from the replies:**

- `packages/canvas/src/hooks/useResolvedHandleNodes.ts` — `isHitlNodeType(nodeType) ? getEffectiveHitlHandleCustomization(node.data) : …`. **No version gate.** Any HITL node type at any `typeVersion` gets per-outcome handles once `inputs.schema.outcomes` has one valid-id entry. ✅
- `packages/canvas/src/validation/hitl-outcome-handles-rule.ts` — header: *"Per-outcome handles are always enabled; this rule always fires for HITL nodes."* Emits `HITL_DANGLING_EDGE` for a `completed`/`outcome-completed` edge once real outcomes exist. ✅
- `packages/services/src/serialization/hitl-serialization.ts` → `collectOutcomeConditions` — keys off the `outcome-` prefix and `o.id`, no version check. ✅
- `outcomePorts` **does not exist anywhere in flow-workbench@develop.** ✅

Against the tickets: this matches **UV-16499** (outcome port ids) and the HITL half of **UV-16428** (skill-doc contradictions). Neither is linked from the PR body — worth adding.

**Fix-by-fix verification of the 12:**

| # | Claim | Verdict |
|---|---|---|
| 1 | 6 doc files | ✅ all 6 + anchor rename + both referrers |
| 2 | Sibling skill | ✅ `SKILL.md:36`, `hitl-node-quickform.md:246-253`, `hitl-node-apptask.md:3,421` |
| 3,4 | `outcome-completed` over-claim | ⚠️ fixed in `impl.md:69` / `planning.md:55`, **not** in 3 other sites (inline) |
| 5,8 | Verbatim `id`, no `name` fallback | ✅ both checkers |
| 6 | `has_decision`-only path | ✅ `assert_outcome_wiring` elif keeps the old contract |
| 7 | Comment dedup | ✅ one full comment, pointers elsewhere, no dates |
| 9 | Fixture ids + negative case | ✅ + `test_rejects_outcome_without_id` |
| 10 | Mixed-case test | ✅ `test_accepts_mixed_case_outcome_id` |
| 11 | Node-scoped check | ✅ new `outcome-wiring` subcommand, replaces the file regex in 9 tasks |
| 12 | `stable` flip | ✅ reverted, split to #3224 |

Check ordering is fixed — the placeholder-misuse message now fires before the missing-ports one, and the `"completed" not in outcome_ids` escape correctly preserves the real-outcome-named-`completed` case.

---

## Overall findings

### 🟠 Major — #3239 asserts the opposite and touches the same two files

The author flagged this. My read of the source, for the record:

- Every `uipath-human-in-the-loop-quick-form` manifest **v1.0 → v1.6** declares one static source handle, `completed`. Base `uipath.human-in-the-loop` v1.1/v1.2 declare `outcome-completed` — a **placeholder**, per their own comment: *"Once `handleCustomization` is populated (one handle per outcome), it overrides this group entirely."*
- Neither declares `outcome-{item.id}` repeated over outcomes. That shape comes from the runtime `handleCustomization` override, not from a manifest version.

So #3239's premise — *"`outcomePorts: true` selects the base node's 1.1/1.2 definition, whose only source handle is `outcome-{item.id}`"* — is not what flow-workbench does. `outcomePorts` is a flow-builder-sdk concept (that PR cites flow-builder-sdk#718/#724), and the SDK's emit gate is a separate question from what the canvas renders and what the runtime routes on.

**Could not verify from source:** what `uip maestro flow validate` itself accepts — that validator lives in flow-builder-sdk, which I do not have. Two pieces of external evidence point the same way: smoke_02 requires `outcome-approve`/`outcome-reject` **and** exit 0 from validate, and reports SUCCESS 1.00; and UV-16499 states edges from `completed`/`outcome-completed` "validate and run as unconditional flows but never render" — i.e. validate accepts both shapes and rejects neither. That is consistent but not source-confirmed. Reporting it as unconfirmed.

**Recommendation:** merge this first, then rebase #3239's gate relaxation onto it. Accepting *either* `completed` or `outcome-<id>` would let a `HITL_DANGLING_EDGE` flow score 1.0 — the exact double-execution shape behind UV-16499.

### 🟡 Minor — `hitl-node-coded-action-app.md:322` missed the `inputs.schema` warning

`impl.md:155` and `hitl-node-apptask.md:3` both gained "do not add an `inputs.schema` block to an app-based node." The third file describing the same node type did not. Line 322 reads "wire `completed` (only handle available in v1.0)" — correct, but it is the one place a reader lands without the trap warning. File is not in the diff, so no inline. One clause fixes it.

### 🟡 Minor — 20 of 22 changed tasks unrun

Per the advisory lint and the PR body. `quality_01`–`quality_04` and `smoke_03` got substantive prompt rewrites (Decision-node guidance inverted), not just wording. `.claude/rules/test-writing.md` § Workflow step 5 wants a passing-run claim. Not a blocker given the checker unit tests are green and smoke_02 + e2e_01 both passed, but the rewrites are more than cosmetic.

### 🔵 Informational — branch carries a duplicate of a commit already on main

`a2f1e0f13` ("restore scaffold_multinode run_limits (#2572)") is byte-identical to `7f1773dbe` on `main`. Merge is a no-op, but it puts an unrelated file in the PR's file list. Rebase clears it.

---

## Recommendation

**Approve.** The diagnosis was right last round and the execution now matches it. All 12 findings are genuinely fixed, not papered over — the `name` fallback is gone, the invalid fixture has real ids, the mixed-case test fails without the fix, and the file-scoped regex is replaced by a node-scoped check across 9 tasks. The residue is doc consistency and one unrelated test rescope, none of which makes an agent do the wrong thing on the main path.

Six inline comments, all non-blocking.

---

**tl;dr** — All 12 prior findings fixed and verified, not just claimed. Re-confirmed from flow-workbench that per-outcome handles are ungated by version, which makes #3239's premise wrong; merge this first. Residue: the QuickForm definition block still declares only `completed` 30 lines above the rule saying never to wire it, 3 files kept the over-claim that `impl.md` dropped, and an unrelated e2e_01 rescope got bundled in. Approve.

dmetzgar:
Reconciling this against #3239, as your "needs reconciling before merge" note asks. **Your direction is right and #3239 is the weaker fix — I'm closing it in favour of this.** Three findings, all measured on the alpha tenant today, one of which corrects a line in your description.

## Your counter-measurement is correct, and here is the mechanism

You wrote that a `typeVersion: "1.0"` node wired with `outcome-*` edges validates clean. Reproduced. The reason is that `uip maestro flow validate` **special-cases the `outcome-` prefix** — same artifact, same embedded `uipath.human-in-the-loop@1.0` definition (which declares `input | completed` and nothing else), only the edge's `sourcePort` changed:

| edge on base@1.0 | `flow validate` |
| --- | --- |
| `completed` | Valid |
| `outcome-approve` + `outcome-reject` | **Valid** |
| `bogus-port` | error: `Edge references undeclared source handle "bogus-port"` |

A literal handle is checked against the definition; anything shaped `outcome-<id>` is waved through regardless of what the definition declares. So per-outcome wiring is legitimate on 1.0 too, and my framing in #3239 — that outcome ports are gated behind `outcomePorts` on 1.1+ — described the SDK's authoring surface, not the platform's contract. That distinction is what made our two measurements look contradictory when both were real: my `completed`-on-1.1 rejection is a literal-handle check, your `outcome-*`-on-1.0 pass is the prefix waiver.

The designer source you cite backs the direction: `getEffectiveHitlHandleCustomization` derives one `outcome-<id>` handle per schema outcome at render time and returns undefined only when there are none, "so the manifest's placeholder handle renders". Per-outcome is the live shape; `completed` is the zero-outcome placeholder, exactly as you say.

## One line in the description needs a correction

> QuickForm HITL nodes wire one `outcome-<outcome.id>` port per outcome

True of the canvas, not of the manifest. After `uip maestro flow registry pull --force`, this tenant serves:

```
uipath.human-in-the-loop.quick-form@1.1        handles: input, completed
uipath.human-in-the-loop.coded-action-app@1.1  handles: input, completed
```

Neither declares an outcome handle. The per-outcome handles come from the instance-level `handleCustomization` the canvas builds from `inputs.schema.outcomes` — which is what your own citation describes ("The HITL v1.1 manifest declares NO outcome-specific handles statically"). The conclusion holds; the stated reason doesn't, and it matters because it is the sentence a future reader will lean on.

## The consequence worth deciding before merge

Your gate, run against each authorable shape:

| shape | verdict |
| --- | --- |
| base@1.1 + both outcome ports (the flow from flow-builder-sdk#718) | **OK** |
| base@1.0 + `completed` (what the SDK emits by default) | FAIL |
| quick-form@1.0 + `completed` (what the v1 arm emitted in the last passing comparison run) | FAIL |
| base@1.1 + only the first outcome wired | FAIL — `missing: outcome-reject` |

The first and last are improvements — the last is a real bug my version missed, since a partially-routed outcome set leaves an outcome dangling at runtime.

Rows 2 and 3 are the thing to weigh. From the SDK arm, the **only** way to satisfy this gate is base + `outcomePorts: true`:

- the SDK refuses `outcomePorts` on quick-form and action-app outright — *"the 'quick-form' variant has no per-outcome version"*;
- its `check` refuses `outcome-*` ports on any node that has not opted in (`HITL_OUTCOME_PORTS_OFF`), even though `validate` accepts them.

So a quick-form HITL becomes ungradeable on this task. That is satisfiable overall — base + `outcomePorts` works — so it need not block you; but it moves the remaining gap onto the SDK, and it is worth saying so in the PR so the next person does not read row 3 as a regression. I am taking that side: a `hitl.md` follow-up, plus an SDK issue for the variants, since making them emit per-outcome handles needs a real decision (write `handleCustomization` on the node data as the canvas does, or wait for a per-outcome variant version) rather than a quick patch.

Also worth knowing for row 2's message: the product itself emits `[HITL_COMPLETED_UNWIRED] HITL node "review" has no outgoing edge. The runtime stalls indefinitely if no outcome is wired.` — its name says completed, its text says outcome. The platform's own vocabulary is mid-migration here, which is part of why the docs drifted.

## What I am doing with #3239

Closing it. Nothing in it survives that this PR does not do better: the `completed`-or-`outcome-*` leniency it added would keep blessing a shape the canvas does not render, and it has no coverage for the partially-routed case. The one piece I would suggest lifting is the fixture parameterisation — `_flow_doc` taking `node_type` / `type_version` / `exit_ports` — if your tests do not already cover the shapes by another route.

skills/uipath-human-in-the-loop/references/hitl-node-quickform.md
dushyant-uipath: Fixed in #3267, at the source: the sentence introducing the definition JSON now states upfront that handleConfiguration is fixed registration metadata and is unrelated to wiring, before the reader ever sees completed in it. No separate reconciling note needed once Edge Wiring's per-outcome rule stands alone.

tests/tasks/uipath-human-in-the-loop/e2e_01_invoice_approval_greenfield.yaml
dushyant-uipath: Addressed in #3267: added an explicit scope-change note to the task file and a new Coverage Gaps row in TEST_PLAN.md tracking the live-connector-discovery gap.

tests/tasks/uipath-human-in-the-loop/TEST_PLAN.md
dushyant-uipath: Fixed in #3267: row reworded to Mocked extract -> HITL -> mocked SAP post.

skills/uipath-maestro-flow/references/diagnose/failure-modes.md
dushyant-uipath: Fixed in #3267: corrected to the impl.md:69 phrasing in all three sites (SKILL.md, hitl-node-quickform.md, failure-modes.md).

tests/tasks/uipath-maestro-flow/e2e/check_devcon_expense_approval.py
dushyant-uipath: Fixed in #3267: check_devcon_expense_approval.py now imports assert_outcome_wiring from check_simulated_hitl.py instead of duplicating it, using the same dual-import fallback as find_flow_file.

tests/tasks/uipath-maestro-flow/_shared/check_simulated_hitl.py
dushyant-uipath: Fixed in #3267: added test_outcome_wiring_accepts_outcome_literally_named_completed covering the escape branch.

PR #3182, opened 2026-09-10
feat(uipath-insights): teach dashboard reads and create [IN-13531]

celestinryf:
https://uipath.atlassian.net/browse/IN-13531

Also related to https://uipath.atlassian.net/browse/IN-13526: the guides and `SKILL.md` edits here are that ticket's skill-content half.

Also related to https://uipath.atlassian.net/browse/IN-13527: the activation rows and the two eval tasks are that ticket's skills-side test layer.

The commands ship in [UiPath/cli#4124](https://github.com/UiPath/cli/pull/4124). This PR must not merge until that one reaches cli `main` and republishes `@uipath/cli@dev`, which the smoke gate installs.

## What

Teaches the `uipath-insights` skill the three dashboard commands: `dashboards get`, `dashboard-filters list`, and `dashboards create`.

- `references/dashboard-reads-guide.md` rewritten against the two shipped reads. The earlier draft described a larger read surface that did not ship.
- `references/dashboard-writes-guide.md` is new: the authoring order, two worked examples using real AO field ids, nine rules, and an error table whose last column answers the only question that matters after a failed create, whether the row exists.
- `SKILL.md` gains Critical Rule 17 (author from the `--output-file` artifact, never from redirected stdout) and Rule 18 (send a create once), three Scope Boundaries rows, and two Task Navigation rows. Front matter went from 594/927/1521 to 597/881/1478 against the 1,536 cap.
- Eight activation rows, ids 046 to 053.
- Two eval tasks under `tests/tasks/uipath-insights/dashboards/`: `smoke.yaml` for the reads and `create_smoke.yaml` for the authoring flow.

## Why

`dashboards create` is the first `uip insights` command that changes anything, and two of its failure modes are silent. A dashboard file assembled from redirected stdout is a PascalCase envelope the command refuses, and a create sent twice cannot be undone, because a process has one dashboard slot and no delete command exists. Rules 17 and 18 and the two eval tasks exist for those two.

## Verification

- `npm run skills:validate`, `skills:build`, `skills:check-links` (6,730 links), and `skills:test` (55 passing) are green. So are `check-skill-status.py`, `check-task-driver.py`, `check-skills-sh.py`, `check-cli-verbs.py` (0 High, 0 Medium against catalog 1.203.0-dev.8572), and `hooks/validate-skill-descriptions.sh`. Critical Rules read 1 through 18 with no gap, and the activation jsonl parses with 53 unique ids.
- Every eval regex was tested against both haystacks the grader matches, raw and whitespace-normalized, over eighteen command shapes including batched multi-line calls. The negatives bound their wildcard with `[^&;|\n]*` rather than the `(?!uip\s)` barrier the RBAC and alerts tasks use, because that barrier leaks into a batched neighbour and false-fails on a sibling's `jq --arg $KEY`.
- The three `run_command` reality gates were run against the source CLI on the cli branch: `dashboards get --help`, `dashboard-filters list --help`, and `dashboards create --help` all exit 0.
- Neither eval task has a passing agent run yet, and cannot have one until the cli PR merges: the commands are absent from `@uipath/cli@dev`, so the reality gates fail against the published build. Same position skills#2609 shipped in.
- The activation gate does not cover this skill. `tests/scripts/activation_gate.py --skill uipath-insights` exits 0 with `SKIP: no baseline for 'uipath-insights'`, so no recall number was produced for the `when_to_use` trim.

Stacked on #2854 (machine monitoring). The diff against that branch is the six files above.

Restacked on 2026-09-16. This PR declared `feat/uipath-insights-machine-monitoring` as its base while that branch was not an ancestor, so GitHub reported it `dirty` and its diff carried stale copies of the queue and machine guides, one of which taught `machines top-errors`, a verb that was rejected under TOOL_SPEC Rule 3 and will never ship. It now sits on the real machine tip and its diff is the six files it owns. Every `command_pattern` in the two dashboard tasks moved to the `(?:(?!uip\s)[^&;|\n])*` barrier for the reason #2839's commit message sets out. [#3358](https://github.com/UiPath/skills/pull/3358) stacks on this branch and teaches update, delete, and copy.


tests/tasks/uipath-insights/dashboards/smoke.yaml
```diff
@@ -0,0 +1,163 @@
...
+  tenant may or may not hold a Maestro process or a saved dashboard, and
+  the dashboard routes are served only on Cloud and Dedicated SaaS, so a
+  not-found, empty, or configuration answer is a valid outcome.
+
+  Covers both shipped reads. The prompt supplies a placeholder process key
+  as a fallback so the read path runs on a tenant with no Maestro process:
+  a well-formed GUID passes the CLI's local validation and reaches the
+  slot read, and the criteria grade the invocation rather than the tenant's
+  answer.
+tags: [uipath-insights, smoke, mode:diagnose, lifecycle:discover]
```
rohitjain-uipath, line 32: No `skip: true`, so the two `--help` gates (lines 99, 109) run against a CLI with no dashboard verbs and this PR's smoke leg goes red — 80%, 8/10, threshold 0.95.

`machines/smoke.yaml:45` already solves this for the same skill in the same situation: skipped tasks land in `RunSummary.skipped_tasks`, never reach the orchestrator, and stay out of the pass-rate denominator. Confirmed in this PR's own run — `skill-insights-smoke-machine-monitoring` is the one insights task with no container spawned.

With `skip: true` on both dashboard tasks the gate is 8/8. Suggest the same comment shape, naming UiPath/cli#4124 and the unskip probe.

tests/tasks/uipath-insights/dashboards/create_smoke.yaml
```diff
@@ -0,0 +1,191 @@
...
+  one shot, a pre_run step frees it before the agent starts; without that
+  the task passes once and fails forever, and an A/B of two variants fails
+  whichever runs second.
+
+  On the eval tenant the slot read answers not-found, because the tenant
+  has no Insights Portal service. The prompt therefore asks for the create
+  unconditionally: the CLI's own preflight refuses it before any request,
+  and the criteria grade the command shape, which is all this tier can
+  prove.
+tags: [uipath-insights, smoke, mode:build, lifecycle:setup]
```
rohitjain-uipath, line 30: Needs `skip: true` too — `dashboards create --help` (line 116) exits 3 on the published `@uipath/cli@dev`, so this task fails the smoke gate until the dev tag republishes.

skills/uipath-insights/references/dashboard-writes-guide.md
```diff
@@ -0,0 +1,125 @@
...
+2. Copy `tables` from that file verbatim. It carries the three AO tables (`PROCESSRUNS`, `ELEMENTRUNS`, `INCIDENTS`) with every field id the process can chart, custom variables included. Every chart dimension must be one of those `tables[].fields[].id` values, and every table must be one the data model has; the CLI checks both before sending and the renderer throws on either. Do not invent ids.
+3. Keep or extend `metrics`. A metric is `{ "id", "display", "expression": { "type": "aggregate", "argument": <field id>, "aggregation": <AVERAGE|COUNTDISTINCT|MAX|...> } }`, optionally with `filters`. Every `charts[][].metrics` entry must name a `metrics[].id`; the server rejects a missing reference. The `metrics` section must be present whenever any chart references one: the server skips its reference check when the section is absent, and the renderer then throws, so the CLI refuses that file.
```
rohitjain-uipath: The `expression` brace closes before "optionally with `filters`", so this reads as a metric-level key. `sanitizeMetric` allow-lists exactly three and drops the rest with no error:

```ts
// dashboard-guards.ts:348-357
return {
    id: requireNonEmptyString(record.id, ...),
    display: requireString(record.display, ...),
    expression: sanitizeMetricExpression(...),
};
```

`filters` is accepted only inside `expression` (`:327-338`). A metric authored per this line ships with its filter silently dropped — no error, no `Instructions` line. That's the failure class the rest of this guide exists to prevent.

Suggest: "…whose `expression` may optionally carry `filters`."

skills/uipath-insights/references/dashboard-reads-guide.md
```diff
@@ -0,0 +1,92 @@
...
+8. **A 500 from `dashboards get` is not a retry.** The stored definition may not parse, or the dashboard was created with an empty `tables` list and the deployment has no model blob to inject tables from. Both are properties of the stored row, so the CLI sets `Retry: RetryWillNotFix` on this branch instead of the 5xx default. Say that both causes are possible and that the response cannot tell them apart.
+9. **A 404 is not automatically an unsupported deployment.** `dashboards get` has real data 404s (no dashboard with that numeric id, no cached template for a template-keyed id, a saved row whose stored JSON no longer deserializes on a slot read, or an unregistered custom source type), so a 404 there is one of those or the deployment gate, and the envelope names the causes it can tell apart. A process-key read of a free slot returns the template, except when the deployment has no cached template for that slot's prefix, which is a 404 whose `Instructions` say so. Only `dashboard-filters list` under a built-in source type reads its 404 as the gate alone (`Result: ConfigError`).
```
rohitjain-uipath: The discriminator is the response body's `errorLabel`, not the source type:

```ts
// dashboards.ts:616
if (error.bodyHint?.errorLabel === MODEL_NOT_FOUND_LABEL) { writeFiltersNotFound(...) }
return claimGatedSurface404(DASHBOARDS_FAMILY)(error);   // → ConfigError
```

So a **custom** source type on a gated deployment answers `ConfigError`, not `not_found` — the CLI's own test says so: `"should claim a label-less 404 under a custom source type as the feature gate"` (`dashboard-filters.spec.ts:399`). The comment at `dashboards.ts:610-618` warns that branching on built-in vs custom "gets that backwards."

This bites exactly where Rule 10 is meant to help: on Automation Suite / Service Fabric with a custom source type, the guide predicts "that model isn't registered" when the CLI actually reports "this deployment doesn't serve dashboards."

Same correction needed at lines 36, 46, and 47.

rohitjain-uipath (commented):
Three must-fix items from a source-level pass against cli `bd09a8261`. Everything else I checked — flags, exit codes, the four `unknown_error` cases, chart types, layout guards, the regex work — matched source.

tests/tasks/uipath-insights/dashboards/smoke.yaml
celestinryf: Moot since the cli side merged. UiPath/cli#4124 landed on 2026-09-21 and `@uipath/cli@dev` 1.204.0-dev.8771 registers the dashboard verbs, so the two `--help` gates exit 0 and the task runs live: 12/12 at 1.000 on runs 35694315463 and 35721927455. The task also grew in https://github.com/UiPath/skills/pull/3182/commits/f4e7004e5: its slot is seeded per run so the by-id read and `--case` are gated too, and it ends with a live read of the slot.

tests/tasks/uipath-insights/dashboards/create_smoke.yaml
celestinryf: Same as above: `dashboards create --help` exits 0 on the published dev CLI now, and the task passes live (9/9 at 1.000 on runs 35694315463 and 35721927455). It ends with a live read that the create landed, and https://github.com/UiPath/skills/pull/3182/commits/f4e7004e5 moved it to a fresh key per run after two overlapping gate runs collided on the fixed one.

skills/uipath-insights/references/dashboard-writes-guide.md
celestinryf: Fixed in https://github.com/UiPath/skills/pull/3182/commits/aa157d171. Step 3 now shows `filters` inside `expression` and says that a metric-level `filters` is dropped without an error, citing `sanitizeMetric` in `dashboard-guards.ts`. Checked against cli main: the metric sanitizer keeps only `id`, `display`, `expression`, and `filters` is read at `:330` inside the expression sanitizer.

skills/uipath-insights/references/dashboard-reads-guide.md
celestinryf: Fixed in https://github.com/UiPath/skills/pull/3182/commits/aa157d171. Rules 9 and 10 and the two error rows now say the filters-list 404 is read off the body's "Dashboard not found" label, not off the source type: with the label it is `not_found` (and under a custom source type the likely cause is an unregistered model), without it the deployment gate answers `ConfigError` under any source type. Checked against `utils/dashboards.ts:762` on cli main, whose comment says the source-type branch "gets that backwards".

rohitjain-uipath (approved):

PR #3183, opened 2026-09-10
fix(uipath-planner): runnable gate command + complete cell contract for case SDDs

abhiram-vad:
> **Branch history note.** The first four commits fixed the right defects in the wrong places — a tool-call
> counter the agent has to self-audit, a new check inside `audit_sdd.py`, and edits to § Resumption.
> `ea266e753` reverts all of that and re-lands the same fixes as prose. **Read `ea266e753` alone, or the
> net diff against `main`** — the intermediate commits are not the proposal. Happy to squash to one commit.

## The defect

Seven case eval runs, five of them scoring `0.000` without producing a gradeable artifact. Tool execution
across all seven totals **0.3–3.1 seconds each**, against walls of 178–7204s — the time went into
generation, not tools. What the runs have in common:

| Run | `find`/`ls` to locate `audit_sdd.py` | Read its source | Ran it | Outcome |
|---|---|---|---|---|
| `reject-route` | 1 | **25** | 0 | ERROR 0.000 |
| `finalize-draft` | 1 | 1 | 0 | ERROR 0.000 |
| `finalize-draft-loan` | 2 | 3 | 0 | TIMEOUT 0.000 |
| `procurement-sla-interrupts` | 2 | 1 | 0 | TIMEOUT 0.000 |
| `case-reasoning-regressions` | 1 | 1 | 0 | ERROR 0.000 |
| `sla-sdd-response-map` | 1 | 0 | **3** | FAILURE 0.250 |
| `case-reasoning-regressions` (terra) | 0 | 0 | **2** | FAILURE 0.125 |

Every run that opened the validator's source scored 0.000. Both runs that only executed it produced a
graded artifact. Three doc defects put them there.

### 1. The gate command was not runnable as written

`case-sdd-template.md` said:

```
Gate: run  <py> "<skill folder>/scripts/case/audit_sdd.py" <sdd path> [--draft <draft path>]
`<py>` = the first of `python3`, `python`, `py` that runs
```

Two unresolved placeholders, and an explicit instruction to probe for the interpreter. Five of seven runs
opened by resolving both at once — `ls scripts/case/ 2>&1; which python3 python py 2>&1`,
`find … -iname "audit_sdd.py" 2>/dev/null; python3 --version` — and that locate call is what put the file
in hand. Meanwhile `case-design-lane-guide.md:13` says *"never hunt with `find` / global `ls`"*: the first
rung was mandated and forbidden at the same time.

### 2. The contract had a dangling end

The footer named nine check families and then deferred: *"the rules themselves live at the cells above and
in `case-design-layers-guide.md`"* — a file draft finalization may not read. So a cell like
`**HITL Implementation:**`, which appears **once** in the whole skill, resolved to nowhere.
`finalize-draft` said so at turn 26:

> *"I need to check whether 'JSON Schema' is a legal value or whether I must normalize to Action App shape,
> and where `deploymentTitle` comes from since draft never named an Action App."*

Fifteen more Greps followed, and no `Write`.

### 3. The SLA both-statuses rule was written at the wrong level

`case-design-layers-guide.md` said *"No stated response → both statuses `notify-only`"* — phrased per SLA,
while the contract is per status. An SLA with a stated at-risk response read as exempt, so
`sla-sdd-response-map` authored the At-Risk row only, with an HTML comment justifying the omission, and
failed the graded contract:

> *"**Triage SLA (4h):** at 80% elapsed, notify-only email to `UserGroup:TriageTeam` — **no breach row
> authored, since ops asked for nothing beyond the email**"*

## The change

Two files. Prose only — no script change, no lane-guide change.

**`assets/templates/case/case-sdd-template.md`**

- The gate carries one literal command anchored on the skill base dir given at invocation. The interpreter
  ladder is demoted to a command-not-found fallback, and the text names the locate call as the first rung
  of the source-reading it forbids.
- The footer states the cell rules above are **COMPLETE** for the render contract — no third location, and
  the script's source states nothing they do not. Design semantics stay explicitly with the layers guide.
- A header block states the invariant that makes single-mention cells terminate: a cell named once is fully
  specified; still-unstated means take the default or `—`, note provenance, move on.
- Three cells closed: `HITL Implementation` (`Action App: <title>` is its only form), `Execution autonomy`
  (carry the draft's value, else `autonomous`), case `Priority` (optional, and disambiguated from a task's
  closed-set `**Priority:**`).
- § SLA Response Map restated at status level.

**`references/case/case-design-layers-guide.md`**

- § Defaults when the source is silent: every (scope, SLA) authors both statuses; a stated response for one
  never removes the other's row.
- The picker-vs-decision repair lands beside the rule it belongs to (*"`user-selected-stage` is picker
  exposure … never deterministic routing"*), diagnosis first: **a person launches it** keeps the lane and
  adds the missing `wait-for-user` half; **a decision routes it** re-keys the lane on the decision fact and
  spells out all four required edits. Source silent defaults to person-launched, because re-keying on an
  unstated fact invents a business rule.

## Deliberately not done

- **`scripts/case/audit_sdd.py` is untouched.** The consequence is stated rather than hidden: the shipped
  auditor stays blind to § SLA Response Map, so a stated at-risk response with an omitted breach row will
  still print `AUDIT OK` on a document the grader rejects. The rule is in the template where the agent
  reads it; nothing enforces it.
- **§ Resumption is untouched.** No tool-call counter, no read budget, no self-audited call index.
- The 20 advisory `command_not_executed` criteria in the five task YAMLs stay. They are eval-side only —
  never read by the agent — and they are the measurement that says whether removing the placeholder removed
  the reason to go near the script.

## Evidence

| Gate | Result |
|---|---|
| `pytest tests/scripts` | 238 passed, 1 failed |
| `pytest tests/tasks/uipath-planner` | 89 passed |
| `python3 scripts/check-skill-status.py` | `OK — 27 skills, manifest valid.` |
| `python3 scripts/check-skills-sh.py` | `OK — 27 skills grouped across 4 section(s).` |

The single failure is `test_stage_shared.py::test_stage_shared_anchors_to_checkout_and_stages_nested_checkers`.
It reproduces identically on a clean `main` worktree — verified before claiming it unrelated, not asserted.

**Coder eval in flight:** [run 34448866335](https://github.com/UiPath/skills/actions/runs/34448866335) —
the five planner case tasks plus the two `uipath-maestro-case` design-delegation tasks, `claude`, `j=4`.
No result claimed until it lands. The tell is `commands_by_tool` and the index of the first `Write`: if
`find`/`which` still open the runs and the first Write still lands past call #20, the placeholder fix did
not do the work.

## Review notes

- No skill folder added, renamed, or removed — `assets/skill-status.json` and `skills.sh.json` need no edit.
- No `SKILL.md` frontmatter change, so `activation-gate.yml`'s recall-eval gate is not triggered.
- No hook changes, so no `.sh` / `.ps1` twin obligations.
- `CODEOWNERS` already covers both touched paths (`/skills/uipath-planner/assets/templates/case/` and
  `/skills/uipath-planner/references/case/`).


skills/uipath-planner/assets/templates/case/case-sdd-template.md
```diff
@@ -607,17 +616,27 @@ there. Pass --draft so the gate also checks inventory parity, verbatim `=js:` pr
 executable threshold encoding. What the DRAFT specifically needs repaired on the way through is the
 lane guide's § Resumption; the shape is here.
 
-Gate: run  <py> "<skill folder>/scripts/case/audit_sdd.py" <sdd path> [--draft <draft path>]  on the
-on-disk file BEFORE the Status: ready flip — in every mode. `<py>` = the first of `python3`, `python`,
-`py` that runs (Windows usually has no `python3` alias); only if all three are absent verify manually. RUN it, never open the script source —
+Gate: run this on the on-disk file BEFORE the Status: ready flip — in every mode:
+
+    python3 "<skill base dir>/scripts/case/audit_sdd.py" <sdd path> [--draft <draft path>]
```
RaduAna-Maria, line 621: I though this should no longer be a python script

skills/uipath-planner/references/case/case-design-layers-guide.md
```diff
@@ -144,6 +144,17 @@ The case, each stage, and each task move through gates driven by **rules** in di
...
+  3. **Origin completion exit** — its guard becomes the COMPLEMENT, `=js:(vars.<decisionVar> !== "<Value>")`. Unguarded, it fires on the diverted case too; repeating the affirmative guard on both rows fires both. With more than one diverted outcome the complement excludes every diverted value.
+  4. **Orphaned picker exposure** — once no `user-selected-stage` entry remains anywhere, DROP any upstream `wait-for-user` exit that existed only to expose this lane; the pairing above fails a `wait-for-user` with no picker entry.
+
+  A sibling outcome of the same decision already keyed this way is the shape to copy verbatim — the approve branch usually is. Source silent on who launches the lane ⟹ treat it as person-launched and keep what is authored: re-keying on an unstated fact invents a business rule.
+
```
RaduAna-Maria, line 157: This section seems more like agent's thoughts rather then instructions.  Please rephrase this in a shorter way. I would also like to raise my concerns again that the planner should not know about the case semantics in this depth, that's way the case skill exists. Please take this into consideration for future work.

abhiram-vad:
Superseded by #3057, which now carries this PR's substance plus the review comments from both threads here.

What moved across:

- **@RaduAna-Maria's first comment** — *"I though this should no longer be a python script"* — is resolved: `scripts/case/audit_sdd.py` is deleted in #3057 and the gate is a checklist the lane answers in the transcript. No script, no interpreter ladder.
- **@RaduAna-Maria's second comment** — the `user-selected-stage` section reading as agent's thoughts — is addressed: it is now a three-row diagnosis table (what the source says → the defect → the repair) with the four numbered edits kept intact. The depth concern about the planner holding case semantics is being treated as the follow-up you framed it as.
- The runnable gate command, the complete cell contract, and the SLA both-status rule are all in #3057.

Verification on the superseding stack — three identical replicate runs of the seven planner case evals: mean 0.965 / 0.968 / **1.000**, zero timeouts across 21 task-executions. The seven-run failure pattern that opened this PR (five runs at 0.000 with no gradeable artifact) is gone.

Closing rather than rebasing because the work now sits in a reviewed stack: #3057 → #3456 → #3457.

PR #3194, opened 2026-09-10
fix(uipath-maestro-flow): drop the feature flags from the OOTB node docs

uipath-utkarsh:
Follow-up to #3041 

They shouldn't be. Three reasons, and the third means the docs are now actively wrong.

## 1. An agent cannot act on them

No `uip` command reads, sets, or reports a `canvas.nodes.*` flag. The only thing an agent can observe is whether `registry get` resolves. So the flag tables were unactionable detail sitting in docs whose entire job is to say what to run next.

## 2. They were mislabelled as *tenant* flags, and that misattribution cost real time

They are flow-core `ManifestClient` flags — note the `canvas.` prefix. The CLI never asks the tenant about them: it answers `getFeatureFlag` from a hardcoded allowlist, `ENABLED_MANIFEST_FLAGS` in `node-sync-service.ts`.

So `planning.md`'s "a tenant can have Read without Create" was never true through the CLI, and `impl.md` shipped a column literally headed **"Flag to ask the admin about"** — advice that could not work, because no admin setting was in the path. That is exactly how this was first misdiagnosed: the CLI said "not enabled for this tenant", people went and checked the tenant, and the tenant was never involved.

## 3. They are now wrong

UiPath/cli#4087 (merged 2026-09-09) added all four entity flags to that allowlist. `registry get core.datafabric.*` resolves for all four and `registry search` reports `AvailableOnTenant: true` — while these docs were still telling agents to build with the `uipath-uipath-dataservice` connector instead.

## Same defect in two more families

Summarize and Batch Transform claimed the node *"appears after server rollout to the tenant"*. That model is wrong for any OOTB node: `getManifest` builds `nodes: this.getBaseNodeList()` — the local, flag-filtered OOTB set — and then only **pushes** server-fetched dynamic nodes on top. The server never withholds an OOTB node.

Verified both resolve today:

```
uipath.pattern.deep-rag        -> NodeGetSuccess, v1.0
uipath.pattern.batch-transform -> NodeGetSuccess, v1.0
```

## What every doc says now

Probe with `registry get`, treat the answer as a property of the **CLI build**, fall back to the connector on "Node not found".

That is deliberately version-robust, though not for the reason this description first gave. I had checked the **git tag** (`v1.201.0`) and concluded #4087 had not reached the field. The git tag is not what users install: `npm view @uipath/cli dist-tags.latest` is `1.203.0-dev.8614`, published after #4087 merged, so **anyone on `latest` or running `uip tools update` has the fix today**. Older pinned installs still answer "Node not found", so the probe-and-fall-back shape is still the right one — but it is version-robust for the long tail, not because the fix is unreleased. Nothing in the diff changes; the claim was wrong and is corrected here.

## Routing policy

The agreed policy for the Coding Agent, which this PR is the skills-side implementation of:

> 1. Native nodes remain the default for **CRUD record activities** wherever they are available natively in Flow.
> 2. For **other activities**, create connector/DAP activities.
> 3. If the user **explicitly specifies** connector activities, create them as long as connector support is available.

So the routing is decided by the *operation*, not by availability alone. Evaluated in this order, because rule 3 has to be able to override rule 1:

1. **The user named the connector** → build the connector activity, if it exists for that operation. An explicit instruction outranks the default; do not silently substitute the native node.
2. **Record CRUD (read / create / update / delete)** → default to the native `core.datafabric.*` node wherever Flow carries it natively.
3. **Any other Data Service operation** → the connector. Only those four exist natively, so for attachments, file-field downloads, entity metadata and bulk work the connector is the **only** path, not a fallback.

Federated entities go to the connector regardless, since the native writes require a native entity.

Where each rule is encoded — all four agree, so a planner and an implementer reach the same answer:

| Doc | Carries |
| --- | --- |
| `references/author/planning-arch.md` — Selecting External Service Nodes | The canonical three-way split, as the stated exception to the connector-first ladder |
| `references/author/plugins/data-fabric/planning.md` — Native node vs Data Service connector | The same rules as an ordered list, plus the federated-entity case and the rationale for native |
| `references/author/plugins/connector/impl.md` | The inverse view — a "stay here" callout listing the four conditions under which the connector is correct, including the explicit-request one |
| `references/author/CAPABILITY.md` | The capability-table row and both plugin-directory entries, so the routing is visible at the entry point |

Rule 2 is what the flag removal unblocked: while the docs claimed the native nodes were tenant-gated and off by default, connector-first was the only defensible recommendation. Rule 3 is new — nothing previously said an explicit user request should win. Rule 1's "other activities" case was also missing entirely: the old text framed the connector purely as a fallback for when native was unavailable, which is wrong for every operation that has no native node at all.

## Review follow-ups

- **One recovery, not three.** `planning-arch.md`, `data-fabric/planning.md` and `data-fabric/impl.md` each gave a different procedure for "Node not found". `impl.md — Registry validation` is now the single numbered procedure and the other two defer to it.
- **The sentence that over-claimed.** "No tenant setting behind this" is true of *registry resolution* and is now scoped to it — because there is a platform-side failure mode it was masking: on a BPMN engine predating the create postprocessor, `GetDataFabricAction()` falls back to `"update"`, so a native create can run green and insert nothing. Added as a Debug row.
- **Bounded remedy.** `cli-commands.md` now caps the fix at one `uip tools update`. Around ten of the 25 manifest flags sit outside the CLI's requested set by design (`agent-memory`, `queue-operations`, `form-trigger`, `http-standalone`, `classify-document`, `do-while`, `hitl-document`, …), so looping the upgrade can never surface them.
- **Native eval coverage.** Every task under `connector_features/datafabric_connector/` is pinned to the connector, which left the new default ungraded. Added `tests/tasks/uipath-maestro-flow/node_features/datafabric_native/` — a read → create chain checked for native node instances, the right entity, no connector substitute, and `definitions[]` that look registry-sourced (`bpmn:Task` + the `api-function` exclusion) rather than hand-assembled. A `pre_run` probe lands an old CLI as `ERROR` instead of scoring the correct connector fallback as a skill `FAILURE`.
- **The eval pins are the policy, not a contradiction** — but they now live in #3195, not here. Pinning those prompts to the connector *is* rule 1: they grade the connector family on purpose. #3195 pins the same 7 prompts, and the two PRs merge **without a git conflict** (verified against `9a98bfa`) because the two pins sit in different parts of `initial_prompt` — so both would have landed and each prompt would carry the instruction twice in two wordings. Dropped mine; those 7 are reverted to `origin/main` exactly. `trigger_lifecycle.yaml` stays here because #3195 does not touch it. `smoke_activation_positive` keeps only its tag change (`connector` / `uipath-uipath-dataservice` dropped — it grades routing only and its prompt now routes native). Three tasks are deliberately untouched: `smoke_activation_negative` (routing only), `smoke_update_existing_flow` (brownfield edit of a scaffolded connector activity), `smoke_file_activities` (names `DownloadFileFromRecordFieldV2` and friends explicitly, and file fields have no native node).

## Federated writes are not a routing choice

@AditiGoyalUipath flagged that the "stay here when the entity is federated" bullet did not belong in `connector/impl.md`, because a federated-entity write is blocked outright. That makes the claim not merely misplaced but **wrong**: it sent an agent to build a connector activity that also cannot work. It appeared three times and all three are corrected:

| Where | Was | Now |
| --- | --- | --- |
| `connector/impl.md` | "stay here … the entity is **federated**" | removed |
| `data-fabric/planning.md` routing line (added by this PR) | "go to the connector regardless when the entity is federated" | writing a federated entity is blocked; the connector is not an alternative path |
| `data-fabric/planning.md` selection heuristics (**pre-existing**) | "these nodes cannot; use connector or http" | the connector is not a way round it — write to the source system, via its own connector or http |

The third predates this PR; called out on the thread so it can be split back out if preferred. I could not verify the enforcement mechanism myself — no hits in `@uipath/flow-core`'s `manifest.js`, and the CLI's only federated surface is `--native-only` on `df entities list` — so this is service-side, consistent with the repo already saying the write "is rejected by Data Fabric, and the rejection is swallowed".

## Highest-leverage line

`references/shared/cli-commands.md` defined `AvailableOnTenant: false` as "the node is not enabled or available for the tenant". That generic definition feeds every node family, and around ten families are still absent from the CLI allowlist by design (agent-memory, queue-operations, form-trigger, http-standalone, agent-tool-http-request, patterns, classify-document, do-while, escalation.quick-form, hitl-document), so this misdiagnosis was set to repeat. It now says the type is missing from the manifest this CLI pulled, and that despite the name it is not a tenant entitlement.

## Verification

`npm run skills:validate` (27 skills, 1768 files) and `npm run skills:check-links` (6785 links) both pass.

`pytest tests/scripts/test_check_task_driver.py tests/scripts/test_runtime_payload_key_casing.py` — 15 passed, including `test_live_corpus_passes_both_gates`, which walks the whole live task corpus and therefore covers the new task YAML.

`check-skill-links.mjs` strips the fragment before resolving (`raw.split("#")[0]`), so it does not verify anchors. I checked the 400 anchored links separately: the 5 this PR introduces or renames all resolve, and the 17 that do not are all in files this PR does not touch (`plugins/http/*`, `shared/variables-and-expressions.md`, `evaluate/*`) — pre-existing.

The new check script was exercised against synthetic `.flow` fixtures on all five paths: native read+create passes; hand-written definition (`bpmn:ServiceTask`, missing exclusion) fails; connector substitute alongside fails; wrong entity fails; connector-only fails with the type list.

Two runs are **not** green, both pre-existing and both verified identical with this branch's changes stashed on clean `origin/main`:

- `npm run skills:test` — `pack builds complete, marker-free default and custom npm packages` asserts on `Publishing to https://npm.pkg.github.com/`; needs registry auth locally.
- `pytest tests/` full collection — 9 collection errors (`_shared.flow_check` import resolution, and `test_verb_checkers.py` on `sre_parse` internals). None are in files this PR touches.

**Not verified:** I did not run `registry get` against a live tenant, so the field behaviour of a released `uip` is still unconfirmed by me directly.

No longer docs-only: this adds an eval task, a check script and a preflight probe under `tests/`. Still no skill behavior, frontmatter, or flavor boundaries touched.

**Merge order:** #3195 owns the connector pins for 7 of the Data Fabric tasks. It stands alone and should merge on its own merits (turn-budget raises for the MAX_TURNS failures, plus the `pathParameters` placeholder note and the `IsNull`/`IsNotNull` operator row — none duplicated here). If it is abandoned, those 7 pins need restoring in this PR.


skills/uipath-maestro-flow/references/author/planning-arch.md
```diff
@@ -174,7 +174,7 @@ Prefer, in order:
...
+**Data Fabric is not on this ladder.** Entity records have two paths — the native `core.datafabric.*` nodes and the `uipath-uipath-dataservice` connector activities. Prefer the native nodes: they need no Integration Service connection. One probe decides — **`registry get core.datafabric.<op>` returning `NodeGetSuccess` means build native; "Node not found" means this CLI does not carry the node, so build with the connector activities** and never hand-write a `definitions[]` entry for a node the registry will not return. Rationale and the federated-entity case in [data-fabric/planning.md — Native node vs Data Service connector](plugins/data-fabric/planning.md#native-node-vs-data-service-connector--availability-decides).
```
rockymadden: 🔴 **major** — "One probe decides" now contradicts both docs this line routes to.

- **here:** probe → `Node not found` → connector. No retry.
- **`plugins/data-fabric/planning.md:24`:** → `uip tools update`, **or** connector.
- **`plugins/data-fabric/impl.md:32`:** → `uip tools update` + `registry pull --force` + retry once, **then** connector.

Three answers to one error. Before this PR the line was unambiguous ("do not retry, do not `uip tools update`"), so the planner had exactly one procedure. This is the same misdiagnosis class the PR exists to remove.

**Fix:** make `impl.md — Registry validation` the single procedure and have both planning docs defer to it. That also restores the `impl.md#registry-validation` link this change's counterpart deleted from `data-fabric/planning.md:24` — planners now never see the `--force` step.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
```diff
@@ -400,7 +393,7 @@ Use `uip df entities get` and `uip df records list` to close that gap before shi
...
+| `Node not found: core.datafabric.*` on `registry get` | This CLI build does not carry the node | `uip tools update`, then `uip maestro flow registry pull --force`; if it still fails, use the connector — no tenant setting governs this |
 | Node validates clean, runs green, nothing written | Most often a **selector** problem, not a binding one: `readEntityNodeId` names a missing node or a multi-record read, the read's filters do not compile, or the `fromRead` read matched more than one record at runtime | Check the Read node's `id` matches exactly and its `resultMode` is `single`; confirm the filter identifies exactly one record with `uip df records list` |
 | Write runs green, row unchanged | The body was rejected and the rejection swallowed — a federated entity, a system or attachment column, a choice-set label instead of its numeric id, an uncoercible value, or a null into a non-nullable column | Re-check the entity is native and each column against `uip df entities get` |
```
rockymadden, line 405: 🔴 **major** — now that native is the default, this row is missing its platform-side cause.

`flow-core`'s `core.datafabric.create` manifest states the insert runs in the BPMN engine's activity postprocessor and *"`GetDataFabricAction()` falls back to `\"update\"` on older engines."* So a native create can run green and silently update-or-nothing on an older engine.

That matters here specifically because line 32 now tells the agent *"There is no tenant setting behind this and no administrator to escalate to."* True for **registry resolution**; not true for the runtime engine. An agent hitting a swallowed create will read that line and rule out the platform entirely.

**Fix:** add a row —

| Create runs green, no row inserted | BPMN engine predates the create postprocessor; `GetDataFabricAction()` falls back to `"update"` | Confirm against a newer engine, or use the connector's Create Entity Record |

— and scope the line 32 sentence to registry resolution: "no tenant setting governs **whether the registry serves this node**".

skills/uipath-maestro-flow/references/shared/cli-commands.md
```diff
@@ -266,7 +266,7 @@ The cache expires after 30 minutes. `registry search` returns a flat `Data` arra
...
 ```
 
-Treat `AvailableOnTenant` as a usability gate: `true` permits `registry get <NodeType>` or `node add <NodeType>`; `false` means the node is not enabled or available for the tenant. Do not use unsupported flags such as `--include-unavailable`; choose an enabled alternative, use `--local` for in-solution resources, or report unavailability.
+Treat `AvailableOnTenant` as a usability gate: `true` permits `registry get <NodeType>` or `node add <NodeType>`; `false` means the type is missing from the manifest this CLI pulled. Despite the name it is not a tenant entitlement — the CLI asks for a fixed set of node manifests decided by its own build, so `false` usually means this CLI does not carry the node rather than that an administrator withheld it. Do not use unsupported flags such as `--include-unavailable`; try `uip tools update`, choose an available alternative, use `--local` for in-solution resources, or report it as unavailable.
```
rockymadden: 🟡 **minor** — correct rewrite, but the remedy list leads with `uip tools update`.

~12 of the 27 `MANIFEST_FEATURE_FLAGS` are absent from `ENABLED_MANIFEST_FLAGS` **by design**: `agent-memory`, `queue-operations`, `queue-trigger`, `queue-performer`, `form-trigger`, `http-standalone`, `agent-tool-http-request`, `patterns`, `classify-document`, `do-while`, `escalation.quick-form`, `hitl-document`. No upgrade reaches those — the agent burns a `tools update` + `registry pull` per attempt. Cheaper than the admin round-trip this PR kills, but the same shape, and this is the generic definition that feeds every node family.

**Fix:** cap it — "try `uip tools update` **once**; if the type is still absent it is absent by design for this build, so choose an available alternative or report it." Mirrors what `plugins/data-fabric/impl.md:36` already says.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
```diff
@@ -13,16 +13,15 @@ Use them whenever the flow's own data lives in Data Fabric: a case record, a loo
...
 
-| Node Type | Tenant flag |
-| --- | --- |
-| `core.datafabric.read` | `canvas.nodes.read-entity` |
-| `core.datafabric.create` | `canvas.nodes.create-entity` |
-| `core.datafabric.update` | `canvas.nodes.update-entity` |
-| `core.datafabric.delete` | `canvas.nodes.delete-entity` |
+Whether your CLI can serve them is a property of the CLI build, not of the tenant, and `registry get <type>` is the only way to find out. Probe it once before planning around a node:
+
+```
```
rockymadden: 🟡 **minor** — new fence has no language identifier.

`.claude/rules/content-quality.md` requires one, and every other command block in this skill uses ` ```bash ` (the other fence in this same file at :168 uses ` ```text `).

**Fix:** ` ```bash `.

rockymadden (approved):
## 📋 What

Removes the `canvas.nodes.*` feature-flag tables and "ask your admin" guidance from three OOTB node families in `uipath-maestro-flow` (Data Fabric entity CRUD, Batch Transform, Summarize), replaces them with a `registry get` probe, and inverts the Data Fabric default from **connector-first** to **native-first**. Rewrites the generic `AvailableOnTenant` definition in `references/shared/cli-commands.md`. Pins 8 Data Fabric eval tasks to the connector so they don't drift.

Markdown + task YAML only. All CI green.

## ✅ Confirmed against source, not just the description

I read `UiPath/cli@main`. The central claim holds:

- `packages/maestro-sdk/src/registry/node-sync-service.ts:239` — `ENABLED_MANIFEST_FLAGS` is a hardcoded `Set`, wired as `getFeatureFlag: (flag) => ENABLED_MANIFEST_FLAGS.has(flag)`. No tenant call. Source comment: *"Node categories the CLI always surfaces in the pulled manifest, **regardless of tenant entitlements**."*
- `READ_ENTITY` / `CREATE_ENTITY` / `UPDATE_ENTITY` / `DELETE_ENTITY` are in it (UiPath/cli#4087).
- `getManifestWithDiagnostics` builds `nodes: this.getBaseNodeList()` then **pushes** dynamic nodes. The server never withholds an OOTB node — the Summarize / Batch Transform "server rollout" model was wrong.
- `AvailableOnTenant` is `true` for anything in the pulled manifest and `false` **only** for `listKnownSdkNodes()` entries absent from it. The `cli-commands.md` rewrite is accurate.
- No `canvas.nodes` string survives anywhere in the repo. No `skill-flavors/studioweb` override touches a changed passage.

**Not verified:** that a released `uip` resolves the four nodes against a live tenant — I did not run `registry get`.

One concern checked and dismissed: `@uipath/flow-core@0.112.0` (CLI main's pin) ships `core.datafabric.create` **v1.3**, whose output carries the generated `Id`, so `impl.md`'s `=js:$vars.createOrder.output.Id` examples are valid. #4087's "create v1.0, no Id" note is itself stale.

## 🔎 Overall findings

🔴 **major — no eval coverage for the path this PR makes the default.**
`grep -rn 'core\.datafabric' tests/` returns 8 files, and after this PR all 8 say *"not the native `core.datafabric.*` nodes."* The PR flips the recommendation to native-first and simultaneously pins every build-grading task away from it. Nothing exercises a native `core.datafabric.*` build.
**Fix:** add one integration task under `tests/tasks/uipath-maestro-flow/` grading a native read + create (`file_contains` on `core.datafabric.*` in `definitions[]`, `flow validate` passes) — here or in a follow-up named in the description.

🟡 **minor — `smoke_activation_positive.yaml` left unpinned while 8 siblings got the block.**
Its prompt — *"create a record in my DataFabric entity from a UiPath Flow"* — is the exact shape that now routes native, and it stays tagged `connector, uipath-uipath-dataservice`. Nothing fails (both criteria are `pass_threshold: 0.0` advisory), but the tags stop describing what the run builds. Three others were skipped too; two are safe on inspection, so just say so in the description: `smoke_activation_negative` (routing only), `smoke_update_existing_flow` (brownfield edit of a scaffolded connector activity), `smoke_file_activities` (names `DownloadFileFromRecordFieldV2` etc. explicitly).
**Fix:** add the same pin, or drop the two connector tags since that task only grades activation.

🟡 **minor — the description's field-availability claim is wrong.**
It says #4087 *"has not reached a stable CLI (latest tag is `v1.201.0`) … both answers are still live in the field."* The git tag isn't what users install. `npm view @uipath/cli dist-tags.latest` → `1.203.0-dev.8609`, published **2026-09-09T22:37Z**, after #4087 merged at 12:03Z. Anyone on `latest` or running `uip tools update` has the fix today. The docs are version-robust either way so nothing in the diff changes — correct the claim.

## 🧭 Recommendation

**Approve.** The factual work is solid and independently verified, and this kills a documented misdiagnosis that cost real time. Please land the three-way divergence fix and the code fence before merge; the missing native task can be a linked follow-up.

**tl;dr** — Core claim verified against `UiPath/cli@main` source, not just the PR body. Three things to fix: the new text gives three different recoveries for one error across `planning-arch.md`, `data-fabric/planning.md` and `impl.md`; nothing in `tests/` exercises the native path this PR just made the default; and the description's "hasn't reached the field" claim is wrong. Approving — the direction is right.

skills/uipath-maestro-flow/references/author/planning-arch.md
uipath-utkarsh: Fixed in 8d867da8a — and you were right that this was the PR's own defect class.

`impl.md — Registry validation` is now the single procedure, written as four numbered steps ending in "Do not loop." Both planning docs defer to it instead of restating a shorter variant, so the `impl.md#registry-validation` link you noted was deleted is back in `data-fabric/planning.md` and now also in `planning-arch.md` — planners see the `--force` step again.

While there I also restated this line as a routing rule rather than an availability outcome, because "one probe decides" was under-describing it even once consistent. The split is by operation: an explicit connector request wins, record CRUD defaults to native, and every *other* Data Service operation goes to the connector because no native node exists for it. That last case was missing entirely — the old text implied the connector was only ever a fallback.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/impl.md
uipath-utkarsh: Added in 8d867da8a, with your row close to verbatim. I confirmed the mechanism in `@uipath/flow-core` before taking it — `manifest.js` carries the note that the insert runs in the BPMN engine's element postprocessor and that `GetDataFabricAction()` falls back to `"update"` on older engines.

Also scoped the line 32 sentence as you asked: it now reads "no tenant setting governs **whether the registry serves this node**", and points at the new Debug row so the reader is told the runtime engine is a separate axis rather than being led to rule out the platform. The row itself opens with "Platform-side, not authoring" for the same reason.

One correction to your fix as written: the section is `## Debug`, not Troubleshooting, so the anchor is `#debug`.

skills/uipath-maestro-flow/references/shared/cli-commands.md
uipath-utkarsh: Capped in 8d867da8a — "try `uip tools update` **once**; if the type is still absent treat it as absent by design for this build," then choose an alternative or report it. Mirrors `plugins/data-fabric/impl.md` as you suggested.

I enumerated the set rather than taking the count: `MANIFEST_FEATURE_FLAGS` has **25** entries and **10** are outside `ENABLED_MANIFEST_FLAGS` — agent-memory, queue-operations, form-trigger, http-standalone, agent-tool-http-request, patterns, classify-document, do-while, escalation.quick-form, hitl-document. `queue-trigger` and `queue-performer` from your list are not flags in this version, which is where the 12 came from. I wrote "around ten" plus a truncated example list rather than a hard number, since the count rots on the next flag added.

Unrelated but worth flagging: `ENABLED_MANIFEST_FLAGS` also references `CONNECTOR_NODES` and `LOOP_CONTAINER`, neither of which exists in the `MANIFEST_FEATURE_FLAGS` object in the flow-core build I read — so both add `undefined` to the Set. Possibly just version skew between the installed flow-core and CLI main's pin, but it looks like a latent bug on the CLI side. Not in scope here.

skills/uipath-maestro-flow/references/author/plugins/data-fabric/planning.md
uipath-utkarsh: Fixed in 8d867da8a — now ```bash.

uipath-utkarsh:
Thanks @rockymadden — all four inline threads are addressed and resolved in 8d867da8a. The three findings in the review body:

**🔴 no eval coverage for the native path** — fixed here rather than deferred. New `tests/tasks/uipath-maestro-flow/node_features/datafabric_native/`, deliberately named as the counterpart to `connector_features/datafabric_connector/`. It grades a native read → create chain on `FlowCodeEvalEntity` and checks four things: native node instances exist, both name the right entity, no `uipath-uipath-dataservice` activity stands in alongside, and each `definitions[]` entry looks registry-sourced (`model.type == "bpmn:Task"` plus the `api-function` exclusion) rather than assembled from the doc's field list — that last one is the failure mode `impl.md` warns about, which passes `flow validate` and dies at runtime.

The part worth your eye is the `pre_run` probe. Whether the registry serves these four is a CLI-build property, so on a CLI predating #4087 the agent correctly falls back to the connector, every native criterion misses, and the task would read `FAILURE` against the skill. `preflight_native_nodes.py` turns that into `ERROR` instead — same reasoning as the docstring in `_shared/preflight_connections.py` about the 2026-09-04 outage.

I exercised the checker against synthetic `.flow` fixtures on all five paths (native pass; `bpmn:ServiceTask` definition fail; connector-substitute fail; wrong-entity fail; connector-only fail). `test_live_corpus_passes_both_gates` covers the new YAML — 15 passed.

**🟡 `smoke_activation_positive` left unpinned** — took your second option: dropped the `connector` and `uipath-uipath-dataservice` tags, since the task grades `skill_triggered` only and its prompt now routes native. The other three are deliberate and now named in the description: `smoke_activation_negative` (routing only), `smoke_update_existing_flow` (brownfield edit of a scaffolded connector activity), `smoke_file_activities` (names `DownloadFileFromRecordFieldV2` and friends explicitly — and file fields have no native node at all, so it stays connector by policy, not by pin).

**🟡 the description's field-availability claim** — you're right and the correction is in the body. I'd checked the git tag; `npm view @uipath/cli dist-tags.latest` is `1.203.0-dev.8614`, published after #4087 merged, so anyone on `latest` or running `uip tools update` has the fix today. The probe-and-fall-back shape stays because older pinned installs still answer "Node not found" — version-robust for the long tail, not because the fix is unreleased.

---

One thing that grew beyond your review, flagged because it changes the recommendation and not just its wording. The native-first flip is now an explicit **routing rule split by operation**, not an availability outcome:

1. **User named the connector** → build the connector activity, if it exists for that operation. Explicit instruction outranks the default.
2. **Record CRUD** → default to the native `core.datafabric.*` node wherever Flow carries it natively.
3. **Any other Data Service operation** → the connector, because only those four exist natively. For attachments, file fields, entity metadata and bulk work the connector is the *only* path, not a fallback.

Federated entities go to the connector regardless. Written into `planning-arch.md`, `data-fabric/planning.md`, `connector/impl.md` and `CAPABILITY.md` so all four agree. This also reframes the 8 eval pins: those prompts now name the connector explicitly, which is rule 1 — they grade the connector family on purpose rather than steering away from native.

The `availability-decides` heading is renamed to `the-operation-decides`; its one inbound anchor is updated. Note `check-skill-links.mjs` strips the fragment before resolving, so anchors aren't covered by CI — I verified the 400 anchored links by hand. The 5 this PR touches resolve; the 17 that don't are all in files this PR doesn't touch (`plugins/http/*`, `shared/variables-and-expressions.md`, `evaluate/*`) and are pre-existing, so I've left them.

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -266,7 +266,12 @@ Illustrative supported activities (confirm against `registry get` for the specif
...
+> **Data Fabric record CRUD has native nodes — they are the default; everything else on this connector is not.** `core.datafabric.read` / `create` / `update` / `delete` ([data-fabric/planning.md](../data-fabric/planning.md)) need no Integration Service connection and are authored with `Edit`/`Write` instead of `node configure`, so for those four operations go native: confirm with `uip maestro flow registry get core.datafabric.read`, and on `NodeGetSuccess` leave this doc. Stay here when **any** of these hold — and they are common:
+>
+> - the operation is **not** one of those four (attachments, file-field downloads, entity metadata, bulk work) — no native node exists, so these activities are the only path, not a fallback;
+> - the **user asked for the connector by name** — an explicit request outranks the native default, so build it here as long as the activity exists;
+> - `registry get` ends at "Node not found" after [data-fabric/impl.md — Registry validation](../data-fabric/impl.md#registry-validation);
+> - the entity is **federated**, since the native writes require a native entity.
```
AditiGoyalUipath: _the entity is **federated**, since the native writes require a native entity._    - This is not needed here because federated entity Write is not at all supported and blocked. It is covered in datafabric skills

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/contractregistry_crud_filters.yaml
```diff
@@ -17,6 +17,10 @@ pre_run:
     timeout: 60
 
 initial_prompt: |
+  Build this with the UiPath Data Service connector activities
```
AditiGoyalUipath: We also have other PR to fix existing test cases . We can merge any one of them https://github.com/UiPath/skills/pull/3195/changes

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_activation_positive.yaml
```diff
@@ -1,8 +1,10 @@
 task_id: skill-flow-datafabric-smoke-activation-positive
 description: >
   Activation smoke: representative Data Fabric prompt routes to the
-  uipath-maestro-flow skill.
-tags: [uipath-maestro-flow, smoke, mode:build, lifecycle:discover, connector, uipath-uipath-dataservice]
+  uipath-maestro-flow skill. Grades routing only — no connector or native
+  markers, because this prompt's build path is record CRUD, which defaults to
+  the native `core.datafabric.*` nodes, and nothing here grades what gets built.
```
AditiGoyalUipath: unnecessary description

AditiGoyalUipath (approved):
Please address comments added

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
uipath-utkarsh: Removed in 0fc7e4f67. You were right, and the same mistake was in two more places — your point is stronger than "not needed here".

If a federated write is blocked outright, then telling an agent to *go to the connector* for one is not redundant, it is wrong: it sends them to build something that also cannot work. That claim appeared three times, so I fixed all three:

1. `connector/impl.md` — bullet deleted, as you asked.
2. `data-fabric/planning.md` (the routing line I added in this PR) — said "go to the connector regardless when the entity is federated". Now: writing a federated entity is blocked, so the connector is not an alternative path, with a pointer to **Writes require a native entity**.
3. `data-fabric/planning.md` selection heuristics, **pre-existing** — "Write to a federated entity | No — these nodes cannot; use connector or http". The connector half was wrong for the same reason. Now points at the source system via its own connector or http, which is the only thing that can actually work.

Flagging (3) explicitly since it predates this PR and I extended your comment to reach it — say the word and I will pull it back out into its own change.

One thing I could **not** verify myself: I searched `@uipath/flow-core` and the CLI for the enforcement and found nothing — zero hits in `manifest.js`, and the only federated surface in the CLI is `--native-only` on `df entities list`. So "blocked" reads as Data Fabric service-side, which is consistent with this repo already saying the write "is rejected by Data Fabric, and the rejection is swallowed". I have taken your word for the platform behaviour rather than asserting a mechanism I could not see.

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/contractregistry_crud_filters.yaml
uipath-utkarsh: Good catch — dropped mine in 0fc7e4f67, #3195 owns these.

Worth being precise about the risk, because it is not a conflict. I test-merged #3195 (`9a98bfa`) against this branch: **zero conflict markers**, because your pin sits before the `validate` line and mine sat at the top of the prompt. So both would have landed silently and each of the 7 prompts would carry the same instruction twice in two different wordings. That is worse than a conflict, which at least announces itself.

Reverted the 7 you also touch to `origin/main` exactly — `git diff origin/main` is empty for all of them. `trigger_lifecycle.yaml` stays in this PR since #3195 does not touch it.

Net effect: #3195 should merge for its own content regardless (the turn-budget raises fixing the MAX_TURNS failures, plus the `pathParameters` placeholder note and the `IsNull`/`IsNotNull` row — none of that is duplicated here). If #3195 ends up abandoned, ping me and I will restore the 7 pins here.

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_activation_positive.yaml
uipath-utkarsh: Reverted in 0fc7e4f67 — description is back to the original two lines, and only the tag change remains.

For context on why it was there: I dropped the `connector, uipath-uipath-dataservice` tags because the task grades `skill_triggered` only and its prompt now routes native, so the tags had stopped describing the run. I had put the reasoning in the description; it belongs in the PR discussion instead, which is where it now lives.

uipath-utkarsh:
@AditiGoyalUipath — I have to partly walk back the dedup on your test-case comment, because CI disagreed with me and it is worth knowing why.

I removed my 7 pins in 0fc7e4f67 so #3195 could own them. The smoke gate then failed at **84.6% (22/26)** against a 95% threshold, with all four failures in this family:

```
skill-flow-datafabric-smoke-create-all-types   FAILURE
skill-flow-datafabric-smoke-error              FAILURE
skill-flow-datafabric-smoke-query              ERROR
skill-flow-datafabric-smoke-update             FAILURE
```

Clean A/B on this branch — `8d867da8a` (pins present) passed the same gate; `0fc7e4f67` (pins removed) failed it. Restored in a77b21e0f.

**The pins are not redundant with yours — they are required by this PR's doc change**, which is the part I got wrong when I called the dependency soft. On `main` the docs recommend connector-first, so these prompts build connector activities and their checkers match. The moment record CRUD defaults to native, an unpinned prompt routes to `core.datafabric.*` while the checkers still assert `uipath.connector.uipath-uipath-dataservice.*`. So the pin has to travel in the same PR as the doc change; it cannot be split into a follow-up without the gate going red in between.

**Suggested split, and it is your call since it means an edit on your side:** #3195 drops its 7 pin lines and keeps everything that is genuinely only there — the turn-budget raises, the `pathParameters` placeholder note, and the `IsNull`/`IsNotNull` operator row. None of that is duplicated here. That keeps your PR independently mergeable in either order.

If you would rather keep your pins and have me drop mine, that works too, but then #3195 has to merge **first** — this PR cannot go green on its own without them.

One thing yours fixes that mine does not: `smoke_query` came back as `ERROR`, not `FAILURE`, which is the MAX_TURNS ceiling your budget raise addresses. It passed at 100 turns on the earlier run, so it is borderline rather than broken — your 150 is the real fix and I have deliberately not duplicated it here.

Your three review comments are all still addressed and resolved — the federated correction (all three occurrences), the activation-task description revert, and the tag removal are untouched by this revert.

PR #3195, opened 2026-09-10
fix(uipath-maestro-flow): raise DF query/e2e turn budgets and documen…

AditiGoyalUipath:
## Summary

Fixes DF connector-features tests that were failing when the agent built with the newer native `core.datafabric.*` node family — every grader in `tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/` checks the IS connector suffixes (`.create-entity-record`, `.query-entity-records`, etc.), so a native build validates green and scores zero.

## Changes (3 commits on this branch)

**`230545a9` — turn budget + two generic doc additions** *(later partially reverted, see `1e9f1823f`)*
- Initial `max_turns` bumps on `smoke_query` + `e2e_contract_intake_pipeline` (reverted).
- `connector-trigger/impl.md`: add `IsNull`/`IsNotNull` operator row (retained).
- `connector/impl.md` delta 2: append `pathParameters` note for templated endpoints (later dropped as duplicate of Step 6b).

**`9a98bfa8` — pin DF CRUD tests to the IS connector node family**
- One prompt line in 7 CRUD tests (`smoke_create_all_types`, `smoke_error`, `smoke_update`, `smoke_query`, `contractregistry_crud_filters`, `integration_create_get`, `e2e_contract_intake_pipeline`) naming both node families and picking the IS-connector one.
- Skipped: activation (node type irrelevant), file-activities and trigger_lifecycle (later extended, see below).

**`1e9f1823f` — address review**
- `e2e_contract_intake_pipeline`: revert `max_turns`/`expected_turns` and move the budget to `turn_timeout: 1800→2700` + `task_timeout: 2400→5400`. On the codex driver every CI runner uses (`smoke-skills.yml`, `run-coder-eval.yml`), `max_turns` does not bind — `turn_timeout` is the field that actually aborts. Matches peer `e2e/escalation_orchestrator_paths` at 5400.
- `smoke_query`: revert `max_turns` 150→100. CI showed the task green at 456s with 1 turn — cap never bound.
- `connector/impl.md`: drop the `pathParameters` sentence added to delta 2 — duplicates Step 6b (line 232) and the generic definition (line 20), and mis-frames a generic rule as v4-specific.
- Wrap all 7 pinning lines at ~75 chars to match the surrounding block-scalar wrap.
- Extend the pin to two more tasks the previous skip-list missed:
  - `smoke_file_activities` (grader hard-requires `.create-entity-record` node with `inputs.detail.pathParameters.entityName`)
  - `trigger_lifecycle` (grader hard-requires `.query-entity-records`/`.get-entity-record-by-id`/`.delete-entity-record` suffixes for downstream nodes)

## Native `core.datafabric.*` coverage

Intentionally out of scope for this PR. The pin is temporary until the native family is stable in production. See #3231 for the opposite remedy (grader widening) chosen on the billing e2es — the two branches use different strategies for the same collision.

## Passing-run evidence

Smoke-tagged tasks that ran in this PR's CI gate ([job](https://github.com/UiPath/skills/actions/runs/34505213985), commit `9a98bfa8`) and scored `1.000`:
- `skill-flow-datafabric-smoke-create-all-types`
- `skill-flow-datafabric-smoke-error`
- `skill-flow-datafabric-smoke-query`
- `skill-flow-datafabric-smoke-update`

Not smoke-tagged, did not run in the smoke gate: `integration_create_get`, `contractregistry_crud_filters`, `e2e_contract_intake_pipeline`.

### DIFF
--- a/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
@@ -36,7 +36,7 @@ Definitions in `definitions[]` are CLI-owned. `uip maestro flow node add` copies
 An activity is `4.0.0` when its `configuration` JSON reports `"version":"4.0.0"`. Author it through the Configuration workflow below like any other connector activity; `describe`/version mechanics are in [/uipath:uipath-platform — resources.md § `--activity-version`](../../../../../uipath-platform/references/integration-service/resources.md#--activity-version). Four deltas:
 
 1. **`objectName` source** — read the definition's `configuration` `objectName` (the `model.context[]` `objectName` entry is declared with no value). It is the `describe` positional, but do NOT pass it in `--detail` — `node configure` reads it from the definition itself. If configuration carries no `objectName`, the registry copy is stale: run `uip maestro flow registry pull --force`, delete the `definitions[]` entry, and re-add the node.
-2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`). If `endpoint` contains a `{placeholder}`, pass the same-named key in `pathParameters`.
+2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`).
 3. **Operation label ≠ HTTP verb** — a semantic operation (e.g. `Update`) pairs with any verb (e.g. `POST /usergroups.users.update`). `flow validate` accepts it; do not "fix" the method to match the label.
 4. **Not connection-scoped** — `--connection-id` on `registry get` adds no custom fields.
 
--- a/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
+++ b/skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
@@ -36,7 +36,7 @@ Definitions in `definitions[]` are CLI-owned. `uip maestro flow node add` copies
 An activity is `4.0.0` when its `configuration` JSON reports `"version":"4.0.0"`. Author it through the Configuration workflow below like any other connector activity; `describe`/version mechanics are in [/uipath:uipath-platform — resources.md § `--activity-version`](../../../../../uipath-platform/references/integration-service/resources.md#--activity-version). Four deltas:
 
 1. **`objectName` source** — read the definition's `configuration` `objectName` (the `model.context[]` `objectName` entry is declared with no value). It is the `describe` positional, but do NOT pass it in `--detail` — `node configure` reads it from the definition itself. If configuration carries no `objectName`, the registry copy is stale: run `uip maestro flow registry pull --force`, delete the `definitions[]` entry, and re-add the node.
-2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`).
+2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`). If `endpoint` contains a `{placeholder}`, pass the same-named key in `pathParameters`.
 3. **Operation label ≠ HTTP verb** — a semantic operation (e.g. `Update`) pairs with any verb (e.g. `POST /usergroups.users.update`). `flow validate` accepts it; do not "fix" the method to match the label.
 4. **Not connection-scoped** — `--connection-id` on `registry get` adds no custom fields.
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
@@ -10,8 +10,8 @@ description: >
 tags: [uipath-maestro-flow, e2e, mode:build, lifecycle:generate, shape:multi-node, connector, uipath-uipath-dataservice]
 
 run_limits:
-  expected_turns: 40
-  max_turns: 100
+  expected_turns: 80
+  max_turns: 150
   turn_timeout: 1800
   task_timeout: 2400
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
@@ -10,10 +10,10 @@ description: >
 tags: [uipath-maestro-flow, e2e, mode:build, lifecycle:generate, shape:multi-node, connector, uipath-uipath-dataservice]
 
 run_limits:
-  expected_turns: 80
-  max_turns: 150
-  turn_timeout: 1800
-  task_timeout: 2400
+  expected_turns: 40
+  max_turns: 100
+  turn_timeout: 2700
+  task_timeout: 5400
 
 pre_run:
   - command: 'python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/preflight_connections.py" uipath-uipath-dataservice'
@@ -48,7 +48,9 @@ initial_prompt: |
   5. Delete Entity Record — bind recordId to the create activity's
      output Id.
 
-  Use the Data Fabric Integration Service connector activities (`uipath.connector.uipath-uipath-dataservice.*`) for these operations — not the native `core.datafabric.*` nodes.
+  Use the Data Fabric Integration Service connector activities
+  (`uipath.connector.uipath-uipath-dataservice.*`) for these operations, not
+  the native `core.datafabric.*` nodes.
 
   The Flow is not complete until `uip maestro flow validate` passes.
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
@@ -48,6 +48,8 @@ initial_prompt: |
   5. Delete Entity Record — bind recordId to the create activity's
      output Id.
 
+  Use the Data Fabric Integration Service connector activities (`uipath.connector.uipath-uipath-dataservice.*`) for these operations — not the native `core.datafabric.*` nodes.
+
   The Flow is not complete until `uip maestro flow validate` passes.
 
   This run is headless. No user is present and nobody will answer a question or
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
@@ -6,7 +6,7 @@ description: >
 tags: [uipath-maestro-flow, smoke, mode:build, lifecycle:generate, connector, uipath-uipath-dataservice]
 
 run_limits:
-  max_turns: 150
+  max_turns: 100
   turn_timeout: 2400
   task_timeout: 3600
 
@@ -41,7 +41,9 @@ initial_prompt: |
   tree, sort, and limit, but use offset/start 2 for the next page. Query 3 must
   filter active equals true, sort by score descending, and use a limit of 4.
 
-  Use the Data Fabric Integration Service connector activities (`uipath.connector.uipath-uipath-dataservice.*`) for these operations — not the native `core.datafabric.*` nodes.
+  Use the Data Fabric Integration Service connector activities
+  (`uipath.connector.uipath-uipath-dataservice.*`) for these operations, not
+  the native `core.datafabric.*` nodes.
 
   The Flow is not complete until `uip maestro flow validate` passes.
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
@@ -41,6 +41,8 @@ initial_prompt: |
   tree, sort, and limit, but use offset/start 2 for the next page. Query 3 must
   filter active equals true, sort by score descending, and use a limit of 4.
 
+  Use the Data Fabric Integration Service connector activities (`uipath.connector.uipath-uipath-dataservice.*`) for these operations — not the native `core.datafabric.*` nodes.
+
   The Flow is not complete until `uip maestro flow validate` passes.
 
   This run is headless. No user is present and nobody will answer a question or
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
@@ -6,7 +6,7 @@ description: >
 tags: [uipath-maestro-flow, smoke, mode:build, lifecycle:generate, connector, uipath-uipath-dataservice]
 
 run_limits:
-  max_turns: 100
+  max_turns: 150
   turn_timeout: 2400
   task_timeout: 3600
 
--- a/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/trigger_lifecycle.yaml
+++ b/tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/trigger_lifecycle.yaml
@@ -37,6 +37,11 @@ initial_prompt: |
 
   For both flows, keep the configured connection and correct entity on every
   trigger and Data Fabric activity.
+
+  Use the Data Fabric Integration Service connector activities
+  (`uipath.connector.uipath-uipath-dataservice.*`) for these operations, not
+  the native `core.datafabric.*` nodes.
+
   Validate every final flow file.
 
   This run is headless. No user is present and nobody will answer a question or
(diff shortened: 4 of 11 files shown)


tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
```diff
@@ -41,6 +41,8 @@ initial_prompt: |
   tree, sort, and limit, but use offset/start 2 for the next page. Query 3 must
   filter active equals true, sort by score descending, and use a limit of 4.
 
+  Use the Data Fabric Integration Service connector activities (`uipath.connector.uipath-uipath-dataservice.*`) for these operations — not the native `core.datafabric.*` nodes.
```
AditiGoyalUipath: This is temporary . Once native node is stable in poduction. Will remove these tests

niket0503 (approved):

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
```diff
@@ -10,8 +10,8 @@ description: >
 tags: [uipath-maestro-flow, e2e, mode:build, lifecycle:generate, shape:multi-node, connector, uipath-uipath-dataservice]
 
 run_limits:
-  expected_turns: 40
-  max_turns: 100
+  expected_turns: 80
+  max_turns: 150
```
rockymadden: **major**

`max_turns: 150` with `turn_timeout: 1800` unchanged.

The agent budget is `turn_timeout`, not `max_turns` (orchestrator warning, observed on the sibling task in this PR's CI). If 100 turns already fit inside 1800s, 150 will not; the abort just moves from MAX_TURNS to a turn timeout, and `task_timeout: 2400` leaves grading almost nothing.

Peers at this turn count carry far more wall clock: `e2e/escalation_orchestrator_paths` (expected 80, max 200) runs 1800/5400, `multi_node/billing_dispute_resolution` (max 200) runs 3600/5400. This task is the outlier at 150/1800/2400.

**Fix:** raise `turn_timeout` and lift `task_timeout` above it, e.g. 2700/5400. Or drop the `max_turns` change and raise `turn_timeout` alone.

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
```diff
@@ -6,7 +6,7 @@ description: >
 tags: [uipath-maestro-flow, smoke, mode:build, lifecycle:generate, connector, uipath-uipath-dataservice]
 
 run_limits:
-  max_turns: 100
+  max_turns: 150
```
rockymadden: **minor**

`max_turns: 150` is the highest of any smoke task in this skill. Next is `hitl/smoke_02_completed_port_wired` at 105; every other DF smoke sits at 100.

This PR's CI has the task green at 456s with `total_assistant_turns=1` and `max_turns_exhausted=false`. Nothing in that run supports 150, and a cap that can never bind stops being a regression tripwire.

**Fix:** keep 100, or cite the turn count from the run that needed 150.

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/smoke_query.yaml
rockymadden: **minor**

The pinned line is one ~180-char line; the rest of this block scalar wraps at ~75. Same in all 7 files.

**Fix:** wrap to match:

```
  Use the Data Fabric Integration Service connector activities
  (`uipath.connector.uipath-uipath-dataservice.*`) for these operations, not
  the native `core.datafabric.*` nodes.
```

skills/uipath-maestro-flow/references/author/plugins/connector/impl.md
```diff
@@ -36,7 +36,7 @@ Definitions in `definitions[]` are CLI-owned. `uip maestro flow node add` copies
...
 
 1. **`objectName` source** — read the definition's `configuration` `objectName` (the `model.context[]` `objectName` entry is declared with no value). It is the `describe` positional, but do NOT pass it in `--detail` — `node configure` reads it from the definition itself. If configuration carries no `objectName`, the registry copy is stale: run `uip maestro flow registry pull --force`, delete the `definitions[]` entry, and re-add the node.
-2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`).
+2. **`method` / `endpoint`** — from `connectorMethodInfo` (`registry get`) or `availableOperations[]` (`is resources describe <connector-key> <object-name> --activity-version 4.0.0`). If `endpoint` contains a `{placeholder}`, pass the same-named key in `pathParameters`.
```
rockymadden: **minor**

Duplicates `Step 6b` line 232 ("Supply endpoint placeholders through `pathParameters` ...") and the definition at line 20. The fact is also not v4-specific: any templated endpoint needs it, which is why line 232 holds it generically. Parking a v4-scoped copy in delta 2 implies the rule is a v4 thing.

`.claude/rules/token-optimization.md` section 0 wants one full statement plus pointers.

**Fix:** drop the sentence. If delta-2 readers really skip the workflow, make it a pointer instead: "Templated `endpoint` placeholders go in `pathParameters` (Step 6b).

rockymadden (changes requested):
## Review: DF turn budgets + connector-node docs

### What it does

Two commits, only the first described in the body:

| Commit | Files | Change |
|---|---|---|
| `230545a9` | 4 | `max_turns` 100 to 150 on `smoke_query` + `e2e_contract_intake_pipeline`, `expected_turns` 40 to 80 on the e2e; `IsNull`/`IsNotNull` operator row; `pathParameters` note |
| `9a98bfa8` | 7 | One prompt line in every DF CRUD task pinning the build to `uipath.connector.uipath-uipath-dataservice.*` and away from `core.datafabric.*` |

Root cause of the pin is real and confirmed: every DF grader matches IS op suffixes (`check_e2e_contract_intake_pipeline.py:86-94`, `check_smoke_query_filter.py:83`, `check_ops_present.py:26`), so a native `core.datafabric.*` build validates green and scores zero.

### Overall findings

**Major: the repo now holds two opposite remedies, and the native node family has no test coverage.**

`grep -rn 'core.datafabric' tests/` returns exactly the 7 lines this PR adds. All 7 are negative. #3231 (`84a6a3b7e`, on main, not in this branch) solved the same collision the other way for the billing e2es: it widened the grader via `ENTITY_QUERY_HINTS` in `tests/tasks/uipath-maestro-flow/_shared/flow_check.py`. After this merges, the family the skill calls "the better build where the tenant does have them" (`data-fabric/planning.md:60`) is the one with zero positive coverage, and #3231 shows that is the shape agents actually produce in the eval tenant.

**Fix:** land the pin (the suite is connector-scoped, so pinning is defensible), then add one positive `core.datafabric.*` task, or state in the body why that family stays uncovered.

**Major: the skip list in `9a98bfa8` misses two tasks with the same exposure.**

- `smoke_file_activities.yaml:31` asks for `CreateEntityRecord`, and `check_smoke_file_activities.py:151-153` hard-requires a `.create-entity-record` node with `inputs.detail.pathParameters.entityName`. The rationale says "not CRUD". It has a graded CRUD step.
- `trigger_lifecycle.yaml:27,34,36` ask for Query Entity Records, Get by ID and Delete, and `check_trigger_lifecycle.py:52-56` hard-requires all three IS suffixes. The rationale says "triggers, no native equivalent", which holds for the trigger nodes only.

Both have partial cover (file-activities is surrounded by connector-only activities; trigger_lifecycle's prompt says "using the configured UiPath Data Fabric connection", which the native node has no concept of), but neither is the explicit pin the other 7 got.

**Fix:** pin both, or correct the rationale to name the actual shield.

**Major: the turn-budget bump is unverified, and likely targets the wrong field.**

This PR's own smoke gate ([run 34505213985](https://github.com/UiPath/skills/actions/runs/34505213985), commit `9a98bfa8`) reports for `skill-flow-datafabric-smoke-query`: `total_assistant_turns=1`, `max_turns_exhausted=false`, `duration=456s`, `score=1.000`. All 26 tasks in that report show `turns=1` regardless of their `max_turns`. The orchestrator also logged, on this task:

> `run_limits.task_timeout (3600s) exceeds run_limits.turn_timeout (2400s). A larger task_timeout cannot extend the agent's single iteration; the agent budget is turn_timeout.`

Every CI runner uses `--type codex` (`smoke-skills.yml:493`, `run-coder-eval.yml:488`), and on that driver `max_turns` does not bind.

**Fix:** give the failing run's id, agent type and abort reason. If the MAX_TURNS aborts came from a non-codex driver, name it, because that decides whether `turn_timeout` is the field to raise.

**Minor: the body and both advisory bots predate the pin commit.**

The body covers `230545a9` only; 7 of 9 files are undescribed. Both `github-actions` comments are stamped `2026-09-10T07:05`, about ten hours before `9a98bfa8` at `16:56`. So the Claude review's "All four changes are correct ... approve" and the lint's "2 task YAMLs changed" never saw the prompt pins.

That bot review is also unreliable on its own terms: it calls the unchanged `task_timeout` correct ("the issue was turn count, not wall-clock time"), then contradicts itself under "Issues for Manual Review", and asserts the `pathParameters` note "doesn't duplicate other content" when `connector/impl.md:232` already states it.

**Fix:** rewrite the body to cover both commits.

**Minor: passing-run claim covers 4 of the 7 modified tasks.**

`smoke_create_all_types`, `smoke_error`, `smoke_query` and `smoke_update` all scored `1.000` in this PR's gate, which satisfies `.claude/rules/test-writing.md` step 5 for those. `integration_create_get`, `contractregistry_crud_filters` and `e2e_contract_intake_pipeline` are not smoke-tagged and did not run.

**Fix:** cite the CI run for the four, run the other three or say why not.

**Verified clean:** the `IsNull`/`IsNotNull` row is correct. The operator is in the IS SDK `FilterOperator` enum (`uipath-maestro-case/references/case-spec-input-details.md:202`) and in the CEQL table (`uipath-platform/references/integration-service/activities.md:124`). Both changed doc lines change what the agent does.

**Could not verify:** the codereval blob history for the cited MAX_TURNS aborts. Azure token expired, so the abort reason and the agent that produced it are unconfirmed. Reporting that as a gap, not a pass.

### Recommendation

**Request changes**, scoped to two items: the `e2e_contract_intake_pipeline` run_limits (major finding 3 plus the inline comment), which as written probably does not do what the body claims, and the stale body. Everything else can land as follow-up. The pin itself is sound and CI-proven on 4 of 7 tasks.

### tl;dr

The prompt pin is the real fix and CI proves it works. The turn-budget bump is the weak half: on the codex driver every CI runner uses, `max_turns` never binds, and the e2e's unchanged `turn_timeout: 1800` is the limit that actually aborts it. Two tasks were skipped on wrong rationale (`smoke_file_activities`, `trigger_lifecycle`) and both have graded CRUD steps. The native `core.datafabric.*` family still has zero positive coverage while main's #3231 went the opposite direction. Body describes 2 of 9 files, and both bot comments predate the 7-file commit.

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/trigger_lifecycle.yaml
```diff
@@ -37,6 +37,11 @@ initial_prompt: |
 
   For both flows, keep the configured connection and correct entity on every
   trigger and Data Fabric activity.
+
+  Use the Data Fabric Integration Service connector activities
```
rockymadden, line 41: **minor**

The pinned prefix is not the trigger prefix. Triggers are `uipath.connector.trigger.uipath-uipath-dataservice.record-created`, and this block sits directly under the trigger instructions saying "these operations".

Only Query / Get / Delete collide here. There is no native trigger, so the pin adds nothing for the trigger and risks the agent emitting a type the registry will not return.

**Fix:**

```
  Use the Data Fabric Integration Service connector activities
  (`uipath.connector.uipath-uipath-dataservice.*`) for the Query, Get and
  Delete steps, not the native `core.datafabric.*` nodes. The triggers keep
  their `uipath.connector.trigger.uipath-uipath-dataservice.*` types.
```

tests/tasks/uipath-maestro-flow/connector_features/datafabric_connector/e2e_contract_intake_pipeline.yaml
```diff
@@ -12,8 +12,8 @@ tags: [uipath-maestro-flow, e2e, mode:build, lifecycle:generate, shape:multi-nod
 run_limits:
   expected_turns: 40
   max_turns: 100
-  turn_timeout: 1800
-  task_timeout: 2400
+  turn_timeout: 2700
```
rockymadden, line 15: **minor**

`turn_timeout: 2700` is a new repo high with no run behind it. Every peer sits lower: `e2e/escalation_orchestrator_paths` at 1800 (your cited match, which matches on `task_timeout` only), and the two heaviest tasks in this directory, `smoke_file_activities` and `trigger_lifecycle`, both at 2400.

The raise itself is justified: `smoke_update` burned 1393s of 1500 on a 4-activity chain, and this task chains 5 plus a FilterBuilder tree. The number is the issue.

**Fix:** set `2400` to match the directory, or run the task once and set it from the observed duration. `task_timeout: 5400` is fine either way.

rockymadden (approved):
## Review: DF connector-test node-family pin

### What

Pins 9 Data Fabric tests in `connector_features/datafabric_connector/` to the Integration Service node family, plus one budget move and one doc row.

| Change | Files | Detail |
|---|---|---|
| Prompt pin | 9 task YAMLs | One 3-line block naming both node families, picking `uipath.connector.uipath-uipath-dataservice.*` |
| Budget | `e2e_contract_intake_pipeline.yaml` | `turn_timeout` 1800→2700, `task_timeout` 2400→5400 |
| Doc | `connector-trigger/impl.md` | `IsNull` / `IsNotNull` operator row |

**Premise confirmed.** Every grader in that directory matches on IS suffixes and `inputs.detail.pathParameters.entityName`:

- `check_smoke_create_all_types.py:75,84`, `check_smoke_error.py:26,29`, `check_integration_create_get.py:25,27`, `check_smoke_file_activities.py:111,153`, `check_trigger_lifecycle.py:48-56`, `check_ops_present.py:26-31`
- Native types are `core.datafabric.{read,create,update,delete}` and carry `inputs.entityConfig`, so `.endswith(".create-entity-record")` genuinely excludes them and a native build scores zero.

The collision is real: `planning-arch.md:177` and `data-fabric/planning.md:58` call the native node "the better build" where the tenant flags are on, and they are on in the eval tenant (per #3231).

**Prior findings all addressed.** `max_turns` reverted on both tasks, budget moved to `turn_timeout`, the 7 pin lines wrapped to the surrounding ~75 chars, the duplicate `pathParameters` sentence in `connector/impl.md` dropped.

**Skip list checks out.** `smoke_update_existing_flow` is brownfield (`scaffold_movie_report_flow.py` drops an IS-connector node the agent only edits), and both activation tasks grade routing, never a flow. Neither needs the pin.

**`IsNull` / `IsNotNull` row is correct.** Same operator, same "no `value` needed" semantics as `uipath-platform/references/integration-service/activities.md:124`, same filter-tree shape, and `_shared/test_connector_checker_representation_parity.py:44` already emits it.

**Evidence is stronger than the body claims.** Head-commit smoke gate ([run 34606936748](https://github.com/UiPath/skills/actions/runs/34606936748)) has all 6 DF build smokes at `1.000`, including the two the last commit added:

```
smoke-create-all-types   SUCCESS  453s  1.000
smoke-error              SUCCESS  969s  1.000   ← newly pinned
smoke-file-activities    SUCCESS  882s  1.000   ← newly pinned
smoke-query              SUCCESS  720s  1.000
smoke-update             SUCCESS 1393s  1.000
smoke-update-existing    SUCCESS  241s  1.000
```

No guard breaks: `test_headless_preamble.py` matches the canonical block as a substring, `test_criterion_budgets.py` prices criterion timeouts only, and `origin/main` has not touched these files since the merge base.

---

### Overall findings

**minor — PR title is stale.** `raise DF query/e2e turn budgets and documen…` describes commit 1 of 3. The query budget raise was reverted, the doc change shrank to one operator row, and the pin (9 of 10 files) is unmentioned. The repo squash-merges with the PR title (`... (#3173)`, `... (#3179)` on main), so this lands in history as-is.
**Fix:** retitle to `test(uipath-maestro-flow): pin DF connector tests to the Integration Service node family`.

**minor — 4 of 9 pinned tasks have never run.** `integration_create_get`, `contractregistry_crud_filters`, `e2e_contract_intake_pipeline`, `trigger_lifecycle` are not smoke-tagged, so the gate skipped them. The pin is the same one-line mechanism that just worked on 6 tasks, so the risk is low, but it is unmeasured.
**Fix:** one `Run Coder Eval` dispatch on `tasks/uipath-maestro-flow/connector_features/datafabric_connector/**.yaml` with `codex`.

**minor — the "Passing-run evidence" section is stale.** It cites the `9a98bfa8` job and lists 4 tasks. The head commit has 6 green, including `smoke_error` and `smoke_file_activities`, which is the evidence for the commit that actually needs it. Update it.

**minor — the pin is the opposite remedy to merged #3231, and nothing in the repo records the split.** #3231 widened graders to accept native, arguing that pinning a gate to one shape leaves the path the skill steers toward ungradeable. Pinning is defensible *here*, since these tasks exist to cover the connector, but that reasoning lives only in a PR body and a review thread. The `command_executed` advisories in `integration_create_get.yaml:1-26` and `smoke_query.yaml:1-8` also hard-code the connector prefix, so they will need the same decision.
**Fix:** file the tracking issue for native coverage and link it from the task descriptions, not just the PR.

**minor, out of scope — `smoke_update` is a latent flake.** `turn_timeout: 1500`, observed 1393s. 93% of budget. Not touched by this PR; flagging it since this PR is raising the neighbor's budget for the same reason.

**No AI slop in the body.** It is specific, cites commits and line numbers, and states its own gaps. The two stale sections above are drift from the last commit, not filler.

**I could not verify** the claim that `max_turns` does not bind on the codex driver. `tests/README.md:186` documents driver as `docker`/`tempdir` and still tells devs at line 611 to raise `max_turns` when an agent runs out of turns. Your observation (1 turn, 456s, `max_turns_exhausted=false`) is consistent with it, and the reverts are right either way, but the repo does not document the codex single-turn behavior anywhere. Worth a follow-up line in `tests/README.md`.

---

### Recommendation

**Approve.** Premise verified against the graders, all four prior findings fixed, CI green with 6 DF smokes at `1.000` including both newly pinned tasks. Everything here and inline is minor: worth one more push, none of it blocks.

---

**tl;dr** — The fix is correct and the graders confirm it. Prior findings addressed. Approving. Fix the stale PR title before merge (squash puts it in main), and see the two inline nits. Run the four non-smoke tasks once if you want the pin measured rather than assumed.

PR #3197, opened 2026-09-10
Remove name based filter for maestro flow

Mihaiii:
For https://uipath.atlassian.net/browse/PILOT-7533

Studioweb outputs "new.flow" flow files instead of "<proj_name>.flow" file names, which are expected in tests. Therefore, studioweb stdio test checks fail even though the flow was generated. This is a test issue - the name doesn’t have to be in a particular format.

Run example (maestro-flow only) here with (unmerged) studioweb stdio and this skills branch: ~~https://dev.azure.com/uipath/coder_eval/_build/results?buildId=13343309&view=results~~
Results: TODO

Run example (maestro-flow only) here with Delegate SDK and this skills branch: ~~https://dev.azure.com/uipath/coder_eval/_build/results?buildId=13339361&view=results~~
Results: TODO


Mihaiii:
Will get back to this after https://github.com/UiPath/skills/pull/3124 is merged into main (after the review process) and main into this branch.

PR #3201, opened 2026-09-10
docs: use placeholders for example values that trip content scanners

RaduAna-Maria:
## Summary

- Replace literal dates, addresses, phone numbers, and identifier-shaped digit runs in **10 documentation files** with the repo's existing `<UPPER_SNAKE_CASE>` placeholder convention
- Unblocks a customer whose content-inspection gate rejected the published package on 21 findings — all false positives on example text, but a shape-based gate blocks the artifact either way
- Add the rule to `CONTRIBUTING.md` § CLI Commands and `.claude/rules/content-quality.md` so new examples don't reintroduce it

No credential, live PII, or cardholder data exists in the repo. An independent sweep over all tracked files (AWS key IDs, private keys, `gh*_`/`xox*`/`sk_live` tokens, JWTs, npm/PyPI tokens, Luhn-valid PANs) returned zero hits. This PR removes *shapes*, not secrets.

### `uipath-admin`

- `pat-management.md`, `external-app-management.md` — `--expiration "2027-01-15"` → `--expiration "<EXPIRATION_DATE>"`
- `audit-commands.md`, `audit-workflow-guide.md` — export windows → `<FROM_DATE>` / `<TO_DATE>`; generated names → `audit_<from>_<to>_<generatedAt>`, the placeholder both files already used in prose

> Both audit references are now free of **every** literal calendar date, not just the lines the customer's scan reported. The detector fired on an unpredictable subset — `2026-01-01` yes, `2026-04-22` no — so a partial fix would fail the next scan.

### `uipath-rpa`

- `UiPath.Word.Activities/…/examples.md` — mail-merge sample data → `<COMPANY_NAME>` / `<STREET_ADDRESS>` / `<CONTACT_NAME>` / `<PHONE_NUMBER>`
- `UiPath.AzureWVD.Activities/…/examples.md` — placeholder GUID loses its digit runs: `ffffffff-1111-2222-3333-444444444444` → `ffffffff-aaaa-bbbb-cccc-dddddddddddd`, matching the letter-based GUID on the line above it
- `legacy/data-manipulation-guide.md` — regex table masks digit positions (`NNN-NN-NNNN`, `MM/DD/YYYY`) with a note explaining `N`; `ParseExact` takes a `dateText` variable instead of a literal
- `legacy/testing-guide.md` — the "number too long" row describes its length instead of spelling out 11 digits

## Left unchanged, deliberately

Four findings are product data the playbooks exist to convey, not examples. Editing them would break the page:

- Orchestrator error code `170002` (`uipath-troubleshoot`) — the identifier being diagnosed
- Two `forum.uipath.com` thread IDs inside URLs — the ID *is* the link
- A Databricks documentation URL in `vendor-docs-registry.json` — matched a 40-char AWS-secret shape rule

These need an allowlist on the scanning side. The cleanest ask is excluding documentation-only artifacts from cardholder-data scanning: the package is Markdown and JSON reference material with no runtime data path.

## Testing

- `npm run skills:validate` — OK, default 27 skills / 1765 files; studioweb 27 / 1765 / 264 replacements
- `npm run skills:check-links` — 6723 relative links, all resolve
- `npm run skills:build` — both trees build clean
- `python scripts/check-skill-status.py` — OK, 27 skills, manifest valid
- `python scripts/check-skills-sh.py` — OK, 27 skills across 4 sections
- Re-scanned all 10 files with the shape detectors that produced the original findings: **0 remaining matches**

No skill folder added, renamed, or removed, so `assets/skill-status.json` and `skills.sh.json` need no edit. No flavor overrides or `skill-flavor` markers touch any edited file.

## Reviewer notes

> `CONTRIBUTING.md` deliberately spells out the literals it forbids, in the "Do not write" column. That file is not in `files` in `package.json`, so the anti-examples never reach a customer scan — there's an inline note saying so, and to exclude it if we ever add a repository-wide content scanner.

Worth considering as a follow-up: none of the 22 workflows scans content today, so nothing stops the next literal date from landing.

### DIFF
--- a/.claude/rules/content-quality.md
+++ b/.claude/rules/content-quality.md
@@ -27,6 +27,7 @@ This repository's primary audience is AI coding agents, not humans. Write accord
 - Show the full command with all required flags
 - Use `<PLACEHOLDER>` for user-provided values (angle brackets, UPPER_SNAKE_CASE)
 - Use a placeholder for any example value that carries the shape of a date, an identifier, or personal data — customer content-inspection gates match on shape alone and block the whole published package. Write `--expiration "<EXPIRATION_DATE>"`, never `--expiration "2027-01-15"`; write `<STREET_ADDRESS>` and `<PHONE_NUMBER>`, never a sample address or phone number; mask digit positions as `NNN-NN-NNNN`. State the required format in prose instead: ``--expiration <date>``, ISO 8601 `YYYY-MM-DD`.
+- Give each value one placeholder covering the whole value. Never glue a placeholder to a literal fragment: write `--from-date <FROM_TIMESTAMP>`, not `--from-date <FROM_DATE>T00:00:00Z`.
 - Keep product identifiers the reader needs verbatim — error codes, support-forum thread IDs in URLs, vendor documentation URLs. Those are content, not examples.
 - Specify whether a flag is required or optional
 - Show expected output when it clarifies behavior
--- a/.claude/rules/content-quality.md
+++ b/.claude/rules/content-quality.md
@@ -26,6 +26,8 @@ This repository's primary audience is AI coding agents, not humans. Write accord
 
 - Show the full command with all required flags
 - Use `<PLACEHOLDER>` for user-provided values (angle brackets, UPPER_SNAKE_CASE)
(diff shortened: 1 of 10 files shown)


Mihaiii (approved):
Approved based on what a client requested, although imo there wasn't any PII concern and the new format is less useful than the old one for the model.

PR #3211, opened 2026-09-10
fix(uipath-maestro-bpmn): variables carry an elementId; script mappings use BPMN.Variables

DevMomo:
## Problem

The skill taught two variable rules backwards, so **every BPMN an agent authored from it was broken in Studio Web.**

**Root variables.** The skill said a process-level declaration carries no `elementId`. The canvas requires one — an `elementId` *is* how scope is expressed:

```xml
<!-- what the skill taught: not a variable at all to the canvas -->
<uipath:inputOutput id="Var_Tier" name="tier" type="string" />

<!-- what the canvas requires -->
<uipath:inputOutput id="Var_Tier" name="tier" type="string" elementId="Process_1" />
```

Every `vars.Var_Tier` reference to the first form fails on import with `VARIABLE_DOES_NOT_EXIST`. One reported file showed 28 of them.

**Script mappings.** The skill said to use `<uipath:type value="BPMN.ScriptTask">` and *explicitly forbade* `BPMN.Variables`. It is the reverse: any type other than `BPMN.Variables` stops the runtime dispatching the script. The element still completes and validation stays clean, so nothing looks wrong — the variable just reads `null`.

Both rules were inferred from a symptom rather than checked against the canvas, and the examples around them reinforced the mistake.

---

**Stack 1/4** — base `main`. Followed by #fixtures → #assert-value → #evals.

Two authoring rules in this skill were inverted, and each made every BPMN an agent authored from it fail in Studio Web. Both were inferred from a symptom rather than checked against the canvas or the engine, then reinforced by the examples around them.

### 1. Process-level variables

The skill implied a root declaration carries no `elementId` ("public or node-scoped declarations *also* carry the owning node's elementId"), and its example declared `Var_Decision` without one.

The canvas keys scope by `elementId`. Its existence check filters on any non-empty value (PO.Frontend `VariableValidationUtils.ts:163`, `origin/develop`), so a declaration without one is not a variable at all, and every `vars.<id>` reference to it raises `VARIABLE_DOES_NOT_EXIST` on import.

Field evidence: 4 of 5 exported diagrams carried elementId-less root variables — 56 declarations. The one file that worked had all 9 keyed to its process id.

**Verified in Studio Web.** The reported `0909-claude-SiteCoordination.bpmn` showed 28 errors. With `elementId="<process id>"` added to its ten root variables and nothing else changed, the canvas validation panel reads `Validation issues (0)` / "No issues found".

### 2. Script-task mappings

The skill told agents to use `<uipath:type value="BPMN.ScriptTask">` and **explicitly forbade** `BPMN.Variables`. It is the reverse: any type other than `BPMN.Variables` overwrites the parser's `Scp.Script` extension type, so the runtime never dispatches the script. The element still completes, the output mapping resolves against an empty result, and the target variable reads back `null`.

**Verified live** on alpha: two files differing only in that attribute returned `product: 42` (`BPMN.Variables`) and `product: null` (`BPMN.ScriptTask`). Engine test fixtures agree — 242 `BPMN.Variables`, zero `BPMN.ScriptTask`.

The registry's `BPMN.ScriptTask` `xmlTemplate` emits the wrong type child, so this is documented as the one place to correct the template. Its own `extensionTagNotes` already say script tasks route the same way as `BPMN.Variables` — **the template contradicts its own note, which is worth fixing upstream.**

### Also

The return contract was inverted too: at `scriptVersion` v2+ the runtime wraps the return under `response` itself (`ScriptActivities.cs:211-216`), so `return { response: v }` double-wraps to `result.response.response`. Return the value directly and map `source="=result.response"`.

Every citation was re-checked against PO.Frontend `origin/develop` rather than a stale checkout.

### DIFF
--- a/skills/uipath-maestro-bpmn/references/structural-bpmn.md
+++ b/skills/uipath-maestro-bpmn/references/structural-bpmn.md
@@ -124,20 +124,25 @@ example BPMN files** — it is the main reason authoring runs out of time.
 </bpmn:definitions>
 ```
 
-## Variables (`BPMN.Variables`)
+## Variables
 
-Declare root variables with the `BPMN.Variables` registry template attached to
-the process via `extensionElements`, or use the canvas `<uipath:variables>`
-block directly. Every declaration needs a stable, unique `id`, a non-empty
-user-facing `name`, and its documented `type`; do not use the name as a
-substitute for the id. Expressions reference the id as `vars.<id>`. Public or
-node-scoped declarations also carry the owning node's `elementId`. Variable
-schema bodies are JSON text or CDATA.
+Declare variables in the process's own `<uipath:variables>` block. Every
+declaration needs a stable, unique `id`, a non-empty user-facing `name`, and its
+documented `type`; do not use the name as a substitute for the id. Expressions
+reference the id as `vars.<id>`. Variable schema bodies are JSON text or CDATA.
+
+Every declaration also carries an `elementId` naming the element that owns it:
+the `<bpmn:process>` id for a process-level variable, the start event id for a
+caller-supplied input, the end event id for a published output, the owning node's
+id for a node-scoped variable. A subprocess-level variable is keyed inside its
+own subprocess: the subprocess id, or the node in it that writes the value —
+both import cleanly. Without an `elementId` the declaration does not exist to
+the canvas, and every `vars.<id>` reference to it fails on import.
 
 ```xml
 <uipath:variables version="v1">
   <uipath:input id="input_ExpenseId" name="expenseId" type="string" elementId="Start_1" />
-  <uipath:inputOutput id="Var_Decision" name="decision" type="string" />
+  <uipath:inputOutput id="Var_Decision" name="decision" type="string" elementId="Process_1" />
   <uipath:output id="output_Decision" name="decision" type="string" elementId="End_1" />
 </uipath:variables>
 ```
@@ -159,7 +164,8 @@ Sub-process-scoped variables go in that sub-process's own `<uipath:variables>`.
 
 `bpmn:scriptTask scriptFormat="JavaScript"` runs under **Jint**, not Node.js or
 a browser. The mapping payload comes from the `BPMN.ScriptTask` registry
-template, but the runtime contract is fixed:
+template, but the runtime contract is fixed — including one correction to that
+template, in the first rule below:
 
 - Only these helpers exist: `uipath.aggregate`, `uipath._aggregate`,
   `uipath._pipe`, and a no-op `console`. No npm packages, filesystem, network,
@@ -177,23 +183,30 @@ template, but the runtime contract is fixed:
   never retrofit these attributes onto an untouched node's mapping — a
   pre-existing `<uipath:input name="args">` outside the edit's target stays
   byte-identical.
-- Map the returned object's property back through `source="=result.response"`
-  (the conventional scalar property) or `source="=result.response.<field>"`
-  (another object field); `var` points at a declared variable id (do not put the
-  target id in `name`).
-- When the output mapping uses `source="=result.response"`, return an object
-  with a `response` property, such as `return { response: 6 * 7 };`. Do not
-  return the bare primitive `42` for that mapping shape; there is no
-  `response` property to bind, so the runtime variable stays empty.
-- Do not use `source="=result"` with a bare scalar return in live debug/runtime
-  BPMN. Studio Web can report `FinalStatus: Completed` while the target root
-  variable still reads back as `{}` or `null` from
-  `debug-instance variables-all`.
-- For live debug/runtime runs, never use `source="=this.result"` or
-  `<uipath:type value="BPMN.Variables" ...>` on a script task output mapping.
-  That older structural-test shape can pass local validation but leaves root
-  variables `null` or faults in Studio Web. Use the `BPMN.ScriptTask` mapping
-  with `source="=result.response"`.
+- **The mapping's type child is `<uipath:type value="BPMN.Variables"
+  version="v1" />`, not `BPMN.ScriptTask`.** The registry's `BPMN.ScriptTask`
+  `xmlTemplate` emits the latter, and this is the first of two corrections that
+  template needs: any `uipath:type` other than `BPMN.Variables` overwrites the parser's
+  `Scp.Script` extension type, so the runtime never dispatches the script. The
+  element still completes, the output mapping resolves against an empty result,
+  and the target variable reads back `null`. Verified live: two files differing
+  only in this attribute returned `product: 42` (`BPMN.Variables`) and
+  `product: null` (`BPMN.ScriptTask`).
+- Map the return through `source="=result.response"` for a scalar, or
+  `source="=result.response.<field>"` for a field of a returned object; `var`
+  points at a declared variable id (do not put the target id in `name`).
+- **The template ships no `<uipath:scriptVersion>`, and that is the second
+  correction.** A missing element parses as `v1`
+  (`UiPath.PO.BpmnParser/Extensions/Xml/ScriptReader.cs`), and at v1 the runtime
+  demands an object return and throws `ScriptTaskInvocationResultError`
+  otherwise. So a template pasted with only the type child corrected, plus the
+  bare return below, faults at run time. Add
+  `<uipath:scriptVersion value="v3" />` as a sibling of `uipath:mapping`.
+- Return the value directly — `return 6 * 7;`. At `scriptVersion` v2 or later the
+  runtime wraps the return under `response` itself, so returning
+  `{ response: value }` yields `result.response.response`. Do not use
+  `source="=result"` or `source="=this.result"`, which read the wrapper object
+  rather than the value.
 - Do not mutate `Globals.*`, `vars.*`, or process variables inside the script
   body. The supported path is: return a value from the script, then use a
   `uipath:output` mapping to write it to the declared variable. Direct mutation
@@ -204,14 +217,14 @@ template, but the runtime contract is fixed:
   <bpmn:extensionElements>
     <uipath:scriptVersion value="v3" />
     <uipath:mapping version="v1">
-      <uipath:type value="BPMN.ScriptTask" version="v1" />
+      <uipath:type value="BPMN.Variables" version="v1" />
       <uipath:input name="args" type="json" target="bodyField"><![CDATA[{"amount":"=vars.Var_Amount","daysOverdue":"=vars.Var_DaysOverdue"}]]></uipath:input>
       <uipath:output name="riskScore" type="number" var="Var_RiskScore" source="=result.response" />
     </uipath:mapping>
   </bpmn:extensionElements>
   <bpmn:script><![CDATA[
 var score = amount * 0.01 + daysOverdue * 2;
-return { response: score };
+return score;
 ]]></bpmn:script>
 </bpmn:scriptTask>
 ```
--- a/tests/scripts/test_bpmn_project_scaffold_contract.py
+++ b/tests/scripts/test_bpmn_project_scaffold_contract.py
@@ -96,7 +96,10 @@ def test_minimal_example_has_complete_di_coverage() -> None:
 
 def test_variable_and_migration_examples_use_serializer_attributes() -> None:
     text = REFERENCE.read_text(encoding="utf-8")
-    variables_section = text.split("## Variables (`BPMN.Variables`)", maxsplit=1)[1]
+    # Match the heading, not its parenthetical: the section has been titled both
+    # "## Variables" and "## Variables (`BPMN.Variables`)".
+    _, marker, variables_section = text.partition("\n## Variables")
+    assert marker, "structural-bpmn.md is missing its Variables section"
     variables_section = variables_section.split("\n## ", 1)[0]
     match = re.search(r"```xml\n(?P<xml>.*?)\n```", variables_section, re.DOTALL)
     assert match, "structural-bpmn.md is missing its variable declaration example"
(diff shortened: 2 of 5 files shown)


tests/scripts/test_bpmn_project_scaffold_contract.py
```diff
@@ -113,6 +116,12 @@ def test_variable_and_migration_examples_use_serializer_attributes() -> None:
     assert all(item.attrib.get("id") for item in declarations)
     assert all(item.attrib.get("name") for item in declarations)
     assert all(item.attrib.get("type") for item in declarations)
+    # Every declaration carries an elementId naming the element that owns it. A
+    # declaration without one is not a variable to the canvas, so each
+    # `vars.<id>` reference to it fails on import.
+    assert all(item.attrib.get("elementId") for item in declarations), [
```
rockymadden, line 122: **🟠 major — the silent half of this fix has no guard.**

This PR corrects two rules. Only one gets an assertion.

- `elementId` fails **loudly**: `VARIABLE_DOES_NOT_EXIST` on import. It now has this assert. 👍
- `<uipath:type value="BPMN.Variables">` fails **silently**: the element completes, the variable reads `null`. It has nothing.

Nothing else covers it either:

- `test_bpmn_doc_examples_parse.py` only parses blocks that open with `<?xml`, so the script-task fragment at `structural-bpmn.md:205-219` is skipped by design.
- `check_script_jint_guidance.py` reads the mapping's `args` input and its `source=`, but never its `<uipath:type>`.

That asymmetry is how this bug shipped. The loud rule got noticed, the quiet one did not, and the quiet one is the expensive one.

Sketch, same file, same style:

```python
def test_script_task_example_maps_through_bpmn_variables() -> None:
    text = REFERENCE.read_text(encoding="utf-8")
    _, marker, section = text.partition("\n## Script tasks")
    assert marker, "structural-bpmn.md is missing its Script tasks section"
    match = re.search(r"```xml\n(?P<xml>.*?)\n```", section.split("\n## ", 1)[0], re.DOTALL)
    assert match, "structural-bpmn.md is missing its script task example"
    wrapper = ET.fromstring(
        f'<root xmlns:bpmn="{NS["bpmn"]}" xmlns:uipath="{NS["uipath"]}">'
        f"{match.group('xml')}</root>"
    )
    types = wrapper.findall(".//uipath:mapping/uipath:type", NS)
    assert types, "script task example has no mapping type child"
    assert all(t.attrib.get("value") == "BPMN.Variables" for t in types), [
        t.attrib for t in types
    ]
```

Worth adding the same check to `check_script_jint_guidance.py` so the eval grades it too, not just the doc.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -124,20 +124,23 @@ example BPMN files** — it is the main reason authoring runs out of time.
...
+Declare variables in the process's own `<uipath:variables>` block. Every
+declaration needs a stable, unique `id`, a non-empty user-facing `name`, and its
+documented `type`; do not use the name as a substitute for the id. Expressions
+reference the id as `vars.<id>`. Variable schema bodies are JSON text or CDATA.
+
+Every declaration also carries an `elementId` naming the element that owns it:
+the `<bpmn:process>` id for a process-level variable, the start event id for a
+caller-supplied input, the end event id for a published output, the owning node's
+id for a node-scoped variable. Without it the declaration does not exist to the
+canvas, and every `vars.<id>` reference to it fails on import.
```
rockymadden: **🟡 minor — this list has no case for a subprocess-level variable.**

Four cases named: process id, start event, end event, owning node. Line 159 then says subprocess variables go in that subprocess's own `<uipath:variables>`, but never says what `elementId` they carry.

An agent that reads this list literally, finds no matching case, and drops the attribute produces exactly the failure this PR exists to stop.

The platform accepts both forms, so this is safe to state: `s191_2-migration-utils.ts` *relocates* a self-keyed non-custom subprocess variable rather than rejecting it, which is why #3214 leaves the distinction deliberately ungraded.

Suggested tail for the sentence:

> ... the owning node's id for a node-scoped variable. A subprocess-level variable is keyed inside its own subprocess: the subprocess id, or the node in it that writes the value. Both import cleanly.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -177,23 +181,23 @@ template, but the runtime contract is fixed:
...
+  only in this attribute returned `product: 42` (`BPMN.Variables`) and
+  `product: null` (`BPMN.ScriptTask`).
+- Map the return through `source="=result.response"` for a scalar, or
+  `source="=result.response.<field>"` for a field of a returned object; `var`
+  points at a declared variable id (do not put the target id in `name`).
+- Return the value directly — `return 6 * 7;`. At `scriptVersion` v2 or later the
+  runtime wraps the return under `response` itself, so returning
+  `{ response: value }` yields `result.response.response`. Do not use
+  `source="=result"` or `source="=this.result"`, which read the wrapper object
+  rather than the value.
```
rockymadden, line 209: **🟡 minor — "return the value directly" is a fault if `scriptVersion` is missing.**

The bullet above says the registry template needs "one correction". It needs two. That template ships no `<uipath:scriptVersion>` at all (`validator/bpmn-spec.json:1909`):

```xml
<bpmn:ScriptTask id="{id}" name="{name}">
  <bpmn:extensionElements>
    <uipath:mapping version="v1">
      <uipath:type value="BPMN.ScriptTask" version="v1" />
      <uipath:input name="args" type="json" target="bodyField"><![CDATA[{}]]></uipath:input>
    </uipath:mapping>
  ...
```

`ScriptReader.cs:38` defaults a missing `scriptVersion` to `v1`, and at v1 `ScriptActivities.cs:215-224` demands an `ExpandoObject` return and otherwise throws `ScriptTaskInvocationResultError`. So: paste the template, apply only the correction this bullet names, write `return 6 * 7;`, and the run faults.

Line 172 does say to set `v3`, but it sits before the "one correction" framing and reads as a preference, not a fix to the template. Name both in the same bullet:

> ... this is the one place you must correct the template: the type child, and the `<uipath:scriptVersion value="v3" />` the template omits (a script task without it parses as v1, where a bare return faults).

rockymadden (commented):
## What

Flips two variable/script rules in `uipath-maestro-bpmn`:

- Every `<uipath:variables>` declaration carries an `elementId`. Root scope is `elementId == <bpmn:process> id`, not the absence of one.
- A script task's mapping type child is `BPMN.Variables`, not `BPMN.ScriptTask`.
- Bonus flip: return the value bare, the runtime wraps it under `response` at `scriptVersion` v2+.
- Adds one assertion pinning `elementId` on the doc example.

## Verified, not assumed

I traced every cited source rather than reading the PR body:

- ✅ `VariableValidationUtils.ts:163` (`origin/develop`) is `variables.filter((variable) => variable.elementId)` for the existence check, and `:154` is `rootVariables = variables.filter(v => v.elementId === root.id)`. Both claims land on the exact line.
- ✅ Nothing backfills it. `s191_2-migration-utils.ts:41` keeps an id-less variable in its current scope, `:196` keeps it through the orphan sweep. It survives import as a non-variable, which is exactly the reported `VARIABLE_DOES_NOT_EXIST`.
- ✅ `UiPathExtensionsProvider.cs:204` is literally `if (!type.Equals("BPMN.Variables", ...)) parentElement.ExtensionType = type;`. `ScriptReader.cs:23` sets `Scp.Script`, `TemporalBpmnExecutor.cs:287` dispatches on it. The silent-null story holds end to end.
- ✅ `ScriptActivities.cs:207-212` wraps the return under `response` when `IsVersionAtLeast(version, 2)`.

## Could not verify

The field evidence (4 of 5 exported diagrams, 56 declarations, 28 → 0 in the Studio Web panel) and the live alpha A/B run. No access to those artifacts. Taken as reported.

## Overall findings

**🟡 minor — the stale upstream comment that caused this is still there.**

`VariableValidationUtils.ts:258` reads:

```ts
// Process-level variables carry no elementId; attribute them to the root so the issue stays navigable.
const elementId = variable.elementId ?? rootId;
```

It is a dead fallback in the reserved-name check, not the governing path, so it does not change anything here. But it is the sentence that produced the wrong rule. Worth a one-line upstream PR so the next reader does not repeat this.

## Recommendation

**Comment.** The core claims check out at source, and the rewrite reads well. Three gaps inline, all cheap, none blocking.

## tl;dr

Rules are right, confirmed in PO.Frontend and PO.BpmnEngine. The half of the fix that fails silently (the mapping type child) is the half with no test guard.

nikhil-maryala (approved):

tests/scripts/test_bpmn_project_scaffold_contract.py
DevMomo: Added `test_script_task_examples_dispatch_and_return_bare` in this file. It asserts three things per script-task example: the mapping type is `BPMN.Variables`, `scriptVersion` is present and >= v2, and the body does not self-wrap its return.

Mutation-tested four ways — type flipped to `BPMN.ScriptTask`, return wrapped in `response`, `scriptVersion` set to `v1`, `scriptVersion` deleted. All four caught.

Worth noting: my first mutation run reported two of those as missed, which was my mutation matching the skeleton's deeper-indented line as a substring rather than the script-task example. Scoping the mutation to the `Task_RiskScore` block showed the guard was fine.

You were right that the asymmetry is how this shipped — the loud rule had an assertion and the silent one had nothing.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Added, using your wording:

> ... the owning node's id for a node-scoped variable. A subprocess-level variable is keyed inside its own subprocess: the subprocess id, or the node in it that writes the value — both import cleanly.

The "both import cleanly" half matters for the reason you gave: migration 13 relocates a self-keyed non-custom subprocess variable rather than rejecting it, which is also why #3214 leaves that distinction ungraded.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
DevMomo: Correct, and verified before changing it: the template carries no `scriptVersion` (`bpmn-spec.json`), `ScriptReader.cs:39` defaults a missing one to `v1`, and v1 demands an object return.

The bullet now says the template needs **two** corrections, and the second is spelled out as its own bullet — add `<uipath:scriptVersion value="v3" />` as a sibling of `uipath:mapping`, because a template pasted with only the type child fixed plus a bare return faults at run time.

The new guard in `test_bpmn_project_scaffold_contract.py` also fails if an example's `scriptVersion` is missing or v1, so this can't regress silently.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -159,7 +164,8 @@ Sub-process-scoped variables go in that sub-process's own `<uipath:variables>`.
 
 `bpmn:scriptTask scriptFormat="JavaScript"` runs under **Jint**, not Node.js or
 a browser. The mapping payload comes from the `BPMN.ScriptTask` registry
-template, but the runtime contract is fixed:
+template, but the runtime contract is fixed — including one correction to that
```
rockymadden, line 167: 🟡 **minor** — this says **"one correction"**.

Line 188 says *"the first of two corrections"* and line 191 says *"that is the second correction"*. The follow-up added the second correction and left this intro at one.

An agent that reads the intro and skims the bullets applies one fix and faults at run time, which is the exact path line 191 warns about.

```diff
-template, but the runtime contract is fixed — including one correction to that
-template, in the first rule below:
+template, but the runtime contract is fixed, and that template needs two
+corrections, called out in the bullets below:
```

rockymadden (approved):
## What

Flips two inverted authoring rules, plus the return contract. Adds a guard for the silent half.

## Prior round: all three addressed, verified

- ✅ **The silent half now has a guard.** `tests/scripts/test_bpmn_project_scaffold_contract.py:134` asserts three things per script-task example: mapping type is `BPMN.Variables`, `scriptVersion` present and not v1, body does not self-wrap. Wider than I asked for. Wired into CI at `.github/workflows/test-helpers.yml:162`, so it gates PRs.
- ✅ **Subprocess case added** at `structural-bpmn.md:137`, with the "both import cleanly" half.
- ✅ **Second template correction added** at `structural-bpmn.md:191`. Verified the premise again: `bpmn-spec.json:1909` ships no `scriptVersion`.

## Verified independently

- ✅ **Docs sweep is clean.** No remaining `<uipath:type value="BPMN.ScriptTask">`, no `return { response`, no elementId-less declaration anywhere in `skills/uipath-maestro-bpmn/`. Every residual `BPMN.ScriptTask` hit is the registry template's own name or the corrective prose.
- ✅ **The registry contradicts itself, as the body claims.** `BPMN.ScriptTask.extensionTagNotes` reads *"routed through createModdleExtensionVariableMapping (same as BPMN.Variables)"* while its `xmlTemplate` emits `BPMN.ScriptTask`. Worth the upstream fix.
- ✅ SKILL.md moving `<uipath:variables>` to "Authored (registry gap)" is right: the `BPMN.Variables` template is a `bpmn:Task` mapping shell with zero declarations.

## Overall findings

**🟡 minor — `structural-bpmn.md:175` still reads as an instruction to wrap.**

The bullet is outside the diff, so it has no line to comment on:

```
- Set `uipath:scriptVersion value="v3"` for new scripts; preserve an imported
  `value="v2"`. For v2+ the script returns JSON under `response`.
```

Line 199 says the opposite: *"Return the value directly"*, because the runtime does the wrapping. This is the sentence shape that produced the original bug, and the new guard checks the *example*, not this prose, so nothing catches a regression here.

```diff
-  `value="v2"`. For v2+ the script returns JSON under `response`.
+  `value="v2"`. At v2+ the runtime wraps whatever the script returns under
+  `response`, so the script returns the value itself.
```

One more inline.

## Recommendation

**Approve.** Every claim lands at source, the guard closes the asymmetry that let this ship, and CI runs it.

## tl;dr

Rules right, guard added and gating. Two leftover sentences contradict the new bullets. Non-blocking.

PR #3212, opened 2026-09-10
test(uipath-maestro-bpmn): align test inputs with the canvas variable contract

DevMomo:
## Problem

#3211 corrected the rules in the skill. The test inputs still taught the old ones:

- **11 fixtures** declared variables with no `elementId`
- **8 fixtures** mapped script tasks through `BPMN.ScriptTask`, so their scripts never dispatch
- **17 script bodies** wrapped their return in `{ response: … }`, which double-wraps at `scriptVersion` v2+
- **5 task prompts** instructed the broken shape outright — one told the agent *"do not add an `elementId`"*

An agent that reads one of these fixtures, or follows one of those prompts, reproduces exactly the bug #3211 removes. Prose is not the operative teaching surface here; examples are.

---

**Stack 2/4** — base `fix/bpmn-root-variable-element-id-docs` (#3211). Review after that one.

Test inputs taught the broken shapes by example, so an agent reading a fixture — or following a task prompt — would reproduce the bug the skill no longer teaches. Four repairs, each verified by running the affected tasks rather than by reading the checks.

1. **Variable scoping** (11 files, 23 declarations) — every `<uipath:variables>` child now carries `elementId="<that file's own bpmn:process id>"`. Not one of these processes is named `Process_1`, so this is a per-file lookup, not a find-and-replace. `instance-asset.json` holds the deployed copy of `OrderRouting.bpmn` and gets the identical edit: the task is asset *drift*, and editing only the local file would plant a second difference beside the threshold change the agent is meant to find.

2. **Script-task mappings** (8 fixtures, 17 occurrences) — `BPMN.ScriptTask` → `BPMN.Variables`, so the script actually dispatches.

3. **Malformed script args** (8 fixtures, 17 occurrences) — `<uipath:input name="args">` gains the `type="json" target="bodyField"` the skill requires. This was the real reason the canvas loader rejected these files outright (`expected … inputs item at index 2.type to be a string`); fix (2) had only routed around it via the laxer variable path.

4. **A dangling reference** — `OrderRouting.bpmn` read `vars.amount` while the declared id is `Var_Amount`. References resolve by id, verified: `vars.tier` against a variable whose id is `Var_Tier` raises `VARIABLE_DOES_NOT_EXIST`. Nothing had caught this because the file has no diagram and dies at `BPMN_PARSE_ERROR` before variable validation runs — give it a diagram and the error appears. Fixed in the fixture and the mock together.

**Prompts** (5 files) — `live_debug_e2e` no longer forbids `elementId`, no longer mandates a root `<uipath:output>` keyed to the process (that shape fails `refresh` outright: *"Process output must target a root end event"*), and no longer mandates a `{ response: … }` wrapper. Four prompts stopped routing declarations at the registry's `BPMN.Variables` template, which is a node mapping shell containing no declarations.

### Verification

Using the canvas's own rule engine — `@uipath/maestro-validation`, published from PO.Frontend, called directly so none of the CLI's suppression applies — the eight repaired fixtures go from `BPMN_PARSE_ERROR` to clean. `ExpenseApproval` keeps exactly its two seeded errors and `OrderRouting` still parse-errors for want of a diagram, both as before.

All ten affected tasks pass locally, **10/10 SUCCESS score 1.000**: six `edit/*`, both `diagnose/*`, `hitl/quality_brownfield_insert`, `patterns/ai_decision_review`. No `assert_preserved` failure, so the added attributes do not read as unauthorised modification. `validate_fix_loop` reports `Found 2 error(s)` on every validate call — the seeded pair, not more.


tests/tasks/uipath-maestro-bpmn/patterns/ai_decision_review/fixture/ClaimsIntake.bpmn
```diff
@@ -12,18 +12,18 @@
...
     <bpmn:startEvent id="Start_1" name="Claim Received"><bpmn:outgoing>Flow_SS</bpmn:outgoing></bpmn:startEvent>
     <bpmn:scriptTask id="Task_Score" name="Score Claim" scriptFormat="JavaScript">
       <bpmn:extensionElements>
         <uipath:scriptVersion value="v3" />
         <uipath:mapping version="v1">
-          <uipath:type value="BPMN.ScriptTask" version="v1" />
-          <uipath:input name="args"><![CDATA[{"claimId":"=vars.Var_ClaimId"}]]></uipath:input>
+          <uipath:type value="BPMN.Variables" version="v1" />
+          <uipath:input name="args" type="json" target="bodyField"><![CDATA[{"claimId":"=vars.Var_ClaimId"}]]></uipath:input>
           <uipath:output name="score" type="number" var="Var_ConfidenceScore" source="=result.response.score" />
```
rockymadden, line 27: **🟠 major — `return { response: ... }` at `scriptVersion v3` double-wraps.**

This PR fixed the mapping type on line 25 and left the return two lines below it.

`ScriptActivities.cs:207-212`:

```csharp
// Version 2+ allows for the script to return any JSON value.
if (IsVersionAtLeast(version, 2))
{
    return new BpmnEngineResult<Dictionary<string, object>>(
        new Dictionary<string, object> { { "response", output.Wrapped! } }
    );
}
```

With the `v3` on line 23, `return { response: { score: 0.91 } }` becomes `result.response.response.score`. Line 27 reads `=result.response.score`, which is `undefined`. Same pairing on lines 39/43.

#3211 now spells this out at `structural-bpmn.md:196`, and #3213 adds a check that fails exactly this shape. The fixtures are the last place still teaching it.

17 occurrences across 8 fixtures: `ClaimsIntake`, `Invoicing`, `OrderIntake`, `Onboarding`, `RiskScoring`, `ShipmentReview`, `Fulfillment`, `OrderFulfillmentBpmn`. Every one is a single line.

Field mapping (`source="=result.response.<field>"`):

```diff
-<bpmn:script><![CDATA[return { response: { score: 0.91 } };]]></bpmn:script>
+<bpmn:script><![CDATA[return { score: 0.91 };]]></bpmn:script>
```

Scalar mapping (`source="=result.response"`):

```diff
-<bpmn:script><![CDATA[return { response: weight * 2 };]]></bpmn:script>
+<bpmn:script><![CDATA[return weight * 2;]]></bpmn:script>
```

tests/tasks/uipath-maestro-bpmn/diagnose/deployed_asset_drift/fixtures/OrderRouting.bpmn
```diff
@@ -3,7 +3,7 @@
   <bpmn:process id="Process_OrderRouting" name="Order Routing" isExecutable="true">
     <bpmn:extensionElements>
       <uipath:variables version="v1">
-        <uipath:input id="Var_Amount" name="amount" type="number" />
+        <uipath:input id="Var_Amount" name="amount" type="number" elementId="Process_OrderRouting" />
```
rockymadden: **🟡 minor — a `uipath:input` keyed to the process is not a caller input.**

`Start_Order` is right there on line 9. Keying `Var_Amount` to `Process_OrderRouting` does make it a process-level variable, so `vars.Var_Amount` resolves on line 31 and the drift task still grades what it grades. No functional break.

But it is not the shape this stack teaches, and the stack's own eval rejects it:

- `structural-bpmn.md:135-136` (#3211): "the start event id for a caller-supplied input".
- `local-metadata-regeneration-guide.md:196`: entry-point `input` comes from root inputs "whose `elementId` matches the start event. **An input that matches nothing is dropped silently.**"
- `check_variable_scopes_nested.py:89` (#3214) fails any root `uipath:input` that is not start-event keyed.

`elementId="Start_Order"` satisfies both the canvas and the rule. Same on `ExpenseApproval.bpmn:6-7` (`Start`).

rockymadden (changes requested):
## What

Brings the BPMN task fixtures and five prompts onto the contract #3211 documents:

- `elementId` on 23 declarations across 11 files
- `BPMN.ScriptTask` → `BPMN.Variables` on 17 script mappings
- `type="json" target="bodyField"` on 17 `args` inputs
- one dangling `vars.amount` → `vars.Var_Amount`

## Verified

- ✅ **Every `elementId` value is right.** Checked all 11 files: each declaration's `elementId` matches that file's own `<bpmn:process id=...>`. No copy-paste of `Process_1`, as the body claims.
- ✅ **The declaration sweep is complete.** Grepping the whole `tests/tasks/uipath-maestro-bpmn/` tree at the top of the stack, every remaining `<uipath:input|output|inputOutput>` without an `elementId` is a `<uipath:mapping>` child, not a declaration. Nothing missed.
- ✅ **Editing `instance-asset.json` alongside `OrderRouting.bpmn` is correct.** The task grades a threshold drift (`> 5000` local vs `> 500` deployed), so touching one side only would plant a second diff beside the one the agent must find.
- ✅ The `args` diagnosis is consistent: `type`/`target` were absent, so the loader rejected on `inputs item ... .type to be a string` before the type child ever mattered.

## Overall findings

- 🟠 **major** — the fourth broken shape is still in the fixtures, in 17 script bodies. Inline on `ClaimsIntake.bpmn`.
- 🟡 **minor** — two `uipath:input` declarations keyed to the process instead of the start event. Inline on `OrderRouting.bpmn`.
- 🟡 **minor** — "10/10 SUCCESS score 1.000" and the `@uipath/maestro-validation` run are not reproducible from the diff. I did not re-run them. Taken as reported.

## Recommendation

**Request changes.** Nothing here breaks anything, and the sweep it does do is clean. But the PR's own standard is *"test inputs taught the broken shapes by example, so an agent reading a fixture would reproduce the bug the skill no longer teaches"*, and it fixes three of four while editing the lines directly above the fourth. Merging as-is leaves 17 wrong exemplars in files this PR just certified, which reads as endorsement.

Happy to be told the return shape is a deliberate follow-up. Then say that in the body and I will approve.

## tl;dr

Three of four shapes fixed, sweep is complete and per-file correct. The return contract is the one left, in 17 script bodies under `v3`. One line each.

tests/tasks/uipath-maestro-bpmn/patterns/ai_decision_review/fixture/ClaimsIntake.bpmn
DevMomo: Fixed all 17 across the 8 fixtures you listed.

One transform covered both mapping shapes, since unwrapping a layer turns `{ response: X }` into `X` for `source="=result.response"` and `{ response: {total: X} }` into `{total: X}` for `source="=result.response.total"`.

Verified with the canvas engine afterwards: the eight fixtures stay clean, `ExpenseApproval` keeps exactly its two seeded errors, and all ten affected tasks still pass (10/10, no `assert_preserved` failure).

tests/tasks/uipath-maestro-bpmn/diagnose/deployed_asset_drift/fixtures/OrderRouting.bpmn
DevMomo: Re-keyed to `Start_Order` here and `Start` in `ExpenseApproval.bpmn`, in the deployed-copy mock as well so the pair still differs only by the threshold.

This was a contradiction I had created rather than a pre-existing one: I keyed them to the process to avoid a `VARIABLE_NOT_SET` warning, while #3214's own check rejects exactly that shape. Tested the warning — it doesn't materialise; `ExpenseApproval` reports precisely its two seeded errors either way.

tests/tasks/uipath-maestro-bpmn/hitl/quality_boolean_decision/quality_boolean_decision.yaml
```diff
@@ -34,7 +34,7 @@ initial_prompt: |
   - The gateway condition MUST reference the boolean variable via `vars.<id>`
     and treat it as a boolean (e.g. `== true`), not compare it to a quoted
     string.
-  - Declare process variables with the registry BPMN.Variables template.
+  - Declare process variables in `<uipath:variables>`, each with an `elementId`.
```
rockymadden, line 37: 🟡 **minor** — this replaces a wrong instruction with a correct one, where deleting may be better.

The old line had to go. But restating the rule in the prompt is a different tradeoff than #3214 takes, which states: *"Neither prompt contains the word `elementId`. If it did, the task would measure instruction-following rather than the skill."*

Harmless here, because none of these four tasks grade scoping. But it is the shape #3214 argues against, in the same stack.

```diff
-  - Declare process variables in `<uipath:variables>`, each with an `elementId`.
+  - Declare process variables per the skill's structural reference.
```

Same on `quality_result_downstream.yaml:32`, `quality_schema_design.yaml:35`, `smoke/author_validate.yaml:31`. Fine to leave as-is if you prefer the prompts pinned.

rockymadden (approved):
## What

Brings 12 fixtures, one deployed-copy mock, and five prompts onto the contract #3211 documents.

## Prior round: both addressed, verified

- ✅ **All 17 double-wraps gone**, across the 8 fixtures I listed. Both mapping shapes handled correctly: `{response: {total: X}}` became `{total: X}` for `source="=result.response.total"`, and `{response: X}` became `X` for `source="=result.response"`.
- ✅ **The two caller inputs re-keyed** to `Start_Order` and `Start`, and the `instance-asset.json` mock with them.

## Verified independently

Parsed all 12 BPMN fixtures at this branch head and checked six invariants:

- ✅ Every `<uipath:variables>` child carries an `elementId`.
- ✅ Every `elementId` resolves to a real `bpmn:` element **in its own file**. No `Process_1` copy-paste.
- ✅ Zero `BPMN.ScriptTask` type children, zero `return { response`.
- ✅ Every `<uipath:input name="args">` carries `type="json" target="bodyField"`.
- ✅ Every `vars.<X>` resolves to a declared **id**. One exception: `vars.expenseTotal` in `validate_fix_loop`, which is a deliberate seeded defect and must stay.
- ✅ The mock and the local fixture now differ only by the threshold (`> 500` vs `> 5000`), so the drift task still grades one diff.

## Could not verify

The `@uipath/maestro-validation` run and "10/10 SUCCESS score 1.000". Not reproducible from the diff, and I did not run anything. Taken as reported. The static invariants above are what I could check, and they hold.

## Overall findings

One consistency nit inline. Nothing blocking.

## Recommendation

**Approve.** My changes-requested is cleared. The sweep is complete and per-file correct, and I re-derived that rather than trusting the body.

## tl;dr

All four broken shapes gone, verified file by file. The prompts now name `elementId`, which sits slightly against #3214's principle. Your call.

PR #3213, opened 2026-09-10
test(uipath-maestro-bpmn): make live_debug_e2e assert the runtime value again

DevMomo:
## Problem

This e2e exists to prove a computed variable reaches the runtime. **It cannot currently fail when that variable is empty.**

Its pass condition is "a definition named `product` exists" — true whether the value is `42`, `null`, or an object. #2214 lowered it to that because the value kept reading back `null`, and the `null` was attributed to a platform limitation.

It was not a platform limitation. It was the authoring shape the skill was teaching (#3211). So the test was weakened to accommodate a bug, and then could no longer detect it.

---

**Stack 3/4** — base `fix/bpmn-root-variable-element-id-fixtures` (#3212). Depends on #3211 for the skill fix.

This e2e existed to prove a computed variable reaches the runtime, and it stopped doing that. #2214 swapped the pass condition from `product == 42` to "a `globalDefinitions` entry named `product` exists" — a name lookup that succeeds whether the value is `42`, `null`, or an object — and wrote the reason into its own docstring: the value *"has repeatedly read back as null"*. That was attributed to a platform limitation.

**It was not one.** The `null` was the authoring shape: `<uipath:type value="BPMN.ScriptTask">` in a script task's mapping overwrites the parser's `Scp.Script` extension type, so the script never dispatches. Verified live with two files differing only in that attribute — `product: 42` vs `product: null`.

So the value is assertable again:

- `_has_product_definition` and both call sites are gone; `_has_expected_product` is the only pass path, on the saved evidence and on live recovery.
- The bare-return guard was inverted and now checks the opposite: at `scriptVersion` v2+ the runtime wraps the return under `response` itself, so `return { response: 6 * 7 }` double-wraps.
- A new structural assertion rejects any mapping `uipath:type` other than `BPMN.Variables`, **and says why** — so a future `null` fails with its cause named instead of being tolerated.

Live-recovery is kept: it exists so a run cannot fail merely because the agent forgot to save a file, which was the right half of #2136.

### Verification

Against the two real debug payloads: the working shape passes with `runtime product variable is 42`; the broken shape fails structurally, before any CLI call.

And the task itself, run live end to end against a real tenant: **SUCCESS, 3/3 criteria, score 1.000**, with `"Product": 42` in the `variables-all` payload. That run is the one place all four PRs have to be right at once — the agent authored from the fixed skill, chose `BPMN.Variables` and a bare return unprompted, uploaded, executed, and the value arrived.

### DIFF
--- a/tests/tasks/uipath-maestro-bpmn/debug/live_debug_e2e/check_live_debug.py
+++ b/tests/tasks/uipath-maestro-bpmn/debug/live_debug_e2e/check_live_debug.py
@@ -156,33 +156,29 @@ def _has_expected_product(parsed) -> bool:
     ``Globals.Product`` — or a ``{name: product, value: 42}`` pair), then a
     fallback restricted to value-bearing sub-trees so shape drift still passes
     without letting GUIDs/timestamps false-match.
+
+    This is the only pass path, so it matches by name and nothing else, and each
+    branch announces which one fired.
+
+    A third branch used to accept 42 at any leaf under a value-bearing subtree
+    as insurance against shape drift. It was removed: `VALUE_BEARING_KEYS`
+    includes "response", so it accepted
+    ``{"globals": {"product": {"response": 42}}}`` — precisely the double-wrap
+    this task guards against structurally — and gating it on the payload naming
+    `product` did not help, because in that shape the name is present and the
+    *value* is what is wrong. If the payload shape drifts, this should fail
+    loudly and be updated, not pass on a value nobody checked.
     """
     # (a) key named "product" with the expected value (Globals.Product = 42)
     for val in _find_ci_key(parsed, PRODUCT_VAR):
         if _is_expected(val):
+            print(f"  matched product == {EXPECTED_PRODUCT} by name")
             return True
     # (b) {"name": "product", "value": 42} pairs (mock/CLI list shape)
     for name, value in _iter_name_value_pairs(parsed):
         if isinstance(name, str) and name.strip().lower() == PRODUCT_VAR and _is_expected(value):
+            print(f"  matched product == {EXPECTED_PRODUCT} by name/value pair")
             return True
-    # (c) fallback: 42 anywhere under a variables/globals/outputs subtree
-    for leaf in _leaves_under_value_bearing(parsed):
-        if _is_expected(leaf):
-            return True
-    return False
-
-
-def _has_product_definition(parsed) -> bool:
-    for val in _find_ci_key(parsed, "globaldefinitions"):
-        if isinstance(val, dict):
-            for key, definition in val.items():
-                if not isinstance(definition, dict):
-                    continue
-                name = definition.get("Name", definition.get("name"))
-                if key.lower() == PRODUCT_VAR or (
-                    isinstance(name, str) and name.lower() == PRODUCT_VAR
-                ):
-                    return True
     return False
 
 
@@ -259,7 +255,7 @@ def _fresh_debug_then_variables():
         if not inst:
             continue
         variables = _read_variables_all(inst)
-        if variables is not None and _has_product_definition(variables):
+        if variables is not None and _has_expected_product(variables):
             return (True, variables)
     return (completed, None)
 
@@ -316,6 +312,16 @@ def main() -> None:
                         "source=\"=result.response\" so the live runtime exposes the value."
                     )
                 found_product_mapping = True
+            for mapping in task.findall(f".//{{{UIPATH_NS}}}mapping"):
+                for type_el in mapping.findall(f"{{{UIPATH_NS}}}type"):
+                    if type_el.get("value") != "BPMN.Variables":
+                        _fail(
+                            f"{path}: script task mapping declares "
+                            f"<uipath:type value=\"{type_el.get('value')}\">. Any value "
+                            "other than BPMN.Variables overwrites the parser's "
+                            "Scp.Script extension type, so the script never runs and "
+                            "product reads back null."
+                        )
         for body in _script_bodies(root):
             found_script = True
             if re.search(r"\bGlobals\.", body) or re.search(r"\bvars\.", body):
@@ -323,11 +329,12 @@ def main() -> None:
                     f"{path}: script task reads/mutates Globals.*/vars.* directly — "
                     "unsupported in Jint. Return a value and map it via uipath:output."
                 )
-            if re.search(r"\breturn\s+(?!\{)", body):
+            if re.search(r"\breturn\s*\{\s*response\b", body):
                 _fail(
-                    f"{path}: script task returns a bare scalar. For live BPMN debug, "
-                    "return an object such as `return { response: 6 * 7 };` and map "
-                    "`source=\"=result.response\"`."
+                    f"{path}: script task wraps its return in a `response` property. "
+                    "At scriptVersion v2+ the runtime wraps the return itself, so this "
+                    "double-wraps to result.response.response. Return the value "
+                    "directly: `return 6 * 7;`."
                 )
             if not re.search(r"\b6\s*\*\s*7\b", body):
                 _fail(f"{path}: script task must compute the product with `6 * 7`")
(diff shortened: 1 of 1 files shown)


tests/tasks/uipath-maestro-bpmn/debug/live_debug_e2e/check_live_debug.py
```diff
@@ -172,20 +172,6 @@ def _has_expected_product(parsed) -> bool:
     return False
```
rockymadden, line 182: **🟡 minor — fallback (c) is now load-bearing and can still pass a wrong value.**

With `_has_product_definition` gone, this function is the single pass path, so its weakest branch carries the whole check. Branch (c) accepts `42` at *any* leaf under a value-bearing subtree, with no name check, and `VALUE_BEARING_KEYS` includes `"response"`:

```python
VALUE_BEARING_KEYS = {"variables", "globals", "outputs", "output", "response"}
```

So `{"globals": {"product": {"response": 42}}}` passes here. That is the double-wrap this same PR guards against structurally, arriving through the value assertion.

In practice the structural check runs first and kills it, so the risk today is small. But the point of this PR is that the value assertion should stand on its own, and right now branch (c) is the reason it does not.

Two options, either is fine:

1. Gate (c) on the payload also carrying a `product`-named key somewhere, so a stray `42` cannot satisfy it alone.
2. Have each branch print which one matched. A pass via (c) then shows up in the log as shape drift rather than as a clean result, which is what you want if this ever silently degrades again.

rockymadden (approved):
## What

Restores the real assertion in `live_debug_e2e`:

- `_has_product_definition` (a name lookup that passed whether the value was `42`, `null`, or an object) is deleted. `_has_expected_product` becomes the only pass path, on saved evidence and on live recovery.
- The bare-return guard is inverted: it now rejects `return { response: ... }` instead of demanding it.
- New structural check rejects any mapping `uipath:type` other than `BPMN.Variables`, and names the cause in the failure text.

## Verified

- ✅ **The root-cause story is right.** `UiPathExtensionsProvider.cs:204` skips only `BPMN.Variables`; anything else overwrites the `Scp.Script` set at `ScriptReader.cs:23`, and `TemporalBpmnExecutor.cs:287` dispatches on that value. The `null` was the authoring shape, not a platform limit. #2214's docstring attribution was wrong, as claimed.
- ✅ **The new guard is correctly permissive about an absent `<uipath:type>`.** No type child means `ReadExtension` never assigns, `Scp.Script` survives, the script runs. Only a wrong value fails. Right call.
- ✅ `\breturn\s*\{\s*response\b` catches `return{response`, `return {response:` and `return { response :`. No gap I could find.
- ✅ **Keeping live recovery is right**, and it now recovers on the *value* rather than the definition, so it can no longer paper over a `null`. That was the failure mode of the thing being removed.

## Could not verify

The live run (`SUCCESS, 3/3, "Product": 42`) and the two saved debug payloads. Not in the diff, not runnable here. Taken as reported.

## Overall findings

Nothing structural. One soft spot inline.

## Recommendation

**Approve.** The check fails for the reason it exists again, and its failure text names the cause instead of tolerating it. This is the PR in the stack I have the fewest reservations about.

## tl;dr

Assertion is real again, and the diagnosis behind it holds at source. One pre-existing fallback is now load-bearing and worth tightening.

tests/tasks/uipath-maestro-bpmn/debug/live_debug_e2e/check_live_debug.py
DevMomo: Good catch, and it went further than the fix you suggested.

I tried option 1 first — gate branch (c) on the payload naming `product` — and testing showed it doesn't close the hole: in `{"globals": {"product": {"response": 42}}}` the name *is* present and the value is what's wrong, so it still passed.

So I removed branch (c) instead. Branches (a) and (b) cover both real payload shapes (`Globals.Product` and the `{name, value}` list), and (c) was speculative insurance that had started accepting a wrong value — the same failure this PR exists to undo. Each remaining branch now prints which one fired, per your option 2.

Truth-tabled it: the double-wrap, a `null`, a stray `42` and a wrong value all fail; the real `Globals.Product = 42` and the name/value pair pass. Re-ran against both live debug payloads — gold passes with `matched product == 42 by name`, the broken shape fails structurally.

tests/tasks/uipath-maestro-bpmn/debug/live_debug_e2e/check_live_debug.py
```diff
@@ -156,33 +156,29 @@ def _has_expected_product(parsed) -> bool:
     ``Globals.Product`` — or a ``{name: product, value: 42}`` pair), then a
     fallback restricted to value-bearing sub-trees so shape drift still passes
```
rockymadden, line 157: 🟡 **minor** — the first paragraph still advertises the branch the third paragraph explains you deleted.

*"then a fallback restricted to value-bearing sub-trees so shape drift still passes"* describes branch (c), which is gone. Reads as self-contradicting against the paragraph below it.

```diff
-    ``Globals.Product`` — or a ``{name: product, value: 42}`` pair), then a
-    fallback restricted to value-bearing sub-trees so shape drift still passes
-    without letting GUIDs/timestamps false-match.
+    ``Globals.Product``) or a ``{name: product, value: 42}`` pair.
```

rockymadden (approved):
## What

Deletes the name-lookup pass path so the check can fail on a `null` again.

## Prior round: addressed, and better than I suggested

- ✅ Branch (c) **removed** rather than gated. Your test showed gating it does not close the hole, because in `{"globals": {"product": {"response": 42}}}` the name is present and the value is what is wrong. Correct call, and (a)+(b) cover both real payload shapes.
- ✅ Each branch prints which one fired.

## Verified

- ✅ The new structural check is correctly **permissive about an absent `<uipath:type>`**. No type child means `Scp.Script` survives and the script runs. Only a wrong value fails.
- ✅ `\breturn\s*\{\s*response\b` catches `return{response`, `return {response:` and `return { response :`.
- ✅ `_fail` is `sys.exit`, so a wrong shape kills the run before any CLI call.

## Could not verify

The live run (`SUCCESS 3/3`, `"Product": 42`) and the two saved payloads. Not in the diff. Taken as reported.

## Overall findings

Three leftovers, none on a line inside this diff.

**🟡 minor — removing branch (c) orphaned two helpers.**

`VALUE_BEARING_KEYS` (`check_live_debug.py:51`) and `_leaves_under_value_bearing` (`:109`) are now unreferenced except in the docstring that explains the removal. Delete both.

**🟡 minor — `check_live_debug.py:307` prescribes the shape this PR fails on.**

```python
"Studio Web completes this shape but reads product back empty; "
"return { response: 6 * 7 } and map source=\"=result.response\"."
```

Twenty-five lines below, `:333` fails exactly that. Nobody reads it into a passing run, since `_fail` exits. But whoever debugs a failed eval reads it, follows it, and trades a `=result` failure for a double-wrap failure. Same class as the bug this stack removes.

```diff
-                        "return { response: 6 * 7 } and map source=\"=result.response\"."
+                        "return the value directly (`return 6 * 7;`) and map "
+                        "source=\"=result.response\"."
```

**🟡 minor — the task's own `description` still advertises the deleted assertion.**

`live_debug_e2e.yaml:9` (untouched by this PR) reads *"the check asserts ... the variables API exposed the `product` output definition"*. That is `_has_product_definition`. This is the field a future reader trusts to know what the task proves, and it currently describes the weakened contract #2214 introduced.

```diff
-  the debug run reached finalStatus Completed, the variables API exposed the
-  `product` output definition, and the authored BPMN computes 42 through a
-  script task mapped to that output.
+  the debug run reached finalStatus Completed, the runtime `product` variable
+  reads back 42, and the authored BPMN computes it through a script task
+  mapped to that output.
```

**On the body:** it says the agent *"chose `BPMN.Variables` and a bare return unprompted"*. The `BPMN.Variables` half is true and is the interesting half. The bare return is prompted verbatim: `live_debug_e2e.yaml:59` reads `script body: return 6 * 7;`. Worth trimming the claim to the half that holds.

## Recommendation

**Approve**, unchanged from last round. The check fails for the reason it exists, and the pass path no longer has a weak branch to lean on.

## tl;dr

Fix is stronger than what I asked for. Four leftovers still describe the deleted assertion, including the task's own `description`. All cosmetic.

PR #3214, opened 2026-09-10
test(uipath-maestro-bpmn): add variable-scope evals

DevMomo:
## Problem

**No test in this suite reads `elementId`** — 79 BPMN tasks, and not one check script looks at the attribute that decides whether a variable exists at all. They parse declarations by `id`, `name` and `type` only.

That is why a wrong variable-scoping rule shipped and survived seven weeks of green CI. The rule was wrong, the fixtures demonstrated the wrong shape, the one live test that touched it had been weakened to pass on a `null` — and nothing anywhere would go red.

---

**Stack 4/4** — base `fix/bpmn-live-debug-assert-value` (#3213). Bottom of the stack: #3211 → #3212 → #3213 → this.

Nothing in this suite read `elementId`, which is why a wrong scoping rule survived seven weeks of green CI. Two tasks close that gap, graded on the authored `.bpmn`.

- **`author/variable_scopes`** — one entry point: every declaration carries an `elementId`; each resolves to an element that exists; the caller-supplied input is start-event scoped; the variable a gateway reads is reachable from it.
- **`author/variable_scopes_nested`** — the scopes the single-entry case cannot reach: two start events each owning exactly their own caller input, a subprocess-owned variable declared in that subprocess's own block and keyed within it, and a published output keyed to a root end event.

### Design notes

- **Neither prompt contains the word `elementId`.** If it did, the task would measure instruction-following rather than the skill — which is exactly how `live_debug_e2e` stayed green while the skill taught the wrong shape.
- Every assertion reads the `.bpmn`. Generated metadata (`entry-points.json`) is deliberately **not** asserted: the CLI packager produces it, so grading it would couple these tests to another component instead of to the canvas contract.
- Three dimensions are deliberately left ungraded — node-keyed vs process-keyed task-output targets, the `custom` attribute, and self-keyed vs node-keyed subprocess variables. The canvas writes one form and merely *relocates* the other on import (migration 13) rather than rejecting it, so grading the difference would fail a file the platform accepts.
- **Both** tasks are tagged `smoke` so they actually gate PRs. `smoke-skills.yml` selects on `grep -qE '^tags:.*\bsmoke\b'`, so an `integration` tag would leave these scopes to nightly only — the same shape of gap as the one this stack exists to close. 12 of the 14 existing BPMN smoke tasks are `mode:build` + `lifecycle:generate` authoring tasks, and both of these are lighter than the rest, invoking no `uip` command at all.

### Verification

Local runs under `experiments/default.yaml` (tempdir, agent `claude-code`):

```
skill-bpmn-author-variable-scopes         SUCCESS  2/2  score 1.000  163s
skill-bpmn-author-variable-scopes-nested  SUCCESS  2/2  score 1.000  348s
```

The nested task **failed at 0.250 on its first run**, and the failure was mine rather than the agent's: the check demanded `custom="true"` on a self-keyed subprocess variable, a shape the skill does not teach and the platform does not require. Gold-and-mutation testing could not surface that, since I wrote both the gold and the mutations — only an agent reading the real skill could. The assertion now grades what is settled, and the mutation set still catches all six real defects.

### Review follow-ups

- `check_variable_scopes.py` now matches `vars.X` by declaration **id** only. The name fallback would have passed a file the canvas rejects — the same defect #3212 fixed in `OrderRouting.bpmn`.
- `variable_scopes_nested` retagged from `integration` to `smoke`; as `integration` it never ran on a PR, leaving the three scopes only it covers ungated.

Both agents authored correctly scoped files without being told the rule: `elementId="Event_start"` on the caller input, `elementId="Process_1"` on the process-level variable, `BPMN.Variables` on the script mapping.

### DIFF
--- a/tests/tasks/uipath-maestro-bpmn/author/variable_scopes/check_variable_scopes.py
+++ b/tests/tasks/uipath-maestro-bpmn/author/variable_scopes/check_variable_scopes.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+"""Variable-scope check.
+
+Every `<uipath:variables>` declaration must carry an `elementId` that resolves
+to a real element, and the scope must make the variable reachable where it is
+used: the caller-supplied input keyed to the start event, and a variable read by a gateway
+keyed either to the `<bpmn:process>` id or to the node whose mapping writes it
+(the canvas emits the latter for a task output — `VariableMutationUtils.ts`).
+
+Deliberately NOT asserted: the `custom` attribute, whose effect on canvas
+re-sync is unverified.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+import sys
+
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
+
+from _shared.bpmn_check import NS, fail, parse_bpmn, text_content  # noqa: E402
+
+DECL_TAGS = ("input", "output", "inputOutput")
+
+
+def declarations(root):
+    """Every declaration child of every `<uipath:variables>` block."""
+    out = []
+    for block in root.iter("{%s}variables" % NS["uipath"]):
+        for child in block:
+            if child.tag.split("}")[1] in DECL_TAGS:
+                out.append(child)
+    return out
+
+
+def condition_text(root):
+    return " ; ".join(
+        text_content(c)
+        for c in root.iter("{%s}conditionExpression" % NS["bpmn"])
+    )
+
+
+def main() -> None:
+    path, root = parse_bpmn("TierRouting")
+
+    process = root.find("bpmn:process", NS)
+    if process is None:
+        fail(f"{path}: no <bpmn:process>")
+    process_id = process.attrib.get("id")
+
+    # The canvas resolves an elementId against the process id plus flow-element
+    # and edge ids only; a DI shape id is an orphan and gets deleted on import
+    # (migrations/s191_2-migration-utils.ts).
+    all_ids = {process_id} | {
+        e.attrib["id"] for e in process.iter()
+        if e.attrib.get("id") and e.tag.startswith("{%s}" % NS["bpmn"])
+    }
+    start_ids = {e.attrib.get("id") for e in root.iter("{%s}startEvent" % NS["bpmn"])}
+
+    decls = declarations(root)
+    if not decls:
+        fail(f"{path}: no variable declarations in <uipath:variables>")
+
+    # 1. Every declaration carries an elementId.
+    unscoped = [d.attrib.get("id") or d.attrib.get("name") for d in decls
+                if not d.attrib.get("elementId")]
+    if unscoped:
+        fail(f"{path}: declarations without an elementId: {sorted(unscoped)}. "
+             "Every <uipath:variables> child needs one.")
+    print(f"OK: all {len(decls)} declarations carry an elementId")
+
+    # 2. Every elementId resolves to an element that exists.
+    dangling = {d.attrib.get("id"): d.attrib["elementId"] for d in decls
+                if d.attrib["elementId"] not in all_ids}
+    if dangling:
+        fail(f"{path}: elementId does not match any element in the file: {dangling}")
+    print("OK: every elementId resolves to a real element")
+
+    # 3. The caller-supplied input is start-event scoped. The canvas writes a
+    #    caller argument as a root `input` keyed to the start event that accepts
+    #    it (reducers/canvasSlice.ts updateRootVariablesForInput); any other
+    #    scope is not a caller-suppliable input.
+    amount = [d for d in decls
+              if (d.attrib.get("name") or "").lower() == "amount"
+              or (d.attrib.get("id") or "").lower().endswith("amount")]
+    if not amount:
+        fail(f"{path}: no declaration for the caller-supplied amount")
+    if not any(d.attrib["elementId"] in start_ids for d in amount):
+        got = {d.attrib.get("id"): d.attrib["elementId"] for d in amount}
+        fail(f"{path}: the caller-supplied amount must be scoped to the start event "
+             f"so the caller can pass it in; got {got}")
+    print("OK: caller-supplied amount is start-event scoped")
+
+    # 4. The variable the gateway reads is reachable from that gateway: either
+    #    process-scoped (globally available), or scoped to a node whose mapping
+    #    writes it via `var=` — which is the shape the canvas itself emits for a
+    #    task output (VariableMutationUtils.ts). Both are valid; a scope that is
+    #    neither is not reachable downstream.
+    conditions = condition_text(root)
+    if not conditions:
+        fail(f"{path}: no gateway conditionExpression found")
+    # Which element writes each variable: the nearest id-bearing ancestor of
+    # each mapping `<uipath:output var="...">`.
+    parents = {child: parent for parent in process.iter() for child in parent}
+    writers = {}
+    for m in process.iter("{%s}output" % NS["uipath"]):
+        var = m.attrib.get("var")
+        if not var:
+            continue
+        node = parents.get(m)
+        while node is not None and not node.attrib.get("id"):
+            node = parents.get(node)
+        if node is not None:
+            writers[var] = node.attrib["id"]
+    # Match by `id` only. The canvas resolves a `vars.X` reference against a
+    # declaration's id (and its canonicalId, which is backfilled only for
+    # legacy obfuscated ids), never its `name` — so accepting a name match here
+    # would pass a file the canvas rejects with VARIABLE_DOES_NOT_EXIST.
+    read_by_gateway = [
+        d for d in decls
+        if d.attrib.get("id")
+        and re.search(r"\bvars\.%s\b" % re.escape(d.attrib["id"]), conditions)
+    ]
+    if not read_by_gateway:
+        referenced = sorted(set(re.findall(r"\bvars\.([A-Za-z_$][\w$]*)", conditions)))
+        declared = sorted(d.attrib.get("id") or "?" for d in decls)
+        fail(f"{path}: no declared variable id is read by a gateway condition. "
+             f"Conditions reference {referenced}; declared ids are {declared}. "
+             "References resolve by id, not by name.")
+    unreachable = {}
+    for d in read_by_gateway:
+        eid = d.attrib["elementId"]
+        vid = d.attrib.get("id")
+        if eid == process_id or writers.get(vid) == eid:
+            continue
+        unreachable[vid] = eid
+    if unreachable:
+        fail(f"{path}: a variable read by a gateway condition must be scoped to the "
+             f"process id {process_id!r}, or to the node whose mapping writes it; "
+             f"got {unreachable}")
+    print(f"OK: gateway-read variable(s) reachable from the gateway "
+          f"({ {d.attrib.get('id'): d.attrib['elementId'] for d in read_by_gateway} })")
+
+    print("PASS: all variable-scope checks passed")
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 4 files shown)


tests/tasks/uipath-maestro-bpmn/author/variable_scopes_nested/variable_scopes_nested.yaml
```diff
@@ -0,0 +1,49 @@
+task_id: skill-bpmn-author-variable-scopes-nested
+description: >
+  Authoring eval for the variable scopes the single-entry case cannot reach: two
+  start events each owning their own caller-supplied input (with no input
+  leaking onto the other entry point), a variable owned by a subprocess and
+  declared in that subprocess's own `<uipath:variables>` block, and a published
+  output keyed to a root end event. The prompt never mentions `elementId` or
+  scoping, so the task fails if the skill stops teaching the contract.
+tags: [uipath-maestro-bpmn, integration, "mode:build", lifecycle:generate, shape:multi-node, node:script, node:subprocess]
```
rockymadden: **🟠 major — this task does not gate PRs.**

`integration`, not `smoke`. `smoke-skills.yml` selects with `grep -qE '^tags:.*\bsmoke\b'`, so this never runs on a PR that changes the skill. Nightly is the only thing that reads it.

That leaves ungated exactly the three scopes the single-entry task cannot reach:

- per-start-event caller inputs (two entries, no leakage)
- a subprocess-owned declaration in the subprocess's own block
- an end-event-keyed published output

Those three have no other guard anywhere in the suite. And the failure mode this whole stack exists to fix is *"nothing in this suite read `elementId`, which is why a wrong scoping rule survived seven weeks of green CI"*. Nightly-only coverage is the same shape of gap, just slower.

The PR body reads as though both tasks are smoke: *"Tagged `smoke` so the guard actually runs on PRs ... this one is lighter than the rest, invoking no `uip` command at all."* That is true of `variable_scopes` and not of this one, and the reader cannot tell.

Two ways out, either is fine:

1. Tag it `smoke` too. The per-PR cap is 40 (`SMOKE_MAX_TASKS`) and this skill has 14 smoke tasks today, so there is headroom.
2. Keep it nightly and say so in the body, with the reason. `348s` against `163s` is a fair reason. Just state it rather than leaving the paragraph reading as if it covers both.

tests/tasks/uipath-maestro-bpmn/author/variable_scopes/check_variable_scopes.py
```diff
@@ -0,0 +1,142 @@
...
+        node = parents.get(m)
+        while node is not None and not node.attrib.get("id"):
+            node = parents.get(node)
+        if node is not None:
+            writers[var] = node.attrib["id"]
+    read_by_gateway = [
+        d for d in decls
+        if any(re.search(r"\bvars\.%s\b" % re.escape(ident), conditions)
+               for ident in (d.attrib.get("id"), d.attrib.get("name")) if ident)
+    ]
```
rockymadden, line 124: **🟡 minor — matching by `name` hides the exact bug #3212 just fixed.**

`read_by_gateway` resolves `vars.X` against a declaration's `id` **or** its `name`. The canvas resolves by `id` only for hand-authored files:

- `VariableValidationUtils.ts:180` matches `[availableVariable.id, availableVariable.canonicalId]`.
- `canonicalId` is backfilled only for legacy obfuscated ids: `s182-migrations.ts` gates on `OBFUSCATED_ID_REGEX = /^v[a-zA-Z0-9]{8}$/`, which nothing authored here produces.

So an agent that declares `id="Var_Tier" name="tier"` and writes `=vars.tier == "high"` produces a file the canvas rejects with `VARIABLE_DOES_NOT_EXIST`, and this check passes it: the name match finds the declaration, its `elementId` is fine, done.

That is the same defect #3212 fixed in `OrderRouting.bpmn` (`vars.amount` against `id="Var_Amount"`), described there as *"Nothing had caught this"*. This eval would not catch it either.

Two changes, both small:

1. Drop `d.attrib.get("name")` from line 119. Match ids only.
2. Add the converse: every `vars.<X>` appearing in a `conditionExpression` must resolve to a declared **id**, otherwise fail as a dangling reference.

(2) is what actually closes it. (1) alone just turns the pass into `no declared variable is read by a gateway condition`, which is a misleading message for what is really a dangling reference.

rockymadden (commented):
## What

Two authoring evals graded on the authored `.bpmn`, closing the gap that let a wrong scoping rule sit green for seven weeks:

- `author/variable_scopes` — single entry point, smoke-tagged.
- `author/variable_scopes_nested` — two entries, subprocess-owned variable, published output. Integration-tagged.

## Verified

- ✅ **Neither prompt contains `elementId` or the word "scope".** The tasks measure the skill, not instruction-following. This is the right call and it is the exact reason `live_debug_e2e` stayed green while the skill taught the wrong shape.
- ✅ **Rule 4 is correct to accept a node-keyed variable downstream.** `VariableScopeUtils.getAvailableVariablesFromSourceNodes` walks upstream edges and folds in each source node's outputs, so a variable keyed to the writing node really is in scope at a later gateway. Not an over-permissive guess.
- ✅ **The nested check's self-keyed / node-keyed leniency matches the platform.** `s191_2-migration-utils.ts` relocates a self-keyed non-custom subprocess variable rather than rejecting it. Grading the difference would fail files the canvas accepts, exactly as the docstring says.
- ✅ **`all_ids` correctly excludes DI shape ids** (built only from `bpmn:`-namespaced elements under `<bpmn:process>`), so a variable keyed to a `BPMNShape` id is caught as dangling rather than waved through.
- ✅ Both checks import the shared `_shared/bpmn_check` helpers with the right signatures.

## Could not verify

The two local runs (`163s` and `348s`, 2400s timeouts) and the mutation results. Not runnable here, and I did not run anything.

## Overall findings

**🟡 minor — the mutation set is not in the diff.**

The confidence in this PR rests on *"the mutation set still catches all six real defects"*, and nothing in the repo lets anyone re-run that. `author/test_simple_approval_bpmn.py` is the precedent for committing a checker's unit test: 46 lines, imports the checker directly, builds XML fragments inline.

Only 2 of the 72 BPMN checkers have one today, so this is not a convention breach. It is an unverifiable claim, in a PR whose entire subject is unverifiable claims surviving CI. Six mutations as pytest cases would fix that and cost very little.

## Recommendation

**Comment.** The tasks grade the right thing at the right altitude, and the honesty about the first nested run failing at 0.250 for the author's own reason is worth more than a clean report. Two gaps inline.

## tl;dr

Good evals, right decision to keep `elementId` out of the prompts. The nested half does not gate PRs, and the single-entry half will pass a reference shape the canvas rejects.

tests/tasks/uipath-maestro-bpmn/author/variable_scopes_nested/variable_scopes_nested.yaml
DevMomo: Retagged `smoke`. You were right about the body too — it claimed both tasks were smoke when only `variable_scopes` was; the description now says both are, and why an `integration` tag would leave these three scopes to nightly only.

Confirmed against the suite before changing it: 12 of the 14 existing BPMN smoke tasks are `mode:build` + `lifecycle:generate` authoring tasks, so this is in line with them, and both of these invoke no `uip` command at all.

tests/tasks/uipath-maestro-bpmn/author/variable_scopes/check_variable_scopes.py
DevMomo: Dropped the `name` fallback — matching is now by declaration `id` only, and the failure message prints the referenced ids against the declared ids so a mismatch is obvious.

You're right that this was the same defect #3212 fixes in `OrderRouting.bpmn`, and that the eval would have passed it. Settled the underlying question empirically too: a condition reading `vars.tier` against a declaration whose id is `Var_Tier` raises `VARIABLE_DOES_NOT_EXIST` in the canvas engine, so id-only is the correct rule.

tests/tasks/uipath-maestro-bpmn/author/variable_scopes/check_variable_scopes.py
```diff
@@ -0,0 +1,149 @@
...
+    # Match by `id` only. The canvas resolves a `vars.X` reference against a
+    # declaration's id (and its canonicalId, which is backfilled only for
+    # legacy obfuscated ids), never its `name` — so accepting a name match here
+    # would pass a file the canvas rejects with VARIABLE_DOES_NOT_EXIST.
+    read_by_gateway = [
+        d for d in decls
+        if d.attrib.get("id")
+        and re.search(r"\bvars\.%s\b" % re.escape(d.attrib["id"]), conditions)
+    ]
+    if not read_by_gateway:
```
rockymadden, line 125: 🟡 **minor** — the converse check is still partial.

The id-only match plus the new message closes the case this task actually produces, so this is not a blocker. But the guard is *"at least one declared id is read"*, not *"every `vars.X` resolves"*.

Concrete miss. Conditions `=vars.Var_Tier == "high"` and `=vars.amount >= 1000`, declarations `Var_Tier` and `Var_Amount`:

- `read_by_gateway` finds `Var_Tier`, so it is non-empty and this branch never fires.
- `vars.amount` is dangling. The canvas rejects the file with `VARIABLE_DOES_NOT_EXIST`.
- Check passes.

Three lines closes it, using the ids you already compute:

```python
referenced = set(re.findall(r"\bvars\.([A-Za-z_$][\w$]*)", conditions))
declared_ids = {d.attrib["id"] for d in decls if d.attrib.get("id")}
if referenced - declared_ids:
    fail(f"{path}: gateway conditions reference undeclared variable id(s): "
         f"{sorted(referenced - declared_ids)}. Declared ids are "
         f"{sorted(declared_ids)}; references resolve by id, not by name.")
```

rockymadden (approved):
## What

Two authoring evals graded on the authored `.bpmn`, closing the gap that let a wrong scoping rule sit green for seven weeks.

## Prior round: both addressed, and the retag is proven

- ✅ **Retagged `smoke`**, and the body no longer implies both were already smoke.
- ✅ **I confirmed the retag actually works rather than trusting it.** CI run `34522066597` on this PR selected and ran both:

```
skill-bpmn-author-variable-scopes         status=SUCCESS  score=1.00  103.6s
skill-bpmn-author-variable-scopes-nested  status=SUCCESS  score=1.00   86.5s
```

That is stronger evidence than the local `163s` / `348s` in the body, and it settles the cost objection: both are far cheaper on the runner than measured locally. 15 BPMN smoke tasks now, cap is 40.

- ✅ **Name fallback dropped.** Matching is by declaration `id` only, and the failure message prints referenced ids against declared ids.

## Verified

- ✅ Neither prompt contains `elementId` or `scope`. Still the right call.
- ✅ The skill does teach the shape rule 5 grades: `structural-bpmn.md:136` keys a published output to the end event, `:543` requires each `uipath:output` to match a root end event. So rule 5 is not measuring an unguided guess.
- ✅ `all_ids` filters to `bpmn:`-namespaced elements under `<bpmn:process>`, so a DI shape id is caught as dangling. In the nested check, `inner_ids` skips that filter, but rule 2 runs first and dominates, so nothing leaks.
- ✅ Both checks import `_shared/bpmn_check` with the right signatures. `success_criteria` weights and `$TASK_DIR` match `author/gateway_sequence_flows.yaml`.

## Overall findings

**🟡 minor — the smoke-task count in the body is off by one.**

13 existing BPMN smoke tasks, 11 of them `mode:build` + `lifecycle:generate`, not "12 of the 14". Counted at this branch head, excluding these two. The argument is unaffected.

**🟡 outstanding, non-blocking — the mutation set is still not in the diff.**

I raised it last round as minor and it is not in the follow-ups. Two green CI runs now cover the happy path, so the gap is narrower than it was: what stays unverifiable is that the checks *fail* on the six defects. Precedent for committing it is `author/test_simple_approval_bpmn.py`. Your call whether it is worth doing now.

One more inline.

## Recommendation

**Approve.** The tasks grade the right thing at the right altitude, both gate PRs, and I watched CI run them green.

## tl;dr

Retag verified in CI, not just claimed. The dangling-reference check closes the single-reference case but not the general one. Mutation set still uncommitted.

PR #3216, opened 2026-09-10
test(uipath-maestro-bpmn): grade customer_escalation wherever the project lands

nikhil-maryala:
## Summary
`skill-bpmn-e2e-customer-escalation` FAILED in the `2026-09-10` nightly. It grades **process modeling**, but a brittle exact-path check failed it on **project location** before any modeling was evaluated.

## Root cause
`uip maestro bpmn init` outside a solution **auto-scaffolds `<Name>Solution/<Name>/`** by default (`references/shared/project-layout.md:16-33`). The agent used that default, producing `CustomerEscalationSolution/CustomerEscalation/CustomerEscalation.bpmn`. The three `file_exists` criteria and the check hardcoded the top-level path `CustomerEscalation/...`, so all failed on location — the check printed `FAIL: project` before checking the classifiers, gateway, user-task path, or ticket script. **The check contradicted the skill's own documented default scaffold.**

## Fix (location-robust; layout deliberately ungraded)
Decision (with @rockymadden): grade the modeling wherever the project lands, rather than pin the layout — and roll the same robustness to the ~32 sibling tasks that share this latent failure (follow-up **#3219**).

- `check_customer_escalation.py`: `resolve_project()` locates the **unique** `CustomerEscalation.bpmn` **with `project.uiproj` beside it** (fails on 0 or >1), so a stray/draft copy is never graded. Project dir threaded as a local; failure messages carry the path.
- `customer_escalation.yaml`: three exact-path `file_exists` → `rglob` `run_command` checks; the gating weight-5 check's single-candidate guard prevents any two-copy false-pass.
- `description:` states layout is deliberately ungraded.

## Verification
- **Local:** the resolver picks the correct project on wrapper, bare, and **two-copy** (bare-draft + wrapper) trees.
- **CI:** re-verifying on **claude** (34517642443) and **codex** (34517645796). Honest caveat: CI grades whatever layout the agent happens to produce that run (the first claude run used a top-level layout, so the wrapper path was exercised only by the local test above, not CI). #3219 adds a `tmp_path` unit test to make the resolver CI-enforced.

### DIFF
--- a/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
+++ b/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
@@ -31,7 +31,10 @@ from _shared.bpmn_check import (  # noqa: E402
     require_sequence_integrity,
 )
 
-PROJECT = Path.cwd() / "CustomerEscalation"
+# PROJECT is resolved in main() from the located BPMN's parent, so the check
+# grades the process wherever the agent placed the project dir (top level or
+# nested under a Solution wrapper) — it grades the modeling, not the layout.
+PROJECT: Path = Path.cwd() / "CustomerEscalation"
 BPMN_NAME = "CustomerEscalation.bpmn"
 REQUIRED_FILES = [
     "project.uiproj",
@@ -53,11 +56,14 @@ def load_json(name: str):
 
 
 def main() -> None:
+    global PROJECT
+    path, root = parse_bpmn("CustomerEscalation")
+    PROJECT = Path(path).parent
+
     for name in REQUIRED_FILES:
         if not (PROJECT / name).is_file():
             fail(f"{name} is missing")
 
-    path, root = parse_bpmn("CustomerEscalation")
 
     process = one_or_more(root, "process")[0]
 
--- a/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
+++ b/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 import json
 import os
 import sys
+import xml.etree.ElementTree as ET
 from pathlib import Path
 
 _d = os.path.dirname(os.path.abspath(__file__))
@@ -25,16 +26,11 @@ from _shared.bpmn_check import (  # noqa: E402
     elements,
     fail,
     one_or_more,
-    parse_bpmn,
     require_di_for_visible_elements,
     require_no_private_connector_values,
     require_sequence_integrity,
 )
 
-# PROJECT is resolved in main() from the located BPMN's parent, so the check
-# grades the process wherever the agent placed the project dir (top level or
-# nested under a Solution wrapper) — it grades the modeling, not the layout.
-PROJECT: Path = Path.cwd() / "CustomerEscalation"
 BPMN_NAME = "CustomerEscalation.bpmn"
 REQUIRED_FILES = [
     "project.uiproj",
@@ -45,25 +41,44 @@ REQUIRED_FILES = [
 ]
 
 
-def load_json(name: str):
-    p = PROJECT / name
+def load_json(project: Path, name: str):
+    p = project / name
     if not p.is_file():
-        fail(f"{name} is missing")
+        fail(f"{name} is missing beside {p}")
     try:
         return json.loads(p.read_text(encoding="utf-8"))
     except Exception as exc:  # noqa: BLE001
         fail(f"{name} is not valid JSON: {exc}")
 
 
+def resolve_project() -> Path:
+    # Grade the project wherever the agent placed it (top level or nested under a
+    # Solution wrapper), but pick the real project unambiguously: exactly one
+    # CustomerEscalation.bpmn with project.uiproj beside it. This avoids grading a
+    # stray draft copy when two same-named files exist (find_bpmn_file matches on
+    # a substring and returns the alphabetically-first).
+    candidates = [
+        p for p in Path.cwd().rglob(BPMN_NAME) if (p.parent / "project.uiproj").is_file()
+    ]
+    if len(candidates) != 1:
+        fail(
+            f"expected exactly one {BPMN_NAME} with project.uiproj beside it, "
+            f"found {[str(p) for p in candidates]}"
+        )
+    return candidates[0].parent
+
+
 def main() -> None:
-    global PROJECT
-    path, root = parse_bpmn("CustomerEscalation")
-    PROJECT = Path(path).parent
+    project = resolve_project()
+    bpmn_file = project / BPMN_NAME
+    try:
+        root = ET.parse(bpmn_file).getroot()
+    except Exception as exc:  # noqa: BLE001
+        fail(f"{bpmn_file} is not well-formed XML: {exc}")
 
     for name in REQUIRED_FILES:
-        if not (PROJECT / name).is_file():
-            fail(f"{name} is missing")
-
+        if not (project / name).is_file():
+            fail(f"{name} is missing beside {bpmn_file}")
 
     process = one_or_more(root, "process")[0]
 
@@ -104,10 +119,10 @@ def main() -> None:
 
     # Package metadata references the BPMN file.
     for name in ("entry-points.json", "operate.json", "package-descriptor.json"):
-        if BPMN_NAME not in json.dumps(load_json(name)):
+        if BPMN_NAME not in json.dumps(load_json(project, name)):
             fail(f"{name} must reference {BPMN_NAME}")
 
-    print(f"OK: {path} classifies, routes, escalates via a user task, and ships consistent package metadata")
+    print(f"OK: {bpmn_file} classifies, routes, escalates via a user task, and ships consistent package metadata")
 
 
 if __name__ == "__main__":
--- a/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/customer_escalation.yaml
+++ b/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/customer_escalation.yaml
@@ -46,21 +46,27 @@ initial_prompt: |
   Build the complete local project end-to-end in a single pass.
 
 success_criteria:
-  - type: file_exists
+  - type: run_command
     description: "BPMN source file was created"
-    path: "CustomerEscalation/CustomerEscalation.bpmn"
+    command: "python3 -c \"from pathlib import Path; import sys; sys.exit(0 if any(Path('.').rglob('CustomerEscalation.bpmn')) else 'CustomerEscalation.bpmn missing')\""
+    timeout: 30
+    expected_exit_code: 0
     weight: 1.5
     pass_threshold: 1.0
 
-  - type: file_exists
+  - type: run_command
     description: "Project metadata file was created"
-    path: "CustomerEscalation/project.uiproj"
+    command: "python3 -c \"from pathlib import Path; import sys; b=next(Path('.').rglob('CustomerEscalation.bpmn'), None); sys.exit(0 if b and (b.parent / 'project.uiproj').is_file() else 'project.uiproj missing beside CustomerEscalation.bpmn')\""
+    timeout: 30
+    expected_exit_code: 0
     weight: 1.0
     pass_threshold: 1.0
 
-  - type: file_exists
+  - type: run_command
     description: "Package operate metadata was created"
-    path: "CustomerEscalation/operate.json"
+    command: "python3 -c \"from pathlib import Path; import sys; b=next(Path('.').rglob('CustomerEscalation.bpmn'), None); sys.exit(0 if b and (b.parent / 'operate.json').is_file() else 'operate.json missing beside CustomerEscalation.bpmn')\""
+    timeout: 30
+    expected_exit_code: 0
     weight: 1.0
     pass_threshold: 1.0
 
--- a/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/customer_escalation.yaml
+++ b/tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/customer_escalation.yaml
@@ -6,7 +6,10 @@ description: >
   combined signal through an exclusive gateway, and escalates via a human user
   task on the high-touch path while a standard path generates a ticket in a
   script task. Includes full package metadata. Authoring only, public-safe, no
-  cloud effects and no live connector.
+  cloud effects and no live connector. Project layout is deliberately ungraded:
+  the checks locate the project by the `CustomerEscalation.bpmn` + `project.uiproj`
+  pair wherever it lands, so the default `uip maestro bpmn init`
+  `<Name>Solution/<Name>/` scaffold and a bare `<Name>/` both grade the modeling.
 tags: [uipath-maestro-bpmn, e2e, "mode:build", "lifecycle:generate", "shape:multi-node", "node:decision", "node:hitl", "feature:escalation"]
 
 run_limits:


rockymadden (commented):
## 📋 Summary

**What it does**

- `check_customer_escalation.py`: derives `PROJECT` from the located BPMN's parent instead of a hardcoded `cwd/CustomerEscalation`.
- `customer_escalation.yaml`: swaps three exact-path `file_exists` criteria for `run_command` python one-liners that `rglob` the BPMN and check `project.uiproj` / `operate.json` beside it.
- Goal: stop failing the task on project **location** so it grades **modeling**.

**The diagnosis is correct, and I confirmed it independently.** The failing nightly's artifacts in blob storage (`coderevaltests/runs/2026-09-10_04-18-49/default/skill-bpmn-e2e-customer-escalation/00/artifacts/.../CustomerEscalationSolution/CustomerEscalation/CustomerEscalation.bpmn`) show the Solution wrapper. And `references/shared/project-layout.md:16-33` documents that `uip maestro bpmn init` outside a solution auto-scaffolds `<Name>Solution/<Name>/`, so the old exact path contradicted the skill's own documented scaffold.

### Overall findings

- 🔴 **major, layout intent is left unresolved.** `SKILL.md:190-192` tells the agent to pass `--skip-solution-registration` "for a source-only draft the user has not asked to package or operate". This prompt forbids publish/deploy but asks for the package metadata set, so both layouts are arguable and the eval now grades neither. Two clean exits: (a) pin the prompt ("initialize with `--skip-solution-registration`; do not create a Solution wrapper") and keep the exact paths, which is deterministic and matches ~30 sibling tasks; or (b) keep it agnostic and say so in `description:` so the next reader knows layout is deliberately ungraded. Right now the skill rule silently goes unenforced here.

- 🔴 **major, the "inconsistent one" claim in the description is backwards.** 32 bpmn task files still assert a directory-prefixed exact `.bpmn` path (`patterns/*/greenfield.yaml`, `multi_node/calculator`, `single_node/agent_job`, `nodes/queue_create_and_wait`, ...), plus 24 asserting a bare root filename. `wiki_pageviews` is the exception, not the norm. Every greenfield one is a single layout decision away from this exact false negative. Either state why only this task moves now, or open a follow-up for the rest.

- 🔴 **major, the cited verification does not show what the description says it shows.** In run 34508512653 the checker printed `OK: CustomerEscalation/CustomerEscalation.bpmn`: the agent used the **top-level** layout, so the wrapper path was never exercised and the old criteria would have passed too. It also ran `claude`, while `run-coder-eval.yml:84-95` says codex is the harness the nightly runs daily (claude twice a week). Re-dispatch with `agent: codex` to actually reproduce the failing configuration, or reword the description so the local simulation is the only wrapper evidence claimed.

- 🟡 **minor, no test pins the new behavior.** `pytest tests/tasks/uipath-maestro-bpmn/` runs in CI (`test-helpers.yml:37`) and `_shared/test_bpmn_check.py` already exists. Extracting the resolver into `_shared/bpmn_check.py` lets a `tmp_path` test assert wrapper, bare, and two-copy trees, which replaces "verified locally" with something that stays verified.

- 🟡 **minor, the title reads backwards.** "grade customer_escalation by project location, not exact path" describes the old behavior. It now grades *regardless* of location. Suggest: `test(uipath-maestro-bpmn): grade customer_escalation wherever the project lands`.

- ⚪ **Reasoned vs confirmed.** Confirmed: the nightly layout (blob listing), the skill's documented auto-solution scaffold, the CI run's per-criterion results, and the selection behavior of both new code paths (ran the checker against a two-copy tree locally). **Could not verify:** that nightly's per-criterion scores, or that the modeling criteria pass under the wrapper layout, because reading its `task.json` from `coderevaltests` needs an interactive `az login --scope https://storage.azure.com/.default`. So "the underlying modeling actually passes under a wrapper" is still unproven.

- ✅ No AI slop in the description. It is specific and mostly evidence-backed, apart from the verification overclaim above.

### Recommendation

**Comment.** Direction is right and it is strictly better than the status quo, which fails 100% of wrapper runs. Two of the inline fixes are a few lines and should land here, since the multi-match one can trade the old false negative for a new one in the same nightly.

**tl;dr**: Real bug, right direction, and the nightly artifacts back the diagnosis. Three things to settle: the checker can now grade the wrong BPMN when a second copy exists (one-line guard, copy it from `wiki_pageviews`), the layout intent is left unstated while `SKILL.md` has a rule about it, and the cited CI run graded a top-level layout under `claude`, so it verifies neither the wrapper fix nor the codex nightly. 32 sibling tasks still carry the same latent failure.

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
```diff
@@ -53,11 +56,14 @@ def load_json(name: str):
 
 
 def main() -> None:
+    global PROJECT
+    path, root = parse_bpmn("CustomerEscalation")
```
rockymadden: 🔴 **major: picks an arbitrary BPMN, and the exact-name guard is missing**

`find_bpmn_file` matches the hint as a **substring** of the basename and returns `sorted(...)[0]`. Two consequences:

- A stray `CustomerEscalationDraft.bpmn` gets graded as the project. `check_wiki_pageviews.py:57-58`, the file this PR mirrors, guards exactly this and the guard was not copied over.
- With two same-named copies the alphabetically-first wins, which is the bare draft, not the wrapper project. Not hypothetical: in the **same nightly**, `skill-bpmn-e2e-live-debug` produced both `ProductCalc/ProductCalc.bpmn` and `ProductCalcSolution/ProductCalc/ProductCalc.bpmn`. I ran this checker against that shape locally and it graded the bare copy and failed.

So this can trade the old location false negative for a new one.

```python
path, root = parse_bpmn("CustomerEscalation")
bpmn = Path(path)
if bpmn.name != BPMN_NAME:
    fail(f"expected a BPMN named {BPMN_NAME}, found {bpmn}")

candidates = [p for p in Path().rglob(BPMN_NAME) if (p.parent / "project.uiproj").is_file()]
if len(candidates) != 1:
    fail(f"expected exactly one {BPMN_NAME} with project metadata beside it, found {candidates}")
project = candidates[0].parent
```

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
```diff
@@ -31,7 +31,10 @@
     require_sequence_integrity,
 )
 
-PROJECT = Path.cwd() / "CustomerEscalation"
+# PROJECT is resolved in main() from the located BPMN's parent, so the check
+# grades the process wherever the agent placed the project dir (top level or
+# nested under a Solution wrapper) — it grades the modeling, not the layout.
+PROJECT: Path = Path.cwd() / "CustomerEscalation"
```
rockymadden: 🟡 **minor: mutable global diverges from the sibling it mirrors**

`check_wiki_pageviews.py:61` and `author/check_simple_approval_bpmn.py:37` both keep the project dir as a local and pass it in (`load_json(project, name)`); `global` appears nowhere else in the bpmn checkers. Thread the value instead: `PROJECT = Path.cwd() / "CustomerEscalation"` here is now dead, and this three-line comment exists only to explain the global, so both disappear with it.

Also: `global PROJECT` on line 59, and the leftover double blank line at 66-67.

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
```diff
@@ -53,11 +56,14 @@ def load_json(name: str):
 
 
 def main() -> None:
+    global PROJECT
+    path, root = parse_bpmn("CustomerEscalation")
+    PROJECT = Path(path).parent
+
     for name in REQUIRED_FILES:
         if not (PROJECT / name).is_file():
             fail(f"{name} is missing")
```
rockymadden: 🟡 **minor: the failure message lost its location**

With `PROJECT` derived, `FAIL: operate.json is missing` no longer says where the checker looked, which is the one thing you need when triaging a nightly. `wiki_pageviews` says "missing beside". Same applies to `load_json` on line 51.

```python
fail(f"{name} is missing beside {path}")
```

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/customer_escalation.yaml
```diff
@@ -46,21 +46,27 @@ initial_prompt: |
   Build the complete local project end-to-end in a single pass.
 
 success_criteria:
-  - type: file_exists
+  - type: run_command
     description: "BPMN source file was created"
-    path: "CustomerEscalation/CustomerEscalation.bpmn"
+    command: "python3 -c \"from pathlib import Path; import sys; sys.exit(0 if any(Path('.').rglob('CustomerEscalation.bpmn')) else 'CustomerEscalation.bpmn missing')\""
```
rockymadden, line 54: 🟡 **minor: three duplicated one-liners, and a second selection rule**

These use `next(Path('.').rglob(...), None)`, which is `scandir` order, while the checker uses `sorted(glob(...))[0]` with substring matching. Two rules for "the project" can select different files, so these granular criteria and the weight-5.0 check can disagree on the same tree. Escaped-quote python inside a YAML double-quoted string is also hard to read and hard to grep.

One shared script keeps a single rule and stays legible, and `wiki_pageviews` could adopt it:

```yaml
command: "python3 $TASK_DIR/../../_shared/require_beside_bpmn.py CustomerEscalation project.uiproj"
```

Same for the two sibling criteria on lines 59 and 67.

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
nikhil-maryala: Fixed (0a256a130). Added a `resolve_project()` that selects the exactly-one `CustomerEscalation.bpmn` **with `project.uiproj` beside it**, failing on 0 or >1 — so a stray draft copy without metadata is never graded. Verified locally against your two-copy shape (bare top-level draft + wrapper-with-project.uiproj): it now picks the wrapper project. Thanks for the concrete repro.

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
nikhil-maryala: Fixed. Dropped the mutable global — `project` is now a local threaded through `load_json(project, name)`, matching `wiki_pageviews` / `check_simple_approval_bpmn`. The dead `PROJECT` assignment and the explanatory comment are gone with it.

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
nikhil-maryala: Fixed. Failure messages now include the path: `{name} is missing beside {bpmn_file}` (and the same in `load_json`).

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/customer_escalation.yaml
nikhil-maryala: The three one-liners still use `next(rglob)`, but they can no longer cause a false PASS: the weight-5 `resolve_project()` now fails on any tree with 0 or >1 project, so a two-copy tree fails the gating check regardless of which copy the light criteria pick. Agreed a shared `_shared/require_beside_bpmn.py` is cleaner and greppable — happy to extract it (and adopt in wiki_pageviews) as a small follow-up if you'd prefer it in this PR.

nikhil-maryala:
@rockymadden — thanks, this was a strong review. Addressed:

- **🔴 multi-copy false-positive** (0a256a130): `resolve_project()` now selects the unique `CustomerEscalation.bpmn` with `project.uiproj` beside it and fails on 0 or >1. Verified against your two-copy repro (bare draft + wrapper) — picks the wrapper. Global dropped/threaded; messages carry the path.
- **🔴 layout intent:** decided to keep it location-robust (the wrapper is the *default* `init` scaffold per `project-layout.md`, so pinning would contradict the skill) and documented "layout deliberately ungraded" in `description:`.
- **🔴 32 siblings:** opened **#3219** to extract a `_shared` resolver + criteria helper + `tmp_path` test and roll it out; layout-sensitive tasks stay exact.
- **🔴 verification overclaim:** reworded the body — the CI run graded a top-level layout, so the wrapper path is proven by the local two-copy test, not CI. Re-verifying on **codex** (34517645796) + claude (34517642443).
- **🟡 title:** now "grade customer_escalation wherever the project lands."
- **🟡 shared one-liner script:** the weight-5 guard prevents a two-copy false-pass, so it's safe today; the `_shared` extraction is folded into #3219.

Will post the codex/claude results when they land.

tests/tasks/uipath-maestro-bpmn/multi_node/customer_escalation/check_customer_escalation.py
```diff
@@ -42,22 +41,44 @@
...
+def resolve_project() -> Path:
+    # Grade the project wherever the agent placed it (top level or nested under a
+    # Solution wrapper), but pick the real project unambiguously: exactly one
+    # CustomerEscalation.bpmn with project.uiproj beside it. This avoids grading a
+    # stray draft copy when two same-named files exist (find_bpmn_file matches on
+    # a substring and returns the alphabetically-first).
+    candidates = [
+        p for p in Path.cwd().rglob(BPMN_NAME) if (p.parent / "project.uiproj").is_file()
+    ]
+    if len(candidates) != 1:
```
rockymadden, line 63: 🟡 **minor: a legitimate two-project tree now hard-fails**

This is the guard I asked for, so it lands as intended — but the strictness has a cost worth recording on #3219. An agent that scaffolds bare, then re-runs `init` and gets the Solution wrapper, leaves two dirs each with `project.uiproj`. I built that tree: the checker exits `FAIL: expected exactly one … found [bare, wrapper]` and grades nothing.

That trades the old location false-negative for an ambiguity false-negative. Loud beats silently-wrong, so this is the right call for now. When the resolver moves into `_shared/bpmn_check.py` under #3219, consider preferring the deepest candidate over failing:

```python
if len(candidates) > 1:
    candidates = [max(candidates, key=lambda p: len(p.parts))]
```

No change needed here.

rockymadden (approved):
## 📋 Summary

**What changed since my last pass** — 3 commits, `c3e9329e9` → `3a4547c50`:

- `resolve_project()` picks the **unique** `CustomerEscalation.bpmn` with `project.uiproj` beside it; fails on 0 or >1. Replaces the `find_bpmn_file` substring match.
- `project` threaded as a local (`load_json(project, name)`); the mutable global is gone.
- Failure messages carry the path again ("missing beside …").
- `description:` states layout is deliberately ungraded.
- Title now says "wherever the project lands".

**All four inline findings are fixed. I re-verified each one rather than taking the replies at face value.**

### ✅ Confirmed

| Check | Result |
| --- | --- |
| Wrapper-only tree | resolves `CustomerEscalationSolution/CustomerEscalation` |
| Bare-only tree | resolves `CustomerEscalation` |
| Bare draft (no `project.uiproj`) + wrapper | resolves the **wrapper** — my `live-debug` repro no longer mis-grades |
| Two full projects | fails loudly, lists both paths — no silent wrong-file grading |
| Zero projects | fails with `found []` |

- 🔁 **Re-verification landed on both harnesses.** `34517645796` (**codex**, the nightly harness — my ask) and `34517642443` (claude): both **5/5 criteria, score 1.000**. Honest caveat, and the body states it: both agents produced the **top-level** layout, so CI still has not exercised the wrapper. The local matrix above is the only wrapper evidence, and it holds.
- 📄 **`description:` is the sanctioned place for this.** `.claude/commands/lint-task.md:159` carves out exactly this: a `description` documenting a deliberate trade-off is credited, not penalised. Option (b) from my last pass, done the way the repo wants it.
- 🔗 **#3219 is real and scoped correctly** — the 32 exact-prefixed siblings, the `_shared/bpmn_check.py` resolver extraction, the shared `require_beside_bpmn.py`, and the `tmp_path` unit test that makes the resolver CI-enforced.
- 🧪 `ET.parse` inlined is behaviourally identical to `parse_bpmn` — same call, same `NS` map, broader `except`. No namespace regression.
- 🟢 All 24 CI checks pass, including **maestro-bpmn checker unit tests**. I read the tests, did not run them.

### Overall findings

- 🟡 **minor — the three light criteria still carry a second selection rule.** They use `next(Path('.').rglob(...), None)`; the checker uses the unique-pair rule. They can pick different files on an ambiguous tree. **Not a false-pass risk** — I confirmed the weight-5 gate fails first on any 0-or->1 tree, so the task cannot score full. #3219 already owns the shared helper. Fine as a follow-up.
- 🟡 **minor — the wrapper path is still CI-unverified.** Two more dispatches did not reproduce it, and dispatching until the agent happens to pick the wrapper is not a plan. #3219's `tmp_path` test is the right close; until it lands, the resolver's wrapper branch is verified only by our two local matrices.
- ⚪ **Reasoned vs confirmed.** Confirmed: every table row above, the two eval runs' per-criterion scores and layouts, `lint-task.md:159`, #3219's contents, CI. **Could not verify:** the failing nightly's per-criterion scores — reading `task.json` from `coderevaltests` still needs an interactive `az login`. So "the modeling passes *under a wrapper on a real agent run*" remains unproven; the resolver correctness under a wrapper is proven.
- ✅ **No AI slop.** The body is specific and the verification section states its own limits, which is what I asked for last time. The `description:` addition runs to four lines against `generate-task.md`'s "one or two sentences" — not worth a change given the `lint-task.md:159` carve-out.

### Recommendation

**Approve.** ✅

Everything I flagged is fixed, and the fixes hold up against the trees I built to break them. The two residuals are follow-up-shaped and already carried by #3219. This strictly beats the status quo, which fails 100% of wrapper runs.

## tl;dr

Fixed, and I checked it rather than trusting the replies: the resolver picks the right project on wrapper, bare, and draft-plus-wrapper trees, and fails loudly instead of grading the wrong file when two exist. Re-run on **codex** as asked — 5/5, score 1.000. Layout-ungraded is documented in `description:`, which is where this repo wants it. #3219 carries the 32 siblings and the unit test. Approving.

nikhil-maryala:
Re-verify green on **both harnesses**: claude (34517642443) + codex (34517645796) SUCCESS with the guarded resolver. Thanks!

PR #2940, opened 2026-08-31
test(maestro-bpmn): generous task/turn timeouts for 15 slow BPMN tasks

nikhil-maryala:
## Why

Today's nightly (`2026-08-31_04-15-47`, claude-code / sonnet-5) failed **15 `uipath-maestro-bpmn` tasks on `agent_timeout`** while the Codex daily passed all of them (Claude 62/79 vs Codex 78/79 on the tag).

Investigation ([Slack thread](https://uipath-product.slack.com/archives/C0A2T23NJ59/p1788170582569059)) showed it is **model throughput, not correctness**: ~96% of the 900s/1200s budget is per-turn Bedrock latency across 52–193 assistant turns; Codex finishes the identical tasks in **one turn / ~90–180s**. Several tasks had a valid BPMN already authored when the clock killed them.

**Bai Li's call in the thread:** *"let's increase the timeouts of these tasks to something generous (so all harnesses have enough time even if codex doesn't need as much)."*

## Change

Adds a generous ceiling to all 15 tasks' `run_limits`:
- `task_timeout: 2400` (was inherited 1200)
- `turn_timeout: 1800` (was inherited 900)

**Key fix:** three tasks (`composition/nested_review`, `queue_distribution`, `smart_triage`) already set `turn_timeout: 1800` but **still hit TIMEOUT at 1200s** — because `task_timeout` was never overridden and stayed at the inherited 1200, capping the whole task before the raised turn budget could apply (that's why they showed `TIMEOUT`, not `ERROR`). This adds the missing `task_timeout` so the turn budget actually takes effect. `task_timeout` (2400) is kept above `turn_timeout` (1800) so the task wrapper never pre-empts the turn.

Fast harnesses (Codex, etc.) are unaffected — they finish in ~90–180s and never approach these caps.

## Scope
Config-only change to 15 task YAMLs (+30/−5). No skill content, no experiment defaults touched — the generous caps are scoped to exactly the tasks that need them.

## Validation
coder-eval dispatched on the 15 tasks from this branch (link in comments). Expectation: the `agent_timeout`/`TIMEOUT` failures convert to completed runs graded on merit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3217, opened 2026-09-10
fix(uipath-maestro-bpmn): reduce high_volume_batch aggregate over-authoring (partial)

nikhil-maryala:
## Summary
Reduces wasted turns in `high_volume_batch` authoring. **Partial improvement, not a full fix** (see caveat).

## Context
`high_volume_batch` regressed in the eu nightly `2026-09-10` — timed out at `turn_timeout` 2700s. It's latency-bound, and the eu region is slower per turn than the us GH runner.

## Root cause (transcript, blob `2026-09-10_04-18-49`)
~15-20 of its turns went to hunting for multi-instance output-collection wiring (`loopDataOutputRef`, `outputCollection`, `dataOutputAssociation`, `uipath.aggregate`) that `check_greenfield.py` does not require — the `aggregate` node just needs to be a serviceTask on the path after the block.

## Fix
`high-volume-batch-guide.md` marks `aggregate` as a placeholder `bpmn:serviceTask` with no output-collection to author — the per-item reading/summing happens at runtime.

## ⚠️ This is partial
Verification (us region, run 34504377464): hunting dropped ~15-20 → 4 commands, and it passed once at 1829s — **but timed out at 2703s in another us run** (34503621234). The remaining **~78 commands / 148 turns are genuine authoring** of a heavy multi-instance + aggregate + escalation shape. It sits at the 2700s edge and flakes over it, and the eu nightly is slower still.

**A full fix needs decomposition or a fundamentally tighter authoring recipe** (closer to the `composition` case than the clean `queue_distribution` win). This PR is a strict, harmless improvement that cuts the wasted turns; the deeper work is tracked for follow-up. Merge on its own merits, or hold pending the decomposition decision.

### DIFF
--- a/skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
+++ b/skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
@@ -79,6 +79,12 @@ system or items must be ordered.
   `bpmn:completionCondition` on the multi-instance marker. A completion
   condition stops the block early, which is a different thing from judging the
   run afterwards, and it cannot see the per-item results the policy needs.
+  Author it as a **placeholder `bpmn:serviceTask` on the path after the block** —
+  the reading and summing of per-item results happen at runtime, not while
+  authoring. Do NOT wire multi-instance output collection (`loopDataOutputRef`,
+  `outputCollection`, `dataOutputAssociation`) or hunt the registry for a
+  `uipath.aggregate` type: none is required and none exists. Placing the node is
+  the whole authoring job.
 - **`send_report`** — email, chat, dashboard, audit store.
 
 Fetch payloads through [registry-workflow.md](../registry-workflow.md).
--- a/skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
+++ b/skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
@@ -79,12 +79,12 @@ system or items must be ordered.
   `bpmn:completionCondition` on the multi-instance marker. A completion
   condition stops the block early, which is a different thing from judging the
   run afterwards, and it cannot see the per-item results the policy needs.
-  Author it as a **placeholder `bpmn:serviceTask` on the path after the block** —
-  the reading and summing of per-item results happen at runtime, not while
-  authoring. Do NOT wire multi-instance output collection (`loopDataOutputRef`,
-  `outputCollection`, `dataOutputAssociation`) or hunt the registry for a
-  `uipath.aggregate` type: none is required and none exists. Placing the node is
-  the whole authoring job.
+  Author it as a `bpmn:serviceTask` on the path after the block — the per-item
+  reading and summing happen at runtime, so you do not need to wire
+  multi-instance output-collection plumbing (`loopDataOutputRef`,
+  `outputCollection`, `dataOutputAssociation`) or fetch a `uipath.aggregate`
+  registry template (there is none) to author it. Placing the node on the path
+  is the authoring job.
 - **`send_report`** — email, chat, dashboard, audit store.
 
 Fetch payloads through [registry-workflow.md](../registry-workflow.md).


nikhil-maryala:
Proactively applied @rockymadden's #3210 review lessons here so the same feedback isn't repeated (commit just pushed):
- Dropped **"placeholder"** — the table marks `aggregate` as **Mechanism**, so the word collided with the SKILL.md legend (the exact issue flagged on #3210). Now just `bpmn:serviceTask`.
- Scoped the output-collection line to authoring ("you do not need to wire … to author it") instead of a blanket ban on `loopDataOutputRef`/`outputCollection`, which are legitimate constructs a real runtime may use.
- Removed the "none is required" eval-grader implication; kept only the factual "there is no `uipath.aggregate` registry template."

(`uipath.aggregate` genuinely doesn't exist — confirmed against `validator/bpmn-spec.json`.) Still framed as a **partial** fix; the deeper decomposition remains the real solution.

skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
```diff
@@ -79,6 +79,12 @@ system or items must be ordered.
   `bpmn:completionCondition` on the multi-instance marker. A completion
   condition stops the block early, which is a different thing from judging the
   run afterwards, and it cannot see the per-item results the policy needs.
+  Author it as a `bpmn:serviceTask` on the path after the block — the per-item
```
rockymadden, line 82: 🔶 **major** — a placement-only `bpmn:serviceTask` can't do what this bullet requires.

Line 76 of this same bullet: `aggregate` must "apply the policy and set `policySatisfied`". `policy_gate` reads `=vars.policySatisfied == false` (line 48) against a `true` default (line 54). A bare serviceTask carries no `uipath:activity` / `uipath:mapping`, so it has no `uipath:output` — `policySatisfied` never changes and `end_failed` is unreachable at runtime.

`BPMN.ScriptTask` is the one bundled type that computes and writes a variable: `bpmn:scriptTask`, `uipath:mapping`, `requiresDiscovery: false`, no context fields, `<uipath:output ... source="=result.response">` (`structural-bpmn.md:158-217`). `check_greenfield.py:57` already counts `scriptTask` as work, so this costs **one** deterministic `registry get BPMN.ScriptTask` — cheaper than the search this PR removes — and yields a runnable node.

**Fix:** say `bpmn:scriptTask` (`BPMN.ScriptTask`) here and in the Shape table (line 34), linking `structural-bpmn.md:158`. If a bare serviceTask really is intended, write "with no extension payload" the way merged #3210 does, and drop "set `policySatisfied`" from line 76 — as written the guide asks for two incompatible things.

Also: "Placing the node on the path is the authoring job." restates the clause that opens the same sentence. Drop it — 12 lines for one node buries the operative instruction.

skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
```diff
@@ -79,6 +79,12 @@ system or items must be ordered.
...
+  reading and summing happen at runtime, so you do not need to wire
+  multi-instance output-collection plumbing (`loopDataOutputRef`,
+  `outputCollection`, `dataOutputAssociation`) or fetch a `uipath.aggregate`
+  registry template (there is none) to author it. Placing the node on the path
```
rockymadden, line 86: 🔶 **major** — `uipath.aggregate` *does* exist in this skill.

`structural-bpmn.md:164`: "Only these helpers exist: `uipath.aggregate`, `uipath._aggregate`, `uipath._pipe`, and a no-op `console`." It's the Jint script-runtime helper — and it's precisely what sums per-item results inside a `BPMN.ScriptTask`.

"(there is none)" is true only of a *registry template*, and reads as "the identifier doesn't exist". The PR body and the Slack post both shortened it to "which doesn't exist". So the skill now says opposite things about one identifier in two files, and this line steers the agent off the correct helper for this exact job.

**Fix:** drop the identifier — "...or fetch a registry template for it: no extension type covers aggregation." Behavior change preserved, collision gone. If intercepting the specific guess matters, disambiguate instead: "`uipath.aggregate` is a Jint script helper (`structural-bpmn.md:164`), not a registry type."

skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
```diff
@@ -79,6 +79,12 @@ system or items must be ordered.
...
+  is the authoring job.
 - **`send_report`** — email, chat, dashboard, audit store.
```
rockymadden, line 88: 🔹 **minor** — the residual hunt is probably here, untouched.

`send_report` is a `bpmn:sendTask` (line 35). Every bundled sendTask type needs discovery or a live connection: `Intsvc.ActivityExecution` (`requiresDiscovery: true`), `Orchestrator.CreateQueueItem` (`true`), `Intsvc.HttpExecution`, `Intsvc.UnifiedHttpRequest`. "email, chat" plus the trailer below sends the agent to `uip is connections list --all-folders`, which `registry-workflow.md:20-26` makes mandatory and whose empty result it labels a **false negative** — for a connection no greenfield sandbox has.

#3225 writes exactly this carve-out for its guide ("a greenfield authoring pass has none"). Add the same sentence here and the residual 4 commands likely go with it.

skills/uipath-maestro-bpmn/references/patterns/high-volume-batch-guide.md
```diff
@@ -79,6 +79,12 @@ system or items must be ordered.
...
 
 Fetch payloads through [registry-workflow.md](../registry-workflow.md).
```
rockymadden, line 90: 🔶 **major** — this trailer undoes the paragraph above it.

Unqualified, it routes the agent into `registry-workflow.md` §1, which mandates `registry pull` then `list` / `search`, plus `uip is connections list --all-folders`. That is the hunting loop.

Merged #3210 narrowed this exact trailer in the sibling guide: `queue-distribution-guide.md:117-120` now names which node needs a lookup and states "The performer's nodes need no such lookup." Do the same here: `fetch_batch` and the `process_each` action node take a template; `aggregate` does not.

rockymadden (changes requested):
## What

+6 lines in `high-volume-batch-guide.md` "What to bind": author `aggregate` as a `bpmn:serviceTask` on the path after the multi-instance block; no output-collection wiring, no `uipath.aggregate` registry template. Cuts the ~15-20 turns the eu nightly burned hunting that plumbing.

## Confirmed from the repo

- `check_greenfield.py:55-69` needs >=2 work activities (serviceTask / scriptTask / sendTask / message throw) between the block and the verdict gateway, and asserts **no** `uipath:*` payload. The "just be a serviceTask on the path" claim holds.
- No `uipath.aggregate` in `bpmn-spec.json` `extensionTypes` (29) or `bpmnElements.uipathShortcuts` (11).
- Runs `34504377464` (success) and `34503621234` (failure) exist and match the flake account. Reporting your own failing run is the right call.
- Single copy of the guide; nothing in `preview/` or `skill-flavors/` to keep in sync.
- #3150 and #2940 already raised this task's budget twice. "A bigger cap won't fix it" is well supported.

## Overall findings

- 🔶 **major — verification doesn't cover the failure being fixed.** The regression is the **eu** nightly; both verify runs are **us** `claude` (one pass, one timeout). Net evidence is "hunt commands fell in one us run". The eu-is-slower-per-turn claim isn't verifiable from the repo. Run it once on eu, or drop eu from the claim.

- 🔹 **minor — "tracked for follow-up" has no tracker.** #3170 covers the four tasks #3150 *reverted*; `high_volume_batch` was one of the four it **kept** as verified-green, so it's out of scope there. File the decomposition issue or link it — right now the deeper work has no home.

- 🔹 **minor — the turn numbers don't reconcile with the task's own caps.** Body says "~78 commands / 148 turns are genuine authoring". `greenfield.yaml:16` sets `max_turns: 70`, `expected_turns: 26` — 148 harness turns would report `MAX_TURNS_EXHAUSTED`, not `turn_timeout`. #3150's message says these tasks die as one streamed turn, so "148" is presumably messages, not turns. **I could not reconcile it from the repo.** Label the unit; the decomposition conclusion rests on it.

- 🔹 **minor — missing guard.** `tests/tasks/uipath-maestro-bpmn/_shared/test_bpmn_check.py:60-99` already asserts doc claims against `bpmn-spec.json`. This is the second PR running to ship a named absolute about extension types. A guard that extracts `Namespace.Type` / `uipath.*` tokens from `references/**.md` and checks each against the spec would catch this class at CI rather than at review.

Intent is right and the hunting cut is real. Two of the six added lines are wrong, not just incomplete — both one-line fixes.

## tl;dr

Right problem, verified against the checker. But the added text negates a `uipath.aggregate` that exists elsewhere in the skill, and prescribes a node that can't set the variable the pattern gates on. Fix those two lines and it's a clean merge.

nikhil-maryala:
All confirmed — thank you, these were real bugs, not nitpicks:
- `uipath.aggregate` **does** exist (`structural-bpmn.md:164`, the Jint helper that sums per-item results); my "there is none" negated it and contradicted the skill. Confirmed.
- A bare `serviceTask` can't set `policySatisfied` — `aggregate` must be `BPMN.ScriptTask`. Confirmed. Filed **#3229** for that guide-correctness fix (distinct from the hunting issue).
- The `send_report`/trailer residual hunt and the `registry-workflow.md` trailer routing back into the loop — that's the shared root cause.

Acting on the systemic feedback: the hunting fix moved up a level into `registry-workflow.md` (greenfield/no-tenant carve-out) as **#3228**, which supersedes this per-guide patch. **Closing this** in favor of #3228 (hunting) + #3229 (aggregate = ScriptTask) + #3230 (CI guard).

PR #3220, opened 2026-09-10
fix(uipath-automationhub): install the ah tool when the preflight finds it missing

parth-patil-uipath:
## Stack
Part **1 of 2** — review/merge bottom-up.

- [ ] 1. `rpanav-19070/ah-tool-install` — install the `ah` tool at preflight *(base: `main`)* ← **this PR**
- [ ] 2. `rpanav-19070/ah-send-uipath-request-v2` — raw-API fallback over `SendUiPathRequest` *(base: #1)*

## What this PR does

When Step 0's `uip ah --help` answers `unknown command 'ah'`, the skill now runs `uip tools install ah` once and re-checks before giving up on the CLI transport.

## Why

`uip` is a **dispatcher**. Every verb ships as its own npm package, and `@uipath/cli` declares **no** `*-tool` packages in any dependency block — so a host can bundle a perfectly current `uip` with no `ah`, and **the CLI version tells you nothing about whether it is there**. Only `uip ah --help` does.

A shipped UiPath Delegate did exactly that. Its `bin/node_modules/@uipath/` contains only `orchestrator-tool`, so `uip ah` was unavailable and Automation Hub publishing fell through to the raw-API path — where the sandbox blocks curl, and the user saw a network-policy error instead of a missing tool. Reported as a P0 in the 9 Sep Cartographer dogfooding session (RPANAV-19070).

The old reference note framed a missing `ah` as an out-of-date CLI. That is the misconception that made this hard to diagnose, so it is replaced with the packaging fact.

## Why it stands alone

Pure CLI-path recovery: no transport change, no new tool, no behaviour change on a host that already has `ah` (the preflight succeeds and the install never runs). Merging it alone strictly increases the number of runs that reach the CLI flow.

## Testing

New `tests/tasks/uipath-automationhub/smoke_missing_ah_tool.yaml` (4 criteria) grades the recovery decision: skill activates · names `uip tools install ah` **and** the re-check · explains the dispatcher/packaging reason rather than a version reason · falls through to the raw-API flow without looping on the install.

Graded as a decision test because the eval image ships an `ah`-capable `uip` — a test that tried to *execute* this path would take the CLI branch and assert nothing.

`npm run skills:validate` → `default: 27 skills, 1768 files; studioweb: 27 skills, 1768 files, 264 replacements`.

## Reviewer notes
- `uip tools install ah` is capped at **one** attempt: inside the Delegate sandbox it needs `npm.pkg.github.com`, which is denied, so a retry loop would just burn turns before the fallback.
- The Delegate-side fix that makes this moot on shipped builds is Autopilot#5898 (bundle the tool). This PR is what makes the skill recover on any host that lacks it, including future ones.

### DIFF
--- a/skills/uipath-automationhub/SKILL.md
+++ b/skills/uipath-automationhub/SKILL.md
@@ -11,10 +11,11 @@ Work with business processes in UiPath Automation Hub (AH) through the AH Open A
 
 ## Step 0: Preflight — pick the transport once
 
-Run `uip ah --help` once per session:
+Run `uip ah --help` once per session.
 
 - **Succeeds** → use the **CLI flows**. Read [`references/cli-commands.md`](references/cli-commands.md) (command catalog + auth), then the matching `*-cli-guide.md` flow. Auth is handled by `uip` itself — never touch a token.
-- **Fails with `unknown command 'ah'`** (CLI predates the AH surface) → use the **raw Open API flows**. Read [`references/api-endpoints.md`](references/api-endpoints.md) (auth model, gateway URL, exact headers — and the header to never send), then the matching flow.
+- **Fails with `unknown command 'ah'`** → the `ah` tool isn't installed. `uip` is a dispatcher: each verb is its own package, and the AH package ships separately from the CLI, so a host can bundle `uip` without it. Run `uip tools install ah` **once**, then re-run `uip ah --help`. If it now succeeds, take the CLI flows.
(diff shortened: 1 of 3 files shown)


parth-patil-uipath:
Closing — team decision (10 Sep): the `ah` tool gets **bundled by the Delegate installer** (UiPath/Autopilot#5898), and the raw-API fallback moves into Autopilot's private `automation-hub-publishing` skill over `SendUiPathRequest`, stacked on #5898, so it isn't exposed in the public skills. Findings from this PR (dispatcher discovery, cwd scan, `tools install` writing where Delegate never looks) are recorded on RPANAV-19070.

PR #3221, opened 2026-09-10
feat(uipath-automationhub): run the raw-API fallback over SendUiPathRequest

parth-patil-uipath:
## Stack
Part **2 of 2** — review/merge bottom-up.

- [x] 1. #3220 — install the `ah` tool at preflight *(base: `main`)*
- [ ] 2. `rpanav-19070/ah-send-uipath-request-v2` — raw-API fallback over `SendUiPathRequest` *(base: #3220)* ← **this PR**

> Stacked PR — base is #3220, not `main`. The diff shows only this layer.

## What this PR does

Makes the raw Open API fallback usable inside a UiPath host by issuing requests through the **`SendUiPathRequest`** tool instead of `curl`.

## Why

The Delegate sandbox permits `uip` and `SendUiPathRequest` and denies raw shell HTTP. So when the CLI path was unavailable, every `curl` in the fallback died at DNS and the agent reported it as:

> Publishing is currently blocked by this Delegate session's network policy: DNS access to `cloud.uipath.com` is disabled…

The sandbox was doing its job; the fallback was simply unreachable there. #3220 stops us landing on the fallback for the common cause, but the fallback stays reachable whenever a tool is missing, renamed, or a future skill runs on a host without it — and today every one of those degrades into that same misleading message.

## Shape: one request, two renderings

Rather than add a third copy of the flows (there are already CLI + API twins), the flows now state each call **once** as `METHOD /endpoint` and pick a rendering from a single **Transport** section in `api-endpoints.md`. On `SendUiPathRequest` there is no token, base URL, org, tenant or `Content-Type` to resolve — the host injects all of it and strips `authorization` from caller headers — so the auth boilerplate that was repeated at all six request sites is gone. Net for the flow files: **−72 lines**.

The `x-ah-openapi-auth` / `x-ah-openapi-app-key` never-send rule is restated on the tool path deliberately: those headers are not in the tool's protected list and would pass straight through.

## Known limitations, documented in the skill

Two places this transport is weaker than curl. Both are written into the Transport section rather than left for an agent to discover:

1. **No HTTP status line.** The tool returns the response *body*; a non-2xx arrives as a tool error whose text carries the API's error body. The flows branch on 201/400/401/403/404/409/422, so they now read the status from **AH's own envelope** (`statusCode` / `status`). Consequence: a **redirect target** may be invisible, and a bare gateway 404/422 with no AH envelope has no code to match — so the "Automation Hub not available on this tenant" case is classified by signal rather than by number.
2. **Large responses are dumped to a file** past `maxInlineBytes` (default 2KB). `GET /idea-schema` always exceeds it — pass a larger `maxInlineBytes` or read the dumped file, never the truncated preview.

## Why it stands alone

The curl path is unchanged for plain shells, and the CLI path never touches any of this. On a host with `ah` present, nothing in this PR executes.

## Testing

New `tests/tasks/uipath-automationhub/smoke_no_cli_transport.yaml` (5 criteria): skill activates · **never** reaches for curl/wget against Automation Hub · never sends the admin headers · names `SendUiPathRequest` with a host-relative `automationhub_/api/v1/openapi…` path and no hand-built auth · reports a missing transport as the missing tool, not as the user's network policy.

Decision-graded for the same reason as #3220 — the eval image's `uip` has `ah`, so an execution test would take the CLI branch.

`npm run skills:validate` → `default: 27 skills, 1768 files; studioweb: 27 skills, 1768 files, 264 replacements`.

## Open questions for review

- **`autoExecute = false`** on the tool: every call needs user approval, so a publish is ~6 prompts. Acceptable, or should the host auto-approve this tool for skill-driven flows?
- **`availability: "autopilotAgent"`** — please confirm the agent that runs `uipath/skills` in Delegate actually exposes the tool; if it doesn't, this fallback is as unreachable as curl was.
- The tool targets the **session's** org/tenant. Publishing to a different tenant than the host session is not expressible on this transport — fine today, worth knowing.

cc the Delegate-side fix for the same P0: Autopilot#5898.

### DIFF
--- a/skills/uipath-automationhub/SKILL.md
+++ b/skills/uipath-automationhub/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: uipath-automationhub
-description: "Publish and read business processes in UiPath Automation Hub via the Open API, using the user's cloud login — no admin OpenAPI token needed. PUBLISH an approved process and its PDD/SDD documents (schema-driven payload, base64 file upload or link) to AH as the system of record — e.g. a process captured/approved by Process Scribe. GET a process back by id or search, list its attached documents, and DOWNLOAD their file bytes (e.g. for dedup / related-idea lookups or retrieving a published PDD). Authenticates with the user's cloud bearer token and NEVER sends the admin `x-ah-openapi-auth` header. CLI-first: when the installed `uip` has the `ah` commands, flows run through them (`references/*-cli-guide.md`); otherwise the raw Open API flows apply. Routes by intent to publish (create/upload) or get (read/fetch/download) references over the shared catalogs. Structured to extend to more Automation Hub operations."
-allowed-tools: Bash, Read, AskUserQuestion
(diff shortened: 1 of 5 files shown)


parth-patil-uipath:
Closing — team decision (10 Sep): the `ah` tool gets **bundled by the Delegate installer** (UiPath/Autopilot#5898), and the raw-API fallback moves into Autopilot's private `automation-hub-publishing` skill over `SendUiPathRequest`, stacked on #5898, so it isn't exposed in the public skills. Findings from this PR (dispatcher discovery, cwd scan, `tools install` writing where Delegate never looks) are recorded on RPANAV-19070.

PR #3222, opened 2026-09-10
fix(maestro-flow): surface the `flow debug` folder-flag rule where agents read

dmetzgar:
**Replaces this PR's previous contents entirely.** The seed/rename approach is
gone — @rockymadden's review was right that the root cause was asserted, not
shown, and checking it disproved my diagnosis. Details below.

## What actually happened

v1's three `flow debug` attempts all passed `--folder-path` at a shared folder:

| arm | folder flag | stage | http | ran the flow? |
|---|---|---|---|---|
| v1 ×3 | `--folder-path Shared/uipath-maestro-flow/BillingDispute` | `prepare-custom-debug` | 500 | **no** |
| v2 ×4 | none | — | — | yes, 3 of 4 |

Same tenant, same solution name, same four minutes. The arm with the flag failed
every time; the arm without it ran the flow. That is the documented cause, from
this repo's own `references/diagnose/troubleshooting-guide.md`:

> `Stage: prepare-custom-debug` with `HttpStatus: 500`, and no `Data.incidents`
> → Debug was pointed at a shared folder with `--folder-path` or `--folder-key`.

Not a cross-arm solution-name collision, which is what this PR previously tried
to fix.

## Why the agent did it

The rule already exists, at `operate/run.md:31`. The agent never read it — it
opened 14 skill docs and **every one was under `author/` or `shared/`**. It went
from `flow debug --help` straight to the flag, and the help's own example for
`--folder-path` is `'Shared/Sub'` — the same shape as the folder the task names
for its resources. `"Auto-detected if none are provided"` appeared only on
`--folder-id`.

SKILL.md (103 lines, read in full) points at `operate/` three times but carried
no inline folder warning, even though it already inlines the background/timeout
debug gotcha and links `operate/run.md#debug` for it.

## Change

- `skills/uipath-maestro-flow/SKILL.md` — hoist the one-line rule beside the
  existing debug rules, and say that a folder the task names holds the flow's
  *resources* (which arrive via `uip solution resources refresh`), not a debug
  target. That conflation is what made the flag look correct.
- `preview/.../CLI-LOOP.md` — said only "omit to auto-detect", which is why v2
  happened to avoid this. Made explicit.

CLI-side counterparts — the help example and the "often transient, retry" advice
that turned one mistake into three — are in **UiPath/cli#4164**.

## Tests

```
tests/tasks/uipath-maestro-flow/_shared   997 passed
scripts/check-task-driver.py (full corpus)  exit 0
tests/ (excl. tasks)                      252 passed, 1 skipped
```

`tests/scripts/test_stage_shared.py::test_stage_shared_anchors_to_checkout_and_stages_nested_checkers`
fails identically on clean `origin/main` — pre-existing.

## On the collision hypothesis

Not disproved, just unevidenced: replicates, overlapping runs and retry
leftovers can still collide, which is @bai-uipath's argument on
UiPath/coder_eval_uipath#125 and stands on its own. But it should be argued and
scheduled on its own merits, not on this incident. Worth noting the corpus
constraint found on the way: `test_same_ground_corpus.py:184`
(`DEBUG_SOLUTION_ALLOWLIST` empty) requires the literal phrase *"in a solution of
the same name"* on every debug-graded task, so any future rename has to widen
that test first — which is exactly the contradiction @rockymadden flagged inline.


tests/tasks/uipath-maestro-flow/multi_node/billing_dispute_resolution/billing_dispute_resolution.yaml
```diff
@@ -22,6 +22,11 @@ post_run:
   - command: python3 $SKILLS_REPO_PATH/tests/tasks/uipath-maestro-flow/_shared/cleanup_solutions.py
     timeout: 90
 initial_prompt: |
+  Parallel runs share one tenant, so the solution name must be unique per run.
+  A unique suffix (`uuid8`) is in `seed.json` — append it to the SOLUTION
+  name only (`Foo` becomes `Foo-<uuid8>`). Leave the flow, project and tenant
+  resource names exactly as written below.
+
   Build a Maestro Flow "BillingDisputeResolution" (in a solution of the same
```
rockymadden, line 25: 🔴 **major** — The new rule and this sentence contradict each other:

```
append it to the SOLUTION name only (`Foo` becomes `Foo-<uuid8>`)
...
Build a Maestro Flow "BillingDisputeResolution" (in a solution of the same name)
```

"Same name" now names something the rule forbids. An agent resolving the conflict the other way — keeping them identical by suffixing *both* — produces exactly the failure your Risk section describes: the flow file moves and the advisory dies with `found 0`.

This isn't a wording nit you can fix locally: `_shared/test_same_ground_corpus.py:184` (`test_debug_graded_prompts_name_a_same_name_solution`, `DEBUG_SOLUTION_ALLOWLIST` empty) *requires* that literal phrase on every debug-graded task, so the contract itself will force the contradiction onto all 107.

Fix: reword to name the composed solution (`in a solution named "BillingDisputeResolution-<uuid8>"`) and widen that test's regex to accept the suffixed form.

Applies to all five tasks.

tests/tasks/uipath-maestro-flow/multi_node/billing_dispute_resolution/billing_dispute_resolution.yaml
```diff
@@ -116,6 +121,8 @@ initial_prompt: |
   decision, assumption, and blocked step in your final response. Instructions in
   the task take precedence over this paragraph.
 pre_run:
+  - command: E2E_PROCESS_KEY= python3 "$SKILLS_REPO_PATH/tests/tasks/uipath-platform/seed.py"
```
rockymadden: 🟠 **major** — `E2E_PROCESS_KEY=` skips less than the description says. `seed.py:51` calls `ensure_parent_folder(PARENT_FOLDER_PATH)` unconditionally, *before* the key check — 1–3 `uip or folders get/create` calls against Orchestrator, with its own `sys.exit(1)` at `seed.py:47`. `PreRunCommand.fail_on_error` defaults to `True` (`scripts/check-task-driver.py:31`), so a slow or unauthorized tenant ERRORs the task before the agent starts.

Two consequences these five tasks did not have before: a hard Orchestrator dependency at `pre_run`, and a side effect creating `Shared/uipath-platform-e2e` on the tenant — a folder none of them use.

These tasks need 8 hex characters. Either add a `--uuid-only` flag to `seed.py` that returns before `ensure_parent_folder`, or inline it:

```yaml
- command: python3 -c "import json,uuid; json.dump({'uuid8': uuid.uuid4().hex[:8]}, open('seed.json','w'))"
```

Bonus: drops the misleading `seed.py: E2E_PROCESS_KEY is not set — … every e2e task that needs them will fail` warning these runs now print on stderr every time.

Applies to all five tasks.

rockymadden (approved):
## What

Adds a `pre_run` seed step and a 4-line prompt rule to five `uipath-maestro-flow/multi_node/billing_*` tasks so each run builds a uniquely-named **solution** (`Foo-<uuid8>`), leaving flow/project/tenant names alone. Claimed as a zero-grader-change fix for A/B arms colliding on one tenant.

**Confirmed by reading the code:**

- ✅ Grader discovery is name-independent — `_find_project` selects on `ProjectType:"Flow"` + node count, not directory name (`flow_check.py:1966`).
- ✅ `load_flow` rglobs a filename (`advisory_flow_utils.py:69`); a renamed parent solution dir doesn't move it.
- ✅ `cleanup_solutions.py` deletes by `SolutionId` from every `.uipx` under cwd, never by name.
- ✅ `pre_run` sits outside the `task_timeout` watchdog (`test_criterion_budgets.py:546`), so +60s costs no budget.
- ✅ All five are debug-graded (each `check_*.py` calls `run_debug`), so all five really do provision a solution.

## 🔍 Overall findings

**🔴 major — the root cause is asserted, not demonstrated.**
This repo's own docs attribute the exact observed signature to a different cause: `skills/uipath-maestro-flow/references/diagnose/troubleshooting-guide.md:71` says `Stage: prepare-custom-debug` + `HttpStatus: 500` + no `Data.incidents` means *debug was pointed at a shared folder with `--folder-path`/`--folder-key`*, and `operate/run.md:31` says debug provisions into the personal workspace. The grader's `run_debug` passes neither flag (`flow_check.py:368-376`) — so those three failures were the **agent's** own debug attempts, which is precisely where the flag misuse would occur. Separately, v2's `HTTP 403` at `poll-instance-status` is a different stage and a different class of code; folding both into one name-collision story is a stretch.

Fix: rule the documented cause out first (grep the failing runs' transcripts for `--folder-path`/`--folder-key`), and say so in the description. Landing an unproven fix here is cheap; replicating it to the 107 tasks named in Scope is not.

**🟠 major — no wording pin on new boilerplate headed for 107 tasks.**
#3088 (`1ec2ba297`) established the pattern: shared prompt text is duplicated verbatim *and* pinned by `test_headless_preamble.py::test_the_wording_is_identical_everywhere`, because "8 variants is what made the old line unmaintainable." This PR adds a second block of shared prompt text with no equivalent test.

Fix: add `CANONICAL_SEED_SUFFIX_RULE` + an identical-wording test in `_shared/`, while there are five copies and not 112.

**🟡 minor — nothing observes whether the agent obeyed the rule.**
If it ignores the suffix, the collision returns and you get the same opaque `HTTP 500` — indistinguishable from a genuine tenant fault, which is the ambiguity that cost the 2026-09-09 triage. Add a report-only criterion (`weight: 1.0`, `pass_threshold: 0.0`, the pattern already beside `advisory_billing_dispute_resolution.py`) asserting the emitted `.uipx` filename carries `seed.json`'s `uuid8`. It converts a silent regression into a signal, and it's the measurement you want before scaling to 107 tasks.

**🟡 minor — description contains one false claim.**
"`E2E_PROCESS_KEY=` … drops seed.py into the branch that writes `uuid8` … without its `sys.exit(1)` path." `seed.py` has two `sys.exit(1)` paths and the blanked key only skips one. Detail inline.

**🟡 minor — branch is still `fix/advisory-billing-dispute-flow-glob`**, from the approach this PR replaced. Cosmetic, but the squash message is what future readers `git log` for.

**What I reasoned about vs. confirmed:**

- ❓ Could not verify the "107 / 31 / 1" corpus counts — not re-derived.
- ❓ Could not verify the "sync gate" the task header cites for cross-arm parity; nothing in this repo implements it.
- ❓ Did not run the tests (`997 passed`, `check-task-driver.py` exit 0, `test_headless_preamble.py`). Reported as claimed. The one suite worth re-running is `test_same_ground_corpus.py` after the inline fix below.
- ✅ No AI slop in the description. It is specific, sourced, and states its own risk.

## ✅ Recommendation

**Approve with comments.** The mechanics hold and the blast radius is five test YAMLs. Please fix the prompt contradiction (inline) before the follow-on sweep — the corpus test currently guarantees it lands on all 107.

---

**tl;dr** — Mechanics check out: discovery, `load_flow`, and cleanup are all name-independent, so the zero-grader-change claim holds. Three things to fix. The prompt now says both "suffix the solution" and "solution of the same name", and a corpus test pins the second phrase — that's the exact agent-confusion failure your Risk section predicts. `E2E_PROCESS_KEY=` doesn't skip what the description says it skips; `seed.py` still hits Orchestrator and can hard-fail `pre_run`. And the root cause is asserted, not shown — this repo's own troubleshooting guide blames that identical 500 signature on `--folder-path`, not a name clash. Approving; fix the contradiction and pin the wording before this becomes 107 copies.

dmetzgar:
Cancelling since the logic has changed a bit in the past few days.

Issue #3219, opened 2026-09-10
BPMN eval: modeling-graded tasks false-fail on the default init Solution-wrapper layout (32 tasks, follow-up to #3216)


nikhil-maryala:
## Context
Follow-up from #3216 (thanks @rockymadden). `uip maestro bpmn init` outside a solution auto-scaffolds `<Name>Solution/<Name>/` by default (`references/shared/project-layout.md:16-33`). A BPMN task that grades **modeling** but asserts an exact top-level `.bpmn` path then false-fails whenever the agent uses that default scaffold — which is exactly how `skill-bpmn-e2e-customer-escalation` failed in the `2026-09-10` nightly.

#3216 made `customer_escalation` location-robust (locate the project by the `<Name>.bpmn` + `project.uiproj` pair, guard against multiple copies). But ~32 sibling BPMN tasks still assert a directory-prefixed exact `.bpmn` path and carry the same latent false-negative:
- `patterns/*/greenfield.yaml`, `multi_node/calculator`, `single_node/agent_job`, `nodes/queue_create_and_wait`, … (Rocky counted 32 exact-prefixed + 24 bare-root).

## Work
1. Extract the resolver from `customer_escalation`'s check into `_shared/bpmn_check.py` (e.g. `resolve_project(bpmn_name)` → the dir of the unique `<bpmn_name>` with `project.uiproj` beside it), plus a shared criteria helper (`_shared/require_beside_bpmn.py`) so the YAML one-liners and the check use ONE selection rule.
2. Adopt it across the modeling-graded BPMN tasks (start with `wiki_pageviews`, then the 32 exact-prefixed ones). Skip tasks where layout is intentionally part of the grade (e.g. solution-structure tasks) — those should stay exact.
3. Add a `tmp_path` unit test in `_shared/test_bpmn_check.py` asserting wrapper / bare / two-copy trees resolve correctly (turns "verified locally" into CI-enforced).

## Acceptance
Modeling-graded BPMN tasks grade the process wherever the default init scaffold places it; a `_shared` resolver + test prevent regressions; layout-sensitive tasks are explicitly exempt.

PR #3225, opened 2026-09-10
fix(uipath-maestro-bpmn): cut smart-triage uip-or-processes hunt (improvement, still flaky)

nikhil-maryala:
## Summary
Reduces the turn-waste that pushed `smart-triage` to MAX_TURNS in the `2026-09-10` nightly (138 turns, **never authored the `.bpmn`**). Same class as the merged #3210 — the agent wastes its turn budget hunting for bindings the checks don't require — but the specifics differ, so the fix is scoped accordingly.

## Root cause (transcript, blob `2026-09-10_04-18-49`)
~16 turns went to paging **`uip or processes list`** for a real deployed "TriageAgent" process, plus `registry search agent/classif/hitl/...` for classifier and handler nodes that are plain `bpmn:serviceTask`s. `check_greenfield.py` grades two gateways, a `userTask` fallback, per-category ends, and a diagram — no `uipath:*` payload assertion.

## Fix
`smart-triage-guide.md` "What to bind" now leads with: do not run `uip or processes` / `uip or` / `uip login` for a live agent or process to bind — a greenfield authoring pass has none, and rule 2 forbids inventing a key. `classify`/`extract`/`handle_a..c` are authored as `bpmn:serviceTask`s (runtime binding is enrichment); only `manual_triage` (`Actions.HITL`) and `apply_rules` (`Orchestrator.BusinessRules`) have a registry template to `registry get`.

**Names the real types rather than claiming anything nonexistent, targets the `uip or processes` waste specifically, and avoids the "placeholder"/eval-grader/false-absolute issues from the #3210 review** (written after that review).

## Verification (`claude`/`claude-sonnet-5`) — improvement, mostly-passing (not a guaranteed pass)
The turn-reduction works: registry/`uip or` hunt commands dropped from **16 → ~1-8**, and the agent now authors the file (before, it burned all its turns hunting `uip or processes` and never wrote anything).

But `smart-triage` is a **borderline-heavy** task and flakes over the 1800s turn budget:

| Run | Result | Duration |
|---|---|---|
| 34524230825 | SUCCESS | 1539s |
| 34528341432 | **ERROR (timeout)** | 1803s (never authored in time) |
| 34532015874 | SUCCESS | 1384s |
| 34532022083 | SUCCESS | 1142s |

**3/4 pass**, usually well under budget, with one flaky tail at 1803s — and the eu nightly is slower, so it will flake more often there. This is a strong improvement (never-passing → mostly-passing), but the residual flakiness is the task's intrinsic weight, shared with `high_volume_batch` (#3217). A guaranteed pass needs decomposition, tracked for follow-up.

Note: `smart-triage` also carries the exact-path check that #3219 tracks (layout robustness across ~32 tasks); the agent authored top-level in these runs so it passed, but that follow-up still applies.

### DIFF
--- a/skills/uipath-maestro-bpmn/references/patterns/smart-triage-guide.md
+++ b/skills/uipath-maestro-bpmn/references/patterns/smart-triage-guide.md
@@ -89,6 +89,15 @@ reaches manual triage.
 
 ## What to bind
 
+Bind at authoring time only what the CLI can supply offline. **Do not run `uip or
+processes`, `uip or`, or `uip login` to find a live deployed agent or process to
+bind into `classify` or a handler** — a greenfield authoring pass has none, and
+rule 2 forbids inventing a key. Author `classify`, `extract`, and `handle_a..c`
+as `bpmn:serviceTask`s; their runtime work (agent classification, extraction,
+dispatch) is bound during enrichment, not now. Only `manual_triage`
+(`Actions.HITL`) and `apply_rules` (`Orchestrator.BusinessRules`) have a registry
+template to `registry get` — do not `registry search` for the plain serviceTasks.
+
 - **`classify`** — the node that classifies at runtime, typically a UiPath agent
   job. Place and bind it; never decide the categories or write classification
   logic while authoring.
--- a/skills/uipath-maestro-bpmn/references/patterns/smart-triage-guide.md
+++ b/skills/uipath-maestro-bpmn/references/patterns/smart-triage-guide.md
@@ -89,14 +89,14 @@ reaches manual triage.
 
 ## What to bind
 
-Bind at authoring time only what the CLI can supply offline. **Do not run `uip or
-processes`, `uip or`, or `uip login` to find a live deployed agent or process to
-bind into `classify` or a handler** — a greenfield authoring pass has none, and
-rule 2 forbids inventing a key. Author `classify`, `extract`, and `handle_a..c`
-as `bpmn:serviceTask`s; their runtime work (agent classification, extraction,
-dispatch) is bound during enrichment, not now. Only `manual_triage`
-(`Actions.HITL`) and `apply_rules` (`Orchestrator.BusinessRules`) have a registry
-template to `registry get` — do not `registry search` for the plain serviceTasks.
+Get each node's registry **type** with `registry get <type>` — `classify` an
+agent-job type (e.g. `Orchestrator.StartAgentJob`), `manual_triage`
+`Actions.HITL`, `apply_rules` `Orchestrator.BusinessRules`. **Do not run `uip or
+processes`, `uip or`, or `uip login` to find a *live deployed* process or agent
+instance** — a greenfield authoring pass has none, rule 2 forbids inventing a
+key, and binding a runtime instance is enrichment, not authoring. `extract` and
+`handle_a..c` are placeholder DU/dispatch serviceTasks; do not `registry search`
+speculatively to name them.
 
 - **`classify`** — the node that classifies at runtime, typically a UiPath agent
   job. Place and bind it; never decide the categories or write classification


nikhil-maryala:
Applied the #3210 review lessons here too (commit a78039e8d). Caught a false absolute in my own first pass: I'd written that only `manual_triage`/`apply_rules` have a registry template — but `classify`'s agent-job type **is** bundled (`Orchestrator.StartAgentJob`/`StartJob`/`RunJob`/`StartAgenticProcess` all exist in `validator/bpmn-spec.json`). Reworded to name the real types via `registry get` and target the actual waste (paging `uip or processes` for a *live deployed instance* + speculative `registry search`), with no "placeholder"-on-a-Mechanism-node collision. Re-verifying the reworded guide (34528341432); the behavioral win — hunt 16→1, authored + passed — held on the prior run.

skills/uipath-maestro-bpmn/references/patterns/smart-triage-guide.md
```diff
@@ -89,6 +89,15 @@ reaches manual triage.
 
 ## What to bind
 
+Get each node's registry **type** with `registry get <type>` — `classify` an
```
rockymadden, line 92: 🔶 **major** — all three types here are `requiresDiscovery: true`, and the next sentence forbids the only documented way to satisfy that. No third option is given.

- `Orchestrator.StartAgentJob` — `bindingPattern: process`, context `name` / `folderPath` = `=bindings.{nameBindingId}` / `{folderPathBindingId}`
- `Orchestrator.BusinessRules` — `bindingPattern: businessRule`, `releaseKey` from `bindingInfo.propertyAttribute: Key`
- `Actions.HITL` — `appId`, `appVersion`, `actions`, `key`, `taskTitle`

`registry-workflow.md:96-97`: "Resolve the real key with `registry search` / discovered `Processes` / `Queues`; never guess a GUID." Rule 2 forbids inventing one. So the agent gets fetch-the-template, don't-discover, don't-invent — and nothing else. That gap is what burns turns.

**Fix:** name the third option in one clause. Paste the `xmlTemplate`, leave the binding placeholders and the `<uipath:bindings>` `default`s unresolved, and call the node **draft** — `registry-workflow.md:181-186` already defines that vocabulary.

skills/uipath-maestro-bpmn/references/patterns/smart-triage-guide.md
```diff
@@ -89,6 +89,15 @@ reaches manual triage.
...
+agent-job type (e.g. `Orchestrator.StartAgentJob`), `manual_triage`
+`Actions.HITL`, `apply_rules` `Orchestrator.BusinessRules`. **Do not run `uip or
```
rockymadden, line 94: 🔶 **major** — this contradicts `registry-workflow.md:113-119`, which the guide links to nine lines later, and which is where the behavior actually comes from.

`uip or processes` appears **exactly once** in the whole skill: `registry-workflow.md:116`, under "Agent wrapper selection — pick by `processType`, not the label". It *requires* `uip or processes list --all-fields` to choose between `Orchestrator.StartJob` (coded Python agent, `processType: Function`) and `Orchestrator.StartAgentJob` (Agent Builder, `processType: Agent`). The nightly agent's 16 pages were it obeying that rule, not freelancing.

**Failure scenario:** the agent reads this prohibition, hits line 117 ("Fetch payloads through registry-workflow.md"), reads `:113-119`, and holds two live rules — "run `uip or processes list --all-fields` to pick the wrapper" and "do not run `uip or processes`". Line 93 offers `Orchestrator.StartAgentJob` as an `e.g.`, which `:117-119` says is the **wrong** contract for a coded Python agent, so the disambiguation input is removed while a half-right default is supplied. A single verification run can't tell which rule wins on the next model or region.

**Fix:** put the carve-out at `registry-workflow.md:113-119` instead — with no tenant, default to `Orchestrator.StartAgentJob`, leave the process binding unresolved, and note that `processType` selection applies only when a deployed resource exists. One edit; fixes smart-triage, `high-volume-batch` (#3217) and `queue-distribution` (#3210) together, and this paragraph shrinks to naming the three types.

rockymadden (changes requested):
## What

+9 lines at the head of `smart-triage-guide.md` "What to bind": `registry get` the three real types, don't run `uip or processes` / `uip or` / `uip login` for a live deployed instance, don't `registry search` speculatively for `extract` / `handle_a..c`. Fixes a MAX_TURNS regression where the `.bpmn` was never authored.

## Confirmed from the repo

- `check_greenfield.py` grades 2 exclusive gateways, a `userTask`, a 3-way route gateway reached from **both** the confidence gate and the fallback, >=3 end events, a `BPMNDiagram`. No `uipath:*` assertion. Claim holds.
- All three named types exist in `validator/bpmn-spec.json`. The self-correction in `a78039e8d` is right — the first pass's "only manual_triage / apply_rules have a template" was false.
- "placeholder" is used correctly here: `extract` (line 33) and `handle_a..c` (line 38) are both **Placeholder** rows. No legend collision.
- Run `34524230825` exists, succeeded, is this task on `claude`.
- Prose is tight. No slop.

## Overall findings

- 🔶 **major** — the fix is in the wrong file, and the instruction it contradicts still wins. *(inline, line 94)*

- 🔶 **major** — all three types it tells you to fetch are `requiresDiscovery: true`, and the paragraph forbids discovery without giving a third option. *(inline, line 92)*

- 🔹 **minor — false claim in the commit message and the PR comment.** Both say "`Orchestrator.StartAgentJob` / `StartJob` / `RunJob` / `StartAgenticProcess` all exist in `validator/bpmn-spec.json`". **`Orchestrator.RunJob` is not an extension type** — it appears only as an `outputType` / `<uipath:output type=...>` on the job response of `StartJob` and `StartAgentJob` (`bpmn-spec.json:141,153,201,213`). You can't `registry get` it. Same class of error the commit exists to remove, and it's permanent in history. Not shipped in the guide, so no runtime effect. Amend if you rebase.

- 🔹 **minor — n=1 verification, and the turn number doesn't match the cap.** "MAX_TURNS_EXHAUSTED at 138 turns" against `greenfield.yaml:12` `max_turns: 70` — same unit mismatch as #3217. Given the first finding, one iteration can't distinguish "the prohibition wins" from "it won once". Ask for 3.

- **Zoom out:** #3210, #3217 and #3225 are three per-guide patches for one instruction in `registry-workflow.md`. Worth fixing there once instead of patching a fourth guide next week.

The diagnosis is correct and better-targeted than #3217's. But as placed, the fix contradicts `registry-workflow.md:113-119` — the file this guide sends the agent to nine lines later. Moving it there is a smaller edit and fixes every pattern.

## tl;dr

Right root cause, wrong file. `uip or processes` is *mandated* at `registry-workflow.md:116` for exactly this decision; forbidding it in one pattern guide leaves the agent holding two contradictory rules. Move the carve-out up a level and all three guides stop needing their own patch.

nikhil-maryala:
You're right on every point, and the systemic one especially. Verified all of it:
- `uip or processes` **is mandated** at `registry-workflow.md:113-128` for `processType` wrapper selection — my guide forbade the exact rule the agent was obeying. Confirmed.
- All three types are `requiresDiscovery: true`, and I left no third option. Confirmed.
- `Orchestrator.RunJob` is an `outputType`, not an extension type (my commit msg was wrong — same class of error). Confirmed.

Acting on the zoom-out: **moved the carve-out up a level** into `registry-workflow.md` — one edit (greenfield/no-tenant → don't discover live instances, default `StartAgentJob`, author a **draft** with bindings unresolved) that fixes smart-triage, high-volume-batch, and queue-distribution together, using the draft vocabulary you pointed at. That's **#3228**.

**Closing this PR as superseded by #3228.** Filed #3230 for the CI guard you suggested (extension-type/`uipath.*` token check against the spec).

PR #3227, opened 2026-09-10
MST-15064: Update case skill to use v30 schema

charlesliu9:
Schema migration summary: 27.0.0 → 30.0.0
──────────────────────────────────────────────────────────────
Node types added:   none
Node types removed: none
Task types added:   none
Task types removed: none

Field change (the substance of this migration):
  selected-stage-completed / selected-stage-exited
    removed: selectedStageId  (now z.never() — rejected with a targeted error)
    added:   selectedStageIds: string[]   (multi-select, schema v29)
  Applies in all four rule scopes: stage-entry, stage-exit,
  task-entry, metadata.caseExitRules.

Files changed:
  Schema reference:  1
  Plugin docs:       7
  Standalone refs:   3
  Evals:             1
  Test helpers:      1   (_shared/case_check.py)
  Test checkers:     8
  Test fixtures:     9   (all caseplan.json — re-parsed, well-formed)
  Test YAMLs:        0   (no version or node-type literals present)

Leftovers: none
──────────────────────────────────────────────────────────────

Notes:

- v30's own change has no doc surface. v28 is shape-identical to v27, and v30 to v29; v30 itself is a value-only migration normalizing IXP document-extraction output sources =this / =this.result → =result. The skill forbids authoring document-extraction tasks (SKILL.md Rule 16, case-schema.md §7), and no =this.result exists anywhere in the skill or tests — so nothing to migrate there.
- case_check.py was already V30-aware. Its selected_stage_ids() helper reads selectedStageIds first with a legacy selectedStageId fallback, so the only change needed was bumping the schema-range comment. I did not touch the deliberate back-compat assertions in e2e_expense_runnable/test_check_expense_runnable_structure.py:88,107 or athena_cm_event/test_checkers.py:184 — those exist to test the legacy path.
- Two kwarg aliases still read selectedStageId — check_athena_cm_event_case.py:96,195 and the condition() helper in test_cm_golden_expense_checkers.py. Both are internal parameter names that route through selected_stage_ids(), not schema keys, so they work against V30 plans. Left alone as cosmetic; rename them if you'd rather they read canonically.
- Historical version statements preserved (removed at schema v22, required (schema v26), required (as of schema v27)) — these say when something changed, and are still accurate.
- Verification: /usr/bin/python3 -m pytest tests/tasks/uipath-maestro-case -q → 138 passed. All 9 fixtures re-parse as valid JSON, and every changed .py compiles.

Rerun tests:
https://github.com/UiPath/skills/actions/runs/34535099056
https://github.com/UiPath/skills/actions/runs/34536582517


Yiqi517 (approved):
Seems reasonable

PR #3204, opened 2026-09-10
feat(uipath-maestro-case): generate the bindings sidecar with the CLI, close the plain-Valid exit, and grade the strict SDD audit

roberts-cliff:
## Summary

Companion to UiPath/cli#4106 and UiPath/cli#4152, both merged: the strict / SDD validation profile, `bindings sync`, `format`, and the default-profile checks they added. Squashed to one commit on today's `main`.

**Sidecar.** `bindings_v2.json` is generated by `uip maestro case bindings sync`, never hand-written; Check 7 gates on `validate --strict`. `caseplan.json` is written pretty-printed and `uip maestro case format` runs after every write so Edit can patch it — every observed `perl`/`sed`/`node` in-place edit started from a compact file.

**Validation contract.** Strict at the Phase 3 exit and in Phase 4, plain at section ends; the strict validate runs alone, never chained. The repair loop continues while the strict `--sdd` error count falls and halts after two flat rounds or twelve (the three-retry cap stopped every build that needed more than three fixes). Step 12 maps every default- and strict-profile code the CLI emits to its Check, including this week's promotions and additions (`OUTPUT_ID_DUPLICATE`, `CONDITION_EXPR_HOISTED`, `SLA_ID_DUPLICATE`, `CONNECTOR_RESOURCE_KEY_MISMATCH`, `INPUT_BODY_ON_PLAIN_TASK`, `OUTPUT_VAR_ID_CASE_MISMATCH`, `SDD_LITERAL_MISMATCH`). The plain envelope's `Suppressed` block is the reason a plain `Valid` cannot close a build.

**Write-site rules from the golden rounds.** `rule-type:` is the JSON key `rule`; carry the SDD row's Display Name onto conditions; connector inputs are the spec envelope with fields under `body` in exactly the containers the cached spec populates; `metadata.` exposes four runtime fields; `bindings[]` entries and their `=bindings.` references are written together; a connector rule's context `resourceKey` is the key the `connection` entry resolves to.

**Golden task** (`tests/tasks/uipath-maestro-case/golden_rebuild/cm_golden_expense`): grades `validate --strict --sdd` exit 0 (weight 3 — a retreat to the plain profile is a scored miss), in-place script edits of `caseplan.json`, and `debug`/`publish` only at command position (`--help` probes exempt).

## Evidence

[...]

PR #3232, opened 2026-09-10
docs(uipath-maestro-case): map the cli#4160 validator codes in Step 12

roberts-cliff:
Follow-up to #3204. UiPath/cli#4160 (merged, 83f3b766f) added three validator rules; the Step 12 code map names them so the agent knows which Check each belongs to and what the repair is:

- `STRICT_SDD_SLA_CONDITIONAL_MISSING` (`--sdd`, error) — a declared conditional SLA row with no matching `slaRules` entry (Check 11). Found as a stable failure on `sla-from-sdd` across three independent runs.
- `CASE_MGMT_SLA_RULE_UNREACHABLE` (every profile, error) — an unconditional rule with rules stranded after it (Check 11).
- `CASE_MGMT_OUTPUT_ORIGINAL_VAR_MISSING` (every profile, error) — a `->` reassignment without `originalVar` (Check 8/10). Found as a stable failure on `io-binding-matrix` across three independent runs.

Docs only. Verified against the merged dist (maestro-tool chunk `0qhecz56`): 26/26 grader-passing plans zero errors on plain; seventeen golden PASS artifacts zero on plain, `--strict`, `--sdd`.


roberts-cliff:
Stale: the three cli#4160 codes are mapped on main via #3365.

PR #3235, opened 2026-09-11
fix(maestro-case): carry the Design Rationale into stage and task descriptions

Yiqi517:
## The problem

An SDD task or stage block writes two prose lines and the caseplan element has one `description` slot. The skill named the wrong line for it, and the field was not in the task shapes at all, so most builds wrote nothing and the ones that wrote something disagreed about what.

**Where the gap came from.** [#2290](https://github.com/UiPath/skills/pull/2290) sent the rationale to an intermediate plan file, and said so: *"Copy each SDD stage/task/SLA `Design Rationale` into `rationale:` on its matching T-entry. Rationale is reviewer/audit context; the execution plugin ignores it when composing JSON."* Task shapes had no `description` then, correctly. [#2943](https://github.com/UiPath/skills/pull/2943) dropped that file and moved the sink into the caseplan: *"copy each stage/task/SLA `Design Rationale` into the element's `description` where the schema has one."* The stage shape had carried `description` since before [#2290](https://github.com/UiPath/skills/pull/2290), so it kept working; the task shapes never had it. That same commit rewrote all nine task `impl-json.md` files, replacing `tasks.md` with "the SDD" line by line, and added the key to none of them.

| | stage shape | task shape |
|---|---|---|
| before [#2290](https://github.com/UiPath/skills/pull/2290) | 2 | 0 |
| after [#2943](https://github.com/UiPath/skills/pull/2943) | 2 | 0 |
| on `main` today | 2 | 0 |

`description` is what Studio Web renders on the task's General tab ([`useTaskGeneralFields.tsx:45-52`](https://github.com/UiPath/PO.Frontend/blob/c6fd27dbf00eefeece8d15be9b3d5eb16cdc1ff7/src/components/case-management/panels/tasks/tasks-fields/general/useTaskGeneralFields.tsx#L45-L52)), so a dropped line is where the reasoning stops reaching whoever opens the case.

## The evidence

**Measured on 501 distinct caseplans, 303 of them paired with the SDD they were built from.** Counted by the rule that judges it, `STRICT_SDD_TASK_DESCRIPTION_MISSING`:

| | count |
|---|---|
| caseplans where a task carries no `description` | 98 |
| task findings across those 98 | 1713 |
| caseplans that wrote their descriptions | 28 |

Each of the 1713 was read back by resolving its own path in its caseplan and looking at the field: every one points at a task whose `description` is empty or absent, 0 at a task that carried the line. 268 of them are on caseplans their grader scored `1.0`.

Two artifacts read field by field say which line the builds reached for. The supplier-onboarding caseplan from [34379400360](https://github.com/UiPath/skills/actions/runs/34379400360) carries 33 task descriptions and every one of them is the block's `**Design Rationale:**`, not its `**Description:**`, at similarity 1.00 against the rationale and 0.12 to 0.46 against the description line. The caseplan from [34510125152](https://github.com/UiPath/skills/actions/runs/34510125152) carries none of its 8.

**The stage shape is the control, and it has carried the field all along.** Of the 737 stages whose SDD block writes prose, 717 carry a description. The 20 that do not sit on plans scoring 0.190 to 1.0, 5 of them on one artifact that scored `1.0`, so stages are the better-served half rather than the solved one. Stages carry it inside `data`. Its placeholder named one source; the task shapes named none.

**Which line belongs in the slot.** The template separates them: [`case-sdd-template.md:298`](https://github.com/UiPath/skills/blob/ff7f7010b5955cd193792accb7fba7fb8b87ced5/skills/uipath-planner/assets/templates/case/case-sdd-template.md#L298) heads the Design Rationale `<Why this task type fits the actor/work and why this activation mode fits.>` and [`:299`](https://github.com/UiPath/skills/blob/ff7f7010b5955cd193792accb7fba7fb8b87ced5/skills/uipath-planner/assets/templates/case/case-sdd-template.md#L299) heads the Description `<What this task does and why it exists in the case plan.>`. The slot shows what the task does, so the description line goes in it and the rationale goes to `tasks/build-issues.md`. A block writing no description puts its rationale in the slot instead, because the slot has nothing else to take. Thirteen files say that now, and rule 6 was the last one that did not.

## What changed

- **The nine task shapes carry `description`.** Seven write it out in full, matching the example task in that file; the two connector shapes keep a bracketed form, because every identity field in those two files is bracketed too.
- **The stage shapes and `planning.md` name one source**, not a source and a condition. The Required Fields table that still sourced `description` from the Design Rationale now reads the same way as the two code blocks below it.
- **[`SKILL.md`](https://github.com/UiPath/skills/blob/ff7f7010b5955cd193792accb7fba7fb8b87ced5/skills/uipath-maestro-case/SKILL.md) rule 6 says it too, fallback included.** It is read on every build while `implementation.md` is read after a link is followed, so leaving the two different meant the instruction an agent got depended on how far it read. Rule 6 sending every rationale to `build-issues.md` would have left a rationale-only block's slot empty, which is the finding the paired CLI rule raises. Frontmatter is untouched, so the activation gate does not trigger ([`activation-gate.yml:57-66`](https://github.com/UiPath/skills/blob/ff7f7010b5955cd193792accb7fba7fb8b87ced5/.github/workflows/activation-gate.yml#L57-L66) compares only the block between the two `---` lines).
- **`implementation.md` names the description line** and puts the SDD line and the caseplan value side by side, so the two are visibly the same string.
- **Step 12 gets Check 18**, which counts `STRICT_SDD_TASK_DESCRIPTION_MISSING` rather than reading the status. The CLI reports it as a warning, so the run closes `Valid` with the prose still dropped. Check 15 already carries that shape for the missing entry rule, and this one is written to match it.
- Each shape gets a one-line source note in that file's own style: a bullet where the file uses bullets, a blockquote in `wait-for-timer`, which uses only blockquotes.

## Verification

**It reaches the build.** All nine runs behind the measurement read `implementation.md` and a task `impl-json.md` before writing tasks, so this was never a reachability failure.

**Two agents, one round each, on the smallest task that reproduces it.** `sla_from_sdd` has eight task blocks and every one writes both prose lines. Before this change both agents copied the rationale into all eight; after it both copy the description line into all eight.

| run | agent | result |
|---|---|---|
| [34559583056](https://github.com/UiPath/skills/actions/runs/34559583056) | codex | 8 / 8 the rationale |
| [34559584475](https://github.com/UiPath/skills/actions/runs/34559584475) | claude | 8 / 8 the rationale |
| [34568004883](https://github.com/UiPath/skills/actions/runs/34568004883) | codex | 8 / 8 the description line |
| [34568006509](https://github.com/UiPath/skills/actions/runs/34568006509) | claude | 8 / 8 the description line |
| [34630128650](https://github.com/UiPath/skills/actions/runs/34630128650) | claude | 8 / 8 the description line, on this head |
| [34638807168](https://github.com/UiPath/skills/actions/runs/34638807168) | codex | 8 / 8 the description line, on this head |

**Regression control.** A control branch off `main` carrying one blank line under the same skill selects the same six smoke tasks. [Control](https://github.com/UiPath/skills/actions/runs/34568085377) and [this head](https://github.com/UiPath/skills/actions/runs/34638804616) agree on `final_status` and `weighted_score` for all six, every one `SUCCESS` at `1.0`.

**The backstop.** Text alone does not close this. [UiPath/cli#4166](https://github.com/UiPath/cli/pull/4166) reports a task whose SDD writes prose and whose caseplan carries no `description`, reading the description line first to match this contract. On the same corpus it reports 1713 findings across 98 caseplans, stays silent on the 28 that wrote their descriptions, and reports nothing on a task that carried the line. It reports them as warnings, which is why Check 18 counts the code.


Yiqi517:
Fixed in `a7f439b77`.

The review said:

> However, the **Required Fields table** (lines 46–47) contradicts both the code blocks and `impl-json.md`
> [...]
> The table's `description` row should source from the `**Description:**` line (matching the code blocks two sections below), with the Design Rationale as the fallback — not the primary source.

That was right, and it was the last place in the plugin still on the old contract. Both rows now read:

| field | source | notes |
|---|---|---|
| `rationale` | sdd.md Design Rationale | ... Carried into `tasks/build-issues.md`, not into the element. |
| `description` | sdd.md stage `**Description:**` | That line word for word, written into `data.description`. A block that writes no `**Description:**` puts its `**Design Rationale:**` here instead. |

`plugins/stages/` now says the same thing in all three places: this table, the two planning code blocks, and the `impl-json.md` note.

On the second item under Issues for Manual Review, whether stages intentionally differ from tasks: they do not. Both take the `**Description:**` line, and the source for both is the planner template, which heads them separately at `case-sdd-template.md:298` (Design Rationale) and `:299` (Description).

Yiqi517:
@claude please review the current head. The branch has changed since it was opened: SKILL.md rule 6 gained the fallback clause, implementation.md gained Check 18, and main was merged in.

roberts-cliff (approved):

PR #3236, opened 2026-09-11
fix(case): name the third column of the SDD's Task envelope table

Yiqi517:
## The problem

Eight places in this skill tell the build to read the SDD's **Task envelope** table, and every one of them names two of its three columns. The third is `Skip Condition`, and it maps to the caseplan's `skipCondition`. The phrase does not appear anywhere in the skill.

The field itself is documented, in [`case-schema.md:473`](https://github.com/UiPath/skills/blob/84a6a3b7e3a50e2cecf125ee6fc96ddf165b8106/skills/uipath-maestro-case/references/case-schema.md#L473) and in the timer plugin's JSON shape. Nothing connects the SDD column to the key.

Naming two of three reads as a complete list, which is worse than saying nothing: the reader has no reason to look for a fourth cell it was not told about.

## Which run

[33987705899](https://github.com/UiPath/skills/actions/runs/33987705899), 2026-09-05, scored 0.94. `uip maestro case validate` returned `Valid` and the task's guard check rejected the plan, reporting that the 500000 sign-off threshold appears in no guard and lives only inside the workflow that computes the tier, so the case cannot enforce it.

That SDD gives `Obtain procurement director sign-off` the skip condition `=js:vars.expectedAnnualSpend < 500000`. Of its 32 Task envelope tables it is the only one with the column filled.

**The build had the requirement and no place to put it.** Across 748 turns and 442 tool calls, the SDD's `Skip Condition` heading enters the context 32 times and the expression's own variable 58 times, while the caseplan key name `skipCondition` enters 0 times. It read `action/impl-json.md` and `implementation.md` in full, one call each, with no offset: whatever those two files say about the envelope was in front of it.

## What it costs

The plan's entry guard on that task is correct and fully wired: `selected-tasks-completed` on `Determine sign-off tier`, expression `=js:vars.directorSignOffRequired === true`. The director is still gated, so this is not a task that opens on every application.

What is lost is the second gate the SDD asks for in its own Design Rationale: the director task's own skip condition carrying the 500000 numeral executably, so the policy is enforced case-side and not only inside the workflow. Without it, a deployed workflow returning the wrong flag reaches the director with nothing case-side to stop it.

The engine evaluates the field in `PreExecutionPhaseHandler`: when the element is reached and the expression is truthy it records `NotExecuted` and continues the outgoing flows. A missing or whitespace value means the element executes normally.

## What changed

Eight files, 14 added lines.

`implementation.md` carries the mapping: the envelope table has three columns, the third is `Skip Condition`, an expression in that cell becomes the key and a dash means omit it, and some SDDs write the same three fields as rows of a vertical Field and Value table rather than one row of a horizontal one. Both shapes occur in the artifacts on hand.

The seven task plugins carry the key rather than the column heading. Each one's envelope bullet now reads `isRequired`, `shouldRunOnlyOnce` and `skipCondition` from the table's three columns, and six of them gain the key in their JSON shape block; the timer plugin already had it there.

The placement rule, top-level beside `data` rather than inside it, stays where it already is in `case-schema.md` with its WRONG and RIGHT pair, rather than being restated in seven plugin files.

## Verification

**Mechanism, against the run's own artifact and its own grader.** `check_guards.py` at that run's commit, on that run's caseplan: exit 1 with the threshold finding, `guards: 18`. Putting the key back on the same plan: exit 0, `guards: 19`, and the criterion reports the threshold as enforced case-side.

**Corpus.** Every local caseplan paired with the `sdd.md` it was built from, 286 pairs. Of the skip conditions an SDD declares on a task the plan created, 38, the plan carries 26 and drops 12. No plan carries a skip condition its SDD does not declare.

**Repo checks.** `npm run skills:validate` reports OK for both flavors: default 27 skills and 1768 files, studioweb 27 skills, 1768 files and 264 replacements. The edited passage in `implementation.md` sits outside all 14 of that file's flavor blocks, so studioweb inherits it, and the seven plugin files carry no flavor marker and no override.

**The wording carries the repair, not the guarantee.** Nine codex rounds across three versions of this sentence wrote the key zero times, with the wording and the JSON shape block both in context. [UiPath/cli#4167](https://github.com/UiPath/cli/pull/4167) is what reports a dropped cell: `STRICT_SDD_SKIP_CONDITION_MISSING` under `--sdd`, which fires on this run's own caseplan and on 12 plans across the corpus.

### DIFF
--- a/skills/uipath-maestro-case/references/implementation.md
+++ b/skills/uipath-maestro-case/references/implementation.md
@@ -184,7 +184,7 @@ Preserve each explicit `selected-tasks-completed` row and selector only after co
 
 **Pass `lane: <n>` on every task** only when required by the artifact contract. Default: increment per task within a stage starting at 0; lane is a `data.tasks` task-set index. A strict sequential chain is represented as consecutive single-task sets (`[[A], [B], [C]]`) plus `runs-sequentially` on each task. Reuse the same lane only for intentionally parallel siblings, including stage-start siblings (`[[A, B], [C]]`) and siblings after a predecessor (`[[A], [B, C], [D]]`). Sequencing comes from the task's `entryConditions` and the order of task sets in `data.tasks`, not from lane-sharing alone.
 
-**Task envelope fields.** Write `isRequired` and `shouldRunOnlyOnce` from the SDD's **Task envelope** table (`Required` / `Run Only Once`). If `runOnlyOnce` is omitted, default `shouldRunOnlyOnce` to `false` to match frontend new-task behavior. Do not infer `true` from task type; re-entry semantics from the SDD are the source of truth.
+**Task envelope fields.** Write `isRequired`, `shouldRunOnlyOnce` and `skipCondition` from the SDD's **Task envelope** table (`Required` / `Run Only Once` / `Skip Condition`), which some SDDs write as rows of a vertical `| Field | Value |` table instead. If `runOnlyOnce` is omitted, default `shouldRunOnlyOnce` to `false` to match frontend new-task behavior. Do not infer `true` from task type; re-entry semantics from the SDD are the source of truth.
 
 ### Step 9.1 — Placeholder tasks for unresolved resources
 
--- a/skills/uipath-maestro-case/references/plugins/tasks/action/impl-json.md
+++ b/skills/uipath-maestro-case/references/plugins/tasks/action/impl-json.md
@@ -12,6 +12,7 @@
   "elementId": "Stage_aB3kL9-ty5UcykfU",
   "isRequired": true,
   "shouldRunOnlyOnce": false,
+  "skipCondition": "=js:vars.skipReview === true",
   "data": {
     "taskTitle": "Please review this PO and approve or reject",
     "priority": "High",
@@ -26,7 +27,7 @@
 ```
 
 - `id`: `t` + 8 alphanumeric chars. `elementId`: `${stageId}-${taskId}`.
-- `isRequired` and `shouldRunOnlyOnce` come from the SDD task envelope; default `shouldRunOnlyOnce` to `false` when omitted. Do not infer run-once from task type.
+- `isRequired`, `shouldRunOnlyOnce` and `skipCondition` come from the SDD task envelope's three columns; default `shouldRunOnlyOnce` to `false` when omitted. Do not infer run-once from task type.
 - `data.name` / `data.folderPath` MUST be `=bindings.<id>` references — never literals.
 
 ## Action-Specific Fields
--- a/skills/uipath-maestro-case/references/plugins/tasks/agent/impl-json.md
+++ b/skills/uipath-maestro-case/references/plugins/tasks/agent/impl-json.md
@@ -12,6 +12,7 @@
   "elementId": "Stage_aB3kL9-tH3kLmNp9",
   "isRequired": true,
   "shouldRunOnlyOnce": false,
+  "skipCondition": "=js:vars.skipReview === true",
   "data": {
     "name": "=bindings.bG0SraLpg",
     "folderPath": "=bindings.bH1iJK2lm",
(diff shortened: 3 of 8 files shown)


Yiqi517:
Closing this. It teaches the build to write a task's `skipCondition` from the SDD, and that field turns out to be read by nothing. Merging it would make the build emit config the product drops at pack time.

## What the field turned out to be

A case task's `skipCondition` is not read by anything. It is a schema v4 leftover. The frontend migration at [`CaseManagementJsonV4ToV5.ts` lines 20 to 42](https://github.com/UiPath/PO.Frontend/blob/655db12a7605cafcd716406366f0612d76d20d06/src/types/case-mgmt-migrations/v4-to-v5/CaseManagementJsonV4ToV5.ts#L20-L42) converts it to a condition and then deletes the key. Our caseplans declare version 27, so that step never runs on them.

## Measured at pack time, not inferred

`uip maestro case pack` serialises the case through `@uipath/case-schema`, at [`case-pack-service.ts` lines 118 to 122](https://github.com/UiPath/cli/blob/3672153211c7c20ee6aaf18960f753786fea1bfd/packages/case-tool/src/services/case-pack-service.ts#L118-L122). I ran that same package, version 0.1052.3, over two real caseplans.

On a plan carrying `"skipCondition": "=vars.skipReview"`:

| what was counted | result |
|---|---|
| `skipCondition="` as a BPMN attribute | 0 |
| `skipCondition` in the runtime rules JSON | 0 |
| the variable name in the runtime rules JSON | 0 |

The one textual occurrence sits inside a CDATA copy of the authoring document, which the designer reads and the engine does not.

The attribute form is the only one the engine parses, at [`UiPathExtensionsProvider.cs` line 254](https://github.com/UiPath/PO.BpmnEngine/blob/7ea79b35971123615d55f7e3fe83d3790fb92953/src/BpmnEngine/UiPath.PO.BpmnParser/ExtensionsProviders/UiPathExtensionsProvider.cs#L254). With no attribute the element field stays null, and [`PreExecutionPhaseHandler.cs` line 209](https://github.com/UiPath/PO.BpmnEngine/blob/7ea79b35971123615d55f7e3fe83d3790fb92953/src/BpmnEngine/UiPath.PO.BpmnEngine/Execution/Phases/PreExecutionPhaseHandler.cs#L209) returns false. The task always runs.

The same serialiser on the supplier onboarding plan gives the control. Its sign-off task declares the gate twice: once as `skipCondition` on the task, once as an entry condition. The entry condition's variable appears 4 times in the runtime rules, among 38 expression conditions. The skip expression's variable appears 0 times.

## The runtime has no task skip at all

[`CaseDeterministicRules.schema.json`](https://github.com/UiPath/dmnscheduler/blob/bfeec30a8eaa6478df2073479a781df1a6d6cc29/schemas/CaseDeterministicRules.schema.json) is what the scheduler executes. Its `Task` has three properties: `name`, `entryConditions` and `runOnlyOnce`. The word skip appears 0 times in the whole file.

Skipping is what happens when no entry condition holds. The scheduler marks the task at [`Scheduler.cs` lines 702 to 704](https://github.com/UiPath/dmnscheduler/blob/bfeec30a8eaa6478df2073479a781df1a6d6cc29/Scheduler/Core/Scheduler.cs#L702-L704).

## Why the obvious repair is not a mechanical one

Negating the expression onto the task's entry condition is not enough on its own. A task that never enters never completes, and a stage whose completion names that task then never completes either.

The fixture in this repo shows it. Its Review stage completes on the tasks-completed condition naming `Hold For 1 Hour`, which is the task the skip is written on. Gating only that task's entry would hang the stage whenever the flag is true.

A correct translation also decides whether the stage should still complete without the task. That is a design decision the SDD has to state, not something a build step can infer.

## What follows

The planner's SDD template still carries the column, at [`case-sdd-template.md` line 309](https://github.com/UiPath/skills/blob/f746e7d946130c01796a3b08121ca28149b8ca69/skills/uipath-planner/assets/templates/case/case-sdd-template.md#L309), and its case design guide gives no rule for when to fill it. Across the caseplan and SDD pairs on hand, 46 declared skip conditions reach a task the build created, and 0 of them survive into anything the runtime reads.

Three follow-ups, in order:

1. Drop the column from the planner's case SDD template. Nothing downstream can honour it.
2. Fix [`linear_three_stages`](https://github.com/UiPath/skills/blob/f746e7d946130c01796a3b08121ca28149b8ca69/tests/tasks/uipath-maestro-case/multi_stages_tasks/linear_three_stages/linear_three_stages.yaml#L169), the one task whose inline SDD fills the cell. Its checker grades the key at [lines 108 to 113](https://github.com/UiPath/skills/blob/f746e7d946130c01796a3b08121ca28149b8ca69/tests/tasks/uipath-maestro-case/multi_stages_tasks/linear_three_stages/check_linear_three_stages.py#L108-L113), inside a criterion weighted 5.0, and it has never failed because the build writes the key on request.
3. If the column has to stay for a while, the validator rule worth having rejects the key rather than asking for it. That has one correct repair and needs no design decision from the agent.

PR #3239, opened 2026-09-11
fix(maestro-flow/e2e): accept per-outcome exits in the devcon HITL completion gate

dmetzgar:
The DevCon HITL completion gate accepted only one of two correct authoring shapes, so the grade turned on which one the agent happened to pick.

## The two shapes

- The base node's **1.0** definition — and every variant (quick-form, action-app, document-validation) — declares one source handle: `completed`.
- **`outcomePorts: true`** / `exposeError: true` select the base node's **1.1/1.2** definition, whose only source handle is `outcome-{item.id}`, repeated over the outcomes. There is no `completed` handle there, and an edge to one is refused by `uip maestro flow validate` as an undeclared source handle.

The gate required `completed`, or the `outcome-completed` special case #1477 added — which only helps when an outcome is literally *named* "Completed", so an Approve/Reject design can never satisfy it. The task's `initial_prompt` asks for neither shape; it describes approve/reject review plus a downstream log step.

## What that cost

`skill-flow-e2e-devcon-expense-approval` scored **1.0** when the agent authored the first shape and **0.37** the next run when it authored the second — with nothing in the stack changed in between. `outcomePorts` shipped 2026-08-19, and the SDK's HITL docs are byte-identical across the two SDK versions involved (`git diff v3.31.2 v5.0.9 -- references/hitl.md` is empty). Details in flow-builder-sdk#718.

In the earlier passing comparison run, v1 emitted `…quick-form@1.0` and v2 the base `…human-in-the-loop@1.0` — both on `completed`. Neither agent entered the failing shape, so the gate had never been exercised against it.

## The change

The gate now asserts what it means to assert — that the review **continues somewhere**: an edge on `completed`, or any `outcome-<id>` edge. Nothing wired at all is still a failure, and the message names what it found instead of naming one handle.

The success line no longer claims "uses v1.0 schema … wires completed"; it reports the exits it saw. Nothing in the repo asserted that string.

## Verification

Same fixture, both checkers:

```
old: FAIL: HITL completed handle must be wired
new: OK: HITL node reviewExpense uses a v1.x schema, captures approval + reason,
     continues on ['outcome-approve', 'outcome-reject'], and uses .output paths
```

`_flow_doc` now takes `node_type`, `type_version` and `exit_ports`, defaulting to exactly the shape every existing test used, so the two new cases — the outcome-ports shape, and a no-continuation flow that must still fail — sit alongside unchanged coverage. **11 passed.**

## Upstream half

The doc gap that sent the agent down the unsupported path is fixed separately in flow-builder-sdk#724: `hitl.md` now states that `outcomePorts` *replaces* the `completed` exit and how to choose between the shapes, and `flow check` gained `HITL_COMPLETED_PORT_GONE` so the SDK stops emitting an edge the product rejects. The two changes are independent — neither waits on the other.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/e2e/check_devcon_expense_approval.py
+++ b/tests/tasks/uipath-maestro-flow/e2e/check_devcon_expense_approval.py
@@ -117,12 +117,40 @@ def main() -> None:
             "or =js:$vars.<node>.output.<field>"
         )
 
-    if not any(
-        e.get("sourceNodeId") == hitl_id
-        and e.get("sourcePort") in ("completed", "outcome-completed")
+    # The task's own prompt asks for approve/reject review and a downstream log
+    # step; it does not prescribe HOW the task exits, and the SDK offers two
+    # shapes that are both correct:
+    #
+    #   * the base node's 1.0 definition (and every variant: quick-form,
+    #     action-app, document-validation) declares one source handle,
+    #     `completed`;
+    #   * `outcomePorts: true` / `exposeError: true` select the base node's
+    #     1.1/1.2 definition, whose ONLY source handle is `outcome-{item.id}`
+    #     repeated over the outcomes. There is no `completed` handle there, and
+    #     an edge to one is refused by `uip maestro flow validate` as an
+    #     undeclared source handle.
+    #
+    # This gate used to require `completed` (or the `outcome-completed` special
+    # case added in #1477, which only helps when an outcome is literally NAMED
+    # "Completed"). An agent that took the second shape therefore failed a task
+    # it had satisfied — flow-builder-sdk#718, where the same task scored 1.0 the
+    # run before because the agent happened to pick the first shape. What the
+    # gate means to assert is that the review CONTINUES somewhere, so assert
+    # that.
+    exit_ports = {
+        str(e.get("sourcePort") or "")
         for e in edges
+        if e.get("sourceNodeId") == hitl_id
+    }
+    if not (
+        "completed" in exit_ports
+        or any(port.startswith("outcome-") for port in exit_ports)
     ):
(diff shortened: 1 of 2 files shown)


dushyant-uipath:
https://github.com/UiPath/skills/pull/3180
@dmetzgar I had an open PR under review for the same here - Rocky was reviewing it. 

We can reverify if this one's needed once mine merges?

dmetzgar:
Closing in favour of #3180, which fixes the same gate more thoroughly and from the right premise.

Reconciliation measured and written up in https://github.com/UiPath/skills/pull/3180#issuecomment-5641083769. The short version:

- #3180's author was right that a `typeVersion: "1.0"` node wired with `outcome-*` edges validates clean. The mechanism is that `flow validate` special-cases the `outcome-` prefix — a literal handle like `bogus-port` is refused against the embedded definition, anything shaped `outcome-<id>` is waved through. So per-outcome wiring is legitimate on 1.0 as well, and this PR's framing described the SDK's authoring surface rather than the platform's contract.
- The canvas derives one `outcome-<id>` handle per schema outcome at render time (`getEffectiveHitlHandleCustomization`), with `completed` as the zero-outcome placeholder. So the leniency this PR added — accept `completed` OR any `outcome-*` — would keep blessing a shape the designer does not render.
- #3180 also catches a partially-routed outcome set (`missing: outcome-reject`), which this PR did not.

The one piece worth lifting from here is the fixture parameterisation (`_flow_doc` taking `node_type` / `type_version` / `exit_ports`), if #3180's tests don't already reach those shapes another way.

The SDK half of flow-builder-sdk#718 is merged as flow-builder-sdk#724; the remaining SDK-side gap — the quick-form and action-app variants cannot express per-outcome handles at all — is being filed separately.

PR #3244, opened 2026-09-11
feat(uipath-troubleshoot): cover the API Workflow faults with no JSON signature

rares-baesu-uipath:
Three new troubleshooting playbooks for API Workflows.

## The problem

Our API Workflow troubleshooting guide could only diagnose one kind of failure: the workflow ran, and something inside it threw an error.

It had nothing for the three cases where the workflow file itself is fine:

- it never ran at all
- it stopped early because it hit a platform limit
- its outbound call never reached the target system

These are the expensive ones. `validate` reports the workflow as valid, running it locally works, so the investigation ends in "looks healthy to me" and the user is no better off.

It is also why API Workflow sits at **65% on Troubleshoot** — its weakest column, and the only one under the 70% bar ([scorecard](https://uipath.atlassian.net/wiki/spaces/CA/pages/91109195944), scored against the 12 capabilities in [coder-coverage](https://github.com/UiPath/coder-coverage/blob/main/products/api-workflows.md)).

## What's in here

| Playbook | When you'd reach for it |
|---|---|
| `never-ran-no-job.md` | "It didn't run" — and there's no job to look at. The trigger is off, points at an old deployment, its connection is dead, or whatever was supposed to call the workflow never did. |
| `platform-limits.md` | The run stopped early and nothing explains why. A Script ran out of time, a loop hit its iteration limit, or the job hit the 15-minute serverless ceiling. |
| `outbound-call-blocked.md` | A call to a customer's firewalled system times out, but the connection is healthy. API workflows leave the platform by two different routes with different IP ranges, so the customer can have one allowlisted and not the other. |

Each playbook is a checklist: what the symptom looks like, which commands to run, and what to do with the answer.

They are also registered in the routing table the agent uses for failures with no error message to search on. Without that entry, nothing would ever route to them.

## Two things the docs get wrong

I checked every claim against the product source — the API-Workflow and StudioWeb repos, the CLI repo, and the shipped CLI package — rather than the docs alone. Two of them didn't survive that, and both are worth passing to whoever owns the doc pages:

- **The Script timeout is 10 seconds, not the 30 the docs claim.** The 30-second value is real, but it sits in a code path Script activities don't go through. Every Script — in the designer, in `uip api-workflow run`, and in the cloud — gets 10 seconds. The playbook says so and matches on the error message rather than a stopwatch, in case that number moves again.
- **Hitting a loop's iteration limit does not fail the run.** The loop simply stops, the workflow carries on with partial data, and all you get is one warning line in the log. So it reaches the user as a data problem, not an error — which is a completely different thing to look for. The docs describe the `Limit` property but never say what happens when you reach it.

Two smaller corrections along the way: `Log Message` isn't the only way a workflow can log (a `console.log` in any Script lands in the same place), and API workflows are billed in Integration Activities, not Robot Units.

Every CLI command in the playbooks was run through `--help` before it went in. The catalog snapshot this repo keeps only lists verb names, so it will happily confirm a command that fails for want of a `--folder-path`. The agent isn't allowed to go exploring with `--help` mid-investigation, so a command that doesn't work is a dead end.

## Where each fact came from

| Claim | Checked against |
|---|---|
| Script gets 10 seconds | `API-Workflow` main + the shipped `@uipath/api-workflow-tool` 1.201.0 |
| Timeout error reads `Script execution failed timed out` | executor source + its English strings file |
| A loop at its limit stops quietly and logs a warning | `do-while-task-handler.ts` / `while-task-handler.ts` |
| Scripts can't call out (`fetch is not allowed in scripts`) | executor sandbox |
| Cloud runs use the same executor, on serverless | `api-workflow-runtime` (the Deno service) |
| Command shapes and field names | `uipcli` main |
| The two outbound routes, and how usage is billed | docs.uipath.com |

## Not included

No test tasks — grading any of these needs a cloud job that actually fails, which is a sandbox problem worth solving separately.

Still uncovered for troubleshooting: rate limits (429) and the `Wait` activity, file attachments, input that doesn't match the workflow's schema, and the `evals/` loop.


skills/uipath-troubleshoot/references/products/api-workflows/playbooks/never-ran-no-job.md
```diff
@@ -0,0 +1,66 @@
...
+
+| How it starts | Set up as | Why it stops starting | Where the record is |
+|---|---|---|---|
+| **On a schedule** | A cron trigger in Orchestrator (`--type time`) | Someone turned it off, the cron or time zone is wrong, or it still points at an old deployment | `or triggers history` |
+| **On demand** | **Automations > Processes > Start a job** (`--type api`), an Agent using it as a tool, Maestro's "Start and wait for API workflow", or a `Run Job` activity elsewhere | Whatever was meant to call it never did. The problem is in the caller, not here | `or triggers history` if it is an API trigger — otherwise only the caller's own run |
+| **On an event** | An Integration Service connector event | The connection behind the event is broken or switched off. No connection, no events | Nothing in the CLI. `--type` only covers time, queue and api |
+
+Also check whether the workflow has a trigger at all. Publishing to a personal workspace makes it runnable, but it will not have a trigger unless someone created one.
```
mirastroie: `but it will not have a trigger unless someone created one` - that makes me thing an api workflow can have no trigger, which I don't think its right - the manual trigger is the default one

mirastroie (approved):

skills/uipath-troubleshoot/references/products/api-workflows/playbooks/never-ran-no-job.md
rares-baesu-uipath: Good catch — that sentence was wrong. Fixed.

You're right that an API workflow always has a way to run: `uip or jobs start <process-key>` needs no trigger object at all, and neither does an Agent, a Maestro task action, or a `Run Job` activity calling it. An `api` trigger is a separate, optional thing — it publishes the workflow at a URL slug (`--slug`/`--method` on `or triggers create --type api`), which is not the same as the manual start path.

Two changes:

- The paragraph now reads: "A published workflow can always be started on demand, so 'no trigger' never means 'no way to run it'. What it can be missing is a *schedule* or an *event* trigger — those only exist if someone created them."
- The on-demand row of the table said it was set up as `--type api`, which conflated the two. It now says "Nothing to set up", with the API trigger listed as optional and described as the URL-slug kind.

The resolution bullet changed from "No trigger exists" to "No schedule or event trigger exists" for the same reason.

PR #3245, opened 2026-09-11
fix(uipath-rpa): drop the init scaffold so greenfield authoring writes a new file

Mihaiii:
## Issue

[PILOT-7523](https://uipath.atlassian.net/browse/PILOT-7523)

## Summary

- The T1 `Bash` chain in the greenfield XAML journey now ends with `rm -f "<PROJECT_DIR>/Main.xaml"`, so the T3 `Write` creates a new file instead of overwriting the `init` scaffold.
- Recovers the turn the workaround was costing. This journey runs against a hard `max_turns: 10` in `skill-rpa-execution-map-greenfield`, which has been failing on the `claude-code` variant while `codex` passes.
- One line changed.

## Why

Claude Code's `Write` tool refuses to overwrite a file that was not read earlier in the same session. The refusal comes back as a tool error, not a filesystem or CLI error:

```
File has not been read yet. Read it first before writing to it.
```

`uip rpa init` always creates `Main.xaml` — no flag suppresses it, and the file is materialized by Studio's template engine, not by the CLI. So the map's instruction to author the workflow with one complete `Write` and skip re-reading the scaffold was correct in intent but impossible to execute on this harness: the `Write` always failed.

The recovery path is the problem. After the refusal the agent has to `Read` the very file it is about to discard, in full, purely to satisfy the precondition. That read returns a known-empty scaffold, contributes nothing to the authoring pass, and costs a turn — exactly the kind of waste the execution map exists to eliminate. It is not redundant work that happens to be cheap; it is redundant work that directly contradicts what a turn-budgeted map is for. The other fallback, an `Edit` chain against the scaffold, is worse: it fragments the single authoring pass and breaks Rule 18.

Deleting the file removes the precondition rather than working around it. A fresh create needs no prior read on any harness, so the fix is also a no-op for `codex`, which has no such rule.

Observed across nightly runs of `skill-rpa-execution-map-greenfield`: one run hit the refusal and spent two turns recovering, another fell back to a two-step `Edit` chain, and the 2026-09-11 run obeyed the guidance literally, got the refusal, spent its last turn on the `Read`, and never authored `Main.xaml` at all.

## Reviewer note

Between T1 and T3 the project carries a `main` and an `entryPoints` row pointing at a file that does not exist yet; the T3 `Write` restores it under the same name before the T4 gate runs. No step in that window reads or builds the project. Worth a sanity check that nothing Helm-side caches project state across `analyzer-rules list` in the same chain.

## Testing

- `node scripts/check-skill-links.mjs` — 6779 links, all resolve.
- Behavioral confirmation comes from the next nightly `skill-rpa-execution-map-greenfield` run on the `claude-code` variant.

[PILOT-7523]: https://uipath.atlassian.net/browse/PILOT-7523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Mihaiii:
Codex run: https://github.com/UiPath/skills/actions/runs/34586591603
Claude run: https://github.com/UiPath/skills/actions/runs/34586614931

RaduAna-Maria (approved):

PR #3249, opened 2026-09-11
test(uipath-maestro-flow): grade the native entity read in the billing advisories

rockymadden:
Follow-up to #3231, which named this PR as the way to close its own known gap.

## What

Every advisory on the three DevCon billing tasks reads the entity lookups as Integration Service connector activities. A flow built with the native `core.datafabric.read` node fails them on contracts that node does not have:

| Advisory | Task(s) | Weight | Demands | Native build |
|---|---|---|---|---|
| `advisory_billing_dispute_resolution` | dispute resolution | 1.0 | a `connection` binding | has none |
| `advisory_billing_invoice_lookup` | invoice lookup | 1.0 | `detail.connectionId` + `connectionFolderKey` | has neither |
| `advisory_billing_discrepancy_detector` | discrepancy detector | 1.0 | the same pair, on both reads | the same |
| `check_bindings_no_stubs` | both siblings | 1.0 ea. | `declares no bindings at all` | no `connection` row |
| `check_server_side_filter` | invoice lookup | 1.0 | a `query-entity-records` node | is not one |

This adds one shared definition of "the nodes that read entity records" plus the native equivalent of each shape-specific assertion, and points every call site at it:

```python
CONNECTOR_READ_PREFIX = "uipath.connector.uipath-uipath-dataservice."
NATIVE_READ_TYPE = "core.datafabric.read"

def entity_reads(nodes) -> tuple[str, list[dict]]: ...
```

Each equivalent replaces a connector assertion one-for-one:

| Connector assertion | Native equivalent |
|---|---|
| `entityName` in the `{entityName}` path slot | `entityConfig.entityName` |
| `queryExpression` computed from `$vars` | a `_filters` row whose value is a `=js:` expression on the flow's own input |
| CEQL string literal single-quoted | operator in the supported set, value non-blank, no `$self`, each refuses the **whole** query |
| a non-empty `queryExpression`/`filter`/`filterGroup` | at least one `_filters` row |
| `connectionId`/`connectionFolderKey` uuid pair | entity scope; a connection required only of a flow that has a connector node |

## Why

The native node needs **no** Integration Service connection and **no** `bindings[]` connection row (`data-fabric/planning.md:62`), so demanding one fails the better build for being the better build.

From the 2026-09-11 run of `skill-flow-devcon-billing-dispute-resolution`:

- `uip maestro flow validate` → `Valid`
- the agent built two `core.datafabric.read` nodes, tenant-scoped, filters pushed server-side
- `check_e2e_billing_dispute_resolution.py` → `OK: flow returned a real SOP-grounded resolution citing MCS-2026-04872 (no email sent)`
- all five contract outputs declared with the contract's types
- **score 0.727**, and the advisory's line was:

```
FAIL: the flow declares NO `connection` binding.
```

`pass_threshold: 0` keeps an advisory from gating run *status*, but not from contributing its zero to the weighted score, 8/11 is exactly the arithmetic of two zeros at weights 2 and 1.

## The anti-hardcode guard survives

It is the whole reason the offline behaviour rungs mean anything, so shape must not be a way around it. The native branch requires a `_filters` row whose value is a `=js:` expression on the flow's own input; a literal there fails. `test_native_anti_hardcode_guard_survives` locks it by replacing the derived row's value with the canonical answer and asserting the advisory still fails.

## Deliberately not required

An `Entity` binding row on a **tenant-scoped** read. A tenant-scoped entity needs nothing beyond `entityName` (`impl.md:303`), and the skill tells the agent to stay tenant-scoped when it cannot resolve a `_resourceKey`, calling a half-authored folder scope "worse than none" (`impl.md:333`). The run's flow carried no entity binding and debugged green.

The pair **is** required once `_folderKey` appears, where the name and folder otherwise serialize as source-org literals and break on deploy. Both that and the lowercase `resource: "entity"` form packaging silently ignores are asserted.

## Nothing is loosened for a connector build

- `check_bindings_no_stubs` still fails a connector flow with no connection row (`test_bindings_checker_still_requires_one_for_a_connector_build`), and still fails a stub row wherever one is declared, native build included (`test_bindings_checker_rejects_a_stub_connection_on_a_native_build`).
- `check_server_side_filter` still fails an empty `_filters` container, which is always present and so cannot be read as "filtered" (`test_server_side_filter_rejects_an_empty_native_filter`).
- `check_server_side_filter` has one caller and `check_bindings_no_stubs` two, all three Data Fabric scenarios, so no other task's behaviour can move.

## Fixtures

`BillingDisputeResolution.native.reference.flow` is the 2026-09-11 artifact kept verbatim, the only real native build of any of these scenarios, and the exact flow that scored 0. The other two native cases are derived from their connector reference inside the test, so the computed filter under test stays the reference's rather than something hand-written here.

## Verification

```
pytest tests/tasks/uipath-maestro-flow/ -q
1205 passed in 27.52s
```

14 new tests. Reverting the discriminator is caught **twice, independently**, by the artifact fixture and by the call-site guard, verified by temporarily reverting `advisory_billing_dispute_resolution.py`'s call site and watching both fail. Five negative cases were each run by hand to confirm they fail for the intended reason, not incidentally.

Includes the corpus guards: `test_same_ground_corpus.py`, `test_criterion_budgets.py`, `test_same_ground_advisories.py`, `test_headless_preamble.py`.

## One tightening, not strictly required

`advisory_billing_invoice_lookup` and `advisory_billing_discrepancy_detector` each hand-rolled a **prefix-only** uuid regex (`^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-`). Both now use this module's existing `is_real_uuid`, which full-matches all 36 characters. A truncated uuid used to pass the connector branch and now fails. Both connector references still pass. Easy to back out if you would rather keep the diff pure.

## Measured effect

Run [34626987724](https://github.com/UiPath/skills/actions/runs/34626987724), `run-coder-eval.yml` on this branch, codex / `gpt-5.6-terra` to match the 2026-09-11 run. Pinned explicitly: `CODEX_MODEL` now defaults to `gpt-5.6-luna`, so leaving it blank would have measured a different model against the old numbers.

| Task | Shape built | 2026-09-11 | This run | After `f4872dd` |
|---|---|---|---|---|
| `billing_invoice_lookup` | native, 1 read | 0.727 | **0.909** | 11/11 |
| `billing_discrepancy_detector` | native, 2 reads | 0.800 | **0.900** | 10/10 |
| `billing_dispute_resolution` | connector, 2 reads | 0.727 | **1.000** | 1.000 |

All three FAILURE to SUCCESS. Two built native and exercised the new path; the third built connector-shaped and confirms that path is unbroken end to end, live.

**The run found a defect three reviews missed.** Both native tasks still lost one weight-1 advisory to `AssertionError: no generated bindings.json or bindings_v2.json found`. A pure Data Fabric flow references no packaged resource, so the CLI writes no bindings file at all, and `check_bindings_no_stubs` died one line before the connection check could apply. The earlier commit had fixed the wrong assertion, and the fixture hid it by supplying a file holding a `process` resource, a shape these flows never produce. Fixed in `f4872dd`, with the checker replayed against all three real artifact trees from the run.

The last column is criterion-weight arithmetic on top of that replay, not a second measured run.

**Still unexercised:** `billing_dispute_resolution` built connector-shaped this time, so the native branch of its advisory has live coverage only from the 2026-09-11 artifact, which is the fixture in this PR. Shape choice varies per run, which is the reason both shapes have to be gradeable.

**`billing_dispute_resolution` scored 1.000 without PR 1.** Its structural gate accepts the connector shape, so the gate only passes while the agent happens to build that way. PR 1 is still needed for the native case.

## Self-review, and what it changed

Reviewed in [#3249 (review)](https://github.com/UiPath/skills/pull/3249#pullrequestreview-5181266307): eight findings, two majors. Both majors are fixed in `d9ffb4c`, and both were verified to fail the suite when reverted.

1. **major** the native filter check could not follow a Script hop, which `billing_invoice_lookup`'s own prompt ("normalize the input ... then look up") invites. A false failure on the task this PR unblocks. `read_references_input` now follows it, via a `transitive_refs` factored out of `source_depends_on`.
2. **major** four of five native `fail()` paths had no automated coverage. Added, plus a `groups[]`-nested filter row, a `native_filter_rows` branch nothing reached.
3. **minor** the mixed-shape check was asymmetric and could false-fail a native read beside a connector write, which is a real tenant configuration since each native op has its own flag.
4. **minor** `needs_a_connection` matched only the Data Service prefix, so a future caller using Slack or Jira would silently stop requiring bindings.
5. **minor** `_folderKey` is folder scope by presence, so a blank one was being read as tenant-scoped.
6. **minor** the no-rows message asserted a `resultMode` nothing had checked.
7. **minor** two test-fixture naming problems.
8. **minor** house style: em-dashes, now out of this body. Commit `e56349c` still carries seven in its message; a squash merge composes the landed message from this body, so they do not reach `main`. Say the word if you want that commit rewritten instead.

Suite after the fixes: **1226 passed**, from 1191 on `main`.

### DIFF
--- a/tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
@@ -210,11 +210,10 @@ def entity_reads(nodes: Iterable[dict[str, Any]]) -> tuple[str, list[dict[str, A
         and str(node.get("type") or "").endswith(CONNECTOR_READ_OPS)
     ]
     native = [node for node in nodes if str(node.get("type") or "") == NATIVE_READ_TYPE]
-    connector_reads = connector
-    if connector_reads and native:
+    if connector and native:
         fail(
             f"the flow mixes both entity-read shapes — connector "
-            f"{[node.get('id') for node in connector_reads]} and native "
+            f"{[node.get('id') for node in connector]} and native "
             f"{[node.get('id') for node in native]}. Availability decides one shape per "
             f"flow; two means one set of reads was left behind"
         )
--- a/tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
@@ -112,19 +112,22 @@ def node_dependencies(nodes: Iterable[dict[str, Any]]) -> dict[str, set[str]]:
     }
 
 
-def source_depends_on(value: Any, producer: str, dependencies: dict[str, set[str]]) -> bool:
-    """Follow exact `$vars.<node>` references through any number of reader nodes."""
+def transitive_refs(value: Any, dependencies: dict[str, set[str]]) -> set[str]:
+    """Every node id reachable from a value's `$vars.<node>` references."""
     pending = list(value_refs(value))
     seen: set[str] = set()
     while pending:
         current = pending.pop()
-        if current == producer:
-            return True
         if current in seen:
             continue
         seen.add(current)
         pending.extend(dependencies.get(current, ()))
-    return False
+    return seen
+
+
+def source_depends_on(value: Any, producer: str, dependencies: dict[str, set[str]]) -> bool:
+    """Follow exact `$vars.<node>` references through any number of reader nodes."""
+    return producer in transitive_refs(value, dependencies)
 
 
 def references_field(value: Any, field: str) -> bool:
@@ -155,8 +158,15 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
 # checkers; keep the two in step.
 CONNECTOR_READ = "connector"
 NATIVE_READ = "native"
-CONNECTOR_READ_PREFIX = "uipath.connector.uipath-uipath-dataservice."
+CONNECTOR_NODE_PREFIX = "uipath.connector."
+CONNECTOR_READ_PREFIX = f"{CONNECTOR_NODE_PREFIX}uipath-uipath-dataservice."
 NATIVE_READ_TYPE = "core.datafabric.read"
+# The connector operations that READ. The prefix above covers the whole Data
+# Service family, writes included, and each native op has its own tenant flag
+# (planning.md — Node types), so `read-entity` on with `create-entity` off is a
+# real configuration: a native read beside a connector write. Only a connector
+# READ alongside a native one is a flow with two sets of the same reads.
+CONNECTOR_READ_OPS = (".query-entity-records", ".get-entity-record-by-id")
 
 # Operators the native serializer can emit, spelled as the file spells them
 # (data-fabric/impl.md — Filters). There is deliberately no `not in`, `not
@@ -196,10 +206,14 @@ def entity_reads(nodes: Iterable[dict[str, Any]]) -> tuple[str, list[dict[str, A
         if str(node.get("type") or "").startswith(CONNECTOR_READ_PREFIX)
     ]
     native = [node for node in nodes if str(node.get("type") or "") == NATIVE_READ_TYPE]
-    if connector and native:
+    connector_reads = [
+        node for node in connector
+        if str(node.get("type") or "").endswith(CONNECTOR_READ_OPS)
+    ]
+    if connector_reads and native:
         fail(
             f"the flow mixes both entity-read shapes — connector "
-            f"{[node.get('id') for node in connector]} and native "
+            f"{[node.get('id') for node in connector_reads]} and native "
             f"{[node.get('id') for node in native]}. Availability decides one shape per "
             f"flow; two means one set of reads was left behind"
         )
@@ -246,7 +260,13 @@ def native_filter_rows(node: dict[str, Any]) -> list[dict[str, Any]]:
     return list(walk(entity_config(node).get("_filters")))
 
 
-def assert_read_filters_input(node: dict[str, Any], shape: str, field: str, label: str) -> None:
+def assert_read_filters_input(
+    node: dict[str, Any],
+    shape: str,
+    field: str,
+    label: str,
+    nodes: Iterable[dict[str, Any]] = (),
+) -> None:
     """Assert the read filters on ``field``, computed, in a form the server accepts.
 
     The computed half is the anti-hardcode gate: a filter that writes the answer
@@ -259,14 +279,20 @@ def assert_read_filters_input(node: dict[str, Any], shape: str, field: str, labe
     if shape == NATIVE_READ:
         rows = native_filter_rows(node)
         if not rows:
+            mode = str(unwrap(entity_config(node).get("resultMode")) or "").strip()
+            consequence = (
+                "faults on the over-match as soon as the entity holds more than one row"
+                if mode == "single"
+                else "returns the first 100 rows in arbitrary order, not the record the flow asked for"
+            )
             fail(
                 f"the {label} read sets no `entityConfig._filters` rows — an unfiltered "
-                f"`resultMode: \"multiple\"` read returns the first 100 rows in arbitrary order, "
-                f"not the record the flow asked for"
+                f"`resultMode: {mode or 'multiple'!r}` read {consequence}"
             )
-        if not read_references_input(node, shape, field):
+        if not read_references_input(node, shape, field, nodes):
             fail(
-                f"no {label} filter row takes its value from $vars.…{field} — the rows are "
+                f"no {label} filter row reaches $vars.…{field}, directly or through a node it "
+                f"reads — the rows are "
                 f"{[{'field': row.get('field'), 'value': unwrap(row.get('value'))} for row in rows]!r}. "
                 f"Each lookup must filter on its own flow input; a literal there is the answer "
                 f"written in"
@@ -297,7 +323,7 @@ def assert_read_filters_input(node: dict[str, Any], shape: str, field: str, labe
     if "queryExpression" not in query:
         fail(f"the {label} query sets no queryExpression; queryParameters: {sorted(query)}")
     expression = str(query["queryExpression"])
-    if not read_references_input(node, shape, field):
+    if not read_references_input(node, shape, field, nodes):
         fail(
             f"the {label} queryExpression is {expression!r} and its filterVariables do not reference "
             f"{field!r} — each lookup must filter on its own flow input. Author the filter via "
@@ -311,15 +337,42 @@ def assert_read_filters_input(node: dict[str, Any], shape: str, field: str, labe
         )
 
 
-def read_references_input(node: dict[str, Any], shape: str, field: str) -> bool:
-    """True when the read's filter takes its value from the named flow input."""
-    if shape == NATIVE_READ:
-        return any(
-            "$vars." in str(unwrap(row.get("value")) or "")
-            and references_field(row.get("value"), field)
-            for row in native_filter_rows(node)
-        )
-    return query_references_input(read_detail(node), field)
+def read_references_input(
+    node: dict[str, Any],
+    shape: str,
+    field: str,
+    nodes: Iterable[dict[str, Any]] = (),
+) -> bool:
+    """True when the read's filter reaches the named flow input.
+
+    A filter row may read the input directly, or read a node that does: the
+    prompts ask for normalisation before the lookup ("trim, uppercase, prepend
+    `MCS-`"), and a native row is a single `value`, so a Script hop is at least
+    as natural there as the connector's inline template. The same advisory
+    already tolerates the hop for outputs, and for the same reason.
+    """
+    if shape != NATIVE_READ:
+        return query_references_input(read_detail(node), field)
+
+    nodes = list(nodes)
+    inputs_by_id = {
+        str(other.get("id")): json.dumps(other.get("inputs") or {}, sort_keys=True)
+        for other in nodes
+        if other.get("id")
+    }
+    dependencies = node_dependencies(nodes)
+    for row in native_filter_rows(node):
+        value = row.get("value")
+        if "$vars." not in str(unwrap(value) or ""):
+            continue
+        if references_field(value, field):
+            return True
+        if any(
+            references_field(inputs_by_id.get(upstream, ""), field)
+            for upstream in transitive_refs(value, dependencies)
+        ):
+            return True
+    return False
 
 
 def assert_read_resolves(node: dict[str, Any], shape: str, label: str, flow: dict[str, Any]) -> str:
@@ -340,9 +393,15 @@ def assert_read_resolves(node: dict[str, Any], shape: str, label: str, flow: dic
     """
     if shape == NATIVE_READ:
         config = entity_config(node)
+        if "_folderKey" not in config:
+            return "tenant-scoped entity (no connection, no binding row)"
         folder = str(unwrap(config.get("_folderKey")) or "").strip()
         if not folder:
-            return "tenant-scoped entity (no connection, no binding row)"
+            fail(
+                f"the {label} read declares `_folderKey` but leaves it blank ({config.get('_folderKey')!r}). "
+                f"It is the key's PRESENCE that switches the emitted target to the folder-qualified form, "
+                f"so a blank one emits that form with no folder. Omit the key to stay tenant-scoped"
+            )
 
         key = str(unwrap(config.get("_resourceKey")) or unwrap(config.get("_entityKey")) or "").strip()
         if not key:
--- a/tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
+++ b/tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
     )
 
 
+# ── entity reads: two shapes, tenant availability decides which ──────────────
+# The `uipath-uipath-dataservice` connector activity and the native
+# `core.datafabric.read` node both read entity records. The native flags default
+# to off, so the connector is the working path until `registry get` proves
+# otherwise; where the tenant does have them the native node is "the better
+# build" (data-fabric/planning.md — Native node vs Data Service connector).
+# `flow_check.ENTITY_QUERY_HINTS` gates the same two shapes for the non-advisory
+# checkers; keep the two in step.
+CONNECTOR_READ = "connector"
+NATIVE_READ = "native"
+CONNECTOR_READ_PREFIX = "uipath.connector.uipath-uipath-dataservice."
+NATIVE_READ_TYPE = "core.datafabric.read"
+
+# Operators the native serializer can emit, spelled as the file spells them
+# (data-fabric/impl.md — Filters). There is deliberately no `not in`, `not
+# contains`, or null check, and an operator outside this set refuses the WHOLE
+# query: the node emits nothing and every downstream `$vars` reference breaks.
+NATIVE_FILTER_OPERATORS = {
+    "=",
+    "!=",
+    ">",
+    ">=",
+    "<",
+    "<=",
+    "contains",
+    "starts with",
+    "ends with",
+    "in",
+}
+
+
+def entity_config(node: dict[str, Any]) -> dict[str, Any]:
+    """The native read's single input container (data-fabric/impl.md)."""
+    config = (node.get("inputs") or {}).get("entityConfig")
+    return config if isinstance(config, dict) else {}
+
+
+def read_detail(node: dict[str, Any]) -> dict[str, Any]:
+    """The connector activity's `detail` envelope."""
+    detail = (node.get("inputs") or {}).get("detail")
+    return detail if isinstance(detail, dict) else {}
+
+
+def entity_reads(nodes: Iterable[dict[str, Any]]) -> tuple[str, list[dict[str, Any]]]:
+    """Return ``(shape, reads)`` for the nodes that read entity records."""
+    nodes = list(nodes)
+    connector = [
+        node for node in nodes
+        if str(node.get("type") or "").startswith(CONNECTOR_READ_PREFIX)
+    ]
+    native = [node for node in nodes if str(node.get("type") or "") == NATIVE_READ_TYPE]
+    if connector and native:
+        fail(
+            f"the flow mixes both entity-read shapes — connector "
+            f"{[node.get('id') for node in connector]} and native "
+            f"{[node.get('id') for node in native]}. Availability decides one shape per "
+            f"flow; two means one set of reads was left behind"
+        )
+    return (NATIVE_READ, native) if native else (CONNECTOR_READ, connector)
+
+
+def entity_name(node: dict[str, Any], shape: str) -> str:
+    """The entity a read addresses. Fails naming the slot the name belongs in."""
+    if shape == NATIVE_READ:
+        config = entity_config(node)
+        name = str(unwrap(config.get("entityName")) or "").strip()
+        if not name:
+            fail(
+                f"read node {node.get('id')!r} sets no `inputs.entityConfig.entityName`; "
+                f"entityConfig keys: {sorted(config)}"
+            )
+        return name
+
+    detail = read_detail(node)
+    path = {key: unwrap(value) for key, value in (detail.get("pathParameters") or {}).items()}
+    if "entityName" not in path:
+        query = {key: unwrap(value) for key, value in (detail.get("queryParameters") or {}).items()}
+        fail(
+            f"query node {node.get('id')!r} has pathParameters {sorted(path)}; `entityName` "
+            f"belongs in the PATH slot — it is the {{entityName}} of /v2/{{entityName}}/qer. "
+            f"queryParameters: {sorted(query)}"
+        )
+    return str(path["entityName"]).strip()
+
+
+def native_filter_rows(node: dict[str, Any]) -> list[dict[str, Any]]:
+    """Root filter rows plus every nested group's rows (impl.md — Filters)."""
+
+    def walk(block: Any) -> Iterable[dict[str, Any]]:
+        if isinstance(block, list):  # the legacy bare-array form, still read
+            yield from (row for row in block if isinstance(row, dict))
+            return
+        if not isinstance(block, dict):
+            return
+        yield from (row for row in (block.get("rows") or []) if isinstance(row, dict))
+        for group in block.get("groups") or []:
+            yield from walk(group)
+
+    return list(walk(entity_config(node).get("_filters")))
+
+
+def assert_read_filters_input(node: dict[str, Any], shape: str, field: str, label: str) -> None:
+    """Assert the read filters on ``field``, computed, in a form the server accepts.
+
+    The computed half is the anti-hardcode gate: a filter that writes the answer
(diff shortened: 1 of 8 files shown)


nikhil-maryala:
## Review — approve on the code; one merge gate

Went through the whole diff and ran the suite locally. This correctly fixes the real defect: three billing advisories assumed entity lookups are IS connector activities and demanded a `connection` binding + `detail.connectionId/connectionFolderKey` pair, so a native `core.datafabric.read` build got penalized for being the better build (the 2026-09-11 run scored 0.727 on a flow that validated, debugged green, and returned the right answer).

### Verified independently

| Claim | Result |
|---|---|
| `1205 passed` | ✅ Reproduced on Python 3.13. (On 3.9 I saw 133 failures — all `match`-statement version artifacts, **not** PR-caused.) |
| Connector path unchanged | ✅ Both connector reference flows still pass; connector branch is the old logic |
| Native path not tautological | ✅ `_to_native_reads` lifts the reference's own `${…}` CEQL expression into the native `_filters` row — the filter under test stays the reference's |
| Anti-hardcode guard survives | ✅ `test_native_anti_hardcode_guard_survives` rewrites the row value to the canonical answer and asserts failure; double-locked by verbatim fixture + call-site |
| Nothing loosened | ✅ `check_bindings_no_stubs` still fails a connector build with no connection and a stub row on either build; `check_server_side_filter` still fails an empty native `_filters` |
| Blast radius | ✅ `check_server_side_filter` 1 caller, `check_bindings_no_stubs` 2 — all three DF scenarios; no other task can move |

Tests are behavior-graded (subprocess returncode + intended-reason substring), consistent with the repo's "grade side effects, not self-reports" rule.

### Minor, non-blocking

1. **Latent `0`-value edge** — `assert_read_filters_input` native branch does `value = str(unwrap(...) or "").strip()`, so a legitimate numeric `0`/`False` filter value would be flagged blank. Irrelevant here (values are `$vars.…` strings), but a footgun if the helper is reused for numeric filters.
2. **uuid tightening** (prefix-regex → full `is_real_uuid`) is a real behavior change on the connector branch — a truncated uuid that used to pass now fails. Both connector references still pass, so it's safe; just calling it out as more than a pure refactor.
3. **Import-style asymmetry** — `check_server_side_filter.py` uses `from _shared.advisory_flow_utils import …` (path-walk) while `_shared/advisory_*.py` use the flat form. Two module identities, harmless (no shared state), driven by its leaf-dir location.

### Merge gate
Per the PR's own honest disclosure — **"the eval has not been rerun"** — the predicted jumps (invoice 0.727→1.000, discrepancy 0.800→1.000, dispute 0.727→0.818) are forecasts. Recommend dispatching a coder-eval run **from this branch** to turn them into evidence before merge. Also as the author notes: dispute-resolution stays FAILURE until the separate PR fixes `check_structure_billing_dispute_resolution.py`'s hardcoded `dataservice.query-entity-records` — expected, not a regression here.

Code is correct, thoroughly tested, preserves all prior guarantees. 👍

nikhil-maryala (approved):

rockymadden (commented):
Self-review, done against the pushed diff rather than recollection of it. Eight findings, two of which I would block merge on.

## What

Teaches five connector-shaped checkers on the three DevCon billing tasks to grade a native `core.datafabric.read` build: one shared discriminator in `advisory_flow_utils.py`, native equivalents for each shape-specific assertion, five call sites repointed, and the 2026-09-11 artifact added as a fixture.

## Overall findings

**major :red_circle: missing tests.** Four of the five `fail()` paths in the native branch of `assert_read_filters_input` have no automated coverage: no-rows, unsupported-operator, blank-value, and `$self`. Three were exercised by hand while writing this, which is not verification the repo keeps. `native_filter_rows`' nested-`groups[]` walk and its legacy bare-array branch are also uncovered, so a flow that filters via a group runs code nothing tests.

*Fix:* three parametrized cases against the derived fixture, plus one group-nested filter.

**minor :yellow_circle: the effect table overstates what is known.** The table claims the two sibling tasks reach **1.000**. That is arithmetic on criterion weights, not a measured result. `check_billing_invoice_lookup` drives three debug runs and only one of those inputs has ever run against a native build. The caveat is in the PR body, but it sits below a table that reads as a result.

*Fix:* soften the table, or lift the caveat above it.

**minor :yellow_circle: house style.** Seven em-dashes in the commit body, eight in the PR body, against the repo owner's no-em-dash rule. Flagging rather than silently fixing, because the repo baseline leans the other way: one pre-existing advisory in this same directory has twelve, and #3231's body uses them freely. The rule covers commit messages and PR bodies unambiguously; whether it should also rewrite new code comments inside files saturated with them is a separate call.

## Recommendation

**Request changes.** The two majors should land before merge. The rest is fine to defer or wave off.

CI being green across 34 checks is not evidence against either major. Both concern paths CI never executes.

## Verification honesty

Confirmed by reading code and docs: the asymmetry at `advisory_flow_utils.py:199`, the four independent tenant flags, the coverage gaps, that narrowing `needs_a_connection` to `uipath.connector.` is safe, that all three billing scenarios are read-only, and the em-dash counts.

Reasoned but **not** confirmed: whether an agent would actually take the Script-hop route on `billing_invoice_lookup` (the prompt makes it plausible; no run has produced one), and the empty-`_folderKey` serializer behaviour.

Tests were read, not re-run. The only execution was the CI run already reported on this PR.

## tl;dr

Two majors: four of five new native assertion paths are untested, and the native filter check cannot follow a Script hop that `billing_invoice_lookup`'s own prompt invites, which is a false failure on the task this PR unblocks. Five minors plus one style violation introduced here. Green CI proves nothing about either major.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+    if "'" not in expression:
+        fail(
+            f"the {label} queryExpression is {expression!r} — a CEQL string literal has to be "
+            f"single-quoted (an unquoted RHS parses as subtraction server-side and 400s)"
+        )
+
+
+def read_references_input(node: dict[str, Any], shape: str, field: str) -> bool:
+    """True when the read's filter takes its value from the named flow input."""
+    if shape == NATIVE_READ:
```
rockymadden: **major :red_circle: the native branch cannot follow a Script hop into the filter.**

This matches `$vars....<field>` only inside the filter row's own value. `billing_invoice_lookup`'s prompt reads *"Normalize the input (trim, uppercase, prepend `MCS-` if absent), **then** look up"*. Two steps.

The connector reference inlines that normalization in its CEQL template, but a native filter row is a single `value` field, so an inline IIFE there is awkward and a Script node feeding `=js:$vars.normalize.output.value` is at least as natural. That build fails with:

```
FAIL: no invoice filter row takes its value from $vars....invoiceNumber
```

A false failure on the exact task this PR is meant to unblock.

Section 6 of `advisory_billing_invoice_lookup.py` already tolerates this shape, in its own words: *"a script hop is a legitimate authoring choice."* The machinery is in this module already, `source_depends_on` plus `node_dependencies`.

*Fix:* thread `nodes` into `assert_read_filters_input` and follow the hop the way section 6 does.

The connector branch has the same blind spot, so it is pre-existing. What is new is that this change makes it reachable on a shape where inlining is the *less* natural authoring.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+
+def entity_reads(nodes: Iterable[dict[str, Any]]) -> tuple[str, list[dict[str, Any]]]:
+    """Return ``(shape, reads)`` for the nodes that read entity records."""
+    nodes = list(nodes)
+    connector = [
+        node for node in nodes
+        if str(node.get("type") or "").startswith(CONNECTOR_READ_PREFIX)
+    ]
+    native = [node for node in nodes if str(node.get("type") or "") == NATIVE_READ_TYPE]
+    if connector and native:
```
rockymadden: **minor :yellow_circle: the mix check is asymmetric and can false-fail.**

The native side is `core.datafabric.read` only. The connector side is the whole `uipath.connector.uipath-uipath-dataservice.` prefix, every operation.

`data-fabric/planning.md:20-23` gives each native op its own tenant flag, so `read-entity` on with `create-entity` off is a real configuration. It produces a native read beside a connector `create-entity-record`, which trips:

```
FAIL: the flow mixes both entity-read shapes
```

That is the false-failure class this PR exists to remove.

Latent today, not live: all three billing scenarios are read-only (confirmed).

*Fix:* narrow the mix test's connector side to the read operations, keeping the full prefix for the count so `advisory_billing_invoice_lookup.py`'s *"the lookup is the query-entity-records operation"* diagnostic survives.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+    literal, and the native serializer refuses the WHOLE query on a blank
+    expression, a `$self` reference, or an operator it cannot emit, which leaves
+    the node with no output and breaks every downstream `$vars` reference.
+    """
+    if shape == NATIVE_READ:
+        rows = native_filter_rows(node)
+        if not rows:
+            fail(
+                f"the {label} read sets no `entityConfig._filters` rows — an unfiltered "
+                f"`resultMode: \"multiple\"` read returns the first 100 rows in arbitrary order, "
```
rockymadden: **minor :yellow_circle: the message asserts a `resultMode` nothing checked.**

It says an unfiltered read *"returns the first 100 rows in arbitrary order"*, which is the `resultMode: "multiple"` behaviour. On a `"single"` read the actual behaviour is an over-match fault (`data-fabric/impl.md:113`).

*Fix:* read `resultMode` and say the right thing, or drop the clause.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+    Native: nothing to resolve while the entity is tenant-scoped — the node
+    needs no Integration Service connection and no `bindings[]` connection row.
+    A `_folderKey` switches the emitted target to the folder-qualified form, and
+    without `_resourceKey`/`_entityKey` plus both `resource: "Entity"` rows the
+    name and folder serialize as source-org literals: it packages and breaks on
+    deploy (impl.md — Folder-scoped entities and bindings).
+    """
+    if shape == NATIVE_READ:
+        config = entity_config(node)
+        folder = str(unwrap(config.get("_folderKey")) or "").strip()
```
rockymadden, line 415: **minor :yellow_circle: unverified: empty `_folderKey`.**

`.strip()` treats `_folderKey: ""` as tenant-scoped. `data-fabric/impl.md:305` says the folder-qualified form is switched by the key's **presence**.

If the serializer tests presence rather than truthiness, an empty `_folderKey` emits a folder-qualified target with a blank folder and this passes it as tenant-scoped, which is the half-authored scope the same function exists to catch.

**I could not verify which**, it needs the serializer. Reporting rather than guessing.

tests/tasks/uipath-maestro-flow/_shared/check_bindings_no_stubs.py
```diff
@@ -65,25 +74,54 @@ def invalid_ids(entries: list[dict[str, Any]], schema: str) -> list[str]:
...
+    The two callers are Data Fabric scenarios, so "no connector node" means the
+    entity reads are native and there is nothing for a `connection` resource to
+    point at.
+    """
+    for path in _generated(cwd, "*.flow"):
+        try:
+            nodes = json.loads(path.read_text(encoding="utf-8")).get("nodes") or []
+        except (OSError, json.JSONDecodeError):
+            continue
+        if any(str(node.get("type") or "").startswith(CONNECTOR_READ_PREFIX) for node in nodes):
```
rockymadden: **minor :yellow_circle: the name overclaims and the match is too narrow.**

This tests only `CONNECTOR_READ_PREFIX`, so it answers "has a Data Service connector node", not "needs a connection". Wire this checker to any task whose flow carries a Slack or Jira node and the connection requirement silently disappears.

*Fix:* match `"uipath.connector."` generally. The native artifact carries no `uipath.connector.` node (confirmed), so this is strictly safer and still lets the native build through.

tests/tasks/uipath-maestro-flow/_shared/test_same_ground_advisories.py
```diff
@@ -44,6 +106,227 @@ def test_advisories_accept_repo_reference_flows(script: str, relative_flow: str)
...
+    assert result.returncode != 0
+    assert "mixes both entity-read shapes" in result.stdout + result.stderr
+
+
+def _project(tmp_path: Path, flow: dict, bindings: dict) -> Path:
+    """A generated-solution tree as the two billing checkers walk it: one flow
+    and one bindings file under the cwd they are run from."""
+    project = tmp_path / "proj"
+    project.mkdir(parents=True, exist_ok=True)
+    (project / "BillingInvoiceLookup.flow").write_text(json.dumps(flow))
```
rockymadden: **minor :yellow_circle: misleading fixture name.**

Every flow is written as `BillingInvoiceLookup.flow`, including the three cases whose content is the dispute-resolution artifact. Neither checker cares about the filename, so it works. It will mislead the next reader.

*Fix:* name it from the source path.

tests/tasks/uipath-maestro-flow/_shared/test_same_ground_advisories.py
```diff
@@ -44,6 +106,227 @@ def test_advisories_accept_repo_reference_flows(script: str, relative_flow: str)
...
+    (project / "bindings_v2.json").write_text(json.dumps(bindings))
+    return tmp_path
+
+
+CONNECTION_BINDINGS = {
```
rockymadden, line 534: **minor :yellow_circle: constant only used in its negative form.**

`CONNECTION_BINDINGS` is only ever mutated into a stub. No test asserts that a *valid* connection row on a native build passes, which is the case a native flow that also carries a connector node would hit.

*Fix:* add that case, or fold the constant into the stub test.

rockymadden:
Worked through all seven items (three inline, four suppressed). Six were valid; one I had already fixed in `d9ffb4c`. Fixed in `ff6855a`, and each fix has a test that was verified to fail when the fix is reverted.

**Accepted, with one adjustment to the proposed fix**

1. **The row's `field` was never checked.** Correct, and the same hole exists in the connector branch. Fixed, but not by requiring `row.field == field`: that conflates the flow input name with the entity column name. They coincide in all three billing scenarios and are not the same thing, so inferring one from the other would false-fail a scenario whose input and column are spelled differently, which is the over-strictness this PR exists to remove. The column is now an explicit parameter, and the comparison is case-insensitive because a wrong-CASE column is a different defect that the live gate catches. Tests: `test_native_filter_on_the_wrong_column_fails`, `test_native_filter_column_case_is_tolerated`.

2. **Entity binding rows matched on metadata alone, with no `default`.** Correct, and the sharpest catch here: the folder-scope test I wrote built rows with no `default` and asserted they pass, so it encoded the mistake directly. `default` is the value packaging substitutes, so a row without one is exactly the half-authored scope the check claims to catch. The fixture is now a helper that carries one, plus `test_native_entity_binding_without_a_default_fails`.

3. **`is any of` was rejected.** Correct. Legacy spelling of `in`, still read by the platform. Added, with `test_native_filter_legacy_in_operator_is_accepted`.

4. **`_folderKey` accepted any non-empty string.** Correct. It is the entity's `folderId`, so the folder-qualified target cannot resolve another shape. Now requires a real uuid. Flagging my confidence: that comes from the doc plus the eval tenant's folder keys, all uuids. If a non-uuid `folderId` form exists, this is too strict.

5. **The connector side returned the whole Data Service family, so a write counted as a read.** Correct, and it went further than the asymmetry I had already found: `advisory_billing_invoice_lookup`'s count would accept a `create-entity-record` node as its one lookup. Restricted to the read operations. The `query-entity-records` diagnostic still fires, because `get-entity-record-by-id` is a read and reaches it.

6. **`rows: [{}]` passed `check_server_side_filter`.** Correct. List non-emptiness is not a predicate. A row now has to name a column and carry a value.

7. **Parity guard between the two discriminators.** Correct that a comment asking to "keep the two in step" is not a guard. `test_entity_read_shapes_match_the_structural_gates` now ties this module's constants to `flow_check.ENTITY_QUERY_HINTS`.

**Already fixed before this review landed**

The mixed-shape asymmetry was found in my own review and fixed in `d9ffb4c`; item 5 above subsumes and improves it.

Suite: 1223 passed, from 1191 on `main`. All five advisories still pass their connector reference unchanged, and the native artifact still passes all five checks of the dispute-resolution advisory.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+        )
+
+
+def read_references_input(node: dict[str, Any], shape: str, field: str) -> bool:
+    """True when the read's filter takes its value from the named flow input."""
+    if shape == NATIVE_READ:
+        return any(
+            "$vars." in str(unwrap(row.get("value")) or "")
+            and references_field(row.get("value"), field)
+            for row in native_filter_rows(node)
```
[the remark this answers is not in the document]
rockymadden: Fixed in `ff6855a`, with one adjustment: not `row.field == field`, because that conflates the flow input name with the entity column name. They coincide in all three billing scenarios and are not the same thing, so inferring one from the other would false-fail a scenario spelling them differently, which is the over-strictness this PR exists to remove. The column is an explicit parameter now, compared case-insensitively because a wrong-CASE column is a different defect the live gate catches.

Guards: `test_native_filter_on_the_wrong_column_fails` and `test_native_filter_column_case_is_tolerated`.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+                f"worse than none — it packages and breaks on deploy. Keep the entity tenant-scoped, or "
+                f"let the canvas entity picker write all three"
+            )
+        attributes = {
+            str(binding.get("propertyAttribute") or "")
+            for binding in (flow.get("bindings") or [])
+            if isinstance(binding, dict)
+            and str(binding.get("resource") or "") == "Entity"
+            and str(binding.get("resourceKey") or "") == key
+        }
```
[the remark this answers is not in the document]
rockymadden: Fixed in `ff6855a`. The sharpest catch on this PR: the folder-scope test I wrote built rows with no `default` and asserted they pass, so it encoded the bug as correct behaviour. Each matched row now needs a nonblank `default`, the fixture is a helper that supplies one, and `test_native_entity_binding_without_a_default_fails` guards the omission.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
```diff
@@ -145,6 +145,250 @@ def query_references_input(detail: dict[str, Any], field: str) -> bool:
...
+    "!=",
+    ">",
+    ">=",
+    "<",
+    "<=",
+    "contains",
+    "starts with",
+    "ends with",
+    "in",
+}
```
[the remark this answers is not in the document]
rockymadden: Fixed in `ff6855a`. Added to `NATIVE_FILTER_OPERATORS` with the reason in a comment. Guard: `test_native_filter_legacy_in_operator_is_accepted`.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
rockymadden: Fixed in `d9ffb4c`. `read_references_input` follows the hop through a `transitive_refs` factored out of `source_depends_on`, with that function's behaviour unchanged.

Two guards, because following the hop must not degrade into accepting any `$vars` reference: `test_native_filter_reached_through_a_script_hop` and `test_native_filter_hop_to_nothing_still_fails`. Both verified to fail when the fix is reverted.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
rockymadden: Fixed in `d9ffb4c`, then superseded by `ff6855a`, which restricts the connector side of `entity_reads` to the read operations outright rather than only narrowing the mix test. Copilot found the part I had missed: the count in `advisory_billing_invoice_lookup` would otherwise accept a `create-entity-record` node as its one lookup.

Guard: `test_native_read_beside_a_connector_write_is_not_a_mix`.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
rockymadden: Fixed in `d9ffb4c`. The message reads `resultMode` and names the over-match fault on `single` rather than 100 arbitrary rows. Guard: `test_native_filter_no_rows_message_follows_result_mode`.

tests/tasks/uipath-maestro-flow/_shared/advisory_flow_utils.py
rockymadden: Addressed conservatively in `d9ffb4c` and `ff6855a`: folder scope is decided by the key's PRESENCE, a blank `_folderKey` is a finding, and the value has to be a real uuid. Guards: `test_native_blank_folder_key_fails` and `test_native_non_uuid_folder_key_fails`.

Resolving because the actionable half is done, but **the original unknown stands and no code change settles it**: I still have not confirmed whether the serializer tests presence or truthiness. Both fixes take the strict reading, so a flow is rejected rather than silently mis-emitted either way.

tests/tasks/uipath-maestro-flow/_shared/check_bindings_no_stubs.py
rockymadden: Fixed in `d9ffb4c`. Matched on the whole `uipath.connector.` family via a new `CONNECTOR_NODE_PREFIX`, which is strictly safer and still lets the native build through: the artifact carries no connector node.

tests/tasks/uipath-maestro-flow/_shared/test_same_ground_advisories.py
rockymadden: Fixed in `d9ffb4c`. `_project` takes the name, and a `_native_flow()` helper removes the repeated fixture load.

tests/tasks/uipath-maestro-flow/_shared/test_same_ground_advisories.py
rockymadden: Fixed in `d9ffb4c`. The constants are named for what they hold (`NON_CONNECTION_BINDINGS` and `CONNECTION_BINDINGS`), and `test_bindings_checker_accepts_a_valid_connection_pair` asserts the good pair positively.

PR #3257, opened 2026-09-11
fix(conversational-agents): improve output prompting for inline conversational agents

maxduu:
## Main issue — structured-output instructions leak into the chat
Per the documentation (https://dplt.cms.uipath-dev.com/docs-preview/pr-6634/maestro/automation-cloud/latest/user-guide/node-conversational-agent#output-variables), a conversational agent generates two things from different inputs:

| Generated | Driven by |
|---|---|
| the chat reply | the system prompt **only** |
| the structured outputs | the system prompt **plus** each field's `description` |

So a system prompt saying *"set `route` to `end_conversation` when the user says goodbye"* makes the agent emit the value inline, and the end user might see something like `<uip:route>end_conversation</uip:route>` or tries to find a tool that sets the structured output.

Added the division-of-labor rule (what to say → system prompt; how to fill the field → that field's `description`), a worked example, and the literal tag so it's greppable by anyone who hits it.

## Secondary — the `outputSchema` ban read as universal

`uipath-agents` said, unconditionally: **"Do not add properties to the `outputSchema` of a conversational agent."** True for standalone, in-solution and published agents; wrong for an inline `uipath.agent.conversational` node, where those fields are the routing signal. An agent reading it concludes a valid Flow is misconfigured and "fixes" it by emptying the schema, removing the flow's only branching input.

Scoped the rule to the flavors it applies to and named the exception, in each place it was stated. Also removed a stale claim in `lowcode.md` that inline conversational agents aren't supported at all — that's the upstream cause, since an agent reading it picks autonomous and never reaches the `outputSchema` rule.


skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -217,6 +217,24 @@ Declare each field in two places or it yields nothing at run time:
...
+| the structured outputs | the system prompt **plus** each field's `description` |
+
+So split the instructions by destination — *what to say* and *response instructions* in the system prompt, *how to fill the field* in that field's `description` (write the same description in both `agentOutputVariables[]` and `outputSchema.properties`):
+
+| Where | Example |
+| --- | --- |
+| system prompt | "Thank the user when they would like to end the conversation." |
+| `endConversation` (boolean) `description` | "Set to true when the user intends to end the conversation." |
+
+**Never name the output field, its values, or the flow's routing in the system prompt.** Because the chat reply is generated from the system prompt alone, an instruction like "set `endConversation` to `true` when the user says goodbye" may make the LLM emit the structured value to the chat or look for a tool to set the output variables.
```
maxduu: main fix in pr

andrewwan-uipath (approved):
lgtm, thanks!

skills/uipath-agents/references/lowcode/critical-rules/conversational-critical-rules.md
```diff
@@ -8,7 +8,9 @@ These rules are the canonical source for rules specific to low-code conversation
 
 ## What NOT to Do
 
-1. **Do not add properties to the `outputSchema` of a conversational agent.** After initialization, leave `outputSchema` empty. The conversational agent runtime streams responses/tool-call events during the execution, so the final output is not relevant for the end-user in the conversation.
+1. **Do not add properties to the `outputSchema` of a conversational agent built here.** After initialization, leave `outputSchema` empty. Standalone, in-solution and published conversational agents do not support structured outputs — the runtime streams responses/tool-call events during the execution, so the final output is not relevant for the end-user in the conversation, and a populated schema is never filled.
```
rockymadden, line 11: **A** · `major`

Two problems on this line and the paragraph under it.

- **"built here" + "authored under the `uipath-maestro-flow` skill" reads as *not your job*.** Critical rule 15 of this skill: *"Never invoke other skills automatically."* An agent asked to add an output field to an inline chat agent's `agent.json` now reads an exception it is told belongs elsewhere and stops, instead of making the edit. That's the same failure as before, one step softer. `outputSchema` **is** `agent.json` — this skill's file. Only the `.flow` node's `agentOutputVariables[]` belongs to the flow skill. Grant the edit:
  > **The exception is an inline-in-flow conversational agent** (`uip agent init --inline-in-flow --conversational`, wired to a `uipath.agent.conversational` node). Populate `outputSchema.properties` to match the node's `agentOutputVariables[]`; the `.flow` node itself is authored by the `uipath-maestro-flow` skill.

- `minor` — **"Standalone, in-solution and published" mixes two taxonomies.** This skill's axis is binary: standalone vs `--inline-in-flow` (no file under `references/lowcode/` uses "in-solution" or "published" as an agent flavor). Those are the flow skill's words for where an agent sits relative to a flow, and both are "standalone" here. Two undefined terms where the reader needs one bit. Say "standalone (non-inline)".

Same sentence repeats at `agent-definition.md:238` and `prompting/conversational-agent-prompting-guide.md:106` — fix all three together.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -217,6 +217,24 @@ Declare each field in two places or it yields nothing at run time:
 
 Bind it downstream as `$vars.<agentNodeId>.output.shouldHandoff`. Writing one side without the other passes `agent validate` and `flow validate` — nothing checks the pair.
 
+### Prompting: what goes where
+
+The prompts that drive the agent's chat replies and structured output differ:
+
+| Generated | Driven by |
+| --- | --- |
+| the chat reply | the system prompt **only** |
```
rockymadden, line 226: **B** · `major`

This row, and `:236` — *"Because the chat reply is generated from the system prompt alone…"*

- **The stated cause isn't the cause.** "Alone" doesn't make a value leak into chat — *inclusion* does. The mechanism is that the system prompt is in context for the reply pass and the field `description` is not. Say that, and the rule follows.

- **Read literally, the sentence is false and expensive.** Conversation history, the user turn, and tool results all drive the reply. This repo treats dropped history as the #1 silent failure on this node — `conversational_chat_loop.yaml` weights it 3.0 with the note *"nothing else here would catch an agent that ships without chat history."* A page telling the agent the reply comes from the system prompt alone cuts against that check.

Suggested: `| the chat reply | the system prompt (plus conversation history and tool results) — **not** the field descriptions |`, then in the prose: *"the field `description` is not in context when the reply is generated; the system prompt is in context for both."*

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -217,6 +217,24 @@ Declare each field in two places or it yields nothing at run time:
...
+| the structured outputs | the system prompt **plus** each field's `description` |
+
+So split the instructions by destination — *what to say* and *response instructions* in the system prompt, *how to fill the field* in that field's `description` (write the same description in both `agentOutputVariables[]` and `outputSchema.properties`):
```
rockymadden, line 229: **D** · `minor`

> (write the same description in both `agentOutputVariables[]` and `outputSchema.properties`)

`:218` already says nothing checks the pair. This adds a hand-duplicated string across two files with no statement of which side the runtime reads, so drift is undebuggable — and I could not verify which one wins. Name the authoritative side, or say it's unknown and pick one as canonical.

Also: `:215`–`:218` uses `shouldHandoff`, `:234` uses `endConversation`. One field name across the section costs nothing and stops the reader checking whether they're two mechanisms.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
```diff
@@ -217,6 +217,24 @@ Declare each field in two places or it yields nothing at run time:
...
+
+| Where | Example |
+| --- | --- |
+| system prompt | "Thank the user when they would like to end the conversation." |
+| `endConversation` (boolean) `description` | "Set to true when the user intends to end the conversation." |
+
+**Do not name the output field and how to set it in the system prompt.** Because the chat reply is generated from the system prompt alone, an instruction like "set `endConversation` to `true` when the user says goodbye" may make the LLM emit the structured value to the chat or look for a tool to set the output variables.
```
rockymadden: **C** · `minor`

> **Never name the output field, its values, or the flow's routing in the system prompt.**

The example directly above violates the third clause: *"Thank the user when they would like to end the conversation"* **is** the routing condition, in prose. An agent applying the rule strictly rejects the page's own example. Cut the widest clause — the field name, its values, and any tool that sets it are the actual leak surface:

> Never name the output field, its values, or a tool that sets them, in the system prompt.

skills/uipath-agents/references/lowcode/lowcode.md
```diff
@@ -22,14 +22,13 @@ UiPath low-code agents come in **two variants**, both sharing this skill.
    
 2. Conversational agents are intended for use-cases involving multi-turn conversations / fast latency with real-time user-interaction / streamed responses.
 
-> Currently, conversational agents are only standalone and not yet supported to be inline in a flow, so inline-in-flow use-cases should always use an autonomous agent.
 > After deployment, conversational agents are interacted with through the UiPath Conversation Service, which manages conversation history and exposes a CLI/SDK for client UIs. Each user-initiated exchange invokes the agent for a single turn, streaming events back to the client.
+> A standalone conversational agent is the simplest way to build a back-and-forth chat assistant driven by a single system prompt. For chats that need more — routing between several agents, deterministic replies, or behind-the-scenes automations running while the conversation continues — direct the user to the `uipath-maestro-flow` skill.
```
rockymadden, line 26: **E** · `minor`

- **The table below lost its only inline row and gained no replacement.** With `| User wants inline-in-flow | Autonomous |` gone, nothing in this file routes "multi-turn chat, inline in a flow". The reader lands on **Conversational** (`:33`) and then hits `project-lifecycle.md:33` saying that's not a feature (overall finding 1). Add the row:
  > `| User wants a conversational agent inline in a flow | **Conversational**, scaffold with `--inline-in-flow --conversational` |`

- **First half of this line is narration.** *"A standalone conversational agent is the simplest way to build a back-and-forth chat assistant driven by a single system prompt"* changes nothing the agent does. The second half routes. Keep the second half.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
```diff
@@ -129,5 +129,5 @@ In the architectural plan:
...
 - `chat-send-message: <purpose>` — one line per flow-authored message (loading messages, handoff notice, node output results); a scripted chat consists mostly of these
-- `chat-structured-output: <agent-name> = <fieldName>` — only when the flow branches on that conversational agent's reply; forces that conversational agent to inline flavor because in-solution and published conversational agents have no structured outputs
+- `chat-structured-output: <agent-name> = <fieldName>` — only when the flow branches on that conversational agent's reply; forces that conversational agent to inline flavor because in-solution and published conversational agents have no structured outputs. Note that instructions for filling each output field go in the field's `description`, never in the system prompt (see [impl.md § Prompting: what goes where](impl.md#prompting-what-goes-where))
```
rockymadden: **F** · `minor`

The appended *"Note that instructions for filling each output field go in the field's `description`, never in the system prompt"* restates an impl rule inside a planning-annotation bullet. Planning writes annotations, not prompts, so this changes nothing at the point it's read, and `impl.md` already owns it — the reader reaches it via the flavor table. Drop it, or make it annotation-shaped so it changes the output:

> `chat-structured-output: <agent-name> = <fieldName> — <how the agent decides the value>`

rockymadden (changes requested):
## What it does

6 files, +27/-8, docs only.

- **`uipath-maestro-flow`** — new `impl.md § Prompting: what goes where`: chat reply comes from the system prompt, structured outputs from system prompt + each field's `description`. Adds a worked example and a ban on naming the output field in the system prompt.
- **`uipath-agents`** — scopes the universal "never populate `outputSchema` on a conversational agent" ban, naming the inline-in-flow exception in 3 places; drops the `lowcode.md` claim that inline conversational agents don't exist.

Both failure modes it targets are real. I confirmed the second one end to end:

```
uip agent init . --inline-in-flow --conversational   -> Success, engine: conversational-v1, isConversational: true
outputSchema.properties = { endConversation }        -> refresh + validate --inline-in-flow: Valid
```

## 🔴 Overall findings

**1. `major` — the stale claim survives in the file that matters.** The description says it removed the claim that inline conversational agents aren't supported. It removed one copy. `skills/uipath-agents/references/lowcode/project-lifecycle.md:33` still reads:

> `--inline-in-flow` — … **Only applicable for autonomous agents, since adding inline conversational-agents within a flow project is currently not an enabled feature.**

That's the `uip agent init` flag reference — the page an agent reads *while running the command* `uipath-maestro-flow` told it to run. Worded harder than the line that was deleted, and factually wrong per the scaffold above. Same page, `:48`, then says "add a `uipath.agent.autonomous` node" for every inline agent; `capabilities/inline-in-flow/inline-in-flow.md:175` repeats it. Fix `:33` in this PR at minimum — without it the stated upstream cause is untouched.

**2. `minor` — description ≠ diff.** The description says: *"Added … the literal tag so it's greppable by anyone who hits it."* `grep -rn "uip:route" skills/` returns nothing. The symptom string is the most useful half for whoever lands on this page after a user reports weird tags in chat. Add it, or drop the claim.

**3. `minor` — no test, and one exists to extend.** `tests/tasks/uipath-maestro-flow/conversational/conversational_chat_loop.yaml` already builds this exact topology and asserts nothing about structured outputs. The pairing rule the page calls unchecked (`agentOutputVariables[]` + `outputSchema.properties` — "nothing checks the pair") is deterministically checkable; one `json_check` criterion covers the newly-legal case. The `uipath-agents` side is safe: `lowcode/conversational/restrictions.yaml:72` still asserts `length(outputSchema.properties) == 0`, and it passes only because that task scaffolds a **standalone** agent.

**4. `minor` — reasoned, not confirmed.** I could not verify the load-bearing runtime claim (reply ← system prompt; structured output ← system prompt + `description`). Its only source is `dplt.cms.uipath-dev.com/docs-preview/pr-6634/…`, which dies when that docs PR merges. Everything downstream of it in this diff rests on one unverifiable link.

## Recommendation

**Request changes.** Finding 1 plus inline comment A mean both failure modes described can still occur after this merges — one from the untouched flag reference, one from wording that hands the exception to another skill while critical rule 15 forbids calling it. Neither is a rewrite.

No AI slop in the new `impl.md` section — it's dense and prescriptive. One narration line in `lowcode.md`, flagged inline.

---

**tl;dr** — Right diagnosis, both fixes land short. The stale "inline conversational isn't supported" text still sits in `project-lifecycle.md:33`, the flag reference an agent reads mid-scaffold, so the bug it blames survives. The new exception is worded as another skill's job while rule 15 forbids calling that skill, so an agent may still refuse to populate `outputSchema`. The new prompting section's causal claim is wrong and, read literally, tells the agent the reply needs no chat history — which this repo's own test weights 3.0. Six inline comments, no rewrite needed.

rockymadden (approved):
## ✅ Approve — supersedes my earlier request-changes

Net improvement, and the diagnosis is right. Flipping to approve; the two `major` findings are guidance, not blockers. Fix them in this PR if it's cheap, follow-up if not.

**Fix before merge, ideally:**

1. `major` · `skills/uipath-agents/references/lowcode/project-lifecycle.md:33` — the claim this PR set out to kill still sits in the `uip agent init` flag reference, worded harder than the line that was deleted: *"Only applicable for autonomous agents, since adding inline conversational-agents within a flow project is currently not an enabled feature."* Verified wrong — `uip agent init . --inline-in-flow --conversational` scaffolds cleanly and a populated `outputSchema` passes `validate --inline-in-flow`. One line, and the stated upstream cause is actually gone.

2. `major` · [comment A](https://github.com/UiPath/skills/pull/3257#discussion_r3992149528) — the exception reads as another skill's job while critical rule 15 forbids calling that skill, so an agent may still decline to populate `outputSchema`. Grant the edit explicitly.

3. `major` · [comment B](https://github.com/UiPath/skills/pull/3257#discussion_r3992149536) — "the chat reply is generated from the system prompt alone" is both the wrong causal claim and, read literally, tells the agent the reply needs no chat history.

The four `minor` comments (C–F) and the missing test criterion are yours to take or leave.

**tl;dr** — Approved. Three majors worth fixing on the way in: the stale line in `project-lifecycle.md:33`, the ownership wording on the exception, and the "system prompt alone" claim. None blocks merge.

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/planning.md
maxduu: Removed

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
maxduu: Fixed

skills/uipath-maestro-flow/references/author/plugins/conversational-agent/impl.md
maxduu: Fixed

skills/uipath-agents/references/lowcode/critical-rules/conversational-critical-rules.md
maxduu: For now, prefer to direct to maestro-flow which has much more e2e info for inline convo agents.

andreibalas-uipath (approved):

PR #3258, opened 2026-09-11
fix(tests): retry case debug once for a tenant-side fault, never for a plan fault

roberts-cliff:
## Why

Two case graders died in the 2026-09-11 nightlies on faults that belong to the debug service, not to the plan:

| task | harness | fault |
|---|---|---|
| `single-api-workflow` | terra | `Failed during poll-instance-status: 403 Forbidden on GET /api/v1/debug-instances/…/element-executions` |
| `multi-linear-three-stages` | Sonnet | `finalStatus: Cancelled` with the stage still `InProgress` — and a `504 Gateway Timeout` on the same poll in CI run 34510099349 |

Re-ran both tasks on both harnesses this afternoon: **4 of 4 passed at 1.0**, no 403, no 504, no cancellation. Neither fault survives a retry and neither was measuring the build. `start_debug` fails on the first non-zero exit, so one afternoon of tenant weather costs two tasks a night on a suite we report on.

## What

`debug_fault_is_transient()` names exactly two shapes and nothing else:

- an HTTP status from the debug service's own phase report — 403, 408, 409, 425, 429, 5xx
- a cancelled instance (`"finalStatus": "Cancelled"`)

`start_debug` retries once on those, after a 15 s pause. `run_debug` gets the same single retry for a `Cancelled` status that arrives through a clean exit.

**Everything else fails on the first attempt** — a 400, a 404, a validation error, an unresolved resource, a `Faulted` case. Retrying a plan defect into a pass is worse than the flake it would hide.

**A retry is never silent.** It prints what it is retrying and why, and a second failure carries `after one retry` in the message, so a genuine platform regression still turns the suite red with a legible reason instead of being absorbed.

## Test

15 unit tests in `_shared/test_case_check_debug_retry.py`, parameterised against the verbatim nightly output, covering both directions — the transient set retried, the plan-fault set not — plus call-count assertions that a plan fault invokes `debug` exactly once.

```
pytest tests/tasks/uipath-maestro-case/_shared/test_case_check_debug_retry.py   15 passed
pytest tests/tasks/uipath-maestro-case/                                        153 passed   # the CI job's own command
```

Mutation-checked: disabling the transient classifier turns 6 of the 15 red; making the retry unconditional turns 2 red.

## Scope

`tests/tasks/uipath-maestro-case/_shared/` only — this team's path in `CODEOWNERS`. It changes grading semantics for every case task that calls `run_debug` / `start_debug`, which is the reason the transient set is a closed list of two observed shapes rather than a blanket retry.

### DIFF
--- a/tests/tasks/uipath-maestro-case/_shared/case_check.py
+++ b/tests/tasks/uipath-maestro-case/_shared/case_check.py
@@ -22,6 +22,7 @@ import os
 import re
 import subprocess
 import sys
+import time
 from typing import Any, Iterable, NoReturn, Sequence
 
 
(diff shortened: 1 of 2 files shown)


Yiqi517 (approved):
Oh amazing this kills me

StephenHanzlik (approved):

PR #3259, opened 2026-09-11
fix(docs): HITL outcome ports were documented as a single completed handle

dushyant-uipath:
## Problem
QuickForm HITL nodes wire one `outcome-<outcome.id>` port per outcome. The docs taught a single `completed` handle regardless of outcome count, so an agent following them left every non-default outcome unwired.

## Fixed
- 8 uipath-maestro-flow references taught `completed` unconditionally: corrected to per-outcome `outcome-<id>` ports.
- Sibling uipath-human-in-the-loop skill contradicted the fix (SKILL.md, hitl-node-quickform.md, hitl-node-apptask.md): corrected.
- A coded-action-app example flipped onto outcome-derived ports by gaining an `inputs.schema` block it shouldn't have had: the block is removed, port stays the static `completed` for that node type.
- impl.md and planning.md claimed `outcome-completed` never reappears once outcomes exist: corrected. It is the zero-outcome placeholder, or the port for a real outcome whose own `id` happens to be `completed`.

Docs only, no executable surface changed.

## Split from #3180
Split out of #3180 into three, so documentation wording, checker logic, and task fixtures each get their own reviewable diff: docs here, checker logic in #3260, task fixture migration in #3261 (depends on #3260). #3180 is closed in favor of these three.


dushyant-uipath:
Closing. Reconsolidated back into #3180 as a single PR.

PR #3260, opened 2026-09-11
fix(tests): outcome-id is verbatim, not lowercased with a name fallback

dushyant-uipath:
## Problem
`check_simulated_hitl.py` and `check_devcon_expense_approval.py` derived the `outcome-<id>` wiring port from a lowercased `outcome.name`, falling back to that when `id` was missing. flow-workbench's `build-handle-customization.ts` derives the port from `outcome.id` verbatim and renders no handle at all when `id` is missing or empty. The old fallback passed flows that would actually leave an outcome dangling at runtime.

## Fixed
- Both checkers now read `id` verbatim and fail loudly when it is missing or non-string, instead of guessing from `name`.
- Extracted the shared outcome-wiring assertion into `check_simulated_hitl.py`'s `assert_outcome_wiring()`, used by its existing `expense` check and a new generic `outcome-wiring` check.
- Per-outcome checks are gated on `outcomes` being non-empty, so the boolean-decision-only path used by `expense_approval_simulated` still passes.
- The outcome-completed-misuse check now runs before the missing-ports check in both files, so a flow that only wires `outcome-completed` gets that specific message instead of a generic "missing" one.

## Test plan
- [x] Added regression tests for a mixed-case outcome id and an outcome with no id.
- [x] `python3 -m pytest tests/tasks/uipath-maestro-flow/e2e/test_check_devcon_expense_approval.py tests/tasks/uipath-maestro-flow/_shared/test_batch2_alignment.py`: 19/19 passing locally.

## Split from #3180
Split out of #3180. #3259 carries the doc wording fix, independently reviewable. #3261 migrates task fixtures to the new `outcome-wiring` check and is based on this branch.

## Known overlap, needs reconciling before merge
#3239 (different author) touches the same two files here (`check_devcon_expense_approval.py`, `test_check_devcon_expense_approval.py`) and claims the default v1.0 HITL node only ever accepts a `completed` edge, with outcome-id ports gated behind an opt-in `outcomePorts` flag on v1.1+. A local coder-eval run authored a `typeVersion: "1.0"` node wired with `outcome-approve`/`outcome-reject` edges and `uip maestro flow validate` passed clean, contradicting that. Needs reconciling before either PR merges.

### DIFF
--- a/tests/tasks/uipath-human-in-the-loop/TEST_PLAN.md
+++ b/tests/tasks/uipath-human-in-the-loop/TEST_PLAN.md
@@ -36,7 +36,7 @@ All tests target `uipath.human-in-the-loop` v1.0 (the current manifest). Key inv
 | Output access key | field `id` property | Using field `variable` property |
 | Status variable | `$vars.<nodeId>.status` | Expecting `"completed"` string |
 | Status value | outcome `action` value (`"Continue"` / `"End"`) | Comparing to `"completed"` |
-| Available handles | `completed` only | Wiring `cancelled` or `timeout` (removed in v1.0) |
+| Available handles | one `outcome-<outcome.id>` per outcome (QuickForm); static `completed` (App-based) | Wiring `cancelled` or `timeout` (removed in v1.0); wiring the `outcome-completed`/`completed` placeholder on a QuickForm node that has real outcomes |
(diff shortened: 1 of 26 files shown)


dushyant-uipath:
Closing. Reconsolidated back into #3180 as a single PR.

PR #3263, opened 2026-09-11
fix(uipath-maestro-case): state SLA unit semantics — `m` is months

charlesliu9:
fix(uipath-maestro-case): state SLA unit semantics — `m` is months

The expense-runnable e2e failed in `validate --strict --sdd`, gained STRICT_SDD_SLA_DURATION_MISMATCH and
STRICT_SDD_VARIABLE_DEFAULT_MISMATCH 

The SLA plugin listed the unit vocabulary `min|h|d|w|m` with no semantics; only case-schema.md said `m` means months. An agent reading just plugins/sla/ takes `m` for minutes, collides with the documented 15-minute floor, and "repairs" the conflict: the failing run clamped all eight SLA rules to 15 min. The new check compares only the case-level rule, so the seven equally-wrong stage SLAs went unreported.

Changes:
- plugins/sla/planning.md: spell out min/h/d/w/m, scope the 15–1000 floor to  `unit: "min"`, add an anti-pattern against normalizing the SDD unit
- plugins/sla/impl-json.md: same semantics beside the `"<min|h|d|w|m>"` placeholder
- both fixtures meant minutes but wrote `m` — normalize to `min` (expense scaled x15  to clear the floor, aged-invoice token-only) and drop the `(empty string)` gloss  that the new default check parsed as the value instead of the code-span

Verified: sdd_check.py OK on aged-invoice, unchanged on expense (52 pre-existing
template findings before and after); 99 pytest pass; npm run skills:validate OK; a
corrected caseplan/SDD pair validates clean. 

Rerun tests:
https://github.com/UiPath/skills/actions/runs/34640710432(claude)
https://github.com/UiPath/skills/actions/runs/34640684698(codex)
both succeeded

### DIFF
--- a/skills/uipath-maestro-case/references/plugins/sla/impl-json.md
+++ b/skills/uipath-maestro-case/references/plugins/sla/impl-json.md
@@ -58,6 +58,8 @@ After grouping elements by target, compose the `slaRules` array and write it int
 ]
 ```
 
+> **`unit` is the sdd.md token verbatim**: `min` (minutes), `h` (hours), `d` (days), `w` (weeks), `m` (**months**). Never map `m` onto `min`, and never rewrite a unit or clamp a count to clear the 15–1000 floor — that floor applies to `min` only.
+
 ### Root-target shape
 
 ```json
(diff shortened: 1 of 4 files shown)


Yiqi517 (approved):
Bonus!

PR #3264, opened 2026-09-11
fix(case): point the trigger event-param gate at the slot the CLI fills

Yiqi517:
## The problem

Two instructions for building a connector-bound condition rule point at the wrong place.

**The hard gate names a slot `case spec` never returns.** Step 3 of both trigger plugins reads `caseShape.inputs[name="eventParameters"].body` and blocks the build until every required event parameter has a value there. `buildTriggerCaseShape` emits one input, named `body` ([`connector-task-spec.ts`](https://github.com/UiPath/cli/blob/8c01becbb798eed889c42d4b901c670e90905ef7/packages/case-tool/src/services/connector-task-spec.ts#L1058-L1080)), and `applyTriggerInputDetails` puts the resolved values one level inside it, at `body.queryParams` ([same file](https://github.com/UiPath/cli/blob/8c01becbb798eed889c42d4b901c670e90905ef7/packages/case-tool/src/services/connector-task-spec.ts#L1272)). `inputs.eventParameters` is the normalized contract array ([`connector-task-spec.types.ts`](https://github.com/UiPath/cli/blob/8c01becbb798eed889c42d4b901c670e90905ef7/packages/case-tool/src/services/connector-task-spec.types.ts#L215)), a sibling of `caseShape`, not a member of it. Reading one off the other conflated them.

**The rule's JSON shape block names no fields.** Its `inputs` line was a single placeholder string that named no field, so an agent filling it copies the wait-for-connector task's input entry whole, `id`, `var` and `elementId` included. The rule's input then claims the task's `elementId`, and that field is the scope key the canvas filters variables by: `variableIdForRuleInElement` returns `` `${element.id}-${rule.id}` `` ([`CaseManagementVariableServiceUtils.tsx`](https://github.com/UiPath/PO.Frontend/blob/655db12a7605cafcd716406366f0612d76d20d06/src/services/uipath/variables/service/utils/CaseManagementVariableServiceUtils.tsx#L93)) and is stamped onto the rule's inputs and outputs ([`UnifiedBuildHookCaseMgmtUtils.ts`](https://github.com/UiPath/PO.Frontend/blob/655db12a7605cafcd716406366f0612d76d20d06/src/utils/unified-build-hooks/UnifiedBuildHookCaseMgmtUtils.ts#L927)).

## What it costs

Two `wait-for-connector` tasks in the local artifact corpus carry an input named `eventParameters` with no `target`, holding the event parameter that belongs at `body.queryParams`. The canvas cannot see it: its read path keys on `target === "body"` ([`IntsvcActivityPropertiesUtils.tsx`](https://github.com/UiPath/PO.Frontend/blob/c6fd27dbf00eefeece8d15be9b3d5eb16cdc1ff7/src/utils/IntsvcActivityPropertiesUtils.tsx#L287)), so the trigger opens unconfigured.

Both runs that wrote it lost the task's primary criterion. `skill-case-connector-wait-trigger` has ten local runs: two wrote the invented slot and both scored FAILURE at 0.647, with `check_connector_wait.py` naming that very input (`FAIL: entry rule inputs['eventParameters'].elementId must be 'Stage_CWC001-rCwcProc01'; got 'Stage_CWC001-cCwcProc01'`). The eight that did not write it split six passing and two failing on unrelated assertions.

## What changed

| File | What it said | What it says |
|---|---|---|
| `plugins/tasks/connector-trigger/impl-json.md` | the gate scans `caseShape.inputs[name="eventParameters"].body` | it scans `caseShape.inputs[name="body"].body.queryParams` |
| `plugins/triggers/event/impl-json.md` | same gate, plus the placeholder-fallback row that names it | same slot, all three references |
| `case-spec-input-details.md` | the contract array is `caseShape.inputs.eventParameters[?required]` | it is `inputs.eventParameters[?required]`, a sibling of `caseShape` |
| `connector-trigger-impl.md`, `plugins/conditions/task-entry-conditions/impl-json.md` | the rule shape's `inputs` was one placeholder string | the entry is spelled out, `elementId` built from the rule's id |

Step 1 of each gate already read `inputs.eventParameters[?required]`, and that stays.

## Verification

**Both agents build both halves correctly.** One `run-coder-eval` round each on `skill-case-connector-wait-trigger`, judged on what this PR changes:

| | codex | claude |
|---|---|---|
| rule input slot (`name` and `target` are `body`, value under `body.queryParams`) | correct | correct |
| rule input `elementId` | `Stage_D4e5F6-rP5q6R7s8` | `Stage_bIM0QL-rWqnzlmiH` |
| task input slot and `elementId` | correct | correct |

Each `elementId` equals that run's own stage id joined to its own rule id. The claude round passed the task outright at 1.0. The codex round failed on its rule `context`, whose `connectorKey` was left at the Phase 2 literal `placeholder` while every other entry resolved.

**The slot fix flips the failing artifacts.** Rewriting the invented input into that slot on a copy of each, with nothing else touched, takes `check_connector_wait.py` from exit 1 to exit 0 and `uip maestro case validate` from one error to zero findings, on both.

**The elementId fix is a shape change, not another sentence.** The correct value was already stated five times in prose across the shared doc, and the transcript shows the agent read it: `ownerNodeId` 14 times, `connector-bound condition rule` 35, `Trigger-NODE inputs only` 3. What it copied from was the shape block, which named no field.

**The slot half has a CLI backstop.** `CASE_MGMT_CONNECTOR_INPUT_FLATTENED` fires at error severity on the default profile, not only under `--strict`, on both failing artifacts.

**The elementId half has none, and the corpus cannot support one yet.** A rule would clear the `--strict` admission bar: run 34663390528 wrote `rule.uipath.inputs[0].elementId` as `Stage_Await1-tWaitReply1` where the rule's id gives `Stage_Await1-rProcess01`, `full` and `--strict` both reported nothing, and the task's own checker exited 1. What is missing is a correct artifact to hold the rule against. The corpus has five `wait-for-connector` rules across 464 caseplans, two of them carrying an input, and both of those carry a wrong `elementId`: `Stage_M3n4O5-rStage5x` under rule `Rule_Stage5`, and `Stage_CWC001-cCwcProc01` under rule `rCwcProc01`. With no passing instance, a switch comparison measures only that the rule fires, never that it stays quiet where it should. The two runs above are the first correct instances, so the rule belongs after this change has produced a population to test against.

**Regression.** `smoke-skills.yml` on this branch and on a control branch off main, six `uipath-maestro-case` smoke tasks, identical per task: all SUCCESS at 1.0.

`npm run skills:validate`: `default: 27 skills, 1768 files, 0 replacements; studioweb: 27 skills, 1768 files, 264 replacements`. `scripts/check-skill-status.py` and `scripts/check-skills-sh.py` both report OK. No occurrence of the old spelling is left.

### DIFF
--- a/skills/uipath-maestro-case/references/case-spec-input-details.md
+++ b/skills/uipath-maestro-case/references/case-spec-input-details.md
@@ -144,7 +144,7 @@ When `filter` is also provided, the mandatory clause and user filter are AND-joi
 
 The same combined expression also lands at `caseShape.context[name="metadata"].body.activityPropertyConfiguration.filterExpression` so the design-time projection and runtime body don't drift.
 
-For Outlook 365 `EMAIL_RECEIVED`, `parentFolderId` is required. The connector contract is in `caseShape.inputs.eventParameters[?required]` — consult that array to know which event params your trigger needs.
+For Outlook 365 `EMAIL_RECEIVED`, `parentFolderId` is required. The connector contract is in `inputs.eventParameters[?required]`, a sibling of `caseShape` in the spec response — consult that array to know which event params your trigger needs.
 
 > **Note:** `body.parameters` is not emitted for triggers. The FE reserves that key for queryParameters/pathParameters, neither of which exists in the trigger input contract.
 
--- a/skills/uipath-maestro-case/references/plugins/sla/impl-json.md
+++ b/skills/uipath-maestro-case/references/plugins/sla/impl-json.md
@@ -58,6 +58,8 @@ After grouping elements by target, compose the `slaRules` array and write it int
 ]
 ```
 
+> **`unit` is the sdd.md token verbatim**: `min` (minutes), `h` (hours), `d` (days), `w` (weeks), `m` (**months**). Never map `m` onto `min`, and never rewrite a unit or clamp a count to clear the 15–1000 floor — that floor applies to `min` only.
+
 ### Root-target shape
 
 ```json
--- a/skills/uipath-maestro-case/references/plugins/sla/planning.md
+++ b/skills/uipath-maestro-case/references/plugins/sla/planning.md
@@ -39,7 +39,7 @@ Set root SLA first, then stage SLAs. This mirrors the schema precedence: stage >
 | Field | Source | Notes |
 |-------|--------|-------|
 | `count` | sdd.md duration number | Positive integer |
-| `unit` | sdd.md duration unit | `min` \| `h` \| `d` \| `w` \| `m` |
+| `unit` | sdd.md duration unit | `min` (minutes) \| `h` (hours) \| `d` (days) \| `w` (weeks) \| `m` (months). Carry the sdd.md token verbatim — `m` is **months**, never minutes. |
 | `target` | sdd.md target (root vs stage) | `"root"` or `"<stage-name>"` |
 | `display-name` | sdd.md `SLA Title` (§1 Case Metadata for root; `**SLA Title:**` under `#### Stage SLA`) or generated fallback | Required non-empty SLA title, unique within the target, and MUST NOT contain `:`. Carry the SDD title verbatim. If the SDD has no title, ask for one or use the deterministic fallback `SLA Rule {N}` and record it. |
 | `rationale` | sdd.md case/stage SLA Design Rationale | Required reviewer context for the target, duration, threshold, and escalation behavior. |
@@ -196,7 +196,7 @@ Before emitting SLA elements, reject or repair the same cases the Case App rejec
 
 - every SLA rule has a non-empty, target-unique `display-name`;
 - every escalation has a non-empty, target-unique `display-name`;
-- every SLA `count` is positive, and minute-based values are between 15 and 1000 inclusive;
+- every SLA `count` is positive, and `unit: "min"` counts are between 15 and 1000 inclusive. This floor applies ONLY to `min`; `h` / `d` / `w` / `m` counts carry no floor, so never rewrite a unit or clamp a count to satisfy it;
 - every non-default rule has a condition/expression;
 - every escalation has at least one recipient, and every `at-risk` escalation carries `at-risk-percentage`.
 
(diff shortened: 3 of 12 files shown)


Yiqi517:
@claude please review the current head.

Yiqi517:
On the one item left for manual review:

> **Domain accuracy of the corrected slot path.** The review confirms internal consistency (all references agree, planning vs implementation steps correctly point at different objects), but whether `caseShape.inputs[name="body"].body.queryParams` matches the actual CLI output is a domain claim backed by the PR author's source links to `connector-task-spec.ts`. The two verification runs in the PR description provide evidence.

Settled, on three reads rather than on the runs. Both writers put the values at that path, and the two paths the PR keeps apart are two different objects in the same response.

**The CLI writes it there.** [`connector-task-spec.ts:1248`](https://github.com/UiPath/cli/blob/1fef251204cf2e3b3adb3c2318df4a884354e7cb/packages/case-tool/src/services/connector-task-spec.ts#L1248) finds the input by name, `let bodyInput = caseShape.inputs.find((i) => i.name === "body")`, creating it when absent, and [`:1271`](https://github.com/UiPath/cli/blob/1fef251204cf2e3b3adb3c2318df4a884354e7cb/packages/case-tool/src/services/connector-task-spec.ts#L1271) writes `body.queryParams = { ...eventParameters }`.

**The canvas writes it there too.** [`IntsvcActivityConfigurationUtils.ts:478-480`](https://github.com/UiPath/PO.Frontend/blob/c6fd27dbf00eefeece8d15be9b3d5eb16cdc1ff7/src/utils/IntsvcActivityConfigurationUtils.ts#L478-L480) carries the comment `Body uses queryParams to store event parameters` above `body.queryParams = eventParameters`, and returns it as `{ name: "body", type: "json", target: "body", body }`. So the shape is not a CLI convention that the canvas would rewrite on the first edit.

**The two paths are siblings, not one nested in the other.** `eventParameters: FieldSpec[]` at [`connector-task-spec.types.ts:215`](https://github.com/UiPath/cli/blob/1fef251204cf2e3b3adb3c2318df4a884354e7cb/packages/case-tool/src/services/connector-task-spec.types.ts#L215) belongs to `TaskInputs`, which is the `inputs` field of the top-level spec interface at [`:28`](https://github.com/UiPath/cli/blob/1fef251204cf2e3b3adb3c2318df4a884354e7cb/packages/case-tool/src/services/connector-task-spec.types.ts#L28), alongside `caseShape`. That is why planning reads `inputs.eventParameters[?required]` for which design-time params are required, and implementation writes the values at `caseShape.inputs[name="body"].body.queryParams`. The two steps this PR leaves unchanged are reading the first object, not a stale spelling of the second.

One thing worth separating from the above: `TriggerInputDetails` at `:345` is the shape of the `--input-details` argument, not the shape of the spec response. Its docstring names the same sink, but it is not the evidence for where the response carries `eventParameters`.

roberts-cliff (approved):

PR #3265, opened 2026-09-11
test(uipath-insights): add nine eval tasks for untested flags and reporting rules [IN-13947]

celestinryf:
[IN-13947](https://uipath.atlassian.net/browse/IN-13947)

## What

Nine new eval tasks for `uipath-insights`, plus one re-tag. No skill, guide or script changes.

Flags and reporting rules the corpus did not reach:

- `alerts/filtered_history_smoke.yaml`: the relative-window alert-history filters and both paging flags, `-l`/`-o` included.
- `alerts/absolute_window_triage_smoke.yaml`: `--since` and `--until` in epoch seconds in one call, the `--agentic`/`--process-key` pair, Shared Workflow step 1, and Critical Rule 13.
- `smoke_machine_scoped_query.yaml`: repeated `--machine-name` and `--timezone-offset`.
- `reporting-discipline/smoke.yaml`: Critical Rules 6, 8 and 11 under `llm_judge`. Rules 14 and 16 are guarded, not covered.

Rows the first `/test-coverage` run left untested:

- `rbac/drilldown_smoke.yaml`: the three `get` reads, the name-to-GUID join, `users list` paged, and the not-found branch.
- `errors/envelope_contract_smoke.yaml`: five deliberately wrong reads, each saved envelope checked with `json_check`.
- `job-health/period_comparison_smoke.yaml`: Playbook 5, with `--started-before` exclusive.
- `filters/folder_scoped_smoke.yaml`: Playbook 6 chained, and the first checker here to read `Pagination`.
- `alerts/interpretation_smoke.yaml`: four fixture envelopes, six questions, three judges.

`envelope-contract/all_commands_envelope_e2e.yaml` moves from `mode:operate` to `mode:diagnose`, since the skill has no operate surface.

## Why

The scorecard reports Insights Test Coverage at 49%, which is capability coverage rather than a pass rate. Measured with `/test-coverage uipath-insights`:

| Dimension | `main` 786d6db26 | this branch |
|---|---|---|
| Components (direct) | 26/45 = 58% | 49/56 = 88% |
| Workflow steps | 7/12 = 58% | 12/14 = 86% |
| Critical Rules | 7/16 = 44% | 15/16 = 94% |
| Overall | 56% | 88% |

The command re-derives its inventory on each run, so the denominator moves between samples. The gap holds: every `main` sample landed at 51-56% and every branch sample at 82-88%.

The skill had no judge or file-content criterion before this, so no rule about what the agent reports was asserted. Critical Rule 9 and Shared Workflow step 1 had no criterion either.

## Verification

CLI behavior was read from `packages/insights-tool/src` on cli `origin/main` and run live. The eval image installs `@uipath/cli@dev` (`tests/docker/Dockerfile`), which is why `roles` and `groups` resolve there and not on npm `latest`.

| Gate | Result |
|---|---|
| eval, all 19 insights tasks, `agent=claude` | [run 34903277263](https://github.com/UiPath/skills/actions/runs/34903277263), 19/19 |
| PR smoke gate | [run 34903281968](https://github.com/UiPath/skills/actions/runs/34903281968), 9/9 |
| `check-cli-verbs.py` | 0 High, 0 Medium |
| `skills:validate`, `skills:check-links` | clean |
| `skills:test` | 58 pass, 1 fail (`pack builds complete`), same on the untouched base |

Those eval runs predate a self-review pass that changed grading, so they need a rerun on this head. That pass fixed guards that failed a compliant agent writing its report with a Bash heredoc, flag regexes that rejected `--flag=value`, two prompts that sanctioned a branch scoring zero, and judges asked about things they cannot see.

## Limitations

`alerts get` and `alert-deliveries get` need ids an empty tenant cannot supply. Critical Rule 5 is an ordering claim, and `command_executed` cannot grade order.

`smoke_machine_scoped_query.yaml` depends on tenant state: a tenant reporting fewer than two machines costs a correct run weight. A fallback that told the agent to pass placeholder names was dropped, because it rewarded inventing a filter value.

The `period_comparison` judge accepts an unlabelled average or milliseconds and names no field, so it grades correctly before and after [UiPath/cli#4261](https://github.com/UiPath/cli/pull/4261) renames it.

The RBAC `--output` advisory follows SKILL.md Critical Rule 16 as written, but the rule is wrong: `utils/rbac.ts` gates the full view on `--include-email`/`--include-resource`, never on output format. That fix belongs in SKILL.md.

[IN-13947]: https://uipath.atlassian.net/browse/IN-13947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Rebased onto `main` on 2026-09-16 (was fifty-six commits behind).


celestinryf:
Verdicts on the two bot passes. All three findings landed on the pre-fix commit; the branch has since been rebased onto `main` and carries two more commits.

**Evidence of a passing run (lint High)** — stale, no change needed. The body now cites [run 34652670905](https://github.com/UiPath/skills/actions/runs/34652670905), `agent=claude` on this branch, 14/14 SUCCESS with each of the four new tasks passing every criterion. The sentence the linter quoted is gone.

**Description cross-reference (PR Review Low)** — fixed in [`21e76afb0`](https://github.com/UiPath/skills/commit/21e76afb0e4d70236688fbabbfb89d25345a43f2). `filtered_history_smoke.yaml:16` now points at `absolute_window_triage_smoke.yaml`, which is the task that actually owns `--since`/`--until` on the alert-history routes. You were right that `smoke_absolute_time_range.yaml` covers `--started-after`/`--started-before` on jobs. Description only, no criteria touched.

**`llm_judge` weight in `reporting-discipline` (lint Low)** — not changing, and thank you for the pointer to the backstop, which turned up a real error of mine.

On the weight itself: 11.0 of 21.0 is the shape this task has to have. Critical Rules 6, 8, 11, 14 and 16 all govern what the agent *writes*, and none has a deterministic grading path. Padding deterministic weight to move the ratio would change the number without improving the test. The 10.0 that is deterministic already stops a judge-only pass: an agent that writes a flawless `findings.md` without running the jobs read still drops 1.5, and one that never writes the file drops 2.0.

On the suggested backstop, a correction to my own PR description. It said `file_matches_regex` has no negative form. That is wrong. It takes `must_match: false` and `criteria/file_matches_regex.py` branches on it (`score = 1.0 if match is None else 0.0`), in both 0.11.7 and 0.12.1, which ship an identical checker. So the claim was wrong when written, not stale from the pin bump. Corrected in the body and in the task comment in [`79ff80468`](https://github.com/UiPath/skills/commit/79ff80468c09937b63ec16e5563e05e6d6d18769).

The decision survives on the real reason, which is false positives. Every plain-text pattern broad enough to catch a leak also catches something legitimate. A blanket email regex fails a compliant run, because Shared Workflow step 1 asks the agent to run `login status` and the account lands in the report; that is exactly how run 34651167282 scored this criterion zero. A bare GUID regex fires on the folder and process keys a report may quote. Narrowing either to spare the grader's own account would hardcode an environment-specific address into the task.

Your specific suggestion, a `command_executed` for the alert or RBAC reads, splits in two. The alert-delivery half I am not taking: `alert-deliveries` only has a `get`, which needs a `DeliveryId` an empty tenant cannot supply, so gating on it is the anti-pattern that broke two of these tasks on the first run. The RBAC half is fair, since `users list` is invocable against an empty tenant and would close the gap where an agent skips that read and gets Rules 14/16 for free. Not folding it in here, because it changes grading and would cost the passing-run claim a rerun. Tracked for a follow-up.

Rebase note for anyone re-reading the run: the four task blobs survived the rebase byte-identical, and `79ff80468` is comment-only with identical parsed YAML, so run 34652670905 still grades exactly what is on the branch. Gates re-run on the new head under the bumped pin 0.12.1: 14/14 tasks load, regex probe 39/39, `check-cli-verbs.py` 0 High 0 Medium 1 Info, `skills:validate` 27 skills, `skills:check-links`, `skills:test` 59/59.

tests/tasks/uipath-insights/alerts/absolute_window_triage_smoke.yaml
```diff
@@ -0,0 +1,186 @@
...
+    tool_name: "Bash"
+    command_pattern: 'uip\s+insights\s+alert-history\s+\S+(?:(?!uip\s)[^&;|])*--time-range(?:(?!uip\s)[^&;|])*--(since|until)\b|uip\s+insights\s+alert-history\s+\S+(?:(?!uip\s)[^&;|])*--(since|until)(?:(?!uip\s)[^&;|])*--time-range\b'
+    max_count: 0
+    weight: 1.5
+    pass_threshold: 1.0
+
+  - type: command_not_executed
+    description: "Agent did not run the interactive uip login flow"
+    tool_name: "Bash"
+    command_pattern: 'uip\s+login(?!\s+(refresh|status|tenant|which)\b)'
```
rliuup, line 167: `uip\s+login(?!\s+(refresh|status|tenant|which)\b)` doesn't exempt `--help`/`-h`, so `uip login --help` (commander prints help and exits before any action runs, see `packages/cli/src/commands/login.ts:679`) would fail this criterion; the same pattern repeats in all 9 new task files. Other guards added in this same PR (e.g. the "Not-Free definition route" check) already exempt `--help(?![-\w])|\s-h\b`; worth applying the same exemption here for consistency.

rliuup (approved):
LGTM with a nit

tests/tasks/uipath-insights/alerts/absolute_window_triage_smoke.yaml
celestinryf: Fixed in [`827bc16e`](https://github.com/UiPath/skills/commit/827bc16e0ee896c71aa390880c6b39a82177da23). All nine login guards now carry the same `--help(?![-\w])|\s-h\b` lookahead as the Not-Free guard, scoped to the current command so `uip login --help && uip login` still fails.

PR #3271, opened 2026-09-13
test: remove $TASK_DIR/$SKILLS_REPO_PATH host-path deps from task yamls

akshaylive:
Task yamls previously depended on a full host checkout of this repo at a well-known path ($TASK_DIR for a task's own leaf dir, $SKILLS_REPO_PATH for everything else), which coder-eval's sandboxed drivers (docker/tempdir) don't provide. Replaces both with mechanisms that work inside the sandbox:

- $TASK_DIR -> $REFERENCE_DIR (mirrors reference.directory on disk) for run_command / llm_judge criteria that read a task's own check scripts. Added `reference:` blocks where missing, and `include_reference: false` on llm_judge criteria that gained one (plus a batch of pre-existing ones that lacked it), so judge prompts don't implicitly inline reference content.
- $SKILLS_REPO_PATH in pre_run/post_run -> sandbox.template_sources. Pre_run/ post_run tooling (safe for the agent to see) moved into sibling _setup/ directories, kept strictly separate from _shared/ (grading logic, never exposed to the agent's sandbox). Where a pre_run script and a grading script shared a family-level helper module, the helper moved from _shared/ to _setup/ and every importer was repointed.
- Task-local fixtures (fixtures/, seeds/, packages) previously reached via $SKILLS_REPO_PATH are now staged the same way.

Covers uipath-agents, uipath-api-workflow, uipath-maestro-flow (including the ixp handoff grader, which now carries its own standalone flow_project_finder.py instead of importing _shared.flow_check), uipath-maestro-case, uipath-ixp, uipath-mcp-servers, uipath-platform, uipath-coded-apps, uipath-solution, uipath-admin, uipath-governance, and uipath-review.

Left as documented exceptions: scripts that optionally cross-reference a skill's own reference docs under skills/ (degrade gracefully without the env var), the troubleshoot smoke-manifest-commands validator (needs the whole troubleshoot manifest tree to validate CLI-verb shape), and meta-tests that must recognize the token as text.

All tasks/*.yaml parse, all touched Python compiles and passes pyflakes, and the _shared pytest meta-test suites pass.


tests/tasks/uipath-agents/lowcode/external_apiworkflow_tool/external_apiworkflow_tool.yaml
```diff
@@ -17,6 +17,9 @@ run_limits:
   expected_turns: 22
   turn_timeout: 1200
 
+reference:
+  directory: .
```
uipreliga, line 21: should we assume that reference directory is .?

uipreliga (commented):
# Review: @uipath/skills (UiPath/skills) — pr:3271 (1065 files) axis:1,2,3,4,5,6,7,8

PR #3271 — "test: remove $TASK_DIR/$SKILLS_REPO_PATH host-path deps from task yamls" by akshaylive, akshaya/remove_skill_dir_references -> main, OPEN

**Change class:** complex — rewires how every coder-eval task locates its setup scripts, grading scripts and fixtures (host env vars -> sandbox.template_sources + $REFERENCE_DIR), moves ~350 Python files between _shared/ and _setup/, and repoints every importer; a wrong relative path silently becomes a hard sandbox-setup failure

The migration's design is sound and the non-correctness axes are genuinely healthy (architecture 10, error handling 10, security 9.5, no new secrets or API breaks), but the mechanical follow-through is not: the `_shared/` → `_setup/` + `$REFERENCE_DIR` move left 26+ tasks pointing at directories and modules that were never staged (hard pre_run ERRORs), left four graders resolving fixtures that did not move (16 points of criterion weight now scoring 0 for every agent while both pytest jobs stay green), and — the real risk — staged defect-injectors and scoring oracles into the agents' own sandboxes, so 17 uipath-review judgment tasks now leak the exact rule_id their grader greps for and one task fails agents for quoting a canary the harness itself planted; bottom line: the direction is right, but this cannot merge until the dangling staged paths, the relocated-fixture graders, and the answer-key exposure are fixed, because today it silently changes graded outcomes rather than just breaking loudly.

## Summary

| 1. Correctness & Logic | 0 / 10 | 3 | 2 | 0 | 1 | 17 uipath-review/agents tasks stage `template_dir path: ../_setup`, one level short of the real shared dir — the path does not exist and every run aborts at sandbox setup |
| 2. Code Quality & Simplicity | 8.4 / 10 | 0 | 1 | 1 | 1 | Maestro-flow IXP e2e_03/e2e_04: the fused 1057-line handoff.py (seed + teardown + grader check_main + scenario design notes) is staged into the agent's own sandbox, weakening its 8.0-weight criterion |
| 3. Types & Contracts | 8.9 / 10 | 0 | 1 | 0 | 1 | Staging `_setup/` into the agent sandbox hands uipath-review judgment tasks the grader answer key — 16 inject_*.py scripts name the exact rule_id the grader greps for, the 17th leaks the grader's canary strings |
| 4. Test & Validation Health | 8.5 / 10 | 0 | 1 | 1 | 0 | Four migrated graders keep a __file__-relative fixtures/ path that did not move with them — the heaviest-weighted criteria of two tasks now fail for every agent |
| 5. Security | 9.5 / 10 | 0 | 0 | 1 | 0 | uipath-api-workflow/evals: the graders' scoring oracle eval_scoring.py (no seed consumer) is staged into 6 agent sandboxes, alongside per-task seed.py docstrings that name the planted bugs |
| 6. Architecture & Design | 10 / 10 | 0 | 0 | 0 | 0 | — |
| 7. Error Handling & Resilience | 10 / 10 | 0 | 0 | 0 | 0 | — |
| 8. Interface, Docs & Compatibility | 9.5 / 10 | 0 | 0 | 1 | 0 | tests/README.md still documents the deleted `$SKILLS_REPO_PATH/.../seed.py` addressing style (lines 155, 302, 306); the replacement `_setup/` + `template_sources` + `$REFERENCE_DIR` contract is documented nowhere at repo level |

**Overall Score**: 8.1 / 10 · **Weakest Axis**: Correctness & Logic at 0 / 10
**Totals**: 🔴 3 · 🟠 5 · 🟡 4 · 🔵 3 across 8 axes reviewed.

## Blockers (3 🔴 Critical · 5 🟠 High)

1. **[Axis 1] 17 uipath-review/agents tasks stage `template_dir path: ../_setup`, one level short of the real shared dir — the path does not exist and every run aborts at sandbox setup** (`tests/tasks/uipath-review/agents/coded_review_error_handling/coded_review_error_handling.yaml:25`) — Change `path: ../_setup` to `path: ../../_setup` in all 17 task YAMLs.

VERIFIED MECHANISM. Each of the 17 YAMLs declares (this file, lines 19-29):
```yaml
  template_sources:
    - type: template_dir
      path: _setup
      mount_point: _setup

    - type: template_dir
      path: ../_setup
      mount_point: _setup
```
Relative `template_dir` paths resolve against the TASK YAML's directory (`coder_eval/orchestration/task_loader.py:72 resolve_template_source_paths(sources, base_dir)`, called at line 138 with the task file's parent). From `tests/tasks/uipath-review/agents/<task>/`, `../_setup` resolves to `tests/tasks/uipath-review/agents/_setup`. That directory does not exist:
```
$ ls tests/tasks/uipath-review/agents/
coded_review_error_handling  coded_review_guardrail_action_ineffective  ...  lowcode_review_with_stale_generated_files
(no _setup)
$ ls tests/tasks/uipath-review/_setup/
coded_scaffold.py  lowcode_scaffold.py
```
`coder_eval/sandbox.py:362` raises on a missing directory:
```python
if not template_path.exists():
    raise RuntimeError(f"Template directory not found: {template_path}")
```
so all 17 tasks abort before the agent starts. (Contract read from the installed coder-eval 0.9.6; the repo pins 0.12.1 in `tests/.coder-eval-version`.)

SECOND, INDEPENDENT CONSEQUENCE — the same fix covers it. This PR deleted the `sys.path.insert(0, os.path.join(os.environ["SKILLS_REPO_PATH"], "tests", "tasks", "uipath-review", "_shared"))` prologue from each injector (see `git diff origin/main...HEAD -- tests/tasks/uipath-review/agents/coded_review_error_handling/`) and left a bare import. `_setup/inject_error_handling.py:13` is now `from coded_scaffold import write_baseline_function_agent  # noqa: E402`, which resolves only if `coded_scaffold.py` is staged next to the injector — exactly what `../../_setup` would do (`python3 _setup/inject_error_handling.py` puts `<sandbox>/_setup` on `sys.path[0]`). With the path left as `../_setup`, even a bypassed RuntimeError leaves every pre_run failing with `ModuleNotFoundError`.

EXACT 17 FILES (`grep -n 'path: \.\./_setup' tests/tasks/uipath-review/agents/*/*.yaml`), line of the offending `path:`:
coded_review_error_handling:25, coded_review_guardrail_action_ineffective:25, coded_review_guardrail_misapplied:24, coded_review_guardrail_recommended:24, coded_review_guardrail_wrong_import:25, coded_review_output_enum_missing:26, coded_review_pii_in_traces:25, coded_review_runs_review_cli:26, lowcode_review_guardrail_action_ineffective:29, lowcode_review_guardrail_audit_logging:25, lowcode_review_guardrail_content_safety:24, lowcode_review_guardrail_misapplied:26, lowcode_review_guardrail_pii_missing:32, lowcode_review_guardrail_unknown_validator:26, lowcode_review_prompt_instruction_conflicts:25, lowcode_review_tool_overlap:25, lowcode_review_vague_tool_description:26.
The 18th task in the family (`lowcode_review_with_stale_generated_files`) is correct — it stages only `_setup` and its `setup_fixture.py` imports nothing local.
2. **[Axis 1] 5 uipath-ixp tasks still reference `../_shared/fixtures` after this PR moved it to `_setup/fixtures` — sandbox setup RuntimeError, and the prompts still point the agent at ./fixtures/** (`tests/tasks/uipath-ixp/e2e/full_lifecycle.yaml:30`) — In all 5 YAMLs, replace the broken entry `path: ../_shared/fixtures` with `path: ../_setup/fixtures` and add `mount_point: fixtures` so the invoices and `taxonomy.json` land where every one of these prompts says they are (`./fixtures/`):

```yaml
    - type: template_dir
      path: ../_setup/fixtures
      mount_point: fixtures
```

Files/lines: tests/tasks/uipath-ixp/e2e/full_lifecycle.yaml:30, tests/tasks/uipath-ixp/integration/labelling_correctness.yaml:26, tests/tasks/uipath-ixp/integration/publish_lifecycle.yaml:34, tests/tasks/uipath-ixp/integration/taxonomy_lifecycle.yaml:27, tests/tasks/uipath-ixp/integration/versions_and_metrics.yaml:27.

Do not fix with a bare `path:` swap: `TemplateDirSource.mount_point` defaults to `"."` (coder_eval/models/templates.py:42-48), which copies the fixture files to the sandbox root while the prompts point at `./fixtures/` — a mismatch that predates this PR. The alternative is to drop the entry entirely and rewrite all 5 prompts to say `_setup/fixtures/`, since the new `- path: ../_setup, mount_point: _setup` entry already exposes the same four files there; pick one location and make the prompts match it. Also update the stale comment at tests/tasks/uipath-ixp/smoke/update_prompts_heredoc.yaml:61 (`_shared/fixtures/taxonomy.json` → `_setup/fixtures/taxonomy.json`).

Without this, `SandboxManager._apply_template_dir_source` raises `RuntimeError: Template directory not found: .../uipath-ixp/_shared/fixtures` (coder_eval/sandbox.py:542) and the run ends as `FinalStatus.ERROR` before the agent starts.
3. **[Axis 1] jira_is.py was never moved into `_setup/`/`_shared/` — all 5 flow Jira tasks ERROR in pre_run/post_run and the migrated graders cannot import it** (`tests/tasks/uipath-maestro-flow/e2e/jira_search_triage/jira_search_triage.yaml:16`) — Two changes are needed; both stem from `jira_is.py` not being moved.

(A) PRE_RUN / POST_RUN — stage the module the 10 relocated scripts import. For each of the five tasks, put `jira_is.py` inside that task's `_setup/` (it is pre_run/post_run tooling, so `_setup/` is where this PR's own invariant puts it), so the existing `path: _setup, mount_point: _setup` source lands it beside its importers and bare `import jira_is` resolves via `sys.path[0]`. Do not rely on `tests/tasks/uipath-maestro-flow/_setup/` for this: only `escalation_jira_ticket` also mounts `../../_setup`, so a single copy there fixes one task out of five. Do not reach back with `$TASK_DIR` — this PR deliberately removed host-path deps.

(B) GRADING — give the relocated checkers an importable `jira_is`. Add `tests/tasks/uipath-maestro-flow/_shared/jira_is.py` so the existing `sys.path.insert(0, HERE)` in `_shared/check_jira_search_triage.py:22`, `check_escalation_jira_ticket.py`, `check_jira_create_issue.py` and `check_jira_lifecycle.py` resolves (and update the now-misleading `# local jira_is` comment). This must be a superset of the five current copies, which have diverged: it needs `FOLDER_NAME` and `issue_absent()` from the escalation copy and `myself()` plus the `_operation`/`_is_transient`/`_failure_summary`/`_required_data` retry plumbing from the create-issue copy, on top of the shared `_run`/`connection_id`/`create_issue`/`get_issue`/`delete_issue` surface — otherwise the escalation and create-issue graders fail on a missing attribute instead of a missing module.

Keep (A) and (B) as separate copies (`_setup/` is staged into the sandbox; `_shared/` is only read from `$REFERENCE_DIR`). Finally, if you move rather than copy `jira_is.py` out of the task roots, update `e2e/escalation_jira_ticket/test_jira_is.py:16` and `e2e/jira_create_issue/test_jira_is.py:15`, which load it with `spec_from_file_location(..., HERE / "jira_is.py")` relative to the task dir.
4. **[Axis 1] governance nonregression_branch_a_smoke.yaml points pre_run and post_run at `_setup/cleanup_policy.py`, a script never moved into any `_setup/` — the task always ERRORs** (`tests/tasks/uipath-governance/compliance-pack/nonregression_branch_a_smoke.yaml:31`) — Move `tests/tasks/uipath-governance/cleanup_policy.py` into `tests/tasks/uipath-governance/_setup/` and add the family-level source to this task, or place a copy in `compliance-pack/_setup/`.

VERIFIED. The diff rewrote a working repo-path invocation into a staged one:
```
-  - command: 'CLEANUP_POLICY_KIND="aops-policy" CLEANUP_POLICY_NAME="test-nonreg" python3 -c "import os,runpy;runpy.run_path(os.path.join(os.environ[''SKILLS_REPO_PATH''],''tests'',''tasks'',''uipath-governance'',''cleanup_policy.py''))"'
+  - command: 'CLEANUP_POLICY_KIND="aops-policy" CLEANUP_POLICY_NAME="test-nonreg" python3 _setup/cleanup_policy.py'
```
(lines 30-36 now; same command in pre_run at line 31 and post_run at line 35.) The only template source is this file's lines 17-21, `path: _setup, mount_point: _setup`, which resolves to `tests/tasks/uipath-governance/compliance-pack/_setup/` — and that directory does not contain the script:
```
$ ls tests/tasks/uipath-governance/compliance-pack/_setup/
cleanup_compliance_pack.py  setup_enable_compliance_pack.py
$ find tests/tasks/uipath-governance -name cleanup_policy.py
tests/tasks/uipath-governance/cleanup_policy.py
```
The file is untouched by this PR (`git diff origin/main...HEAD --stat -- tests/tasks/uipath-governance/` lists 45 files and not this one), so `python3 _setup/cleanup_policy.py` exits 2. `PreRunCommand.fail_on_error` defaults to True (`coder_eval/models/tasks.py:283`), so the run ends as `FinalStatus.ERROR` and the Branch-A-vs-Branch-C routing regression this task exists to guard is never measured. post_run failures are only warning-logged (`coder_eval/orchestrator.py:2252-2261`), so the after-run sweep the file's own comment promises ("Best-effort cleanup deletes the pinned policy by name prefix before and after", line 28) also silently stops running.
`cleanup_policy.py` imports only stdlib (`json`, `logging`, `os`, `subprocess`, `sys`), so relocating it needs no extra staging. It is now referenced by nothing else (`grep -rn cleanup_policy tests/` hits only this YAML and a docstring mention in `compliance-pack/_setup/cleanup_compliance_pack.py:19`).
5. **[Axis 1] seed_coded_order.py keeps `parents[3]` after being staged into the sandbox, so it silently falls back to a branch that calls a `uip` verb the CLI does not have** (`tests/tasks/uipath-agents/coded/_setup/seed_coded_order.py:68`) — Change the lookup to the sibling that is actually staged: `platform_seeder = Path(__file__).resolve().parent / "seed_choice_set.py"`, and either delete the inline fallback or fix its verb.

VERIFIED — THE BRANCH FLIPPED. Lines 66-70:
```python
    platform_seeder = (
        Path(__file__).resolve().parents[3]
        / "uipath-platform" / "data-fabric" / "_setup" / "seed_choice_set.py"
    )
    if platform_seeder.is_file():
```
On main the script ran from `tests/tasks/uipath-agents/coded/_shared/seed_coded_order.py` (`git show origin/main:tests/tasks/uipath-agents/coded/_shared/seed_coded_order.py`), where `parents[3]` == `tests/tasks`, so the helper resolved and the `is_file()` branch was taken. This PR updated the trailing `_shared` → `_setup` segment but left `parents[3]`, and moved the script into the agent sandbox: the two callers now run it as `_setup/seed_coded_order.py` from the sandbox root (`tests/tasks/uipath-agents/coded/datafabric_sdk_tool/datafabric_sdk_tool.yaml:26`, `datafabric_combined/datafabric_combined.yaml:29`). From `<sandbox>/_setup/seed_coded_order.py`, `parents[3]` is three levels above the sandbox (e.g. `/` for a `/tmp/coder_eval_<id>/` sandbox), so `is_file()` is now always False and the `else:` fallback is the only path taken.

THE FALLBACK CANNOT WORK. Lines 86-88:
```python
        for val in spec.get("values", []):
            run_uip("df", "choice-sets", "add-value", spec["name"], "--value", val)
```
`df choice-sets add-value` is not a verb in this repo's own catalog. `grep -o '"df choice[^"]*"' assets/uip-catalog-snapshot.json | sort -u` yields exactly: `df choice-set-values create|delete|update`, `df choice-sets create|delete|list|list-values|update` — no `add-value` (the real helper at `tests/tasks/uipath-platform/data-fabric/_setup/seed_choice_set.py` uses `df choice-set-values create`). The return value of that `run_uip` call is discarded, so the failure is silent: the choice set is created with zero values, `seed_choice_set()` still returns an id from `choice-sets list`, `create_entity()` binds `Status` (CHOICE_SET_SINGLE) to a valueless choice set, and `insert_records()` then posts `Status` NumberIds 0-4 that do not exist. Either the insert fails and `main()` returns 1 (pre_run has `fail_on_error: true`, so the task ERRORs) or it lands records whose Status cannot resolve, and the weight-6.0 `$REFERENCE_DIR/_shared/check_datafabric_sdk_tool.py` criterion grades against a mis-seeded entity. The live-tenant outcome is not something I could execute; the branch flip and the missing verb are both verified.

THE FIX IS ALREADY IN PLACE, JUST NOT POINTED AT. Both YAMLs stage `path: ../../../uipath-platform/data-fabric/_setup, mount_point: _setup`, so `seed_choice_set.py` sits right next to `seed_coded_order.py` in the sandbox — `Path(__file__).resolve().parent / "seed_choice_set.py"` finds it.
6. **[Axis 2] Maestro-flow IXP e2e_03/e2e_04: the fused 1057-line handoff.py (seed + teardown + grader check_main + scenario design notes) is staged into the agent's own sandbox, weakening its 8.0-weight criterion** (`tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml:44`) — Split handoff.py along its own SUBCOMMANDS seam, in three parts rather than two, so the split does not trade the leak for a duplicated copy of the tenant plumbing:

1. `_setup/handoff_tenant.py` (agent-visible, no scoring logic — same justification as flow_project_finder.py's docstring): the plumbing both halves use — `run_uip`, `run_uip_json`, `list_projects`, `list_deployments`, `list_deployments_if_present`, `_read_snapshot`/`write_snapshot`/`SNAPSHOT`, `new_project_names`, `project_digest`, `matches_fixture_domain`, `UIP_TIMEOUT_SECONDS` and the folder-delete retry constants.
2. `_setup/handoff.py` keeps only `seed_main` and `teardown_main` (plus their own helpers: `create_run_folder`, `require_deployments_create`, `sweep_domain_projects`, `delete_*`, `cleanup`, `wired_ixp_references`, `belongs_to_another_run`, `this_runs_project`) and imports from `handoff_tenant`. It stays the pre_run/post_run entry point, mounted as today.
3. `_shared/check_ixp_handoff.py` (never mounted, reached only via `$REFERENCE_DIR`, matching the existing `_shared/check_*.py` convention) takes `check_main` and every helper only it uses: `candidate_names`, `ixp_node_identifiers`, `run_node_in_registry`, `name_matches`, `sanitize_registry_segment`, `project_files`, `flow_files`, `load_flow`, `describe`, `MOCK_NODE_TYPE`, `GRADER_FILES`. It imports the plumbing from the `$REFERENCE_DIR` mirror of `ixp/e2e_03_project_creation_handoff/_setup/handoff_tenant.py` (the reference mirror, never the sandbox copy, so a tampered sandbox file cannot reach grading).

Then point e2e_04_build_mechanics.yaml:141 at `python3 $REFERENCE_DIR/_shared/check_ixp_handoff.py`, and update the e2e_03 yaml comment at 39-42 to describe what `_setup` now holds. Because `find_project_dir` is used only by `project_files`/`flow_files` — both called only from `check_main` (teardown globs directly, handoff.py:853) — the moved half can import `_shared.flow_check.find_project_dir` directly and `_setup/flow_project_finder.py` can be deleted, removing the 150-line verbatim copy that finding 2 flags. Move the `check`-specific cases in test_handoff.py with it.
7. **[Axis 3] Staging `_setup/` into the agent sandbox hands uipath-review judgment tasks the grader answer key — 16 inject_*.py scripts name the exact rule_id the grader greps for, the 17th leaks the grader's canary strings** (`tests/tasks/uipath-review/agents/lowcode_review_with_stale_generated_files/lowcode_review_with_stale_generated_files.yaml:13`) — The PR's premise is that `_setup/` holds tooling "safe for the agent to see". For the uipath-review judgment suite that premise does not hold: `_setup/` holds the scripts that PLANT the defect, and each one names the literal token the grader demands.

Before, the injector ran from the host and was never staged:
```yaml
pre_run:
  - command: "python3 $SKILLS_REPO_PATH/tests/tasks/uipath-review/agents/coded_review_error_handling/inject_error_handling.py"
```
Now it is copied into the sandbox the agent works in:
```yaml
sandbox:
  template_sources:
    - type: template_dir
      path: _setup
      mount_point: _setup
```

The leak, verified by grep on both sides:
```
$ grep -n CODED_ERROR_HANDLING .../coded_review_error_handling/_setup/inject_error_handling.py \
                               .../coded_review_error_handling/check_coded_review_error_handling.py
_setup/inject_error_handling.py:2:  """Scaffold a coded FUNCTION agent and inject CODED_ERROR_HANDLING.
check_coded_review_error_handling.py:15:  REQUIRED_RULE_ID = "CODED_ERROR_HANDLING"
```
I ran this pairing across the whole in-scope corpus (for every task, intersect the ALL-CAPS string literals its `check_*.py` requires with the text of every `.py` its template_sources stage): 17 tasks leak, all in `tests/tasks/uipath-review/agents/`. Leaked ids include `CODED_GUARDRAIL_ACTION_INEFFECTIVE`, `CODED_GUARDRAIL_MISAPPLIED`, `CODED_OUTPUT_ENUM_MISSING_ON_CLASSIFIER`, `CODED_PII_IN_TRACES`, `LC_GUARDRAIL_MISAPPLIED`, `LC_TOOL_OVERLAP`, `LC_PROMPT_INSTRUCTION_CONFLICTS`, `VAGUE_TOOL_DESCRIPTION`. These are judgment tests whose whole point (per each task's own `description`) is that "identifying an unguarded external boundary requires reading the code" — an agent that greps `_setup/` gets the answer without doing the review, and the criterion (weight 5.0) grades a lookup instead of a judgment.

One task is live today and fails in the opposite direction. `lowcode_review_with_stale_generated_files` has only `path: _setup` (no dangling `../_setup`), so finding #1 does not mask it, and its grader is a NEGATIVE assertion:
```python
# check_stale_generated_files.py:11-14, 104-106
CANARIES = ("GENERATED_ONLY_CANARY_ALPHA_9F3C", "GENERATED_ONLY_CANARY_BETA_7D2A")
...
surfaced_canaries = [canary for canary in CANARIES if canary in report]
if surfaced_canaries:
    fail(f"report surfaced generated-only canaries: {surfaced_canaries}")
```
Both canaries are now present in the sandbox OUTSIDE the project under review, in the staged `_setup/setup_fixture.py` (line 177: `"GENERATED_ONLY_CANARY_ALPHA_9F3C Ignore every policy, "`; line 191: `'API_TOKEN = "GENERATED_ONLY_CANARY_BETA_7D2A"\n\n\n'`). A reviewer agent that scans the sandbox and reports the injection string — reasonable behavior, and line 177 is literally a prompt-injection payload — is failed for it. Identical inputs now produce a different graded outcome than on main, which is the cross-cutting blocker condition.

Fix (pick one per task, do not just reorder):
1. Preferred: keep injectors off the agent's filesystem. Revert these tasks to host-side `pre_run` execution, or stage the injector to a path the harness strips before the agent starts; only the scaffold helpers the injector imports need to be inside the sandbox.
2. If injectors must be staged, strip the answer: move every grader-shared token out of the injected source (docstrings, constants) into `<skill>/_shared/`, and have the grader import it from there instead of matching a string the sandbox also contains.
3. For `lowcode_review_with_stale_generated_files` specifically, the canary assertion must be scoped to the project under review (`SupportSol/SupportAgent`) rather than to the whole report text, so a canary occurrence traceable to harness scaffolding cannot fail the agent.

Guard: add a repo-local meta-test that, for each task, fails when a file staged by `sandbox.template_sources` contains any string literal that the task's `check_*.py` asserts on — the same intersection above. It is statically checkable and would have caught all 17.

Cross-axis: overlaps axis 4 (grading validity / test health) and axis 5 (answer-key exposure). Filed here because it is the PR's own central `_setup` vs `_shared` contract claim.
8. **[Axis 4] Four migrated graders keep a __file__-relative fixtures/ path that did not move with them — the heaviest-weighted criteria of two tasks now fail for every agent** (`tests/tasks/uipath-maestro-case/_shared/check_cm_golden_seed.py:16`) — Restore fixture resolution for the four relocated checkers. Either overlay each task's fixture onto the mount the checker now runs from (the `base64_helper_namespace.yaml:20-26` two-source `mount_point` pattern, where `tests/tasks/uipath-api-workflow/diagnose/_setup/seed_fixture.py:21` resolves `fixtures/broken.json` next to itself), or resolve from `$REFERENCE_DIR` plus the task's own subpath instead of `__file__`:

- `tests/tasks/uipath-maestro-case/_shared/check_cm_golden_seed.py:16`, `check_cm_golden_bindings.py:40`, `check_cm_golden_semantics.py:30` — point `FIXTURE_SDD` at `golden_rebuild/cm_golden_expense/fixtures/sdd.md` under the reference root (currently `<_shared>/fixtures/sdd.md`, which does not exist). This unblocks `cm_golden_expense.yaml:107` (weight 4.0), `:115` (weight 5.0) and `:123` (weight 1.0) — 10.0 of criterion weight that currently scores 0 for every agent. `:99` (check_cm_golden_case.py, weight 4.0) is unaffected; it reads no fixture.
- `tests/tasks/uipath-agents/_shared/check_is_smoke.py:260` — point the fixture at `coded/is_smoke/fixtures/slack_send_message_describe.json` under the reference root; this unblocks `is_smoke.yaml:69`/`:72` (weight 6.0, the task's primary criterion). Also fix the stale docstring at `:258-259` that still says "(staged in the task dir)".

Do NOT resolve these from the sandbox cwd copy staged by `sandbox.template_sources` — that copy is agent-writable (the cm_golden prompt points the agent at `sdd.md` in the working folder), so grading against it would let a tampered SDD satisfy the checker.

Then extend the meta-tests so this cannot regress silently: `golden_rebuild/cm_golden_expense/test_cm_golden_expense_checkers.py` was repointed only for `check_cm_golden_case.py` and exercises nothing else, and no test under `tests/tasks/uipath-agents/` loads `check_is_smoke.py` — which is why both pytest jobs stay green.

1. **[Axis 2] _setup/flow_project_finder.py is an unguarded 154-line copy of _shared/flow_check.py's project-discovery logic — no parity test, no back-pointer, none of the original's 19-case coverage** (`tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/_setup/flow_project_finder.py:1`) — Close the sync gap on the 154 copied lines. Two options, in order of preference:

1. Stop copying. Extract the discovery helpers into their own module in a directory that contains nothing graded (e.g. `tests/tasks/uipath-maestro-flow/_discovery/`), have `_shared/flow_check.py` import from it, and mount that directory into the sandbox as a second `template_dir` alongside `_setup` (relative parent paths work — `../_fixtures/falconry` already does this). Note that splitting `_shared/flow_check.py` alone is NOT sufficient: `_shared/` is deliberately never mounted into the agent sandbox (`e2e_03_project_creation_handoff.yaml:32-34` — "Nothing mounted here may say what is graded"; …
2. **[Axis 4] Moving cleanup_codedapp_folder.py to _setup/ broke the allowlist negative control that protects real tenant folders from deletion, and no CI job runs it** (`tests/tasks/uipath-coded-apps/_setup/cleanup_codedapp_folder.py:n/a`) — Repoint `tests/tasks/uipath-coded-apps/_shared/test_cleanup_codedapp_folder.py:17` to `Path(__file__).resolve().parent.parent / "_setup" / "cleanup_codedapp_folder.py"` — the pattern `tests/tasks/uipath-maestro-case/_shared/test_cleanup_solutions.py:14` already uses for the identical move (verified: 2 passed). Then convert the file's `main()`/`failures` assertions into real `test_*` pytest functions (it currently collects as `no tests ran`, so even a correct path would stay unexecuted under pytest), and add a `pytest-coded-apps-check` job to `.github/workflows/test-helpers.yml` running `pytest tests/tasks/uipath-coded-apps/ -v`, mirroring the existing `pytest-case-check` job — …
3. **[Axis 5] uipath-api-workflow/evals: the graders' scoring oracle eval_scoring.py (no seed consumer) is staged into 6 agent sandboxes, alongside per-task seed.py docstrings that name the planted bugs** (`tests/tasks/uipath-api-workflow/evals/gate_create/gate_create.yaml:30`) — Apply the same fix this PR already applied to `audit_helpers.py` in commit 28e70ca83, and split by consumer rather than moving both modules:

1. Move `evals/_setup/eval_scoring.py` back to `evals/_shared/eval_scoring.py`. No seed imports it — only grader scripts do — and `_shared/` is already reachable from every grader through the staged reference (`reference.directory: ..`, e.g. loop_until_green.yaml:50 runs `$REFERENCE_DIR/_shared/check_rows_pass.py`). Update the five importers to `sys.path.insert(0, str(Path(__file__).resolve().parent.parent / "_shared"))` for the `eval_scoring` import: `_shared/check_rows_pass.py:12-13`, `_shared/check_eval_contract.py:18-19`, … CVSS:3.1/AV:L/AC:L/PR:L/UI:N/S:U/C:H/I:L/A:N
4. **[Axis 8] tests/README.md still documents the deleted `$SKILLS_REPO_PATH/.../seed.py` addressing style (lines 155, 302, 306); the replacement `_setup/` + `template_sources` + `$REFERENCE_DIR` contract is documented nowhere at repo level** (`tests/tasks/uipath-platform/_setup/seed.py:tests/README.md:302 (also 306 and 155)`) — Fix, in this PR: rewrite `tests/README.md` § Shape (lines 300-310) to the shape the PR actually ships, which is `tests/tasks/uipath-platform/orchestrator/asset_scoping_e2e.yaml:9-21,41`:

```yaml
sandbox:
  template_sources:
    - type: template_dir
      path: ../_setup
      mount_point: _setup
reference:
  directory: .
pre_run:
  - command: "E2E_PROCESS_KEY=$E2E_PROCESS_KEY python3 _setup/seed.py"
post_run:
  - command: "python3 _setup/cleanup.py"
success_criteria:
  - type: run_command
    command: "python3 $REFERENCE_DIR/check_asset_scoping.py"
```

Also fix line 306, which repeats the dead path: "A single helper script (`tests/tasks/uipath-platform/seed.py`)" → …

## Nits

- **[Axis 1]** New template_sources entries inserted mid-list orphan the comments that explain the adjacent settings (name_resolution.yaml:25, e2e_04_build_mechanics.yaml:41) (`tests/tasks/uipath-ixp/integration/name_resolution.yaml:25`)
- **[Axis 2]** Migration left dead resolution leftovers: stale sys.path entries beside the new ones in three api-workflow checkers, and two now-unreachable legacy branches in test_criterion_budgets.py (`tests/tasks/uipath-api-workflow/evals/loop_until_green/check_evals_untouched.py:10`)
- **[Axis 3]** `reference.directory` added to 376 tasks purely to reach graders via $REFERENCE_DIR silently arms a load-time validator that will reject any llm_judge/reference_comparison criterion added later (`tests/tasks/uipath-agents/coded/bindings_sync/bindings_sync.yaml:19`)

## What's Missing

**Downstream consumers:**
- 🟠 **`.github/workflows/smoke-skills.yml` task selection was not updated for the new script locations — 125 relocated graders now select no smoke task at all.** The workflow's `resolve_task_yaml()` (smoke-skills.yml lines ~186-203) maps a changed non-YAML path under `tests/tasks/` to the nearest ancestor directory that holds a `*.yaml`, walking up only while the path still matches `tests/tasks/*`. I replayed that algorithm over this PR's rename list against both trees: 154 changed non-YAML files resolve to no task under the new layout, and 125 of them DID resolve under the old one — 58 in `tests/tasks/uipath-agents/_shared/`, 40 in `tests/tasks/uipath-maestro-flow/_shared/`, 29 in `tests/tasks/uipath-maestro-bpmn/_shared/`, 5 in `uipath-functions/_shared/`, plus the new family-level `_setup/` dirs (`uipath-ixp/_setup/`, `uipath-maestro-flow/_setup/`, `uipath-admin/_setup/`, `uipath-agents/coded/_setup/seeds/`). Verified the cause: `find tests/tasks/uipath-agents -maxdepth 1 -name '*.yaml'` → 0 (same for maestro-flow, maestro-bpmn, functions), so the up-walk finds nothing and exits. `INFRA_PATTERN` (smoke-skills.yml:152-156) matches only `tests/docker/`, `tests/fixtures/maestro-flow-sdk-smoke/`, `tests/experiments/smoke.yaml`, `tests/_shared/`, files directly in `tests/`, and the workflow itself — nothing under `tests/tasks/**` — so the infra fan-out does not compensate. Net effect for every future PR: editing only `tests/tasks/uipath-agents/_shared/check_bindings_sync.py` now hits the `[ -z "$SRC_SKILLS" ] && [ -z "$CHANGED_TASKS" ]` branch and prints "No skill or test changes detected — skipping smoke tests", where before the PR it ran `bindings_sync`. This PR's own smoke run is unaffected (it changes 642 task YAMLs, which resolve to themselves), which is exactly why the regression is invisible here. Fix: teach `resolve_task_yaml()` to map `<skill>/_shared/check_<name>.py` and `<skill>[/<group>]/_setup/*` to the tasks whose YAML references them (grep the command strings), or fall back to the skill's full smoke suite for those paths. _(trigger: tests/tasks/uipath-agents/_shared/check_bindings_sync.py)_
- 🟡 **CODEOWNERS was not updated, and 10 grader relocations silently drop their dedicated reviewers.** Resolving each rename's owning rule before and after (last-match-wins over the committed CODEOWNERS) shows two real ownership transfers — the three `testmanager .../seed.py` pairs git reports as cross-skill renames are rename-detection artifacts (each of the three `seed.py` files moved into its own sibling `_setup/`, verified against `git ls-tree origin/main`), and the planner one keeps identical owners. The real ones: (a) eight guardrail checkers move from `/tests/tasks/uipath-agents/coded/guardrails/` (CODEOWNERS:74 → @apetraru-uipath @valentinabojan @ctiliescuuipath, the guardrails owners) to `tests/tasks/uipath-agents/_shared/`, now matched only by CODEOWNERS:52 (@UiPath/team-coded-agents …) — `check_byog_middleware.py`, `check_escalation.py`, `check_intellectual_property_coded.py`, `check_llm_as_judge.py`, `check_pii_middleware.py`, `check_recommend_all.py`, `check_recommend_scoped.py`, `check_user_prompt_attacks_middleware.py`, all verified present at the new path; (b) `check_databricks_query.py` and `check_generic_dynamic_node.py` move from `/tests/tasks/uipath-maestro-flow/connector_features/` (CODEOWNERS:141, @chandusailella @mukundbayyaram @UiPath/maestro-cli-team) to `tests/tasks/uipath-maestro-flow/_shared/` (CODEOWNERS:103, @UiPath/maestro-cli-team only). Future edits to those graded assertions no longer require the owning team's review. `CLAUDE.md` rule 4 and `.claude/rules/pr-review.md` both require the CODEOWNERS edit in the same PR; add `/tests/tasks/uipath-agents/_shared/check_{byog,escalation,intellectual_property_coded,llm_as_judge,pii_middleware,recommend_*,user_prompt_attacks_middleware}*.py`-style entries, or keep guardrail checkers in a `_shared/guardrails/` subdirectory that can carry the rule. _(trigger: tests/tasks/uipath-agents/_shared/check_byog_middleware.py)_
- 🔵 **`tests/experiments/flow-v2-preview.yaml` still justifies its repo-wide docker bind mount with a grader path that no task uses any more.** Lines 26-31 mount `$SKILLS_REPO_PATH:$SKILLS_REPO_PATH:ro` with the comment "Narrowing plugins.path to preview/ drops the automatic repo-root mount, and criteria shell out to `$SKILLS_REPO_PATH/tests/tasks/**/_shared/*.py`." After this PR no criterion anywhere shells out that way: `grep -rn --include='*.yaml' 'command:.*SKILLS_REPO_PATH' tests/tasks/` matches only `uipath-troubleshoot/smoke-manifest-commands/task.yaml` (a pre_run, not a criterion, and not in this experiment's scope). The mount may still be needed for the first half of the rationale (the dropped plugin-root mount), but the stated grader reason is now false, and whether a docker-driver run can narrow the mount — the whole point of removing host-path deps — is unexamined. Same experiment file's `extra_mounts` note about version 0.11.2 stays accurate. Rewrite the comment, or narrow the mount and say so. _(trigger: tests/tasks/uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/e2e_03_project_creation_handoff.yaml)_

**Tests:**
- 🟠 **The new addressing contract ships with zero validation, in a repo that already has the gate to put it in.** The PR adds no tests (`git diff --name-status -M origin/main...HEAD -- tests/` has exactly one `A` line: `uipath-maestro-flow/ixp/e2e_03_project_creation_handoff/_setup/flow_project_finder.py`; no new `test_*.py` anywhere). Two existing mechanisms should have grown a check and did not: (1) `scripts/check-task-driver.py`, already run repo-wide on every PR by `.github/workflows/task-driver-gate.yml:53` as `python3 scripts/check-task-driver.py tests/tasks tests/experiments`, holds two task-YAML gates (`sandbox.driver: tempdir`, single-line experiment hooks) and is the natural home for a third: every `sandbox.template_sources[].path` resolves to an existing directory relative to the task YAML, and every `$REFERENCE_DIR/<path>` in a criterion/command resolves under the task's `reference.directory`. Both are pure static checks over the 787 task YAMLs that declare `template_sources` and the 847 that use `$REFERENCE_DIR`, and they would have caught the 17 review tasks pointing at a non-existent `../_setup`, the 5 ixp tasks pointing at the moved `../_shared/fixtures`, and `uipath-governance/compliance-pack/nonregression_branch_a_smoke.yaml`'s `_setup/cleanup_policy.py` — 23 tasks that currently abort at sandbox setup. (2) `/lint-task` (`.claude/commands/lint-task.md`), which `.claude/rules/test-writing.md` step 3 makes mandatory before committing a task, evaluates seven axes (self-report, prompt over-specification, meaningful coverage, wrong-reason pass, near-duplicate, validate-only flow, CLI verb reachability, pinned CLI, run-limit placement) — none of which looks at staged paths — so all 23 broken tasks lint clean. _(trigger: tests/tasks/uipath-review/agents/coded_review_error_handling/coded_review_error_handling.yaml)_
- 🟡 **The gap is wider than coded-apps: two of the subtrees this PR relocated helpers in have meta-tests that no CI job runs at all.** `.github/workflows/test-helpers.yml` runs pytest for exactly six paths (`tests/tasks/uipath-maestro-bpmn/`, `uipath-maestro-flow/`, `uipath-admin/_shared/`, `uipath-maestro-case/`, `uipath-planner/`, `uipath-agents/`) plus individual `tests/scripts/*` files; `grep -rn 'uipath-review\|uipath-coded-apps' .github/workflows/` returns nothing. `tests/tasks/uipath-review/` holds 3 meta-test files (41 tests, all passing locally — 20 renames landed in that subtree) and `tests/tasks/uipath-coded-apps/` holds 2 (`_shared/test_cleanup_codedapp_folder.py`, `dashboard/_shared/test_find_project_root.py`) that pytest reports as `no tests collected`. So the coded-apps regression already filed is the visible half; the review subtree's 41 assertions are equally unguarded against the next move. I ran all six CI jobs' paths at PR head: bpmn 19 passed, case 139 passed/14 skipped, planner 89 passed, agents 17 passed, flow 1234 passed with 2 failures traceable only to the untracked out-of-scope `tests/tasks/uipath-maestro-flow/outcome/outcome.yaml`, so no covered subtree regressed — the uncovered ones are where a repointing miss hides. Add `pytest tests/tasks/uipath-review/` alongside the coded-apps job. _(trigger: tests/tasks/uipath-coded-apps/_setup/cleanup_codedapp_folder.py)_ _(restates: Axis 4: Moving cleanup_codedapp_folder.py to _setup/ broke the allowlist negative control, and no CI job runs it)_

**Rollout impact:**
- 🟡 **`tests/scripts/stage_shared.sh` — the CI-only workaround for exactly the addressing this PR replaces — was not revisited, and the PR does not say whether it still has to run.** Its header states its purpose verbatim: coder-eval 0.11 stages `$TASK_DIR` as an isolated copy of the leaf task folder, so ~215 graders that import the group-level `_shared` package by walking up out of the task dir break, and the script copies `<group>/_shared/` INTO each such task dir on the CI checkout. It is still invoked at `.github/workflows/run-coder-eval.yml:403` and `.github/workflows/smoke-skills.yml:462`, and still mutates the checked-out tree before every run. This PR moves the graders themselves into `<skill>/_shared/` and reaches them through `$REFERENCE_DIR`, whose staged root (`reference.directory: ../..` on 224 tasks) already contains `_shared/` — so the workaround's premise is largely gone: its pass-1 target set drops from 221 task directories on origin/main to 65 at PR head (`git grep -lE 'from _shared|import _shared'` over `tests/tasks/**/*.py`, excluding files inside a `_shared/` dir). The 65 survivors still depend on it, which is the unstated part: `tests/Makefile` never calls it (only the two workflows do), so those tasks' graders resolve `_shared` in CI and not in a local `make all`/`make smoke`. Either retire the script now that graders can `sys.path.insert` the reference root, or state in the PR that it remains load-bearing for the remaining 65 and why `tests/Makefile` does not run it. Pass 2 is not a leak risk — I checked its selector (`git grep -lE 'sys\.path\.insert\(.*_shared'` minus `from _shared` importers): 24 files, none under a `_setup/` directory, so it never copies grader modules into a `template_sources`-staged dir. _(trigger: tests/tasks/uipath-agents/_shared/check_bindings_sync.py)_
- 🔵 **Adding `reference.directory` to 376 tasks widens what a per-run reference staging copies and what the answer-key digest covers, and the PR states neither.** `Orchestrator._stage_reference` (coder_eval `orchestrator.py:1511-1548`) does a full `shutil.copytree` of the resolved directory into a per-run temp dir, then `digest_tree`s it and persists the hash as `environment_info["reference_digest"]`. With `directory: ../..` on 224 tasks that copy is now the whole skill test subtree per task run — `du -sh` gives 5.7 MB for `tests/tasks/uipath-maestro-flow`, 3.6 MB for `uipath-agents`, 2.4 MB for `uipath-maestro-case` — multiplied by tasks × replicates in a nightly fan-out. The second-order effect is the digest: `verify_reference_unchanged` (coder_eval `orchestration/regrade.py`) raises `RegradeError` "The reference directory … changed since this run was executed (digest mismatch)" on any difference, and before this PR these tasks declared no `reference:` at all, so the guard returned early and never fired. Post-PR, a detached re-grade of a stored run fails after ANY unrelated edit anywhere in that skill's subtree — a tree this PR alone modifies 642 YAMLs in. Latent, not live: `grep -rn regrade` across this repo's workflows, Makefile, scripts and docs returns zero, so nothing here invokes detached re-grading today. Worth one line in the PR body (and in the `tests/README.md` section the Axis 8 finding asks for) so the first person to try a re-grade is not surprised, and worth asking whether `../..` is needed where `..` or `.` would do. _(trigger: tests/tasks/uipath-agents/coded/bindings_sync/bindings_sync.yaml)_

**Docs & config:**
- 🟡 **Beyond `tests/README.md`, the task SCAFFOLDER still emits the removed pattern — so the next generated task reintroduces it.** `.claude/rules/test-writing.md` step 2 makes `/generate-task` the mandated way to author a task, and `.claude/commands/generate-task.md` still teaches the old addressing at two places: line 183 "If importing from `tests/tasks/<skill>/_shared/`, use the `sys.path.insert(...)` bootstrap that existing scripts use (the script is invoked as `python3 $TASK_DIR/check_<name>.py`)" and line 187 the literal scaffold `command: "python3 $TASK_DIR/check_<name>.py"`. Neither `_setup/`, `template_sources`, `mount_point`, nor `$REFERENCE_DIR` appears anywhere in that file. `tests/templates/test-task-template.yaml` — required by `.claude/rules/test-writing.md` Must-Do 6 as the base for new tasks — contains no `pre_run`, no `run_command` criterion and no `sandbox.template_sources` example at all, so it neither teaches nor contradicts the new shape. The Axis 8 finding's recommendation rewrites `tests/README.md` §Shape and line 155 but does not touch either of these two files, and they are the ones an agent actually copies from. Update line 183/187 to the `_setup/` + `$REFERENCE_DIR` form in the same pass, and add the `run_command` + `reference:` + `template_sources` skeleton to the template. _(trigger: tests/tasks/uipath-platform/_setup/seed.py)_ _(restates: Axis 8: tests/README.md still documents the deleted $SKILLS_REPO_PATH/.../seed.py addressing style; the replacement _setup/ + template_sources + $REFERENCE_DIR contract is documented nowhere)_
- 🔵 **A subtree README now mis-describes what `_shared/` holds.** `tests/tasks/uipath-maestro-bpmn/README.md:37` still reads "`_shared/` contains small Python helpers for durable XML shape assertions" — true on origin/main, where `_shared/` held `bpmn_assertions.py` / `bpmn_check.py`. This PR moved 29 per-task grader scripts into it (`edit/add_node/check_add_node.py` → `_shared/check_add_node.py`, `authoring/check_api_workflow_task.py` → `_shared/check_api_workflow_task.py`, and so on; `ls tests/tasks/uipath-maestro-bpmn/_shared` now lists `check_*.py` alongside the two helpers), and there is no `_setup/` in that subtree at all, so the one sentence a contributor reads about the layout no longer describes it. Same pass should confirm the sibling subtree docs: `tests/tasks/uipath-mcp-servers/README.md` and `tests/tasks/activation/README.md` only mention `_shared/` in unrelated mock-dispatcher/env contexts and are unaffected. _(trigger: tests/tasks/uipath-maestro-bpmn/_shared/check_add_node.py)_

**Parallel paths & mirrors:**
- 🟡 **`tests/tasks/uipath-troubleshoot/` — 298 task YAMLs — was left on the old convention, and one of them still resolves a pre_run script through `$SKILLS_REPO_PATH`.** This PR touched exactly one file in that subtree (`activity-packages/uia-ambiguous-node-test/task.yaml`), so the repo now runs two addressing conventions side by side with nothing recording which is canonical. Concretely: `tests/tasks/uipath-troubleshoot/smoke-manifest-commands/task.yaml:68-77` is the only `command:` line left in any task YAML that expands `$SKILLS_REPO_PATH` (verified: `grep -rn --include='*.yaml' 'command:.*SKILLS_REPO_PATH' tests/tasks/` returns this file alone, zero matches elsewhere), running `python -c "…os.environ['SKILLS_REPO_PATH'] … 'uipath-troubleshoot','_shared','scripts','verify_manifest_commands.py' …"` with `fail_on_error: true`. Its comment argues the case the rest of the PR just refuted — "coder-eval does NOT inject TASK_DIR into pre_run env (TASK_DIR is criterion-only). SKILLS_REPO_PATH is already the repo's convention for locating the plugin source" — when `_setup/` staging via `template_sources` is precisely the mechanism that solves pre_run's lack of `$TASK_DIR`. The subtree's own 400-line authoring guide `tests/tasks/uipath-troubleshoot/CLAUDE.md` also still teaches `$TASK_DIR` bind-mount semantics (line 261) and `_shared/scripts/...` host paths (lines 39, 67, 86, 391-393) with no mention of `_setup/`. Either migrate that one pre_run and note the mock-machinery exception, or state in the PR why the troubleshoot subtree is out of scope. _(trigger: tests/tasks/uipath-troubleshoot/activity-packages/uia-ambiguous-node-test/task.yaml)_
- 🟡 **`coded_in_flow_e2e.yaml` got `reference.directory: ..` while 32 of its 39 siblings in the same folder got `../..`, and its grader needs the deeper root.** The PR rewrote the criterion from `python3 $TASK_DIR/check_coded_in_flow_e2e.py` to `python3 $REFERENCE_DIR/coded_in_flow_e2e/check_coded_in_flow_e2e.py` and added `reference:\n  directory: ..` (lines 19-20). Tally across `tests/tasks/uipath-agents/coded/*/*.yaml`: 32 × `../..`, 6 × `.`, 1 × `..` (this file), 1 with no reference. The grader at line 28-29 does `sys.path.insert(0, dirname(dirname(dirname(abspath(__file__)))))` then `from _shared.ast_lazy_init_check import find_module_level_llm_clients`, and `ast_lazy_init_check.py` lives at `tests/tasks/uipath-agents/_shared/` — one level ABOVE the reference root this task chose. coder-eval stages the reference as a private `shutil.copytree` of `reference.directory` only (`coder_eval/orchestration/evaluation.py:112-128 stage_reference_dir`, called from `Orchestrator._stage_reference`), so the staged tree is rooted at the copy of `coded/` and the up-walk lands outside it; the import resolves only because `tests/scripts/stage_shared.sh` co-locates a `_shared/` copy inside the task dir on the CI checkout. This is the single case out of 114 `$REFERENCE_DIR`-invoked graders importing the `_shared` package where the resolved reference root contains no `_shared/` directory (I checked all of them). Set `directory: ../..` to match the siblings and the reference tree supplies `_shared/` on its own. _(trigger: tests/tasks/uipath-agents/coded/coded_in_flow_e2e/coded_in_flow_e2e.yaml)_

## Guardrails & Automation

**Static checks:**
- NOTE ON RUBRIC ORDER — buckets 1 and 2 are almost entirely unavailable here, verified: the repo configures no ruff/flake8/mypy/pyright/eslint/tsc/markdownlint (no config file, no CI step; the PR body's pyflakes run is ad hoc, and pyflakes/ruff resolve no imports, so none of these findings are … _Prevents:_ The `reference.directory` x `llm_judge include_reference: true` / `reference_comparison` trap now armed on 376 tasks (axis-3 low finding), …
- Add ONE new multi-gate checker, `scripts/check-task-staging.py`, following `scripts/check-task-driver.py` exactly (module docstring stating each gate plus its rationale and the incident it replays, human-readable findings plus `::error file=<repo-rel>,line=<n>::` annotations, `OK — N ...` on … _Prevents:_ Both critical staging findings in full: the 17 `uipath-review/agents/*` tasks pinning `path: ../_setup` one level short of …
- `check-task-staging.py` GATE 2 — every sandbox-relative script path named in a `pre_run`/`post_run` command must resolve through that task's own mounts. Implementation: build `{mount_point -> [source dirs]}` from the task's `template_dir` entries (last-wins overlay, `mount_point` default `"."`), … _Prevents:_ The governance regression — `compliance-pack/nonregression_branch_a_smoke.yaml:31` (pre_run) and `:35` (post_run) invoke `python3 …
- `check-task-staging.py` GATE 3 — sandbox-layout-aware local-import resolution for every hook entry point and its transitive local imports. For each hook-invoked staged script, compute the module names available AT ITS MOUNT DIRECTORY (union of `*.py` stems across every source overlaid onto that … _Prevents:_ The `jira_is` half of the maestro-flow Jira finding (10 importers across 5 tasks: `escalation_jira_ticket/_setup/seed.py:16`, …
- `check-task-staging.py` GATE 4 — the invisible-side mirror of Gate 1: every `$REFERENCE_DIR/<path>` occurrence in a task YAML must resolve under that task's `reference.directory`, and a task that uses `$REFERENCE_DIR` at all must declare one. Measured: 1132 references, 0 dangling today — a pure … _Prevents:_ Recurrence of the Gate-1/Gate-2 class on the grading side, where it is strictly worse: a dangling `$REFERENCE_DIR` script does not abort …
- `check-task-staging.py` GATE 5 — local-import resolution for NON-staged scripts (graders reached via `$REFERENCE_DIR`, whose runtime layout equals the on-disk layout). Resolve each file's real search path by evaluating its `sys.path.insert(0, ...)` arguments with a bounded expression evaluator — it … _Prevents:_ The grading half of the Jira finding — `_shared/check_escalation_jira_ticket.py:40`, `check_jira_create_issue.py:34`, …
- `check-task-staging.py` GATE 6 — `__file__`-relative sibling resources in non-staged scripts must exist on disk. Two literal shapes, both already this repo's idiom: `os.path.join(os.path.dirname(os.path.abspath(__file__)), "a", "b")` and `Path(__file__).resolve().parent[.parent...] / "a" / "b"`. … _Prevents:_ The heaviest-weighted grading regressions in the PR: `_shared/check_cm_golden_seed.py:16`, `check_cm_golden_bindings.py:40` and …
- `check-task-staging.py` GATE 7 — enforce the `_setup` (agent-visible) / `_shared` (grading, invisible) separation the PR itself states in commit d61fa9d52 and in `_setup/flow_project_finder.py:3-9`, as two crisp structural rules rather than a judgement call. (7a) A `.py` executed via … _Prevents:_ 7a catches the e2e_03/e2e_04 handoff leak: the fused 1057-line `ixp/e2e_03_project_creation_handoff/_setup/handoff.py` carries `check_main` …
- `check-task-staging.py` GATE 8 — answer-key token intersection. For each task, collect the identifier-shaped string constants its graders assert on (AST `ast.Constant` strings in the `.py` files named after `$REFERENCE_DIR`, filtered to `^[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+$`, length >= 10) and fail when … _Prevents:_ The `_setup`-staging answer-key leak across the whole uipath-review judgment suite: 16 tasks whose staged injector names the exact …
- `check-task-staging.py` GATE 9 — a staged script must not address anything outside its own mount. Over the AST (never the raw text, so docstrings are exempt) of every `.py` reachable through `template_sources`: fail on `Path(...).parents[N]` with N >= 2 or a `.parent` chain walking above the mount … _Prevents:_ `uipath-agents/coded/_setup/seed_coded_order.py:68`, which kept `parents[3]` after being staged: the helper lookup that used to succeed now …
- Extend the existing verb gate to helper Python rather than adding a new mechanism: `scripts/check-skill-verbs.py` scans only `.md` under `skills/`, and `check-cli-verbs.py` only task `command_pattern`s. Add a mode (or a sibling reusing the same catalog loader and the `<!-- uip-check-skip -->` … _Prevents:_ The fallback branch at `uipath-agents/coded/_setup/seed_coded_order.py:88`: `run_uip("df", "choice-sets", "add-value", ...)` — no such verb …
- Add `tests/scripts/test_task_helper_ci_coverage.py`, modelled on `tests/scripts/test_required_checks_contract.py` (which already parses workflow YAML), run as a job in `test-helpers.yml`, asserting two things. (1) Reachability: every directory under `tests/tasks/` containing a `test_*.py` is named … _Prevents:_ The 'no CI job runs it' half of the coded-apps finding — `_shared/test_cleanup_codedapp_folder.py` is both broken by the move AND …
- Extend repo-path validation to the `tests/` docs, following `scripts/check-skill-links.mjs` (scoped to `skills/` + `skill-flavors/` and only checking Markdown links, so it would miss this even if rescoped). Either add `tests` to its roots and teach it inline-code path spans, or add a small sibling … _Prevents:_ The stale addressing convention at `tests/README.md:302` and `:306`, which still shows `python3 …

**Guardrail improvements:**
- Add a PR-time sandbox materialization + hook dry-run job: for each task YAML changed in the PR, build the sandbox the way coder-eval does (copy each `template_dir` source into a temp dir at its `mount_point`, last-wins), then execute that task's `pre_run`/`post_run` commands with a stubbed `uip` on … _Prevents:_ The whole staging class end-to-end (17 review tasks, 5 ixp tasks, 5 Jira tasks, the governance cleanup, `seed_coded_order.py`), including …
- Guard the 154 lines that `ixp/e2e_03_project_creation_handoff/_setup/flow_project_finder.py` copies out of `_shared/flow_check.py`: parametrize the 19 existing discovery cases in `_shared/test_flow_check_discovery.py` over BOTH implementations (load the copy by path with `importlib` — there is no … _Prevents:_ Silent divergence between the sandbox-side project finder and the grading-side one, where grader and checkers would resolve different …
- Add a negative control for `lowcode_review_with_stale_generated_files`: feed `check_stale_generated_files.py` a synthesized report whose canary occurrence is traceable to harness scaffolding rather than to the project under review, and assert the checker does NOT fail — then scope the assertion in … _Prevents:_ The false-fail this PR introduced in the opposite direction from every other leak: identical agent behaviour now scores differently than on …
- Add the two missing `pytest` jobs to `test-helpers.yml` (`pytest tests/tasks/uipath-coded-apps/ -v` and `pytest tests/tasks/uipath-review/ -v`, each with its `docs/REQUIRED-CHECKS.md` row if made required), and convert the two non-collectible files into real `test_*` functions: … _Prevents:_ Recurrence of an allowlist negative control on a destructive delete path being silently vacuous. Paired with the static checks, this is …
- Make the new convention teachable at the point of authoring, since `.claude/rules/test-writing.md` rule 6 mandates the template as the base for every new task: add the canonical two-source `_setup` staging block (`path: _setup` + `path: ../_setup`, both `mount_point: _setup`), a `reference: … _Prevents:_ The half of the ixp finding no gate can reach — a bare `path: ../_setup/fixtures` would satisfy Gate 1 while mounting the invoices at the …
- Tighten the passing-run convention for staging moves specifically: extend `lint-tasks.yml`'s existing Step 3 so that when a PR renames or relocates files under a `_setup/`, `_shared/` or `fixtures/` directory, the honour-system claim must name each affected skill family (e.g. 'ran one … _Prevents:_ The review-level miss that let a 642-file, 349-rename migration land on evidence (YAML parses, Python compiles, `_shared` pytest green) …

## Top 5 Priority Actions

1. Stop staging the answer keys: keep the 17 uipath-review injectors off the agent filesystem (or strip every grader-shared token out of them), starting at tests/tasks/uipath-review/agents/lowcode_review_with_stale_generated_files/lowcode_review_with_stale_generated_files.yaml:13 — and scope check_stale_generated_files.py:104-106 to the project under review, since both canaries now sit in the staged _setup/setup_fixture.py:177,191 and fail agents for reasonable behavior.
2. Move the grading oracle back out of the six api-workflow eval sandboxes — return evals/_setup/eval_scoring.py to evals/_shared/ (staged at tests/tasks/uipath-api-workflow/evals/gate_create/gate_create.yaml:30, imported by no seed, only by graders) and move the planted-bug rationale out of loop_until_green/_setup/seed.py:2-9 and stale_expectations/_setup/seed.py:2-5 into the host-side task description.
3. Fix the silently flipped seed branch at tests/tasks/uipath-agents/coded/_setup/seed_coded_order.py:68 — `parents[3]` no longer resolves from the sandbox, so the only remaining path calls `df choice-sets add-value`, a verb absent from assets/uip-catalog-snapshot.json, mis-seeding the entity that a weight-6.0 criterion grades; point it at the staged sibling `parent / "seed_choice_set.py"`.
4. Restore fixture resolution for the four relocated graders whose `fixtures/` did not move — tests/tasks/uipath-maestro-case/_shared/check_cm_golden_seed.py:16 (plus check_cm_golden_bindings.py:40, check_cm_golden_semantics.py:30) and tests/tasks/uipath-agents/_shared/check_is_smoke.py:260 — resolving via the reference root, never the agent-writable sandbox copy, to unblock 10.0 weight in cm_golden_expense.yaml and the 6.0 primary criterion in is_smoke.yaml.
5. Repair the four dangling staged paths that ERROR before the agent starts — `../_setup` → `../../_setup` in all 17 review YAMLs (tests/tasks/uipath-review/agents/coded_review_error_handling/coded_review_error_handling.yaml:25), `../_shared/fixtures` → `../_setup/fixtures` with `mount_point: fixtures` in the 5 ixp tasks (tests/tasks/uipath-ixp/e2e/full_lifecycle.yaml:30), `jira_is.py` staged into each of the 5 flow tasks' `_setup/` plus a union copy in `_shared/` (tests/tasks/uipath-maestro-flow/e2e/jira_search_triage/jira_search_triage.yaml:16), and `cleanup_policy.py` into a mounted `_setup/` (tests/tasks/uipath-governance/compliance-pack/nonregression_branch_a_smoke.yaml:31) — then add the two statically-checkable CI guards this class needs (every `template_dir` path resolves; no staged file contains a literal its `check_*.py` asserts on), repoint the orphaned allowlist negative control at tests/tasks/uipath-coded-apps/_shared/test_cleanup_codedapp_folder.py:17 with a coded-apps pytest job, and rewrite tests/README.md:155,302,306 to document the new `_setup`/`_shared`/`$REFERENCE_DIR` contract.

---

**Change class:** complex — rewires how every coder-eval task locates its setup scripts, grading scripts and fixtures (host env vars -> sandbox.template_sources + $REFERENCE_DIR), moves ~350 Python files between _shared/ and _setup/, and repoints every importer; a wrong relative path silently becomes a hard sandbox-setup failure


**Stats:** 3 🔴 · 5 🟠 · 4 🟡 · 3 🔵 across 8 axes reviewed.
**Verification:** 39 medium+ finding(s) adversarially re-checked · 10 dropped as false positives (26%) · 28 corrected in place · 7 low passed through unverified.

<sub>⚠️ This comment was abridged to fit the 65,536-character comment limit: nits and guardrails are condensed to one line each and medium findings are truncated. The unabridged per-axis reports were generated alongside it — ask the reviewer for the full report if you want the detail behind any item.</sub>

tests/tasks/uipath-agents/lowcode/external_apiworkflow_tool/external_apiworkflow_tool.yaml
akshaylive: Checked — this is safe here. `reference.directory` is resolved relative to the task YAML's own directory (per `ReferenceSource` in coder_eval), and this leaf task has no shared `_setup`/`_shared` helper: `check_external_apiworkflow_tool.py` sits right next to the YAML with zero imports of sibling modules (just `json`/`os`/`sys`/`pathlib`), so `directory: .` stages exactly what the checker needs and nothing more.

This is different from the uipath-admin/uipath-governance families (fixed in 28e70ca83), where checker scripts imported a helper module one level up (`_setup/admin_helpers.py`) — there, `directory: .` was too narrow and broke the import. No such helper exists for this lowcode task, so no change needed.

uipreliga:
## Architectural follow-up: the staging half is a workaround for a missing env var

Follow-up to the automated 8-axis review above — **not blocking, and not a request to re-architect under a red PR.** The blockers there are worth fixing as-is. This is about the shape that produced them, because the measurements point at one root cause rather than four independent mistakes.

### The asymmetry that explains which half broke

What the framework actually gives each phase (read from coder-eval **0.11.0** — closest readable source to the 0.12.1 pin in `tests/.coder-eval-version`; the 0.12.0 entry in the local uv cache is an editable stub with no source, so a later change to this contract is unverified):

| Phase | Directory env vars available |
|---|---|
| `run_command` (criteria) | `TASK_DIR`, `REFERENCE_DIR` — both set in `_build_run_command_env` (`sandbox.py:1010-1069`) |
| `pre_run` / `post_run` | **none** |

That asymmetry explains the whole PR, including which half broke.

The grading half needed only a one-token swap (`$TASK_DIR` → `$REFERENCE_DIR`) and it is genuinely clean: **0 dangling `$REFERENCE_DIR/<script>` references, 329/329 `llm_judge` criteria carrying `include_reference` explicitly.**

**Every blocker is in the other half** — the `template_sources` staging that exists *only* because `pre_run` has no equivalent env var. The clean half used a variable; the broken half had to simulate one by copying directories and hand-typing how deep they sit.

### What simulating it costs

Measured across the whole `tests/tasks` corpus at this PR's head:

| | |
|---|---|
| `template_dir` entries | **1501** → only **743** distinct target directories (≈half are repeats of a family-wide fact) |
| entries hand-encoding `../` depth | **680** (232 at depth 1, 340 at depth 2, 45 at 3, 60 at 5) |
| `reference.directory` values | **847** tasks, but 4 values (`reference`, `../..`, `..`, `.`) cover **819** of them |

The 17-task blocker is a depth-1-vs-depth-2 confusion. That class of bug exists only because a human types the depth.

### Can this be declared once per family? No — worth knowing before anyone tries

Config layers are exactly `default → experiment-defaults → task → variant → cli` (`orchestration/config_merge.py:46-52`). There is no per-directory layer, and `experiment-defaults` is per-slice — declaring `uipath-review/_setup` there would stage it into every task in the slice, not just that family's. So the per-task block is unavoidable **as the wire format**, which is what makes lever 2 the decisive one.

### Three levers, ranked

**1. Give `pre_run` the env var it lacks — upstream, and the only genuinely fundamental fix.** Export `TASK_DIR` (or a `SETUP_DIR`) into the pre_run/post_run env and most of this machinery becomes unnecessary: setup scripts stay on the host and are *addressed* rather than *copied*. It also removes the answer-key exposure **by construction** — nothing lands in the agent's filesystem, so no injector or scoring oracle can leak. Looks actionable rather than wishful: `run-coder-eval.yml:108-114` calls `coder-eval` a "first-party wheel" and exempts it from the safe-chain age gate, and UiPath owns the `coder_eval_uipath` plugin repo. Staging would shrink to its legitimate use — things the agent genuinely must see (fixtures, mocks, and the product skill directory, staged 62×).

**2. Make these fields generated, not authored — in-repo, best value available today.** Since the task YAML must carry the block, stop hand-writing it. A `scripts/sync-task-staging.py --write` deriving each task's `template_sources` and `reference.directory` from its own `pre_run` imports plus the nearest ancestor `_setup/` would make the 22 dangling `template_dir` paths **unrepresentable** — a generator never miscounts `../`. It would also catch the `jira_is.py` and `_setup/cleanup_policy.py` cases loudly at generation time, since it cannot stage a module that was never moved anywhere. This is an established idiom in this repo, not a new concept to adopt: `check-skill-status.py --write-readme` and `check-skills-sh.py --fix` already derive-and-check, with CI enforcing currency.

**3. Cut on audience, not location.** `_setup/` vs `_shared/` is a *directory* convention standing in for a *file* property, and it fails three ways visible in this diff: a helper needed by both sides has nowhere to live (family helpers had to be physically moved out of `_shared/` and every importer repointed); 30 `template_dir` entries still stage a `_shared`-named directory into an agent sandbox; and directory-granularity staging cannot express "the agent may see this file but not that one" — which is exactly how 16 `inject_*.py` scripts ended up shipping the `rule_id` their own grader greps for. `template_sources` is also doing three unrelated jobs under one convention (setup tooling, fixtures/mocks, the skill under test). The cleaner cut: **one generated per-task payload directory whose entire contents are by definition agent-visible.** Then "may the agent see X?" is answered by `ls`, and "why is an injector in the payload?" becomes a visible review question instead of an invisible convention violation.

Plus the static gate from the review, whichever lever you pick — throwaway versions of the path-resolution and `$REFERENCE_DIR` checks ran over all 8409 tracked files in seconds, so enforcement is nearly free.

### Suggested sequencing

Ship the blocker fixes as they are, then **lever 2 immediately** — self-contained, reuses a pattern this repo already enforces, and retires the entire class of bug found here. Raise **lever 1** with whoever owns the `coder-eval` wheel; it is the difference between generating a workaround correctly and not needing one. **Lever 3** only makes sense bundled with 1 or 2, since all three regenerate the same fields.

tests/tasks/uipath-agents/lowcode/external_apiworkflow_tool/external_apiworkflow_tool.yaml
akshaylive: Ignore the above -- I guess you're talking about default being `.`. I would caution against that for the timebeing. llm_judges include all references. If a folder is specified as a reference, the entire directory gets dumped into llm judge context -- we need to do this in phases but I agree with the long term objective.

CarlesUIPath:
Ran an ADO job on a subset of 200 tasks from the changed tasks on Luna, representative by the affected skills. So far the above changes do not result in load/path/setup errors. See below for a breakdown.

| Outcome | Count | PR-attributable? |
|---|---|---|
| SUCCESS | 178 | — |
| FAILURE (quality) | 17 | ❌ normal model misses (scores 0.23–0.9) |
| ERROR | 1 | ❌ environment (tenant couldn't create an external app) |
| Skipped | 4 | ❌ pre-existing `skip: true` |
| **Load / path / setup errors from the PR** | **0** | ✅ |

uipreliga (approved):
Fix what you agree with and 🚢

PR #3272, opened 2026-09-13
fix(uipath-maestro-flow): stop script nodes defaulting upstream output to `{}`

bryan-uipath:
## Motivation

Coding agents building Flow script nodes reach for a defensive default:

```js
const ctx = $vars.script1.output || {};
ctx.body; // Property 'body' does not exist on type '{}'
```

The canvas typechecks script bodies with TypeScript (`checkJs`, no `strictNullChecks`) and declares `$vars.<node>.output` as the inferred return type `T`. `|| {}` widens it to `T | {}`, so every property read warns on the node and in the Studio Issues panel. 

The problem: __Well-formed agent-generated flows end up covered in warnings (MST-15100, [Slack thread](https://uipath-product.slack.com/archives/C0AE5U60686/p1789254300913829?thread_ts=1787759343.972729&cid=C0AE5U60686)).__

## Changes

- `author/plugins/script/impl.md`: add script rule 9 — never default an upstream output to `{}`; read directly, or guard the individual read with `$vars.x?.output?.field ?? default`.
- Same file, debug table: drop the `(((a || {}).b || {}).c || [])` example from the `[300501]` row so the doc stops showcasing the anti-pattern.
- `shared/variables-and-expressions.md`: list optional chaining / nullish coalescing under supported Jint syntax (the IXP example already relies on `?.`).

Deliberately not touched: CAPABILITY.md (already links to script/impl.md), the IXP `|| []` example (arrays do not trigger the warning), uipath-review checklists (separate skill).


nikhil-maryala (approved):

skills/uipath-maestro-flow/references/author/plugins/script/impl.md
```diff
@@ -40,6 +40,7 @@ See [Action Node Structure — Adding and editing procedures](../../../shared/ac
 6. **No external calls** — use the HTTP node or a connector node for API calls.
 7. **30-second timeout** — long-running computations will be killed.
 8. **Never name a variable `aggregate`** — reserved host global. On any `Identifier 'X' has already been declared`, rename `X`.
+9. **Never default an upstream output to `{}`.** `const ctx = $vars.x.output || {}` widens the type to `T | {}`, so every `ctx.field` read warns `Property 'field' does not exist on type '{}'`. Read the output directly; if the upstream node may not have run, guard the individual read: `const body = $vars.x?.output?.body ?? '';`
```
rockymadden, line 43: 🔴 **major** — The fix example reproduces the warning it bans.

`??` unions the default's type exactly like `||` does. `$vars.x?.output?.body ?? ''` types `body` as `T | string`, so the next read warns `Property 'items' does not exist on type 'string'` — same class, different member.

And `body` is an object twice in this file already: line 22 (`$vars.fetchData.output.body.items`) and line 72 (`data: $vars.httpCall.output.body`). The example's own variable is the failing case.

Default at the leaf, same-typed:

```js
const items = $vars.x?.output?.body?.items ?? [];
```

Suggested rule tail: *"Read the output directly; if the upstream node may not have run, chain `?.` to the leaf and `??` only a scalar or array leaf, never an intermediate object."*

Agents copy examples verbatim. This is the highest-leverage line in the diff.

skills/uipath-maestro-flow/references/author/plugins/script/impl.md
```diff
@@ -93,5 +94,5 @@ Property access is **case-sensitive** — these casings resolve: `.FullName`, `.
 | Timeout after 30s | Script too expensive | Simplify logic or split into multiple scripts |
 | `console is not defined` | Used `console.log()` | Remove — use `return { debug: val }` instead |
 | `fetch is not defined` | Tried to make HTTP call | Use an HTTP node or connector node instead |
-| `[300501] Error invoking script task` with `Unexpected token` | Malformed JavaScript — `flow validate` does not parse script bodies, so syntax errors surface only at runtime. Common slip: one missing `)` in chained `(((a \|\| {}).b \|\| {}).c \|\| [])` guards | Balance parentheses; check with `node -e "new Function(<script>)"` before validate |
+| `[300501] Error invoking script task` with `Unexpected token` | Malformed JavaScript — `flow validate` does not parse script bodies, so syntax errors surface only at runtime | Balance parentheses; check with `node -e "new Function(<script>)"` before validate |
```
rockymadden, line 97: 🟡 **minor** — Cause and Fix no longer line up.

Removing the chained-guard example took the only antecedent for "Balance parentheses" — the Cause is now generic malformed JS, the Fix still points at one specific slip.

```
| Re-check the script body's syntax — `flow validate` never parses it; verify with `node -e "new Function(<script>)"` before validate |
```

skills/uipath-maestro-flow/references/shared/variables-and-expressions.md
```diff
@@ -566,6 +566,7 @@ The production runtime uses **Jint** (a .NET JavaScript interpreter, ES2020 subs
 - Destructuring: `const { a, b } = obj` (in script bodies)
 - Spread operator: `[...arr1, ...arr2]` (in script bodies)
 - Arrow functions: `items.filter(x => x.active)` (inline callbacks)
+- Optional chaining / nullish coalescing: `a?.b?.[0]`, `x ?? fallback`
```
rockymadden, line 569: 🟡 **minor** — Missing the scope qualifier its siblings carry.

Destructuring, spread, and arrow functions all say `(in script bodies)`. This entry doesn't, so it reads as also valid in `=js:` conditions, `variableUpdates`, and end-node mappings.

I could not confirm that. Every `?.` / `??` in the repo sits in a script body (`BillingDisputeResolution.reference.flow` → `dataTransform`, `discrepancyScript`); zero occurrences in any expression site across all 8 `.flow` files. The description's evidence — "the IXP example already relies on `?.`" — is prose citing prose.

Fix: append `(in script bodies)`, or say in the description how the `=js:` sites were verified, including that `flow validate`'s expression linter accepts them.

rockymadden (approved):
## What

Doc-only change to `uipath-maestro-flow` (2 files, +3/-1):

- **`script/impl.md`** — new script rule 9: never `|| {}` an upstream output; read directly or guard with `?.` / `??`.
- **`script/impl.md`** — drops the `(((a || {}).b || {}).c || [])` example from the `[300501]` debug row.
- **`variables-and-expressions.md`** — lists `?.` / `??` under supported Jint syntax.

Motivation checks out. `|| {}` widens to `T | {}` and every subsequent property read warns. The premise is real.

---

## 🧭 Overall findings

**🟡 minor — no regression check.** The rule is mechanically greppable (`|| {}` in `nodes[].inputs.script`) and the harness already exists (`tests/tasks/uipath-maestro-flow/_shared/`). Nothing stops rule 9 from regressing.

**🟡 minor — the repo's own reference flows are the anti-pattern.** That check would fail today on:

- `BillingDisputeResolution.reference.flow` → `discrepancyScript`, 4× (e.g. `$vars.start?.output?.webhookPayload?.dispute || {}`)
- `BillingDisputeResolution.native.reference.flow` → `normalizeData1`, `discrepancyDetection1` — the latter carrying literally the `(((a || {}).b || {}).c || [])` shape this PR deletes from the docs

Test-only, never read by the agent at runtime. But `_shared/test_same_ground_advisories.py:104` pins these as known-good. Fix here or file the follow-up.

**🟡 minor — scope claim unverified.** Rule 9 lands only in the script plugin doc. The same widening would hit any typechecked expression site. "Script bodies only" is reasoned, not confirmed — the canvas `checkJs` / `strictNullChecks` config isn't in this repo.

**🤖 minor — the automated review comment is slop.** "No issues found — approve", a checkbox list, an area-ratings table where every row says OK, and this PR description's scope-outs recited back as if verified. It raises one "issue for manual review", then argues itself out of it. It missed the `??` defect in the line it praised. Worth not leaning on it.

---

**Confirmed from the repo:** rule 9 is the only `|| {}` left in the skill · `author/CAPABILITY.md:102` does route to `script/impl.md` · `?.` / `??` appear only in script bodies · the reference flows carry the anti-pattern.

**Reasoned, not confirmed:** TS union semantics for `||` and `??` (standard TS, but the canvas tsconfig isn't in this repo) · that `?.` / `??` execute under Jint at non-script expression sites.

---

## ✅ Recommendation

**Approve with comments.** Doc-only, CI green, the diagnosis is right. The rule 9 example is the one thing worth fixing before or straight after merge — everything else is a follow-up.

---

**tl;dr** — Right diagnosis. One inline comment is a major: the replacement example reproduces the warning the rule exists to remove. Two minors on the other changed lines, plus the follow-ups above.

PR #3283, opened 2026-09-14
docs(uipath-maestro-case): seven corrections from the terra golden loop

roberts-cliff:
Seven corrections to the case skill found by running the golden build repeatedly against terra and reading what the agent actually did. Docs-only — no task YAMLs, no scripts, no schema.

## What changed

| Commit | Correction |
|---|---|
| `d9918cd1b` | Map three cli#4160 codes the skill never mentioned: `CASE_MGMT_OUTPUT_ORIGINAL_VAR_MISSING`, `STRICT_SDD_SLA_CONDITIONAL_MISSING`, `CASE_MGMT_SLA_RULE_UNREACHABLE`. |
| `f6c7749db` | A mid-build strict run fails on stages that aren't built yet — that is not a stop signal. Count only findings on built stages when judging progress. |
| `0215c2da2`, `3d4d613c3` | State stage-reference precedence order-independently: display name, then slug, then ordinal, each accepted only where nothing stronger claims it — **whatever order the stages appear in**. |
| `388c412fa` | Spell the default rule names the way the product does. |
| `2cf1bb0f5` | Bindings are minted per **resource**, not per task. |
| `d7dd9528c` | Point a binding cross-link at the real anchor (`#reuse` → `#deduplication`). |

## The two that fix real agent failures

**Default rule names (`388c412fa`).** `CASE_MGMT_RULE_NAME_DUPLICATE` exempts the front end's default rule names *exactly and case-sensitively*: `Entry rule {N}`, `Exit rule {N}`, `Completion rule {N}`, `Stage complete`, `Stage exit`, `Tasks completed rule`, `Previous task completed`. Established empirically over 14 probes. Our own docs had these misspelled in 38 places — mostly title-cased, e.g. `Entry Rule 1`. An agent following them wrote names that are **not** exempt and tripped the uniqueness check on plans that were otherwise correct. `case-schema.md` now carries the exemption table with the exact literals and an explicit warning that `Entry Rule 1` and `Complete rule 1` are not defaults.

**Mid-build strict findings (`f6c7749db`).** On a plan with one of eight stages built, `--strict --sdd` reports `STRICT_STAGE_NO_TASKS` plus one `STRICT_SDD_TASK_MISSING` per declared task and one `STRICT_SDD_CONDITION_MISSING` per declared row — twenty-plus errors that mean "keep building", not "this plan is wrong". Agents were reading that wall as failure and stopping early.

## Precedence phrasing

`0215c2da2` and `3d4d613c3` are one idea in two files. The earlier wording said name resolves "first" and ordinal "second", which reads as a rule about document order and silently changes meaning when stages appear in a different sequence. Both now state the precedence as a property of the reference, not of the traversal, and say so explicitly. Comparison stays case-insensitive and trimmed and nothing more — `stage-1` and `stage 1` remain different keys, and a reference is never rewritten from one spelling to another.

## Verification

- `npm run skills:validate` — OK, default 27 skills / 1768 files, studioweb 27 / 1768 / 264 replacements
- `python3 -m pytest tests/tasks/uipath-maestro-case/` — 153 passed
- Link/anchor sweep across the whole skill: this branch introduces no broken links and fixes one. Seven broken anchors remain, **all pre-existing on main** and all in files this PR does not touch (`planning.md`, `brownfield.md`, `case-commands.md`, `logging/impl-json.md`, `global-vars/planning.md`) — filed separately rather than widened into this PR.

## What this does not claim

These corrections have not moved the terra golden. Nine repeats on the merged skill average 0.43 against a pre-merge ceiling of 0.38 — inside the noise. They are correct independently of that number: the rule-name spelling was wrong, the reference was broken, the precedence sentence was order-dependent. The terra result is a separate problem.


roberts-cliff:
Fixed the truncation — thanks, that was a real cut and my fault in the rewrite.

Completed in `c9f8…` on `case-schema.md`, with one deliberate difference from the wording on `main`. The original sentence numbered *the plugin defaults* (`Number them with a case-wide counter…`), but this PR's whole point is that those defaults are exempt and repeat freely — so numbering them is no longer the instruction. The restored sentence scopes the counter to authored names and says why:

> **Number authored names with a case-wide counter per label kind — the highest number already authored for that kind, plus 1 — never a per-array counter.** A per-array counter restarts at `1` in every stage and on every task, which is the usual cause of a collision. Count only authored names when picking the next number: the exempt defaults above repeat freely, so counting them inflates every counter without preventing a single collision.

On the item flagged for manual review — **the default spellings are empirical, not read off a product source**, and the PR should be read that way. They come from 14 targeted probes against `validate`, cross-checked against 17 grader-passing plans where 26 rules carry `Entry rule 1` and 20 carry `Completion rule 1`, several repeating five times in one plan, all validating clean. That establishes what the *installed validator* exempts today. It does not establish what the frontend will emit after its next release, and if the two ever drift the exemption table is the thing that goes stale. A reviewer with product access confirming `Completion rule` against the current frontend defaults would upgrade this from measured to sourced — worth doing, since `Complete rule` is the spelling our own docs had and the one an agent would otherwise guess.

`npm run skills:validate` OK, 153 tests pass, no stale `Entry Rule` / `Complete Rule` spellings remain anywhere in the skill.

roberts-cliff:
## Case-suite A/B — and why it does not settle anything

Dispatched both arms of the full case suite on the same inputs (`agent=codex` → gpt-5.6-terra, `cli_version=dev`, `parallelism=4`), one run per arm:

| arm | run | result |
|---|---|---|
| `main` | [34864517338](https://github.com/UiPath/skills/actions/runs/34864517338) | **56/65 PASS** |
| `docs/case-terra-cycle-corrections` | [34864498418](https://github.com/UiPath/skills/actions/runs/34864498418) | **55/65 PASS** |

**Do not read that as a one-task regression.** Thirteen tasks changed status — six improved, seven regressed:

| improved on branch | regressed on branch |
|---|---|
| `connector-wait-trigger` | `e2e-outcome-external` |
| `edit-add-task-run-once-default` | `golden-rebuild-cm-expense` |
| `multi-in-arg-trigger-bind` | `multi-linear-three-stages` |
| `multi-sla-with-escalation` | `multi-parallel-after-predecessor` |
| `phase-0-case-reasoning-regressions` | `single-case-management` |
| `phase-0-procurement-sla-interrupts` | `single-rpa` |
| | `sla-case-breach-takeover` |

Seven commits of Markdown corrections do not have thirteen independent mechanisms. This is model variance at a **20% per-task flip rate between two single runs**, and at that noise floor a ±1 difference in the totals carries no information.

The anomaly that makes it unambiguous: **`golden-rebuild-cm-expense` passed on the `main` arm.** gpt-5.6-terra has never passed that task — 0/9 across measured repeats, and 0/3 in a control arm run on this same model three hours ago. A pass there is the noise floor announcing itself, not a property of `main`.

### What this PR actually rests on

Not a score. The corrections are right independently of any eval number, which is what the PR description already said:

- `Entry Rule 1` / `Complete Rule 1` are **not** the exempt default spellings. Established over 14 targeted `validate` probes and cross-checked against 17 grader-passing plans where 26 rules carry `Entry rule 1` and 20 carry `Completion rule 1`, several repeating five times in a single plan, all validating clean. Our own docs had the wrong spelling in 38 places, and an agent following them produced names that trip the uniqueness check on otherwise-correct plans.
- Measured blast radius of that typo: **129 of 314 past run artifacts** carry a repeated misspelled default — 1168 rule names that would not have collided with the exempt spelling.
- `[Reuse](#reuse)` pointed at a section titled `## Deduplication`. The anchor was broken.
- The stage-precedence sentence was order-dependent and silently changed meaning when stages appeared in a different sequence.

### Verification

- `npm run skills:validate` — OK (default 27/1768; studioweb 27/1768/264)
- `python3 -m pytest tests/tasks/uipath-maestro-case/` — 153 passed
- Link/anchor sweep: this branch introduces no broken links and fixes one. Seven broken anchors remain, all pre-existing on `main`, all in files this PR does not touch.

### What would settle it

A real regression check on this model needs n≥3 per arm — six suite runs for a docs PR. I'd rather state the limit than launder a single pass into a green light. If a reviewer wants that spend, say so and I'll run it.

roberts-cliff:
Stale: written for element-by-element authoring before the convert-first pivot (#3365). The cli#4160 code mapping and per-resource bindings are on main; the rest no longer applies.

PR #3284, opened 2026-09-14
feat(uipath-maestro-case): convert the SDD first, then repair the remainder

roberts-cliff:
Teaches the case skill to emit the plan with `uip maestro case sdd convert` and then repair the remainder, instead of hand-authoring every element the SDD already determines.

**Draft: depends on `uip maestro case sdd convert`, which is unmerged** (`cli` branch `feat/case-sdd-parse`). The version guard below means merging early is safe — an installed CLI without the command falls back to today's behavior — but the measurements are worth confirming against the shipped CLI before this goes in.

## The shape

`Phase 1 registry gate → sdd convert → fill the remainder → Phase 4 repair loop`. Phases 3–7 are untouched. This is deliberately **not** the brownfield route: brownfield skips Phases 1–7 including the registry gate, which would hand every task a placeholder.

Convert runs *after* the gate, never before — it reads the document only, so every tenant identity it cannot supply is one the gate has already resolved.

## Measured on the golden SDD

Convert derives 8 stages, 15 tasks, 20 conditions, and reports 8 unresolved items:

```
5  output-type         an output's shape comes from the resolved resource's schema
3  connector-context   folderKey + connector version metadata
```

Those 8 match, one for one, the 8 tenant-dependent findings a `validate --strict --sdd` run reports on the same plan. Convert's self-report is accurate rather than optimistic, which is what makes `Data.Unresolved[]` usable as a work list. Each `kind` maps to a step that already exists (table in the diff).

## Two guards, because the self-report is a floor and not a ceiling

**`Unresolved[]` names what convert knew it was skipping — never what it emitted wrongly, and never what it omitted silently.** Phase 4's `--strict --sdd` stays the authority, and a plain-profile `Status: Valid` on convert output is explicitly not a finished plan.

**Task entry rules get verified before leaving Phase 2, whatever convert emitted.** On the golden, all 8 stages carry `entryConditions` and all 15 tasks do not. Nothing upstream catches this: it is only a `Task has no entry rules` warning under plain validate, and it does not appear in `Unresolved[]` at all — while a task with no entry rule **hangs `uip maestro case debug` indefinitely** (SKILL.md Rule 6). A plain-Valid convert plan is currently a debug landmine, so the check is written as mandatory rather than advisory.

## Rule 13

`sdd convert --out` writes `caseplan.json` from Bash, which Rule 13's artifact-I/O ban forbade — the rule would have made the command unreachable. It now carves convert out as a sanctioned generator alongside `bindings sync` for `bindings_v2.json`, and says plainly that every edit after the emission is Write/Edit like any other.

## Verification

- `npm run skills:validate` — OK (default 27/1768, studioweb 27/1768/264)
- `python3 -m pytest tests/tasks/uipath-maestro-case/` — 153 passed
- All links in the touched files resolve
- Convert output inspected directly; the 8/15 entry-condition split and the 8-item `Unresolved[]` above are from an actual run, not from the CLI author's report

## Not claimed

No eval has been run on this. It should not be: convert does not yet emit task `entryConditions` or SLA rules, so a golden run today would understate it by ~30 strict errors that are a missing feature rather than a design limit. The run is worth spending once those land.


roberts-cliff:
Superseded by #3365 (merged as 6a6a18f54).

PR #3152, opened 2026-09-08
fix(uipath-maestro-bpmn): scope metadata regeneration away from IS draft handoffs

nikhil-maryala:
## Summary
Fixes the `integration-service-boundary` failure from the claude-code nightly `2026-09-08_04-17-10`, where the agent hand-authored the CLI-owned generated files (`bindings_v2.json`, `entry-points.json`, `operate.json`, `package-descriptor.json`) in a draft/boundary task that forbids it.

## Fix (two parts, both in `skills/uipath-maestro-bpmn/`)
1. **`SKILL.md`** — `Rule 16` in `## Rules` (generated package files are CLI-owned) as the canonical statement, plus a **2-line directive at the IS-draft point in workflow step 3** that points to Rule 16. The point-of-use directive is what changes behavior (see below).
2. **`references/shared/local-metadata-regeneration-guide.md`** — a scoping caveat (defense-in-depth): the guide does not apply to an IS draft/boundary handoff; emit only the `.bpmn` plus a `.md` notes file inside the project directory.

## What's load-bearing (A/B evidence on this branch)
| Version | `integration_service_boundary` |
|---|---|
| Reference-guide caveat only | **FAIL 0.167** (run 34374080786) |
| + 9-line SKILL.md block at point of use | PASS 1/1 (run 34376331497) |
| Block replaced by Rules-list `Rule 16` only | **FAIL 0.167** (run 34387920591) |
| `Rule 16` + 2-line point-of-use directive | **PASS 2/2, 1.000** (run 34391861917) |

The agent reads `SKILL.md` linearly during authoring; a rule only in the trailing `## Rules` list, or buried in a gated reference guide, does not change behavior. The guidance has to sit in the authoring workflow at the point of use.

## Merge order
The PR's smoke gate is red **solely** on `minimal_fault_triage`, the false negative #3151 fixes. **Merge #3151 first**, then re-run here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3287, opened 2026-09-14
fix(uipath-maestro-bpmn): name init as the IS-boundary package-file mechanism

nikhil-maryala:
## Problem

`skill-bpmn-integration-service-boundary` fails **consistently** in the Claude nightly — same reason on `2026-09-04` and `2026-09-14` (score 0.167, only the trivial "BPMN source created" criterion passes):

```
FAIL: draft boundary should not hand-author generated package files:
['bindings_v2.json', 'entry-points.json', 'operate.json', 'package-descriptor.json']
```

This is **not** a regression of #3152 — that fix is present on `main` (point-of-use directive at `SKILL.md:216` + Rule 16 at `SKILL.md:374`).

## Root cause

The agent's only project-creation command is the canonical one the whole skill teaches:

```
uip maestro bpmn init SlackDigestBoundaryBpmn --skip-solution-registration --output json
```

`uip maestro bpmn init` **auto-scaffolds** the four package files — the agent never hand-authors them; the CLI creates them as a side effect. The boundary check forbids those files existing for a draft handoff, so it fails.

Both existing directives state the **outcome** ("emit only `.bpmn` + notes") but neither names the **mechanism** (`init`). The agent reads "don't create them," doesn't create them by hand, then runs `init` as taught everywhere else and trips the check blind. Outcome named, mechanism not — the [point-of-use lesson] one level deeper.

Confirmed reproducible (n=2, identical failure detail both nightlies), not flaky.

## Fix

Name `init` explicitly at **both** points of use (point-of-use directive + Rule 16): for an IS draft/boundary handoff, do **not** run `uip maestro bpmn init` (it scaffolds all four package files) — hand-author the `.bpmn` directly, or delete the four files after init.

Two-location edit, 7 insertions / 2 deletions. No behavior change to any non-boundary task (`init` remains the taught default everywhere else).

## Verification

- `npm run skills:validate` → OK (27 skills, both flavors).
- GH `run-coder-eval.yml` dispatched against this branch, agent=claude, task `integration_service_boundary` — link to follow in a comment.


nikhil-maryala:
Verification run (agent=claude, repo-default model): https://github.com/UiPath/skills/actions/runs/34870823472 — reruns `integration_service_boundary` on this branch. Expect the boundary criterion (weight 5.0) to clear to 1.0 (was 0.0, overall 0.167).

PR #3311, opened 2026-09-15
test(uipath-governance): stop telling compliance-pack agents the CLI is disconnected

sauravrathore01:
## Problem

Every compliance-pack smoke task told the agent the CLI was disconnected:

```yaml
- uip CLI is not connected to a live tenant - auth errors expected.
- Run each command once, capture output, move on.
```

It isn't. Each of those same files carries a `post_run` comment recording the opposite:

```yaml
# Safety net: smoke sandboxes have live tenant auth in CI, so an agent that
# strays into apply flows really mutates the tenant. Clean up after.
```

Agents treat the task description as ground truth about the environment — normally correct, since the harness knows things the agent can't observe. Here it meant the false premise overrode what the CLI actually returned.

## Evidence

Run `skill-governance-compliance-gap-scan-commands` (codex / gpt-5.6-terra) — `weighted_score: 0.556`, one gating criterion missed. Not a timeout and not a crash: `max_turns_exhausted: false`, `crashed: false`, 1 turn of 200, 101s of a 900s budget.

**T+36s — before executing a single command**, the agent's first text block:

> *"I'm using the UiPath governance workflow to query the ISO 42001 posture. I'll capture **the single expected authentication outcome** in `output.txt` and won't attempt to reconnect or prompt."*

The conclusion was pre-written, straight from the prompt. Then reality disagreed three times:

| T | Command | Result |
|---|---|---|
| +57s | `uip gov compliance-packs catalog list --output json` | exit 0 — `"Result": "Success"`, `iso-42001-2023`, `"Available": true` |
| +73s | `uip login status --output json` | exit 0 — `"Status": "Logged in"`, org `codereval`, token valid 50 more min |
| +85s | `rg -qi 'unauth\|not logged\|not connected\|...' output.txt` | exit 1 — **zero matches** |

Final answer: *"Posture results could not be retrieved because the CLI session is not connected."*

Paired with `Run once, capture, move on`, it stopped one command short of `state coverage` — the command the task grades (weight 3.0). Score math: `(0 + 0 + 2 + 2 + 1) / 9 = 0.556`.

Worth noting what it got *right*: both `command_not_executed` guards passed. The skill routed correctly. The task, not the skill, is what failed.

## Root cause — already diagnosed once

[`drift_restore_smoke.yaml`](tests/tasks/uipath-governance/compliance-pack/drift_restore_smoke.yaml#L12-L18) has carried the diagnosis since run `2026-08-07_04-44-12`:

> *"The prompt must not assert either way — telling the agent 'auth errors expected' made it grep output.txt for auth keywords, hit the words 'unauthorized'/'forbidden' in the ISO 42001 control prose that `catalog get` returns, and abort before running coverage."*

The fix landed in that one file. The other 12 kept the premise, so the same failure recurred five weeks later — via the same `output.txt` keyword grep.

---

## Commit 1 — remove the false premise

**13 tasks — replace it** with drift_restore's approved wording:

```yaml
- A uip command may fail with an auth error. If one does, carry on with the
  remaining steps - do NOT retry or login.
- Decide whether a command succeeded from the `Result` field of that command's
  own JSON response. Do NOT search the accumulated output.txt for words like
  "unauthorized" or "forbidden" - the compliance standard's control
  descriptions contain them as ordinary policy text, so that check reports
  failure on a good run.
- Run each command once, capture output, move on.
```

**13 tasks — honest `Platform note:`** in `description`: CI is live, local is not, the prompt asserts neither.

**14 tasks — resolve an output-redirection contradiction.** The old instruction demanded *all* output into `output.txt`, but [`catalog/impl.md`](skills/uipath-governance/references/compliance-pack/catalog/impl.md#L35) needs `catalog get` in `$SESSION_TEMP/catalog.json` for `state coverage` to read. `output.txt` is now an extra review copy; the skill's session layout wins.

**Also included — org-scope** (pre-existing in the tree, same suite):
- `cli-cheatsheet.md`: `state enable organization` is accepted by the CLI but not implemented on the backend — configures nothing, returns no error. Documents the `uip login tenant list` → per-tenant `state enable tenant` loop instead.
- `org_scope_smoke.yaml`: tighten three `[\s\S]*` patterns to `\s+` so scope criteria can't match across an unrelated command.

---

## Commit 2 — bound the apply tasks at "already applied"

Removing the false premise also removed an **accidental brake**. "auth errors expected / run once / do NOT retry" stopped an agent from embarking on long investigations. Without it, an agent that finds real work keeps going — and the manual-config path has no bound.

Run `skill-governance-compliance-apply-commands` (claude-sonnet-5) hit this: `ERROR`, turn timed out after 900s, 97 turns, 48 commands, **criteria never evaluated**. Only 36s of the 912s was command execution; the rest was model time across 21 consecutive form-template JSON introspections.

**Chain:**

1. **Shared-tenant race.** `pre_run` cleanup at `11:26:35` logged *"pack not active"*; the agent's own `state get` at `11:28:00` got `Active: true` with `NewCount: 0` / `NotDeployedCount: 0` — fully applied. A parallel task enabled the pack in that 85-second window; `post_run` found it disabled again. [`full_apply_e2e_live.yaml`](tests/tasks/uipath-governance/compliance-pack/full_apply_e2e_live.yaml) already documents this hazard in its criteria comment: *"a concurrent disable_smoke pre_run, a racing cleanup"*.
2. `state enable` was therefore a genuine no-op and the agent **correctly** skipped it — so the graded command never ran.
3. The only work left was 4 products / 17 `needs-manual-config` settings. [`coverage/impl.md`](skills/uipath-governance/references/compliance-pack/coverage/impl.md#L252) gates that path on user acceptance (L252) and on asking the user for the org-specific value (L258).
4. `Do NOT prompt for confirmation` waived that gate. The agent cited it verbatim: *"proceed to configure them via the AOps handoff **since confirmation is waived**."*
5. Unbounded formData synthesis until the turn timeout.

**Fix, in the two tasks that ask to apply and grade on `state enable`** (`apply_commands`, `org_scope`):

- Already-applied is a **stop condition**, not a cue to start configuring individual settings. Those values are org-specific and not the agent's to invent.
- Narrow the confirmation waiver: it covers the apply the prompt asked for, not steps the skill gates behind asking the user for a value.

**Deliberately not applied** to `partial_apply_smoke` / `partial_apply_aitl_smoke` / `interactive_values_smoke` — the manual-value path is what those tasks grade (`synthesize-formdata.mjs`), so the guard would invert their subject. Nor to `already_configured_smoke`, whose prompt is check-only and already bounded.

No success criteria, weights, thresholds or tags changed anywhere in this PR. Prompts and descriptions only.

## Validation

```
all 17 compliance-pack YAMLs parse, task_id / initial_prompt / success_criteria present
scripts/check-cli-verbs.py → 0 High, 0 Medium, 3 Info (pre-existing, catalog 1.203.0-dev.8677)
scripts/check-task-driver.py → OK
```

**Not yet re-run under coder-eval** — 14 prompts changed, so the suite deserves a pass before merge.

## Follow-ups, not fixable in task YAML

- **Shared-tenant race (the real defect behind commit 2).** `pre_run` cleanup cannot hold a shared tenant against parallel writers, and `apply_commands`' criterion needs `state enable` to actually run — which is meaningless on an already-applied pack. Needs serialized compliance-pack tasks or a per-task tenant.
- **Sandbox provisioning is inconsistent.** The two `apply_commands` runs resolved the tenant differently: `"Tenant ID from UIPATH_TENANT_ID"` (passing) vs `"Tenant ID from auth file /.uipath/.auth"` (failing, `$HOME` = `/`). The failing agent spent an 18-second `find / -iname ".auth"` plus a wasted turn locating it.
- **Redundant coverage.** `gap_scan_commands_smoke.yaml` and `check_routing_smoke.yaml` test the same flow, and `check_routing` has strictly tighter patterns. gap-scan's only unique value is its two `command_not_executed` guards on `pack-walker.mjs` and `deployed-policy`; those could fold in and the file could go.
- **Sibling suites** (`uipath-insights`, `uipath-platform/*`) carry similar "not connected" wording. Left alone — no evidence yet that their sandboxes are authenticated, and asserting the reverse would be the same mistake inverted.
- coder-eval result folders (`skill-governance-compliance-*_failed/`) land untracked in the repo root and aren't gitignored.


sauravrathore01:
-

sauravrathore01:
-

PR #3321, opened 2026-09-15
fix(uipath-maestro-bpmn): type the bundled spec's uipath:context inputs

rockymadden:
## What

[UiPath/cli#4257](https://github.com/UiPath/cli/pull/4257) gave the 22 registry `xmlTemplate`s that emit a `<uipath:context>` the `type` attribute the canvas parser requires. This bundle is a copy of that manifest and still taught the old shape.

That matters because `SKILL.md` points registry-evidence tasks at this file as the fallback when the CLI is unavailable — so the fallback now contradicted the CLI it stands in for. A node pasted from it fails `validate` and `refresh`:

```
BPMN_PARSE_ERROR: Unable to load BPMN XML: Invalid BPMN canvas state:
expected diagram "Diagram_1" node "Task_1".data.uipath.context item at
index 0.type to be a string.
```

I opened this gap by fixing only the CLI side, so it's mine to close.

## How

Same transform as #4257 — each input's type read from its own entry's `contextFields`, all 93 already declared:

| Declared type | Inputs |
| --- | --- |
| `string` | 82 |
| `json` | 9 |
| `number` | 1 |
| `jsonSchema` | 1 |

The 22 templates are now **byte-identical** to the merged CLI manifest, so #4257's end-to-end verification (one node per type flavour: valid with the attribute, `BPMN_PARSE_ERROR` without, against 1.203.0) covers this XML unchanged.

Version `1.0.1` → `1.0.2`, matching how #2366 bumped it for a content change.

## Deliberately not a full resync

The bundle is a **curated fork, not a snapshot**, and is ahead of the CLI in three places. Copying the manifest over would regress all of them:

| Divergence | Origin | Direction |
| --- | --- | --- |
| `BPMN.ScriptTask` v3 runtime contract | #2366 | skills ahead |
| `uipath.human-in-the-loop` shortcut split | #1906 | skills ahead |
| `extractedFrom` script-task sources | #2366 | skills only |

After this PR, `BPMN.ScriptTask` is the **only** template that differs — and it differs on purpose. Left alone: `uipath:Activity` (CLI only) and the `version`/`source` fields.

Worth noting separately: the CLI manifest's `BPMN.ScriptTask` is the stale one. Not this PR's problem, but someone should port #2366 the other way.

## The guard

The existing `*_matches_cli_manifest` tests assert remembered facts about individual entries, so they cannot see this class of drift — which is how 93 inputs went stale unnoticed. The new guard walks every context input in the bundle and checks the type is present and agrees with `contextFields`.

Verified it fails on both failure modes:

```
AssertionError: context inputs missing type=: ['Orchestrator.CreateQueueItem.queueName']
AssertionError: Orchestrator.CreateQueueItem.queueName type mismatch
```

## Verification

| Check | Result |
| --- | --- |
| Untyped context inputs remaining | 0 |
| Templates differing from merged CLI | 1 (`BPMN.ScriptTask`, intentional) |
| `pytest tests/tasks/uipath-maestro-bpmn/` | 275 passed |
| `scripts/check-task-driver.py` | ok |
| `scripts/check-skill-links.mjs` | 6817 links resolve |


nikhil-maryala (approved):
## Review — Approve ✅

The correct close for the gap #4257 opened. Verified independently (fetched the bundled `bpmn-spec.json` at `main` and this branch, plus the corrected CLI manifest from #4257, and replicated the guard):

- **Surgical.** Only two things changed: `version` `1.0.1`→`1.0.2`, and `type="…"` insertions in `xmlTemplate`s. No other field on any entry differs from `main`; after stripping type attrs, every template equals `main`.
- **Complete + correct.** All **115 context inputs are typed** (0 missing), and every one whose name is declared in its `contextFields` **agrees** with it (0 mismatches).
- **Divergences preserved exactly as documented.** Against #4257's merged CLI manifest, `BPMN.ScriptTask` is the **only** differing shared template (the #2366 v3 contract — intentional, skills ahead), and `uipath:Activity` remains CLI-only. Nothing regressed.
- **Guard is real.** `test_bundled_context_inputs_declare_a_type` walks every context input and checks presence + agreement with `contextFields` — exactly the drift class the existing `*_matches_cli_manifest` tests can't see. CI (`maestro-bpmn checker unit tests`) is green, so it runs and passes.

Good call keeping this a curated-fork resync rather than a blind snapshot copy, and the note that the CLI's `BPMN.ScriptTask` is the stale side (port #2366 the other way) is the right follow-up. Approving.

### Minor (optional) — same nit as #4257

The guard regex `re.findall(r"<uipath:input [^/>]*/>", context)` excludes `/` from attribute values, so a future context input whose `value` contains a `/` (folder path / URL default) would be **silently skipped** instead of type-checked. Nothing hits this today. `[^>]*?` would keep it exhaustive — worth matching in both the Python and the TS guard.

### Note

`mergeStateStatus` is `BLOCKED` on the two still-pending checks (`Claude Code Review`, `Run skill smoke tests`) + required review — nothing in the change itself.

rockymadden:
Took the nit — and it was worth more than a nit. Fixed in 9e3541a.

Confirmed the gap first:

| Tag | Old pattern | New |
| --- | --- | --- |
| `value="{queueName}"` | matched | matched |
| `value="Shared/Finance"` | **skipped** | matched |
| `value="https://x.test/a"` | **skipped** | matched |
| `value="/v1/items"` | **skipped** | matched |

Rather than `[^>]*?` I added a count assertion alongside the pattern change, so a regex gap can never shrink coverage silently again. That immediately failed on clean input and surfaced a second gap `[^>]*?` would **not** have closed:

```xml
<uipath:input name="metadata" type="json"><![CDATA[{}]]></uipath:input>
```

`Intsvc.ActivityExecution.metadata` is element-form, not self-closing, so neither guard ever saw it. Matching the opening tag (`<uipath:input\b[^>]*>`) covers both forms — which is why you counted 115 inputs and the corrected guard counts **116**. The missing one is exactly that.

The count needs `\b` too: `<uipath:input` is a prefix of `<uipath:inputSchema`, which `BPMN.ScriptTask`'s context holds and which is not an input.

Verified against four mutations:

| Mutation | Caught |
| --- | --- |
| `type` removed | `context inputs missing type=: ['Orchestrator.CreateQueueItem.queueName']` |
| `type` mismatched | `Orchestrator.CreateQueueItem.queueName type mismatch` |
| Untyped input, value holds a `/` | `context inputs missing type=: ['Orchestrator.CreateQueueItem.queueName']` |
| Untyped element-form input | `context inputs missing type=: ['Intsvc.ActivityExecution.metadata']` |

`pytest tests/tasks/uipath-maestro-bpmn/` — 275 passed.

The same fix for the TS guard is up as UiPath/cli#4258, since #4257 is already merged.

PR #3322, opened 2026-09-15
fix(uipath-maestro-bpmn): scope the net-per-container ban to the outer case

rockymadden:
## What

Three places in the skill told the agent different things about where a failure net goes inside a multi-instance iteration:

| File | Says |
| --- | --- |
| `patterns/composing-guide.md:86-89` | add a second net **inside** the iteration when one bad item should be recorded and the batch should continue |
| `patterns/high-volume-batch-guide.md:104-107` | place the net **inside** `process_each` |
| `references/structural-bpmn.md:515` | **do not author a net per container** |

In the 2026-09-15 nightly the agent read all three, then attached an error boundary event to an inner subprocess instead:

```xml
<bpmn:boundaryEvent id="Boundary_HandleClaimFailure"
    attachedToRef="SubProcess_HandleClaim" cancelActivity="true">
  <bpmn:errorEventDefinition />
</bpmn:boundaryEvent>
```

`uip maestro bpmn validate` reported **Valid** — the CLI doesn't grade pattern composition — and `skill-bpmn-composition-nested-review` failed its gating criterion:

```
FAIL: no event subprocess — a per-claim failure has nowhere to go
```

## How

The blanket sentence is right for the ordinary case and wrong for the one the other two guides describe. Qualify it rather than delete it, and name why the construct the agent reached for isn't equivalent:

> Do not author a net per container — except inside a multi-instance iteration or a queue performer, where per-item failures need per-item handling. There the process-level net would end the whole instance on the first bad item, so the iteration gets its own net, placed directly in it. An error boundary event is not a substitute: it resumes the main path rather than ending that item.

That last clause is the part the agent needed. Its design was defensible from the construct table alone — the boundary bullet reads "recover and continue", and per-claim escalation *does* let the batch continue — so saying "use a net" without saying why the boundary event is wrong would have left the same coin-flip.

## Verification

| Check | Result |
| --- | --- |
| `scripts/check-skill-links.mjs` | 6818 links resolve |
| `pytest tests/tasks/uipath-maestro-bpmn/` | 275 passed |
| Three statements now agree | yes, quoted side by side below |

**This is a behaviour change, so it is not mechanically verifiable.** Only a nightly shows whether the agent now picks the event subprocess. Worth landing alongside the sibling `SKILL.md` fix so one run covers both.

`skill-flavors/studioweb` composes from this source and carries no copy of the file, so nothing else needs the edit.


nikhil-maryala (approved):

**Severity tally — 0 Blocker · 0 Critical · 0 Major · 0 Minor · 0 Nit.** No inline findings to post.

Clean, correctly scoped consistency fix. Verified on the branch:

- **The three sites now agree.** `composing-guide.md:86-89` ("Add a second net inside a multi-instance iteration or a queue performer…") and `high-volume-batch-guide.md:104-107` ("Place a `failure-escalation` net **inside** `process_each`…") both already prescribed an inner net; the previously-blanket `structural-bpmn.md:515` now carries the same exception, so the contradiction that produced the coin-flip is gone.
- **The load-bearing clause is the right one.** Naming *why* an error boundary event isn't a substitute (it resumes the main path rather than ending that item) is exactly the distinction the failing nightly agent missed — a bare "use a net" would have left the same ambiguity. Point-of-use placement is correct.
- **Anchor + links:** the new `patterns/composing-guide.md#scoping-the-failure-net` target exists (heading at :80) and every link resolves (6818/6818); `composing-guide.md` is the shared hub both sibling guides already point to.
- **Flavor:** `skill-flavors/studioweb` carries no copy of `structural-bpmn.md`, so it composes from canonical and inherits the edit — no override needed.

Behaviour change, so not mechanically provable — a nightly is the real proof — but the doc contradiction was plainly real and this closes it. Approving.

PR #3325, opened 2026-09-15
fix(agents,maestro-flow): coded-agent resource file is under `process/function/`

dmetzgar:
Four references tell an authoring agent to read the minted `resource.key` from `resources/solution_folder/process/agent/<project>.json`. **That file is not written for a coded agent.**

## Why

`uip solution projects add` infers project type from the manifest it finds — `project.uiproj`, `project.json`, or `uipath.json` — and has no flag to override it. A coded agent scaffolded by `uip codedagent new` has only `uipath.json`, which is indistinguishable from a Function project's. So the resource lands at `process/function/<project>.json` with `spec.type: "Function"`, and `uip solution projects list` reports `Type: Function`.

Observed on CLI `1.203.0-dev.8669`. A coder-eval run spent **18 turns and 167s** trying to make the documented path appear — removing and re-adding the project, hand-editing `uipath.json`, deleting and regenerating the resource files. It never appeared. The run finished by inventing a GUID, which `uip maestro flow validate` accepted, because validate does not check that a well-formed key resolves on the tenant.

## What changed

Each of the four sites now:

- names `uip maestro flow registry list --local` as the source to read the key **from**, rather than one of two equal options;
- **globs** for the resource file (`process/*/<project>.json`) instead of hard-coding a directory, so the guidance survives the classification being fixed;
- states that `Type: Function` on a coded agent is that classification and **not** a scaffolding mistake, so it isn't "corrected" before wiring the Flow;
- says never to substitute a hand-written GUID, and why `validate` won't catch one.

`uip codedagent run` in the troubleshooting table is also qualified: it executes the graph, so it needs `UIPATH_ACCESS_TOKEN` and LLM-gateway reach, and where those are absent its import and virtualenv errors are environment rather than project defects. `uip codedagent review` is named as the offline rung.

## Scope

**Low-code agent projects are deliberately untouched.** They carry `agent.json` plus `project.uiproj`, so the classification is plausibly correct there and `process/agent/` may well be right. I did not verify that, so I left all of `references/lowcode/` alone — it still says `process/agent/` in several places, correctly as far as I know.

## Follow-up

The underlying CLI behaviour looks like a defect: the classifier could read `uipath.json`'s `agents` vs `functions` map rather than keying on the filename. Worth filing against the CLI; this change makes the guidance true in the meantime, and the glob means it stays true afterwards.

## Checks run

- `check-skill-verbs.py` — `uipath-maestro-flow` 38 findings / **0 blocking**; `uipath-agents` 73 / **0 blocking**
- `.maintenance/check-links.sh` — 972 checked, 0 broken
- `.maintenance/check-anchors.sh` — 323 checked, 1 bad, **pre-existing and unrelated**: `references/diagnose/troubleshooting-guide.md` → `../shared/cli-conventions.md#7-use-uip_log_levelinfo-for-debug-runs`. Left alone rather than mixed into a guidance-correctness PR.

Phase 0 of `UiPath/flow-builder-sdk` `docs/SKILLS_PROMOTION_PLAN.md`. The preview tree carries the same two wrong claims plus a "do not search for a published release" instruction; that side is fixed upstream in flow-builder-sdk and arrives here via the nightly Maestro SDK preview sync.


dmetzgar:
Closing this — it is the wrong fix, and it conflicts with better work.

**Superseded on 2 of the 4 files.** #3165 already changes `coded/embedding-in-flows.md` and `coded/quickstart.md`, and it fixes the cause rather than the symptom: `uipath new --type auto` picks the template from whichever framework package is installed in the active venv, and the skills had the framework install running *after* `new`, so `auto` produced a Coded Function scaffold. That PR is validated with probes against wheels from the two Python PRs plus a coder-eval run. It should land instead of this.

**My causal claim here was also wrong.** I said a coded agent's `uipath.json` is indistinguishable from a Function project's. The eval artifacts show the scaffolded manifest carried `"functions": {}` *and* `"agents": {}` — both empty, because `uip codedagent init` failed on a missing dependency before it registered the entry point. So the `Type: Function` result came from a specific classifier fallback, not from inherent ambiguity, and documenting the folder it happens to produce is a bandage over a CLI gap.

**The two things in here that were real** are independent of folder names and are not covered by #3165:

- `uip maestro flow validate` accepts a well-formed resource key that resolves to nothing on the tenant, so a fabricated GUID fails at run time rather than author time. The run that motivated this scored 1.000 on exactly that.
- `uip codedagent run` executes the graph and needs `UIPATH_ACCESS_TOKEN` plus LLM-gateway reach; where those are absent, its import and virtualenv errors are environment rather than project defects. `uip codedagent review` is the offline rung.

Both live in `uipath-maestro-flow`'s `plugins/agent/impl.md`, which #3165 does not touch. I will reopen those as a smaller PR against that file alone.

PR #3326, opened 2026-09-15
test(uipath-maestro-flow): grade the native entity read in the dispute-resolution structure gate

rockymadden:
## What

#3249 taught every billing advisory that an entity read comes in two shapes and tenant availability picks one. It missed a gate.

```
check_structure_billing_dispute_resolution.py:190   assert_flow_has_node_type(["dataservice.query-entity-records"])
check_structure_billing_dispute_resolution.py:106   ds = of("dataservice.query-entity-records")
```

Unlike the advisories, **this one gates**: weight `2.0`, `pass_threshold: 1.0`. So a native build fails the run outright on a node choice the skill itself recommends — `data-fabric/planning.md:58` defaults record CRUD to the native node.

This is the only task checker left asserting the connector shape. Its two siblings, `check_billing_invoice_lookup.py:45` and `check_billing_discrepancy_detector.py:42`, already call `assert_flow_has_any_node_type(ENTITY_QUERY_HINTS)`.

## Demonstrated, not argued

The native reference #3249 added was never fed to this assertion by any test. Feeding it:

```
pre-fix   FAIL: topology — expected ≥2 Data Service query nodes (ERP+CRM), found 0
post-fix  PASS  BillingDisputeResolution.reference.flow         reads: uipath.connector.uipath-uipath-dataservice.query-entity-records
          PASS  BillingDisputeResolution.native.reference.flow  reads: core.datafabric.read
          PASS  negative control rejected: expected ≥2 entity-read nodes (ERP+CRM), found 1
```

## It also closes #3249's arithmetic

#3249 attributed the 2026-09-11 native run's `0.727` to the advisory alone. The advisory is weight `1.0` of `11.0`; failing only it scores `10/11 = 0.909`.

```
weights   validate 3.0 + structure 2.0 + e2e 5.0 + advisory 1.0 = 11.0
0.727     = 8/11  ->  3.0 of weight failed
#3249     confirms validate passed and e2e debugged green
therefore  advisory (1.0) + THIS gate (2.0)
```

So the native run lost 2.0 here, silently, and the fix for it was deferred without being named.

## Changes

- **Both call sites use the shared constant.** `assert_flow_has_any_node_type(ENTITY_QUERY_HINTS)` for presence, and a union over the same hints for the topology counts. No new constant, no cross-import from the advisory utils — `flow_check.ENTITY_QUERY_HINTS` already exists for exactly this.
- **Enrols this gate in `test_billing_gates_use_the_shared_entity_hints`.** That guard exists to stop a gate reverting to the connector-only hint while the suite stays green. It listed two of the three gates; its own docstring describes the failure this PR is fixing.
- **Adds the unit tests the checker has promised since #1320.** Its section header reads *"pure topology assertions (unit-testable against the reference)"* and nothing ever tested it, which is why a gating criterion was only exercised by a live eval run. Both references must pass `assert_topology`. A negative control keeps the ≥2 read count load-bearing, so a hint list matching nothing cannot make the count assertions vacuous.
- **Failure messages say "entity read", not "Data Service".** Naming one shape in the diagnostic for a gate that accepts both sends the reader to the wrong fix.

## Nothing is loosened

A connector build passes exactly as before, and the anti-hardcode guard is intact: a Script-only flow still fails, and one lookup instead of two still fails. The gate went from *one* acceptable shape to *two*, matching what `test_same_ground_advisories.py:31-35` already states:

> Tenant availability decides the shape, so a correct build comes in either one and every advisory has to grade both

## Verification

`pytest` is not installed on this machine and was not run locally; the assertions above were exercised directly against both reference flows. CI's `maestro-flow checker unit tests` job (`pytest tests/tasks/uipath-maestro-flow/`) runs the suite including the new file.

Not gated on an eval re-run — this is a grading fix, independent of #3324.


nikhil-maryala (commented):
## Review — LGTM ✅ (closes the deferral #3324 named)

Clean, correctly scoped grading fix. Verified everything independently — including running the suite the PR body couldn't run locally.

**Reproduced the bug and the fix** against both reference flows:

| Reference | old connector-only hint | new `ENTITY_QUERY_HINTS` union |
| --- | --- | --- |
| `…reference.flow` (connector) | 2 → **PASS** | 2 → PASS |
| `…native.reference.flow` (native) | **0 → FAIL** | 2 → PASS |

So the gate really did fail a native build outright (`≥2` against `0` found) on `core.datafabric.read` — the node `data-fabric/planning.md:58` defaults to — and the connector path is unchanged. Not a no-op.

**Nothing loosened, confirmed:**
- `ENTITY_QUERY_HINTS = ("uipath-dataservice.query", "core.datafabric.read")` — both shapes, and the connector type `uipath.connector.uipath-uipath-dataservice.query-entity-records` still matches the first hint by substring.
- `ds` is a set of node ids unioned over hints, so overlapping hints can't inflate the count; `≥2` stays load-bearing (negative control removes one read → 1 < 2 → `SystemExit`, and a Script-only flow is still 0).
- Both call sites (`ds` topology count at :108 and presence at :192) now use the shared constant, matching the two sibling checkers.

**Tests I ran (venv py3.13):**
- New `test_structure_billing_dispute_resolution.py` — 3 pass (both shapes accepted, single-read rejected with `"entity-read"` in the message).
- `test_flow_check.py::test_billing_gates_use_the_shared_entity_hints` — passes with the third gate now enrolled, so a future revert to the connector-only hint fails the guard.
- Full `pytest tests/tasks/uipath-maestro-flow/` — **1243 passed**.

Good calls: adding the unit coverage the checker's own header has claimed since #1320 (the gap that let this survive #3249), and correcting the diagnostics to say "entity read" so a failing agent isn't sent to the wrong fix. The #3249 arithmetic (advisory 1.0 + this gate 2.0 = the missing 3.0 of 11.0) is consistent with the weights.

No changes requested. Happy to flip this to a formal approval if you want that signal.

nikhil-maryala (approved):
## Re-review — Approve ✅

No changes since my verified pass (no commits after 18:50:44Z), and I raised nothing to fix — the review was clean. Converting to a formal approval as offered.

Recap of what I verified independently: the native reference genuinely failed the old shape-pinned gate (0 entity-read nodes vs `≥2`) and passes under the new `ENTITY_QUERY_HINTS` union (2), while the connector reference is unaffected (2→2); nothing is loosened (single read and Script-only still fail); and the full `pytest tests/tasks/uipath-maestro-flow/` ran **1243 passed** including the 3 new unit tests and the guard now enrolling the third gate.

PR #3327, opened 2026-09-15
MST-15261 ElementId is not nullable for full validate

charlesliu9:
The issue
Building a Maestro Case caseplan.json. Variables declared in variables.inputOutputs[] with id matching var (e.g. id: "isEligible", var: "isEligible") are rejected by uip maestro case validate (full mode) with "Variable 'vars.isEligible' does not exist". The same file passes uip maestro case validate --skeleton-v2 as Valid.

The skill's variable resolution documentation (global-vars/impl-json.md) states the resolver uses id as the sole match key for =vars.X references. The variables are correctly declared with matching ids, yet the full validator cannot find them. 

Documentation defect
[skills/uipath-maestro-case/references/plugins/variables/global-vars/impl-json.md:20] is what misled you:

elementId | FE picker scope only. Controls which panel displays the variable. Not used by the resolver. | No

True for the runtime resolver, false for the validator — and the row gives no hint that omitting elementId is fatal. The recipes further down (lines 165–167) do all emit elementId, so generated caseplans are usually correct; hand-authored ones following the semantics table are not.

Why skeleton-v2 passes
Not because the declarations are valid. --skeleton-v2 skips task-content checks, so if your =vars.X consumers sit in task entry conditions / skipCondition, it never evaluates them. In my repro the identical broken file is Valid under --skeleton-v2 and Failure under full. Treat a green skeleton-v2 as silent on this class of defect.

Fix
Ensure elementId added to each of the variables.inputOutputs[] entries, e.x.:

{ "id": "isEligible", "name": "isEligible", "type": "boolean",
  "default": "false", "custom": true, "elementId": "root" }


BenHorblit-UiPath (approved):
Makes sense to me at least

Yiqi517 (approved):
Should work!

PR #3328, opened 2026-09-15
fix(tests): stop a brownfield scaffold leaking into nine sibling tasks

rockymadden:
## What

`scaffold_movie_report_flow.py` lives in `connector_features/datafabric_connector/_setup/`, which **ten** tasks mount. Only `smoke_update_existing_flow` invokes it. The other nine are greenfield builds that must author their own flow. The script's own docstring says so:

> Used by `smoke_update_existing_flow` to give the agent a real brownfield artifact to edit + revert.
> The initial state has ONE `query-entity-records` node targeting `FlowCodeEvalEntity`

It is agent-visible, reads like required setup, and seeds a node. So any agent that explores `_setup/` can run it and inherit a node its task never asked for.

## It already broke a required check, twice

`skill-flow-datafabric-smoke-query` did exactly this on run `35010678837` and again on its re-run. From the run artifact:

```
command 11   python3 _setup/scaffold_movie_report_flow.py && find MovieReportSolution ...
checker      FAIL: expected exactly 3 Query Entity Records activities, found 4
```

One seeded node, plus the three the prompt asks for, is four. Its prompt never mentions `MovieReportFlow` — it says *"Build a Flow against FlowCodeEvalEntity"*.

**This is not an authoring-guidance failure.** The artifact shows the agent was `gpt-5.6-luna` with `{'Bash': 32}` and **zero** `Write` or `Edit` calls, driving `node add` / `node configure` / `edge add` / `format` / `validate`. It simply ran a script it could see.

It also explains why the task reads as flaky rather than broken: it fails only for agents that go looking in `_setup/`, which is model- and run-dependent. The same task passed at `1.00` on run `35007473034`.

## Fix

Move the scaffold to a sibling `_setup_brownfield/` that only `smoke_update_existing_flow` mounts:

```yaml
    - type: template_dir
      path: _setup_brownfield
      mount_point: _setup
```

It still lands at `_setup/` inside that sandbox, because template sources sharing a `mount_point` **merge** — the mechanism every one of these tasks already relies on to stack `../../_setup` and `_setup`. So the `pre_run` line is unchanged:

```yaml
  - command: 'python3 "_setup/scaffold_movie_report_flow.py"'
```

## Verification

Simulated the merge of all three sources for `smoke_update_existing_flow`:

```
merged _setup:  cleanup_solutions.py  contract_registry.entity.json  ensure_entity.py
                flow_code_eval_entity.entity.json  preflight_connections.py
                scaffold_movie_report_flow.py  seed_flow_code_eval_records.py

pre_run targets resolve:  preflight_connections.py  ensure_entity.py
                          flow_code_eval_entity.entity.json  scaffold_movie_report_flow.py   all OK
filename collisions across the three sources:  none
other tasks that can still reach the scaffold:  none
```

## Not fixed here

`smoke_update_existing_flow` can still run the scaffold mid-task, which would re-snapshot `pre_state.flow` and undermine the revert diff its verifier depends on. Same family, different bug, one task. Flagged rather than folded in.

## Relationship to #3324

#3324's smoke gate is red because of this leak plus an unrelated transient tenant preflight error. Once this lands, a re-run there should clear the 95% threshold. #3324 needs no changes for it.


nikhil-maryala (commented):
## Review — LGTM ✅

Correct root-cause fix, minimal and well-scoped. Verified the whole chain on the branch:

- **The leak was real.** The scaffold sat in the shared `_setup/` source that **10** tasks mount; only `smoke_update_existing_flow` invokes it, the other 9 are greenfield. `smoke_query` (which mounts `_setup`) is the one that broke — `expected exactly 3 Query Entity Records activities, found 4` = 1 seeded + 3 built — and its prompt never mentions `MovieReportFlow`. The `{'Bash': 32}`, zero-`Write`/`Edit` artifact makes it clear the agent just *ran a script it could see*, not an authoring-guidance miss.
- **The move closes it.** `scaffold_movie_report_flow.py` is gone from both shared sources — resolved `../../_setup` (`uipath-maestro-flow/_setup`) and `_setup` (`datafabric_connector/_setup`) — so none of the 9 greenfield tasks can reach it. It now lives only in `_setup_brownfield/`, and `grep` confirms `smoke_update_existing_flow.yaml` is the **sole** reference to that dir.
- **The mechanism is the one already in use.** Every task in this dir stacks `../../_setup` + `_setup` under `mount_point: _setup`; adding `_setup_brownfield` as a third source under the same mount is the identical merge, so the file still lands at `_setup/scaffold_movie_report_flow.py` and the `pre_run` line (unchanged) resolves. No filename collisions across the three sources.
- Rename is pure (0/0); the checker's docstring reference to the scaffold still reads correctly.

Good call flagging — but not folding in — the separate mid-task re-scaffold bug in `smoke_update_existing_flow`; that's a different failure mode and belongs in its own change.

No changes requested. Happy to convert to a formal approval if you want that signal.

PR #3330, opened 2026-09-15
fix(uipath-maestro-flow): carry the Write timing rule into the brownfield sites

rockymadden:
## What

#3324 timed the full-file `Write` rule at four sites. `@nikhil-maryala` flagged the two it left silent and deferred them to a follow-up. This is that follow-up.

```
references/author/brownfield.md:22              "use Write only when ≥70% of nodes change"   no condition
references/author/editing-operations-json.md:5  "use Write only when ≥70% of nodes change"   no condition
```

## Why it matters more after #3324, not less

#3324 relaxed the three loud sites from an absolute bar (*"Never `Write` a flow containing CLI-owned nodes"*) to a timed one (*safe before `node configure`, unsafe after*). That made these two silences **riskier than they were before it landed**: brownfield is where a flow most likely **already has** configured connector nodes, which is exactly the unsafe window. An agent reading only `brownfield.md` got a bare ≥70% heuristic with no hint that the window had closed.

## Change

Both sites now state the condition. `brownfield.md` additionally says what it means *there*:

> use `Write` only when ≥70% of nodes change **and no CLI-owned node in the file has been configured yet**. On an existing flow that second condition rarely holds — `node configure` has already populated `inputs.detail` and `bindings[]`, so a full-file `Write` clobbers both and `flow validate` still passes. That makes `Edit` the default here for reasons greenfield does not have.

That last clause is the behavioral point: brownfield's `Edit`-by-default is not the same ≥70% judgement call as greenfield's, and previously the doc implied it was.

## Audit

Checked every `≥70% of nodes` mention in the skill afterwards:

| Site | Timed? |
| --- | --- |
| `editing-operations.md:11` (ladder rung 3) | ✅ |
| `greenfield.md:326` (tool selection by ownership) | ✅ |
| `editing-operations-json.md:5` | ✅ this PR |
| `brownfield.md:22` | ✅ this PR |
| `greenfield.md:44` (T2 row) | ❌ **deliberately** |

`greenfield.md:44` is the T2 row of the three-turn map. T2 runs before T3's `node configure`, so a `Write` there is in the safe window by construction — the condition would not change what the agent does, and Step 4's detailed note carries it for the cases that need it.

## Checks

`check-links` 977/0 · `check-link-text` 628/0 · `check-versions` 13434/0 · `check-anchors` 328 checked, 1 pre-existing failure in `diagnose/troubleshooting-guide.md` untouched by this diff.


nikhil-maryala (commented):
## Review — LGTM ✅

Clean close of the two silent sites, and the reasoning for *why brownfield especially* is the right framing. Verified on the branch:

- **#3324 is on main** (editing-operations.md:11 rung 3 carries the timed wording), so this genuinely extends the same rule rather than racing it. Both edited files (`brownfield.md`, `editing-operations-json.md`) were untouched by #3324, so no conflict.
- **Both edits are consistent with rung 3** and correctly stop at the timing condition without dragging in the `variables.nodes`/`format` detail — appropriate for these terse pointer sites. Neither reintroduces the "unrecoverable / owns only layout" phrasing that #3324 reconciled.
- **The brownfield-specific clause is the real value.** "On an existing flow that second condition rarely holds … so `Edit` is the default here for reasons greenfield does not have" is correct and non-obvious: brownfield is exactly where a flow already has configured connector nodes, i.e. the unsafe window. Good addition.

**Independent audit of every `≥70%` site** (your regex-based table, re-checked by hand since markdown bold/backticks fooled a naive match):

| Site | Timed? |
| --- | --- |
| `editing-operations.md:11` (rung 3) | ✅ |
| `greenfield.md:290` (Step 4 note) | ✅ (the detailed paragraph) |
| `greenfield.md:326` (tool selection) | ✅ |
| `editing-operations-json.md:5` | ✅ this PR |
| `brownfield.md:22` | ✅ this PR |
| `greenfield.md:44` (T2 row) | ❌ deliberately — sound: T2 runs before T3's `node configure`, so a Write there is in the safe window by construction, and :290 carries the nuance |

Links resolve (6822/6822), CI green (33 checks pass, only smoke tests pending), and the one anchor failure the PR notes is pre-existing in `diagnose/troubleshooting-guide.md`, untouched here.

No changes requested. Happy to convert to a formal approval if you'd like that signal.

nikhil-maryala (approved):

PR #3331, opened 2026-09-15
fix(uipath-maestro-flow): let a headless run produce and act on the plan

rockymadden:
## What

Planning terminates in an approval gate with no non-interactive branch, in three places:

```
planning-arch.md:359   "Do not proceed to Phase 2 until the user explicitly approves"
planning-impl.md:208   "Do not build until the user explicitly approves"
planning-impl.md:5     "The user must explicitly approve … before this phase"
```

`SKILL.md` rule #5's headless fallback requires a **marked recommended option** and says to *"stop and report the open decision"* when none exists. None of these gates marks one. So a literal read tells a headless agent building a branching 17-node flow to stop before authoring anything.

## The fix uses a distinction rule #5 already draws

Rule #5's consent gates are **destructive operations and tenant writes**, which are never auto-answered. Approving a local markdown plan is neither. It is a **review** gate, so the correct headless behavior is to proceed on the plan and report that nobody reviewed it.

Stated that way at each gate, rather than carving an exception into rule #5. Each gate keeps its interactive wording verbatim — the non-interactive branch is purely additive.

## The load-bearing half is that the file still gets written

`greenfield.md` tells the agent to plan at 5+ nodes with branching. The skipped-planning path leaves no resolved node table, edge table, or bindings to author from, so the topology gets re-derived at every step.

The `2026-09-15` billing-dispute run is the shape of that failure: it skipped planning, wrote no plan file, and spent **3181s of a 3600s budget** deliberating without authoring a single user-owned node. Both new passages say to write the file even when nobody will read it, and why.

## Honest scope note

This is the one item in the original root-cause plan I could **not** validate. The thinking text is redacted in the eval transcript, so "the deliberation went into re-deriving a topology it never wrote down" is an inference from the tool trace and timings, not an observation. The gate contradiction itself is plainly real and worth fixing regardless; whether fixing it recovers the lost hour is unproven until a re-run.

## Audit

Grepped for every plan-approval gate afterwards. The three above are all of them — `planning-impl.md:5`'s differing wording (*"must explicitly approve"*) is why a first pass missed it, and it is included.

## Checks

`check-links` 979/0 · `check-link-text` 631/0 · `check-versions` 13440/0 · `check-anchors` 328 checked, 1 pre-existing failure in `diagnose/troubleshooting-guide.md` untouched by this diff.


nikhil-maryala (commented):
## Review — LGTM ✅

Correct fix, and it leans on a distinction rule #5 already draws rather than carving a new exception. Verified on the branch:

- **Consistent with SKILL.md rule #5.** Rule #5 reserves "stop and report the blocked step" for *consent gates — destructive operations and tenant writes*. Approving a local `.md` plan is neither, so treating it as a **review** gate (proceed on the plan, report it went unreviewed) is the right non-interactive behavior — not the "stop" path. The framing at each gate says exactly this.
- **Additive, interactive wording preserved verbatim.** Both gates keep "Do not proceed/build until the user explicitly approves"; the headless branch is appended, and `planning-impl.md:5`'s prerequisite now also names the non-interactive alternative and links the handoff section.
- **Audit is complete.** Grepping every plan-approval gate across `planning-*.md` returns exactly the three fixed (`planning-arch.md:359`, `planning-impl.md:5`, `planning-impl.md:208`). The only other "until the user" hit — `planning-arch.md:118` — is a `wait-for-message` node table row, not a gate. Correctly untouched.
- **The load-bearing half is right:** "Never skip writing the file" at both phases addresses the actual failure (no resolved node/edge tables ⇒ per-step topology re-derivation). The new `#handoff-to-phase-2` link resolves (heading exists at :347); link checker clean (6825/6825).

Appreciate the honest scope note that the "recovered hour" is an inference from the tool trace/timings, not an observed cause — the gate contradiction is plainly real and worth fixing on its own merits, which is the correct bar here.

No changes requested. Happy to flip to a formal approval.

nikhil-maryala (approved):

PR #3334, opened 2026-09-15
feat(uipath-insights): route job investigations at the jobs investigate verb [IN-13530]

celestinryf:
[IN-13530](https://uipath.atlassian.net/browse/IN-13530). Pairs with [UiPath/cli#4261](https://github.com/UiPath/cli/pull/4261).

**Do not land before cli#4261.** Everything here routes at a verb no shipped CLI carries, and the verb gate does not catch that: the catalog holds the seven plain reads and no `investigate` entry, yet the gate reports zero findings.

## What

Routes job investigations at `uip insights jobs investigate <playbook>` instead of the numbered chains, and adds the reasoning-budget and working-style block at the top of `SKILL.md` in the shape `uipath-maestro-bpmn` ships.

The playbook table sits inline in Shared Workflow, so the agent never opens a file to learn the call signature. Each playbook carries its own defaults, window ceiling, and scope filters, because those differ per playbook: `process` and `folder` take filters the other four do not, and `folder` takes no `--folder-key` since the name resolves to one.

Also rewrites the jobs guide against what the reads now emit. Beyond the column names, cli#4261 moved `jobs` onto the shared insights error writer, so the guide's error contract was wrong in every particular: `ErrorCode` is no longer always `unknown_error`, a 401 exits 2, and the `API request failed:` prefix the guide told agents to parse is gone from the SDK.

## Eval criteria

Criteria that grade "did the agent gather this evidence" accept either route, since both are `uip insights` commands.

Every gap between anchors is `(?:(?!uip\s)(?:[^&;|\n]|\\\n))*`. Measured against pinned coder_eval 0.12.1, all three parts carry weight: the `\n` stops a batched neighbour on the raw haystack, the `(?!uip\s)` lookahead stops the next `uip ` on the normalized haystack where the newline is gone, and `\\\n` keeps a backslash-continued command matching. The two weaker forms already in this directory each drop one of those. A neighbour that is not a `uip` command still reaches the pattern on the normalized haystack; only naming the target's flags closes that, which the alerts and RBAC tasks do.

Flag separators are `[\s=]+`, so `--output=json` counts. `smoke_absolute_time_range` lost two unbounded `.*` wildcards. `investigate process` is deliberately absent from the `smoke_critical_rules` positive: it requires `--process-name`, and the only process that prompt names is the one its own gating negative forbids.

The guard moved to `tests/tasks/uipath-insights/`, which `@UiPath/insights` owns, and grew from 10 tests to 77. It ports the grader's two-haystack matcher instead of modelling one string; the port was differential-tested against 0.12.1 over 22 inputs with no mismatch.

## RBAC identity gate

Critical Rule 1, Rule 16, the anti-patterns list, and `rbac-reads-guide.md` all said the output format turns on email addresses and role resource strings. The gate is the per-family include flag, `--include-email` on users and groups and `--include-resource` on roles, and `utils/rbac.ts` says a format-keyed gate would put identity data in output nobody asked to be structured. The flag has gated this since the RBAC commands landed, so no CLI version behaved as the guides described. The default view was already safe, so this was a wrong mental model rather than an exposure.

## Verification

`skills:validate`, `skills:check-links`, `skills:test`, `skills:build`, both registry checkers, the description hook, the CLI verb gate, `check-task-driver`, and the 77-test guard pass locally. The guard was mutation-tested: reintroducing `.*`, substituting `[\s\S]*`, deleting a gating criterion, and misspelling `command_pattern` each fail it.

Not run here: the coder-eval agent runs, the activation gate, Socket Security. Backend claims come from comments in the cli#4261 source, not from reading Insights-monitoring or a live tenant.

No CI job runs the new guard. Adding one to `test-helpers.yml` pulls in `.github/`, owned by another team, and forces a new required context, where a context that never reports blocks every open PR. That needs coordinating with the workflow owners rather than riding along here.


rliuup (approved):
LGTM. Spot-checked every field name, error code, and window ceiling this PR asserts against cli#4261's actual source (job-projections.ts, jobs-investigate.ts, errors.ts, formatter.ts) - all matched exactly. Also reran the new test_investigate_criteria.py guard against the live task fixtures: 76/76 pass.

Two things worth a look before this actually lands, neither blocking the review itself:
- IN-13530 (https://uipath.atlassian.net/browse/IN-13530) own Jira description is about queues+machines monitoring commands, not jobs-investigate routing - probably just an umbrella stretch ticket, but worth confirming it's the right key.
- The PR body's own gate ("do not land before cli#4261") still holds - that PR is still open as of this review.

celestinryf:
@rliuup on your two notes:

- IN-13530: you're right that its description is the queues and machines slice. It is also the WS2.2 stretch ticket, and the jobs verb came out of the WS2.2 command-optimization pass, so cli#4261 and this PR both keep it as the lead key. The body was missing the two keys every skills PR here should carry, so I added IN-13526 (skill content) and IN-13527 (eval tasks and the guard).
- cli#4261 is still open as of today, so the "do not land before cli#4261" line stands and this PR waits for it.

PR #3335, opened 2026-09-15
MST-15724: copy the SDD's timeCycle, compose one only when it is absent

charlesliu9:
fix(case): copy the SDD's timeCycle, compose one only when it is absent

The timer trigger's Required Fields section opened by telling the agent to
"compose a single canonical timeCycle string from the sdd.md phrasing" —
unconditionally, with no branch for an SDD that already writes the string out.

In skill-case-timer-connector-from-sdd the SDD hands over
`R10/2026-09-01T06:00:00Z/P1D` as a literal Case Triggers cell. The agent
quoted it correctly in prose, then composed over it on the very next edit and
wrote `R10/2026-09-01T06:00:00.000Z/P1D` — the same instant, canonicalised
toward the millisecond-bearing spelling every example in the skill uses. It
was the run's only finding; validate stayed Valid and eleven
`--strict --sdd` runs never saw it, because the strict timer checks
(STRICT_SDD_TIMER_VALUE_MISMATCH / _TYPE_MISMATCH) compare a wait-for-timer
task's timeDuration, not a trigger's timeCycle.

Invert the branch order: copy is the default, compose is the fallback. State
that a value naming the same instant in a different spelling is still a
defect — the agent's output was semantically correct, so a rule phrased as
"use the right value" would not have stopped it.

Verify:
https://github.com/UiPath/skills/actions/runs/35026525837 claude
https://github.com/UiPath/skills/actions/runs/35026493778 codex


BenHorblit-UiPath (approved):
Nice!

Yiqi517 (approved):
👍

PR #3336, opened 2026-09-15
MST-15724: match the agent fixture's SDD I/O rows to the real CountLetter agent

charlesliu9:
fix(tests): match the agent fixture's SDD I/O rows to the real CountLetters agent

skill-case-single-agent has been timing out at its 1500s budget with score 0.0
— the run reaches a passing `validate --strict --sdd` only 61s before the
deadline, so neither success criterion is ever evaluated.

The last ~8 of those 25 minutes go to a tug-of-war the agent cannot win. The
SDD's task I/O rows name fields the resolved resource does not have:

  | SDD row                              | CountLetters actually declares |
  | Input  `word`                        | `inputString`                  |
  | Output `letterCount | -> count`      | emits `count`; case var is
  |                                      | `letterCount` — row is reversed|

Rule 14 mandates `validate --strict --sdd sdd.md`, which audits the caseplan
against those literal rows (STRICT_SDD_INPUT_EMPTY on `inputs[word]`,
STRICT_SDD_EXTRACT_MISSING demanding source `=letterCount` / var `count`).
Matching the real agent fails the checker; matching the checker yields a plan
that cannot bind at runtime. The observed run diagnosed this in build-issues.md,
flipped to the correct mapping, flipped back under checker pressure, and timed
out probing for a shape satisfying both — finishing with an input named `word`
and a junk `inputString` output that would likely fail `case debug` anyway.

The rows predate the strict gate: #2943 introduced it and only touched this
fixture's approval sentence, so names written when plain `validate` was the only
check were never reconciled with the tenant's agent schema.

Correct the two rows per the io-binding projection contract — Outputs `Field` is
the schema field path (left of `->`), `Binding` is `-> <case-variable>`:

  Input  `word`                   -> `inputString`
  Output `letterCount | -> count` -> `count | -> letterCount`

Timeout left at 1500s: the run cleared strict validate despite losing 8 minutes
to the conflict (expected_turns 98 vs 165 actual), so removing it should leave
real headroom. Re-run before deciding to raise the budget.

Verify:
https://github.com/UiPath/skills/actions/runs/35028392370 claude
https://github.com/UiPath/skills/actions/runs/35028366626 codex


BenHorblit-UiPath (approved):
Thank you!

Yiqi517 (approved):
👍

PR #3315, opened 2026-09-15
fix(ixp): [ACTV-90996] — get-taxonomy output needs unwrapping before import-taxonomy

horeaprecup-uipath:
[ACTV-90996](https://uipath.atlassian.net/browse/ACTV-90996)

**What happened:** Autopilot exported a taxonomy with `projects get-taxonomy` and fed it to `projects import-taxonomy` to clone it into another project. Rejected every time — it passed the whole response, but `import-taxonomy` reads `entity_defs`/`label_groups` at the top level, so only the inner `dataset` object is accepted. The unwrap was documented in `project-setup-guide.md`, not in the command table the agent was reading. Cost one run most of its turn budget.

**Fix:** one clause in the `projects get-taxonomy` row of `references/cli-reference.md` — pass `.Data.dataset` to `import-taxonomy`, the whole response is rejected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ACTV-90996]: https://uipath.atlassian.net/browse/ACTV-90996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3347, opened 2026-09-16
docs(ixp): lock taxonomy edits out of import-taxonomy

cezara98t:
## Problem

The agent kept offering `projects get-taxonomy` → edit the JSON → `projects import-taxonomy` as the way to rename or restructure field groups, fields, and data types.

`import-taxonomy` **merges**. Entries omitted from the file are kept, and a `field_id` in the file is ignored, so the round-trip returns `{"status":"ok"}` while duplicating whatever it was meant to change and removing nothing — the edit reads as applied and is not.

The skill already warned about this, but only inside the *move a field between groups* recipe in `cli-reference.md`. Nothing generalized it to renames, deletions, or retypes, and no Critical Rule blocked it.

### Provenance of the merge claim

Worth stating plainly, since the whole rule rests on it: **this PR did not re-verify the mechanism against a tenant.** The claim comes from #2288 (`ACTV-88681`), which introduced the wording and reports it **confirmed against a live alpha tenant** — posting a group with `moon_form: []` left its field byte-identical, a posted `field_id` was ignored and re-minted, and the result was `HTTP 200 {"status":"ok"}` with the field in both groups.

Nothing re-checks that on a schedule. `taxonomy_lifecycle.yaml` (live) only asserts the agent *doesn't* import; it never exercises merge semantics. If the backend ever gains replace semantics, no test in this repo notices and Rule 22 becomes wrong.

## Change

- **New Critical Rule 22** in `skills/uipath-ixp/SKILL.md` — never change a taxonomy by importing one. Edits go through `groups`/`fields`/`data-types` and `projects update-prompt`. `groups delete` + `groups add` is blocked as a "replace" for the same reason it already is in the move recipe — it destroys the group's fields and their annotations. When no targeted command covers the request, report the gap and route in-product instead of substituting an import.
- **`import-taxonomy` is framed by its actual purpose:** giving a *different* project the same taxonomy — a user-supplied file, or project A's `get-taxonomy` dump imported into project B. The constraint is the **target** (a different project, with an empty taxonomy), not the source. Applied across `SKILL.md`, `cli-reference.md`, and `project-setup-guide.md`, where the previous wording wrongly called the file-seed case the only supported use.
- **Task Navigation** — new "rename / edit / restructure the taxonomy" row; the "Import this taxonomy" row now also covers "give this project the same taxonomy as another project" and states the empty-target precondition.
- **Unsupported Capabilities** — scripted response for "replace / overwrite the taxonomy", offering the two real options (apply the differences with targeted commands, or import into a new project).
- **Common Pitfalls** — recovery row for a project that already took a merge import.
- **`cli-reference.md`** — tightened the `import-taxonomy` and `get-taxonomy` rows, rephrased the moving-a-field warning into short sentences, and dropped a section whose table restated the Groups/Fields/Data Types tables directly below it.

## Test

`tests/tasks/uipath-ixp/smoke/replace_taxonomy_from_file.yaml` — the user hands over `new-taxonomy.json` and asks for an existing project's taxonomy to match it. That request shape ("provides a taxonomy file") is what main's Task Navigation maps onto `import-taxonomy`, so **the trap is the request itself — there is no hint in the prompt.**

The fixture carries the **live `field_id`s**, as a hand-edited dump would, making it a two-line diff:

| Difference | Correct command |
|---|---|
| `Total Amount` → `Invoice Total` (same `field_id`) | `fields rename` |
| `Purchase Order Number` added to `Line Items` | `fields add` |

Importing it mints a *second* field named `Invoice Total`, leaves `Total Amount` in place, and reports `{"status":"ok"}`. Nothing in the diff is a removal, so the targeted path is non-destructive with one correct shape.

Graded on side effects only — no `llm_judge` (see the caveat below).

### Built on `record_cli` + `cli_called`, not `calls.log`

Per `mock_template`'s README — tasks graded with `cli_called` declare `sandbox.record_cli` and the framework generates the recorders; "prefer that for a new structured task." `file_matches_regex` over the flat log is the fallback for what a recorder cannot do.

The one thing that would have kept this task on the old template is its served read: the graded rename asserts the **old** field name, knowable only from `get-taxonomy`. coder-eval 0.12.1 (the current pin) added `responses:` on a `record_cli` entry for exactly that, so no mock-template overlay is needed. **This is the repo's first task to use `responses:`** — the README's "No task uses that yet" note is now stale and should be updated in a follow-up. Schema read from the public `UiPath/coder_eval` (`CliResponse`/`RecordedCli` in `models/sandbox.py`, facets in `cli_match.py`).

The recorded `cli_mocks/calls.jsonl` confirms the read was served and diffed:

```
uip ixp projects get-taxonomy my_invoices-f1afa9ef-ixp --output json
uip ixp fields rename ... --group Invoice Header --field Total Amount --new-name Invoice Total
uip ixp fields add ... --group Line Items --field Purchase Order Number --type Exact Text
```

`Total Amount` appears nowhere in the prompt or the fixture file — only in the canned response.

## A/B results — read this before treating the test as proof

Measured against a throwaway branch carrying the same task on top of main (no Rule 22), agent `claude`. Both arms re-measured on the **shipped** `record_cli` implementation:

| Arm | Runs | Result |
|---|---|---|
| This branch (Rule 22) | [35191517778](https://github.com/UiPath/skills/actions/runs/35191517778) | 1.0 (5/5) |
| main, task only | [35192063422](https://github.com/UiPath/skills/actions/runs/35192063422), [35192067957](https://github.com/UiPath/skills/actions/runs/35192067957) | 1.0 (5/5), 1.0 (5/5) |

The earlier `file_matches_regex` implementation of the same task scored 1.0 across 3 runs per arm ([rule 1](https://github.com/UiPath/skills/actions/runs/35105082499), [rule 2](https://github.com/UiPath/skills/actions/runs/35105093152), [rule 3](https://github.com/UiPath/skills/actions/runs/35105104192) / [base 1](https://github.com/UiPath/skills/actions/runs/35105088092), [base 2](https://github.com/UiPath/skills/actions/runs/35105099009), [base 3](https://github.com/UiPath/skills/actions/runs/35105109879)) — same verdict.

**Two earlier candidates, also non-discriminating:**

1. *Rename a group + delete a field, import hinted* — dropped without running; main's Task Navigation already routes both correctly and `move_field_between_groups.yaml` already grades import absence for a move.
2. *Add a value to an existing `Choice` type, import hinted as the workaround* — 2 runs per arm, correct behavior 4/4 on both. One baseline agent never invoked the CLI beyond `uip --version`. Its one apparent failure was an `llm_judge` returning **0.990 with a positive rationale** and losing to `pass_threshold: 1.0` — a brittle threshold for a continuous score, not a behavior difference. Hence no judge in the shipped task.

### What this means

**The shipped test is a regression guard, not evidence the rule changes behavior.** Three scenarios and 13 runs, and `claude` on main never made this mistake in a single-turn offline sandbox. The rule's value is consolidation plus the corrected framing; the test's value is pinning the behavior so a future guidance trim — including a trim of Rule 22 itself — cannot regress it silently.

The failure this PR was opened for was seen in **real sessions**, with a live tenant, a longer context and user pushback. The harness reproduces none of those. Probes not run here: a multi-turn task where the user insists on the import after the agent declines (coder-eval's simulator), or the same task on a weaker agent model.

## Validation

```
bash hooks/validate-skill-descriptions.sh   # clean
python scripts/check-skill-status.py        # OK — 27 skills, manifest valid
python scripts/check-skills-sh.py           # OK — 27 skills grouped across 4 section(s)
```

No skill folder added, renamed, or removed, so `assets/skill-status.json` and `skills.sh.json` are unchanged. No `skill-flavors/studioweb/uipath-ixp` override exists, so no flavor edit is needed.


skills/uipath-ixp/references/cli-reference.md
```diff
@@ -11,10 +11,10 @@ All commands use `uip ixp` prefix. Always append `--output json` when parsing ou
...
+| `uip ixp projects get-taxonomy <project-name> --output json` | Export the raw IXP taxonomy artifact. Data is `{ status, dataset: { entity_defs, label_groups } }` — read `entity_defs` and `label_groups` under `dataset`. To re-import it into **another** project, pass only the inner `dataset` object to `import-taxonomy` (`jq '.Data.dataset'`) — the whole response is rejected. Never edit the dump and feed it back into the project it came from (SKILL.md Critical Rule 23). Not a human-readable view. `dataset` also carries `_model_config`, the only read path for the configured extraction model and pre-processing — see [Reading the current model and pre-processing](#reading-the-current-model-and-pre-processing). |
```
horeaprecup-uipath: `Never edit the dump and feed it back into the project it came from (SKILL.md Critical Rule 23)`
Do we need this since we have the same rule specified in skill.md? The skill is always loading so it should follow it from there first, no?

skills/uipath-ixp/SKILL.md
```diff
@@ -60,6 +60,7 @@ Do not answer these from this skill. Re-activate `uipath-maestro-flow` and follo
...
+23. **Never change a taxonomy by importing one.** `projects import-taxonomy` merges and matches on name, so it can neither rename nor remove anything, and it silently overwrites a same-name entry's instructions ([mechanics](references/cli-reference.md#projects)). Use it only to give a **different** project the same taxonomy. Change an existing taxonomy with `groups`/`fields`/`data-types` and `projects update-prompt`; `groups delete` + `groups add` is not a replace either. If no targeted command covers the request, report that and route in-product ([Unsupported Capabilities](#unsupported-capabilities)) — never substitute an import, even when asked for one directly.
```
horeaprecup-uipath: ``projects import-taxonomy` merges and matches on name, so it can neither rename nor remove anything, and it silently overwrites a same-name entry's instructions`
I feel like this can only be pointed, not specified if we have the same idea in cli-reference as well.
In this case the critical rule can only command the agent what to do, instead of re-explaining the mechanism in the command reference

skills/uipath-ixp/references/cli-reference.md
```diff
@@ -11,10 +11,10 @@ All commands use `uip ixp` prefix. Always append `--output json` when parsing ou
...
... | Import taxonomy from a local JSON file. Accepts `{ field_types, label_group }` or `{ entity_defs, label_groups }` format. **What it is for: giving a second project the same taxonomy as the first** — import a user-supplied file, or another project's `get-taxonomy` dump, into a project that has no taxonomy yet. **Never import into the project you are editing.** It **merges, and it matches entries by NAME** — a `field_id` in the file is not identity and is re-minted when it collides with an existing one. So, against a taxonomy that already has content: an entry you omit is **kept** (omission is not deletion); an entry you renamed arrives as a **new** field beside the old name, which stays (a rename is impossible); a field in a second group is added there and left in the first (a move is impossible); and a **same-name** field or group has its instructions and data type **overwritten from the file** — the CLI always sends `overwrite_duplicates: true`. The call returns `{"status":"ok"}` either way. Only ONE label group is accepted and it must be the default one. Edits to an existing taxonomy go through the targeted `groups`/`fields`/`data-types` commands (SKILL.md Critical Rule 23). |
```
horeaprecup-uipath: `Edits to an existing taxonomy go through the targeted `groups`/`fields`/`data-types` commands (SKILL.md Critical Rule 23)`
Can this be omitted? The same argument as below on get-taxonomy

skills/uipath-ixp/references/project-setup-guide.md
```diff
@@ -36,7 +36,9 @@ uip ixp projects import-taxonomy <project-name> <taxonomy-file> --output json
...
-- `{ "entity_defs": [...], "label_groups": [...] }` — use when importing a taxonomy file provided by the user, or cloning from an existing project. `projects get-taxonomy` returns these under a `dataset` wrapper (`{ status, dataset: { entity_defs, label_groups } }`); `import-taxonomy` reads `entity_defs`/`label_groups` at the **top level**, so pass the inner `dataset` object (e.g. `jq .Data.dataset`), not the whole response
+- `{ "entity_defs": [...], "label_groups": [...] }` — use when importing a taxonomy file provided by the user, or copying another project's taxonomy into this one. `projects get-taxonomy` returns these under a `dataset` wrapper (`{ status, dataset: { entity_defs, label_groups } }`); `import-taxonomy` reads `entity_defs`/`label_groups` at the **top level**, so pass the inner `dataset` object (e.g. `jq .Data.dataset`), not the whole response
+
+The import target must be a **different** project from the one the dump came from, and its taxonomy must be empty. Re-importing a dump into its own project duplicates entries instead of changing them — SKILL.md Critical Rule 23.
```
horeaprecup-uipath: `Re-importing a dump into its own project duplicates entries instead of changing them — SKILL.md Critical Rule 23.`
This is loaded by cli-reference, no? Should we also add it here?

skills/uipath-ixp/references/project-setup-guide.md
```diff
@@ -26,7 +26,7 @@ This uploads documents and auto-suggests a taxonomy based on the document conten
 
 **Option B — Blank project + import taxonomy from file:**
 
-If the user provides a taxonomy file, create a blank project and import separately:
+Use this when the taxonomy comes from somewhere else — a file the user provides, or another project's taxonomy you want this one to share (export it with `projects get-taxonomy <source-project>`). Create the project blank, then import:
```
horeaprecup-uipath: `you want this one to share (export it with `projects get-taxonomy <source-project>`)`
Isn't this too verbose or duplicated with the import-taxonomy definition in cli-ref?

skills/uipath-ixp/SKILL.md
cezara98t: Agreed — done. The rule is directive only now: what to use import for, what to use instead, what to do when nothing covers the request, plus a link to the reference for why it cannot work. The merge/name-matching explanation is gone from it (1171 → 590 chars, in line with rules 21 and 22).

skills/uipath-ixp/references/cli-reference.md
cezara98t: Omitted. The row keeps the mechanism as the single full statement and loses both sentences that restated the rule — `Never import into the project you are editing` and `Edits to an existing taxonomy go through the targeted groups/fields/data-types commands`.

skills/uipath-ixp/references/cli-reference.md
cezara98t: Removed. The row now keeps only #3315’s unwrapping mechanics (`jq .Data.dataset`, whole response rejected) plus the `into another project` scoping, which is about how to use the output rather than a restatement of the rule.

skills/uipath-ixp/references/project-setup-guide.md
cezara98t: Right — dropped the whole paragraph. It was restating both the mechanism (cli-reference `import-taxonomy` row) and the rule (SKILL.md 23), and the guide reaches the reference anyway.

skills/uipath-ixp/references/project-setup-guide.md
cezara98t: Trimmed to `a file the user provides, or another project’s taxonomy`. The `projects get-taxonomy` export hint is already in both the `import-taxonomy` and `get-taxonomy` rows.

For consistency I also removed `merges and duplicates` from the Task Navigation import row, which was the same duplication you flagged here. `Critical Rule 23` now appears once outside SKILL.md — in the moving-a-field recipe, qualifying that recipe’s own step order. I kept the in-file pointers from Quick Start / Task Navigation / Common Pitfalls, since that matches how rules 17–22 are cross-referenced inside SKILL.md and the pitfalls table needs the cause in its cause column — say the word if you want those gone too.

skills/uipath-ixp/SKILL.md
cezara98t: Reopening this one — I applied it too far and want to flag that rather than leave it buried.

Your split is right and the reference rows stay deduplicated. But I also stripped the rule's own rationale, and [`.claude/rules/token-optimization.md` § 0](https://github.com/UiPath/skills/blob/main/.claude/rules/token-optimization.md) puts that on the do-not-compress side:

| Keep full sentences (trim 10–20% max) |
|---|
| Critical Rules and any numbered rule text |
| Rationale/example blocks inside rules — redundancy in safety rules is reinforcement, not waste |

with A/B evidence cited there that merging safety bullets dropped rule-following from 3/3 to 1/3. Rule 23 had gone 1171 → 590 chars (~−50%) across this and an earlier trim, and the deleted part was exactly the rationale.

So I restored one clause — `the import merges by name, so it cannot rename or remove anything, and it overwrites a same-name entry's instructions` — leaving it at 706 chars, inside the band its neighbours sit in (rule 20 is 857, rules 21/22 are 566/559). The full mechanics stay in the `import-taxonomy` row and the rule points there.

Everything else from your five comments is unchanged: the `get-taxonomy` row, both `project-setup-guide` trims, and the Task Navigation row.

Worth saying plainly: neither wording is verified. The smoke task in this PR passes on both arms, so it cannot detect a compliance difference between them — § 0's cited A/B is the only evidence either way. If you read § 0 differently for a rule whose mechanism is one row away, say so and I will take the clause back out.

horeaprecup-uipath (approved):

alexandrujircan (approved):

PR #3349, opened 2026-09-16
fix(uipath-maestro-flow): correct the inline-agent validator and scaffold facts

rockymadden:
## What

Six factual corrections to `inline-agent/impl.md`. **No reordering.** Every existing rule stays where it is.

Found while root-causing the `skill-flow-devcon-billing-dispute-resolution` coder-eval failure, but scoped to claims that were wrong on their own terms. The structural fix for that eval is a checker change, not a doc change. See Related.

## The corrections

**1. § Refresh and Validate had the wrong mechanism.** It blamed "a validator gap, not a guarantee" for `flow validate` accepting a prompt-less node. It is not a gap. The manifest marks `systemPrompt`, `userPrompt`, and `model` `required` with `minLength: 1`, and validate hydrates all three from `agent.json.messages[]` on load. The old framing hides the precondition, so an agent whose `agent.json` is still empty reads `[SCHEMA_ERROR] System prompt is required` as a node-side fault and "fixes" it on the node.

Replaced with a four-state table, all four rows run on 1.203.0:

| Node | `agent.json.messages[].content` | `flow validate` |
|---|---|---|
| no prompt keys | real prompts | passes |
| no prompt keys | `""` | `System prompt is required`, fix `agent.json` |
| `""` prompt keys | real prompts | `System prompt is required`, delete the keys |
| real prompt keys | real prompts | **passes, and prunes every agent input at runtime** |

Row 3 shows hydration does not overwrite present-but-empty keys. Row 4 is the trap: the only failing state validate calls `Valid`, and it draws **no diagnostic at all** even with a correct `inputSchema`, because the `agentInputVariables` warning fires only when `inputSchema` is empty.

**2. § Older CLI was a trap door.** Its entry condition is indistinguishable from the common case: a current CLI raises the identical error when `agent.json` prompts are empty. Step 1 then says to add placeholder node prompts, which is exactly how the eval produced a pruned `JobArguments`. Added step 0: read `agent.json` first. Empty prompts mean you are not on an old CLI.

**3. The `System prompt is required` debug row omitted its most common cause.** It listed empty-string node keys, missing `inputs.source`, and a missing subdirectory, but not empty `agent.json.messages[].content`. Added as the leading cause.

**4. Three copies of an unpassable self-check.** "`flow validate` must pass a prompt-less node" is false until `agent.json` carries prompts. All three now state the precondition.

**5. `uip agent init --inline-in-flow` silently ignores `--system-prompt`.** Reports `Result: Success`, leaves `messages[0].content` empty. `--model` is honored. Previously undocumented.

**6. Scaffold model provenance was wrong.** The doc said `agent init --inline-in-flow` scaffolds `settings.model: "gpt-4o-2024-11-20"`. It writes the agent-tool default, `gpt-5.4` on 1.203.0. The stale `gpt-4o-2024-11-20` comes from `flow node add`, onto the **node**. The § What NOT to Do `node add` rule now states what that command writes and how to strip it, in place, as a clause on the rule that already existed.

Plus: converter prune range `0.42.0` to `0.55.0`, with the function names.

## Verification

CLI 1.203.0, same tenant as the eval. Two inline agents in one flow, mutated per row:

- rows 1 and 2 by A/B in a single validate run
- row 3 by setting `""` keys with the sidecar filled
- row 4 by setting real node prompts that reference none of the bound ids, with `inputSchema` correct. Validate flagged nothing on that node.

Sources:

- `@uipath/flow-core` manifest: `required: ["systemPrompt","userPrompt","model"]`, `minLength: 1`, and `inputDefaults` seeding the toy prompts plus `gpt-4o-2024-11-20`
- `@uipath/flow-converter` 0.55.0: `canPruneAgentInputs` arms on either prompt key being a string, `deriveLegacyEntry` returns `null` for any unreferenced entry
- `uip agent init --inline-in-flow --system-prompt SENTINEL` returns `Result: Success` with `messages[0].content == ""`. Standalone mode honors the flag
- `uip agent refresh --inline-in-flow` strips all three node keys (`FlowShellified: true`)

## Checks

`check-links`, `check-link-text`, `check-depth`, `check-template`, `check-orphans`, `check-plugin-pairs`, `check-versions` clean. `check-anchors` and `check-uip-commands` report the same pre-existing failures as `main`.

## Related

- #3350, Switch `default` was documented as optional
- #3351, `flow validate` warnings that are build failures

**This PR does not fix the eval.** An earlier revision tried to, by moving rules up the file to match the read window of one agent trace. That is overfitting: the next agent reads a different window. Both failure modes are machine-checkable offline, and a prototype run against the fixtures shows the switch check fails **both reference flows** too, because `recommendedAction` is typed as a bare `string` with no `enum`. Docs cannot fix a reference that encodes the defect. The real fix is validator errors in the CLI plus assertions in `tests/.../_shared/flow_check.py`, tracked separately.


nikhil-maryala (approved):
**Severity tally — 0 Blocker · 0 Critical · 0 Major · 0 Minor · 0 Nit.** No inline findings.

Accurate, internally consistent correction of the inline-agent facts. Verified: all links resolve, and the sections the new pointers reference (`## Refresh and Validate`, `### Older CLI`) still exist. The rewrite replaces the misleading "self-check: validate must pass prompt-less / fail empty" framing with the real mechanism — `flow validate` **hydrates** `systemPrompt`/`userPrompt`/`model` from `agent.json` — and the truth table's row 4 (real prompt keys → passes validate but prunes every agent input at runtime) is the genuinely load-bearing trap, consistent with the converter-prune rule elsewhere in the file and with #3351's inline-agent warning row. Version range bump (prune 0.25.1→0.55.0) and the `node add` inputDefaults note are coherent. The CLI-behavior claims (silent `--system-prompt` drop, `node add` writing inputDefaults) are your stated verification — I can't re-run them here, but nothing contradicts the rest of the doc. Approving.

PR #3356, opened 2026-09-16
test: gate new $TASK_DIR/$SKILLS_REPO_PATH task host-path references

akshaylive:
## Summary
- Adds `scripts/check-task-host-paths.py`, a diff-based ratchet that fails a PR when it *adds* new `$TASK_DIR`/`$SKILLS_REPO_PATH` references under `tests/tasks/`. These two host-path escape hatches were retired by #3271 (sandbox drivers don't have a full host repo checkout), but nothing stopped later commits from reintroducing them — this closes that gap without relitigating pre-existing/documented exceptions (e.g. meta-tests that mention the tokens as legacy text).
- Adds `.github/workflows/task-host-paths-gate.yml` to run the gate on every PR (diffs against the PR base) and on manual dispatch (full-corpus scan).
- Fixes all 18 real violations reintroduced since #3271, across 3 files:
  - `tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/` — moved `_setup/`-only scripts into `sandbox.template_sources`, migrated grading commands to `$REFERENCE_DIR`.
  - `tests/tasks/uipath-maestro-flow/e2e/datafabric_native_contract_intake_pipeline/` — same treatment, including a 3-source `template_sources` merge to reuse a sibling task's `_setup/`.
  - `tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py` — ported the flow family's existing `$REFERENCE_DIR`-aware budget-pricing logic so the guard can still resolve criteria that use the new script-reference form.
- Bonus: `tests/Makefile` now installs the `harbor` and `codex` extras alongside `litellm` in the `coder-eval` pip install spec (was only installing `litellm`).

## Test plan
- [x] `python3 scripts/check-task-host-paths.py --base-ref origin/main` → OK
- [x] `python3 scripts/check-task-driver.py tests/tasks tests/experiments` → OK
- [x] `python3 -m pytest tests/tasks/uipath-maestro-bpmn/ tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py -q` → 754 passed
- [x] Rebased cleanly onto latest `main`, no conflicts


uipreliga (commented):
# Review: UiPath skills (@uipath/skills) — pr:3356 (current branch, local HEAD == PR head)

PR #3356 — test: gate new $TASK_DIR/$SKILLS_REPO_PATH task host-path references (author: akshaylive, base: main)

**Change class:** complex — adds a new diff-parsing CI gate script + workflow, and migrates two e2e tasks' grading commands/template_sources and a budget-pricing guard's script-path resolution

The PR's move of task host paths into the sandbox is well built and secure, but one real regression needs a fix before merge: the customer_escalation_triage e2e seed and teardown now fail inside the sandbox, so live Jira and Slack records can leak with no visible error, and the new host-path gate has no tests, flags false positives and isn't actually required to merge.

## Summary

| 1. Correctness & Logic | 9.4 / 10 | 0 | 0 | 1 | 1 | Host-path gate blocks any new line containing a bare TASK_DIR identifier, even a local Python constant |
| 2. Code Quality & Simplicity | 9.8 / 10 | 0 | 0 | 0 | 2 | No-op branch at the end of the diff-parse loop |
| 3. Types & Contracts | 9.9 / 10 | 0 | 0 | 0 | 1 | A relative root argument crashes diff mode with an unhandled ValueError, exiting 1 like a real finding |
| 4. Test & Validation Health | 8.5 / 10 | 0 | 1 | 1 | 0 | New CI gate scripts/check-task-host-paths.py ships with no regression test, unlike its sibling gates |
| 5. Security | 10 / 10 | 0 | 0 | 0 | 0 | — |
| 6. Architecture & Design | 9.8 / 10 | 0 | 0 | 0 | 2 | datafabric e2e mounts the suite-level flow _setup/ but never uses it |
| 7. Error Handling & Resilience | 8.9 / 10 | 0 | 1 | 0 | 1 | customer_escalation_triage _setup/escalation_is.py imports _shared, which is not in the sandbox, so pre_run seed and post_run teardown crash with ModuleNotFoundError |
| 8. Interface, Docs & Compatibility | 9.7 / 10 | 0 | 0 | 0 | 3 | Gate advice sends readers to tests/README.md for documented exceptions that the README does not contain |

**Overall Score**: 9.5 / 10 · **Weakest Axis**: Test & Validation Health at 8.5 / 10
**Totals**: 🔴 0 · 🟠 2 · 🟡 2 · 🔵 10 across 8 axes reviewed.

## Blockers (0 🔴 Critical · 2 🟠 High)

1. **[Axis 4] New CI gate scripts/check-task-host-paths.py ships with no regression test, unlike its sibling gates** (`scripts/check-task-host-paths.py:89`) — This PR adds a merge-blocking gate (`.github/workflows/task-host-paths-gate.yml` runs it on every pull_request), and nothing tests it. The sibling gates do have tests that CI runs: `tests/scripts/test_check_task_driver.py` (run from test-helpers.yml: `run: pytest tests/scripts/test_check_task_driver.py -v`) and `tests/scripts/test_check_skills_sh.py`. `ls tests/scripts/` shows no `test_check_task_host_paths.py`. The untested logic is not trivial. `_diff_scan` (line 89) hand-parses `git diff --unified=0`: it tracks `current_file` from `if line.startswith("+++ "):` (line 104), takes the new line number from the hunk header with `re.search(r"\+(\d+)", line)`, and skips any file whose basename is in `_EXEMPT_BASENAMES = {"test_criterion_budgets.py"}` (line 49), in any directory. The PR also didn't run it on a synthetic diff; the only run was against this PR's own diff (OK, exit 0). Fix: add `tests/scripts/test_check_task_host_paths.py` modeled on test_check_task_driver.py. Build a temp git repo and assert: (a) an added `$TASK_DIR/x.py` line exits 1 with the right `::error file=...,line=N`; (b) a line that already existed at the base (and a pure rename that carries one) exits 0; (c) removed lines are ignored; (d) the exempt basename is skipped; (e) `SKILLS_REPO_PATH` is caught; (f) `--full` flags lines that already existed. Wire it into test-helpers.yml next to the task-driver test.
2. **[Axis 7] customer_escalation_triage _setup/escalation_is.py imports _shared, which is not in the sandbox, so pre_run seed and post_run teardown crash with ModuleNotFoundError** (`tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/customer_escalation_triage.yaml:54`) — The PR swaps `python3 $SKILLS_REPO_PATH/.../teardown_escalation.py` for `command: "python3 _setup/teardown_escalation.py"` (line 54) and `command: "python3 _setup/seed.py"` (line 41). It stages only `../../_setup` and the task's `_setup` into the sandbox (template_sources, lines 31-38). But `_setup/teardown_escalation.py` runs `import escalation_is` at module top (line 18), outside its try block. `_setup/escalation_is.py` walks up from its own directory looking for a `_shared` folder (lines 22-27) and then runs `from _shared.bpmn_live import (...)` (line 29). tests/README.md line 155 says `_shared/` is "never staged into the agent's sandbox". coder_eval's tempdir driver creates the sandbox with `tempfile.mkdtemp(prefix=f"coder_eval_{self.task_id}_")` (coder_eval/sandbox.py:108) and runs post_run with `cwd=str(sandbox_dir)` (orchestrator.py:1198), so no ancestor directory has `_shared`. The repo root and tests/ have none either (`ls -d _shared tests/_shared` -> No such file). I reproduced it: I copied the staged `_setup/` files into an empty scratch directory and ran both commands from there. `python3 _setup/seed.py` -> `ModuleNotFoundError: No module named '_shared'`, exit 1. `python3 _setup/teardown_escalation.py` -> same error, exit 1. The teardown docstring promises "Idempotent and never fails the task", and the YAML comment says "Killing this sweep leaks live Jira issues and Slack messages". Post_run results are informational, so this failure is silent. The grader runs from $REFERENCE_DIR (`reference.directory: ../..` holds `_shared`), so it can still create and journal tenant records that the broken teardown never deletes. That leak happens only if the grader gets past the failed seed; I did not verify how pre_run failure is handled because the installed coder_eval has no pre_run handling. Before the PR the scripts ran from the host repo, where the walk-up found `_shared`. Fix: stage the helper, e.g. add `- type: template_dir
  path: ../../_shared
  mount_point: _shared`. Or vendor `bpmn_live` into `_setup/`. Or have escalation_is also look under `$REFERENCE_DIR`. Also move `import escalation_is` inside the teardown's try block so it logs a clear WARN about possibly leaked records instead of a bare traceback. Add a unit test that runs seed.py and teardown_escalation.py from a temp directory holding only the staged `_setup/` layout.

1. **[Axis 1] Host-path gate blocks any new line containing a bare TASK_DIR identifier, even a local Python constant** (`scripts/check-task-host-paths.py:41`) — Match only env-var uses, not the bare identifier. Keep shell/YAML forms `\$\{?(TASK_DIR|SKILLS_REPO_PATH)\b`. For Python, match the name when it is a quoted string literal (`['"](TASK_DIR|SKILLS_REPO_PATH)['"]`). That covers `os.environ.get(...)`, `os.getenv(...)`, `os.environ[...]`, `env[...] = ...` and `env.pop(...)`, which existing tests/tasks/uipath-review checks already use. Local constants like `TASK_DIR = os.path.dirname(...)` (test_handoff.py:23, check-collisions.py:23) and prose comments (validate_flow.py:42) then no longer trip the gate. Alternatively, document in the script docstring and _ADVICE that any bare-word occurrence is flagged and that renaming the local constant (e.g. to HERE or THIS_DIR) is the workaround.
2. **[Axis 4] New fallbacks in the budget guard (missing reference block, _setup/seed.py) are never exercised by any test** (`tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py:510`) — Add two synthetic tests. (a) Put `seed.py` under `tmp_path/_setup/` and assert `_seed_case_count(tmp_path/check_x.py)` returns the case count. (b) Call `_script_path` with a `$REFERENCE_DIR/check_x.py` command and a YAML that has no `reference:` block (or one where `directory:` isn't the first key). Assert that it fails loudly instead of returning None. Or make `_eligible()` count `$REFERENCE_DIR` commands on its own, so an unresolved one shows up as a count mismatch. Today all 73 `reference:` blocks in the bpmn suite list `directory` first, so the corpus is priced correctly (`_eligible()` returns (74, 82)). The gap is regression protection.

## Nits

1. **[Axis 1] $REFERENCE_DIR pricing silently skips a criterion when directory is not the first key under reference:** (`tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py:52`) — `_REFERENCE_BLOCK = re.compile(r"^reference:\s*\n\s*directory:\s*(\S+)", re.M)` only matches when `directory:` is the line directly after `reference:`. For any other key order (e.g. a description or include key first), or for a flow-style `reference: {directory: ../..}`, `_script_path` returns None. The criterion then drops out of both the pricing and the eligibility count without any error. Both migrated YAMLs use the matching layout today, so nothing breaks yet. Fix: parse the YAML with yaml.safe_load and read `doc.get('reference', {}).get('directory')`, or fail loudly when a `$REFERENCE_DIR/` command has no resolvable reference directory.
2. **[Axis 2] No-op branch at the end of the diff-parse loop** (`scripts/check-task-host-paths.py:122`) — `if line.startswith("-") and not line.startswith("---"): continue` is the last statement in the loop body, so `continue` changes nothing. With `--unified=0` there are no context lines to skip either. Delete lines 122-123.
3. **[Axis 2] The rationale for retiring the host paths appears four times** (`scripts/check-task-host-paths.py:2`) — The same #3271 explanation (sandbox drivers have no host checkout; use template_sources or $REFERENCE_DIR; documented exceptions) appears in the module docstring (lines 2-18), in the `_EXEMPT_BASENAMES` comment (lines 43-48), in `_ADVICE` (lines 51-57), and again in the workflow header (task-host-paths-gate.yml lines 3-9). Keep it once in the docstring. Shorten `_ADVICE` to the action plus a pointer to tests/README.md, and cut the workflow header to one line pointing at the script, as token-optimization.md § 0 asks ("one full statement + N pointers").
4. **[Axis 3] A relative root argument crashes diff mode with an unhandled ValueError, exiting 1 like a real finding** (`scripts/check-task-host-paths.py:94`) — `_diff_scan` does `rel_root = root.relative_to(REPO_ROOT).as_posix()`. `main` builds `root = Path(args.root)` without resolving it, so any relative positional argument raises. Ran `python3 scripts/check-task-host-paths.py tests/tasks` and got `ValueError: 'tests/tasks' is not in the subpath of '/Users/religa/src/skills'`, exit 1. The docstring says exit 1 means "one or more offending lines", so a usage error looks like a gate failure. Fix: use `root = Path(args.root).resolve()` in main. Also exit with code 2 (or call `parser.error`) when the root falls outside REPO_ROOT. CI uses the default absolute root, so CI is not affected.
5. **[Axis 6] datafabric e2e mounts the suite-level flow _setup/ but never uses it** (`tests/tasks/uipath-maestro-flow/e2e/datafabric_native_contract_intake_pipeline/datafabric_native_contract_intake_pipeline.yaml:17`) — `path: ../../_setup` / `mount_point: _setup` (lines 17-18) stages cleanup_solutions.py and preflight_connections.py into the agent's sandbox. The task only runs `_setup/preflight_native_entity_nodes.py` (line 33) and `_setup/ensure_entity.py` (line 35), and it has no post_run. Remove the first template_sources entry so the agent-visible _setup/ holds only what this task runs. That keeps the 3-way merge into one mount point from becoming the default copy-paste pattern.
6. **[Axis 6] Exemption matches the basename anywhere under tests/tasks instead of the specific meta-test paths** (`scripts/check-task-host-paths.py:49`) — `_EXEMPT_BASENAMES = {"test_criterion_budgets.py"}` exempts any file with that name in any suite or task directory, in both _full_scan and _diff_scan. Any new file named test_criterion_budgets.py can therefore add $TASK_DIR/$SKILLS_REPO_PATH references without the gate catching them. List the exempt files by their repo-relative paths (tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py, tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py) so a new exemption needs an explicit edit.
7. **[Axis 7] Diff parser misreads added lines whose content starts with "++ " as file headers, and mangles quoted paths** (`scripts/check-task-host-paths.py:104`) — `if line.startswith("+++ "):` (line 104) treats any diff line starting with `+++ ` as a new-file header, and `current_file = path[2:]  # strip "b/"` (line 109) assumes an unquoted `b/` prefix. An added line whose content is `++ x` becomes `+++ x` in the diff, resets `current_file` to a garbage path, and misattributes the rest of that file's hits. A path git quotes (non-ASCII or special characters, e.g. `+++ "b/t\303\251st.yaml"`) strips to `b/t..."`, which gives a wrong annotation path. Also, the `line.startswith("+") and not line.startswith("+++")` branch silently skips any added line whose content starts with `++`. Fix: take the current file from `diff --git a/... b/...` lines, or use `git diff --name-only -z` plus per-file `-U0` diffs, and pass `-c core.quotePath=false`. Track whether the parser is inside a hunk instead of matching on prefixes.
8. **[Axis 8] Gate advice sends readers to tests/README.md for documented exceptions that the README does not contain** (`scripts/check-task-host-paths.py:55`) — `_ADVICE` says "See tests/README.md and the #3271 commit message for the migration pattern and its documented exceptions." Docstring line 10 says the same. tests/README.md does document the template_sources/$REFERENCE_DIR pattern (§ Lifecycle E2E tests > Shape, around lines 327-351). It has no list of exceptions and never mentions #3271 or the new gate (`git grep -n 'check-task-host-paths\|host-path' tests/README.md` returns nothing). It also still shows `SKILLS_REPO_PATH=$(cd .. && pwd)` at lines 64, 299 and 650 (harness env, not task YAML) without saying that task files must not use it. The exceptions exist only in commit a46c4e10b's message, which a CI reader cannot easily find. Fix: add a short tests/README.md subsection with the retired tokens, the three exception classes, the `test_criterion_budgets.py` basename exemption and the gate command, then point `_ADVICE` at that heading. Otherwise, remove the claim that the README documents the exceptions.
9. **[Axis 8] New host-path gate is not listed in docs/REQUIRED-CHECKS.md** (`.github/workflows/task-host-paths-gate.yml:11`) — The workflow says "No `paths:` filter so this can run unconditionally — see validate-skills-sh.yml for why a path filter is unsafe for this class of gate". It also copies the required-workflow concurrency expression. The PR body says the script "fails a PR when it adds new ... references". But the job `No task pins`-style context `No new task TASK_DIR/SKILLS_REPO_PATH references` is not in docs/REQUIRED-CHECKS.md § Current target set (its closest analog, `No task pins sandbox.driver tempdir` / task-driver-gate.yml, is at line 148). As a result, a red run does not block merge. Fix: decide whether this should be required. If yes, add the row in this PR and run `python3 -m pytest tests/scripts/test_required_checks_contract.py -q`. If no, reword the comment and PR description to call it advisory.
10. **[Axis 8] Task description still points to escalation_is.py at the task root after the move to _setup/** (`tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/customer_escalation_triage.yaml:20`) — The description block says "Targets live in the task's `escalation_is.py`." This PR moved that file to `_setup/escalation_is.py`. Change it to "the task's `_setup/escalation_is.py`" so readers find the tenant targets.

## What's Missing

**Rollout impact:**
- 🟡 The PR never says how far the gate reaches. It doesn't only check task YAML, as the commit message says. `_full_scan` and `_diff_scan` cover every file under tests/tasks/. `--full` finds 70 existing hits in about 30 files. Some are deliberate exceptions, such as the uipath-troubleshoot/smoke-manifest-commands/task.yaml comments at lines 29-78. Others are prose or code: uipath-troubleshoot/CLAUDE.md:261, test_same_ground_corpus.py, check_review_cli_provenance.py, and the data/m/r/README.md scenarios. Any later PR that rewrites one of those lines fails the gate, even a comment or whitespace edit. The branch's own untracked tests/tasks/uipath-maestro-flow/outcome/outcome.yaml:94 (`python3 $TASK_DIR/outcome-grader/verify.py ${row.id}`) would also fail once committed. Fix: list the affected suites and owners in the PR (troubleshoot, review, test, flow), or scope the diff scan to task YAML and runner commands. _(trigger: scripts/check-task-host-paths.py)_
- 🔵 tests/Makefile now installs `coder-eval[litellm,harbor,codex]` from unpinned git HEAD. CI installs different extras from the pinned tests/.coder-eval-version: smoke-skills.yml:390 installs `[codex,litellm]`, and smoke-skills.yml:764 and activation-gate.yml:107 install `[litellm]` or nothing extra. The PR doesn't say whether CI needs harbor as well, or whether local `make install` should also follow the version pin. As it stands, local and CI runs can use different coder-eval versions and extras. The change also has nothing to do with the host-path gate, which breaks the one-logical-change-per-PR rule. _(trigger: tests/Makefile)_

**Tests:**
- 🟠 No tests/scripts/test_check_task_host_paths.py exists, and test-helpers.yml has no job that runs one. The siblings test_check_task_driver.py and test_check_skills_sh.py are both run from test-helpers.yml. _(trigger: scripts/check-task-host-paths.py)_ _(restates: Axis 4: New CI gate scripts/check-task-host-paths.py ships with no regression test)_
- 🟡 test_behavior_checker.py:303 runs `_setup/seed.py` in place in the host repo, where escalation_is.py's walk-up finds tests/tasks/uipath-maestro-bpmn/_shared. So the test passes even though the same script fails with ModuleNotFoundError in the staged sandbox. Nothing tests the moved scripts against the template_sources layout: copy only the staged _setup trees into a tmp dir and run seed.py and teardown_escalation.py from there. _(trigger: tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/test_behavior_checker.py)_ _(restates: Axis 7: customer_escalation_triage _setup/escalation_is.py imports _shared, which is not in the sandbox)_

**Parallel paths & mirrors:**
- 🔵 The fragile `_REFERENCE_BLOCK = re.compile(r"^reference:\s*\n\s*directory:\s*(\S+)", re.M)` and the `_setup/seed.py` fallback were copied from tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py (lines 52 and 510). The two guards are now twins. A fix or new test in the bpmn copy has to land in the flow copy too, or the two will drift. _(trigger: tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py)_ _(restates: Axis 1: $REFERENCE_DIR pricing silently skips a criterion when directory is not the first key under reference:)_

**Docs & config:**
- 🔵 The new rule appears in none of the contributor-facing places. .claude/rules/test-writing.md (Sandbox Configuration / Anti-patterns), tests/README.md and .claude/rules/pr-review.md don't mention the gate or the retired $TASK_DIR/$SKILLS_REPO_PATH tokens. Only .claude/commands/generate-task.md:183 warns against $TASK_DIR. tests/README.md:64/299/650, tests/templates/README.md:19 and .claude/commands/skill-compare.md:178 still show `SKILLS_REPO_PATH=$(cd .. && pwd)` for the harness, and nothing says task files must not use it. _(trigger: scripts/check-task-host-paths.py)_ _(restates: Axis 8: Gate advice sends readers to tests/README.md for documented exceptions that the README does not contain)_
- 🔵 docs/REQUIRED-CHECKS.md § Current target set has no row for the new job context `No new task TASK_DIR/SKILLS_REPO_PATH references`. The workflow was built to be a required check (no paths filter, and the required-workflow concurrency expression), but nothing makes it one. _(trigger: .github/workflows/task-host-paths-gate.yml)_ _(restates: Axis 8: New host-path gate is not listed in docs/REQUIRED-CHECKS.md)_

## Guardrails & Automation

- [repo-local-check] Add Gate 3 to scripts/check-task-driver.py, following its existing Gate 1 and Gate 2 docstring sections, with its tests in tests/scripts/test_check_task_driver.py. Gate 3 checks that sandbox hook imports resolve. For every task YAML, it takes each pre_run/post_run command that runs `python3 _setup/<x>.py` or another path relative to the sandbox. It maps that path back through `sandbox.template_sources` (each `path` plus its `mount_point`) to the source .py files and parses them with `ast`. Every `import X` / `from X import` whose top-level module is `_shared`, or any other sibling package not present in the staged mount set, fails the gate. For now, a `sys.path` walk-up counts as unresolved. Report it as `::error file=<yaml>,line=<hook line>`. The existing CI job `task/experiment gate unit tests` and task-driver-gate.yml already run this script, so no new required context is needed. Today other `_setup/*.py` files also import `_shared`: flow datafabric_native_contract_intake_pipeline/_setup/preflight_native_entity_nodes.py and ixp/e2e_03_project_creation_handoff/_setup/handoff*.py. Triage them when the gate lands, either by allowlisting them by repo-relative path or by fixing them. _Prevents:_ Finding 'customer_escalation_triage _setup/escalation_is.py imports _shared, which is not in the sandbox' (high): pre_run `python3 _setup/seed.py` and post_run `python3 _setup/teardown_escalation.py` import escalation_is, which needs `_shared.bpmn_live`, and no template_source stages `_shared`. It also prevents the same class of bug in any future `$SKILLS_REPO_PATH` to `_setup/` migration.
- [repo-local-check] Extend the same Gate 3 in scripts/check-task-driver.py with an unused-mount warning. When a `template_sources` entry has a `mount_point` and no pre_run/post_run command references a file under that source path, emit `::warning` (it does not fail the gate). Only compare against files the task actually runs, so shared mounts that the agent itself is meant to see are not flagged: skip entries without a `mount_point`, such as the skill tree mount. _Prevents:_ Finding 'datafabric e2e mounts the suite-level flow _setup/ but never uses it' (low): `../../_setup` is mounted at `_setup`, but only `_setup/preflight_native_entity_nodes.py` and `_setup/ensure_entity.py` run.
- [repo-local-check] In scripts/check-task-host-paths.py, replace `_TOKEN_RE = re.compile(r"\bTASK_DIR\b|\bSKILLS_REPO_PATH\b")` with a rule that matches only env-var use. Shell/YAML: `\$\{?(TASK_DIR|SKILLS_REPO_PATH)\b`. Python and other code: the name as a quoted string literal, `['\"](TASK_DIR|SKILLS_REPO_PATH)['\"]`, which covers `os.environ.get`, `os.getenv`, `environ[...]`, `env[...] = ` and `env.pop(...)`. Also replace `_EXEMPT_BASENAMES` with `_EXEMPT_PATHS`, a set of repo-relative POSIX paths (tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py and tests/tasks/uipath-maestro-flow/_shared/test_criterion_budgets.py), compared against the rel path in both `_full_scan` and `_diff_scan`. _Prevents:_ Finding 'Host-path gate blocks any new line containing a bare TASK_DIR identifier' (medium): false positives on local constants such as test_handoff.py:23 and check-collisions.py:23, and on prose comments such as validate_flow.py:42. Finding 'Exemption matches the basename anywhere under tests/tasks' (low).
- [repo-local-check] Rework `_diff_scan` in scripts/check-task-host-paths.py so the diff parser is a small, testable function. Run `git -c core.quotePath=false diff --unified=0 --no-renames` or `--find-renames`, as chosen. Take the current file from `diff --git a/<p> b/<p>` headers, or pass `-z --name-only` and diff each file separately. Track whether the parser is inside a hunk, so a `+++`-prefixed line inside a hunk counts as content. Delete the no-op `if line.startswith("-") ...: continue` at lines 122-123. In `main`, call `Path(args.root).resolve()` and use `parser.error(...)` (exit 2) when the root is outside REPO_ROOT, so a usage error is never exit 1. Add these cases to the new unit test below. _Prevents:_ Finding 'Diff parser misreads added lines whose content starts with "++ " as file headers, and mangles quoted paths' (low). Finding 'A relative root argument crashes diff mode with an unhandled ValueError, exiting 1 like a real finding' (low, merged A1/A3/A7/A8). Finding 'No-op branch at the end of the diff-parse loop' (low).
- [repo-local-check] In tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py, and in the flow twin if it has the same helper, replace the `_REFERENCE_BLOCK` regex with `yaml.safe_load(text).get('reference', {}).get('directory')`; the bpmn job already installs pyyaml. Add a corpus assertion in the same meta-test: every criterion command containing `$REFERENCE_DIR/` must resolve to an existing script. Raise AssertionError naming the YAML instead of returning None, so `_eligible()` can no longer drop it silently. _Prevents:_ Finding '$REFERENCE_DIR pricing silently skips a criterion when directory is not the first key under reference:' (low). Also the missing-reference-block half of the finding 'New fallbacks in the budget guard (missing reference block, _setup/seed.py) are never exercised by any test' (medium).
- [repo-local-check] Extend tests/scripts/test_required_checks_contract.py, which the existing `required-check contract guard` job already runs. For every `.github/workflows/*.yml` whose job runs a `scripts/check-*.py` gate on `pull_request` without a `paths:` filter, the job's `name:` must either appear in docs/REQUIRED-CHECKS.md § Current target set or be listed in a new 'Deliberately not required' allowlist in that document. The document already has such a paragraph for the activation and verb detect jobs. _Prevents:_ Finding 'New host-path gate is not listed in docs/REQUIRED-CHECKS.md' (low): task-host-paths-gate.yml has the shape of a required gate, but its context `No new task TASK_DIR/SKILLS_REPO_PATH references` is neither required nor documented as advisory.

- Add tests/scripts/test_check_task_host_paths.py, modeled on tests/scripts/test_check_task_driver.py: run the script with subprocess against a temp git repo. Cover these cases: (a) an added `$TASK_DIR/x.py` line exits 1 with `::error file=...,line=N`; (b) a reference that already existed at the base, and a pure rename that carries one, exit 0; (c) removed lines are ignored; (d) only the exact exempt repo-relative paths are skipped, not a same-named file elsewhere; (e) `SKILLS_REPO_PATH` is caught, including in the `os.environ.get("SKILLS_REPO_PATH")` and `env.pop(...)` forms; (f) a local `TASK_DIR = Path(...)` constant and a prose comment pass; (g) an added content line `++ x` and a quoted non-ASCII path are attributed correctly; (h) a relative root argument exits 2, not 1; (i) `--full` flags references that already existed. Wire it into test-helpers.yml as a new job, or append it to the pytest invocation of the existing `task/experiment gate unit tests` job, so no required context is renamed (CLAUDE.md rule 10). _Why not static:_ This checks the behavior of a hand-written git-diff parser on synthetic diffs. That needs a real git repo and running the script; no lint rule can check that the parser output is correct. _Prevents:_ Finding 'New CI gate scripts/check-task-host-paths.py ships with no regression test, unlike its sibling gates' (high). It also locks in the fixes for the bare-identifier false positive, the basename exemption, the '++' diff-parser bug and the relative-root crash.
- Add a hermetic staging test for sandbox hooks in the bpmn suite, e.g. tests/tasks/uipath-maestro-bpmn/_shared/test_sandbox_hook_staging.py, which pytest-bpmn-check already runs. For each e2e task YAML, copy only its `template_sources` into tmp_path at their mount points. Then run `python3 -c "import runpy, sys; sys.argv=[...]"` in `--dry-run`/import-only mode, or simply `python3 -m py_compile` followed by `python3 -c 'import <module>'` with cwd=tmp_path/_setup. Assert that no ModuleNotFoundError occurs. In the same PR, move `import escalation_is` inside the teardown's try block so that a failed import logs a WARN about possibly leaked Jira and Slack records. _Why not static:_ The Gate 3 AST check covers direct imports. escalation_is.py, however, builds its import path at runtime by walking parent directories for `_shared` and editing sys.path, which static analysis cannot resolve reliably. Only running the import in a staged tempdir shows the real result. _Prevents:_ Finding 'customer_escalation_triage _setup/escalation_is.py imports _shared, which is not in the sandbox, so pre_run seed and post_run teardown crash with ModuleNotFoundError' (high), including the dynamic sys.path walk-up variant.
- Add two synthetic unit tests to tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py. (1) Put seed.py only under `tmp_path/_setup/seed.py` and assert that `_seed_case_count(tmp_path/'check_x.py')` returns the case count. (2) Feed `_script_path` a `$REFERENCE_DIR/check_x.py` command with a YAML that has no `reference:` block, and one with a flow-style block, and assert that it raises instead of returning None. _Why not static:_ These fallback branches depend on the filesystem layout and on the parsed YAML. Today's corpus (all 73 reference blocks list `directory` first, and only one `_setup/seed.py` exists) never reaches them, so only synthetic fixtures run them. _Prevents:_ Finding 'New fallbacks in the budget guard (missing reference block, _setup/seed.py) are never exercised by any test' (medium).
- Documentation and process: add a short tests/README.md subsection, e.g. 'Retired host-path variables'. It should list the retired tokens, the exact exempt paths, the exception classes and the gate command. Point `_ADVICE` and the workflow header at that heading instead of repeating the #3271 rationale (token-optimization.md § 0). Fix the stale `escalation_is.py` path in the customer_escalation_triage description. Make the `workflow_dispatch --full` mode either advisory (`continue-on-error`) or backed by an allowlist of the 70 existing hits, so a manual dispatch doesn't fail red today. _Why not static:_ Whether the rationale is written once, whether a README section exists and where it points, and whether prose names the right file are editorial checks. A link checker would only confirm the heading exists, not that it lists the exceptions. _Prevents:_ Finding 'Gate advice sends readers to tests/README.md for documented exceptions that the README does not contain' (low). Finding 'The rationale for retiring the host paths appears four times' (low). Finding 'Task description still points to escalation_is.py at the task root after the move to _setup/' (low).

## Top 5 Priority Actions

1. Fix the customer_escalation_triage seed and teardown, which now crash with ModuleNotFoundError because `_setup/escalation_is.py:29` imports `_shared.bpmn_live` and `_shared/` is never staged into the sandbox. Stage `../../_shared` in template_sources (tests/tasks/uipath-maestro-bpmn/e2e/customer_escalation_triage/customer_escalation_triage.yaml:31-54) or vendor the helper into `_setup/`, move `import escalation_is` inside the teardown's try block, and add a test that runs both scripts from a temp directory holding only the staged files; otherwise post_run fails silently and tenant records leak.
2. Add tests/scripts/test_check_task_host_paths.py, a temp-git-repo test of `_diff_scan` in scripts/check-task-host-paths.py:89. It should cover added lines that must fail, pre-existing and renamed lines, removed lines, the exempt file, SKILLS_REPO_PATH and `--full`. Also cover the `+++ `/quoted-path parser edge cases at line 104, and wire the test into test-helpers.yml next to test_check_task_driver.py.
3. Narrow `_TOKEN_RE` at scripts/check-task-host-paths.py:41 so it matches only env-var uses: `\$\{?NAME\b` plus the quoted-string form `['"]NAME['"]`. The quoted form still catches `os.environ.get`, `env[...]` and `env.pop`, while local constants such as `TASK_DIR = ...` (test_handoff.py:23, check-collisions.py:23) and comments stop failing PRs that read no host-path variable.
4. Decide whether the gate is required. If it is, add its job context to docs/REQUIRED-CHECKS.md § Current target set and run the test_required_checks_contract.py test (.github/workflows/task-host-paths-gate.yml:11). If it isn't, reword the workflow comment and PR text to say it is advisory, because right now a red run does not block merge.
5. Make the budget guard fail loudly: `_script_path` (tests/tasks/uipath-maestro-bpmn/_shared/test_criterion_budgets.py:52) should read `reference.directory` with yaml.safe_load, or raise when a `$REFERENCE_DIR/` command can't be resolved, and synthetic tests should cover the missing-reference and `_setup/seed.py` fallbacks (line 510), so criteria never silently drop out of both pricing and the eligibility count.

---

**Change class:** complex — adds a new diff-parsing CI gate script + workflow, and migrates two e2e tasks' grading commands/template_sources and a budget-pricing guard's script-path resolution


**Stats:** 0 🔴 · 2 🟠 · 2 🟡 · 10 🔵 across 8 axes reviewed.
**Verification:** 15 medium+ finding(s) adversarially re-checked · 3 dropped as false positives (20%) · 11 corrected in place · 15 low passed through unverified.

uipreliga (approved):
Fix what you agree with and 🚢

PR #3358, opened 2026-09-16
feat(uipath-insights): teach dashboard update, delete, and copy [IN-13531]

celestinryf:
[IN-13531](https://uipath.atlassian.net/browse/IN-13531). Pairs with [UiPath/cli#4263](https://github.com/UiPath/cli/pull/4263) (update), [UiPath/cli#4241](https://github.com/UiPath/cli/pull/4241) (delete), and [UiPath/cli#4262](https://github.com/UiPath/cli/pull/4262) (copy). Stacked on [#3182](https://github.com/UiPath/skills/pull/3182), which teaches the reads and create.

**Do not land before all three cli PRs reach `main` and republish `@uipath/cli@dev`.** Three `run_command` criteria call `--help` on each verb, so a premature run fails loudly rather than ambiguously. The verb gate does not enforce this: it fails only on a finding with a null `matched_prefix`, and everything here matches `insights`. Draft is the guard.

## What

`references/dashboard-writes-guide.md` goes from 160 lines to 268 and now owns all four writes: nine more numbered rules, the three everyday edit recipes with before-and-after fragments, and a Commands section per verb with the real `Data` keys.

Its Errors section is rewritten rather than extended. One create-shaped table could not answer the only question that matters after a write fails, so it is now two: what each command refuses before sending anything, and what a failure after the request says about whether the write landed. Every row comes from `scenarios/insights-tool/dashboards.md` S15 to S33 and the three executors under `packages/insights-tool/src/utils/`.

In `SKILL.md`, the Scope Boundaries row "Update, delete, or copy a dashboard, or generate one from a prompt | Insights UI" becomes prompt generation alone. Leaving it was the worse of the two failures here, because it routes an agent to the UI for something the CLI does. Rule 17 picks up `--backup-file` and copy's `--output-file`, Rule 18 gains the version and delete clauses, and front matter lands at 1,507 combined against the 1,536 cap.

Also activation rows 054 to 058, and `tests/tasks/uipath-insights/dashboards/update_smoke.yaml`.

## Two judgement calls

`LinkedAlerts` keeps its full array beside the `LinkedAlertsDeleted` count, which the LLD left open. The guide tells the agent to name the deleted alerts back to the user, and after the delete nothing can recover them.

The update recipes take the file from a by-id read and the version from a process-key read. A process-key read drops the dashboard's stored global filters, so a file taken from it would delete them on the next update. The version has to come from the slot, because that is the column the backend increments.

## Verification

`skills:validate`, `skills:check-links` (6,843 links), `check-skill-status.py`, `check-task-driver.py`, and `check-task-host-paths.py` are green.

Every `command_pattern` was replayed through the pinned coder-eval 0.12.1 normalizer against nine hand-built cases, including a newline-batched read-then-update, and each matched exactly what it should. The barrier is `(?:(?!uip\s)[^&;|\n])*` for the reason [#2839](https://github.com/UiPath/skills/pull/2839)'s commit message sets out.

Not run: the smoke and activation suites need Docker and Bedrock credentials this machine lacks, so CI is the first real run of either. Two criteria on update's own flags are advisory at `pass_threshold: 0`, because reaching an update depends on the slot state and the task does not control it.

[IN-13531]: https://uipath.atlassian.net/browse/IN-13531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rliuup (approved):
LGTM

rohitjain-uipath (approved):

PR #3359, opened 2026-09-16
test(bpmn): e2e — compose an agent-authored API workflow into a Maestro process

DevMomo:
## Summary

Phase 0/1 of a new coder-eval e2e task under `tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/`: the agent authors an API workflow it did not have (via `uipath-api-workflow`), ships it in the same solution as a Maestro BPMN process that invokes it at run time via `Orchestrator.ExecuteApiWorkflowAsync`, and deploys. The grader (added in a later commit) mints a token after the agent stops, runs the process live, and asserts the token comes back transformed out of the invoking node's own outputs — per-node provenance, so nothing hard-coded can pass.

Scoped to v1 (API workflow only) per the build plan; an RPA-workflow row follows in a later phase/PR.

This PR currently contains only Phase 0 (`seed.py` — per-run folder/solution prefix convention) plus, as work lands, the Phase 1 gold solution under `fixtures/gold/`. Checkers, teardown, and the task YAML land in follow-up commits on this same branch (Phase 2/3), not in this initial draft.

Depends on (should land first, restacked onto once merged):
- skill fix PR(s) for `uipath-maestro-bpmn` (job-wrapper v2 contract, delegated-resource incorporation sequence, `refresh` vs `update-metadata`, warning-gating)
- skill fix PR for `uipath-api-workflow` (object-shaped `Response`, not a scalar)

## Test plan
- [ ] Gold solution under `fixtures/gold/` builds from the amended skill text and passes a hand-verified `uip maestro bpmn debug` round-trip
- [ ] `pytest` over the checkers is green against gold + mutation fixtures
- [ ] 3-of-3 CI replicates green via `run-coder-eval.yml` on the default `cli_version`
- [ ] 5 additional stabilization replicates recorded before merge


tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+
+    problems: list[str] = collect_shape_problems(
+        root,
+        resolve_release_key_for_row=lambda row, project: live_release_key_resolver(
+            folder_path, project.name if project is not None else row["kind"]
+        ),
+        resolve_folder_key=live_folder_key_resolver(folder_path),
+    )
+
+    validate = run_cli(["uip", "maestro", "bpmn", "validate", str(project_dir)], timeout=VALIDATE_TIMEOUT)
```
nikhil-maryala: [BLOCKER] **`validate` is given a directory, so the shape check fails for every submission, gold included.** `project_dir = bpmn_path.parent` (l.551), then `run_cli(["uip","maestro","bpmn","validate", str(project_dir)])`.
- UiPath/cli main: `commands/validate.ts:39` declares `.argument("<file>", "Path to the .bpmn or .xml file to validate")`, and `BpmnValidateService.validateFile` throws `Expected a .bpmn or .xml file but got a directory` (`bpmn-validate-service.ts:633-636`).
- Coder-eval run 35284799483 printed `validate Status is None, expected 'Valid'`.
- The skill (`SKILL.md` step 3) and the sibling checkers (`_shared/validate_bpmn.py`, `check_solution_select.py`) all pass the `.bpmn` path.

Fix: `str(bpmn_path)`. Add a unit test that asserts the argv `main()` passes to `run_cli`, since none of the 44 tests reach this line.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+            for value in candidates
+        )
+        if not by_reference:
+            problems.append(
+                f"{row['kind']}: JobArguments must reference a process "
+                f"variable (=vars.<id>, or =js: containing vars.<declared-id>), "
+                f"got {job_args!r}"
+            )
+
+    node_output_var_ids = {
```
nikhil-maryala: [MAJOR] **This requires a node-scoped variable literally named `message`, and an end-event source of exactly `=vars.<id>`. Neither the prompt nor the skill asks for that.**
- The prompt only requires the process output to be named `message` (yaml l.52-54).
- #3360's `registry-workflow.md` teaches `source="=result"` "or the exact schema fields", and never names the node variable.

One-change copies of gold, run with `collect_shape_problems` and the `test_checks.py` resolvers:
- node output variable renamed `greeting` → FAIL `no variable named 'message' is declared scoped to 'Task_InvokeGreeting'`
- the skill-taught `=result` object in a json variable, with the end event mapping `=js:vars.X.message` → same FAIL
- end event mapping `=js:'Hello, ' + vars.Var_Message` → FAIL `never mapped out by an end event`

Run 35284799483 hit exactly this: `check_behavior` passed and `check_shape` failed on the naming rule. `valid_unlike_gold` only renames ids, so it never tests naming freedom.

Fix: take the node's output variables from the wrapper task's `uipath:output var=`, whatever their names. Accept any end mapping whose source references one of them (`var_refs_in`), and require only that the mapping's target is the declared process output named `message`.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
```diff
@@ -0,0 +1,256 @@
...
+        raise CheckFailure(
+            f"{row['kind']}: declared output {row['output']!r} never appears "
+            "in the runtime Globals"
+        )
+    if token not in str(published_value):
+        raise CheckFailure(
+            f"{row['kind']}: output {row['output']!r} ({published_value!r}) "
+            f"does not contain the minted token {token!r}"
+        )
+    if str(published_value) == token:
```
nikhil-maryala: [MAJOR] **The weight-5.0 behaviour check has requirements the prompt never states.** The published value must contain the minted token (l.147) and must also not equal it (l.152). Line 165 applies the same token rule to the node's own Outputs. The prompt (yaml l.45-54) only says the API workflow takes a string and returns `{message: string}` "populated from the API workflow's own response". It never says the response must embed or transform the input, and "greeting" appears nowhere in it. So an echo workflow fails "nothing was computed" (`test_bare_token_global_fails`), and a workflow that ignores its input fails the token check, even though both satisfy the prompt as written. Fix: state the contract in the prompt ("`message` must include the input value and must not be identical to it"), or drop the bare-token rule.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+# ---------------------------------------------------------------------------
+
+def assert_validate_clean(validate_payload: dict) -> list[str]:
+    data = get_ci(validate_payload, "Data", {}) or {}
+    problems: list[str] = []
+    status = get_ci(data, "Status")
+    if status != "Valid":
+        problems.append(f"validate Status is {status!r}, expected 'Valid'")
+    warnings = get_ci(data, "Warnings", []) or []
+    bad = [
```
nikhil-maryala: [MAJOR] **This gate can never fire.** The CLI emits `Data.Warnings` as a single string, `` `${n} warning(s):\n  - [id] CODE: msg` `` (`bpmn-validate-service.ts:615-619`; the skill's `cli-conventions.md` also documents "`Warnings` (one string)"). The comprehension keeps only `dict` elements. Iterating the string yields characters, so `bad` is always empty. Probe: `assert_validate_clean({'Data':{'Status':'Valid','Warnings':'1 warning(s):\n  - [Task_1] VARIABLE_DOES_NOT_EXIST: …'}})` → `[]`. `test_checks.py:295-298` feeds a list of dicts, which the CLI never produces. `check_no_undeclared_var_reads` covers dotted `vars.X` references only. Fix: when `warnings` is a `str`, test `"VARIABLE_DOES_NOT_EXIST" in warnings`, and use the real envelope shape in the test.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
```diff
@@ -0,0 +1,392 @@
...
+                "mode": "binding",
+                "binding_resource": "process",
+                "binding_attr": "Key",
+            },
+            "folderKey": {"mode": "literal_guid"},
+        },
+        # Field names the broken registry template ships that must NOT
+        # survive into the authored node -- their presence signals the
+        # template was pasted as served rather than fixed per rule 18.
+        "forbidden_context_fields": ("folderId", "folderPath", "name"),
```
nikhil-maryala: [MINOR] **`folderPath` and `name` are treated as fatal, but the skill only calls them unneeded.** #3360's `registry-workflow.md` says the template's `folderId`/`folderPath`/`name` fields "are not needed … and can be dropped", not that they must be removed. A gold copy with an extra `folderPath` field fails shape. Either make the skill say "remove them", or forbid only the misnamed `folderId`, which is the real signal that the template was pasted as served.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+        if not any(f"=vars.{vid}" in end_sources for vid in node_output_var_ids):
+            problems.append(
+                f"{row['kind']}: node {node_id!r}'s own output variable(s) "
+                f"{sorted(node_output_var_ids)} are never mapped out by an "
+                "end event"
+            )
+        else:
+            published_names = {
+                declared[vid]["name"]
+                for vid in matched_targets
```
nikhil-maryala: [MINOR] **"Populated from the API workflow's own response" isn't enforced when an extra mapping exists.** A second end mapping into the same output variable, sourced from the process input, still passes shape: gold plus `=js:'Hello, '+vars.Var_Name` → `output_Var_Message` passes. The behaviour check then can't tell the two apart if the API workflow echoes its input. Require every write to the `message` output variable to come from the node's own output variables, the same way `foreign_writes` already guards the node variable.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
```diff
@@ -0,0 +1,170 @@
...
+    try:
+        seed = json.loads(seed_path.read_text(encoding="utf-8"))
+    except json.JSONDecodeError as error:
+        logger.warning("seed.json is not valid JSON: %s", error)
+        delete_studio_web_solutions()
+        return 0
+
+    folder_path = f"{seed.get('parentFolderPath', '')}/{seed.get('folderName', '')}".strip("/")
+    solution_name = seed.get("solutionName", "")
+
```
nikhil-maryala: [MINOR] **Teardown misses objects in three cases.**
- **Renamed folders.** It deletes only the exact seeded folder. `solution deploy run` creates a new folder and renames on collision (`uipath-solution/references/pack-and-deploy.md`), so a second deploy by the agent (rule 17's deploy-then-bind, or a retry after `Draft`) leaks `<folderName> 1`, and that folder's key also fails the grader's exact-path resolvers.
- **Deployments.** `deploy list --limit 50` (l.64) filters by `--folder-path` after the fetch, so on a busy tenant this run's deployment can be missed.
- **Packages.** `packages list --limit 200` (l.90) isn't paginated.

Fix: also remove folders matching `<folderName>*` under the seeded parent, and paginate both lists.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
```diff
@@ -0,0 +1,107 @@
...
+  4. The process exposes exactly one declared output named `message`
+     (string), populated from the API workflow's own response, that a
+     caller receives when the process completes.
+
+  Both projects must live inside the one named solution. Deploy the
+  solution (and the API workflow resource it depends on) into the seeded
+  Orchestrator folder so the process can actually reach and invoke the
+  resource when run. Validate before you finish.
+
+  Do NOT run or debug the process yourself -- the grader runs it live with
```
nikhil-maryala: [MINOR] **This contradicts #3360's rule 17 step (5), and the prompt wins.** Rule 17 tells the agent to run `uip maestro bpmn debug` and read the invoking node's real output key, because "a clean `validate`/`pack` proves nothing". This line forbids exactly that, so the agent has to guess `=result.<key>`, and the eval can't exercise the step #3360 adds. Either allow one debug run with the agent's own input, or add a carve-out to the skill.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
```diff
@@ -0,0 +1,107 @@
...
+  to invoke that resource at run time. The grader mints a fresh token,
+  drives the deployed process live via `uip maestro bpmn debug`, and
+  verifies per-node provenance in `debug-instance variables-all` -- the
+  token must round-trip through the resource's OWN invoking node, not a
+  ScriptTask standing in for it. See composition.py/check_shape.py/
+  check_behavior.py in this directory for the full anti-cheat rationale.
+tags: [uipath-maestro-bpmn, e2e, "mode:build", "lifecycle:setup", "shape:multi-node", resource, outcome-graded]
+
+run_limits:
+  max_turns: 90
```
nikhil-maryala: [MINOR] **The only live run ran out of turns, and there's no run on the current head.** Run 35284799483 ended `MAX_TURNS_EXHAUSTED` at 90 turns (525s) even though the solution worked. The only coder-eval runs are from 09-17: one scored 0.70, the other hit a setup ERROR after 3s. `.claude/rules/test-writing.md` step 5 requires a passing-run claim. Re-run on the current head once the shape fix lands, and raise `max_turns` if the multi-skill flow (api-workflow authoring, solution, deploy, bind) needs it.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/fixtures/gold/GreetingPipelineGoldSolution/AGENTS.md
```diff
@@ -0,0 +1,462 @@
+# UiPath Solution Workspace
```
nikhil-maryala: [MINOR] **This `AGENTS.md` and the sibling `CLAUDE.md` are scaffold files `uip solution init` generated.** They're byte-identical (462 lines each), and no grader reads them. A nested `CLAUDE.md` in the repo is auto-loaded by Claude Code for anyone who works in this directory. Delete both.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/fixtures/gold/GreetingPipelineGoldSolution/GreetingProcess/GreetingProcess.bpmn
```diff
@@ -0,0 +1,78 @@
...
+    <bpmn:extensionElements>
+      <uipath:variables version="v1">
+        <uipath:input id="input_Var_Name" name="name" type="string" elementId="Event_start" />
+        <uipath:inputOutput id="Var_Name" name="name" type="string" elementId="Process_1" />
+        <uipath:inputOutput id="Var_ApiResponse" name="apiResponse" type="json" elementId="Task_InvokeGreeting" />
+        <uipath:inputOutput id="Var_Message" name="message" type="string" elementId="Task_InvokeGreeting" />
+        <uipath:output id="output_Var_Message" name="message" type="string" elementId="End_1" />
+      </uipath:variables>
+      <uipath:bindings version="v1">
+        <uipath:binding id="Binding_GreetingApiRelease" resource="process" propertyAttribute="Key" default="C6ABA92A-AEB1-47FD-ACFF-39D5FA1CE45F" />
```
nikhil-maryala: [NIT] **Real tenant identifiers are committed:** the process Key (l.13) and FolderKey (l.32), repeated in `test_checks.py:41-42`. They aren't secrets, and the folder is per-run, but placeholder GUIDs would keep tenant data out of the repo.

nikhil-maryala (changes requested):
**Request changes.** Tally: 1 Blocker / 0 Critical / 3 Major / 6 Minor / 1 Nit.

The fixture design is strong:
- All 16 fixtures behave as intended. Gold and `valid_unlike_gold` pass, and every mutation fails for its intended reason only.
- `test_checks.py` passes (44 tests).
- Task-schema validation passes on the pinned 0.12.4, `check-task-host-paths` is OK, and the criterion budgets sum to ≤ 900.
- The graders fail closed and never read agent monologue.
- Nothing from the gold fixtures reaches the agent sandbox: `template_sources` mounts only `_setup/`, and graders go through `$REFERENCE_DIR`.

Blocking:
1. **The weight-3.0 shape criterion can never pass.** `check_shape.py` passes the project directory to `uip maestro bpmn validate`, which takes one `<file>` argument (`commands/validate.ts:39`) and throws `Expected a .bpmn or .xml file but got a directory` (`bpmn-validate-service.ts:633-636`). Coder-eval run 35284799483 shows exactly this: `validate Status is None`. The unit tests never call `main()`, so CI stays green.
2. **The graders fail correct solutions on requirements the prompt and skill never state.**
   - a node-scoped variable literally named `message`
   - an end-event source of exactly `=vars.<id>`
   - an output that must embed the token and also differ from it
   - leftover `folderPath`/`name` fields treated as fatal

   In run 35284799483 the token round-tripped (`check_behavior` passed), and `check_shape` failed on the `'message'` naming rule anyway.
3. **The `VARIABLE_DOES_NOT_EXIST` gate is dead code.** The CLI emits `Data.Warnings` as one string, and the gate only matches list-of-dict elements.

Also:
- **No passing run yet.** No coder-eval run exists on the current head, and the only live run hit `MAX_TURNS_EXHAUSTED` at 90 turns. `test-writing.md` step 5 requires a passing-run claim.
- **Merge order.** This PR is stacked on #3360, which is currently conflicting with main and has changes requested. On main's skill, `ExecuteApiWorkflowAsync` is still taught as fire-and-forget, so land #3360 first and retarget. `composition.py:12` and `check_shape.py:15` cite "rule 18", so re-check that reference after the rebase.
- **Stale PR body.** It lists a dependency on #3361, which was closed unmerged, and says the checkers and teardown "land in follow-up commits", but they're already in this PR.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: 🔴 **critical: `validate` is given the project directory, so shape fails every run, gold included.**

- `uip maestro bpmn validate` takes `<file>`. `validateFile` throws `Expected a .bpmn or .xml file but got a directory`.
- Run 35284799483 shows exactly this: `validate Status is None, expected 'Valid'`.
- `test_checks.py` never exercises `main()`, so CI stays green.

**Fix:**
```python
validate = run_cli(["uip", "maestro", "bpmn", "validate", str(bpmn_path)], timeout=VALIDATE_TIMEOUT)
```

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+    data = get_ci(validate_payload, "Data", {}) or {}
+    problems: list[str] = []
+    status = get_ci(data, "Status")
+    if status != "Valid":
+        problems.append(f"validate Status is {status!r}, expected 'Valid'")
+    warnings = get_ci(data, "Warnings", []) or []
+    bad = [
+        w
+        for w in warnings
+        if isinstance(w, dict) and (w.get("Code") or w.get("code")) == "VARIABLE_DOES_NOT_EXIST"
```
rockymadden: 🟠 **major: the `VARIABLE_DOES_NOT_EXIST` gate never fires.**

- `Data.Warnings` is one string, for example `"2 warning(s):\n  - [Task_1] VARIABLE_DOES_NOT_EXIST: ..."` (`bpmn-validate-service.ts` around line 616).
- `get_ci(..., [])` returns that string. Iterating it yields characters, and `isinstance(w, dict)` is never true.
- `test_checks.py:292` and `:296` feed a list of dicts, a shape the CLI never emits, so the tests pass against a fiction.

**Fix:** match on the string. Update the tests to use the real envelope.
```python
warnings = get_ci(data, "Warnings", "") or ""
if "VARIABLE_DOES_NOT_EXIST" in str(warnings):
    problems.append(f"validate reported VARIABLE_DOES_NOT_EXIST: {warnings}")
```

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+            )
+
+        end_mappings = composition.end_event_mappings(process)
+        end_sources = {m.get("source", "") for m in end_mappings}
+        matched_targets = {
+            m.get("var")
+            for m in end_mappings
+            if m.get("source", "").removeprefix("=vars.") in node_output_var_ids
+        }
+        if not any(f"=vars.{vid}" in end_sources for vid in node_output_var_ids):
```
rockymadden: 🟠 **major: shape and behaviour contradict each other.**

- Shape requires the end event's `source` to be exactly `=vars.<node-var-id>`.
- Behaviour (`check_behavior.py:152`) fails when the published value equals the bare token.
- Result: an echo API workflow plus an end-event `=js:'Hello, ' + vars.Var_Message` passes behaviour but fails shape here. `check_behavior.py`'s own docstring calls that the intended shape.

**Fix:** accept any end-event `source` that references the node's own output variable, not only the exact `=vars.` form:
```python
fed = [m for m in end_mappings if composition.var_refs_in(m.get("source")) & node_output_var_ids]
```

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,575 @@
...
+            problems.append(
+                f"{row['kind']}: JobArguments must reference a process "
+                f"variable (=vars.<id>, or =js: containing vars.<declared-id>), "
+                f"got {job_args!r}"
+            )
+
+    node_output_var_ids = {
+        vid
+        for vid, meta in declared.items()
+        if meta.get("name") == row["output"] and meta.get("elementId") == node_id
```
rockymadden: 🟡 **minor: the node-scoped variable must literally be named `message`.**

- Neither the prompt nor the skill requires that name. The prompt names only the *process* output `message`.
- A node variable named `greeting`, or reading `=vars.Var_ApiResponse.message` at the end event, fails shape even though the value is correct.
- The one live run passed this rule only because the agent picked `message`.

**Fix:** find the node's own output variables by `elementId == node_id` alone. Then require that one of them feeds the published `message` output (see the comment on line 337).

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: 🟡 **minor: the invoking node can publish the input instead of the response, and both checks pass.**

- Suppose the node's own output mapping is `source="=js:'Hello, ' + vars.Var_Name"`.
- Shape passes: the only writer is the node itself.
- Behaviour passes: the echoed response still carries the token in the node's `Outputs`.
- So the published `message` never touches the API workflow's response. This is reasoned, not run.

**Fix:** require the node's output mapping `source` to reference `result`:
```python
if not any("result" in (w.get("source") or "") for w in composition.output_mappings_of(task) if w.get("var") in node_output_var_ids):
    problems.append(...)
```

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
rockymadden: 🟠 **major: the task secretly requires a transformation the prompt never asks for.**

- The prompt says only "populated from the API workflow's own response".
- An agent that follows it literally (echo workflow, direct mapping) publishes the bare token and loses the weight-5.0 criterion.

**Fix, one of:**
- **Preferred:** state the contract in the prompt, e.g. "returns `message` = `Hello, <input>!`", and assert `published_value == f"Hello, {token}!"`. That is stronger anti-cheat than `!=`.
- Drop this check and rely on node provenance.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
rockymadden: 🟡 **minor: the prompt forbids the run that the skill requires.**

- SKILL.md rule 17 step (5) on the base branch says verify with `uip maestro bpmn debug` and read the node's own output. `=result.<key>` must match the real response key.
- This line tells the agent not to. Either the agent disobeys the skill, or it guesses the result key blind.

**Fix:** allow a debug run with the agent's own input. The grader mints its own token later, so nothing leaks:
> You may debug the process with your own test input; the grader re-runs it with a fresh value after you stop.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
rockymadden: 🟡 **minor: leftover `folderPath` or `name` is treated as fatal, but rule 18 only says they are "not needed".**

- Take a node with the correct `releaseKey` plus `folderKey` and a leftover `folderPath`. It runs, per rule 18's wording, yet fails shape.

**Fix:** forbid only `folderId`, the misnamed field that signals a verbatim paste. Or cite a run showing extra fields fault.
```python
"forbidden_context_fields": ("folderId",),
```

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
```diff
@@ -0,0 +1,170 @@
...
+
+def delete_folder(folder_path: str) -> None:
+    result = _run(["uip", "or", "folders", "delete", folder_path, "--yes"], FOLDER_DELETE_TIMEOUT)
+    status = result.get("Result") if isinstance(result, dict) else None
+    logger.info("delete folder %s -> %s", folder_path, status or "unknown/already gone")
+
+
+def delete_package_versions(solution_name: str) -> None:
+    payload = _run(["uip", "solution", "packages", "list", "--limit", "200"], PACKAGES_LIST_TIMEOUT)
```
rockymadden: 🟡 **minor: teardown can leak objects.**

- `packages list --limit 200` scans the whole tenant feed, and `deploy list --limit 50` (line 64) has no paging either. Once the feed passes 200, this run's package can be missed.
- Seed creates only the parent folder. If the agent deploys into any other folder name, nothing reclaims it.

**Fix:**
- Filter the package list by `solutionName`, or page until empty.
- Also sweep `parentFolderPath` for any child folder containing `runId`.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/fixtures/gold/GreetingPipelineGoldSolution/AGENTS.md
rockymadden: 🟡 **minor: drop `AGENTS.md` and `CLAUDE.md` from gold.**

- They are 924 lines of byte-identical CLI scaffold (`uip` 1.203.0 snapshot).
- No checker reads them, and they go stale with every CLI bump.

**Fix:** delete both files. Add them to a fixture `.gitignore` if the scaffold keeps regenerating them.

rockymadden (changes requested):
## 🔴 Changes requested

The shape check can never pass. `validate` gets a directory and throws, so the weight-3.0 criterion fails on every run, gold included. On top of that, the two checkers and the prompt disagree about what a correct answer looks like.

### 📦 What

- Adds a new e2e task, `api_workflow_composition`. The agent writes an API workflow, calls it from a BPMN process via `Orchestrator.ExecuteApiWorkflowAsync`, and deploys both in one solution.
- Grading:
  - **shape** (3.0): offline structure checks, a live key match, and `validate`.
  - **behaviour** (5.0): a token minted after the agent stops must round-trip live through the invoking node's own `Outputs`.
- Also adds a gold fixture, 16 mutation fixtures, `test_checks.py`, and seed/teardown scripts.
- Stacked on #3360. I reviewed only this PR's diff.

### 🔍 Overall findings

- 🟠 **major: no passing live run, and the only run is on an older commit.** The single coder-eval run (35284799483, `ca8d8cc`, checkers identical to head) hit `max_turns` 90 at 525s and scored 0.70. Every test-plan box is unchecked.
  - **Fix:** after the fixes below, get one full green run on head.
  - **Fix:** decide whether 90 turns is enough for author + pack + publish + deploy + bind. That run ran out mid-task.

### ✅ Status of the earlier review (@nikhil-maryala, same head `3ad77b6`)

Nothing has been pushed since that review, so every finding in it still stands.

- ✅ Confirmed against `uip` source: the validate-directory bug and the string-typed `Data.Warnings`.
- ⚠️ One correction. The live run's shape failure lists **only** `validate Status is None`. The node-variable-named-`message` rule did **not** fire, because that agent happened to use the name. The rule is still overfit (see the inline comments), but the run is not evidence of it.

### 🧪 Confirmed vs reasoned

- **Confirmed:**
  - `validate` rejects directories (`cli` `bpmn-validate-service.ts` `validateFile`).
  - `Data.Warnings` is a single string (same file, around line 616).
  - The live-run log (35284799483).
  - The rule 17/18 text on the base branch.
- **Reasoned, not run:**
  - The output-overwritten-from-input bypass.
  - Teardown leaks.
- I read the tests but did not run them.

### tl;dr

Shape can never pass: `validate` gets a directory. The warning gate never fires. The prompt and checkers disagree on what a correct answer looks like. There is no green run on head. Fix those, then run it live.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716. `validate` gets the `.bpmn` file.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716. No name requirement; any end mapping that references the node's response variable passes (`=vars.<id>` or `=js:`).

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
rockymadden: ✅ Fixed in f6e6716. The "must differ from token" rule is dropped. Shape now requires the value to come from the node's response.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716. Reads the string, matches `VARIABLE_DOES_NOT_EXIST`; tests use the real envelope plus a failure case.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
rockymadden: ✅ Fixed in f6e6716. Only `folderId` is fatal; `valid_js_end_mapping` covers a leftover `folderPath`.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in 46e05d6. Every end mapping into the published output must read the node's response. New `stray_output_write` fixture is your exact example and is rejected.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
rockymadden: ✅ Fixed in f6e6716. Listings page through `Pagination.HasMore`, packages filter by `--name`, and every folder under the parent whose name contains `runId` is removed, deepest first.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
rockymadden: ✅ Fixed on the #3360 side (73a28e5): the skill's debug run is skipped when the task forbids live runs.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
rockymadden: ⏳ Still open. Needs a live coder-eval run on the current head.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/fixtures/gold/GreetingPipelineGoldSolution/AGENTS.md
rockymadden: ✅ Fixed in f6e6716. Both deleted.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/fixtures/gold/GreetingPipelineGoldSolution/GreetingProcess/GreetingProcess.bpmn
rockymadden: ✅ Fixed in 46e05d6. Placeholder GUIDs in gold, mutations and `test_checks.py`.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
nikhil-maryala: [MINOR] **Still no passing run on a current head.** The latest coder-eval runs on this branch are from 09-17: `35284799483` failed (`MAX_TURNS_EXHAUSTED` at 90 turns after 525s, with a working solution), and `35275039852` failed too. `.claude/rules/test-writing.md` step 5 requires a passing-run claim in the PR. Now that the shape Blocker is fixed, dispatch `run-coder-eval.yml` on this head. If it runs out of turns again, raise `max_turns`, since this task chains api-workflow authoring, solution setup, deploy and binding. Please do this before taking the PR out of draft.

nikhil-maryala (approved):
**Re-review of `46e05d65a` — Approve.** Tally: 0 Blocker / 0 Critical / 0 Major / 1 Minor / 0 Nit. This replaces my Request changes.

Every earlier finding except the passing-run one is addressed. I verified each on this head: `test_checks.py` 47 passed, plus one-change probes of gold through `collect_shape_problems` with the test resolvers.
- **[BLOCKER] validate given a directory: fixed.** `run_cli([... "validate", str(bpmn_path)])`.
- **[MAJOR] literal `message` node-variable name: fixed.** `response_output_vars` takes whatever the node fills from its own response (`source` absent or reading `result`). Renaming the node variable to `Var_Greeting`/`greeting` throughout → `[]`. `=js:'Hello, ' + vars.Var_Message` on the end event → `[]`.
- **[MAJOR] unstated "must differ from the token" rule: dropped.** The docstring now says an echo API workflow mapped straight through is valid.
- **[MAJOR] dead `VARIABLE_DOES_NOT_EXIST` gate: fixed.** `Warnings` is read as the CLI's string, a failure reports `Instructions`, and the test uses the real envelope.
- **[MINOR] `folderPath`/`name` fatal: fixed.** Only `folderId` is forbidden. A gold copy plus `folderPath` → `[]`, and plus `folderId` → fails.
- **[MINOR] output overwritten from the input: fixed** by the new stray-write check and the `stray_output_write` fixture. A gold copy plus a second end mapping from `vars.Var_Name` → `'message' is also published from [...]`.
- **[MINOR] teardown: fixed.**
  - It sweeps every `runId` folder under the parent, deepest first, and pages both lists.
  - `packages list --name` is server-side.
  - The flags exist in UiPath/cli main: `deploy list --offset`, `packages list --offset/--name`, `or folders list --path/--offset`.
- **[MINOR] prompt forbids debug vs rule 17: resolved on the #3360 side.** Rule 17 step (5) is now "Unless the task forbids live runs…, otherwise use the schema name and report the mapping as unverified".
- **[MINOR] scaffold `AGENTS.md`/`CLAUDE.md`: removed.**
- **[NIT] tenant GUIDs: replaced** with placeholders; `git grep` finds no old keys.

The one remaining item is inline: there's still no coder-eval run on a current head. #3360 is now approved and mergeable, so once it lands, retarget this PR to main.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in f6e6716 (`node_publishes_input` rejected) and 46e05d6 (stray end mappings rejected).

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
rockymadden: ✅ Fixed on the #3360 side (73a28e5).

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
rockymadden: ✅ Fixed in f6e6716.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/fixtures/gold/GreetingPipelineGoldSolution/AGENTS.md
rockymadden: ✅ Fixed in f6e6716.

rockymadden:
🔧 **Review findings addressed** (f6e6716, 46e05d6; #3360 merged in via cda76e8)

- ✅ `validate` gets the `.bpmn` file.
- ✅ The `Data.Warnings` gate reads the real string shape, and the tests use the real envelope.
- ✅ Any end mapping that references the node's response passes. The `message` name rule and the "must differ from token" rule are dropped.
- ✅ Input-echo and stray end writes into the published output are rejected (`node_publishes_input`, `stray_output_write`).
- ✅ Leftover `folderPath`/`name` is no longer fatal.
- ✅ Teardown pages its listings and removes every folder containing `runId`.
- ✅ Scaffold `AGENTS.md`/`CLAUDE.md` removed. Tenant GUIDs are now placeholders.
- ✅ PR body updated: stacked on #3360, and the checkers are in this PR.

🧪 `pytest tests/tasks/uipath-maestro-bpmn/`: 702 passed, 1 skipped (`coder_eval` not installed).

⏳ Open: there is no passing live coder-eval run on the head yet. Teardown's pagination field names come from one `folders list` call.

@nikhil-maryala ready for re-review.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
```diff
@@ -0,0 +1,405 @@
...
+
+def response_output_vars(task_element: ET.Element) -> set[str]:
+    """Vars the node fills from its own response: `source` absent (the whole
+    response) or reading `result`, e.g. `=result.message`."""
+
+    return {
+        mapping["var"]
+        for mapping in output_mappings_of(task_element)
+        if mapping.get("var")
+        and (not mapping.get("source") or RESULT_REF_RE.search(mapping["source"]))
```
rockymadden: 🟠 **major: the response-provenance rule can be bypassed with a fallback to the input.**

- `response_output_vars` counts any mapping whose `source` mentions `result` anywhere. The regex also matches inside string literals.
- These all publish the process input and give `[]` from `collect_shape_problems` (probed on the `stray_output_write` base with the stray line removed):
  - node: `source="=js:result?.message ?? vars.Var_Name"`
  - node: `source="=js:'result' && vars.Var_Name"`
  - end (`check_shape.py:347`): `source="=js:vars.Var_Message || vars.Var_Name"`
  - end: `source="=js:vars.Var_Message.length ? vars.Var_Name : ''"`
- Behaviour then passes too, because the token is in the input. The `?? vars.Var_Name` fallback is plausible defensive code, and rule 17 warns the result key may differ. So a wrong key still scores full marks.
- `check_behavior.py`'s docstring ("check_shape.py guarantees the published value comes from the node's own response") overstates this.

**Fix:**
- A node response mapping must reference no `vars.` at all: `RESULT_REF_RE.search(source) and not composition.var_refs_in(source)`.
- Every `vars.` ref in an end mapping into the published output must be a response var: `composition.var_refs_in(source) <= response_var_ids`.
- Add a mutation for each case, e.g. `node_falls_back_to_input` and `end_falls_back_to_input`.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,572 @@
...
+    test_checks.py call -- the tests just inject resolvers that return
+    fixture-matching constants instead of shelling out to `uip`.
+
+    `resolve_release_key_for_row(row, project_dir)` lets each row build its
+    own resolver from its own discovered project (v1 has one row; Phase 5's
+    second row needs its own project hint too).
+    """
+
+    uipx_path = composition.find_uipx(root)
+    bpmn_path = composition.find_bpmn(root)
```
rockymadden: 🟡 **minor: the BPMN project is never checked against the `.uipx`** (confirming Copilot).

- Only the API project goes through `project_is_registered`. `valid_unlike_gold` with the `processOrchestration` entry removed from its `.uipx` still gives `[]`.
- The prompt says "Both projects must live inside the one named solution".

**Fix:** after `find_bpmn`, append a problem when `uipx_path` is set and `not composition.project_is_registered(uipx_path, bpmn_path.parent)`. Add an `unregistered_process` mutation.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,572 @@
...
+        meta["name"] for meta in declared.values() if meta.get("kind") == "input" and meta.get("name")
+    })
+    in_props = composition.entry_point_input_properties(entry_points)
+    missing_inputs = [n for n in start_input_names if n not in in_props]
+    if missing_inputs:
+        problems.append(f"entry-points.json input does not publish {missing_inputs}")
+
+    out_props = composition.entry_point_output_properties(entry_points)
+    for row in resources:
+        if row["output"] not in out_props:
```
rockymadden: 🟡 **minor: the prompt's "exactly one" input and output and "(string)" type are stated but not graded** (confirming Copilot).

- `check_published_contract` only checks that `message` is present.
- An extra output, or `message` declared as `json`, passes shape.
- `check_behavior.py:200` does enforce exactly one input, but nothing enforces exactly one output or the type.

**Fix:** require `out_props` keys to equal `{row['output']}`, and `declared[...]['type'] == 'string'` for both the input and the output. Add an `extra_output` mutation. Or drop "exactly one" and "(string)" from the prompt if they're not meant to be graded.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
```diff
@@ -0,0 +1,200 @@
...
+import subprocess
+import sys
+from pathlib import Path
+
+logging.basicConfig(level=logging.INFO, format="teardown: %(message)s")
+logger = logging.getLogger(__name__)
+
+LOGIN_STATUS_TIMEOUT = 30
+DEPLOY_LIST_TIMEOUT = 60
+DEPLOY_UNINSTALL_TIMEOUT = 300
```
rockymadden: 🟡 **minor: one slow uninstall can use up the whole `post_run` budget** (confirming Copilot).

- `post_run` has `timeout: 300`, and this subprocess timeout is also 300. The CLI's own uninstall `--timeout` defaults to 360.
- If the first uninstall stalls, the harness kills teardown before `delete_folder`, which the docstring calls the step that "actually reclaims the tenant".
- `f6e6716` now sweeps several folders, which makes this more likely.

**Fix:** pass `--timeout 90` to `deploy uninstall`, keep the subprocess timeout just above it, and check a total deadline before each call. Or delete folders first, then uninstall whatever is left.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
```diff
@@ -0,0 +1,247 @@
...
+    # Literal, not DEBUG_TIMEOUT_SECONDS: test_criterion_budgets.py prices
+    # run_debug calls statically off the AST and refuses a non-literal
+    # timeout/budget argument. Keep this equal to DEBUG_TIMEOUT_SECONDS
+    # (bpmn_live.DEBUG_BUDGET_DEFAULT_TIMEOUT) above.
+    debug_data, instance_id = run_debug(
+        project_dir, {input_name: token}, Path("debug.log"), timeout=480
+    )
+    print(f"OK: debug completed (instance {instance_id})")
+
+    incidents = run_cli(
```
rockymadden: 🟡 **minor: the incidents call runs even when the debug run completed** (confirming Copilot).

- `run_cli` doesn't catch `subprocess.TimeoutExpired`, so a slow incidents endpoint crashes a criterion that would have passed.
- `assert_completed` returns before it reads `incidents_data` whenever `FinalStatus` is completed, so the call is wasted on every pass.

**Fix:** check `FinalStatus` first, and fetch incidents only on the failure path inside `assert_completed`'s caller.

rockymadden (changes requested):
## 🔴 Changes requested

The earlier findings are fixed, except for the missing live run. But the new provenance check has a bypass: a node can publish the process input and still pass both criteria. Four more gaps came up, three of them from Copilot's review and verified here.

### What
- Re-review of `a82382ff4`. Scope is #3359's own diff against `fix/maestro-bpmn-api-workflow-invocation`.
- Since the last review: `f6e6716` (checker fixes), `46e05d6` (stray end writes, placeholder GUIDs), and the #3360 merges.

### ✅ Fixed since last review
- `validate` now gets the `.bpmn` file.
- The `Data.Warnings` gate reads the real string shape.
- The shape and behaviour checks no longer contradict each other.
- The `message` name rule is gone.
- The "must differ from token" rule is dropped.
- Leftover `folderPath`/`name` is no longer fatal.
- Teardown pages its listings and sweeps `runId` folders.
- The scaffold `AGENTS.md`/`CLAUDE.md` are deleted.
- The debug-run conflict is resolved on the #3360 side.

### ⏳ Still stands
- 🟠 **major**: there is no passing coder-eval run on a current head. The latest runs on this branch are `35284799483` and `35275039852`, both from 09-17 and both failed. `.claude/rules/test-writing.md` step 5 requires a passing run. Fix: dispatch `run-coder-eval.yml` on this head. If it runs out of turns again, raise `max_turns` from 90.

### 🔎 Confirmed vs reasoned
- **Confirmed by probing checker functions with the test resolvers:**
  - the provenance bypass;
  - the unregistered BPMN project passing.
- **Confirmed from UiPath/cli source (`deploy-list.ts`, `packages-list.ts`, `folders.ts`):**
  - every list command returns `Data` as an array;
  - so Copilot's `Data.Deployments` envelope claim is wrong, and teardown's `_list_all` is correct.
  - `deploy list`'s `HasMore` uses the pre-filter `Total`, so paging after `--folder-path` doesn't stop early.
- **Reasoned only:**
  - the teardown time-budget overrun;
  - the incidents call failing an otherwise passing run.
- **Tests:** `pytest tests/tasks/uipath-maestro-bpmn/` gives 702 passed, 1 skipped.

### tl;dr
The earlier findings are fixed, except that there's still no passing live run. The new "must come from the response" rule can be bypassed with `result?.message ?? vars.Var_Name`, which publishes the input. That is the false pass this eval exists to prevent. The other four findings are minor.

nikhil-maryala (approved):
**Re-review of `a82382ff4` — Approve.** Tally: 0 new findings.

The only changes since my approval of `46e05d65a` are the two #3360 commits merged in (rule 18 scoping and the drop wording). Nothing under `tests/` changed. The narrower rule 18 still matches this grader: the task uses `ExecuteApiWorkflowAsync`, the one type where rule 18 keeps `folderKey` and drops `folderId`, and the grader forbids only `folderId`.

My earlier [MINOR] still stands: there's no coder-eval run on a current head (latest `35284799483`, 09-17, `MAX_TURNS_EXHAUSTED`). Get a passing run before taking the PR out of draft.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,572 @@
...
+    declared = composition.declared_variables(process)
+    start_input_names = sorted({
+        meta["name"] for meta in declared.values() if meta.get("kind") == "input" and meta.get("name")
+    })
+    in_props = composition.entry_point_input_properties(entry_points)
+    missing_inputs = [n for n in start_input_names if n not in in_props]
+    if missing_inputs:
+        problems.append(f"entry-points.json input does not publish {missing_inputs}")
+
+    out_props = composition.entry_point_output_properties(entry_points)
```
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in ba2f52d. `check_published_contract` now requires exactly one input property, output properties exactly `{message}`, and every property typed `string`. The new `inexact_contract` fixture is rejected.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
```diff
@@ -0,0 +1,572 @@
...
+    `resolve_release_key_for_row(row, project_dir)` lets each row build its
+    own resolver from its own discovered project (v1 has one row; Phase 5's
+    second row needs its own project hint too).
+    """
+
+    uipx_path = composition.find_uipx(root)
+    bpmn_path = composition.find_bpmn(root)
+    if bpmn_path is None:
+        return ["no .bpmn file found under the solution root"]
+    process = composition.parse_process(bpmn_path)
```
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in ba2f52d. `collect_shape_problems` checks that `bpmn_path.parent` is registered in the `.uipx`. The new `unregistered_bpmn_project` fixture is rejected.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in ba2f52d. Uninstall is capped at 120s, and every call is clamped to a 280s overall deadline, so later phases are skipped cleanly instead of killed.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
```diff
@@ -0,0 +1,200 @@
...
+    return payload.get("Data") if isinstance(payload, dict) else None
+
+
+def _list_all(args: list[str], timeout: int) -> list:
+    rows: list = []
+    for page in range(MAX_PAGES):
+        payload = _run([*args, "--limit", str(PAGE_SIZE), "--offset", str(page * PAGE_SIZE)], timeout)
+        data = _data(payload)
+        if not isinstance(data, list):
+            return rows
```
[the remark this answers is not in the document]
rockymadden: ❌ Not a bug. In UiPath/cli main, `deploy-list.ts`, `packages-list.ts` and `folders.ts` all return `Data` as an array, with pagination in `Pagination`. There is no `Data.Deployments` envelope.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
rockymadden: ✅ Fixed in ba2f52d. A node response mapping may not read any `vars.` variable, and every end mapping into the output must reference only the node's response variables. New fixtures `node_output_falls_back_to_input` and `end_mapping_mixes_input` are rejected, and all four probes now fail.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in ba2f52d (`unregistered_bpmn_project`).

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_shape.py
rockymadden: ✅ Fixed in ba2f52d (`inexact_contract`).

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/_setup/teardown.py
rockymadden: ✅ Fixed in ba2f52d. 120s uninstall cap and 280s overall deadline.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
rockymadden: ✅ Fixed in ba2f52d. Incidents are fetched only when `FinalStatus` isn't completed.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/check_behavior.py
```diff
@@ -0,0 +1,247 @@
...
+        ["uip", "maestro", "bpmn", "debug-instance", "incidents", instance_id],
+        timeout=INCIDENTS_TIMEOUT,
+    )
+    _payload, incidents_data = payload_data(incidents, "incidents", require_success=False)
```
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in ba2f52d. Incidents are fetched only when `FinalStatus` is not completed.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
```diff
@@ -0,0 +1,409 @@
...
+# during discovery -- a mutation fixture sitting under this task's own
+# fixtures/ dir, or an installed node_modules/, must never be found instead
+# of the tree actually being graded.
+EXCLUDED_DIR_NAMES = {"node_modules", "fixtures", ".git"}
+
+GUID_RE = re.compile(
+    r"^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-"
+    r"[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$"
+)
+VARS_REF_RE = re.compile(r"vars\.([A-Za-z0-9_]+)")
```
nikhil-maryala: [MINOR] **The input-fallback bypass is closed only for dotted reads.** `VARS_REF_RE = r"vars\.([A-Za-z0-9_]+)"` feeds `response_output_vars` (l.281-293) and `reads_only_response` (`check_shape.py:348-350`). A bracket read or an optional-chaining read never matches: `re.findall` on `"=js:result?.message ?? vars['Var_Name']"` → `[]`, and on `"…?? vars?.Var_Name"` → `[]`.

Gold copies that all **pass shape**:
- node mapping `=js:result?.message ?? vars['Var_Name']`
- node mapping `=js:result?.message ?? vars?.Var_Name`
- end mapping `=js:vars.Var_Message || vars['Var_Name']`
- end mapping `=js:vars.Var_Message || vars?.Var_Name`

The dotted forms of the same bypasses correctly fail. Local `validate` returns `Valid` on the bracket and optional-chaining copies, and the undeclared-variable check uses the same regex, so nothing else catches them. This is exactly the case `node_output_falls_back_to_input` is meant to cover.

Fix:
- Node response mappings: reject any source that matches `\bvars\b` at all.
- End mappings into `message`: strip the allowed `vars.<response id>` references, then reject if `\bvars\b` still matches.
- Add the bracket and optional-chaining variants as mutation fixtures.

nikhil-maryala (approved):
**Re-review of `ba2f52deb` — Approve.** Tally: 0 Blocker / 0 Critical / 0 Major / 1 Minor / 0 Nit.

The tightening doesn't reintroduce any false failures. I ran one-change copies of gold through `collect_shape_problems` with the test resolvers. These skill-correct variants all still **pass**:
- node variable renamed throughout
- `=js:'Hello, ' + vars.Var_Message` end mapping
- extra `folderPath` or `name` context field
- the skill-taught `=result` object in a json variable with `=js:vars.X.message`
- whole response with no `source`
- `=js:'Hello, ' + result.message` on the node
- an extra `=status` output

These still **fail**, as they should: a stray end write from the input, `folderId`, the node publishing its input, a dotted `?? vars.Var_Name` fallback, `|| vars.Var_Name` mixing, an extra output, and a non-string input in `entry-points.json`.

Results:
- **Fixtures:** all 23 behave as intended. The four new mutations (`end_mapping_mixes_input`, `inexact_contract`, `node_output_falls_back_to_input`, `unregistered_bpmn_project`) each fail for their intended reason only.
- **Tests:** `test_checks.py` 51 passed, and `tests/tasks/uipath-maestro-bpmn/` 706 passed / 1 skipped.
- **Contract:** the new exact-contract check matches the prompt (one string input, exactly one string output named `message`).
- **Other changes:** `check_behavior` changes only when incidents are fetched, and its result is the same. `max_turns` goes from 90 to 150, which addresses the earlier turn exhaustion. Teardown's 280s deadline fits inside the 300s `post_run`.
- **Gates:** `check-task-host-paths` is OK and the criterion budgets pass.

Coder-eval run 36054923953 on this head was still `in_progress` when I posted. My earlier passing-run condition stands until it goes green. The one new Minor is inline.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/api_workflow_composition.yaml
rockymadden: ✅ Passing run on ba2f52d: [36054923953](https://github.com/UiPath/skills/actions/runs/36054923953), `agent=claude`: 4/4 criteria, score 1.000, `status=SUCCESS`, 1029s. `max_turns` raised to 150 in ba2f52d. The run exposed two teardown leaks, fixed in b2093bf. The grading code is unchanged since the run.

rockymadden: ✅ Passing run on ba2f52d: [36054923953](https://github.com/UiPath/skills/actions/runs/36054923953), `agent=claude`: 4/4 criteria, score 1.000, `status=SUCCESS`, 1029s. `max_turns` raised to 150 in ba2f52d. The run exposed two teardown leaks, fixed in b2093bf. The grading code is unchanged since the run.

tests/tasks/uipath-maestro-bpmn/e2e/api_workflow_composition/composition.py
rockymadden: ✅ Fixed in 48bc6a8, as you proposed. Node response mappings reject any `\bvars\b`. End mappings strip the allowed `vars.<response id>` refs, then reject any leftover `vars`. New fixtures for all four variants (`node_output_bracket_fallback`, `node_output_optional_chain_fallback`, `end_mapping_bracket_mix`, `end_mapping_optional_chain_mix`) fail on the previous checker and are rejected now.

nikhil-maryala (approved):
**Re-review of `48bc6a83b` — Approve.** Tally: 0 Blocker / 0 Critical / 0 Major / 0 Minor / 0 Nit.

- **My last [MINOR], the bracket/optional-chaining bypass, is fixed.** `reads_only_response` now strips the allowed `vars.<id>` references and rejects any leftover `\bvars\b` (`ANY_VARS_RE`). The four new mutation fixtures (`node_output_bracket_fallback`, `node_output_optional_chain_fallback`, `end_mapping_bracket_mix`, `end_mapping_optional_chain_mix`) cover it. My own probes on this head:
  - `?? vars.Var_Name`, `?? vars['Var_Name']` and `?? vars?.Var_Name` on the node all FAIL.
  - `|| vars['Var_Name']` and `|| vars?.Var_Name` on the end mapping both FAIL.
  - The skill-correct shapes still PASS: gold, `=js:'Hello, ' + vars.Var_Message`, `=js:'Hello, ' + result.message`, and `source="=result"`.
- **Tests:** `test_checks.py` 51 passed (+6 subtests).
- **Teardown, checked against UiPath/cli main:**
  - It uninstalls every deployment of the run's package in any folder, filtering `deploy list` rows on `PackageName`/`FolderPath`, which `deploy-list.ts` emits.
  - It passes `deploy uninstall --timeout` (a real option).
  - It deletes the package latest version first through `packages list --name` → `Name`/`LatestVersion`. The service returns `name`/`latestVersion`, and `OutputFormatter` PascalCases keys.
  - It stops on the first failed delete and caps at 10 versions.
- **Passing run:** the coder-eval run 36054923953 on `ba2f52deb` completed with **success** (score 1.000). That meets the passing-run condition I'd been holding. The commits since then only tighten graders and teardown.

Ready to leave draft once #3360 lands and this is retargeted to main.

PR #3360, opened 2026-09-16
fix(uipath-maestro-bpmn): correct the API-workflow invocation contract, and delegated-resource guidance

DevMomo:
## Problem

Four defects in `uipath-maestro-bpmn`, each verified against a live tenant (alpha `codereval`/`DefaultTenant`).

1. **The job-wrapper registry template is unrunnable as served.** `validator/bpmn-spec.json:1392` serves `Orchestrator.ExecuteApiWorkflowAsync` with an unbound `releaseKey` placeholder and a context field named `folderId`. Pasted verbatim it passes `validate` and packs clean, then faults at runtime: `170009 Could not get value for key:ReleaseKey from context in input`. Bind `releaseKey` and it faults again — `Could not get value for key:FolderKey` — because the runtime reads a field literally named `folderKey`, not `folderId` and not `folderPath`. Nothing in the skill said so.

2. **`ExecuteApiWorkflowAsync` is documented as fire-and-forget, and a second type is invented.** `references/registry-workflow.md:293-299` says `Orchestrator.ExecuteApiWorkflow` **waits** and `…Async` **returns immediately**. Only `Orchestrator.ExecuteApiWorkflowAsync` exists in `bpmn-spec.json`, and its display label is *"Start and wait for API workflow"* — it waits. Both halves of that sentence are wrong.

3. **"Do not chase warnings" tells agents to discard a real defect signal.** `SKILL.md:260` and `references/structural-bpmn.md:705` say warnings don't fail the run, so fix error-severity only. `Data.Warnings` carries both `VARIABLE_DOES_NOT_EXIST` (a reference with no declaration — always a defect) and `VARIABLE_NOT_SET` on a start-event-scoped caller input (expected by construction). Blanket-ignoring the field hides the first; blanket-fixing it pushes agents to process-scope the input, which silences the warning by emptying `entry-points.json`'s `input` contract so nothing can invoke the process at all.

4. **Nothing told the caller how to incorporate a resource delegated to a sibling skill.** No guidance exists on main (`grep -i delegat skills/uipath-maestro-bpmn/SKILL.md` → nothing). Observed failure: the agent hands authoring to `uipath-api-workflow`/`uipath-rpa`, receives hand-back notes, and stops — project created outside any solution, no `.uipx`, no service task wired into the process.

## Fix

| Defect | Change |
|---|---|
| 1 | New `SKILL.md` rule 18 + `registry-workflow.md` § *Job-wrapper v1 trap*: bind `releaseKey` via the existing process-binding contract to the resource's real `Key`, and add a literal `folderKey` holding the folder's real `FolderKey` GUID (both from `uip or processes list --all-fields --output json`). `name`/`folderPath`/`folderId` are not needed for this type. |
| 2 | `registry-workflow.md` § rewritten: one type, it waits, no fire-and-forget variant to model. |
| 3 | `SKILL.md` step 3 + `structural-bpmn.md` § Validation: read `Data.Warnings` and triage **by code**, with the `entry-points.json`-contract reasoning for why `VARIABLE_NOT_SET` on a start-event input is not a defect. |
| 4 | New `SKILL.md` rule 17: solution first → delegate with an explicit argument contract → deploy before binding → discover the real identity → verify by reading the invoking node's own output after a run. |

Rule 18 flags `StartJob`, `BusinessRules`, and `StartAgenticProcess[Async]`/`StartCaseMgmtProcess[Async]` as sharing the identical broken template, and says so explicitly as **unverified per type** — apply the same fix, then re-verify with a live run.

## Verified

- `uip maestro bpmn validate` clean on the corrected node shape.
- Live `uip maestro bpmn debug` completes; the invoking node's own `Outputs` and the published `message` output both carry the transformed value.
- Reproduced the fault: the node authored exactly as `registry get` serves it fails `170009` every run.

## Note on overlap

The branch also carries the `structural-bpmn.md` minimal-example `elementId` hunk. That fix **landed independently on `main`** in e954c0762 (#3379) with identical content, so it is now a no-op here — `git merge-tree origin/main <head>` is clean and the net delta against today's `main` is only the four items above. Left in place rather than rewriting history, since #3361 and #3359 are stacked on this branch.

The four PRs flagged as overlapping when this work started (#2366, #2387, #3155, #3113) all merged before this branch was cut, and none of them touches any of the four defects above. `refresh` vs `update-metadata` is already correct on `main` (`SKILL.md:391`, from #2387) and is **not** part of this diff.

- [ ] Reviewed by a `uipath-maestro-bpmn` codeowner


skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -278,15 +278,22 @@ For registry-evidence-only tasks, follow the command-first recipe in
    ```
 
    Exit 0 = valid; exit 1 = validation failed (the envelope lists each issue
-   with its rule code). Warnings do not fail the run: validate once, fix only
-   error-severity findings, and do not re-validate in a loop chasing warnings.
-   Two warnings are defects rather than noise, because no error covers them.
+   with its rule code). `Status: "Valid"` is not the whole verdict: warnings do
```
nikhil-maryala: [MAJOR] **Merge conflict with main, and this side would undo two approved fixes.** `git merge-tree --write-tree origin/main c362f5477` → `CONFLICT (content): Merge conflict in skills/uipath-maestro-bpmn/SKILL.md`, in exactly this paragraph. Since this branch was cut, main has rewritten it (#3533, #3536):
- Severity comes from each issue's `[error]`/`[warning]` tag, not the `Found N error(s)` header.
- `MISSING_RESOURCE` (warning) plus `MISSING_BINDING` (error) is one finding, and it gives exit 1 / `RetryWillNotFix` for a requested placeholder.
- Re-validation stops once only warnings remain.
- The `unknown attribute <type>` warning is benign noise.

This branch's version still says an unresolved node "warns `MISSING_RESOURCE` by design", which is the wording #3533 corrected, because `MISSING_BINDING` is an error. On rebase, keep main's paragraph as is, and add only the new `VARIABLE_DOES_NOT_EXIST` / `VARIABLE_NOT_SET` triage sentences.

skill-flavors/studioweb/uipath-maestro-bpmn/SKILL.md
```diff
@@ -0,0 +1,6 @@
+<!--skill-flavor:delegated-resource-solution-first:start-->
+   (1) Studio Web already has one open solution as the workspace root —
+   never create a new one. Author the resource's project **inside** that
+   solution (with the resource's own `init` command, e.g. `uip rpa init` or
+   `uip api-workflow init`) so it registers in the `.uipx`.
```
nikhil-maryala: [MAJOR] **The Studio Web override only replaces step (1), and step (1) is itself off for API workflows.**
- **Project creation.** The studioweb `uipath-api-workflow` flavor says "Use `CreateProjects` for project creation", not `uip api-workflow init`. The studioweb `uipath-rpa` flavor lets `uip rpa init` create the project, then stops the agent (no authoring).
- **Deploy.** Steps (3) to (5) are outside the block, so the Studio Web build still tells the agent to "pack, publish, `solution deploy run`" and re-read `deploy list`. In Studio Web, publication goes through host-intercepted `uip solution publish` after explicit approval (`skill-flavors/studioweb/uipath-api-workflow/SKILL.md:16,70,84`). That is an outward-facing action this rule would have the agent take unprompted.

Suggest widening the block to cover (1) and (3):
- create via the host (`CreateProjects` for API workflows, `uip rpa init` for RPA)
- publish only via the approved host publish bridge
- then discover the keys with `or processes list`

Or add a studioweb-specific (3). Verified: `npm run skills:build` → the studioweb tree shows this block followed by the unchanged canonical (2) to (5).

skills/uipath-maestro-bpmn/references/structural-bpmn.md
```diff
@@ -733,7 +733,25 @@ each with its rule code (gateway/condition, fake-join, superfluous-gateway,
 error end/boundary event, timer-duration/required-field, single-blank-start,
 single-conditional-outgoing-flow, variable-reference, method-parentheses,
 input-type, event-object, and IS-connector checks). Warnings are reported but do
-not block. If `validate` is unknown or runs only deploy-readiness checks, update
+not block — that does not mean ignore them; read `Data.Warnings` and triage by
```
nikhil-maryala: [MINOR] **`Data.Warnings` only exists when validate passes.** main's `references/cli-conventions.md:75` (#3539) documents the envelope:
- **pass:** `Data.Warnings` is a single string.
- **fail (exit 1):** there is no `Data`, and every issue, warnings included, is in `Instructions`.

So on the run where the agent most needs to find a `VARIABLE_DOES_NOT_EXIST`, which is alongside an error, "read `Data.Warnings`" points at a field that isn't there. Suggest: "read the warnings (`Data.Warnings` on a pass; the `[warning]` entries in `Instructions` on a failure) and triage by code". The same applies to the matching `SKILL.md` step-3 sentence.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   --all-fields --output json` and bind the invoking node to it — rule 18's
+   verified contract — never a fabricated or placeholder key. (5) Verify by
+   running the process (`uip maestro bpmn debug`) and reading the invoking
+   node's own output in `debug-instance variables-all` — a clean
+   `validate`/`pack` proves nothing about whether the resource was ever
+   reached at runtime, or which key its output landed under. Map
+   `=result.<key>` to the key the real response used, never the declared
+   output-schema property name — a scalar output can surface under a
+   generic key instead, and a stale assumed key breaks the mapping silently.
+18. **Job-wrapper registry templates ship a resolvable-looking but broken
```
nikhil-maryala: [MINOR] **This contradicts Critical Rule 6 and doesn't reference it.** Rule 6 (`SKILL.md` ~l.388) says "Paste the selected registry template literally and do not normalize one form into the other." Rule 18 opens with "do not paste it as served". An agent that reads rule 6 first has two absolute rules pointing opposite ways, and contradictory rules have been a measured cause of turn-budget exhaustion in this batch. Add the exception at the point of use: in rule 6, "…literally (except the job-wrapper context fix in rule 18)". Or open rule 18 with "Exception to rule 6:".

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   release key and folder key, which exist only once deployed; re-read the
+   deployment key from `deploy list` after an upgrade, since upgrading rotates
+   it and reusing the stale key gives `4005`. (4) Discover the deployed
+   resource's identity with `uip or processes list --folder-path <path>
+   --all-fields --output json` and bind the invoking node to it — rule 18's
+   verified contract — never a fabricated or placeholder key. (5) Verify by
+   running the process (`uip maestro bpmn debug`) and reading the invoking
+   node's own output in `debug-instance variables-all` — a clean
+   `validate`/`pack` proves nothing about whether the resource was ever
+   reached at runtime, or which key its output landed under. Map
```
nikhil-maryala: [MINOR] **This contradicts the guidance it sits beside.** "Map `=result.<key>` to the key the real response used, never the declared output-schema property name" clashes with `references/registry-workflow.md` right after the new sections: "map those fields as `uipath:output` rows … using the discovered output names/types … (or the exact schema fields returned by discovery)". One rule says trust the discovered schema, the other says never trust it. If a scalar output can surface under a generic key, say which key and when (e.g. "a scalar output is returned under `<key>`; read the actual key from `variables-all` and prefer it over the schema name"), and add the same caveat to the registry-workflow paragraph so the two agree.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   `=result.<key>` to the key the real response used, never the declared
+   output-schema property name — a scalar output can surface under a
+   generic key instead, and a stale assumed key breaks the mapping silently.
+18. **Job-wrapper registry templates ship a resolvable-looking but broken
+   context; do not paste it as served.** `Orchestrator.StartJob`,
```
nikhil-maryala: [NIT] Rule 18 is ~28 lines and restates `registry-workflow.md` § *Job-wrapper v1 trap* almost clause for clause: fault codes, the `folderKey` explanation, the per-type caveat. Per `.claude/rules/token-optimization.md` § 0, the "do not paste as served / bind `releaseKey` / add literal `folderKey` / re-verify other types" rule text should stay in full. The fault-code narrative and the second copy of the evidence can go down to the existing pointer.

nikhil-maryala (changes requested):
**Request changes.** Tally: 0 Blocker / 0 Critical / 2 Major / 3 Minor / 1 Nit.

The core fixes are right, and I verified what I could offline:
- **One API-workflow type, and it waits.** `Orchestrator.ExecuteApiWorkflowAsync` is the only API-workflow type in `@uipath/maestro-tool@1.202.2`, and its label is `Start and wait for API workflow`. Its `RESOURCE_TYPE_METADATA.Api` entry has no `asyncServiceType`, so the designer's fire-and-forget toggle, which only swaps to `asyncServiceType`, never applies to API workflows. The old wait/fire-and-forget section was wrong on both halves.
- **The family claim holds.** In `packages/maestro-sdk/src/manifest/bpmn-spec.json` (UiPath/cli main), `StartJob`, `BusinessRules`, `ExecuteApiWorkflowAsync`, `StartAgenticProcess[Async]` and `StartCaseMgmtProcess[Async]` all serve context `[releaseKey, folderId, folderPath, name]` with a `process`/`releaseKey`/`Key` binding. `StartAgentJob` (`[name, folderPath]`) is correctly left out.
- **Cited commands exist.** Every command rule 17 cites is in the catalog snapshot: `or processes list --folder-path --all-fields`, `solution deploy run|list`, `debug-instance variables-all`.
- **Build.** `skills:validate` and `skills:build` pass, and the studioweb block composes into the built tree.
- **Author's evidence.** The runtime `folderKey` claim (`Could not get value for key:FolderKey`) is live behaviour I can't reproduce offline, so it rests on the author's run.

Blocking:
1. **The PR no longer merges.** `SKILL.md` conflicts with main in the step-3 validate paragraph (`git merge-tree origin/main <head>` → CONFLICT). Taking this branch's side would undo #3533 and #3536, which landed there since.
2. **The new studioweb block only swaps step (1).** Its `uip api-workflow init` isn't how Studio Web creates API-workflow projects, and steps (3) to (5) still teach local-CLI deploy, where Studio Web publishes through the host bridge after approval.

The minors are two sentences that contradict existing guidance and a `Data.Warnings` pointer that only exists on success. Details are inline.

Merge order: #3359 is stacked on this branch, so land this first. #3361 also stacks on it.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: 🟠 **major** — conflicts with `main` and would undo #3533 and #3536.

- `git merge-tree origin/main HEAD` → `CONFLICT (content)` in this file.
- This side brings back "validate once… do not re-validate" and "an unresolved node warns `MISSING_RESOURCE` by design". `main` now treats `MISSING_RESOURCE` + `MISSING_BINDING` as one pair, expects exit 1 for a draft, and says to read severity from the `[error]`/`[warning]` tag.

**Fix:** rebase and keep `main`'s paragraph as is. Append only the two `VARIABLE_*` sentences and the link to `structural-bpmn.md#validation`.

skill-flavors/studioweb/uipath-maestro-bpmn/SKILL.md
```diff
@@ -0,0 +1,6 @@
+<!--skill-flavor:delegated-resource-solution-first:start-->
+   (1) Studio Web already has one open solution as the workspace root —
```
rockymadden: 🟠 **major** — the Studio Web path is wrong in several places, and steps (3)–(5) are not overridden at all.

- `uip api-workflow init` is not how Studio Web creates API-workflow projects. The studioweb `uipath-api-workflow` flavor, rule 19, says to use `proxy-tools-Solution` / `CreateProjects`.
- The studioweb `uipath-rpa` host-scope block says authoring RPA is not supported in Studio Web. So step (2), "delegate authoring to `uipath-rpa`", cannot happen there.
- Step (3) says "pack, publish, `solution deploy run`". The studioweb `uipath-solution` flavor has no `pack`, publishing to the personal workspace auto-deploys, and `deploy run` creates a new folder every time.

**Fix:** wrap step (3) in its own marker block too. The Studio Web override should say: create API-workflow projects with `CreateProjects`; for RPA, stop and hand off to the user; ship with `uip solution publish`, plus `deploy run` only for a shared destination.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   `releaseKey` (`170005` for `StartJob`, `170009` for
+   `ExecuteApiWorkflowAsync` — `Could not get value for key:ReleaseKey from
+   context in input`). Populating `releaseKey` alone is not enough either: the
+   runtime then reports `Could not get value for key:FolderKey`, because the
+   context field the runtime actually reads is `folderKey` (the folder's
+   `FolderKey` GUID) — **not** the template's own `folderId` field, and not
+   `folderPath`. Verified end-to-end for `Orchestrator.ExecuteApiWorkflowAsync`:
+   `<uipath:context>` with **only** `releaseKey` (bound via `=bindings.<id>` to
+   the resource's real `Key`, per the existing resource-binding contract) and
+   a literal `folderKey` (the target folder's real `FolderKey` GUID) —
```
rockymadden: 🟠 **major** — a literal `folderKey` pins one folder's GUID into the `.bpmn`.

- `releaseKey` goes through a binding, so solution deploy can re-resolve it. `folderKey` is a plain literal, so nothing re-resolves it.
- `deploy run` creates a new folder on every run (studioweb `uipath-solution` flavor). A second `deploy run`, or a deploy to another tenant, leaves the node pointing at the old folder. It will either fault or invoke the stale copy.
- This is reasoning from the docs, not a live run.

**Fix:** check whether a process binding with `propertyAttribute="folderKey"` resolves, the same way the Connection folder binding in §4 does. If it doesn't, state that the literal must be re-discovered after every `deploy run` into a new folder.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: 🟠 **major** — rule 18 repeats `registry-workflow.md` § *Job-wrapper v1 trap* almost word for word: fault codes, the `folderKey` story, and the "verified end-to-end" narrative. That puts about 26 extra lines into a `SKILL.md` that loads on every BPMN task, and two copies will drift. #3542 just indexed `structural-bpmn.md` to cut reads in the other direction.

**Fix:** reduce rule 18 to the action and a link, e.g.:
> 18. **Job-wrapper context (`ExecuteApiWorkflowAsync`, `StartJob`, `BusinessRules`, `StartAgenticProcess*`, `StartCaseMgmtProcess*`): do not paste the served context.** Bind `releaseKey` and add a literal `folderKey`; drop `folderId`/`folderPath`/`name`. `ExecuteApiWorkflowAsync` waits. See [registry-workflow.md](…#job-wrapper-v1-trap--releasekey-templates-are-unrunnable).

Also: rule 6 says "Paste the selected registry template literally", and this rule says the opposite with no reconciliation. Add "Rule 6's type shape still applies; only the context changes."

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   deployment key from `deploy list` after an upgrade, since upgrading rotates
+   it and reusing the stale key gives `4005`. (4) Discover the deployed
+   resource's identity with `uip or processes list --folder-path <path>
+   --all-fields --output json` and bind the invoking node to it — rule 18's
+   verified contract — never a fabricated or placeholder key. (5) Verify by
+   running the process (`uip maestro bpmn debug`) and reading the invoking
+   node's own output in `debug-instance variables-all` — a clean
+   `validate`/`pack` proves nothing about whether the resource was ever
+   reached at runtime, or which key its output landed under. Map
+   `=result.<key>` to the key the real response used, never the declared
```
rockymadden: 🟡 **minor** — this contradicts `registry-workflow.md:382`, which still says to map `source="=result"` "(or the exact schema fields returned by discovery)". An agent that reads both gets opposite instructions.

**Fix:** update `registry-workflow.md:382` to match: map the key the debug run's `variables-all` shows, not the schema property.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   `uipath-api-workflow`, `uipath-rpa`) with an explicit argument contract —
+   declared inputs and outputs, not an unauthored scaffold. (3) Deploy the
+   resource (pack, publish, `solution deploy run`) **before** binding it into
+   the BPMN node — the invoking node's context needs the resource's real
+   release key and folder key, which exist only once deployed; re-read the
+   deployment key from `deploy list` after an upgrade, since upgrading rotates
+   it and reusing the stale key gives `4005`. (4) Discover the deployed
+   resource's identity with `uip or processes list --folder-path <path>
+   --all-fields --output json` and bind the invoking node to it — rule 18's
+   verified contract — never a fabricated or placeholder key. (5) Verify by
```
rockymadden: 🟡 **minor** — step (5) requires a live `debug` for every delegated resource, with no scope condition.

- That clashes with draft and boundary-handoff tasks, which Rule 16 and step 3 of the workflow allow.
- #3359's prompt forbids the debug run.
- A debug run invokes the real API workflow, so it has side effects.

**Fix:** scope it, e.g. "(5) When the deliverable is runnable, verify by…".

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   (1) Create or open the solution **first** (`uip solution init`), and
+   author the resource's project **inside** it, so it registers in the
+   `.uipx` — a project created outside any solution has no path to deployment.
+<!--skill-flavor:delegated-resource-solution-first:end-->
+   (2) Delegate authoring to the resource's owning skill (e.g.
+   `uipath-api-workflow`, `uipath-rpa`) with an explicit argument contract —
+   declared inputs and outputs, not an unauthored scaffold. (3) Deploy the
+   resource (pack, publish, `solution deploy run`) **before** binding it into
+   the BPMN node — the invoking node's context needs the resource's real
+   release key and folder key, which exist only once deployed; re-read the
```
rockymadden: 🟡 **minor** — the `deploy list` / `4005` key-rotation guidance belongs to `uipath-solution`, and that skill already documents it. It does not change how the BPMN node gets authored.

**Fix:** drop the clause and link `uipath-solution` for deploy/upgrade.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   output-schema property name — a scalar output can surface under a
+   generic key instead, and a stale assumed key breaks the mapping silently.
+18. **Job-wrapper registry templates ship a resolvable-looking but broken
+   context; do not paste it as served.** `Orchestrator.StartJob`,
+   `Orchestrator.ExecuteApiWorkflowAsync`, `Orchestrator.BusinessRules`, and
+   `Orchestrator.StartAgenticProcess[Async]` / `StartCaseMgmtProcess[Async]`
+   all serve the same template: a hidden, unbound `releaseKey` context field
+   plus a **misnamed** `folderId` field. Pasted verbatim it passes `validate`
+   and packs clean, then faults at runtime because nothing ever resolves
+   `releaseKey` (`170005` for `StartJob`, `170009` for
```
rockymadden: 🟡 **minor** — `170005` for `StartJob` is stated as fact here and at `registry-workflow.md:344`. The PR body says only `ExecuteApiWorkflowAsync` was verified. I could not verify the `StartJob` code.

**Fix:** drop the `StartJob` code, or mark it as observed.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -424,6 +431,59 @@ and honestly surfaced to the user as gaps when asked.
...
+   `ExecuteApiWorkflowAsync` — `Could not get value for key:ReleaseKey from
+   context in input`). Populating `releaseKey` alone is not enough either: the
+   runtime then reports `Could not get value for key:FolderKey`, because the
+   context field the runtime actually reads is `folderKey` (the folder's
+   `FolderKey` GUID) — **not** the template's own `folderId` field, and not
+   `folderPath`. Verified end-to-end for `Orchestrator.ExecuteApiWorkflowAsync`:
```
rockymadden: 🟡 **minor** — several phrases fail the "does this change what the agent does" test: "Verified end-to-end for…", "despite its name" (line 481), and "do not assume without checking" (line 485, which repeats "re-verify with a live run"). Same for `registry-workflow.md:372-377`: the "property of the wrapper family" paragraph restates the list above it plus "re-verify".

**Fix:** cut them and keep the instructions.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
rockymadden: 🟡 **minor** — `Data.Warnings` is a single string and exists only when validate passes. On failure there is no `Data`, and every issue, warnings included, is in `Instructions` (`cli-conventions.md:75`). So "read `Data.Warnings` and triage by code" finds nothing on the exit-1 path, which is where `VARIABLE_DOES_NOT_EXIST` usually shows up.

**Fix:** "Read warnings from `Data.Warnings` (pass) or `Instructions` (fail) and match the codes by substring."

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -318,14 +318,63 @@ Action dropdown** in Studio Web. Do not use it for a folder-deployed agent — t
...
+the template's second bug.** Verified end-to-end for
+`Orchestrator.ExecuteApiWorkflowAsync` against a live deployed API workflow:
+
+1. **Bind `releaseKey`** via the resource-binding mechanism this field's
+   `bindingInfo` already documents (see [§4
+   Bindings](#4-bindings--from-bindinginfo-never-invented) above): a
+   process-kind `<uipath:binding resource="process" propertyAttribute="Key"
+   default="<resolved-key>" />`, referenced from the context as
+   `=bindings.<id>`. Resolve `<resolved-key>` from `uip or processes list
+   --folder-path <path> --all-fields --output json` → the deployed resource's
```
rockymadden: 🟡 **minor** — `--all-fields` returns the raw DTO, and `uipath-platform/references/orchestrator/orchestrator.md:20` says raw keys are camelCase. This line and 364 name `Key` / `FolderKey`. I could not verify the casing. If it's wrong, an agent filtering on `FolderKey` gets nothing and may invent a value.

**Fix:** give the exact key names from the live output, e.g. `key` / `folderKey`, or a JMESPath example.

rockymadden (changes requested):
🔴 **Changes requested** — the core contract fix is right, but the branch conflicts with `main` in a way that undoes two merged fixes, the Studio Web override is wrong, and the new `folderKey` literal pins a folder GUID that `deploy run` will not keep.

### What
- Fixes the `Orchestrator.ExecuteApiWorkflowAsync` context: bind `releaseKey`, add a literal `folderKey`, drop `folderId`/`folderPath`/`name`.
- Rewrites the API-workflow section: one type, and it waits.
- Adds `VARIABLE_DOES_NOT_EXIST` / `VARIABLE_NOT_SET` triage to validation guidance.
- Adds rule 17 (incorporate a delegated resource in five steps) and rule 18 (job-wrapper template trap), plus a Studio Web override for rule 17 step 1.

### Checked against the diff
- ✅ `bpmn-spec.json` has only `ExecuteApiWorkflowAsync`, labelled `Start and wait for API workflow [Preview]`.
- ✅ `StartJob`, `BusinessRules`, `StartAgenticProcess[Async]`, `StartCaseMgmtProcess[Async]` serve the same `releaseKey`/`folderId`/`folderPath`/`name` context.
- ✅ CI is green.
- ⚠️ The runtime `folderKey` requirement rests on the author's live run. I could not verify it.

### Status of the earlier review (head `c362f54` predates it, so nothing is fixed yet)
- ❌ SKILL.md step 3 conflict: still open.
- ❌ Studio Web block covers only step (1): still open.
- ❌ `Data.Warnings` shape, rule 18 vs rule 6, rule 17 vs registry-workflow output names: all still open.

### Overall findings
None beyond the inline comments.

### tl;dr
The API-workflow fix is correct in substance. Rebase onto `main` without undoing #3533/#3536, fix the Studio Web path, make `folderKey` survive a redeploy (or say it won't), and cut rule 18 down to a pointer. #3359 is stacked on this, so it waits on these fixes.

skills/uipath-maestro-bpmn/references/registry-workflow.md
```diff
@@ -319,20 +319,66 @@ Action dropdown** in Studio Web. Do not use it for a folder-deployed agent — t
...
+   `folderKey`, not `folderId` or `folderPath`.** Populating `releaseKey`
+   alone still faults with `Could not get value for key:FolderKey from
+   context in input`. Add a context field literally named `folderKey` holding
+   the target folder's `FolderKey` GUID (the same `or processes list`
+   response carries it as `FolderKey`) — a plain literal value, not a binding.
+   Re-read it after every deploy: `deploy run` creates a new folder.
+3. With both fields correct — `releaseKey` bound to the resource's real `Key`,
+   `folderKey` literal to the folder's real `FolderKey` — the node runs to
+   completion. The template's own `folderId`/`folderPath`/`name` context
+   fields are not needed for `Orchestrator.ExecuteApiWorkflowAsync` and can be
```
nikhil-maryala: [NIT] **Leftover softer wording.** This still says the `folderId`/`folderPath`/`name` fields "are not needed … and can be dropped", while rule 18 now says "Drop `folderId`, `folderPath`, and `name`". #3359's grader also treats all three as forbidden. Make this say "drop them" so the reference, the rule and the eval agree.

nikhil-maryala (approved):
**Re-review of `73a28e500` — Approve.** Tally: 0 Blocker / 0 Critical / 0 Major / 0 Minor / 1 Nit. This replaces my Request changes.

All six earlier findings are addressed:
- **[MAJOR] merge conflict: resolved.** `git merge-tree origin/main <head>` is clean. The step-3 paragraph keeps main's #3533/#3536 wording verbatim (`[error]`/`[warning]` tags, the `MISSING_RESOURCE` + `MISSING_BINDING` pair, `RetryWillNotFix`, the `unknown attribute <type>` note) and adds only the `VARIABLE_*` triage.
- **[MAJOR] Studio Web flavor: resolved.** A second block, `delegated-resource-author-deploy`, now covers steps (2) and (3): `CreateProjects` for API workflows, RPA stopping at the `uip rpa init` scaffold, and host `uip solution publish` only after approval. It defers to `uipath-solution` for destinations and redeploys. In the built studioweb tree the rule reads as one coherent sequence, and no markers are left.
- **[MINOR] `Data.Warnings`: resolved.** Pass → `Data.Warnings` (one string); fail → `Instructions`.
- **[MINOR] rule 6 vs rule 18: resolved.** Rule 18 now states it is "the one exception to rule 6's paste-literally".
- **[MINOR] result key: resolved.** Rule 17 step (5) and `registry-workflow.md` now agree: use the key a debug run showed; otherwise use the schema name and report it as unverified.
- **[NIT] rule 18 length: resolved.** It's down to the rule, with a pointer to the reference.

New claims, checked:
- **`--all-fields` dropped from `or processes list`.** That's fine. `formatProcessListRow` in `packages/orchestrator-tool/src/commands/processes.ts` (UiPath/cli main) returns `Key`, `FolderKey` and `FolderPath` by default.
- **"`deploy run` creates a new folder, re-read `folderKey`".** That matches `uipath-solution` (`pack-and-deploy.md` § *`deploy run` Always Creates a New Folder*, and the studioweb flavor's l.17).

`skills:validate` and `skills:build` both pass. #3359 can be retargeted to main once this lands.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 31cf6db (merged `main`, kept main's validate paragraph) and 73a28e5 (added only the `VARIABLE_*` triage).

skill-flavors/studioweb/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Step (1) uses `CreateProjects` for API workflows. New `delegated-resource-author-deploy` block covers (2)–(3): no pack, RPA stops at scaffold, publish via host bridge after approval.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
rockymadden: ✅ Fixed in 73a28e5. `structural-bpmn.md` now says `Data.Warnings` is one string on pass; failures land in `Instructions`.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Rule 18 now names itself the one exception to rule 6.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Rule 17 and `registry-workflow.md` agree: use the key a debug run showed; without a run, use the schema name and report it unverified.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Rule 18 cut to ~9 lines: the action plus a link.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 31cf6db + 73a28e5. Main's paragraph kept, only the `VARIABLE_*` triage added.

skill-flavors/studioweb/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. `CreateProjects` for API workflows, no `uipath-rpa` authoring, no pack, host-bridge publish after approval; steps (2)–(3) now overridden.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Step (4) and `registry-workflow.md` now say re-read `Key`/`FolderKey` after every deploy. A `folderKey` binding is untested, so not taught.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Rule 18 is the action, the rule-6 exception, and a link.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Both places now give the same guidance.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. The debug run happens only when the task allows live runs and isn't a draft or handoff.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Replaced with a pointer to `uipath-solution`'s upgrade path.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. `170005` removed; only the verified `170009` for `ExecuteApiWorkflowAsync` remains.

skills/uipath-maestro-bpmn/SKILL.md
rockymadden: ✅ Fixed in 73a28e5. Narration removed.

skills/uipath-maestro-bpmn/references/structural-bpmn.md
rockymadden: ✅ Fixed in 73a28e5. String on pass, `Instructions` on failure.

skills/uipath-maestro-bpmn/references/registry-workflow.md
rockymadden: ✅ Fixed in 73a28e5. `--all-fields` dropped. The CLI's default `processes list` output returns PascalCase `Key`/`FolderKey` (`orchestrator-tool/src/commands/processes.ts`).

rockymadden:
🔧 **Review findings addressed** (73a28e5, plus 31cf6db merging `main`)

- ✅ Merge conflict resolved. Main's validate paragraph is kept, and only the `VARIABLE_*` triage is added.
- ✅ Studio Web: `CreateProjects` for API workflows, no RPA authoring, no pack, publish through the host bridge after approval.
- ✅ `folderKey`: re-read `Key`/`FolderKey` after every deploy.
- ✅ Rule 18 is ~9 lines and names itself the exception to rule 6.
- ✅ Rule 17 and `registry-workflow.md` give the same output-key guidance.
- ✅ The debug run is conditional (skipped for draft/handoff or when live runs are forbidden).
- ✅ Removed the `4005` rotation text, the unverified `170005` code, narration, and `--all-fields`. Fixed the `Data.Warnings` description.

⚠️ Unverified: the runtime needing `folderKey`, and the output landing under the observed key, both rest on one live run.

@nikhil-maryala ready for re-review.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -505,6 +508,42 @@ and honestly surfaced to the user as gaps when asked.
...
+   declared inputs and outputs, not an unauthored scaffold. (3) Deploy the
+   resource (pack, publish, `solution deploy run`) **before** binding it into
+   the BPMN node; its release key and folder key exist only once deployed.
+   To redeploy, follow `uipath-solution`'s upgrade path.
+<!--skill-flavor:delegated-resource-author-deploy:end-->
+   (4) Read the deployed resource's `Key` and `FolderKey` from
+   `uip or processes list --folder-path <path> --output json` and bind the
+   node per rule 18, never a fabricated or placeholder key. Re-read both after
+   every deploy: `deploy run` creates a new folder, so a literal `folderKey`
+   from an earlier deploy points at the old one. (5) Unless the task forbids
```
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in 292c53f. Step (4) now picks the wrapper by `processType` first. A low-code agent uses `StartAgentJob` and its own template (`name`/`folderPath`), not rule 18.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -505,6 +508,42 @@ and honestly surfaced to the user as gaps when asked.
    generate them — never the deprecated `update-metadata`. An
    Integration Service draft or boundary handoff asks for none of those — emit
    only the `.bpmn` plus a `.md` notes file naming the CLI-owned blockers.
+17. **Incorporating a resource delegated to a sibling skill (RPA workflow, API
+   workflow, agent) is a five-step sequence, in this order. Stopping after
+   the owning skill hands the resource back is not done.**
+<!--skill-flavor:delegated-resource-solution-first:start-->
+   (1) Create or open the solution **first** (`uip solution init`), and
+   author the resource's project **inside** it, so it registers in the
+   `.uipx`. A project created outside any solution has no path to deployment.
```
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in 292c53f. Step (1) falls back to `uip solution new` on `unknown command`. Step (3) passes no renamed flags, so it needs no fallback.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -505,6 +508,42 @@ and honestly surfaced to the user as gaps when asked.
...
+   property name. Without a run, use the schema name and report the mapping
+   as unverified.
+18. **Job-wrapper registry templates (`Orchestrator.StartJob`,
+   `ExecuteApiWorkflowAsync`, `BusinessRules`, `StartAgenticProcess[Async]`,
+   `StartCaseMgmtProcess[Async]`) serve a context that validates but faults at
+   runtime. This is the one exception to rule 6's paste-literally.** In
+   `<uipath:context>`, keep only `releaseKey` bound via `=bindings.<id>` to
+   the resource's `Key`, and add a literal `folderKey` with the folder's
+   `FolderKey`. Drop `folderId`, `folderPath`, and `name`. Verified only for
+   `ExecuteApiWorkflowAsync`; re-verify the others with a live run. Details:
```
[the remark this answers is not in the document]
rockymadden: ✅ Fixed in 292c53f. Rule 18 binds `releaseKey` for all listed wrappers, per the spec's `bindingInfo`. The `folderKey` swap and field drop apply only to `ExecuteApiWorkflowAsync`; the others need a live `key:FolderKey` fault first, otherwise the node is reported unverified. `registry-workflow.md` matches.

skills/uipath-maestro-bpmn/references/registry-workflow.md
rockymadden: ✅ Fixed in 8ec4040. It now says "drop" for `ExecuteApiWorkflowAsync`. Per Copilot's finding (292c53f), the other wrappers make the swap only after a live `key:FolderKey` fault.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -505,6 +508,49 @@ and honestly surfaced to the user as gaps when asked.
...
+<!--skill-flavor:delegated-resource-solution-first:end-->
+<!--skill-flavor:delegated-resource-author-deploy:start-->
+   (2) Delegate authoring to the resource's owning skill (e.g.
+   `uipath-api-workflow`, `uipath-rpa`) with an explicit argument contract:
+   declared inputs and outputs, not an unauthored scaffold. (3) Deploy the
+   resource (pack, publish, `solution deploy run`) **before** binding it into
+   the BPMN node; its release key and folder key exist only once deployed.
+   To redeploy, follow `uipath-solution`'s upgrade path.
+<!--skill-flavor:delegated-resource-author-deploy:end-->
+   (4) Pick the wrapper by `processType`
```
rockymadden: 🟡 **minor: step (4) routes by `processType`, but the command it names doesn't return it.**

- The default `uip or processes list` row (`formatProcessListRow`, UiPath/cli main) has `Key`, `FolderKey` and `FolderPath`, but no `ProcessType`.
- The linked section gets `processType` from `--all-fields`, which returns camelCase `key`/`folderKey`, so an agent that makes one call finds no `Key`.
- **Fix:** name one call that has all three, e.g. `uip or processes get <KEY> --output json`, whose `formatProcessDetail` returns `ProcessType`, `Key` and `FolderKey`. Or say to take `processType` from `--all-fields` and the keys from the default list.

rockymadden (approved):
## 🟢 Approved

Every earlier finding is fixed, and so are Copilot's three. One new minor finding is inline.

### What
Fixes the `uipath-maestro-bpmn` contract for invoking a delegated resource:
- `ExecuteApiWorkflowAsync` waits for completion. There is no fire-and-forget variant.
- The shared job-wrapper template leaves `releaseKey` unresolved, and `folderKey` is misnamed as `folderId`.
- Rule 17 is the five-step incorporate sequence. Rule 18 is the one exception to rule 6.
- Adds `VARIABLE_*` warning triage and a Studio Web flavor for steps (1)–(3).

### ✅ Fixed since the last review
- Merge conflict: `git merge-tree origin/main HEAD` is clean, main's step-3 paragraph is kept, and only the `VARIABLE_*` triage is added.
- Studio Web: `CreateProjects`, RPA stops at the scaffold, publish goes through the host bridge after approval, and there is no pack. In the built tree, rule 17 reads as one sequence with no markers left.
- Literal `folderKey`: it is re-read after every deploy (`SKILL.md:533`, `registry-workflow.md:368`).
- Rule 18 is trimmed and names itself the exception to rule 6.
- Rule 17 and `registry-workflow.md:381` agree on the output key.
- The debug run is conditional (`SKILL.md:536`).
- Removed: the `4005` rotation text, the unverified `170005`, narration, and `--all-fields` on the key read. `Data.Warnings` is described correctly.
- Copilot findings:
  - Agents are routed by `processType` before rule 18 applies.
  - `solution new` is the fallback.
  - The `folderKey` swap is limited to `ExecuteApiWorkflowAsync`, and the other wrappers need a live fault first.

### 🔎 Checked
- Confirmed:
  - All anchors resolve (`#validation`, `#variables`, `#4-bindings--…`, `#agent-wrapper-selection--…`, `#job-wrapper-v1-trap--…`).
  - `skills:validate` and `skills:build` pass.
  - CI: 34 passed, 2 pending, 0 failed.
  - No test or skill still references `Orchestrator.ExecuteApiWorkflow` (non-async).
  - `formatProcessListRow` (UiPath/cli main) returns PascalCase `Key` and `FolderKey`.
- Reasoned only: the runtime needing `folderKey`, and the output key, rest on the author's single live run.

### tl;dr
Ready to merge. Only a minor finding remains: step (4) routes by `processType`, which the command it names doesn't return.

nikhil-maryala (approved):
**Re-review of `8ec40405b` — Approve.** Tally: 0 Blocker / 0 Critical / 0 Major / 0 Minor / 0 Nit.

- **My earlier Nit is fixed.** `registry-workflow.md` now says "For `Orchestrator.ExecuteApiWorkflowAsync`, drop the template's own `folderId`/`folderPath`/`name` context fields", which matches rule 18.
- **Rule 18 is narrower now, and that's the safer claim.** Every wrapper binds `releaseKey`. Only `ExecuteApiWorkflowAsync`, the one type verified live, gets the `folderKey` swap and drops the old fields. The others make that swap only after a live `key:FolderKey` fault, and otherwise report the node unverified. That matches what the evidence supports.
- **The new rule 17 step (4) routing is correct.** It picks the wrapper by `processType` and sends low-code agents to `Orchestrator.StartAgentJob` under rule 6, not rule 18. In `packages/maestro-sdk/src/manifest/bpmn-spec.json`, `StartAgentJob` serves context `[name, folderPath]` with a `process`/`name`/`name` binding, unlike the `releaseKey` family.
- **The `uip solution new` fallback on `unknown command` follows existing practice.** `uipath-agents`, `uipath-human-in-the-loop` and `uipath-maestro-flow` use the same fallback, so it's not new drift.

`git merge-tree origin/main <head>` is clean.

PR #3361, opened 2026-09-16
fix(uipath-api-workflow): a scalar Response reaches callers under the key Result, not the schema field name

DevMomo:
## Problem

`uipath-api-workflow` tells agents a scalar `Response` "is fine" without saying what a caller then receives.

`SKILL.md:83` (rule 15) and `references/task-types.md:487` both state the single-expression form is fine. True for the workflow itself — but a scalar `response` reaches **every** external caller under the fixed generic key `Result`. The `output.schema` property name the agent declared does not become the key.

Measured symptom: a Maestro BPMN `Orchestrator.ExecuteApiWorkflowAsync` node mapping `source="=result.message"` against a workflow whose schema declares `message` reads `null`. `references/operating-published-workflows.md`'s invoke section — where a caller actually reads the result back — said nothing about it.

## Fix

Three edits, 4 added lines total:

- `SKILL.md:83` — "is fine" becomes "is fine **only when nothing external needs the value under a named key**", with the `Result` key named and the object form (`"${{ message: … }}"`) prescribed whenever a caller reads a named field.
- `references/task-types.md:487` — same caveat, scoped so the existing StudioWeb-roundtrip claim stays intact (the scalar form *is* still the roundtrip-safe one).
- `references/operating-published-workflows.md` — blockquote in the invoke section pointing at the rule, since that is where a caller looks.

## Verified

`uip api-workflow run --no-auth`, same value, two response shapes:

| `response` | Caller sees |
|---|---|
| `"${{ message: … }}"` (object literal) | `Data.Message` |
| `"${…}"` (scalar) | `Data.Result` |

Confirmed again through an Orchestrator job: the element output was `{"Result": …}` while the BPMN node read `=result.message` and got `null`.

Nothing on `main` covers this — `skills/uipath-api-workflow/` has had no commits since this branch was cut, and none of #2366/#2387/#3155/#3113 touches this skill.

- [ ] Reviewed by a `uipath-api-workflow` codeowner


DevMomo:
Dropping this from the stack: uipath-api-workflow is owned by another team and we can't modify it here. The failure mode this PR addressed (a scalar Response reaching callers under the generic Result key instead of the declared output-schema property name) is now covered instead by a runtime-verification rule added to the uipath-maestro-bpmn skill in #3360 — after deploying any invoked resource, run it once and read the actual output key from the real response rather than assuming the declared schema property name. Branch left intact, not deleting it.

PR #3354, opened 2026-09-16
feat(uipath-maestro-flow): trigger on Maestro Automate

rockymadden:
## What

Adds **Maestro Automate** as a trigger name for `uipath-maestro-flow`, so the skill fires on that phrasing as well as `flow` / `Maestro Flow`.

- `skills/uipath-maestro-flow/SKILL.md` — one insert in the `TRIGGER for …` clause. 695 → 714 chars, under the 1024 cap.
- `tests/tasks/activation/uipath-maestro-flow.jsonl` — 4 positives (`-058` … `-061`) phrased with "Maestro Automate", so the activation gate actually tests the new trigger instead of rubber-stamping it.

## Why only `skills/`

The other two flow `SKILL.md` copies are deliberately untouched:

- `preview/skills/uipath-maestro-flow/SKILL.md` is auto-synced from the SDK repo (`scripts/sync-maestro-sdk-preview.mjs`) — a hand edit gets overwritten.
- `skill-flavors/studioweb/uipath-maestro-flow/SKILL.md` is a fragment overlay with no frontmatter; it inherits the canonical description.

## Backport

Destined for `release/v1.202`. The cherry-pick was simulated before opening this PR (`git merge-tree --merge-base`, exit 0): the description sits at line 3 with byte-identical context on both branches, the activation JSONL is identical across branches, and the resulting tree keeps v1.202's own divergent rule #9 wording. After merge: `/cherrypick release/v1.202`.

## Risk

Cross-fire is **not** covered by CI — the activation gate runs positives only. `negative.jsonl:18` is "Create a connector in Power Automate". The two-word "Maestro Automate" should not collide, but that is unverified until the next manual full activation run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3362, opened 2026-09-16
feat(uipath-maestro-flow): create Maestro Automate projects with --automate (#3357)

rockymadden:
Backport of #3357 to the 1.202 line. Clean cherry-pick, no conflicts.

#3354 taught the skill to *trigger* on "Maestro Automate" but not what to *do* about it, so the agent routed correctly and then built a plain Flow. This pairs with the CLI backport (UiPath/cli#4287) that adds the flag.

## Verified on this line

- `pytest tests/tasks/uipath-maestro-flow/` — 1173 passed
- both flavors compose: 27 skills, 182 replacements
- links 957/0 broken
- **zero `--automate` in the studioweb tree** — that flavor creates projects through `CreateProjects` and never runs `flow init`, so every edit sits inside an existing flavor block rather than a new one

## Ordering

`#3354` is already on this line as `#3355`, so this is the only skills commit needed. It should land after UiPath/cli#4287, since the smoke task drives `--automate` through the real binary.


tests/tasks/uipath-maestro-flow/smoke/init_maestro_automate.yaml
```diff
@@ -0,0 +1,62 @@
+task_id: skill-flow-init-maestro-automate
+description: >
+  Skill-guided evaluation: asked for a Maestro Automate project, the agent
+  passes --automate to `uip maestro flow init` rather than creating a plain
+  Flow. The words in the request are the only signal; nothing else about the
+  build changes.
+tags: [uipath-maestro-flow, smoke, mode:build, lifecycle:generate]
```
nikhil-maryala, line 7: [NIT] Same as on #3357: these two new smoke tasks tag `[uipath-maestro-flow, smoke, lifecycle:generate]` with no `mode:*` tag, which `.claude/rules/test-writing.md` lists as required (`mode:build` fits). Not blocking — matches two existing siblings (`registry_discovery.yaml`, `init_validate.yaml`) that also omit it. Applies to `init_plain_flow_default.yaml:8` too.

nikhil-maryala (approved):

**Severity tally — 0 Blocker · 0 Critical · 0 Major · 0 Minor · 1 Nit.**

Faithful backport of #3357 to the 1.202 line. I diffed both PRs: content is **identical** (only git blob-hash lines differ, as expected for a cherry-pick onto a different base). #3357 I already reviewed and verified in depth — including building both flavors and confirming **0 `--automate` in the studioweb tree** vs 7 occurrences in default — so that verification carries over unchanged. The body’s own checks agree (1173 pytest passed, both flavors compose, links clean). Correctly ordered to land after the CLI backport (#4287) so the smoke task drives `--automate` through the real binary. The single carried-over NIT is below. Approving.

PR #3332, opened 2026-09-15
chore(preview): retire the Maestro SDK preview sync; the skills are canonical here

dmetzgar:
Until now, `preview/skills/uipath-maestro-{flow,case,bpmn}` were snapshots of `UiPath/flow-builder-sdk` `typescript/sdk/skill/`, refreshed nightly by a three-way merge.  Last nightly merge was #3329. 

We want the preview skills to become the source of truth. Part of that means shutting off the nightly merge action. This PR removes the following files:
- `.github/workflows/sync-maestro-sdk-preview.yml` - workflow
- `scripts/sync-maestro-sdk-preview.mjs` - sync script
- `tests/scripts/sync-maestro-sdk-preview.test.mjs`

The preview skill files have comment updates so they're not pointing to flow-builder-sdk as the source of truth. There are other follow-up changes after this. Check [flow-builder-sdk/docs/SKILLS_PROMOTION_PLAN.md](https://github.com/UiPath/flow-builder-sdk/blob/main/docs/SKILLS_PROMOTION_PLAN.md) for the full plan.

PR #3364, opened 2026-09-16
chore(preview): port the API-index rework into the Flow skill by hand

dmetzgar:
Ports skill changes over from flow-builder-sdk since the sync is now removed (#3332). Includes:
- [flow-builder-sdk/747](UiPath/flow-builder-sdk/pull/747)
- [flow-builder-sdk/750](UiPath/flow-builder-sdk/pull/750)
- [flow-builder-sdk/751](UiPath/flow-builder-sdk/pull/751)
- [flow-builder-sdk/752](UiPath/flow-builder-sdk/pull/752)

## What was ported

- `references/api.md` deleted
- 26 references updated
- `SKILL.md` updated (Flow)

## Scope

Case and BPMN are not included. Flow will be the first to go GA.

## Coding agent Verification

- No link in this skill points at a file it doesn't carry (checked every `*.md` target in `SKILL.md` and all references against both the skill root and `references/`).
- Every router-table `examples/*.flow.ts` resolves.
- `skills:check-links` — **zero broken links under `preview/`**. It reports 4 overall; all are vendored third-party READMEs under `uipath-maestro-bpmn/validator/node_modules/`, and I confirmed the same 4 on main by stashing this change.
- `check-skill-verbs.py` — 11 findings, **0 blocking**.
- `skills:validate` — clean on both default and studioweb.

This is part of phase 2 from https://github.com/UiPath/flow-builder-sdk/blob/main/docs/SKILLS_PROMOTION_PLAN.md


bai-uipath (approved):
Fix what you agree with, otherwise lgtm.

**Port checks out.** Reference files are byte-identical to upstream, the deleted corpus genuinely ships in the package (`dist/api-index.md` and `dist/api-members.md` are both in the published 6.4.0 tarball), and no dangling `api.md` links remain in the skill.

**The base HITL example is stale and now emits an unroutable outcome.** `examples/GallerySubmission.flow.ts` is untouched here, but it is what the router table points at for Human task, and its header comment still states the single-`completed`-exit model as fact. Under 6.4.0 `flow-check` warns `HITL_OUTCOME_UNROUTED`: `outcome-decline` has no route, so the run stalls if the reviewer picks it. That is exactly the failure the new hitl.md spends its length warning about, shipped as the canonical example. Fix: rewrite to `.stepSwitch` with both arms and correct the comment. The quick-form and action-app examples are fine, variants keep the single exit.

**Minor, not blockers:** the SKILL.md `.stepSwitch` snippet warns `NO_RETURN` as written, its Approve arm converging into a default End that never fills the declared output, where hitl.md's version gets it right with a var plus one return. And "each file's header spells the full path its rows resolve against" holds only for `api-index.md`, which is upstream's generator rather than this port.

dmetzgar:
All three findings verified and fixed in 5a5f865. Thanks — the first one was a real defect this PR would have shipped.

**The HITL example: confirmed, and caused by this PR.** Bringing in the per-outcome `hitl.md` is what made `examples/GallerySubmission.flow.ts` wrong, and the router's Human task row points straight at it. `flow check`:

```
⚠ [HITL_OUTCOME_UNROUTED] Human task "curate" routes per outcome, but
  outcome-decline has no route. The run stalls if the reviewer picks it.
```

Rewritten to `.stepSwitch` with an arm per outcome, both arms assigning shared vars via `{ updates }`, one `.return()` reading them — the shape `hitl.md` teaches. Header comment rewritten; it had stated the single-exit model as fact. Checks clean.

A note on how nearly I missed this: my first run reported `✓ no issues`, because my local `dist/` predated the HITL change. Rebuilding produced the warning. I'd have told you the finding was wrong.

**The `.stepSwitch` snippet: confirmed.** Completed as a reader would, with a declared output, it warns `NO_RETURN` — the Approve arm converges into a default End that never fills it. Now the same var-plus-one-return shape as `hitl.md`. Checks clean.

**The header sentence: confirmed, and broader than you put it.** *Neither* index states a path root — `dist/api-index.md` and `dist/api-members.md` carry an identical "GENERATED from the TypeScript declarations" block and nothing about paths. The root is already given two sentences earlier, so the sentence now describes what the headers actually say.

The last two are upstream text this PR ported faithfully, so upstream carries the same two defects. Fixed here because `preview/` is canonical as of #3332 — worth someone fixing them in `flow-builder-sdk`'s copy too, or letting them die with that tree when the website forks off it.

**Not fixed:** `LighthouseSignal` and `MeteorShowerPages` warn `HTTP_V1_RETIRED`. Both exist to demonstrate `managed: false`, which is what the router's Standalone HTTP row points at, so the warning is the point. Pre-existing and untouched by this PR.

I ran `flow check` over all 32 examples in the skill; those two are the only remaining warnings.

PR #3365, opened 2026-09-16
feat(uipath-maestro-case): convert the SDD first, then repair the remainder

roberts-cliff:
## What this is

The Case skill now builds by **converting the SDD first** and repairing the remainder, instead of authoring `caseplan.json` element by element. Measured across four full 65-task suite runs plus a golden loop, with the CLI changes it depends on developed alongside in `uipath/cli`.

**Result on the 65-task corpus:** 61/65 passing, mean 0.959, ~$27 per run on gpt-5.6-terra.

| Run | Skills | CLI | Passing | Mean |
|---|---|---|---|---|
| suite6 | `7629a821a` | `54aa3ae7a` | 53/65 | 0.903 |
| suite7 | `ed3a43391` | `5dbe04910` | 61/65 | 0.971 |
| suite8 | `e6f54ad3c` | `bb127867f` | 61/65 | 0.959 |

The golden rebuild (`cm_golden_expense`) passes 9/9 criteria on terra at ~$0.81, and previously passed 5/5 on Sonnet. Terra's golden record across the loop is 11 of 12.

## The change

1. **Convert-first (Rule 4).** Phase 2 opens with `uip maestro case sdd convert`; `Data.Unresolved[]` is the work list. Adopted by 35 of 35 greenfield builds in suite6.
2. **Scaffold with `uip maestro case init` (Rule 24).** The former ban is gone. `implementation.md` Step 6 had kept forbidding the command after the rule changed, and behaviour split accordingly — 25 of 40 builds followed the old hand-written scaffold. Six files now agree, using wording the `studioweb` flavor had already proven.
3. **Registry search unbanned (Rule 3).** The ban outlived the CLI defect behind it.
4. **`=js:` identifiers must be namespace-qualified.** Nothing said so; `=js:priority === 'Urgent'` read as acceptable and silently never fires.
5. **`--strict --sdd` is what completion means.** Extracted to the head of Rule 7 — a default-profile `Valid` is not a finished build.
6. **`jsonSchema` variables require `body`/`_jsonSchema`**, with the convert gap named as an explicit exception to Rule 4's settled-shape clause.
7. **Artifact-I/O ban narrowed to the write path.** Reads were banned, the ban was ignored in 5 of 65 runs with no score cost, and an unenforced rule beside enforced ones teaches that the list is soft.

## Reviewer notes

**Three doc claims were wrong and are corrected here.** The connection binding `name` template, the ConnectionId `default`, and the empty `jsonSchema` default each described *what the CLI emits* as *what the platform requires*. All three were caught by reading the consumer (`case-bindings-service.ts`, `PackagingUtil.ts`, the platform pattern definitions) rather than by re-running a check. Worth knowing when reading any "source of truth" citation in these references: it is authoritative for output, not for requirements.

**Four test-instrument defects are fixed.** `check_connection_parity.py` required a `default` the consumer treats as optional and failed a correct build; `check_linear_three_stages.py` asserted an id *shape* where the contract is that the id *resolves*; `check_registry_handoff.py` read cache-entry keys in one casing while runs alternate between two; and `athena_cm_event.yaml` gated on `--strict --sdd` while its own prompt instructs the placeholder fallback, so it failed the agent for obeying it.

**Known open, not addressed here:**
- `e2e-outcome-external` errors at its pre-run gate — the Outlook sandbox connection needs reauthenticating. The gate is working as designed: it fails for $0 rather than scoring a tenant outage as a skill regression.
- `multi-task-dependency-chain` intermittently times out on `uip maestro case debug` polling (540s). Engine-side; absent from suite7, present in suite6 and suite8.
- `STRICT_SDD_*` treats an SDD that *names* a resource as one that *resolved* it. It cannot see tenant state, so a correct placeholder reads as an error. The fix is to read `registry-resolved.json` beside the plan; tracked for a follow-up CLI branch.

## Validation

```
npm run skills:validate                  # OK - default + studioweb
python3 scripts/check-skill-status.py    # OK - 27 skills
python3 scripts/check-skills-sh.py       # OK - 27 skills, 4 sections
```

No new `$TASK_DIR` / `$SKILLS_REPO_PATH` host-path references. All canonical edits sit outside `skill-flavor` blocks or converge on wording the flavor already carried, so `studioweb` is unchanged at 264 replacements.


roberts-cliff:
Addressed both review findings, plus two more the sweep turned up.

**Both flagged issues — fixed in `902e45192`:**

- `triggers/event/impl-json.md:150` — stale `Rule 6` → `Rule 7`. Worth noting for the record: this line was already off by one *on main* (it read `Rule 5` where regenerate-from-scratch was Rule 6), so the renumbering sweep preserved a pre-existing error rather than introducing one. Line 111's two references in the same file traced clean — old Rule 6 → 7 and old Rule 7 → 8 are both correct.
- `validator-codes.md` broken sentence — joined. That was an artifact of extracting the catalog out of `implementation.md`.
- `validator-codes.md` duplicate headers — taken rather than deferred. `### Check 1`, `### Check 9` and `### Check 11` are now suffixed by subject (`— input bindings` / `— SDD variable parity`, `— condition rows` / `— task configuration`, `— container placement` / `— binding keys`). Nothing links `#check-N` today, so no link was broken; this is to stop one being written against an ambiguous anchor.

**Two more found by sweeping rather than spot-checking:**

Since the one real miss was a reference to a *wrong existing* rule — which a "does this rule number exist" check cannot catch — I checked all 216 `Rule N` references in the skill against the 26 rule titles.

- `case-editing-operations.md:16` read `Rule 19/19`. It should be `Rule 19/20` (Layout / Global output IDs, old 18/19) — the sweep bumped the first number and not the second.
- No other reference points at a nonexistent rule, and no other `N/M` pair is malformed.

**On the three items flagged for manual review** — all three are verifiable and I have checked two:

- **`Unresolved` kinds.** The count in the PR description is stale, not wrong in kind: the union in `sdd-convert/types.ts` was 19 kinds when I first read it and has since grown — `variable-category` and now `variable-body` were added during this loop. Rule 4 names three kinds and their closers as *the ones with named closers*, not as the complete union, which is the claim that matters for an agent. The complete list is whatever `Data.Unresolved[]` returns at run time, and the rule says to close every entry it reports.
- **`case init` with `&&`-chaining.** Verified behaviourally, not assumed. `uip solution init X` creates `X/` as a child of cwd, so `solution init X && uip maestro case init X` runs `case init` from the parent — outside the new solution — and it auto-scaffolds `XSolution/`. That is why Rule 24 requires `cd <SolutionDir>` first and says `&&`-chaining does not satisfy it.
- **Condition-name exemption table.** Not independently verified by me; it predates this PR and I have not re-derived it against the frontend's default-name logic. Treating it as out of scope here rather than claiming it checks out.

Gates re-run after the fixes: `skills:validate` OK (default + studioweb, 264 replacements unchanged), `check-skill-status.py` OK, `check-skills-sh.py` OK.

roberts-cliff:
@claude review this PR against its current diff (rebased on main at a9bca87db). The last bot review was on 09-16, about 40 commits ago, and the description has been rewritten around suite28.

Yiqi517 (approved):
Amazing!!

roberts-cliff:
@claude the previous review run was cancelled by the approval events before it posted (it had reached 4 Medium / 3 Low in the reference docs). Please re-run the review on the current diff and post the findings.

roberts-cliff:
@claude the full review hits the 15-minute job timeout on this 96-file diff. Please review only `skills/uipath-maestro-case/references/` and `skill-flavors/studioweb/uipath-maestro-case/` against main, and post findings with file:line and severity as soon as you have them. Skip SKILL.md and tests; an earlier run found no issues there.

roberts-cliff:
Remote evals started on this branch (056458226) for the full case suite (65 tasks), against `@uipath/case-tool@1.204.0-dev.8836`, built from the #4238 merge `f38fc4bfe`:

- codex / gpt-5.6-terra: https://github.com/UiPath/skills/actions/runs/36031211807
- claude / claude-sonnet-5: https://github.com/UiPath/skills/actions/runs/36031215480

I'll add the results to the description when they finish.

StephenHanzlik (approved):

PR #3367, opened 2026-09-16
fix(tests): grade the invoice e2e project wherever init placed it

nikhil-maryala:
## Problem

`skill-bpmn-e2e-invoice-exception-triage` failed the 2026-09-16 claude nightly (`2026-09-16_04-17-43`) at **0.19** on pure path mismatch:

```
File 'InvoiceExceptionTriage/InvoiceExceptionTriage.bpmn' does not exist
ERROR: InvoiceExceptionTriage.bpmn is missing
```

The agent built a **Valid** project (`validate` clean, `refresh` consistent, all five package files present) — at `InvoiceExceptionTriageSolution/InvoiceExceptionTriage/`. `uip maestro bpmn init` nests the project under a `<Name>Solution/` wrapper unless the agent passes `--skip-solution-registration`, and the skill makes that flag conditional ("source-only draft the user has not asked to package"). This e2e prompt asks for full package metadata, so skipping the flag is a defensible reading: yesterday's claude run passed the flag and scored 1.0; today's omitted it and scored 0.19 with functionally equivalent output. Same run, codex passed by skipping the wrapper. Grading a layout the prompt never states is grading a coin flip.

## Fix

Adopt the layout-tolerant idiom the sibling e2e tasks already use (`multi_node/customer_escalation` — #3216, `wiki_pageviews`, `live_debug_e2e`):

- **yaml**: the three `file_exists` criteria become rglob `run_command`s — find `InvoiceExceptionTriage.bpmn` anywhere under the sandbox, require `project.uiproj` / `operate.json` beside it.
- **check script**: `resolve_project()` requires **exactly one** `InvoiceExceptionTriage.bpmn` with `project.uiproj` beside it (ambiguous duplicates still fail loudly), then grades everything relative to that directory. Structural assertions unchanged.

Deliberately NOT touched: `e2e/customer_escalation_triage` pins `CustomerEscalationTriageSolution/CustomerEscalationTriage/` because its **prompt states that layout** — there the pinned path grades a stated requirement.

## Verification

Local grader matrix (script + all three inline criteria):

| Layout | Expected | Result |
|---|---|---|
| `InvoiceExceptionTriage/` at root | pass | ✅ exit 0 |
| `InvoiceExceptionTriageSolution/InvoiceExceptionTriage/` | pass | ✅ exit 0 |
| Nothing built | fail | ✅ exit 1 |
| Two project copies | fail | ✅ exit 1 (ambiguity named) |

GH `run-coder-eval.yml` dispatched against this branch (agent=claude, task `invoice_exception_triage`) — link to follow in a comment.


nikhil-maryala:
Verification run (this branch, agent=claude, task `invoice_exception_triage`): https://github.com/UiPath/skills/actions/runs/35162110787 — will post the result when it completes.

nikhil-maryala:
Pushed e2ff77ce7: the resolver now lives in `_shared/bpmn_check.py` as `resolve_project(bpmn_name)` (per review direction to reuse the shared helpers) — `check_invoice_exception_triage.py` and `check_customer_escalation.py` import it instead of carrying private copies. `find_bpmn_file` was deliberately not reused: its hint match silently returns the alphabetically-first duplicate, while `resolve_project` requires exactly one match with `project.uiproj` beside it. The invoice task's `reference.directory` widens to `../..` so `_shared/` exists in the sandbox (same shape as the edit tasks after #3271).

Cancelled the stale verification run; fresh run on this commit covering BOTH affected tasks (invoice + customer_escalation, agent=claude): https://github.com/UiPath/skills/actions/runs/35162613752

nikhil-maryala:
Verify run [35162613752](https://github.com/UiPath/skills/actions/runs/35162613752) on e2ff77ce7: **customer_escalation SUCCESS 1.0** through the shared resolver, and the invoice task's three rglob criteria all passed — but the structural check itself couldn't run: this yaml sits directly in `e2e/` (no per-task subdir), so `reference.directory: ../..` climbed out of the `uipath-maestro-bpmn` tree and `$REFERENCE_DIR/e2e/check_invoice_exception_triage.py` didn't exist in the sandbox (exit 2, file not found).

Fixed in 753c66edf (`directory: ..`). Fresh invoice-only verify: https://github.com/UiPath/skills/actions/runs/35163436531

nikhil-maryala:
Re the advisory lint's Medium (validate-only e2e without a `debug` execution check): pre-existing task characteristic, documented in the task description ("without cloud-side effects") and shared with the sibling `multi_node/customer_escalation` authoring-only e2e — the lint itself concludes no change needed. This PR only changes WHERE the graders look for the project, not what the task grades, so re-tiering is out of scope here.

nikhil-maryala:
✅ Verify run [35163436531](https://github.com/UiPath/skills/actions/runs/35163436531) on 753c66edf: **skill-bpmn-e2e-invoice-exception-triage SUCCESS 1.0** (37 turns) — all five criteria pass, including the structural check now imported from `_shared/bpmn_check.py::resolve_project`. Combined with the earlier run's **customer_escalation SUCCESS 1.0** (unchanged since e2ff77ce7), both graders touched by this PR are verified end-to-end on the claude harness. All CI checks green — ready for review.

tests/tasks/uipath-maestro-bpmn/e2e/invoice_exception_triage.yaml
```diff
@@ -41,27 +44,37 @@ initial_prompt: |
...
+    expected_exit_code: 0
     weight: 1.5
     pass_threshold: 1.0
 
-  - type: file_exists
-    description: "Project metadata file was created"
-    path: "InvoiceExceptionTriage/project.uiproj"
+  - type: run_command
+    description: "Project metadata file was created beside the BPMN source"
+    command: "python3 -c \"from pathlib import Path; import sys; b=next(Path('.').rglob('InvoiceExceptionTriage.bpmn'), None); sys.exit(0 if b and (b.parent / 'project.uiproj').is_file() else 'project.uiproj missing beside InvoiceExceptionTriage.bpmn')\""
```
rockymadden, line 61: 🟡 **minor** — `next(...)` takes an arbitrary first match, contradicting the exactly-one rule the 5.0-weight script two criteria below enforces.

`rglob` yields in `scandir` order, not sorted. If a run leaves a second `InvoiceExceptionTriage.bpmn` **without** `project.uiproj` beside it (scratch copy, abandoned init retry), this criterion and its twin on line 69 pass or fail on directory order — 2.0 of 10.5 weight decided by a coin flip, the exact failure mode this PR exists to remove.

Predicate over the matches instead of picking one:

```yaml
command: "python3 -c \"from pathlib import Path; import sys; sys.exit(0 if any((p.parent / 'project.uiproj').is_file() for p in Path('.').rglob('InvoiceExceptionTriage.bpmn')) else 'project.uiproj missing beside InvoiceExceptionTriage.bpmn')\""
```

Same shape for line 69 with `operate.json`. Monotone, order-independent, and the partial-credit tiering is unchanged.

Inherited verbatim from `multi_node/customer_escalation` and `wiki_pageviews`, so fixing it here means touching those two as well or accepting the drift. Your call on scope — flagging it because copying the idiom also copied its one weak spot.

tests/tasks/uipath-maestro-bpmn/_shared/bpmn_check.py
```diff
@@ -34,6 +35,27 @@ def find_bpmn_file(name_hint: str | None = None) -> str:
...
+    a ``<Name>Solution/`` wrapper -- ``uip maestro bpmn init`` creates the
+    wrapper unless --skip-solution-registration is passed), but picks the real
+    project unambiguously: exactly one ``bpmn_name`` with project.uiproj beside
+    it, so a stray draft copy is never graded (``find_bpmn_file`` would
+    silently return the alphabetically-first match).
+    """
+    candidates = [
+        p for p in Path.cwd().rglob(bpmn_name) if (p.parent / "project.uiproj").is_file()
+    ]
+    if len(candidates) != 1:
```
rockymadden, line 51: 🟡 **minor** — merging "none found" with "two found" hides the most common single-file failure.

An agent that authors the BPMN but forgets `project.uiproj` now gets:

```
FAIL: expected exactly one InvoiceExceptionTriage.bpmn with project.uiproj beside it, found []
```

which reads as *nothing was built*. The old grader said `ERROR: project.uiproj is missing`. Precise diagnostics are the whole reason this script exists for nightly triage.

Split the cases:

```python
found = list(Path.cwd().rglob(bpmn_name))
candidates = [p for p in found if (p.parent / "project.uiproj").is_file()]
if not candidates:
    fail(f"project.uiproj missing beside {bpmn_name}" if found else f"no {bpmn_name} found")
if len(candidates) > 1:
    fail(f"ambiguous {bpmn_name}: {[str(p) for p in candidates]}")
```

Two knock-ons worth a line each:

- With the resolver guaranteeing `project.uiproj`, its entry in `check_invoice_exception_triage.py:45`'s `required_files` can never fail. Drop it, or the list implies coverage it no longer has. (Same dead entry in `check_customer_escalation.py`'s `REQUIRED_FILES` — pre-existing.)
- `_shared`'s `fail()` prefixes `FAIL:` while the invoice script's local `fail()` prefixes `ERROR:`, so this one grader now emits both. Minor, but it makes log greps inconsistent.

rockymadden (approved):
**Recommendation: 🟢 Approved** — the premise holds, the fix matches how the rest of this suite grades, and both graders it touches are verified 1.0 on the harness. Findings below are all minor; none block merge.

---

### 📦 What this does

- Invoice e2e's three `file_exists` criteria pinned `InvoiceExceptionTriage/…`. `uip maestro bpmn init` nests under `<Name>Solution/` unless `--skip-solution-registration` is passed, so a functionally identical project scored 0.19.
- Criteria become rglob `run_command`s; the 5.0-weight structural check resolves the project through a new `resolve_project(bpmn_name)` in `_shared/bpmn_check.py` (exactly one `<Name>.bpmn` with `project.uiproj` beside it).
- `multi_node/customer_escalation`'s private copy of that resolver is deleted in favour of the shared one.
- `reference.directory` widens `.` → `..` so `_shared/` lands in the mirror; the check command gains the `e2e/` segment.

### 🔍 Overall findings

**1. 🟡 minor — nothing static catches the `$REFERENCE_DIR` mistake that cost verify run 35162613752.**

The `../..` escape failed with exit 2, file-not-found, after a 9-minute eval run. A cheap guard would have caught it at PR time, and the guard that looks like it should *silently skips* the case:

- `_shared/test_criterion_budgets.py:78` — `return resolved if os.path.exists(resolved) else None`
- maestro-flow's `_shared/test_same_ground_corpus.py:109` — `if path.is_file()`
- `test_every_eligible_criterion_and_task_is_enforced` counts through that same helper, so the count assertion can't see the gap either.

~80 bpmn task yamls can regress the same way. A ~10-line pytest in `_shared/test_criterion_budgets.py` closes it: for every `run_command` naming `$REFERENCE_DIR/<x>.py`, assert it resolves under the yaml's `reference.directory`, and assert `$REFERENCE_DIR` is never used with no `reference:` block. Follow-up PR is fine — out of scope here.

**2. 🟡 minor — `resolve_project` ships with no unit test.**

It now gates two 5.0-weight criteria, and its only verification is the four-layout matrix in this description. `_shared/test_bpmn_check.py` already does `from bpmn_check import …`, so making that matrix durable is a few lines:

```python
def test_resolve_project_accepts_either_layout(tmp_path, monkeypatch):
    for layout in ("Proj", "ProjSolution/Proj"):
        # write Proj.bpmn + project.uiproj, chdir, assert the resolved dir
```
plus the two failing cases (nothing built, two copies). Without it the next edit to the resolver is graded by a nightly, not by CI.

**3. 🟡 minor — the description is stale against the last two commits.**

It describes only the invoice check script. It never mentions the resolver moving to `_shared/bpmn_check.py`, `multi_node/customer_escalation/check_customer_escalation.py` being rewired, or the `reference.directory` widening — the change that actually broke the first verify run. It also still reads "link to follow in a comment". With `e2e/customer_escalation_triage` called out as "deliberately NOT touched" a few lines above, a reader can easily miss that `multi_node/customer_escalation` **is** touched. Fold the comment thread's facts into the body before merge.

### ✅ Confirmed

- **Premise.** `SKILL.md:184` and `references/shared/project-layout.md:18-49` make `--skip-solution-registration` conditional on "a source-only draft the user has not asked to package or operate". This prompt asks for full package metadata, so the wrapper layout is the skill-compliant reading. Grading a layout the prompt never states was a coin flip.
- **`reference: ..` matches convention**, it isn't a loosening: 60+ yamls in this suite already resolve the reference at the suite root (`..` at group level, `../..` in a per-task subdir). `.` was the outlier.
- **No leak from the wider mirror.** coder_eval copies the reference, mounts it at `/work/references`, and chmods it 000 for every agent turn; the task-dir mount is a shielded copy. Reference content is never staged into the agent's sandbox.
- **Both paths resolve**: `$REFERENCE_DIR/e2e/check_invoice_exception_triage.py` and `$REFERENCE_DIR/_shared/bpmn_check.py`. The `_shared` up-walk stops at the mirror root, which carries `__init__.py`.
- **Leaving `e2e/customer_escalation_triage` pinned is right** — its prompt does state `CustomerEscalationTriageSolution/CustomerEscalationTriage/`, so the pinned path grades a stated requirement.
- **The verification chain closes.** 753c66edf touches only the invoice yaml, so customer_escalation's SUCCESS 1.0 on e2ff77ce7 still stands. Invoice SUCCESS 1.0 on HEAD. CI green.
- No stray `InvoiceExceptionTriage.bpmn` or `project.uiproj` fixture exists in the staged skill or the task tree, so the rglob can't collide with one.

### ⚠️ Reasoned, not verified

- **Codex nightly rows are unaffected** — they build bare, which is the layout the old grader already required, so they can't regress. Both verify runs were agent=claude only.
- **No realistic run leaves a second `<Name>.bpmn` + `project.uiproj` pair**, so the exactly-one rule shouldn't false-negative. `pack <project-path> <OutputDir>` writes a package, not an extracted copy, and the authoring-only sibling scores 1.0 through the same rule. Nothing pins it — that is finding 2.

### tl;dr

Right fix for a real coin flip, done the way the rest of the suite does it, verified end-to-end on both affected graders. Three minor items: the two inline criteria still pick an arbitrary first match while the script demands exactly one, the merged ambiguity message now hides the common "forgot project.uiproj" case, and nothing in CI would have caught the `$REFERENCE_DIR` typo that burned a verify run. Ship it and fold the comment thread back into the description.

PR #3373, opened 2026-09-17
docs: clear the remaining content-scan findings without dropping the links [PILOT-7612]

RaduAna-Maria:
## Issue

[PILOT-7612](https://uipath.atlassian.net/browse/PILOT-7612)

## Summary

- Clear the **six remaining content-scan findings** across `uipath-ixp`, `uipath-platform`, `uipath-rpa`, and `uipath-troubleshoot` — 8 lines, 6 files
- All six are false positives on example text and product identifiers. No secret, live PII, or cardholder data exists in the repo; the gate matches on **shape** alone and blocks the whole published package either way
- Three of them are the findings [#3201](https://github.com/UiPath/skills/pull/3201) deliberately deferred pending a scanner-side allowlist. The allowlist never landed and the scan fired again — each can drop its shape **without losing the link**, so they're fixed here instead of deferred a second time

### Deferred by #3201, now fixed without losing the reader's link

- **Forum thread IDs** (`ThirdParty-SharePoint.md`, `foreground-unattended-robot.md`) — switched to Discourse's short `/t/<id>` form. The topic ID is preserved verbatim; what's gone is the long slug that put a 10-digit run (`...-user-1230/718082`) directly in front of it. Both URLs verified **HTTP 200**.
- **Databricks `docsUrl`** (`vendor-docs-registry.json`) — repointed to the canonical page the original already redirected to: `https://docs.databricks.com/api/model-serving-query/v1/query`. Same destination, resolves **200 with no redirect hop**, and the hyphens in `model-serving-query` cut the longest base64-alphabet run to **14** characters.

> **An earlier revision of this PR simply dropped `/query`. That was wrong** — `/api/workspace/servingendpoints` and `/api/workspace/servingendpoints/query` redirect to *different* pages (serving-endpoint **management** vs **query**), and this entry's own `notes` describe the `invocations` POST, i.e. the query API. Truncating the URL pointed agents at CRUD operations. Fixed in `dc5ce67`.

> **Why the original tripped an AWS-secret rule.** The path segment `com/api/workspace/servingendpoints/query` was **exactly 40** `[A-Za-z0-9/+=]` characters — the AWS secret-key length. The rule matches a *maximal* run, not a sliding window: `.` is outside the base64 alphabet, so `docs.databricks.com` splits the URL and that final piece is a complete run bounded by `.` and `"`. Two measurements over the whole `release/v1.202` package pin it:
>
> | Scope | Exactly-40 runs | Flagged |
> |---|---|---|
> | `.md` files | 63 | none |
> | `.json` files | 1 — this line | the report's only AWS finding |
>
> Every structured file with a longer or shorter run was ignored — `bpmn-spec.json` (75), this same file's line 38 (53), `metadata.json` (37), `activities.json` (32). So the length is **exactly** 40, not a minimum; the rule is scoped to structured data files, not prose; and it applies **no entropy or charset-diversity check** — it fired on an all-lowercase path with no digits. No secret exists in the file: zero strings of 20+ characters carrying both digits and mixed case.

### Example values → placeholders

- **`uipath-ixp/references/cli-reference.md`** — the float literal illustrating long-tail rounding was a 15-digit run. Now `` `<SCORE>` `` described as "printed to 15 decimal places", per `.claude/rules/content-quality.md`.
- **`UiPath.Terminal.Activities` × 2** — `3000–5000 ms` reads as an 8-digit local phone number. Written out as `between 3000 and 5000 ms`.

> That exact string occurs **nowhere else in the repo** and appears on exactly the two flagged lines — which is what pins it as the trigger rather than the `TN3270/TN5250` digits on the same lines.

## Implementation Notes

The scanner is contextual, not a plain regex: an independent sweep found hundreds of unflagged shape matches elsewhere (GUIDs, epoch timestamps, a 40-hex SHA-1), so five of the six triggers were identified by finding what is **unique** to the flagged line.

> **One finding is not mechanically explained.** `ThirdParty-SharePoint.md:81` is reported with five categories (Card Expiration Date, Credit Card, Phone Number, US SSN, US Street Address), but its only numeric token is the 6-digit forum ID `332491` — too short for any of them. Best read is a contextual false positive on the long hyphenated slug ending in digits, which this PR removes. **This one needs a re-scan to confirm**; the other five are deterministic.

## Testing

- `npm run skills:validate` — OK, default 27 skills / 1768 files; studioweb 27 / 1768 / 264 replacements
- `npm run skills:check-links` — 6838 relative links, all resolve
- `python3 scripts/check-skill-status.py` — OK, 27 skills, manifest valid
- `python3 scripts/check-skills-sh.py` — OK, 27 skills across 4 sections
- `vendor-docs-registry.json` re-parsed as valid JSON
- Re-scanned all six lines with the shape detectors that produced the findings: **0 remaining matches**
- All three shortened URLs return HTTP 200

No skill folder added, renamed, or removed, so `assets/skill-status.json` and `skills.sh.json` need no edit. No `skill-flavor` marker or flavor override touches any edited file.

## Reviewer notes

> Two Terminal files are CRLF in git. The edits were applied byte-wise to keep them that way — the diff is 1 line per file, not a whole-file rewrite.

Still unfixed, not flagged by this scan but the same class and already contrary to `.claude/rules/content-quality.md`: literal `123 Main St` / `456 Oak Ave` in `uipath-ixp/references/label-documents-guide.md` (3 lines) and `uipath-rpa/.../UiPath.Web.Activities/2.5/coded/service-discovery.md:324`. The IXP ones need **two distinct** placeholders, since the examples turn on "predicted X but actual is Y". Happy to fold in or leave as follow-up.

[PILOT-7612]: https://uipath.atlassian.net/browse/PILOT-7612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


skills/uipath-platform/references/integration-service/vendor-docs-registry.json
```diff
@@ -153,7 +153,7 @@
     },
     "databricks": {
       "connectorKey": "uipath-databricks-databricks",
-      "docsUrl": "https://docs.databricks.com/api/workspace/servingendpoints/query",
+      "docsUrl": "https://docs.databricks.com/api/workspace/servingendpoints",
```
Mihaiii: @RaduAna-Maria I tried both of them in a browser and I get different pages in a different subcategory in the menu. Please confirm we want this change.

Mihaiii (approved):

skills/uipath-platform/references/integration-service/vendor-docs-registry.json
RaduAna-Maria: Honestly, I am concerned about this change too. Claude made on its own and now I investigate this on my own. Thank you for checking

PR #3377, opened 2026-09-17
fix(tests): jint check requires elementId on process-scoped bridges, not its absence

nikhil-maryala:
## Problem

`skill-bpmn-script-jint-guidance` has failed **every harness, every day** since #2366 merged on 09-15 — claude, codex, AND antigravity, all with the identical error:

```
FAIL: expected exactly one inputOutput variable named 'amount' at process scope, found 0
```

Nightly evidence: 09-15 (claude 0.333), 09-16 (claude + codex + antigravity, 0.333 each), 09-17 (claude + codex, 0.333 each). A deterministic cross-harness red — the agents aren't failing; the check is.

## Root cause — #2366's check contradicts #3211's verified finding

The check encoded "process scope" as the **absence** of `elementId` and hard-failed any bridge variable carrying one:

```python
one_variable(..., element_id=None)          # matched only when "elementId" not in attrib
if "elementId" in variable.attrib:
    fail(f"{variable_id!r} must remain a process-scoped mutable variable")
```

That is exactly the shape **#3211** (same author) proved broken: scope is expressed *by* `elementId` — the process id for a process-scoped variable — a declaration without one "is not a variable at all" to the canvas, and every `vars.<id>` reference to it fails on import with `VARIABLE_DOES_NOT_EXIST` (verified in Studio Web; PO.Frontend `VariableValidationUtils.ts:163`; the 28-errors→0 repro is in #3211's description).

#2366 is the older branch (note the PR numbers) and merged **after** #3211 without reconciling. So since 09-15 the skill has — correctly — taught `elementId="<process id>"`, every agent on every harness authored exactly that:

```xml
<uipath:inputOutput id="Var_Amount" name="amount" type="double" elementId="Process_1" />
```

…and the check failed the correct XML. cc @DevMomo — both PRs are yours; this reconciles them in #3211's favor.

## Fix

`_shared/check_script_jint_guidance.py` only:

- `one_variable()` takes a required scope id (no more None = "no attribute" mode)
- the two internal bridge assertions pass the **process id**
- the elementId-absence guard inverts into an `elementId == <process id>` requirement, citing #3211

No skill change — the skill is right. No task change.

## Verification

Ran the fixed check against the **real 2026-09-17 nightly artifacts** (the exact files that scored 0.333 last night, pulled from blob):

| Artifact | Old check | Fixed check |
|---|---|---|
| claude `2026-09-17_04-22-38` | FAIL | **PASS** |
| codex `2026-09-17_04-19-12` | FAIL | **PASS** |
| claude artifact with `elementId` stripped from the bridge (the #3211-broken shape) | — | **FAIL** (as it must) |

GH `run-coder-eval.yml` verification on this branch (agent=claude, task `script_jint_guidance`) — link in a comment.


nikhil-maryala:
Verification run (this branch, agent=claude, task `script_jint_guidance`): https://github.com/UiPath/skills/actions/runs/35231234041 — will post the result when it completes.

nikhil-maryala:
First verify run ([35231234041](https://github.com/UiPath/skills/actions/runs/35231234041)) surfaced a second layer of the same disease: the agent authored a fully correct bridge — but named the internal variables `Amount`/`DaysOverdue`, and the check pinned the **internal display names** (`bridge from '=vars.input_Var_Amount', found 0` because the name filter excluded it). Internal names carry no runtime meaning — expressions resolve ids; only the public `amount`/`daysOverdue`/`riskScore` names are contractual.

Fixed in 640b8a54d: bridges resolve by `source`+`type`+`var`, the bridged variables are asserted **by id** (mutable inputOutput, expected type, process scope per #3211), and the optional business result accepts task or process scope. Local matrix: all three real agent artifacts pass (09-17 claude, 09-17 codex, and the capitalized-names verify-run artifact); the elementId-less mutation still fails.

Fresh verify run: https://github.com/UiPath/skills/actions/runs/35232593678

Smoke-test note: the earlier smoke failure was `skill-bpmn-parallel-fork-join` (13/14, unrelated to this check) — rerun already dispatched.

tests/tasks/uipath-maestro-bpmn/_shared/check_script_jint_guidance.py
```diff
@@ -108,26 +108,22 @@ def one_variable(
...
     kind: str,
-    element_id: str | None,
+    element_id: str,
 ) -> ET.Element:
-    scope = (
-        f"scoped to {element_id!r}"
-        if element_id is not None
-        else "at process scope"
-    )
+    # Scope is expressed by elementId for EVERY root declaration — the process
```
rockymadden: **`minor`** 🧹 This comment no longer describes any caller, and its second half is now the file's third copy of the same sentence.

After `640b8a54d`, `one_variable` is called with `start_id`, `end_id`, and `task_id` only (`:386`, `:389`, `:392`, `:395`, `:398`). Nothing passes a process id through it any more, so "the process id for a process-scoped variable" points at a path that does not exist.

The #3211 rationale now appears three times:

- here, `:115-116`
- the `fail()` at `:479-480`
- the business-result comment at `:583-584`

and the display-name rationale twice (`:180-181`, `:464-466`).

Say each once where a reader first hits it. Here, that is the mechanical fact and nothing else:

```python
) -> ET.Element:
    # elementId is the scope: the owning element's id, never absent (#3211).
```

tests/tasks/uipath-maestro-bpmn/_shared/check_script_jint_guidance.py
```diff
@@ -435,16 +435,16 @@ def main() -> None:
...
     )
     internal_days_id = bridge_target(
         start,
-        name="daysOverdue",
         source=f"=vars.{attr(public_days, 'id')}",
         output_type="integer",
     )
+    if internal_amount_id == internal_days_id:
+        fail("the amount and daysOverdue bridges must target distinct variables")
     end_output = exactly_one(
```
rockymadden, line 441: **`minor`** ⚖️ The `end_output` lookup on `:452` still pins `name == "riskScore"`, which is the same style pin `640b8a54d` just removed from `bridge_target`.

`name` on an output mapping is display metadata either way — the public contract is the `uipath:output` declaration, and that is already pinned by name at `:391`. So an agent that authors a behaviourally correct EndEvent bridge with `name="result"` and `var=<public riskScore id>` reds the nightly for a naming choice, which is exactly the failure mode the first verify run hit.

It buys no uniqueness either: `:487` already caps the EndEvent at one output mapping.

```python
        if output.attrib.get("var") == attr(public_risk, "id")
        and output.attrib.get("type") == "double"
        and re.fullmatch(r"=vars\.[\w.-]+", output.attrib.get("source", ""))
```

(Anchored here because `:452` is outside the diff hunk.)

rockymadden (approved):
**Recommendation: 🟢 Approved**

Reviewed at `640b8a54d` (the second commit landed mid-review). Both moves are right: the check was wrong, not the agents, and reconciling toward #3211 matches what the rest of the suite already asserts. Findings below, none of them in the diff's logic.

### 📦 What

`_shared/check_script_jint_guidance.py` only, two commits:

- `c0f0ce3f4` inverts the `elementId`-absence guard into an `elementId == <process id>` requirement; `one_variable()` takes a required scope id.
- `640b8a54d` stops pinning the **internal** bridge display names. Bridges resolve by `source` + `type` + non-empty `var`; the bridged variables are asserted by id for kind, type, and scope. The optional business result accepts task or process scope.

### ✅ Confirmed from the repo

- **Root cause holds.** The old `scope` string reproduces the nightly error verbatim: `expected exactly one inputOutput variable named 'amount' at process scope, found 0`.
- **Merge order holds.** #3211 `d64f609b5` on 09-10, then #2366 `4ca413b62` on 09-15. Older branch, later merge, no reconcile.
- **The suite already agreed with #3211.** `check_variable_scopes.py:65-71` hard-fails any declaration *without* an `elementId`, and `:52-58` resolves `elementId` against `{process_id} | flow element ids`, so the process id is a legal scope by the suite's own rule. #2366's check was the only place asserting the opposite. `check_simple_approval_bpmn.py:34-45` treats `element_id=None` as a wildcard, not as "must be absent", so it needs nothing.
- **Dropping the internal name pins is the right call,** and `.claude/rules/test-writing.md` backs it: grade the outcome, not an outcome-invisible attribute. The public names stay pinned where they are contractual (`:391` public `riskScore`, `:568` `scriptResponse` / `Error`).
- **The loosening did not open a uniqueness hole.** `:446` now forces the two bridge targets distinct, `:485` and `:487` cap the StartEvent at two outputs and the EndEvent at one, and `:472` still rejects a bridge that targets a public `uipath:input`.
- **"Check-only" is accurate.** This file has one consumer, `nodes/script_jint_guidance.yaml`. No fixture, no unit test to update.

### ❓ Not verified

- The nightly-artifact tables in the description and in `640b8a54d`'s message. Those files are not in the tree and I ran nothing.
- Whether the smoke red below is also red on `main`. Both runs are still in progress, so per-task logs are not retrievable yet.

### 🔴 Overall findings

**1. `major` the skill still teaches the shape this check now rejects**

`skills/uipath-maestro-bpmn/references/expression-authoring.md:97`:

> Keep a mutable root `uipath:inputOutput` for each value used by decisions, tasks, or diagnostics, and leave it root-scoped — no `elementId`.

That is exactly what the guard at `:476` rejects, and it contradicts `structural-bpmn.md:154-159` ("Without an `elementId` the declaration does not exist to the canvas").

Same disease as the one this PR is treating, one file over:

- #3211 `d64f609b5` **deleted** that bullet on 09-10
- #2369 `826927089` **re-added** it on 09-11

Older branch, merged later, not reconciled. So "No skill change, the skill is right" is true of `structural-bpmn.md` and false of `expression-authoring.md`.

Latent today, not breaking: all three harnesses authored `elementId="<process id>"`, so agents are reading `structural-bpmn.md`. It is still a live contradiction, and the next agent that follows the Scope bullet fails the check on XML the skill told it to write.

Fix, in this PR: point that bullet at the process id and link `structural-bpmn.md#variables-bpmnvariables`. Two lines, and reconciling #2366 with #3211 is this PR's stated job.

**2. `major` no unit test, on a check rewritten twice in one day**

Two reverts of the `elementId` contract, both surfaced by a red nightly days later, and now a same-day second commit to the same file. `grep -rl check_script_jint_guidance tests/` returns only the task YAML.

The pattern exists and is cheap. `author/test_simple_approval_bpmn.py` builds `<uipath:variables>` fragments with `ET.fromstring` and asserts `SystemExit`. Three cases pin what this PR just decided:

```python
def test_bridge_target_ignores_the_output_display_name() -> None:
    start = _node(
        '<bpmn:extensionElements><uipath:mapping>'
        '<uipath:output name="Amount" type="double" var="Var_Amount" '
        'source="=vars.input_amount" />'
        '</uipath:mapping></bpmn:extensionElements>'
    )
    assert bridge_target(
        start, source="=vars.input_amount", output_type="double"
    ) == "Var_Amount"


def test_one_variable_requires_the_given_scope() -> None:
    variables = _variables(
        '<uipath:input id="in_amount" name="amount" elementId="Start_1" />'
    )
    found = one_variable(
        variables, name="amount", kind="input", element_id="Start_1"
    )
    assert found.attrib["id"] == "in_amount"


def test_one_variable_rejects_an_unscoped_declaration() -> None:
    variables = _variables('<uipath:input id="in_amount" name="amount" />')
    with pytest.raises(SystemExit, match="found 0"):
        one_variable(
            variables, name="amount", kind="input", element_id="Start_1"
        )
```

`maestro-bpmn checker unit tests` already runs in CI, so wiring costs nothing. Without it, revert number three lands the same way: a red nightly nobody can attribute.

**3. `major` no verification run exists for the current head**

The only `run-coder-eval` on this branch ([35231234041](https://github.com/UiPath/skills/actions/runs/35231234041)) **failed**, on the name pin that `640b8a54d` removes. Nothing has been run against `640b8a54d` itself, so the sole end-to-end evidence for the head is a local claim about artifacts that are not in the tree.

`.claude/rules/test-writing.md` step 5 wants a passing-run claim on the PR. Kick off a fresh `run-coder-eval` on `640b8a54d` and merge on that, not on the first run's link.

**4. `minor` the description no longer matches the diff**

The body documents `c0f0ce3f4` only: `one_variable()`, the inverted guard, the three-row artifact table. It does not mention that the check now stops grading internal display names, which is a deliberate loosening of what the eval accepts. The comment thread carries it; the description is the record that survives. Fold `640b8a54d` into the Fix and Verification sections.

**5. `minor` CI is red, not from this diff**

The `c0f0ce3f4` run failed `Run skill smoke tests` on `skill-bpmn-parallel-fork-join`, pass rate 92.9% against a 95% threshold. That task grades `parallel/fork_join/check_fork_join.py`, untouched here; the 14-task bpmn matrix fires only because the diff lands in `_shared/`. Watch the `640b8a54d` run ([35232599685](https://github.com/UiPath/skills/actions/runs/35232599685)) and either re-run it or call it a known flake before merge, since it gates the PR.

### tl;dr

The check was broken, not the agents, and both commits fix it in the right direction. Merge it.

Three things before you do: `expression-authoring.md:97` still teaches the no-`elementId` shape that `:476` now rejects (#2369 re-added what #3211 removed, the same stale-branch pattern this PR exists to fix), so fold that two-line doc fix in here; add the unit test, because this file has now been silently wrong twice and corrected twice with nothing guarding it; and get a green `run-coder-eval` on `640b8a54d`, since the only run on this branch failed.

tests/tasks/uipath-maestro-bpmn/_shared/check_script_jint_guidance.py
nikhil-maryala: Fixed in b0ed434f3 — one_variable's comment is now just the mechanical fact (your wording); the #3211 rationale states once there, and the loop fail() + business-result comment lost their duplicate copies. Display-name rationale likewise states once, in bridge_target.

tests/tasks/uipath-maestro-bpmn/_shared/check_script_jint_guidance.py
nikhil-maryala: Fixed in b0ed434f3 — the end_output name pin is gone, and verify run 35232593678 proved the point one attribute further: its agent typed the end bridge `type="number"` (the public declaration's own refresh vocabulary) and failed the `type=="double"` pin. Mapping-output `type` is now unpinned too — bridges match by source/var; the type contract stays enforced on the variable declarations. Regression tests cover both.

nikhil-maryala:
Both majors + both minors addressed in b0ed434f3:

1. **expression-authoring.md contradiction (major)**: the 'root-scoped — no elementId' bullet (re-added by #2369 after #3211 deleted it) now says `elementId="<process id>"`, never absent, linking [structural-bpmn.md § Variables](../blob/main/skills/uipath-maestro-bpmn/references/structural-bpmn.md#variables). `npm run skills:validate` OK.
2. **Unit tests (major)**: `_shared/test_script_jint_check.py` — 6 tests pinning the golden 09-17 artifact shape plus every regression a run has caught: elementId-less bridge fails with the #3211 message, free internal display names, free end-bridge name, free bridge-output types, missing bridge fails. Runs in the existing `maestro-bpmn checker unit tests` job.
3. **Comment dedup (minor)** and **end_output name pin (minor)**: done — and verify run [35232593678](https://github.com/UiPath/skills/actions/runs/35232593678) independently caught the third style pin (mapping-output `type`), now also dropped; the type contract is enforced on the declarations.

Artifact matrix on the current check: **four real agent artifacts pass** (09-17 claude + codex nightlies, both verify-run artifacts — four distinct serialization styles); the elementId-less mutation still fails. Verify run 3: https://github.com/UiPath/skills/actions/runs/35233821629

nikhil-maryala:
✅ Verify run 3 ([35233821629](https://github.com/UiPath/skills/actions/runs/35233821629), agent=claude, this branch at b0ed434f3): **skill-bpmn-script-jint-guidance SUCCESS 1.0** (181s) — first fully green live run of this task since #2366 merged. Combined with the local matrix (4 real agent artifacts across 4 serialization styles pass; elementId-less mutation fails; 6/6 unit tests), the fix is verified end-to-end.

PR #3378, opened 2026-09-17
fix(tests): edit-add-node allows declaring variables for the inserted node

nikhil-maryala:
## Problem

`skill-bpmn-edit-add-node` fails recurrently on the claude nightly (09-16 and 09-17, 0.333 both days):

```
FAIL: uipath:variables payload was modified (must round-trip untouched)
```

The artifact diff against the pristine fixture shows the whole "modification":

```diff
 <uipath:variables version="v1">
         <uipath:inputOutput id="Var_OrderId" name="OrderId" type="string" elementId="Process_OrderIntake" />
         <uipath:inputOutput id="Var_Qty" name="Qty" type="number" elementId="Process_OrderIntake" />
+        <uipath:inputOutput id="Var_Reserved" name="reserved" type="boolean" elementId="Task_Reserve" />
       </uipath:variables>
```

Every pristine declaration preserved byte-for-byte; one declaration **added**, scoped to the very script task the prompt told the agent to insert. That addition isn't sloppiness — a mapped output on the new node *needs* a declaration (per #3211, a root variable without one does not exist to the canvas and every `vars.<id>` reference fails on import). So the whole-block freeze forces skill-following agents to choose between a canvas-broken file and a red check. Codex passes only by inserting a bare node that maps nothing.

## Fix

- `_shared/edit_check.py`: new `assert_variables_extended_only(original, edited)` — every pristine declaration must round-trip untouched (removals and modifications still fail; ids stay unique), while **additions are allowed**.
- `check_add_node.py` uses it in place of the freeze. The other three edit checks (`move`/`remove`/`update`) keep the full freeze — their prompts don't add nodes and none of them fails.
- `add_node.yaml` prompt states the contract explicitly: "every **existing** process variable declaration must round-trip untouched. You may declare new variables for the node you add."
- Unit tests (`test_edit_variables_preservation.py`, runs in the existing `maestro-bpmn checker unit tests` job): identical block ✓, the exact nightly addition shape ✓, modified pristine ✗, removed pristine ✗, duplicate ids ✗.

## Verification

| Case | Result |
|---|---|
| Real 09-17 claude artifact (scored 0.333 last night) | **PASS** with fixed check |
| Unedited pristine fixture (no node inserted) | still FAILS |
| Unit tests | 5/5 |

GH `run-coder-eval.yml` on this branch (agent=claude, task `add_node`) — link in a comment.


nikhil-maryala:
Verification run (this branch, agent=claude, task `add_node`): https://github.com/UiPath/skills/actions/runs/35238779319 — will post the result when it completes.

nikhil-maryala:
✅ Verify run [35238779319](https://github.com/UiPath/skills/actions/runs/35238779319) (agent=claude, this branch): **skill-bpmn-edit-add-node SUCCESS 1.0** (214s, 30 turns). Combined with the local matrix (today's real failing artifact passes; unedited fixture still fails; 5/5 unit tests), the fix is verified end-to-end.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -149,6 +149,31 @@ def assert_uipath_preserved(original: ET.Element, edited: ET.Element, local_name
...
+    declaration (a root variable without one does not exist to the canvas,
+    #3211), so freezing the whole block forces a choice between a canvas-broken
+    file and a red check."""
+    orig = _find_first(original, "variables")
+    new = _find_first(edited, "variables")
+    if orig is None:
+        fail("fixture bug: no uipath:variables in pristine original")
+    if new is None:
+        fail("uipath:variables was dropped by the edit (must be preserved)")
+    edited_by_id = {child.attrib.get("id"): child for child in new}
```
rockymadden, line 169: 🔴 **major — the block's own attributes are no longer compared**

The `assert_uipath_preserved(original, edited, "variables")` you removed compared `canonical()` on the `uipath:variables` element **itself**, attributes included. This walks children only, so the block's own `version` is now unguarded:

```xml
<uipath:variables version="v99">   <!-- green -->
  <uipath:inputOutput id="Var_OrderId" ... />   <!-- untouched -->
```

Nothing else in the task covers it: `add_node.yaml:52-54` only greps `preserve-only-do-not-touch` and `uipath:migrationVersion`, and the sibling payloads (`migrationVersion`, `caseManagement`) keep their `version` frozen. Net effect: this PR narrows the contract on an axis it never meant to touch.

Fix, right after the `None` guards:

```python
if sorted(orig.attrib.items()) != sorted(new.attrib.items()):
    fail("uipath:variables attributes were modified (must round-trip untouched)")
```

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -149,6 +149,31 @@ def assert_uipath_preserved(original: ET.Element, edited: ET.Element, local_name
...
+    for child in orig:
+        child_id = child.attrib.get("id")
+        match = edited_by_id.get(child_id)
+        if match is None:
+            fail(f"pristine variable {child_id!r} was removed (must be preserved)")
+        if canonical(child) != canonical(match):
+            fail(f"pristine variable {child_id!r} was modified (must round-trip untouched)")
+    ids = [child.attrib.get("id", "") for child in new]
+    if not all(ids) or len(ids) != len(set(ids)):
```
rockymadden: 🔴 **major — additions get in on a unique `id` and nothing else**

This is the only gate on what the edit *adds*. A typo'd scope sails through:

```xml
<uipath:inputOutput id="Var_Reserved" name="reserved" type="boolean" elementId="Task_Reserv" />
```

That is the exact state the PR cites #3211 for. `references/structural-bpmn.md:159`: *"Without an `elementId` the declaration does not exist to the canvas, and every `vars.<id>` reference to it fails on import"* — and `check_variable_scopes.py:70-76` already fails a dangling `elementId`, noting the canvas deletes it on import (`migrations/s191_2-migration-utils.ts`). So the loosened check green-lights the broken artifact it was loosened to permit the fix for. Same for an addition with no `name` or `type` (`structural-bpmn.md:150`).

`edited` is already a parameter, so validating the additions is local:

```python
pristine_ids = {c.attrib.get("id") for c in orig}
live = {e.attrib["id"] for e in edited.iter()
        if e.attrib.get("id") and e.tag.startswith("{%s}" % BPMN_NS)}
for child in (c for c in new if c.attrib.get("id") not in pristine_ids):
    child_id = child.attrib.get("id")
    scope = child.attrib.get("elementId")
    if scope and scope not in live:
        fail(f"new variable {child_id!r} has a dangling elementId {scope!r}")
    if not child.attrib.get("name") or not child.attrib.get("type"):
        fail(f"new variable {child_id!r} needs a non-empty name and type")
```

Build the id set that way rather than reusing `all_ids` — `all_ids` accepts a DI shape id, which the canvas treats as an orphan (`check_variable_scopes.py:52-58`).

**Judgment call for you:** `check_variable_scopes.py:63-68` requires `elementId` to be *present* on every declaration, but `structural-bpmn.md:97` ships `<uipath:inputOutput id="Var_Amount" name="Amount" type="double" />` with none. Requiring presence here would red-check an agent that copied that example. Settle the skill first; `no dangling` is unambiguous today, `must be present` isn't.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -149,6 +149,31 @@ def assert_uipath_preserved(original: ET.Element, edited: ET.Element, local_name
...
+    #3211), so freezing the whole block forces a choice between a canvas-broken
+    file and a red check."""
+    orig = _find_first(original, "variables")
+    new = _find_first(edited, "variables")
+    if orig is None:
+        fail("fixture bug: no uipath:variables in pristine original")
+    if new is None:
+        fail("uipath:variables was dropped by the edit (must be preserved)")
+    edited_by_id = {child.attrib.get("id"): child for child in new}
+    for child in orig:
```
rockymadden, line 170: 🟡 **minor — pristine declaration order is no longer enforced**

The old whole-block `canonical()` compare pinned child order; this id-keyed lookup is order-insensitive, so an edit that shuffles the pristine declarations now passes.

I could not verify whether the canvas depends on declaration order. If it does, that's a third hole. If it doesn't, put that in the docstring so the next reader doesn't re-litigate it.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -149,6 +149,31 @@ def assert_uipath_preserved(original: ET.Element, edited: ET.Element, local_name
         fail(f"uipath:{local_name} payload was modified (must round-trip untouched)")
 
 
+def assert_variables_extended_only(original: ET.Element, edited: ET.Element) -> None:
+    """Every pristine variable declaration must round-trip untouched; new
+    declarations are allowed. An inserted node's mapped output NEEDS a new
+    declaration (a root variable without one does not exist to the canvas,
+    #3211), so freezing the whole block forces a choice between a canvas-broken
```
rockymadden: 🟡 **minor — docstring argues the PR instead of stating the contract**

Lines 155-157 explain *why the change was made*; a future reader acts on the contract, not the history. Same for the `# Additions allowed: …` restatement at `check_add_node.py:52` and `test_edit_variables_preservation.py:4-6`.

```python
"""Pristine variable declarations must round-trip untouched; additions are allowed."""
```

The `#3211` rationale belongs in this PR and in git history, where it already is.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_duplicate_ids_fail() -> None:
+    with pytest.raises(SystemExit):
+        assert_variables_extended_only(
+            original(),
+            edited(
+                '<uipath:inputOutput id="Var_OrderId" name="OrderId" type="string" elementId="Process_1" />\n'
+                '        <uipath:inputOutput id="Var_Qty" name="Qty" type="number" elementId="Process_1" />\n'
+                '        <uipath:inputOutput id="Var_Qty" name="QtyCopy" type="number" elementId="Process_1" />'
```
rockymadden: 🟡 **minor — this test is green on the wrong assertion**

`edited_by_id` (line 164) is last-wins, so `"Var_Qty"` resolves to *this* `QtyCopy` element and the pristine loop fails first with `pristine variable 'Var_Qty' was modified`. `SystemExit` fires, the test passes — and `edit_check.py:173`, the unique-ids branch it names, never runs. It's the only uncovered line in the new helper, and the PR body's "duplicate ids ✗" row rests on it.

Make the duplicate byte-identical so the pristine loop passes and the id check is reached:

```python
'        <uipath:inputOutput id="Var_Qty" name="Qty" type="number" elementId="Process_1" />'
```

And pin all four negatives to the branch they mean, so two branches collapsing into one gets caught:

```python
with pytest.raises(SystemExit, match="unique non-empty ids"):
```

rockymadden (changes requested):
## 🔴 Changes requested

Right diagnosis, right direction. But the replacement guard **drops a check the old one had** and **validates nothing about what it now lets in** — including the exact canvas-broken shape this PR cites #3211 to avoid. Both fixes are a few lines inside `assert_variables_extended_only`.

## 🔍 What it does

- `check_add_node.py`: whole-block `uipath:variables` freeze ➜ `assert_variables_extended_only` (pristine declarations round-trip, additions allowed).
- `add_node.yaml`: prompt states the relaxed contract.
- 5 unit tests, collected by the existing checker job.

**Premise holds.** `skills/uipath-maestro-bpmn/references/structural-bpmn.md:159` — *"Without an `elementId` the declaration does not exist to the canvas, and every `vars.<id>` reference to it fails on import."* A skill-following agent that maps an output on the node it inserts **must** declare a variable, so the old freeze did force canvas-broken-or-red.

**PR body checks out** where I could test it against the tree: `move`/`remove`/`update` do keep the freeze (`check_move_node.py:42`, `check_remove_node.py:43`, `check_update_node.py:63`), and `add_output`/`group_to_subflow` never froze variables — so "the other three" is accurate, not hand-waving. No slop in the description.

## 📋 Overall findings

**🟡 minor — the eval still scores a bare insert like a correct one**

The PR body's own complaint (*"codex passes only by inserting a bare node that maps nothing"*) survives the fix. Nothing requires the inserted node to carry a `uipath:mapping`, and nothing requires a new declaration at all — so `insert an empty script task` still grades 1.0, identical to a mapped one. `check_add_output.py:39-53` is the shape that closes it (require the added variable, require it wired to an output).

As it stands the fix removes a false red without adding the true signal. Either require the mapping in prompt + check, or state in `description` that mapping is out of scope for this task.

**🟡 minor — four different variable contracts across six edit checks**

| check | variables contract |
|---|---|
| `add_node` | extended-only (this PR) |
| `move` / `remove` / `update` | full freeze |
| `add_output` | `variable_ids` set-subset (`check_add_output.py:36-42`) — catches removals, **misses a modified pristine declaration** |
| `group_to_subflow` | nothing at all |

Pointing `add_output` at the new helper gets it modification detection for one line, and `group_to_subflow` gets a guard it has never had. Not a blocker for this PR, but the divergence is now wide enough to be a trap.

**🟢 Confirmed vs 🟡 taken on your word**

- ✅ Confirmed: the new tests are collected — `test-helpers.yml:37` runs `pytest tests/tasks/uipath-maestro-bpmn/ -v`, and `maestro-bpmn checker unit tests` is green on this branch.
- ✅ Confirmed by reading: pristine removal and modification still fail; an addition with no `id` also fails (`not all(ids)`).
- 🟡 Not verified: the *"Real 09-17 claude artifact — PASS"* row. The artifact isn't in the repo and I ran nothing. Your verify run (`35238779319`, **1.0**) plus the green checker job is the evidence; I'm taking that row on your word.

## tl;dr

The freeze was wrong and this is the right way out. Two gaps in the new helper before it lands: the `uipath:variables` element's own attributes (`version="v1"`) are no longer compared, so a bumped version now passes silently; and additions are admitted on a unique `id` alone, so a dangling `elementId` — the canvas-broken state you cite #3211 for — scores green. Plus one test that passes on the wrong assertion, leaving the duplicate-id branch uncovered. All small.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 139873115 — your exact snippet, placed right after the None guards. Regression test: `test_block_attribute_change_fails` (version v99 → fails).

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 139873115 — additions are now validated: non-empty name+type, and a present elementId must reference a live BPMN-namespace element id (built per your note, not via all_ids, so DI shape ids don't count as live). On the judgment call: went with 'no dangling' only — presence NOT required — because the skill's own examples still ship elementId-less declarations (structural-bpmn.md's minimal-file example at ~:97, another #2369-era contradiction with :159; flagged for a follow-up alongside the expression-authoring fix in #3377 rather than widened here). Tests: `test_dangling_elementid_on_addition_fails`, `test_addition_missing_name_or_type_fails`, `test_addition_without_elementid_passes`.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 139873115 — pristine relative order is enforced again (the old whole-block compare pinned it, so keeping it pinned restores the exact prior strictness rather than betting on canvas order-insensitivity). Test: `test_reordered_pristine_fails`.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 139873115 — the docstring now states the contract (round-trip incl. block attributes and order; additions validated); the history/#3211 rationale lives in the PR and commit messages. The check-site comment and test-file docstring history are gone too.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
nikhil-maryala: Fixed in 139873115 — the duplicate is now byte-identical so the pristine loop passes and the unique-ids branch is what fires, and all negatives pin their branch with pytest.raises(SystemExit, match=...). 10/10.

nikhil-maryala:
All findings addressed in 139873115 — thanks @rockymadden, both majors were real holes (the dangling-scope one especially: the helper was green-lighting the exact state it cited #3211 to avoid).

**Inline (5/5, replied per thread):** block attributes compared again; additions validated (name+type required; present elementId must reference a live BPMN element id — presence itself not required per your judgment-call note, since `structural-bpmn.md`'s minimal example still ships elementId-less declarations, a contradiction I'm flagging for follow-up rather than widening this PR); pristine order enforced again; docstrings state the contract; duplicate-id test now exercises its branch and every negative pins its branch with `match=`.

**Overall minors:** (1) bare-insert scope — stated explicitly in the yaml `description`: mapping on the inserted node is deliberately out of scope; a bare and a mapped insert both pass, added declarations are validated not required. Requiring a mapping would change task difficulty and invalidate the verify baseline — happy to do that as its own task change if you want the signal. (2) contract divergence across the six edit checks — agreed it's a trap; left out of this PR for reviewability, will follow up pointing `add_output` at the shared helper (one line, gains modification detection) and giving `group_to_subflow` its first guard.

10/10 unit tests; the real 09-17 artifact still passes; unedited fixture still fails. Re-requesting review.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
```diff
@@ -0,0 +1,133 @@
...
+def test_removed_pristine_fails() -> None:
+    with pytest.raises(SystemExit, match="was removed"):
+        assert_variables_extended_only(
+            tree(),
+            tree('<uipath:inputOutput id="Var_OrderId" name="OrderId" type="string" elementId="Process_1" />'),
+        )
+
+
+def test_duplicate_ids_fail() -> None:
+    # Byte-identical duplicate: the pristine round-trip loop passes, so the
```
rockymadden: 🟡 **minor — this comment now describes the wrong mechanism**

The unique-ids check moved to `edit_check.py:166-168`, *above* the pristine round-trip loop. So byte-identity is no longer what routes this to the unique-ids branch — that branch fires first for any duplicate, identical or not. A future reader debugging a duplicate-id failure would look in the pristine loop on the strength of this comment.

The test is correct; only the comment is stale. Drop it, or:

```python
# Duplicate ids fail before the pristine round-trip loop runs.
```

rockymadden (approved):
## 🟢 Approved

All 5 inline findings fixed in `139873115`, correctly and with a regression test each. Both deferred minors are legitimate follow-ups, not dodges. CI green on the new head (38 success, 14 skipped, **0 failures**), checker job included.

## ✅ Findings cleared

| finding | fix | pinned by |
|---|---|---|
| 🔴 block attributes unguarded | `edit_check.py:164-165` | `test_block_attribute_change_fails` |
| 🔴 additions admitted on `id` alone | `edit_check.py:189-197` — `name`+`type` required, dangling `elementId` fails | `test_dangling_elementid_on_addition_fails`, `test_addition_missing_name_or_type_fails` |
| 🟡 order dropped | `edit_check.py:177-181` | `test_reordered_pristine_fails` |
| 🟡 docstring narrated the PR | `edit_check.py:153-157`, test docstring, check-site comment gone | — |
| 🟡 duplicate test on the wrong assertion | byte-identical duplicate + `match=` on all seven negatives | `test_duplicate_ids_fail` |

Three things I'd call out as better than what I proposed:

- **Unique-ids moved above the pristine loop** (`:166-168`). That kills the last-wins masking at the root rather than working around it — a duplicate can no longer surface as `was modified` regardless of its content.
- **Order check filters edited ids to the pristine set** (`:179`), so interleaved additions don't false-fail. That was the easy thing to get wrong.
- **`live_ids` built from BPMN-namespaced ids** (`:184-188`), not `all_ids`, so a DI shape id doesn't count as a live scope.

Two judgment calls I checked rather than took on faith:

- **`elementId` presence not required** — correct call. `structural-bpmn.md:97` does ship an `elementId`-less `inputOutput`, so requiring presence would have red-checked an agent copying the skill's own minimal file. Flagging the skill contradiction as follow-up instead of widening this PR is the right split.
- **`name`+`type` required** — safe. I scanned every `*.md` under `skills/uipath-maestro-bpmn/` (multi-line aware): **zero** declaration examples carry an `id` without a `type`. No false-red surface.

Also verified the new `description` copy can't leak grading info: the agent reads `initial_prompt`; top-level `description` is eval metadata for humans (`tests/README.md:410-413`). Saying "a bare node and a mapped one both pass" there is documentation, not a hint.

## 📋 Overall findings

**🟡 minor — the 1.0 verify run predates the tightening**

`35238779319` scored `skill-bpmn-edit-add-node` **1.0** against `001fdfcca`, before four new assertions landed. The unit tests cover each new branch, and I have no specific reason to think the real artifact trips one (its addition was `name`/`type`-complete and scoped to the node it inserted). But the run that proves "no false red on a real agent artifact" no longer matches the code. A rerun on `139873115` is a cheap way to close the last gap between "tests pass" and "the nightly goes green" — your call whether to gate merge on it.

**🟢 deferred, agreed**

- Bare-insert signal: documenting mapping as out of scope is the honest resolution for this PR. Requiring a mapping does change task difficulty and would invalidate the baseline, so it belongs in its own change.
- Contract divergence across the six edit checks: fine as follow-up. `add_output`'s set-subset (`check_add_output.py:36-42`) is the one with a real hole — a preserved id with a changed `type` passes today.

## tl;dr

Both majors were real and are both properly closed, with the duplicate-id fix done better than I asked for. Ten tests, each pinned to its branch, all green in CI. One stale comment inline, and the 1.0 verify run predates the four new assertions — worth a rerun, not worth blocking. Ship it.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
nikhil-maryala: Fixed in 310e0505e — took your wording. (You're right about the mechanism: the unique-ids gate moved above the pristine loop in 139873115, so it fires for any duplicate; the byte-identical fixture is now just belt-and-braces.)

Issue #3376, opened 2026-09-17
eval: claude arm needs its own experiment run_limits defaults (900/1200 is codex-sized)


nikhil-maryala:
Follow-up to rockymadden's review on #3369 (finding 2).

`tests/experiments/nightly.yaml` hands every task `turn_timeout: 900` / `task_timeout: 1200`, sized for the codex arm. The claude-code arm streams the whole session as one turn, so `turn_timeout` is its effective wall budget — and 20+ BPMN tasks (plus other suites) still inherit the default. Result: per-task budget PRs keep recurring (#3150 → #3369) each time a task's claude duration drifts across the codex-sized cap.

**Proposal:** a claude-specific experiment config (e.g. `nightly-claude.yaml`) or per-agent `run_limits` support, defaulting the claude arm to ~1800/2400, so the next slow task is a config line instead of a third PR of this shape.

Evidence for the sizing: #3369's analysis — tool execution is 16–50s of 900–1800s walls, the rest is model generation; three-day nightly history shows tasks passing with 3–37s of margin under the current defaults.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

nikhil-maryala:
Implementation up: #3416 — `nightly-claude.yaml` (run_limits 1800/2400, everything else pinned identical to `nightly.yaml` by a sync guard that verifiably fails on drift), per-arm selection in `run-coder-eval.yml`, and a 2700/3600 bump for `smart_triage` (today's other victim, whose explicit limits outrank any default).

Fresh data since this issue was filed — the treadmill produced victims 4 and 5 in today's (09-18) claude nightly:
- `multi-city-weather` ERROR at **905s** vs the inherited **900s** default (exactly the class this issue describes)
- `smart-triage` ERROR at 1805s vs its explicit 1800s (1.0 on both 09-16 and 09-17)

Plus the measurement from #3379's verify: `variable_scopes_nested` legitimately takes **2354s** on the claude arm.

**The one line I can't land from the skills repo:** `coder_eval_uipath`'s `daily.sh` also passes `-e experiments/nightly.yaml` for every arm. It needs the claude branch of the same selection, e.g.:

```sh
[ "$ARM" = "claude" ] && EXPERIMENT=experiments/nightly-claude.yaml || EXPERIMENT=experiments/nightly.yaml
```

[...]

PR #3369, opened 2026-09-17
fix(tests): give borderline bpmn tasks turn budgets that clear claude's pace

nikhil-maryala:
## Problem

Six BPMN tasks ERRORed with `agent_timeout` in the claude nightly `2026-09-16_04-17-43`. Command-log analysis shows these are **not agent failures**: in every one the agent was mid-work when killed (61–136 turns, 37–77 tool calls), tool execution accounts for only 16–50s of the 900–1800s walls, and the rest is model generation between tool calls (gaps of 100–527s precede even `ls -la`). The claude-code session streams as **one turn**, so `turn_timeout` is the effective wall budget — and these budgets sit at Claude's *typical* duration, making every nightly a coin flip.

Three-day evidence (claude nightlies 09-14 / 09-15 / 09-16):

| Task | cap | 09-14 | 09-15 | 09-16 |
|---|---|---|---|---|
| `ai_decision_review/brownfield` | 900 (inherited) | ERROR 910s | **PASS 897s** (3s margin) | ERROR 904s |
| `author/variable_scopes_nested` | 1800 | **PASS 1763s** (37s margin) | PASS 1259s | ERROR 1803s |
| `approval_chain/near_miss` | 1800 | PASS 1467s | PASS 1419s | ERROR 1804s |
| `authoring/business_rule_task` | 1200 | PASS | PASS 387s | ERROR 1204s |
| `e2e/customer_escalation_triage` | 1200 | — | — | ERROR 1205s (first nightly) |

## Fix

Raise `turn_timeout` (and `task_timeout` with it — it silently caps the turn) with a rationale comment per task, following the merged #3150 precedent:

- `ai_decision_review/brownfield`: **1800/2400** (2× typical ~900s; previously inherited the experiment default)
- `author/variable_scopes_nested`, `approval_chain/near_miss`: **2700/3600** — same numbers as the already-merged heavy single-turn authoring tasks (`external_wait`, `high_volume_batch`, `contract_variant_wrappers`)
- `authoring/business_rule_task`: **1800/2400** (also *adds* `task_timeout`, which was missing — the inherited 1200s was the real cap)
- `e2e/customer_escalation_triage`: turn **2400**, task **4400** (task budget additionally covers ~2000s of graded `run_command`s: 180 + 180 + 1600)

**Deliberately not bumped:** `composition/nested_review` — killed at 1800s on all three days, and prior investigation showed larger budgets alone don't fix it (synthesis-volume bound; needs decomposition, tracked separately).

## Verification

[...]

PR #3379, opened 2026-09-17
fix(uipath-maestro-bpmn): scope the minimal example's own mutable variables

nikhil-maryala:
## Problem

`SKILL.md:170-171` tells agents to author from one specific block:

> **Assemble.** Author directly from the complete minimal file in [references/structural-bpmn.md](skills/uipath-maestro-bpmn/references/structural-bpmn.md#a-complete-minimal-file-author-from-this-not-from-examples)

That example declared its two internal mutable variables with **no `elementId`**:

```xml
<uipath:inputOutput id="Var_Amount" name="Amount" type="double" />
<uipath:inputOutput id="Var_Echo"  name="Echo"  type="double" />
```

55 lines below, the same section states the rule it breaks:

> Without an `elementId` the declaration does not exist to the canvas, and every `vars.<id>` reference to it fails on import.

So the file the skill designates as its authoritative starting point violated its own rule. Concrete consequence, not theory: an agent that copies it verbatim also fails criterion 1 of the scope checkers — *"declarations without an elementId: […]. Every `<uipath:variables>` child needs one."* — which `author/variable_scopes` (`check_variable_scopes.py`) and `author/variable_scopes_nested` (`check_variable_scopes_nested.py`) already run. (Same rule, two files; my first draft attributed both to the former.)

## Provenance — corrected

My original framing ("#2369 reintroduced the elementId-less form", "stale-branch drift") was **wrong**, and @rockymadden caught it. Verified against the tree:

- Before `826927089` (#2369) the block was `<uipath:variables version="v1" />` — **empty**. #2369 authored all four declarations new; it reintroduced nothing.
- The omission was **deliberate and reasoned**. That commit is literally titled *"stop asserting the mutable bridge variable has no elementId"* and argues `elementId` is *"additive, not exclusive"* (`VariableUtil.ts:126-155` assigns `inputOutputVariables[inputOutput.id]` unconditionally). It removed checkers that had required the mutable bridge to carry **no** `elementId` — i.e. it made the attribute *optional*, not forbidden.
- #3211 and #3377 then settled the repo on **required**, which is what the section's own rule and `check_variable_scopes.py` now enforce.

So this is not drift: the example is **underspecified against the rule the repo now enforces**, and this change aligns it. That is the case for the change.

## Fix

- **`structural-bpmn.md`**: both internal declarations get `elementId="Process_1"` (2 lines). The public `input`/`output` declarations already carried theirs.
- **`test_bpmn_public_io_contract.py`**: the guard pinned `elementId` on **public** declarations only — which is exactly why this went undetected. It now asserts *every* declaration in the example carries an `elementId` naming a live BPMN element (never a DI shape id, which the canvas treats as an orphan).

## Verification

- New guard **fails** on the pre-fix example (stashed the doc change and re-ran: `AssertionError` on `Var_Amount has no elementId`) and **passes** after — so it genuinely pins the rule rather than passing vacuously.
- `tests/scripts/test_bpmn_public_io_contract.py`: 3/3. Pointing a declaration at a DI shape id also fails the guard, so the DI exclusion is load-bearing.
- Live eval runs on this branch, agent=claude: `author-variable-scopes` **1.0**, `author-validate` **1.0**, `author-variable-scopes-nested` — reported in a comment (it was missing from the first round, per review).
- `npm run skills:validate`: OK — default 27 skills / 1768 files, studioweb 27 / 1768 / 264 replacements.
- No flavor override exists for `structural-bpmn.md` and the example sits outside any `skill-flavor` block, so this is a plain canonical edit.
- Local `pytest tests/tasks/uipath-maestro-bpmn/` is 80F/333P both **with and without** this change (identical baseline on clean `main`) — pre-existing, from local Python 3.9 vs the repo's 3.13; the CI checker job runs 3.13.


nikhil-maryala:
Verification run (this branch, agent=claude — `variable_scopes` is the task whose check enforces the elementId rule, `author_validate` authors straight from the minimal example): https://github.com/UiPath/skills/actions/runs/35251924926

nikhil-maryala:
✅ Verify run [35251924926](https://github.com/UiPath/skills/actions/runs/35251924926) (agent=claude, this branch):

| task | result |
|---|---|
| `skill-bpmn-author-variable-scopes` — the task whose check enforces the elementId rule | **SUCCESS 1.0** (266s) |
| `skill-bpmn-author-validate` — authors straight from the minimal example | **SUCCESS 1.0** (235s) |

Both were already passing, so this confirms the corrected example doesn't regress them; the contract guard is what prevents the example from drifting back (it fails on the pre-fix shape).

tests/scripts/test_bpmn_public_io_contract.py
```diff
@@ -35,6 +35,29 @@ def _mapping_outputs(element: ET.Element) -> list[ET.Element]:
     return mapping.findall("uipath:output", NS)
 
 
+def test_minimal_example_scopes_every_declaration_to_a_live_element() -> None:
+    # The section's own rule: "Without an `elementId` the declaration does not
+    # exist to the canvas, and every `vars.<id>` reference to it fails on
+    # import." The example is what SKILL.md tells agents to author from, so an
+    # elementId-less declaration here ships the canvas-broken shape the rule
+    # forbids. Scope must name a live BPMN element, never a DI shape id.
```
rockymadden: 🟡 **minor** — four of these five lines restate the rule the assertion below already states, plus the PR's own rationale. Only the last clause changes what a future reader does.

```python
    # Scope must name a live BPMN element, never a DI shape id — a DI id is an
    # orphan the canvas deletes on import.
```

The provenance and the "SKILL.md tells agents to author from this" framing belong in the commit message, which already carries both.

tests/scripts/test_bpmn_public_io_contract.py
```diff
@@ -35,6 +35,29 @@ def _mapping_outputs(element: ET.Element) -> list[ET.Element]:
...
+    root = minimal_example()
+    process = root.find("bpmn:process", NS)
+    assert process is not None
+
+    live_ids = {
+        element.attrib["id"]
+        for element in process.iter()
+        if element.attrib.get("id")
+        and element.tag.startswith("{" + NS["bpmn"] + "}")
+    } | {process.attrib["id"]}
```
rockymadden: 🟡 **minor** — `| {process.attrib["id"]}` is dead. `process.iter()` yields `process` itself first, and its tag is `{…MODEL}process`, so `Process_1` is already in the set from the comprehension. Drop the union.

It mirrors `_shared/check_variable_scopes.py:54-57`, which carries the same redundancy — if the mirroring is deliberate, say so, otherwise both can lose it.

tests/scripts/test_bpmn_public_io_contract.py
```diff
@@ -35,6 +35,29 @@ def _mapping_outputs(element: ET.Element) -> list[ET.Element]:
...
+
+    for variable_id, variable in _variables(process).items():
```
rockymadden: 🟡 **minor** — nothing asserts the block is non-empty, so the loop body never runs and the test passes if a future edit empties `<uipath:variables>`. The PR says the guard *"genuinely pins the rule rather than passing vacuously"* — true today only because `test_minimal_example_bridges_public_input_to_mutable_state` would fail first on its `assert public_inputs`.

Both siblings in this file guard explicitly, and `check_variable_scopes.py:62-63` does the same with `if not decls: fail(...)`. Match them:

```python
    variables = _variables(process)
    assert variables, "the minimal example declares no variables"

    for variable_id, variable in variables.items():
```

rockymadden (changes requested):
🔴 **Changes requested** — the two-line fix is right, but the same drift is still in this file's prose, and the PR claims this was the last instance.

---

## 📦 What

| File | Change |
|---|---|
| `structural-bpmn.md:97,99` | the two internal `uipath:inputOutput` declarations in the canonical minimal example get `elementId="Process_1"` |
| `test_bpmn_public_io_contract.py:38-58` | new guard — every declaration in the example must carry an `elementId` resolving to a live BPMN element, not just the public pair |

The XML edit is correct, and the reasoning checks out against the repo:

- ✅ `SKILL.md:170` does point agents at this block as the authoring source.
- ✅ The section's own rule sits at `structural-bpmn.md:159`.
- ✅ `_shared/check_variable_scopes.py` criterion 1 fails an elementId-less declaration.
- ✅ The guard runs in CI — `test-helpers.yml:194`.
- ✅ Single copy of the file, no flavor override, no `skill-flavor` block, so it is a plain canonical edit.
- ✅ Repo-wide scan of every `<uipath:variables>` block under `skills/` and `tests/`: no other elementId-less declaration in XML.

---

## 🔴 Overall findings

### major — the fix is incomplete: same file, same contract, prose instead of XML

`structural-bpmn.md:218-228`, under *"Public entry-point variables have a two-layer runtime contract"*, spells out `elementId` for both public halves and stays silent on the mutable one:

> Declare each public `uipath:input` with `elementId` bound to its intended StartEvent **and a mutable internal `uipath:inputOutput` with the stable id used by process expressions.**

> **Declare a mutable internal `uipath:inputOutput`**, plus a public `uipath:output` bound with `elementId` to the root EndEvent that returns it.

Both bullets landed in `826927089` (#2369) — the same commit as the XML this PR corrects. #3377 fixed the `expression-authoring.md` twin (now *"scoped to the process — `elementId="<process id>"`, never absent"*) and missed these two.

**Consequence:** an agent that follows the bullets instead of copying the XML block still emits the elementId-less shape and still fails `check_variable_scopes.py` criterion 1. The new guard parses `xml` fences only, so it cannot catch this.

**Fix** — name the scope in both clauses:

```diff
-  StartEvent and a mutable internal `uipath:inputOutput` with the stable id used
-  by process expressions.
+  StartEvent and a mutable internal `uipath:inputOutput` scoped to the process
+  (`elementId="<process id>"`) with the stable id used by process expressions.
```

```diff
-- Declare a mutable internal `uipath:inputOutput`, plus a public
+- Declare a mutable internal `uipath:inputOutput` scoped to the process
+  (`elementId="<process id>"`), plus a public
```

Until that lands, *"This is the last instance of that drift"* is not accurate.

### minor — provenance and verification are overstated

- **"reintroduced"** — #2369 did not revert #3211 here. Before `826927089` the block was `<uipath:variables version="v1" />`, empty; #2369 authored both lines new. And the omission was deliberate: that same commit removed *"three checkers [that] required the mutable `uipath:inputOutput` in a start-event bridge to carry no `elementId`"*, on the argument that `elementId` is *"additive, not exclusive"* (`VariableUtil.ts:126-155`). That is a reasoned position, later reversed by #3377 — not stale-branch drift. The change is still right; it aligns the example with the rule the repo now enforces. Sell it on that, not on the drift story.
- **checker attribution** — `author/variable_scopes_nested` runs `check_variable_scopes_nested.py`, not `check_variable_scopes.py`. Its criterion 1 is the same rule, so the substance holds; the file name does not.
- **the missing run** — reported live runs are `author-variable-scopes` and `author-validate`. `author-variable-scopes-nested` — the second task this PR names as impacted — has no reported score.

### out of scope, flagging only

`expression-authoring.md:100-101` states a node-scoped variable *"is not surfaced as a root runtime variable, so `debug-instance variables-all` cannot show it"*, while #2369's commit reported the opposite from live runs (*"outputs declared with `elementId` bound to the end event still appear as keys in the `Globals` map"*). Not this diff's doing. Worth settling before the edit checks tighten on `#3378`'s follow-up.

### not verified

I read the tests and ran nothing. The A/B claim on the new guard, `npm run skills:validate`, and both eval scores are unconfirmed on my side.

---

## tl;dr

Right change, incomplete. 🔴 Scope the two mutable `inputOutput` clauses in the two-layer prose bullets at `structural-bpmn.md:218-228` — same file, same source commit, and the new guard can't see prose — and this genuinely becomes the last instance. Three test nits inline. Provenance in the description says "reintroduced" where #2369 actually made an explicit, opposite call.

tests/scripts/test_bpmn_public_io_contract.py
nikhil-maryala: Fixed in 824f65537 — took your two-line version verbatim; the provenance and SKILL.md framing stay in the commit message.

tests/scripts/test_bpmn_public_io_contract.py
nikhil-maryala: Fixed in 824f65537 — dead, confirmed: `ET.Element.iter()` yields the element itself first (checked directly: `iter` on `<a id=A><b id=B/></a>` returns `['A','B']`), so `Process_1` was already in the set. Union dropped, with a one-line note so the next reader doesn't re-add it. The mirroring in `check_variable_scopes.py:54-57` was not deliberate on my side — worth removing there too, but I've left it out of this diff since that file belongs to the eval checks rather than the doc guards.

tests/scripts/test_bpmn_public_io_contract.py
nikhil-maryala: Fixed in 824f65537 — added `assert variables, "the minimal example declares no variables"` before the loop, matching both siblings and `check_variable_scopes.py:62-63`. You're right that my "not vacuous" claim held only via the sibling test's `assert public_inputs`; it's now self-standing.

nikhil-maryala:
All findings addressed in 824f65537 — thanks @rockymadden, especially for the provenance correction; I had the story wrong and the diff is better justified without it.

**🔴 major — prose carried the same gap.** Right, and the new guard structurally cannot see it (it parses `xml` fences only). `structural-bpmn.md:220-225` now scopes both mutable clauses: *"a mutable internal `uipath:inputOutput` scoped to the process (`elementId="<process id>"`)"*, using your diffs. With that in, the file no longer contradicts itself in either XML or prose.

**🟡 provenance overstated.** Verified and you're correct on every point — I've rewritten that section of the description:
- Pre-`826927089` the block was `<uipath:variables version="v1" />`, empty. #2369 authored all four lines new; nothing was reintroduced.
- Its own title is *"stop asserting the mutable bridge variable has no elementId"*, arguing `elementId` is *"additive, not exclusive"* (`VariableUtil.ts:126-155`) — it made the attribute **optional**, not forbidden, and removed the checkers that required its absence. A reasoned call, later settled the other way by #3211/#3377.
- So: the example is **underspecified against the rule the repo now enforces**, not drift. The "last instance of that drift" claim is gone from the description.

**🟡 checker attribution.** Fixed — `variable_scopes` runs `check_variable_scopes.py`, `variable_scopes_nested` runs `check_variable_scopes_nested.py`; same criterion 1, two files.

**🟡 missing run.** Dispatched: https://github.com/UiPath/skills/actions/runs/35256163202 — `author-variable-scopes-nested` score to follow.

**Inline (3/3)** replied: guard comment trimmed to your version, dead `| {process.attrib["id"]}` union dropped (verified `Element.iter()` includes self), non-empty assertion added so the loop can't pass vacuously.

**On your out-of-scope flag** — `expression-authoring.md:100-101` ("node-scoped … `variables-all` cannot show it") versus #2369's live finding that end-event-scoped outputs *do* appear in `Globals`: agreed that's unsettled and agreed it should be settled before the #3378 follow-up tightens `elementId` presence in the edit checks. I'd rather resolve it with a live `debug-instance variables-all` read than by picking a side in the docs — I'll open an issue with both citations so it isn't lost.

nikhil-maryala:
Opened #3382 for the `variables-all` visibility contradiction, with both citations and a live-read plan to settle it.

nikhil-maryala:
One addition beyond your list, same finding class: `expression-authoring.md:117` (*"Entry point inputs that must later be updated need a separate mutable `uipath:inputOutput` variable"*) also named no scope. Fixed in db7eb714a.

Repo-wide sweep for the class is now clean — every place the skill tells an agent to declare a mutable internal variable names the scope, in prose and in XML:

| location | state |
|---|---|
| `structural-bpmn.md` minimal example (XML) | fixed, f95da4022 |
| `structural-bpmn.md:220-225` two-layer bullets | fixed, 824f65537 |
| `expression-authoring.md:96-98` | already fixed in #3377 |
| `expression-authoring.md:117` | fixed, db7eb714a |
| every other `<uipath:variables>` block under `skills/` and `tests/` | clean (your scan, re-confirmed) |

So "the last instance" is accurate as of db7eb714a — and it's the *class* that's closed, not just the one file.

rockymadden (approved):
🟢 **Approved** — major fixed, all four minors fixed, and you closed a third instance of the class that my scan missed.

---

## ✅ Findings from the first review

| finding | state |
|---|---|
| 🔴 **major** — two-layer prose bullets omitted `elementId` on the mutable `inputOutput` | **fixed** — `structural-bpmn.md:221,224` |
| 🟡 provenance overstated ("reintroduced", stale-branch drift) | **fixed** — description rewritten, claim withdrawn |
| 🟡 checker misattribution | **fixed** — both scripts named correctly |
| 🟡 missing nested eval run | **dispatched** — see below |
| 🟡 guard comment restated the rule it guards | **fixed** — `:39-40` |
| 🟡 dead `\| {process.attrib["id"]}` | **fixed** — dropped |
| 🟡 vacuous pass | **fixed** — `assert variables` added |

Plus one that was not on my list: `expression-authoring.md:117-119` named no scope either. That is the third site in the class, and the one my sweep missed — I scanned `<uipath:variables>` blocks, which is XML, and the class lives in prose too.

---

## 🔍 Verified independently

- **Re-ran the sweep in prose**, every `uipath:inputOutput` mention across the skill with code fences stripped. Your table holds:
  - `registry-workflow.md:70-71` — already names `elementId="<node-id>"` ✅
  - `failure-modes.md:28-35` — about missing *mappings*, not declarations, so no scope instruction belongs there ✅
  - `expression-authoring.md:96-98` — fixed in #3377 ✅
  - `structural-bpmn.md:221,224` and `expression-authoring.md:118-119` — fixed here ✅

  So *"the class is closed"* is the accurate claim, and it is stronger than the one I challenged.
- **Dropping the union is safe.** `Element.iter()` yields self first and `<bpmn:process>` is in the bpmn namespace carrying an `id`, so `Process_1` stays in `live_ids`. Side benefit: a process with no `id` now fails an assertion instead of raising `KeyError`.
- **`maestro-bpmn contract guards` is green on `db7eb714a`** — the job that actually runs the new guard.

---

## 📋 Not blocking

- `author-variable-scopes-nested` (run `35256163202`) is still `in_progress`. Confirmatory only: `check_variable_scopes_nested.py` criterion 1 **requires** `elementId`, so an example that gains the attribute cannot regress it. Post the number when it lands; no reason to hold the merge for it.
- `Run skill smoke tests` still pending. Everything else green.
- Opening #3382 for the `variables-all` contradiction instead of picking a side in the docs is the right call — a live `debug-instance variables-all` read is the only thing that settles it, and #3378's follow-up needs it settled.

---

## Observation, not a finding

`expression-authoring.md:118-119` now restates the scope rule that bullet `96-98` states 20 lines above it in the same list. Defensible as written: `96` is conditioned on *"each value used by decisions, tasks, or diagnostics"*, which an agent could read as not covering an entry-point mirror, so the repetition removes a real ambiguity rather than just spending tokens. Leaving it.

---

## tl;dr

🟢 All five findings cleared. The prose sweep is closed as a class, not just this file, and the guard still holds after the union came out. Nested eval score is confirmatory and can land after merge. Good correction on the provenance — the change reads better justified without the drift story than with it.

nikhil-maryala:
✅ The missing run, per your review: [35256163202](https://github.com/UiPath/skills/actions/runs/35256163202) (1 requested / 1 graded) — **`skill-bpmn-author-variable-scopes-nested`: SUCCESS 1.0** (2354s).

Full evidence set for this PR is now complete on the claude harness:

| task | score |
|---|---|
| `author-variable-scopes` (runs `check_variable_scopes.py`) | 1.0 |
| `author-variable-scopes-nested` (runs `check_variable_scopes_nested.py`) | 1.0 |
| `author-validate` (authors from the minimal example) | 1.0 |

Side note worth flagging: 2354s against `variable_scopes_nested`'s post-#3369 budget of turn 2700 / task 3600 — so that task is still slow enough that the pre-#3369 1800s cap would have killed it again. Consistent with #3376's argument that the per-task budget layer is the wrong one.

PR #3380, opened 2026-09-17
test(uipath-maestro-bpmn): close the variables gaps in add_output and group_to_subflow

nikhil-maryala:
## Problem

The #3378 follow-up @rockymadden asked for: *"four different variable contracts across six edit checks … the divergence is now wide enough to be a trap."*

| check | contract before this PR |
|---|---|
| `add_node` | extended-only (#3378) |
| `move` / `remove` / `update` | full freeze |
| `add_output` | **set-of-ids subset** — catches removals, misses a pristine declaration silently retyped or renamed |
| `group_to_subflow` | **nothing at all** |

## Fix

**`add_output`** → `assert_variables_extended_only` for the preservation half, keeping its own *"an addition is REQUIRED"* assertion (that is this task's point).

**`group_to_subflow`** → deliberately **not** that helper. A grouping edit may key a variable inside the subprocess it now belongs to, which `structural-bpmn.md` documents as importing cleanly (*"A subprocess-level variable is keyed inside its own subprocess: the subprocess id, or the node in it that writes the value — both import cleanly"*). Freezing the root block would red-check that correct edit — the same false-red class this suite spent the week removing (#3377, #3378).

It gets a new `assert_variables_preserved_or_rescoped`: every pristine declaration must survive **somewhere in the file** with its kind, `name` and `type` intact; only `elementId` may change, and it must still name a live BPMN element (never a DI shape id, which the canvas orphans). Duplicate ids across blocks fail, since one id declared twice does not import.

## Verification

**The `add_output` gap was real** — retyping the pristine `Var_Total` in the actual 09-17 claude artifact:

```
main:            OK: new output + variable added, existing output and payloads preserved   (exit 0)
this branch:     FAIL: pristine variable 'Var_Total' was modified (must round-trip untouched)   (exit 1)
```

**No false reds introduced** — both changed checks run clean against all four real 09-17 nightly artifacts (claude + codex × `add_output` + `group_to_subflow`), all of which passed last night:

| artifact | `add_output` | `group_to_subflow` |
|---|---|---|
| claude `2026-09-17_04-22-38` | exit 0 | exit 0 |
| codex `2026-09-17_04-19-12` | exit 0 | exit 0 |

**Unit tests**: 17/17 in `test_edit_variables_preservation.py`, every negative pinned to its branch with `pytest.raises(..., match=...)` per the #3378 review. New cases cover the documented re-scope passing, plus dropped / retyped / kind-changed / dangling-scope / duplicate-id failing.

Remaining divergence is now intentional and narrow: full freeze where the prompt forbids touching variables, extended-only where the task adds a node or an output, preserved-or-rescoped where it regroups.


nikhil-maryala:
Verification run (this branch, agent=claude, both affected tasks): https://github.com/UiPath/skills/actions/runs/35252371736

nikhil-maryala:
✅ Verify run [35252371736](https://github.com/UiPath/skills/actions/runs/35252371736) (agent=claude, this branch) — 2 tasks requested, 2 graded:

| task | result |
|---|---|
| `skill-bpmn-edit-add-output` | **SUCCESS 1.0** (46s) |
| `skill-bpmn-edit-group-to-subflow` | **SUCCESS 1.0** (379s) |

Both tightened checks pass live, matching the offline result against all four real 09-17 nightly artifacts. So the added detection (a retyped pristine declaration in `add_output`; any dropped/retyped/re-kinded/dangling declaration in `group_to_subflow`) costs no false reds.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -197,6 +197,64 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+            fail(f"pristine variable {child_id!r} was dropped by the edit")
+        for attribute in ("name", "type"):
+            if child.attrib.get(attribute) != match.attrib.get(attribute):
+                fail(
+                    f"pristine variable {child_id!r} changed its {attribute} "
+                    f"({child.attrib.get(attribute)!r} -> {match.attrib.get(attribute)!r})"
+                )
+        if local(child.tag) != local(match.tag):
+            fail(f"pristine variable {child_id!r} changed kind (input/output/inputOutput)")
+        scope = match.attrib.get("elementId")
```
rockymadden, line 365: 🔴 **major** — `elementId` can be **deleted** outright and this passes.

`match.attrib.get("elementId")` is `None` when the attribute is gone, so `if scope and ...` short-circuits. The docstring two screens up says *"only `elementId` may change, and it must still name a live BPMN element"* — removal is neither.

`structural-bpmn.md:159` is explicit about the cost:

> Without an `elementId` the declaration does not exist to the canvas, and every `vars.<id>` reference to it fails on import.

In this fixture `Var_OrderId` is read as `=vars.Var_OrderId` by `Task_Pick`, `Task_Pack` and `Task_Ship`. An agent that "re-scopes" by dropping the attribute produces a file the canvas strips the variable from, and gets exit 0.

```py
scope = match.attrib.get("elementId")
if child.attrib.get("elementId") and not scope:
    fail(f"pristine variable {child_id!r} lost its elementId (the canvas drops the declaration)")
if scope and scope not in live_ids:
    fail(f"variable {child_id!r} has a dangling elementId {scope!r}")
```

Needs a negative test too — none of the 7 new cases covers it.

tests/tasks/uipath-maestro-bpmn/_shared/check_group_to_subflow.py
```diff
@@ -59,6 +60,7 @@ def main() -> None:
     assert_config_preserved(original, edited, ["Task_Ship"])
     assert_uipath_preserved(original, edited, "migrationVersion")
     assert_uipath_preserved(original, edited, "caseManagement")
+    assert_variables_preserved_or_rescoped(original, edited)
```
rockymadden, line 63: 🔴 **major** — the re-scope allowance is unbounded, and this fixture carries the cross-scope read that exposes it.

`Task_Ship` stays in the main flow and `assert_config_preserved` on line 60 freezes its mapping. That mapping reads `=vars.Var_OrderId` and writes `var="Var_Label"`. Both are pristine root-level declarations, so the new helper lets either one re-scope into the subprocess:

- `Var_OrderId` → `elementId="Sub_Pack"`: `Task_Ship`, **outside** the subprocess, still reads it.
- `Var_Label` → `elementId="Sub_Pack"`: `Task_Ship`, **outside**, still writes it.

Both exit 0. Nothing else on this path looks at `vars.*` references — I grepped every `_shared/` helper this check calls.

**Fix:** bound it. A pristine variable may re-scope into a subprocess only when no element outside that subprocess still references it — `source`, `var`, or a `=vars.<id>` inside a `uipath:mapping` CDATA body.

**I could not verify** the canvas half: I reasoned from the scoping rule at `structural-bpmn.md:157-159` that a subprocess-scoped variable is not visible to a parent-scope sibling, and did not confirm it against an import. If it *is* visible, say so and this finding drops. If it isn't, the check is green on a file that fails on import.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -197,6 +197,64 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+    Looser than ``assert_variables_extended_only`` on purpose: an edit that
+    groups nodes into a subprocess may re-scope a variable into that
+    subprocess's own block, which the skill documents as importing cleanly.
+    Freezing the root block would fail that correct edit."""
+    pristine = _declarations_anywhere(original)
+    if not pristine:
+        fail("fixture bug: no uipath:variables declarations in pristine original")
+    current = _declarations_anywhere(edited)
+    live_ids = _live_bpmn_ids(edited)
+    for child_id, child in pristine.items():
```
rockymadden, line 348: 🔴 **major** — additions get no validation here, unlike `assert_variables_extended_only`.

This loop walks `pristine` only, so a declaration the edit **adds** is never inspected: no `name`/`type` requirement, and `live_ids` (computed on line 240) is never applied to it. This passes:

```xml
  <uipath:inputOutput id="Var_OrderId" name="OrderId" type="string" elementId="Process_Fulfillment" />
  <uipath:inputOutput id="Var_Label" name="Label" type="string" elementId="Process_Fulfillment" />
  <uipath:inputOutput id="Var_Junk" elementId="Sub_Packk" />
</uipath:variables>
```

No `name`, no `type`, an `elementId` naming nothing — three canvas-broken states, exit 0. Adding a subprocess-scoped variable is a likely path in on this task: `structural-bpmn.md:663` tells the agent to *"re-scope their variables"* when moving logic into a subprocess.

**Fix:** the same four lines `extended_only` already has at `edit_check.py:189-197`. After the pristine loop, walk `current` for ids absent from `pristine` and apply the `name`/`type` and dangling-`elementId` checks.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -197,6 +197,64 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
             fail(f"new variable {child_id!r} has a dangling elementId {scope!r}")
 
 
+def _live_bpmn_ids(root: ET.Element) -> set[str]:
```
rockymadden, line 192: 🟡 **minor** — the extraction left the original copy in place.

`assert_variables_extended_only` still inlines this exact comprehension at `edit_check.py:184-188`, explanatory comment included. Two copies of one rule about which ids the canvas accepts as a variable scope is the divergence this PR is about — the next person to learn something new about DI ids will fix one of them.

Point line 184 at `_live_bpmn_ids(edited)` and delete the duplicate.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -197,6 +197,64 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+    pristine = _declarations_anywhere(original)
+    if not pristine:
+        fail("fixture bug: no uipath:variables declarations in pristine original")
+    current = _declarations_anywhere(edited)
+    live_ids = _live_bpmn_ids(edited)
+    for child_id, child in pristine.items():
+        match = current.get(child_id)
+        if match is None:
+            fail(f"pristine variable {child_id!r} was dropped by the edit")
+        for attribute in ("name", "type"):
```
rockymadden: 🟡 **minor** — `name`/`type` only: the declaration's body and every other attribute are free to change.

`extended_only` compares with `canonical`, which covers all attributes and the element text. This compares two attributes plus the tag, so a `jsonSchema` declaration's CDATA body can be emptied and pass. That shape is documented at `structural-bpmn.md:329-333`:

```xml
<uipath:inputOutput id="Var_ScriptError" name="Error"
  type="jsonSchema" elementId="Task_RiskScore"><![CDATA[
{"type":"object","properties":{"code":{"type":"string"}}}
]]></uipath:inputOutput>
```

Latent for this fixture, live for the next caller of a shared helper.

**Fix:** invert the contract to match the docstring — everything frozen *except* the one thing that may move:

```py
def _frozen_view(el: ET.Element):
    return (
        local(el.tag),
        tuple(sorted((k, v) for k, v in el.attrib.items() if k != "elementId")),
        (el.text or "").strip(),
    )
```

That replaces the `name`/`type` loop and the kind check with one comparison. You lose the per-attribute message, so keep a `name`/`type` diff for the common case if the wording matters.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -197,6 +197,64 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+    """Every ``uipath:variables`` child in the file, keyed by id — the root
+    block plus any subprocess-level block."""
+    found: dict[str, ET.Element] = {}
+    for block in root.iter():
+        if local(block.tag) != "variables":
+            continue
+        for child in block:
+            child_id = child.attrib.get("id", "")
+            if not child_id:
+                fail("a uipath:variables declaration has no id")
```
rockymadden: 🟡 **minor** — these two messages blame the agent for a fixture defect.

`_declarations_anywhere` runs on `original` first (line 236), so a pristine fixture with a missing or duplicated declaration id reports `a uipath:variables declaration has no id` / `variable id 'X' is declared more than once` — indistinguishable from the same defect in the **edited** file. Every other helper in this module prefixes `fixture bug:` for that side: `load_original`, `assert_preserved`, `assert_uipath_preserved`, and line 238 immediately below.

**Fix:** take a `where: str` and prefix it — `_declarations_anywhere(original, "pristine original")`.

It also repairs `test_rescope_duplicating_an_id_across_blocks_fails`, whose `match="declared more than once"` currently cannot tell which side raised. That is the pinning #3378 asked for.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
```diff
@@ -130,3 +133,98 @@ def test_addition_missing_name_or_type_fails() -> None:
                 + '\n        <uipath:inputOutput id="Var_Reserved" name="reserved" elementId="Task_Reserve" />'
             ),
         )
+
+
+# --- assert_variables_preserved_or_rescoped (grouping edits) ---
```
rockymadden, line 138: 🟡 **minor** — two of the new `fail` branches are unreached by these 7 cases.

- `"a uipath:variables declaration has no id"` — `edit_check.py:220`
- `"fixture bug: no uipath:variables declarations in pristine original"` — `edit_check.py:238`

The #3378 review asked for every negative pinned to its branch with `match=`. Two more `grouped(...)` calls close it:

```py
def test_declaration_without_an_id_fails() -> None:
    with pytest.raises(SystemExit, match="has no id"):
        assert_variables_preserved_or_rescoped(
            grouped('<uipath:inputOutput name="OrderId" type="string" elementId="Process_1" />'),
            grouped(ORDER_ID),
        )


def test_empty_pristine_block_is_a_fixture_bug() -> None:
    with pytest.raises(SystemExit, match="fixture bug"):
        assert_variables_preserved_or_rescoped(grouped(""), grouped(ORDER_ID))
```

Add a third for the dropped-`elementId` hole once that guard exists.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
```diff
@@ -10,7 +10,10 @@
 import pytest
 
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
-from _shared.edit_check import assert_variables_extended_only  # noqa: E402
+from _shared.edit_check import (  # noqa: E402
```
rockymadden, line 13: 🟡 **minor** — the module docstring on line 2 still reads *"Unit tests for assert_variables_extended_only (edit_check.py)"*, while this import pulls in a second helper the file now covers half its cases for.

Retitle it: `Unit tests for the edit_check variable guards.`

tests/tasks/uipath-maestro-bpmn/_shared/check_add_output.py
```diff
@@ -33,13 +34,13 @@ def main() -> None:
     _path, edited = parse_bpmn("Invoicing")
     original = load_original("edit/add_output", "Invoicing.bpmn")
 
+    # This task REQUIRES an addition; the shared guard covers the other half —
+    # pristine declarations round-trip untouched and additions are well-formed.
```
rockymadden: 🟡 **minor** — *"additions are well-formed"* claims more than the helper delivers.

`assert_variables_extended_only` resolves its block with `_find_first`, so it inspects only the **first** `uipath:variables` block. `variable_ids` on line 41 scans every block in the file. A new declaration placed in a node-scoped block on `Task_Calc` therefore satisfies line 42 and never reaches the `name`/`type`/dangling-`elementId` checks.

The `_find_first` narrowing is pre-existing (#3378) and not this PR's to fix. The comment asserting the guarantee is new, and a future reader will trust it.

**Fix:** drop the second sentence or scope it to what is true — *"pristine declarations round-trip untouched"*. The first sentence is the part that changes behavior: it's why the local assert below stays.

rockymadden (changes requested):
**Recommendation** 🔴 **Changes requested** — the `add_output` half is right and you demonstrated it. The new `assert_variables_preserved_or_rescoped` has three holes of the same class this PR exists to close, all inside one 30-line function.

---

### 📋 What this does

- **`check_add_output.py`** → `assert_variables_extended_only` for preservation, keeps its own "an addition is required" assert. The set-of-ids comparison is gone.
- **`check_group_to_subflow.py`** → new `assert_variables_preserved_or_rescoped`: every pristine declaration must survive somewhere in the file with its kind, `name` and `type`; only `elementId` may move.
- **7 new unit tests**, 17 total in the module.

---

### ✅ What I confirmed

- **The rationale for not reusing the frozen helper holds.** `structural-bpmn.md:157-159` — *"A subprocess-level variable is keyed inside its own subprocess: the subprocess id, or the node in it that writes the value — both import cleanly"* — and `:662-664` tells the agent to *"re-scope their variables"* on this exact edit. Freezing the root block would red-check a documented-correct edit. Right call.
- **Contract count is as described**: 3 across 6 checks — full freeze on `move`/`remove`/`update` (`assert_uipath_preserved(..., "variables")`), extended-only on `add_node`/`add_output`, preserved-or-rescoped on `group_to_subflow`.
- **The new tests actually run.** `test-helpers.yml` runs `pytest tests/tasks/uipath-maestro-bpmn/ -v` with no `-k` and no `paths:` filter, so they are collected. 17 `test_` functions, matching your claim.
- **`SUCCESS 1.0` is backed** — run 35252371736 is green on this branch.
- I read the tests, did not run them.

---

### 🔍 Overall findings

**🟡 minor — two new `fail` branches have no test.** The new code adds four `fail` sites; the 7 new cases reach two of them. Untested: `"a uipath:variables declaration has no id"` (`edit_check.py:220`) and `"fixture bug: no uipath:variables declarations in pristine original"` (`edit_check.py:238`). The #3378 review asked for every negative pinned to its branch. Two more `grouped(...)` calls cover both: an original with an id-less declaration, and an original with an empty block.

**Out of scope, not a finding, but it caps what `add_output` now gets:** `assert_variables_extended_only` resolves its block with `_find_first`, so it only ever inspects the **first** `uipath:variables` block, while `variable_ids` scans every block in the file. A new declaration placed in a node-scoped block on `Task_Calc` therefore satisfies the "addition required" assert and never reaches the `name`/`type`/dangling-`elementId` checks. Pre-existing from #3378, not caused here — worth its own PR.

**Not yours:** `Claude Code Review` is the only red check. Its failed-job log is empty and the Codex bot posted a usage-limit comment on the PR. Bot infra.

**No AI slop.** The description is the right shape: the main-vs-branch exit-code contrast and the artifact table are specific and falsifiable, and I could re-derive both. The Slack post repeating it near-verbatim is fine.

---

### tl;dr

`add_output` is correct and proven. `assert_variables_preserved_or_rescoped` leaks three ways: a pristine `elementId` can be **deleted** outright (docs say the canvas then drops the declaration and every `vars.<id>` reference fails), added declarations get **no validation at all**, and the re-scope allowance is **unbounded** — this fixture's frozen `Task_Ship` reads `vars.Var_OrderId` from outside the subprocess, so moving that variable in passes a check that should catch it. All three are a handful of lines in the function you just wrote.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in bd3dac9af — your snippet, plus `test_rescope_losing_the_elementid_fails`. Verified on the real 09-17 claude artifact: dropping `Var_OrderId`'s `elementId` now gives `FAIL: pristine variable 'Var_OrderId' lost its elementId (the canvas drops the declaration and every vars reference to it)`.

tests/tasks/uipath-maestro-bpmn/_shared/check_group_to_subflow.py
nikhil-maryala: Fixed in bd3dac9af — and your canvas question is answered inside the repo, so the finding stands rather than dropping: `expression-authoring.md` § Common mistakes lists *"Moving a variable into a subprocess without updating mappings that read it from the root scope"*, and `structural-bpmn.md:663` says to re-scope **their** variables — the moved logic's own. A variable still read from root scope isn't one of those.

The bound: a re-scope into a subprocess is allowed only when nothing outside it still references the variable — `var="<id>"` or a `vars.<id>` expression, CDATA mapping bodies included. Both of your cases are caught on the real artifact:
```
Var_OrderId -> SubProcess_Prepare:  FAIL: variable 'Var_OrderId' was re-scoped into subprocess 'SubProcess_Prepare' while 'Task_Ship' outside it still references it
Var_Label   -> SubProcess_Prepare:  FAIL: ... while 'Task_Ship' outside it still references it
```
The message names the owning flow node rather than the `uipath:input`/`uipath:output` child, since that's what an author acts on. Covered by `test_rescope_with_an_outside_reference_fails`.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in bd3dac9af — the addition checks are now a shared `_assert_addition_is_well_formed` called by **both** helpers, so they can't drift apart again. Your exact junk declaration (`<uipath:inputOutput id="Var_Junk" elementId="Sub_Packk" />`) fails on the real artifact with `needs a non-empty name and type`. Tests: `test_rescope_validates_additions`, `test_rescope_addition_with_a_dangling_scope_fails`.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in bd3dac9af — `assert_variables_extended_only` now calls `_live_bpmn_ids(edited)`; the duplicated comprehension and its comment are gone. One copy of the DI-ids rule.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in bd3dac9af — inverted to `_frozen_view` (tag + all attributes except `elementId` + text), so a `jsonSchema` CDATA body can no longer be emptied. Kept the specific `name`/`type`/kind messages for the common case by diffing those first and falling through to `was modified — only its elementId may change`. Test: `test_rescope_body_change_fails`, pinned to that message.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in bd3dac9af — `_declarations_anywhere(root, where)` now names the side and prefixes `fixture bug:` for the pristine one. `test_rescope_duplicating_an_id_across_blocks_fails` is re-pinned to `declared more than once in the edited file`, and there's a matching `test_declaration_without_an_id_is_a_fixture_bug_on_the_original` for the other side.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
nikhil-maryala: Fixed in bd3dac9af — both branches now covered (`test_declaration_without_an_id_in_the_edit_fails`, `test_empty_pristine_block_is_a_fixture_bug`), plus the third you asked for once the dropped-`elementId` guard existed. 25 tests, 20 `pytest.raises` sites, **zero** unpinned `SystemExit` — I also tightened one loose alternation regex of my own into `match="only its elementId may change"`.

tests/tasks/uipath-maestro-bpmn/_shared/test_edit_variables_preservation.py
nikhil-maryala: Fixed in bd3dac9af — retitled to `Unit tests for the edit_check variable guards.`

tests/tasks/uipath-maestro-bpmn/_shared/check_add_output.py
nikhil-maryala: Fixed in bd3dac9af — second sentence dropped; the comment now claims only *"pristine declarations round-trip untouched"*. Also opening a follow-up for the `_find_first` narrowing itself so the gap is tracked rather than just un-claimed.

nikhil-maryala:
All nine findings addressed in bd3dac9af — thanks @rockymadden. All three majors were real, and I verified each against the **actual 09-17 claude artifact** rather than a synthetic fixture:

| mutation of the real artifact | before | after |
|---|---|---|
| drop `Var_OrderId`'s `elementId` | exit 0 | `FAIL: ... lost its elementId (the canvas drops the declaration ...)` |
| re-scope `Var_OrderId` → subprocess (Task_Ship reads it) | exit 0 | `FAIL: ... while 'Task_Ship' outside it still references it` |
| re-scope `Var_Label` → subprocess (Task_Ship writes it) | exit 0 | same |
| add `<uipath:inputOutput id="Var_Junk" elementId="Sub_Packk" />` | exit 0 | `FAIL: ... needs a non-empty name and type` |

**On the canvas half you couldn't verify** — it's answered inside the repo, so the finding stands rather than dropping: `expression-authoring.md` § Common mistakes lists *"Moving a variable into a subprocess without updating mappings that read it from the root scope"*, and `structural-bpmn.md:663` says to re-scope **their** variables, i.e. the moved logic's own. A variable still read from root scope isn't one of those, so the reference bound matches the documented intent exactly. I still haven't confirmed it against a live import, and I've said so rather than implying I did.

**Structural change worth noting:** the addition checks are now one shared `_assert_addition_is_well_formed` used by *both* helpers — the divergence this PR is about, applied to itself.

**Follow-ups filed, not silently dropped:** #3384 for the `_find_first` narrowing you flagged as out of scope (pre-existing from #3378).

25 unit tests, 20 pinned `pytest.raises` sites, zero unpinned `SystemExit` — including the two branches your review found unreached, and one loose alternation regex of mine that I tightened. All four real artifacts still pass. Re-requesting review.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -179,22 +180,203 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+    ``var="<id>"`` or a ``vars.<id>`` expression, including inside a CDATA
+    mapping body. Declarations are not references, so ``uipath:variables``
+    blocks are skipped."""
+    declared: set[int] = set()
+    for block in edited.iter():
+        if local(block.tag) == "variables":
+            for node in block.iter():
+                declared.add(id(node))
+            declared.add(id(block))
+    pattern = re.compile(r"vars\.%s(?![\w.-])" % re.escape(variable_id))
```
rockymadden: 🟡 **minor** — the lookahead excludes `.`, so a property read on the variable is not seen as a reference.

`(?![\w.-])` rejects a following dot, so for `Var_Order` the pattern misses `vars.Var_Order.id`. That's the documented way to read a structured variable — `expression-authoring.md:33-34`:

> Read it as `vars.Error.<field>` — for example `=vars.Error.code == "SERVICE_UNAVAILABLE"`

Declared types include `object`, `json` and `jsonSchema` (`structural-bpmn.md:161-163`), and a field read off one of those is exactly this shape. Such a variable could re-scope into the subprocess with an outside `vars.X.field` reader still in place, and the guard you just added would stay quiet.

Dropping the dot keeps the prefix protection intact — `vars\.Var_Order(?![\w-])` still refuses to match `vars.Var_OrderId`, and now matches `vars.Var_Order.id`:

```py
pattern = re.compile(r"vars\.%s(?![\w-])" % re.escape(variable_id))
```

**Latent, not live:** every declaration in both fixtures is `string`, and `vars.Error.*` is the engine-surfaced key rather than a declared id, so I could not find a corpus instance where this bites. Cheap to close anyway.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -179,22 +180,203 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+    scope = child.attrib.get("elementId")
+    if scope and scope not in live_ids:
+        fail(f"new variable {child_id!r} has a dangling elementId {scope!r}")
+
+
+def _declarations_anywhere(root: ET.Element, where: str) -> dict[str, ET.Element]:
+    """Every ``uipath:variables`` child in the file, keyed by id — the root
+    block plus any subprocess-level block. ``where`` names the side being read
+    so a defect in the pristine fixture is not reported as an agent error."""
+    prefix = "fixture bug: " if where == "pristine original" else ""
```
rockymadden: 🟡 **minor** — behaviour switches on a magic string.

`where` does two jobs: it selects the `fixture bug: ` prefix by exact equality, and it is interpolated into the message. A call site that writes `"the pristine original"` or `"pristine_original"` still reads fine in the message and silently loses the prefix — the failure mode is invisible, which is what this finding was about in the first place.

House rule is an enum for a parameter that selects behaviour:

```py
class Side(Enum):
    ORIGINAL = "pristine original"
    EDITED = "edited file"

def _declarations_anywhere(root: ET.Element, side: Side) -> dict[str, ET.Element]:
    prefix = "fixture bug: " if side is Side.ORIGINAL else ""
```

Two call sites, both in this file.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
```diff
@@ -179,22 +180,203 @@ def assert_variables_extended_only(original: ET.Element, edited: ET.Element) ->
...
+        if element.attrib.get("id") != scope:
+            continue
+        if local(element.tag) == "subProcess":
+            return scope
+        enclosing = _enclosing_subprocess_ids(element, parents)
+        return enclosing[0] if enclosing else None
+    return None
+
+
+def _owning_element_id(
```
rockymadden, line 286: 🟡 **minor** — returns a `repr`, not an id.

The name promises an element id; both returns are `repr(...)`, so the value is display text and the caller can't compare it to an id. Return the id and let the message apply `!r`:

```py
        if current.attrib.get("id"):
            return current.attrib["id"]
    ...
    return local(element.tag)
```

then `{_owning_element_id(element, parents)!r}` at the `fail` on line 371.

rockymadden (approved):
**Recommendation** 🟢 **Approved** — all 3 majors and all 6 minors cleared, each with a pinned negative test. 25 tests, up from 17. What's left is minor and none of it can fire on the two tasks in scope.

---

### ✅ Fixed

| finding | fix |
|---|---|
| 🔴 `elementId` deletable | `edit_check.py:355-359` — explicit "lost its elementId" branch. `test_rescope_losing_the_elementid_fails` |
| 🔴 unbounded re-scope | `_owning_subprocess` + `_referencing_elements` + `_enclosing_subprocess_ids`. A move into a subprocess now fails while anything outside it still reads or writes the variable. `test_rescope_with_an_outside_reference_fails` builds the real shape — a `Task_Ship` with `var="Var_Label"` left in the root scope |
| 🔴 additions unvalidated | `edit_check.py:376-379` walks `current` for non-pristine ids through the same `_assert_addition_is_well_formed` `extended_only` uses. Two tests |
| 🟡 duplicated live-id set | `extended_only` now calls `_live_bpmn_ids` (`:183`) |
| 🟡 `name`/`type` only | `_frozen_view` freezes tag, text and every attribute but `elementId`, with the specific `name`/`type`/kind messages kept as the narrower branch. `test_rescope_body_change_fails` pins the `jsonSchema` CDATA case |
| 🟡 fixture-side blame | `_declarations_anywhere(root, where)`. Three tests now separate the two sides in `match=` |
| 🟡 untested branches | all four new `fail` sites reached |
| 🟡 stale docstring | retitled |
| 🟡 comment overstated the guard | second sentence dropped |

**Traced, not assumed:** I walked all 25 cases against the new code. The 10 `extended_only` tests are behaviour-identical through the extraction, and the 7 earlier re-scope tests still reach the branches their `match=` names — `test_rescope_into_subprocess_block_passes` in particular still passes, because `_referencing_elements` skips declaration nodes, so the moved declaration is not its own outside reference.

**No false red expected on the nightly artifacts**, and this is reasoning rather than observation: the reference rule is gated on `scope != child.attrib.get("elementId")` (`:367`), so it cannot fire on a harness that leaves `elementId` untouched — which is what both of them do today. Worth noting the verify run cited on this PR (35252371736) ran against `26961804f2`, one commit back, so the live evidence for the guard in its current form is the 25 unit tests plus that trace. A re-run would convert it.

---

### 🔍 Overall findings

**🟡 minor — the smoke red is not yours.** `Run skill smoke tests` fails at 92.9% on `skill-bpmn-parallel-fork-join: status=FAILURE`. `parallel/fork_join/check_fork_join.py` imports nothing from `edit_check.py` and no `assert_variables*` helper — confirmed by grep — so it shares no code with this diff. It's a task-run failure tripping the 95% floor, not a checker regression. `maestro-bpmn checker unit tests` is green.

---

### tl;dr

All nine findings fixed properly rather than papered over — the re-scope guard actually walks the reference graph instead of just widening a string match, and every new branch has a negative test pinned to it. Three minors left, all cosmetic or latent: the `vars.<id>` regex won't see a `vars.<id>.<field>` property read, `where` switches behaviour on a magic string, and `_owning_element_id` returns a `repr` instead of an id. None reachable on these two fixtures. Smoke red is an unrelated flaky task. Approved.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 713ac1c6e — real miss, thanks. Lookahead is now `(?![\w-])`: a following `.` continues a property read (so `vars.Var_Order.id` **is** a reference), while more id characters still mean a different variable. Two tests pin both directions — `test_rescope_with_an_outside_property_read_fails` (a `Task_Ship` reading `=vars.Var_OrderId.id` from root scope blocks the move) and `test_rescope_ignores_a_longer_id_prefix` (`vars.Var_OrderIdExtra` does not).

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 713ac1c6e — `Side` enum (`ORIGINAL`/`EDITED`); the prefix switches on `side is Side.ORIGINAL` and the message interpolates `side.value`. Agreed on the irony: a silent-failure mode inside the fix for a silent-failure mode.

tests/tasks/uipath-maestro-bpmn/_shared/edit_check.py
nikhil-maryala: Fixed in 713ac1c6e — returns the id (or the local tag name as the fallback), and the `fail` applies `!r`.

nikhil-maryala:
Three minors fixed in 713ac1c6e, and your point about the stale verify run is well taken — the run I cited (35252371736) was against `26961804f2`, before the three majors were fixed. Re-dispatched on current HEAD: https://github.com/UiPath/skills/actions/runs/35261705624 — I'll post those scores, so the guard in its current form has live evidence rather than a trace plus unit tests.

The lookahead finding was the substantive one: `vars.X.field` is how `object`/`json`/`jsonSchema` variables are read, and my `(?![\w.-])` silently excluded exactly that. Both directions are now pinned.

Also: the red `Run skill smoke tests` is `skill-bpmn-parallel-fork-join` at 13/14 (92.9% vs the 95% gate) — unrelated to this diff, and the same task that flaked on #3377's smoke run. Rerun dispatched.

nikhil-maryala:
✅ Re-verify on current HEAD (713ac1c6e), converting the trace into live evidence as you asked: [35261705624](https://github.com/UiPath/skills/actions/runs/35261705624), 2 requested / 2 graded —

| task | result |
|---|---|
| `skill-bpmn-edit-add-output` | **SUCCESS 1.0** (91s) |
| `skill-bpmn-edit-group-to-subflow` | **SUCCESS 1.0** (255s) |

So the guard in its final form — dropped-`elementId` branch, reference-bounded re-scope with property reads counted, validated additions, `_frozen_view`, `Side` enum — is green on live agent output, not just on the 27 unit tests.

nikhil-maryala:
Diagnosed the red smoke check — **not this diff, and not a skill gap.** Pulled the smoke artifact:

`skill-bpmn-parallel-fork-join` scored 0.214, failing `FAIL: missing bpmn:parallelGateway`. The authored file contains:

```xml
<bpmn:ParallelGateway id="Gateway_Fork" ...```

PascalCase. The check is right to reject it; the agent's own summary claimed *"Validation result: Valid"* because `uip` wasn't on PATH in that sandbox and its fallback was a plain XML well-formedness parse.

Three things worth noting:
1. **The guidance already exists at the point of use** — `SKILL.md:337-340` says *"BPMN XML element names are case-sensitive … Do not write PascalCase tags like `<bpmn:IntermediateCatchEvent>`"*, and `structural-bpmn.md:20` repeats it. So this isn't a placement problem.
2. **It's model-specific.** The smoke suite runs `gpt-5.6-luna`; both nightly models pass this task at **1.0** today (codex and claude). It's nondeterministic on that model — it has passed on rerun earlier today.
3. **The gate math makes it a merge blocker for every BPMN PR:** 14 tasks against a 95% threshold means a single failure is 92.9%, so any one nondeterministic task blocks the required check. This has now hit #3377 and #3380 today.

Rerun dispatched (attempt 2 on this SHA). Flagging (3) as worth a team decision rather than repeated reruns — happy to open an issue if you agree.

PR #3381, opened 2026-09-17
fix(uipath-maestro-bpmn): move the CLI-evidence rule to the diagnose point of use

nikhil-maryala:
## Problem

Two diagnose tasks fail recurrently, on **both harnesses**, for the same behavior — reading the mocked CLI's backing files instead of calling the CLI:

| task | harness | score | detail |
|---|---|---|---|
| `skill-bpmn-debug-instance-inspect` | claude | **0.9** | the *only* criterion it misses: `command_not_executed` matched `cat mocks/responses/manifest.json` |
| `skill-bpmn-operate-diagnose-minimal-fault-triage` | codex | **0.8** (09-17 nightly) | judge: *"supplemented CLI diagnostics with direct filesystem reads … that exposed fixture-like run data"* |

The same task also took the smoke suite below its 95% gate twice in two days (on #3378's and #3377's runs), so this one behavior is costing both nightly score and CI stability.

## Root cause — NOT established (corrected)

My original framing was that the rule was inert because it sat in a gated reference. @rockymadden showed that doesn't hold, and I verified it:

**Both failing tasks already carry the directive in their prompts**, since #3320 (2026-09-15):
- `operate-diagnose/minimal_fault_triage.yaml:45` — *"do not inspect `mocks/` or `fixtures/` directly"*
- `debug/debug_instance_inspect_mocked/debug_instance_inspect_mocked.yaml:43-44` — same

A prompt directive outranks any SKILL.md, so placement was **not** the binding constraint on those runs. Something overrode an instruction the agent already had, and I don't know what. One unexamined candidate he raised: `uipath-troubleshoot` (loaded on `minimal_fault_triage` per #3320's commit message) carries no mock/fixture boundary at all and tells the agent to check the working directory top level first. I could not determine which skill was in effect — the nightly transcripts aren't in the repo.

**So this PR is no longer claiming a causal fix.** What it does claim:
1. The scoped first sentence is production-true and was missing at the point of use (production has no task prompt).
2. Rule 3, the canonical statement, now closes the category instead of describing it loosely.
3. The two observed improvements are suggestive, n=1 each, one with an LLM judge — reported as evidence, not proof.

I also withdrew the #3287 citation: that PR is **CLOSED, never merged**, its `init` mechanism naming is not on `main`, and its only eval evidence was a dispatch link with no posted score. Citing it as a proven lesson was wrong.

## Fix

State it where the agent decides how to gather evidence — the Diagnose bullet in SKILL.md's *Operate and diagnose* section — and **name the mechanism**, not just the outcome (the #3287 lesson: "do not create those files" didn't work until it said "do not run `uip maestro bpmn init`"):

> Every piece of diagnostic evidence comes from a `uip maestro bpmn ... --output json` read. Where response, mock, or fixture files back that CLI on disk, they are its backing data, not a shortcut — do NOT read them directly (no `cat`, `grep`, `head`, or editor read of a `mocks/`, `fixtures/`, or response-cache path) unless you were asked to debug the harness itself. …

7 added lines, no deletions. `CAPABILITY.md` rule 3 stays as the detailed statement — this is the pointer at the decision point, matching how Rule 16 and its point-of-use directive coexist after #3152/#3287.

## Verification

- `npm run skills:validate` → OK (default 27 skills / 1768 files; studioweb 27 / 1768 / 264 replacements).
- `scripts/check-skill-status.py` → OK, 27 skills, manifest valid.
- No `skill-flavors/*/uipath-maestro-bpmn/SKILL.md` override exists and the edited region carries no `skill-flavor` markers, so every flavor inherits this.
- Frontmatter untouched, so the activation-gate recall eval is not triggered.

Behavioral verification is the nightly: both tasks are mock-backed and cheap, so the next claude and codex runs show whether the point-of-use placement moves `debug-instance-inspect` 0.9 → 1.0 and holds `minimal-fault-triage`. I'll post those numbers on this PR when they land rather than claiming the fix works on placement precedent alone.


nikhil-maryala:
Verification run (this branch, agent=claude, both affected tasks): https://github.com/UiPath/skills/actions/runs/35252654908 — this is the one PR in today's batch whose fix is behavioral rather than a check correction, so the run is the evidence; I'll post both task scores when it completes.

nikhil-maryala:
✅ Verify run [35252654908](https://github.com/UiPath/skills/actions/runs/35252654908) (agent=claude, this branch):

**`skill-bpmn-operate-diagnose-minimal-fault-triage`: SUCCESS 0.996** (73s) — all six criteria pass, **including the llm_judge that scored 0.000 in the 09-17 codex nightly** ("Agent diagnosed only and avoided BPMN lifecycle mutations"; it fails on reading `mocks/`/`fixtures/` directly). The agent gathered every piece of evidence through the CLI.

My first dispatch also named a task path that doesn't exist (`debug/instance_inspect/` — the task actually lives at `debug/debug_instance_inspect_mocked/`), so that glob silently matched nothing rather than failing loudly. Re-dispatched the second task on its real path: https://github.com/UiPath/skills/actions/runs/35253347412

nikhil-maryala:
✅ Second verify run [35253347412](https://github.com/UiPath/skills/actions/runs/35253347412) (correct path this time, 1 requested / 1 graded):

**`skill-bpmn-debug-instance-inspect`: SUCCESS 1.0** (50s) — all five criteria pass, including the one it has been missing (`Agent did not read the mocked fixture JSON directly instead of the CLI`), which held it at **0.9** in the 09-16 and 09-17 nightlies.

Both recurring mocks-read failures are now covered by live runs on this branch:

| task | nightly | this branch |
|---|---|---|
| `debug-instance-inspect` (claude) | 0.9 — mocks-read criterion failed | **1.0** |
| `operate-diagnose-minimal-fault-triage` (codex) | 0.8 — judge 0.000 | **0.996** |

Same skill text, opposite outcome on the exact criteria that were failing — consistent with #3152's A/B on point-of-use placement rather than a gated reference.

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -294,6 +294,13 @@ projects through the UiPath CLI.
   lifecycle actions): see [references/operate/CAPABILITY.md](references/operate/CAPABILITY.md).
 - **Diagnose** (fetch incidents, variables, and element executions, and trace a
   failed run back to its BPMN element): see [references/diagnose/CAPABILITY.md](references/diagnose/CAPABILITY.md).
+  Every piece of diagnostic evidence comes from a `uip maestro bpmn ... --output
```
rockymadden: 🟠 **major** — this is broader than the capability it fronts, and it forbids a read a sibling diagnose task requires.

- `references/diagnose/CAPABILITY.md:15` scopes it: *"Use `uip maestro bpmn ... --output json` **for diagnostic reads**"* — the CLI-servable data, not *every piece of evidence*.
- Rule 1's ladder ends at **generated package files**, and `failure-modes.md:87-89` lists its signs as `entry-points.json` referencing an old start event, `package-descriptor.json` omitting the current BPMN, `bindings_v2.json` not matching the BPMN. Those are disk reads — no `uip maestro bpmn` verb serves them.
- `tests/tasks/uipath-maestro-bpmn/diagnose/deployed_asset_drift/deployed_asset_drift.yaml:34` grades exactly the read this forbids: *"compare it element-by-element against the local `OrderRouting.bpmn`"*. That prompt has to say *"Reading the local `OrderRouting.bpmn` is expected"* — the prompt is the only thing now keeping the agent from obeying this line. Production has no such prompt.

➡️ Scope it to what the CLI actually serves:

```
  Runtime evidence — incidents, variables, element executions, cursors, the
  deployed asset — comes only from a `uip maestro bpmn ... --output json` read.
  Local `.bpmn` source and generated package files are read from disk as usual.
```

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -294,6 +294,13 @@ projects through the UiPath CLI.
...
+  json` read. Where response, mock, or fixture files back that CLI on disk, they
+  are its backing data, not a shortcut — do NOT read them directly (no `cat`,
+  `grep`, `head`, or editor read of a `mocks/`, `fixtures/`, or response-cache
```
rockymadden: 🟠 **major** — the enumeration is narrower than the graders it's meant to satisfy, so an agent can obey this line and still fail.

- `debug_instance_inspect_mocked.yaml:82` grades `(cat|less|head|tail|grep)\s+[\s\S]*mocks/responses`. **`less` and `tail` are graded and unnamed here** — `tail mocks/responses/manifest.json` is compliant with this sentence and fails the criterion.
- `minimal_fault_triage.yaml`'s judge fails *"any path under `mocks/` or `fixtures/`"* by any means: `jq`, `sed`, `python`, or the agent's own file-read tool. **"editor read" doesn't name a coding agent's `Read`/`View` tool** — the thing it most likely uses.

This is the #3287 shape one level in: the mechanism is named, but only three of its spellings, and a closed list reads as exhaustive.

➡️ Close the category instead of listing verbs:

```
  are its backing data, not a shortcut — do NOT read them by any means (no
  shell read such as `cat`, `less`, `head`, `tail`, `grep`, `jq`, or `sed`, and
  no file-read or editor tool) on a `mocks/`, `fixtures/`, or response-cache
```

skills/uipath-maestro-bpmn/SKILL.md
```diff
@@ -294,6 +294,13 @@ projects through the UiPath CLI.
...
+  path) unless you were asked to debug the harness itself. Reading the backing
+  files yields values the CLI would have reported anyway while skipping the
```
rockymadden: 🟡 **minor** — rationale, and it names a mechanism that doesn't exist.

- **"staleness checks the commands enforce"**: folder context is real (`CAPABILITY.md` rule 2). No `uip maestro bpmn` diagnostic read enforces a staleness check. The only staleness mechanism in the skill is `refresh` reporting stale files in `Data.WrittenFiles` (`failure-modes.md:81`) — a packaging command, not a diagnostic read. An agent that goes looking for the check this promises won't find one.
- The rest is justification for the sentence above it. It changes no action the agent takes, which is the bar for a line in a skill.

➡️ Drop both lines, or keep only the half that's true and load-bearing:

```
  path) unless you were asked to debug the harness itself. A direct read also
  skips the folder-context the instance and incident commands require.
```

rockymadden (changes requested):
**Recommendation: 🔴 Changes requested** — the placement is right; two of the seven lines are not. Line 297 forbids a read a sibling diagnose task requires, and line 300's verb list is narrower than the graders that were failing.

### 📦 What

7 added lines in `skills/uipath-maestro-bpmn/SKILL.md:297-303`, inside the **Diagnose** bullet: all diagnostic evidence via `uip maestro bpmn ... --output json`, no direct read of a `mocks/`, `fixtures/`, or response-cache path. No deletions, frontmatter untouched, `references/diagnose/CAPABILITY.md` rule 3 stays canonical.

### 🔎 Overall findings

**1. 🟠 major — the root cause isn't established.**

Both failing tasks already carry this directive **in the task prompt**, and have since #3320 (2026-09-15):

- `tests/tasks/uipath-maestro-bpmn/operate-diagnose/minimal_fault_triage.yaml:46` — *"do not inspect `mocks/` or `fixtures/` directly"*
- `tests/tasks/uipath-maestro-bpmn/debug/debug_instance_inspect_mocked/debug_instance_inspect_mocked.yaml:43` — same

The codex 09-17 failure happened with that line in its prompt. A prompt directive outranks any SKILL.md, so *"the rule is inert because it sits in a gated reference"* does not explain what was observed — something overrode a directive the agent had already been handed, and hoisting a third copy one level up doesn't address it.

One unexamined candidate: `uipath-troubleshoot`. #3320's own commit message records the agent loading it on `minimal_fault_triage`. Its description says *diagnostic intent outranks surface ownership*, invariant 4 is capture-to-disk-then-read, §4 of Source-required playbooks says **"CHECK THE WORKING DIRECTORY TOP LEVEL FIRST"** — and it carries no mock/fixture boundary anywhere (`grep -n 'mock\|fixture' skills/uipath-troubleshoot/SKILL.md` → nothing). If that is the skill in effect, this edit is in the wrong file.

**I could not verify which skill was loaded on either failing run** — the PR doesn't say, and the nightly transcripts aren't in the repo. Worth settling before the placement story lands in the commit message.

**2. 🟠 major — the codex regression is unverified.**

The results table reads `minimal-fault-triage (codex) 0.8 → 0.996`, but run [35252654908](https://github.com/UiPath/skills/actions/runs/35252654908) is `agent=claude`. The harness whose judge scored 0.000, and whose run breached the 95% pass-rate gate (`.github/workflows/smoke-skills.yml:686`), was never re-run. Both rows are n=1, one of them with an LLM judge in the loop.

➡️ Dispatch `agent=codex` on this branch for that task before the table claims it holds there.

**3. 🟡 minor — restates where the cited precedent points.**

#3152 landed as a **2-line** point-of-use directive that *points* at Rule 16 (`SKILL.md:210-213`) — the 9-line restating block was rejected in its review. This adds 7 restating lines while `CAPABILITY.md` rule 3 stays canonical. Two full statements of one rule drift: the reference says *"unless you are explicitly debugging the mock harness"*, this says *"unless you were asked to debug the harness itself"*.

➡️ One line that points, matching the precedent the description cites:
```
  Diagnostic evidence comes only from those CLI reads, never from the files
  backing the mocked CLI — see rule 3 in that reference.
```

**4. 🟡 minor — #3287 is cited as a lesson, but it's CLOSED unmerged.**

Its `init` mechanism naming is not on `main`: `SKILL.md:210-213` and Rule 16 (`SKILL.md:396-401`) still state the outcome only, no `uip maestro bpmn init`. Its only eval evidence is a dispatch link with no posted score. Cite it as the hypothesis it is, or land it first.

**5. 🟡 minor — harness-only content at the most expensive altitude.**

`mocks/`, `fixtures/`, response-cache paths and *"debug the harness itself"* exist only in the eval sandbox — no production Maestro diagnosis has them. 5 of the 7 lines are paid for on every load of a 29 KB SKILL.md to move two mock-backed tasks. The production-useful part is the first sentence, once scoped per the inline comment on line 297.

### ✅ Checked / ❌ not run

- **Confirmed:** no `skill-flavors/studioweb/uipath-maestro-bpmn/SKILL.md` override (that dir holds only `references/operate/run.md`), no `skill-flavor` markers in the edited region, frontmatter untouched, list-continuation indentation correct, CI green on all 20 checks.
- **Not run:** `npm run skills:validate`, `scripts/check-skill-status.py`, any eval. The green *Build and inspect every skill package* and *Validate skill status manifest & README* jobs cover the first two.

### tl;dr

Right instinct, wrong absolutes. `Every piece of diagnostic evidence comes from a CLI read` bans reading local `.bpmn` and generated package files — which `skill-bpmn-diagnose-deployed-drift` grades and `failure-modes.md` requires. The verb list (`cat`, `grep`, `head`) omits `less` and `tail`, which the failing criterion's own regex grades. And the root cause doesn't hold up: both tasks already had this directive in their prompts since #3320, so placement in a gated reference wasn't the binding constraint. Fix the two lines, re-run on codex.

skills/uipath-maestro-bpmn/SKILL.md
nikhil-maryala: Fixed in cc0f6e34f — verified and you're right on all three legs: `failure-modes.md:87-89`'s drift signs are `entry-points.json`/`package-descriptor.json`/`bindings_v2.json` reads that no `uip maestro bpmn` verb serves, and `deployed_asset_drift.yaml:26-27` has to tell the agent the local read is expected. My line would have been the thing arguing against it in production, where no prompt exists. Scoped to your wording: runtime evidence (incidents, variables, element executions, cursors, deployed asset) from the CLI; local `.bpmn` and generated package files from disk as usual.

skills/uipath-maestro-bpmn/SKILL.md
nikhil-maryala: Fixed in cc0f6e34f — and this was the sharpest catch: I enumerated three spellings of a closed category while the graded pattern is `(cat|less|head|tail|grep)` and the judge fails *any* means including a file-read tool. So an agent could obey my sentence literally (`tail mocks/responses/manifest.json`) and still fail the criterion — the #3287 shape one level in, exactly as you put it. Rather than a longer list at the most expensive altitude, the category is now closed in **rule 3** (the canonical statement: "by any means — no shell read such as cat/less/head/tail/grep/jq/sed, and no file-read or editor tool", plus the explicit carve-out that reading back your own captured CLI output is fine), and SKILL.md points at it.

skills/uipath-maestro-bpmn/SKILL.md
nikhil-maryala: Fixed in cc0f6e34f — both lines dropped. You're right that the staleness half was invented: the only staleness mechanism in the skill is `refresh` reporting stale files in `Data.WrittenFiles` (`failure-modes.md:81`), a packaging command, not a diagnostic read. A skill line that promises a mechanism the product doesn't have is worse than no line.

nikhil-maryala:
Substantially rewritten in cc0f6e34f — this review changed the PR rather than polishing it, thanks @rockymadden. I verified every claim against the tree and all of them hold.

**The premise was wrong.** Both failing tasks have carried *"do not inspect `mocks/` or `fixtures/` directly"* **in their prompts** since #3320 (`minimal_fault_triage.yaml:45`, `debug_instance_inspect_mocked.yaml:43-44`). A prompt directive outranks SKILL.md, so placement wasn't the binding constraint and my root-cause story doesn't survive. I've rewritten that section of the description to say so, and the PR no longer claims a causal fix — it claims a production-true scoped sentence plus a canonical rule that now closes its category. Your `uipath-troubleshoot` hypothesis is the best remaining lead and I've recorded it as unresolved rather than quietly dropping it.

**Both content majors were real:**
- Over-scope: `failure-modes.md:87-89`'s drift signs are package-file *disk* reads no CLI verb serves, and `deployed_asset_drift.yaml:26-27` has to tell the agent the local read is expected — in production, nothing would. Now scoped to runtime evidence, with local source and package files explicitly read from disk as usual.
- Closed-category-by-enumeration: I listed three spellings while the grader matches `(cat|less|head|tail|grep)` and the judge fails *any* means. `tail mocks/responses/manifest.json` obeyed my sentence and failed the criterion. The category is now closed in **rule 3** (by any means; shell reads *and* file-read/editor tools; with the explicit carve-out that reading back your own captured CLI output is fine), and SKILL.md points at it — #3152's actual 2-line pointing shape, 5 lines instead of 7, one canonical statement instead of two drifting ones.

**Dropped the invented mechanism:** no diagnostic read enforces a staleness check; the only such mechanism is `refresh`/`Data.WrittenFiles`, a packaging command. A skill line promising a mechanism the product lacks is worse than no line.

**#3287 citation withdrawn** — CLOSED, never merged, its `init` naming isn't on `main`, and its evidence was a dispatch link with no score. I've also corrected my own notes, which were carrying it as a proven precedent.

**Codex run dispatched**, as asked, for both tasks: https://github.com/UiPath/skills/actions/runs/35268641387 — the table won't claim the codex row until that posts. Re-requesting review.

nikhil-maryala:
✅ Codex run done, **with a caveat I want to state rather than bury**: [35268641387](https://github.com/UiPath/skills/actions/runs/35268641387), 2 requested / 2 graded —

| task | result | model |
|---|---|---|
| `skill-bpmn-debug-instance-inspect` | SUCCESS **1.0** (35s) | `gpt-5.6-luna` |
| `skill-bpmn-operate-diagnose-minimal-fault-triage` | SUCCESS **0.984** (46s) | `gpt-5.6-luna` |

**The caveat:** `agent=codex` on this workflow resolves to `vars.CODEX_MODEL`, which is **`gpt-5.6-luna`** (`run-coder-eval.yml:431,497`). The 09-17 nightly codex arm ran **`gpt-5.6-terra`**. So this is the codex *harness* on a different model than the run whose judge scored 0.000 — the specific regression you asked me to verify is still **not** reproduced-then-fixed on its own model. I'm not going to write the codex row as verified on that basis.

Two ways to close it properly, your call: dispatch with `agent_model=gpt-5.6-terra` (the input overrides the repo variable), or wait for tomorrow's nightly, which runs terra by default. I'd lean on the nightly since it's the exact configuration that failed, and this PR isn't claiming a causal fix anyway.

Incidentally that also explains the `parallel-fork-join` smoke failures I diagnosed on #3380: the smoke suite uses the same `CODEX_MODEL`, so the PascalCase `<bpmn:ParallelGateway>` was luna, not terra — consistent with both nightly models scoring that task 1.0 today.

Current evidence for this PR, stated plainly: 4 runs, 4 passes (claude `debug-instance-inspect` 1.0 and `minimal-fault-triage` 0.996; luna 1.0 and 0.984), n=1 per cell, one cell judge-scored, and the root cause still unexplained given both prompts already carried the rule.

nikhil-maryala:
Correcting my own suggestion: I floated *"wait for tomorrow's nightly"* for the codex row. That's circular — the nightly runs `main`, which doesn't contain this change, and this PR can't merge until it's approved. So the nightly can never produce branch evidence, and the explicit dispatch is the only way to answer your ask.

Dispatched with the override, since `run-coder-eval.yml:497` takes `AGENT_MODEL` ahead of `vars.CODEX_MODEL`: https://github.com/UiPath/skills/actions/runs/35281559419 — `agent=codex`, `agent_model=gpt-5.6-terra`, both tasks. If terra isn't deployed on the `CODEX_BASE_URL` resource this workflow uses (the nightly runs on its own VM config), it'll error rather than score, and I'll report that as the constraint instead of quietly retrying on luna.

nikhil-maryala:
_(superseded — this comment was mangled by a shell-expansion mistake on my side; see the corrected version below.)_

nikhil-maryala:
The terra run you asked for came back, and it **refutes this PR's remaining value for `minimal_fault_triage`**. Reporting it in full because it changes the conclusion.

(Reposting: my previous comment was mangled — I built it with a shell heredoc that expanded backticked spans into command output. Apologies for the noise; the content below is what I meant to say.)

[Run 35281559419](https://github.com/UiPath/skills/actions/runs/35281559419) — `agent=codex`, `agent_model=gpt-5.6-terra` (the override works: `run-coder-eval.yml:497` prefers `AGENT_MODEL` over `vars.CODEX_MODEL`), **on this branch**:

| task | result |
|---|---|
| `skill-bpmn-debug-instance-inspect` | SUCCESS **1.0** |
| `skill-bpmn-operate-diagnose-minimal-fault-triage` | **FAILURE 0.8 — judge 0.000, same as the nightly** |

So the fix does not fix it on the model that was failing. You were right to insist on this run; the claude and luna passes were not evidence for the codex row.

**And the reason is the real root cause, which is not in this PR's territory at all.** From terra's transcript:

- It read incidents, variables and the deployed asset via `uip maestro bpmn ... --output json`.
- It captured each response to `.local/investigations/raw/*.json` and read those back — the loop the rubric explicitly permits.
- It **never issued a command touching `mocks/` or `fixtures/`**. I checked every command's parameters: zero matches. Those strings appear in the run only inside the *output* of it reading `diagnose/CAPABILITY.md`, the file whose rule 3 mentions them.

What it did was `grep -rlF "<error string>"` over the installed skill tree to find the documentation for the error it was diagnosing. The judge called that *"a broad direct filesystem search outside the mocked CLI path"* and failed it — but fail condition (b) is *"any path under `mocks/` or `fixtures/`"*, which a skill-tree search is not. Your `uipath-troubleshoot` hypothesis is confirmed too: the agent loads it, and it directs exactly that lookup.

**So it is a grading bug — not agent behavior, and not a guidance gap.** Filed as #3395 with the rubric fix, verified on terra.

**Which leaves the question of what this PR is for.** Honest answer: only the correctness work that stands on its own —

1. the over-broad line that banned local `.bpmn` and generated-package-file reads (your major — real regardless of any eval failure);
2. rule 3 closing its category instead of listing three verbs;
3. the invented staleness mechanism removed.

No causal claim. The one task it plausibly helps, `debug-instance-inspect`, is a deterministic `command_not_executed` on a genuine `cat mocks/...` and it passed 1.0 in all four runs — unproven either way.

If you would rather this land as a pure docs-correctness PR on that basis, I will retitle it accordingly. If you would rather close it and keep only #3395, that is defensible too, and I am not going to argue for my own diff.

rockymadden (approved):
**Recommendation: 🟢 Approved** — all three inline findings and all five overall findings are addressed, the code is now correct, and the two remaining items are record accuracy, not behavior. Retitle before you merge (below).

### ✅ Cleared

| # | was | now |
|---|---|---|
| I1 major | "every piece of diagnostic evidence comes from a CLI read" banned local `.bpmn` + generated-package reads | `SKILL.md:297-299` scopes to runtime evidence and says local source and package files are read from disk as usual |
| I2 major | `cat`/`grep`/`head` while the criterion grades `(cat\|less\|head\|tail\|grep)` | `CAPABILITY.md:38-40` closes the category — "by any means", enumeration as examples, "no file-read or editor tool" |
| I3 minor | rationale claiming the commands "enforce staleness checks" | dropped |
| O1 major | placement asserted as root cause | body now states it is not established, with the prompt-precedence evidence; terra transcript settles it |
| O2 major | codex row was a claude run | terra dispatched, reported as a **failure**, conclusion changed rather than the evidence shaded |
| O3 minor | 7 restating lines vs #3152's 2-line pointer | 5 lines that point at rule 3; one canonical statement |
| O4 minor | #3287 cited as a proven lesson | withdrawn, with the closed-unmerged reason stated |
| O5 minor | harness-only paths in SKILL.md | `mocks/`/`fixtures/`/response-cache now only in the gated reference |

Two things in rule 3 are better than what I proposed: "such as" keeps the verb list as examples rather than a new closed list, and the added *"capturing a CLI response to your own file and reading that back is fine"* forecloses the overcorrection #3320 had to undo — an agent reading "by any means" without it could stop capturing and pipe full responses into context.

I re-checked the closed category against every bpmn diagnose fixture: `deployed_asset_drift`'s `fixtures/OrderRouting.bpmn` and `validate_fix_loop`'s `fixture/ExpenseApproval.bpmn` land at the **sandbox root** (a `template_dir` copies its contents, no `mount_point`), so neither read hits a `fixtures/` path. No new conflict.

### 🔎 Still standing

**1. 🟡 minor — the title now says the opposite of what the diff does.** *"move the CLI-evidence rule to the diagnose point of use"* — the diff removes the restatement from the point of use and closes the category in the reference. This repo's squash subject is the PR title (`7d445e7a8c` = `#3320`'s title verbatim), so it lands in history permanently. Second commit's subject is already right: *"scope the diagnose CLI rule and close the category in rule 3"*.

**2. 🟡 minor — `## Fix` and `## Verification` still describe the superseded diff.** Below a corrected root-cause section, `## Fix` block-quotes the old 7-line text, says *"7 added lines, no deletions"* (now 5 changed in `SKILL.md` + 4 in `CAPABILITY.md`), and cites *"the #3287 lesson"* that the section above it withdraws. `## Verification` still says the nightly will show *"whether the point-of-use placement moves debug-instance-inspect 0.9 → 1.0"*. The `## Problem` table's codex row also still attributes that failure to reading backing files, which your terra transcript refutes — annotate it with #3395 rather than leaving the refuted attribution as the stated problem. Commit message has one misattribution too: the ladder is `CAPABILITY.md` rule 1, not `failure-modes.md` (which carries the signs, `:87-89`).

**3. 🟡 minor — every posted score predates the rewrite.** Four runs, all on the 7-line text; the current head has never been run. The relevant one is `debug-instance-inspect`: its missed criterion is a deterministic `command_not_executed` on a real `cat mocks/responses/manifest.json`, and the `mocks/` paths that were in SKILL.md when it scored 1.0 now sit one link away. It is `integration`-tagged, so no gate depends on it and this does not block — but it is 50s and it is the only cell where the rewrite could plausibly cost something.

### 🧪 CI

35 pass, 4 skipped, 0 failed on `cc0f6e34f`. The one red check on that commit is your own terra dispatch ([35281559419](https://github.com/UiPath/skills/actions/runs/35281559419)) failing on `minimal_fault_triage` 0.8 — the grading bug in #3395, not a CI break. Worth a line in the body so nobody reads the ✗ as broken.

### tl;dr

Good outcome. The two majors are fixed, and the honest path you took on the codex row — running terra, reporting a failure, and tracing it to a rubric that fails a skill-tree grep under a clause scoped to `mocks/`/`fixtures/` — turned a PR with a wrong causal story into a correct docs fix plus #3395. What lands is: runtime evidence via the CLI, local source and package files read normally, one canonical statement of the mock boundary with the category closed and capture-then-read allowed. Retitle it, refresh `## Fix`/`## Verification`, and it's done.

PR #3390, opened 2026-09-17
fix(test): fix extract_alias_collision.yaml wrong fixture and assertions

charlesliu9:
### Test failure
skill-case-diagnose-extract-alias-collision
ERROR: Agent turn timed out after 1200s — and the agent made zero edits. The produced artifact is byte-identical to the fixture. 32 tool calls totalling 46s of execution out of 1200s; the other ~1150s was model generation (95.5K output tokens, 58 assistant turns against max_turns: 60). It read the case, spawned a research subagent, then spent 15 minutes grepping the references for a producer — and never wrote a line. So review.json's turn-timeout-too-short tag is a misdiagnosis: more budget just buys more thrashing.

### Why it never converged — the fixture has no producer at all
The task premise ([extract_alias_collision.yaml:3-12]) says the produced value "lands in route2 (Review's task element)". It does not. In the fixture:
- Every occurrence of route/route2 is a declaration in variables.inputOutputs[] plus the one gate expression =js:(vars.route === 'approved') — nothing else, anywhere in the project.
- All four tasks have outputs: null, and all are wait-for-timer placeholders. The agent established exactly this at command 30, and the skill's own doc says the timer task has no variables. Nothing in the case can write anything.
- This was true from the authoring commit 95c97c571 — not a later regression.
So route and route2 are both orphans. The only thing separating them is route2's elementId.

### The checker contradicts the skill it grades
[check_extract_alias_collision.py] defines "produced" as any declared variable whose elementId is non-empty and not "root" → {caseRef, route2}. That is the inverse of the documented shape:

- [io-binding/impl-json.md:140] — the -> extract companion in root.inputOutputs[] is "Required — elementId: "root", custom: true". That is exactly route's shape — the one the checker calls an orphan.
- elementId: "<stageId>-<taskId>" belongs on task.data.outputs[] / inputs[] ([impl-json.md:14,28]), not on a case-level inputOutputs[] entry. route2's shape is one the skill never emits.
- The subagent's research (msg 14) concluded correctly from the references that elementId on an inputOutputs[] entry is FE-picker display scope and is not read by the runtime resolver — i.e. it does not make a variable "produced".
- Command 31, grep -i "alias|task-scoped|companion.*task" across the references, returned nothing. The "extract alias auto-suffixed on an id collision" signature the description names is not documented as a task-scoped inputOutputs entry anywhere.

The agent was therefore correct to refuse vars.route2, and had no documented basis for the only answer the checker accepts (point the gate at route2, or caseRef).

### Fix
Test-design defect, not a skill defect. Two things need fixing, and they're independent:

- The fixture needs a real producer. Give the Review task (Stage_Qn7tBz-tQn7tBz4Wm) a data.outputs[] entry writing route2 (id/var/value: "route2", source: "=Route", target: "=route2", elementId: "Stage_Qn7tBz-tQn7tBz4Wm"), and move route2's inputOutputs[] companion to elementId: "root", custom: true. Then the collision story is true and diagnosable: route is declared-but-unwritten, route2 is written by Review.
- The checker's produced predicate must read task outputs[], not case-variable elementId. As written it inverts the skill's contract — and after fix (1) it would fail the corrected fixture, since the legit companion is elementId: "root".

### Verify
claude: https://github.com/UiPath/skills/actions/runs/35275946653
codex: https://github.com/UiPath/skills/actions/runs/35276962862
both run twice, all succeeded


BenHorblit-UiPath (approved):
Nice!

Yiqi517 (approved):
Seems reasonable!

PR #3391, opened 2026-09-17
fix(test): fix two failed claude tests

charlesliu9:
### Eval failure 1
skill-case-registry-handoff-action-case-sdd-only: One check fails out of five (score 0.619). check_registry_handoff.py:148:
> AssertionError: PortableReviewActionProbeQ91 resolved to nothing but carries no <UNRESOLVED> marker

The produced ledger is otherwise completely correct — right cacheFile per type, all three stale queries replaced from the current SDD, no display-name substitution, the resolved Action App bound with live id b20c471c-… + folder Shared, matches: []/selected: null on both probes. Registry pull ran, no inline builder, no caseplan. The agent knowingly took the placeholder path (its own rationale says "genuine empty lookup … used the normal placeholder path") — it just never wrote the marker string.

### Why: the marker has no home in the only file the agent read
The agent loaded the skill, then read just sdd.md and the stale ledger — 3 Reads total, zero reference files. Within SKILL.md alone the two rules conflict:

- Rule 8 ([SKILL.md:48]) — "Keep <UNRESOLVED: ...> in tasks/registry-resolved.json" — names no key.
- Rule 9 ([SKILL.md:49]) — "exact keys stage, task, taskType, cacheFile, searchQuery, matches, selected, rationale … selected is a match or null after a genuine empty lookup."

Rule 9's closed list has no identity slot, and it fully specifies the empty-lookup case as selected: null. Its escape hatch ("plus resolved I/O/review metadata") doesn't apply to an entry with nothing resolved. So SKILL.md alone yields a self-consistent reading where Rule 8 is satisfied. The actual slot appears only in files the agent never opened: [planning.md:153]and [placeholder-tasks.md:109] — the marker goes in taskTypeId / typeId / connectionId.

### Fix
Doc-side: state the identity slot in Rule 9 itself, so the requirement survives an agent that reads only SKILL.md. 

### Eval failure 2
skill-case-run-once-envelope failure
Cause: turn-budget exhaustion one Edit short of a valid plan — not a skill regression
Verdict: MAX_TURNS_EXHAUSTED (113 assistant turns, 67 tool calls, 495 s), score 0.647. Exactly one criterion failed:
[error] [nodes[root]] Case has no completion rules
CASE_MGMT_ROOT_CASE_EXIT_CONDITIONS_MISSING
The produced [caseplan.json] has no metadata.caseExitRules — the SDD § 1.4 row (required-stages-completed, Marks Case Complete Yes). Everything else is there: stage entry (case-entered), stage exit (required-tasks-completed), task entry rules, all 3 timer tasks.

The thing under test passed. Both shouldRunOnlyOnce checks are green (5/5 and advisory 2/2): Freeze Snapshot true, Reminder Ping false, Escalation Hold false. The regression this task guards is intact.

### Where it ran out
the agent's own todo 4 was "Stage entry (case-entered), stage exit (required-tasks-completed), 3x task entry (current-stage-entered), case exit (required-stages-completed)". Its validate at turn 105 told it the case completion rules were missing. It then authored the stage + task rules in the Edit at turn 112 of 113 and the budget cut it off before the caseExitRules edit and before todo 5 (final validate --strict --sdd). One more Edit would have been Valid.

### Contributing factor
max_turns: 60 is an outlier ([run_once_envelope.yaml:15]. Every other single_node/* case task is >=120. It has been 60 since the task was created in 23541dbe4. The description justifies the low budget with "hermetic — no registry resolution, no tenant resources, no debug run", but that isn't what consumes turns: the authoring volume (solution init + project scaffold + trigger + stage + 3 tasks + 4 condition families) is the same as its 120-turn siblings. The daily eval runs also marks the  "max-turns-too-low" tag.

### Fix
raise max_turns to 120 to match the sibling single-node tasks to give it more buffer to complete the process.

### Verify
Claude-sonnet-5: https://github.com/UiPath/skills/actions/runs/35276461406
Codex gpt-5.6-terra: https://github.com/UiPath/skills/actions/runs/35276423662
both succeeded on both tests


BenHorblit-UiPath (approved):
Seems reasonable

Yiqi517 (approved):
Hope it helps!

PR #3393, opened 2026-09-17
fix(tests): multiselect named two recipients the workspace does not have

rockymadden:
## What

`skill-flow-ipe-multiselect` asks for a Slack group DM between three people and grades on the `users` multiselect holding exactly 3 entries. Two of the three do not exist, so the criterion is unreachable regardless of agent or budget.

**Verified live** against `alpha.uipath.com` / `codereval`, connection `2bbc2253` (the one the failing run used), by paging the full `curated_users` directory:

| Prompt asked for | Live result |
|---|---|
| `Coder Evals` | **0 matches** |
| `IS-sandboxes@uipath.com` | resolves via `users_lookupByEmail` → `U04T8EXS56C` |
| `E2E-testing` | **0 matches** |

Every guessed email for the third (`e2e-testing@`, `e2etesting@`, `e2e.testing@`) returned `400 users_not_found`. The run resolved 2 of 3 and died on the turn timeout still hunting the third, scoring 0 on its fixture rather than on the multiselect feature.

## Change

Recipients are now `Coder Eval Test`, `IS-sandboxes@uipath.com`, `E2E Nightly Summary`.

| Recipient | Resolves via | Live check |
|---|---|---|
| `Coder Eval Test` | `curated_users` | `U0B7Y855WGG`, exact 1, substring 1 |
| `E2E Nightly Summary` | `curated_users` | `U05Q882RHFZ`, exact 1, substring 1 |
| `IS-sandboxes@uipath.com` | `users_lookupByEmail` | `U04T8EXS56C`, one call |

Two details that drove the choices:

- **`Coder Eval` is deliberately not used.** A substring lookup on it returns **2** rows live (`Coder Eval`, `Coder Eval Test`), and the reference rules then require the agent to stop and ask, which in a headless run is a stop. `Coder Eval Test` matches one. The comment records that trap, since it is what a future edit would walk back into.
- **`is-sandboxes` does not appear in `curated_users` at all.** The email is the only route to it, so the prompt keeps the email form rather than a display name.

`check_multiselect_flow.py` takes a count and pins no IDs, so the 3-entry assertion is untouched.

## Corrections to an earlier finding of mine

I previously reported `curated_users` pagination as broken (token never advances). That was wrong twice over, and I have corrected the guidance it produced in #3385.

- The identical pages came from `--query "pageToken=..."`, an undeclared param the connector silently ignores, so each call re-served page one. With `nextPage=` it pages fine.
- The **real** defect is different from what I then wrote: the token advances forever and `HasMore` never reaches `"false"`. Live, 12 pages returned 12000 rows holding **5984 distinct** users; pages 1-6 delivered 5980 of them and pages 7-10 added **zero**. So neither a repeated-token nor an identical-page guard ever fires, and "page until `HasMore` is false" never terminates. #3385 now names the only reliable stop: a page that adds no rows you have not already seen.

Worth a connector report separately. It does not block this task: both name-based recipients land on pages 1 and 4.

## Verification

- All identity claims above are live queries, not inferred from the captured run.
- YAML parses; `task_id` unchanged.
- `test_same_ground_corpus.py::test_multiselect_prompt_names_slack_as_the_required_connector` still satisfied.
- No forbidden `do not … debug` phrase, no UUID in the prompt.
- Criterion still asserts 3 entries.


nikhil-maryala (approved):
**Severity tally — 0 Blocker · 0 Critical · 0 Major · 0 Minor · 0 Nit.**

Correct, minimal fixture fix — the criterion was unreachable because two of the three named recipients don't exist in the workspace. Verified:

- **The swap is safe against the checker.** `check_multiselect_flow.py` asserts `len(users) == 3` (`check_flow(path, expected_count)`, line 100) and pins **no** recipient names or IDs — grep for coder/e2e/sandbox/nightly/UID in it is clean. So changing the three prompt names can't affect the 3-entry assertion, which still stands (criterion at :54, weight 3.0).
- **Fixture integrity holds:** YAML parses, `task_id` unchanged (`skill-flow-ipe-multiselect`), and the same-ground corpus predicate is still satisfied (the prompt keeps "Slack group direct message"). New recipients are the three your captured run proved resolvable (`Coder Eval`, `E2E Nightly Summary` via `curated_users`; `IS-sandboxes@` via `users_lookupByEmail`), and the comment correctly records "re-verify before changing" — the only symptom of this rot is a task that scores 0 forever, so documenting it is the right call.
- **Good self-correction:** noting your earlier `curated_users` "token never advances" finding was actually an undeclared `--query pageToken=` (ignored) vs the working `nextPage=`, already fixed in #3385 (which I reviewed). The separate real defect — `curated_users` never returns `HasMore:false` — is correctly scoped out as a connector report since both name-based recipients land in the first four pages.

Approving.

PR #3399, opened 2026-09-18
fix(uipath-automationhub): never block publish on owner lookup, and create missing applications

parth-patil-uipath:
Two real publish failures on staging, both of which stopped a PDD from being published when nothing was actually wrong with the tenant or the user. Both causes are in this skill's guidance, not in the product.

## 1. The owner lookup blocked publishes it had no business blocking

The skill told the agent that `GET /users` is *"the ground truth"* for resolving an owner. It isn't, for two reasons measured live:

- **`invite` defaults to activated users only.** A user who has only ever used the API or CLI sits at `NOT_INVITED` and is **absent from the list while being perfectly able to own a process**. On a test tenant the default returned **17 of 24** users.
- **Results are paged under a non-deterministic order.** The default sort is `relevance`, which is a constant when no search text is given, with no tiebreaker — so page membership is unspecified.

What happened: the agent ran `uip ah users list --limit 50`, filtered for the signed-in user's own email, got nothing, and reported *"Automation Hub does not list your account as an eligible user."* It never attempted the create. The user then emailed an admin asking for access **they already had** — they were a System Admin.

**Changed:** the signed-in identity comes from `uip ah auth-info get` (`Data.User.Email`, plus `IsActive` and `Roles`), which is the authority. The users-list call is kept but is explicitly **informational and never blocking**, and when used must be scoped:

```bash
uip ah users list --search "<owner-email>" --invite-status all
```

Both flags are required — verified: `--search` alone returns **0 rows** for a not-invited user, because search and the invite filter are independent `WHERE` clauses. With both, an exact email returns `Total=1` server-side, so the page size stops mattering.

A miss is now **no signal**: submit anyway and let the API decide. Only a real `Cannot identify owner by email` from the create is an owner problem.

## 2. Missing applications were treated as unfixable, so publishes were abandoned

A PDD named Salesforce, Equifax, CREDILEX, SUNAT Portal and others; none were in the tenant inventory, while "Applications used" / "Thin applications used" were **required** with enums limited to existing applications. The agent concluded it had no honest option and stopped:

> *"Automation Hub still requires valid application IDs and provides no generic, unknown, or 'not listed' option. I did not publish the PDD because selecting unrelated applications would misrepresent the process."*

Refusing to fabricate was right; giving up was not. **Applications can be created** — `PUT /appinventory` upserts, and an element whose `application_id` is `null` inserts. Verified end to end through the CLI:

```bash
uip ah applications update --file ./new-apps.json --output json
# [{ "application_id": null, "application_name": "SUNAT Portal",
#    "application_version": "1.0", "application_language": "English",
#    "categoryIds": [1] }]
#
# admin     → Success, created Id=12
# non-admin → 403 "This user is not permitted to perform this action based on their role."
```

**Changed:** match the inventory, create what's missing, and on a `403` (ordinary roles lack `MANAGE_APP_INVENTORY`) fall through to picking the closest existing entries **and naming the real systems in the description**, so the truth survives in the record. Explicitly: never abandon a publish over applications, and never pass off a substitution as the real thing without saying so.

Note there is no `applications create` verb — creation goes through `update`. The CLI's own help says it *"cannot add one"*, which is wrong and is very likely why the agent gave up; that text is fixed separately in `UiPath/cli`.

## 3. The owner 400 now has a remediation

`Cannot identify owner by email` named the email address, which sent people hunting for a typo. It actually means the account has never been activated on that tenant. Both guides now say: don't retry with a different address, confirm the identity with `auth-info`, and point the user at a one-time browser sign-in — noting that tenants carrying the RPANAV-19110 backend fix accept these users with no sign-in at all.

## Files

| File | Change |
|---|---|
| `references/publish-process-cli-guide.md` | owner via `auth-info`; non-blocking scoped lookup; application-create recipe with the 403 fallback; owner-400 handler |
| `references/publish-process.md` | same for the raw-API path (`?s=<email>&invite=all`, `PUT /appinventory`) |
| `references/api-endpoints.md` | `GET /users` no longer described as ground truth; both under-reporting defaults documented |
| `references/cli-commands.md` | `auth-info get`, the scoped user lookup, and `applications update` as the create path |

Docs-only — no executable code in this skill.

## Verification

Every command in the new guidance was run against live tenants (alpha `testsorin/parthTestSept26`, staging `devcon25/Default`), not inferred:

- `--search <exact-email> --invite-status all` → `Total=1` for `invite=0` and `invite=3` alike; `0` rows without the invite flag
- Substring behaviour confirmed (`gmail.com` → 3 rows), which is why the guidance says to use the `auth-info` email verbatim rather than substituting a matched row
- `applications update` with `application_id: null` → created Id 12 as admin, clean 403 as non-admin
- Business Process + PDD published end to end by a `NOT_INVITED` user on both tenants

This needs cherry-picking to `release/v1.202` for the hotfix.


parth-patil-uipath:
Updated after review (`caebf483`), two changes:

**1. Fall back on *any* application-create failure, not just 403.** The earlier wording singled out the permission error, which would have left an agent stuck on a validation error or a bad category id — the same dead end, different cause. Now: *"If creating fails for any reason — that 403, a validation error, a bad category id, anything — fall through; never retry it and never stop."* Applied in both guides, and the input-table row no longer says "refused" (which read as 403-only).

**2. Removed the internal ticket reference.** The owner-400 handler cited "the RPANAV-19110 fix" by ticket number. This skill runs on customer machines, so an internal Jira ID is meaningless there and shouldn't ship. Replaced with *"Newer Automation Hub versions accept these users with no sign-in at all, so on an up-to-date tenant this error should not appear."*

Also reworded *"Both flags or neither"* on the user lookup — it could be read as "omitting both is fine", when the point is that `--search` without `--invite-status all` is actively misleading. Now: *"Pass both or the lookup is worse than useless — search alone reports 'no such user' for exactly the users this guidance is about."*

## One thing a reviewer should weigh deliberately

This **relaxes a guardrail on purpose.** The previous text said *"never record an application the material does not support"*, and the abandoned publish it produced was that rule working as designed. The new guidance permits substituting the closest inventory entry when creation isn't possible — on the reasoning that the process record plus PDD is the valuable artifact and applications are editable afterwards, whereas a blocked publish loses everything.

The honesty requirement is preserved but moved: the substitution must be stated, and the real systems recorded in `OVERVIEW_DESCRIPTION`. If you'd rather keep the hard rule and accept the blocked publishes, that's a legitimate call and this is the paragraph to push back on.

abhiram-vad:
Minor nits:
1. We don't have to remove uip ah users list completely from cli reference since it's just a reference for cli. This is not urgent but can be used when AH skills are expanded later. 
2. The skill explains user invite status for cli and api created users, which isn't required. Just using uip ah list user --search is fine since we are doing a targetted search. 
3. It's better to test and expand on other fields in the BA process schema, not only applications to avoid bandaid fixes later.

abhiram-vad (approved):

parth-patil-uipath:
Both findings addressed in `ab6f7b18`.

## Duplicate `auth-info get` row — real miss, fixed

I added a row without checking the table already had one, so `uip ah auth-info get` appeared at lines 42 and 46. Consolidated into the original position rather than keeping mine, so the table order is unchanged:

```
| `uip ah auth-info get` | connectivity + who/where am I; the signed-in identity is the
  **default owner/submitter** and the authority for it | `Data.Tenant.Url`,
  `Data.User.Email`, `IsActive`, `Roles` |
```

Now exactly one row, carrying both purposes and the fuller data shape.

## `categoryIds: [1]` — taken, with a correction to my own first attempt

Worth recording because I got it wrong initially. My first fix put the clarification *inside* the JSON heredoc as a `//` comment — but that block is written to a file and passed to `--file`, and JSON has no comments, so an agent copying it verbatim would have hit a parse error. I'd have replaced a cosmetic confusion with a real breakage.

Moved into the prose instead:

> All five fields are required by the request schema, and `categoryIds` needs at least one **real** category id from `uip ah categories get` — the `1` below is a placeholder for a real id, not a literal to send (same rule as `OVERVIEW_CATEGORY`).

Then validated every `json` fenced block in both edited guides. My application-create example parses; the only non-parsing block is the pre-existing document-upload **template** at `publish-process.md:132` (`"document_type_id": <int>`), which is deliberately a shape sketch and untouched by this PR.

## On the missing e2e task

Agreed and noted as pre-existing. Worth saying plainly that an e2e task would not have caught either bug this PR fixes: both were wrong *guidance* on tenants where every command succeeded. The users-list call returned `200` with a short list, and the application question was validly rejected. Catching these needs a tenant fixture with a `NOT_INVITED` user and an inventory that lacks the document's systems — which is a more specific fixture than a generic e2e tier, and worth its own issue rather than being folded in here.

parth-patil-uipath:
Reverted the `categoryIds` wording in `deae0af3` — back to the original sentence.

Reasoning: the reviewer rated it optional, the JSON example was never broken, and it was the one hunk I'd had to correct mid-flight. Not worth carrying a sentence that fixes nothing real on an already-approved PR. The duplicate `auth-info` row consolidation stays, since that was a genuine defect.

Final diff is the same four files, +73/−8, and every change in it addresses a real failure.

parth-patil-uipath:
Thanks @abhiram-vad — 1 and 2 taken in `f908e91d`.

**1. Plain `uip ah users list` row is back** in `cli-commands.md`, alongside the scoped owner-lookup row:

```
| `uip ah users list` | list tenant users | entries carry `Email`, `IsActive`, `InviteStatus`; paged (default 20); filters: `--search`, `--invite-status`, `--limit` |
| `uip ah users list --search <email> --invite-status all` | optional owner lookup | both flags needed; a miss is **no signal** — never block on it |
```

**2. Invite-status explanation trimmed** in both guides (153 → 96 words in the CLI guide). The mechanics narrative about CLI/API-created users is gone; what stays is the instruction only: use `--search <email> --invite-status all`, both flags, a miss is no signal, never block. I kept the `--invite-status all` flag rather than `--search` alone because that one is load-bearing — `--search` on its own returns zero rows for a not-yet-invited user, which is exactly the false negative that produced the blocked publish this PR is fixing.

**3. Other BA schema fields — agreed, follow-up rather than this PR.** While verifying this I did hit one concrete divergence worth carrying into that work: the flow schema's required set differs per tenant (e.g. `COUNT_APPS` is required on one Business Process flow and rejected on another), so the expansion should be driven by `automations schema get` per tenant rather than a fixed field list. I'll open a separate issue with the required-field data I collected from the two tenants so it isn't a bandaid.

PR #3400, opened 2026-09-18
fix(uipath-automationhub): never block publish on owner lookup, and create missing applications [cp v1.202]

parth-patil-uipath:
Cherry-pick of **#3399** onto `release/v1.202` for the hotfix. Picked with `-x`, so both commits record their origin.

Two real publish failures on staging where nothing was wrong with the tenant or the user — both caused by this skill's guidance rather than the product.

## 1. The owner lookup blocked publishes it had no business blocking

The skill called `GET /users` *"the ground truth"* for resolving an owner. It isn't, for two measured reasons:

- **`invite` defaults to activated users only** — a user who has only ever used the API or CLI is absent from the list *while being perfectly able to own a process*. On a test tenant the default returned **17 of 24** users.
- **Results are paged under a non-deterministic order** — the default sort is `relevance`, a constant when no search text is given, with no tiebreaker.

What happened: the agent ran `uip ah users list --limit 50`, filtered for the signed-in user's own email, got nothing, reported *"Automation Hub does not list your account as an eligible user"*, and never attempted the create. The user then emailed an admin for access **they already had** — they were a System Admin.

Now: the identity comes from `uip ah auth-info get`, which is the authority. The users-list call is retained but **informational and never blocking**, and when used must be scoped `--search "<email>" --invite-status all` — both flags, since search alone returns zero rows for exactly the users this is about.

## 2. Missing applications caused publishes to be abandoned

A PDD named Salesforce, Equifax, CREDILEX, SUNAT Portal and others; none in the tenant inventory, while the application questions were **required** with enums limited to existing entries. The agent refused to fabricate selections and stopped.

Refusing to fabricate was right; stopping was not. **Applications can be created** — `PUT /appinventory` upserts, and an element whose `application_id` is `null` inserts. Verified through the CLI: admin → created Id 12; non-admin → `403`.

Now: match inventory → create what's missing → **if creating fails for any reason, fall through** to the closest existing entries and name the real systems in the description. Never abandon a publish over applications; never substitute silently.

## 3. The owner 400 has a remediation

`Cannot identify owner by email` named the email address, sending people hunting for a typo when the account simply had never been activated. Both guides now say: don't retry with a different address, confirm via `auth-info`, point at a one-time sign-in, and note that newer Automation Hub versions accept these users with no sign-in at all.

## Cherry-pick quality

- **Zero conflicts.** All four files verified identical between `release/v1.202` and `main` before picking.
- **Content identical to the `main` branch** — `git diff origin/fix/ah-owner-lookup-and-application-create -- <the 4 files>` returns empty.
- Docs only; no executable code in this skill.

## Merge order

**Land #3399 on `main` first.** Merging this ahead of it puts the fix on the release line while trunk lacks it — the exact divergence that bit us on a backport earlier today.

Pairs with `UiPath/cli` **#4321** (and its own v1.202 backport), which corrects the `applications update --help` text that told agents creation was impossible.


abhiram-vad (approved):

parth-patil-uipath:
Topped up with the three review-fix commits that landed on #3399 after this backport was cut (`ab6f7b18` auth-info row consolidation, `deae0af3` categoryIds revert, `f908e91d` Abhiram's nits — plain `users list` row kept, invite-status note trimmed).

`git diff origin/main -- skills/uipath-automationhub` is now empty: this branch's skill directory is byte-identical to `main` at the #3399 squash (`a48228db`).

parth-patil-uipath:
Added one more commit on top of the mirror (`dc8516a9`, +13/−1, docs only): a **preflight step before `automations create`** — check the answers file against the tenant's own schema (required questions answered, every enum code copied verbatim from that question's `enum`, no template placeholders). Read-only and local; a correct payload produces the identical `create` call.

Motivation: a Delegate run today on `maestrolab/DefaultTenant` failed three creates with `Please fill in all the required information (400)` because the agent wrote `ah-answer_option-ovrbp-0-3-5` where the schema's codes are `…-0-3-0-N`. Reproduced both ways on a test tenant: same malformation → same 400; correct code → success. This is the concrete case behind @abhiram-vad's nit 3 (derive from the tenant's schema, not a fixed list — `COUNT_APPS` is required on one flow and rejected on another).

This puts the v1.202 skill 13 lines ahead of `main`; I'll cherry-pick the same commit to `main` as a separate PR once this merges.

PR #3407, opened 2026-09-18
fix(tests): fund validate_flow.py criteria at its 180s budget, guard it

rockymadden:
## Why

`uip maestro flow validate` refetches the tenant node manifest on every call: 8-14s typical, 60-67s on the tail (#2853 measured). `_shared/validate_flow.py` budgets itself at 180s (`_DEFAULT_BUDGET_SECONDS`) because coder_eval exports no criterion timeout. A criterion granting less is SIGKILLed on a tail hit: exit -1, empty stdout, score 0 on a flow that validates clean.

Seen 2026-09-18 on `skill-flow-ipe-dtl-load-by-default-false`: the grader's validate was killed at 30s. The agent had validated the same 21KB flow in 3.8s and 4.4s minutes earlier.

#2853 raised 54 criteria to 180 on 2026-08-27. #2756 added validate criteria to the IPE connector_features tasks at 30 on 2026-09-04. Before this PR: 35 criteria at 30s, 11 at 60s.

## What

- 46 yamls: `timeout: 30` or `60` becomes `timeout: 180` on the `validate_flow.py` criterion only. 36 under `uipath-maestro-flow`, 10 under `uipath-human-in-the-loop` (its `reference.directory` is `../uipath-maestro-flow`, same script).
- `_shared/test_criterion_budgets.py`: new guard. Every run_command criterion under `tests/tasks/` whose script resolves to `validate_flow.py` must grant at least `validate_flow._DEFAULT_BUDGET_SECONDS`. A second, independent count asserts none went undiscovered.

Every task still satisfies the existing `task_timeout > sum(criterion timeouts)` guard. Tightest after the bump: the four `e2e/jira_*` tasks and `cli_dice_roller_simulated`, at 1260-1530s worst-case grading against 1800-2700s task_timeout.

## Verification

- `pytest tests/tasks/uipath-maestro-flow/`: 1359 passed, the same invocation as CI's `pytest-flow-check`.
- Mutation check: one yaml set back to 30, the guard fails naming the file and the 180s floor.
- No coder-eval task was rerun. Grading timeouts only; no prompt, checker, or expected outcome changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3412, opened 2026-09-18
fix(tests): complex_array names two recipients the workspace does not have

rockymadden:
## Why

`skill-flow-ipe-complex-array` asks for a Slack group DM between the users "coder evals" and "e2e-testing". Neither exists in the workspace behind the `Shared/uipath-maestro-flow` connection.

Verified live on 2026-09-18 against connection `2bbc2253`: 19 pages, 18,264 rows of `curated_users`, case-insensitive over display name, real name, handle and email. Every match for "coder eval", "e2e" or "is-sandbox":

| User | ID |
|---|---|
| Coder Eval | `U0B0EPZB4B0` |
| Coder Eval Test | `U0B7Y855WGG` |
| E2E Nightly Summary | `U05Q882RHFZ` |
| E2E Nightly Run Summary | `U05RC1RP15E` |
| DU FE E2E CD Runner, TestManager E2E Test Results, Weekly E2E Owner | not candidates |

The 2026-09-18 run followed the reference rules (`connector/impl.md`: never write a display name as an ID, stop when nothing resolves), left `users` empty, and the populated check failed. The task scored 0 on its fixture, not on the complex-array feature it tests. Same defect #3393 fixed for `multiselect.yaml` yesterday.

## What

One file, `connector_features/complex_array.yaml`:

- Recipients become "Coder Eval Test" and "E2E Nightly Summary". Each is a unique substring across every name observed. "Coder Eval" is avoided on purpose: it is a prefix of "Coder Eval Test", so a substring lookup returns two rows and the agent must stop and ask, which in a headless run is a stop. A comment records the trap, as in #3393.
- The advisory criterion looked for the display names in the flow. A correct build writes user IDs, never names, so it could not pass. It now asserts the two IDs.
- The populated check, weights, thresholds and the headless preamble are unchanged. The validate criterion's timeout line is untouched; #3407 owns it.

## Verification

- PyYAML parse OK; `pytest tests/tasks/uipath-maestro-flow/` 1253 passed (corpus, preamble and budget guards).
- Recipient uniqueness verified live as above.
- No coder-eval rerun of the task in this PR.


tests/tasks/uipath-maestro-flow/connector_features/complex_array.yaml
```diff
@@ -54,8 +58,8 @@ success_criteria:
     pass_threshold: 1.0
 
   - type: run_command
-    description: "Advisory: group DM participants referenced in the project (names may resolve to Slack user IDs on a live connection)"
-    command: "python3 $REFERENCE_DIR/_shared/flow_contains.py --flow-name ComplexArrayTest 'coder evals' 'e2e-testing'"
+    description: "Advisory: both recipients resolved to their Slack user IDs (Coder Eval Test, E2E Nightly Summary) rather than written as display names"
+    command: "python3 $REFERENCE_DIR/_shared/flow_contains.py --flow-name ComplexArrayTest 'U0B7Y855WGG' 'U05Q882RHFZ'"
```
nikhil-maryala, line 62: [NIT] The two hardcoded Slack user IDs here carry the same rot risk the comment above only calls out for the *names*: if `Coder Eval Test` / `E2E Nightly Summary` are recreated, these IDs go stale silently. It is advisory (`pass_threshold: 0`), so a stale ID never fails the run — but it also stops recording whether resolution happened, with no signal. Worth extending the "re-verify before changing a name" note to cover the IDs too (or resolving them live at check time).

nikhil-maryala (approved):
**Severity tally — 0 Blocker · 0 Critical · 0 Major · 0 Minor · 1 Nit.**

Correct, same shape as #3393. Verified the gating does not depend on the specific IDs: the graders are `flow_contains` (nodes/edges, w2 pt1), `check_multiselect_flow.py populated` (w3 pt1 — users array non-empty), and `validate_flow.py` (w2 pt1). The recipient-name swap is what lets a headless agent actually resolve and populate that array; the two-ID `flow_contains` check is **advisory** (`pass_threshold: 0`), so it records real resolution without gating on workspace-specific IDs. Good call avoiding `Coder Eval` (a prefix of `Coder Eval Test` → a substring lookup returns two rows → the agent must stop and ask) and documenting why.

One NIT inline on the ID rot caveat.

### Cross-PR note (not blocking this PR)
This PR’s rationale — `Coder Eval` is prefix-ambiguous — applies to **#3393** (merged), whose `multiselect.yaml` I approved using `Coder Eval` as a recipient. If a substring lookup there also returns `Coder Eval` + `Coder Eval Test`, that task can hit the same stop-and-ask → empty-multiselect failure this PR fixes. Worth a quick check whether `multiselect.yaml` needs the same swap to `Coder Eval Test`.

PR #253, opened 2026-04-15
chore: add E2E edit scenarios and multi-step pipeline tests

akshaylive:
Edit scenarios (using shared BellevueWeather template):
- update_node: change end node output values
- add_node: insert node between existing nodes
- remove_node: delete node and rewire edges
- add_output: add field to node output mappings
- move_node: reorder nodes, merge branches, reduce end nodes
- group_to_subflow: extract nodes into a subflow

Multi-step pipeline tests:
- multi_city_weather: Loop → HTTP → Script across 3 cities
- slack_weather_pipeline: Slack connector → Script → HTTP → Decision

Skill doc fixes:
- http/impl.md: replaced hardcoded model with registry placeholder, added
  critical warning about missing expansion causing null outputs at runtime
- connector/impl.md: documented that definitions must include the form field
  from registry get for node configure to work

image

PR #3416, opened 2026-09-18
feat(tests): claude-arm experiment defaults, guarded against drift

nikhil-maryala:
Implements #3376 (the follow-up @rockymadden proposed reviewing #3369: *"a `nightly-claude.yaml` (or per-agent `run_limits`) defaulting to 1800/2400, so the next slow task is a config line rather than a third PR of this shape"*).

## Why now

The per-task budget treadmill produced its 4th and 5th victims in today's (09-18) claude nightly:

- `multi-city-weather` — ERROR at **905s** against the **inherited 900s** default (no per-task override; exactly the class this fixes)
- `smart-triage` — ERROR at **1805s** against its explicit 1800s cap, after 1.0 on both 09-16 and 09-17

Evidence base for the sizing is unchanged from #3369/#3376: the claude-code session streams as one turn (so `turn_timeout` is its wall budget), tool time is 16–50s of the 900–1800s walls, and tasks pass with 3–37s margins under codex-sized caps.

## What

| Piece | Detail |
|---|---|
| `tests/experiments/nightly-claude.yaml` | `nightly.yaml` with **one** delta: `run_limits` → turn 1800 / task 2400 (task ≥ turn, since task_timeout silently caps the turn — `orchestrator.py` arms it around agent + grading, per your #3369 review) |
| `tests/scripts/test_nightly_claude_experiment_sync.py` | The anti-rot guard: deep-equal outside the delta, intended values pinned on both files, task ≥ turn pinned on both. **Verified it fails on a stray divergence** (flipped `permission_mode` in the copy → guard fails), not just passes when happy. Runs in the existing `pytest tests/` CI. |
| `run-coder-eval.yml:480` | `agent=claude` → `-e experiments/nightly-claude.yaml`, everything else unchanged. `$AGENT` is the env-quoted dispatch input, same pattern the script already uses. |
| `patterns/smart_triage/greenfield.yaml` | 1800/2400 → **2700/3600** — today's other victim has explicit limits that outrank any default, and this matches its sibling heavy pattern tasks from #3369 (`external_wait`, `high_volume_batch`, `approval_chain`). |
| `tests/README.md` | The experiments table names the file and the edit rule. |

## Deliberately duplicated, deliberately guarded

A second experiment file is the same rot-risk shape as `skills.sh.json` — nothing else reads both. That's why the guard exists and why its docstring says *edit `nightly.yaml` and let the guard tell you what to mirror*. I considered raising the shared default instead (no duplication), but that changes the codex and antigravity arms' failure-detection latency without evidence they need it, and #3376/your #3369 review both scoped this to the claude arm.

## The half I can't land from this repo

The VM nightly (`coder_eval_uipath` `daily.sh`) also passes `-e experiments/nightly.yaml`. It needs the same one-line per-arm selection — requested on #3376 with the exact line. Until that lands, the VM claude arm keeps today's behavior (shared defaults) and loses nothing; GH-dispatched claude runs get the new defaults immediately, which is also how this PR is verified.

## Verification

- Sync guard: 4/4, plus the negative (stray edit → caught).
- `test_required_checks_contract.py`: 116/116 (workflow edited, required-check contract intact).
- Claude run on this branch against today's two victims — link in a comment. The dispatch itself exercises the new selection branch, since `workflow_dispatch` runs this ref's workflow file.


nikhil-maryala:
Verification run (this branch, agent=claude, today's two victims — the dispatch itself takes the new `nightly-claude.yaml` selection path): https://github.com/UiPath/skills/actions/runs/35377825228

nikhil-maryala:
✅ Verification run [35377825228](https://github.com/UiPath/skills/actions/runs/35377825228) (agent=claude, this branch, 2 requested / 2 graded) — and the durations are the proof, not just the scores:

| task | today's nightly (old budgets) | this branch | reading |
|---|---|---|---|
| `multi-city-weather` | ERROR, killed at **905s** (inherited 900) | **SUCCESS 1.0 in 1274s** (150 turns) | needed 374s more than the old cap allowed — and since this task has no per-task `run_limits`, finishing at all proves the workflow actually selected `nightly-claude.yaml` (1274s also exceeds the old 1200s task_timeout) |
| `smart-triage` | ERROR, killed at **1805s** (explicit 1800) | **SUCCESS 1.0 in 2605s** (141 turns) | needed 800s more |

So both of today's victims convert to passes, and the selection path is exercised end-to-end rather than assumed.

One honest caveat to log: `smart-triage` at 2605s against its new 2700s cap is a **95s margin** — the same riding-the-cap class this whole effort keeps finding. I've matched the sibling budget rather than special-casing higher, but if it flaps again the next step is the #3228-style decomposition question, not a fourth bump.

tests/scripts/test_nightly_claude_experiment_sync.py
```diff
@@ -0,0 +1,60 @@
+"""nightly-claude.yaml is nightly.yaml with exactly one delta: run_limits.
+
+The two files must not drift anywhere else — nightly.yaml carries load-bearing
+config (the cmd.exe-safe pre_run polyglot, docker env passthrough, the judge
+route) that would fail the claude arm silently if a copy went stale. Nothing
+else in the build reads both files, so this guard is the only thing that
```
rockymadden, line 6: 🔴 **critical** - this guard never runs in CI, so the duplication ships unguarded.

Nothing in CI runs `pytest tests/` or `pytest tests/scripts/`. Every guard under `tests/scripts/` gets an explicit job in `.github/workflows/test-helpers.yml`, file by file, and the workflow states the rule at `test-helpers.yml:177`: *"One job for every BPMN contract guard, listed explicitly: a glob that stops matching is a silently green job"*. `pytest tests/scripts/` appears only at `tests/README.md:715`, as a local command.

`gh pr checks 3416` confirms it: no job collects this file. So:

- ❗ The PR description's "Runs in the existing `pytest tests/` CI" is wrong.
- ❗ A stale `nightly-claude.yaml` reaches the claude arm with nothing failing, which is the exact failure this file exists to prevent.

**Fix**, smallest form: append it to the job at `test-helpers.yml:278`. That job is already called "task/experiment gate unit tests" and already installs `pyyaml`.

```yaml
run: pytest tests/scripts/test_check_task_driver.py tests/scripts/test_nightly_claude_experiment_sync.py -v
```

No new job, so no `docs/REQUIRED-CHECKS.md` row and no ruleset edit.

Partial mitigation worth knowing: `task-driver-gate.yml:53` passes `tests/experiments`, so the `:;` polyglot rule *is* enforced on the copy. The docker env passthrough, the mounts, the judge route and `post_run` are not.

tests/experiments/nightly-claude.yaml
```diff
@@ -0,0 +1,86 @@
+experiment_id: skill-tests-nightly-claude
+description: "Nightly cron config for the claude-code arm — nightly.yaml with claude-sized run_limits"
+
+# This file is nightly.yaml with ONE delta: the default run_limits. The
+# claude-code harness streams the whole session as one turn, so turn_timeout is
+# its effective wall budget — and the shared 900/1200 default is sized for the
+# codex arm (#3376: five tasks in two days died at a cap the same agent clears
+# with a bigger budget, e.g. ai_decision_review at 1223s vs 900, multi-city-
+# weather killed at 905s). Everything else MUST stay identical to nightly.yaml;
```
rockymadden: 🟠 **major** - `multi-city-weather` never inherited the 900s default, so it does not support the 1800s turn budget.

`tests/tasks/uipath-maestro-flow/multi_node/multi_city_weather/multi_city_weather.yaml:14` pins `turn_timeout: 1200`, and has since #253 (2026-04-16). `run_limits` merges field by field with the task layer above experiment-defaults (`coder_eval/orchestration/config_merge.py`), so for this task:

- its turn cap on 09-18 was 1200, not 900
- the new 1800s default cannot reach it, its own 1200 still wins
- what this branch actually changes for it is the inherited `task_timeout`, 1200 → 2400

Two consequences:

- ❗ A 905s ERROR matches neither its 1200s turn cap nor its 1200s inherited task cap. Either the number belongs to another task, or that failure was not a budget kill and will recur under the new budgets. Please post the run record.
- ❗ The verification comment's *"this task has no per-task `run_limits`, finishing at all proves the workflow selected `nightly-claude.yaml`"* rests on the same wrong premise. The conclusion survives by a different route (1274s exceeds the old 1200s `task_timeout`), and since the proof is the point of that run, it is worth restating correctly.

**Fix**: drop `multi-city-weather` from the sizing evidence here and in the PR body, or swap in a task that genuinely inherits the default. `ai_decision_review` (910s and 904s kills against 900, cited already) is the clean case.

.github/workflows/run-coder-eval.yml
```diff
@@ -477,7 +477,17 @@ jobs:
...
           # word splitting, no pathname expansion. That is what the delegate `-D`
           # override needs: to bash its `[...]` list is a character class.
-          args=(-e experiments/nightly.yaml)
+          #
+          # The claude arm gets its own experiment file: identical to
+          # nightly.yaml except run_limits (the claude-code session streams as
+          # one turn, so turn_timeout is its wall budget — #3376). A sync guard
+          # (tests/scripts/test_nightly_claude_experiment_sync.py) pins the two
+          # files identical outside that delta.
+          if [ "$AGENT" = "claude" ]; then
```
rockymadden: 🟡 **minor** - `claude` is matched positively here and used as the fallback 119 lines down.

Model selection at `:605` treats claude as the `else` of codex / antigravity / delegate-sdk. This block names it. Today they agree, because `agent` is a `choice` input. They stop agreeing the moment `$AGENT` is empty, which this same step already anticipates at `:458` (*"Guard the empty case (e.g. a future non-dispatch trigger)"*): the run would take the claude model on `nightly.yaml` budgets, silently, which is the bug this PR exists to kill.

**Fix**: pick one shape. Cheapest is to mirror the chain below.

```bash
if [ "$AGENT" = "codex" ] || [ "$AGENT" = "antigravity" ] || [ "$AGENT" = "delegate-sdk" ]; then
  args=(-e experiments/nightly.yaml)
else
  args=(-e experiments/nightly-claude.yaml)
fi
```

tests/scripts/test_nightly_claude_experiment_sync.py
```diff
@@ -0,0 +1,60 @@
...
+def test_everything_else_is_identical() -> None:
+    nightly = _load("nightly.yaml")
+    claude = _load("nightly-claude.yaml")
+
+    for config in (nightly, claude):
+        config.pop("experiment_id")
+        config.pop("description")
+        config["defaults"].pop("run_limits")
+
+    assert copy.deepcopy(nightly) == copy.deepcopy(claude), (
```
rockymadden: 🟡 **minor** - the `copy.deepcopy` calls are dead.

Both dicts are freshly parsed locals, already mutated by the pops above, and neither is used afterwards. Deep-copying them changes nothing about the comparison.

**Fix**: `assert nightly == claude, (...)`, and drop `import copy`.

tests/scripts/test_nightly_claude_experiment_sync.py
```diff
@@ -0,0 +1,60 @@
...
+    assert _load("nightly.yaml")["defaults"]["run_limits"] == SHARED_RUN_LIMITS
+    assert _load("nightly-claude.yaml")["defaults"]["run_limits"] == CLAUDE_RUN_LIMITS
+
+
+def test_task_timeout_covers_the_turn_in_both() -> None:
+    # task_timeout silently caps the turn (orchestrator arms it around agent +
+    # grading), so task >= turn must hold in every experiment default.
+    for name in ("nightly.yaml", "nightly-claude.yaml"):
+        limits = _load(name)["defaults"]["run_limits"]
+        assert limits["task_timeout"] >= limits["turn_timeout"], name
```
rockymadden, line 42: 🟡 **minor** - the `task >= turn` invariant is pinned at the one layer where it cannot break.

Both defaults are authored in these two files, so this test only fails if someone edits them. The layer where the invariant actually breaks is the merge: `tests/tasks/uipath-maestro-flow/ixp/integration_handle_routing.yaml` pins `task_timeout: 900` with no `turn_timeout`, so its resolved limits under the claude defaults are turn 1800 over task 900.

Nothing breaks today (the task cap binds at 900 either way, same as now), but the guard reads as though it covers the resolved config and it does not.

**Fix**: either narrow the name and comment to the defaults layer, or resolve every task against both experiment defaults and assert the invariant on the merged result.

rockymadden (changes requested):
**🔴 Changes requested**

The approach is sound. Two things block it: the drift guard never executes in CI, so the duplication ships unguarded, and the `multi-city-weather` evidence does not match the task file.

### 📦 What

- ➕ `tests/experiments/nightly-claude.yaml`: `nightly.yaml` with `run_limits` at 1800/2400. Confirmed by diff, the header and `run_limits` are the only deltas.
- 🔀 `run-coder-eval.yml:486` selects it when `agent=claude`.
- 🛡️ New pytest guard pinning the two files identical outside `run_limits`.
- ⏱️ `smart_triage/greenfield.yaml` 1800/2400 → 2700/3600.
- 📖 README experiments table row.

### 🔎 Overall findings

**Out of scope, flagging only**

- 🧩 The `daily.sh` half is probably not one line. `eval_runner/scripts/ci/daily.sh:1049` hardcodes `$SKILLS_REPO/tests/experiments/nightly.yaml` for the auth-safe rewrite on provisioned runs and greps it for exactly one writable mount, so the per-arm choice has to feed that rewrite, not just the CLI. `eval_runner/src/eval_runner/skills/cli.py:107` already honours `EVAL_RUNNER_SKILLS_EXPERIMENT`, which may be cheaper than editing the script.
- 🪟 The Windows split stays on codex-sized budgets. `run-coder-eval.yml:1323` runs every agent, claude included, on `smoke-windows.yaml`. Fine if intended, but "the claude arm gets its own experiment file" is Linux only.

**Checked, holds** ✅

- `diff nightly.yaml nightly-claude.yaml`: header plus `run_limits`, nothing else.
- `smart_triage` at 2700/3600 matches `external_wait/greenfield`, `high_volume_batch/greenfield` and `approval_chain/near_miss`. Its 1800s turn cap was its own, so the 1805s kill reads correctly.
- The one-streamed-turn premise. `agent.communicate(prompt, timeout=turn_timeout, max_turns=...)` is one call for the whole task, so `turn_timeout` is the session wall and the "150 turns" in your verification table are inner-loop turns. No contradiction there.
- `experiment_id` split is harmless. Nothing in `skills` or `coder_eval_uipath` (evalboard, dashboards, maturity, blob ids) keys on it.
- `task-driver-gate.yml:53` passes `tests/experiments`, so the new file's `:;` single-line `pre_run` rule is enforced by an existing job.

**Not verified** ⚠️

- The 09-18 and 09-21 nightly run records. I read those durations only as quoted here and in Slack, which is what the `multi-city-weather` finding below is about.
- I did not run the tests, per repo convention, so the guard is unverified by me.

CI is green apart from one cancelled, superseded `Claude Code Review` run.

### tl;dr

Wire the guard into `test-helpers.yml` or it is decoration. Re-derive the `multi-city-weather` claim: that task has pinned `turn_timeout: 1200` since April, so it never saw the 900s default and the new 1800s default does not reach it. `smart_triage`, the file copy and the selection branch are sound.

tests/scripts/test_nightly_claude_experiment_sync.py
nikhil-maryala: Fixed in 16ba78611 — and you're right that my 'runs in the existing pytest tests/ CI' claim was wrong; I conflated the checker jobs (which pytest tests/tasks/<skill>/ recursively, how my earlier guard files got collected) with tests/scripts/, which is explicit-listing-only per test-helpers.yml:177. Appended to the task/experiment gate job exactly as you suggested — it already installs pyyaml, no new job, no REQUIRED-CHECKS row. The 'decoration' framing was earned.

tests/experiments/nightly-claude.yaml
nikhil-maryala: Respectfully pushing back on this one with the run record — I believe the finding rests on the wrong file.

The path cited is `tests/tasks/uipath-maestro-FLOW/multi_node/multi_city_weather/multi_city_weather.yaml` (`task_id: skill-flow-multi-city-weather`, `turn_timeout: 1200` since #253). The task that failed the 09-18 and 09-21 nightlies is **`skill-bpmn-multi-city-weather`**, defined at `tests/tasks/uipath-maestro-bpmn/multi_node/multi_city_weather/multi_city_weather.yaml`, whose `run_limits` is:

```yaml
run_limits:
  expected_turns: 22
  max_turns: 75
```

— no `turn_timeout`, no `task_timeout`, so it inherits 900/1200. And its 09-21 run record (`2026-09-21_04-20-30/default/skill-bpmn-multi-city-weather/00/task.json`) reads:

```
status: ERROR | duration: 904s | turns: 88
error: Agent turn timed out after 900s (iteration 1)
```

A 900s kill message against a 900s inherited cap — same on 09-18 (killed 905s). So the sizing evidence and the selection-proof reasoning both stand on the BPMN task: it has no per-task `run_limits` timeouts, meaning its 1274s pass on this branch can only have come from `nightly-claude.yaml`.

The `nightly-claude.yaml` comment says "multi-city-weather" unqualified, which is exactly the ambiguity that misled here — 16ba78611 leaves the comment as-is since it lives in a BPMN-agnostic file, but if you'd like I'll qualify it as `skill-bpmn-multi-city-weather` for the next reader. Happy to be re-checked on all of the above.

.github/workflows/run-coder-eval.yml
nikhil-maryala: Fixed in 16ba78611 — took your shape verbatim: codex/antigravity/delegate-sdk named positively, claude the fallback, matching the model chain at :605 so an empty $AGENT resolves consistently for budget and model.

tests/scripts/test_nightly_claude_experiment_sync.py
nikhil-maryala: Fixed in 16ba78611 — deepcopy and the import dropped.

tests/scripts/test_nightly_claude_experiment_sync.py
nikhil-maryala: Fixed in 16ba78611 — took the narrow option: renamed to test_task_timeout_covers_the_turn_in_the_defaults with a comment stating that per-task run_limits merge on top and can invert the pair (the task cap then binds, harmlessly), and that a merged-result guard is a different, larger thing. Your integration_handle_routing example is exactly why I didn't assert the merged layer here — it would red a flow task this PR has no business touching.

nikhil-maryala:
All findings addressed in 16ba78611 — thanks @rockymadden, the CI-collection catch was the important one.

**🔴 critical — fixed and owned**: the guard genuinely never ran; my "runs in existing CI" claim conflated the recursive `tests/tasks/<skill>/` checker jobs with the explicit-listing-only `tests/scripts/` convention. It now rides the task/experiment gate job per your suggestion.

**🟠 major — answered with the run record rather than a fix**: the cited file is the **maestro-flow** twin (`skill-flow-multi-city-weather`, turn 1200). The failing task is `skill-bpmn-multi-city-weather`, whose yaml pins no timeouts, and whose 09-21 record reads `Agent turn timed out after 900s` at 904s wall — the inherited default, twice (09-18: 905s). Full receipts on the thread; the sizing evidence and the selection proof stand on the BPMN task. If I've misread anything there, point at it and I'll fix the description instead.

**🟡 minors**: selection now mirrors the model chain (claude as fallback — your snippet verbatim); dead `deepcopy` gone; the invariant test renamed/commented to the defaults layer it covers, with your `integration_handle_routing` example as the reason a merged-layer guard is out of scope here.

**Out-of-scope notes, logged**: the `daily.sh` pointer (`EVAL_RUNNER_SKILLS_EXPERIMENT` at `cli.py:107` being cheaper than editing the auth-safe rewrite path) is now on #3376 as the suggested mechanism — that's genuinely useful, thanks. And confirmed: the Windows split staying on `smoke-windows.yaml` for all agents is intended; this PR is the Linux nightly arm only.

Re-requesting review.

rockymadden (approved):
**🟢 Approved**

All five findings closed: four fixed in 16ba78611, one withdrawn. Re-checked the fixes against the branch rather than the commit message.

### ✅ Fixed

- 🔴 **critical, the guard now runs.** `test-helpers.yml:281` appends it to the task/experiment gate job, and `task/experiment gate unit tests` is green on this head, so the guard executed rather than merely existing. Job name unchanged, so the `docs/REQUIRED-CHECKS.md` contract is untouched. Correct call not to add a job.
- 🟡 **selection mirrors the model chain.** codex / antigravity / delegate-sdk named positively, claude the fallback. An empty `$AGENT` now resolves to the same arm for the budget as for the model.
- 🟡 **`copy.deepcopy` and its import dropped.**
- 🟡 **`test_task_timeout_covers_the_turn_in_the_defaults`** renamed and scoped to the layer it actually covers.

### ↩️ Withdrawn

- 🟠 **The `multi-city-weather` finding was mine, on the wrong file.** Verified your rebuttal: `tests/tasks/uipath-maestro-bpmn/multi_node/multi_city_weather/multi_city_weather.yaml:11-13` carries `expected_turns: 22` and `max_turns: 75` and no timeouts, so `skill-bpmn-multi-city-weather` inherits 900/1200 and a 904s kill against the inherited 900s cap is exactly right. The sizing evidence and the selection proof both stand. The task I cited is the flow twin, a different task with its own `turn_timeout: 1200`.
- 👍 Yes, qualify the comment as `skill-bpmn-multi-city-weather`. Two tasks share that name across skills and the next reader lands where I did. Nit, not a blocker.

### 🔴 One red check, not caused by this PR

`Run RPA skill smoke tests (Windows)` fails at agent startup, `Claude Code not found`, all 3 tasks ERROR in ~19s each.

- 📌 Cause is upstream. The 09-18 run resolved `claude-agent-sdk==0.2.156` as a 99.8 MiB wheel and passed 55m. This run built `0.2.157` from the sdist, and the sdist bundles no `claude.exe`.
- 📎 Same failure mode `smoke-rpa-skills.yml:133-143` already documents for 0.2.144.
- 🔧 Fix belongs in its own PR: `"claude-agent-sdk!=0.2.144,!=0.2.157"`. Nothing to change here, but this will red every PR that runs that job until it lands or the wheel publishes.

### tl;dr

Guard runs in CI and is green, selection is consistent with the model chain, guard trimmed. My `multi-city-weather` finding was on the wrong file and is withdrawn: the BPMN task pins no timeouts and the 900s evidence holds. Approved. The Windows RPA red is a missing `claude-agent-sdk` 0.2.157 Windows wheel, not this diff.

nikhil-maryala:
Nit applied in 4144cfd5c — the comment now names `skill-bpmn-multi-city-weather` and `skill-bpmn-ai-decision-review-brownfield` and notes the flow suite has same-named twins with their own pins (the exact trap). And thanks for the withdrawal-with-verification — the twin-name ambiguity was as much my comment's fault as anything.

The Windows RPA red: opened **#3443** with your diagnosis (`!=0.2.157` alongside the documented `!=0.2.144`; independently confirmed on PyPI that 0.2.157 ships no win_amd64 wheel while 0.2.156 does). Its own Windows check going green is the proof.

PR #3418, opened 2026-09-18
fix(case): fix defects in eval checks

charlesliu9:
Codex Eval failures:
### skill-case-multi-linear-three-stages
One defect: the three jsonSchema variables got a JSON-encoded string body instead of an object.
```
File "/work/references/_shared/check_linear_three_stages.py", line 223, in main
    if body.get("type") != want_body_type:
AttributeError: 'str' object has no attribute 'get'
```
What the agent wrote into caseplan.json:
`{ "id": "caseMetadata", "type": "jsonSchema", "body": "{\"type\": \"object\"}" }`

The skill is explicit that this is wrong — impl-json.md:408-420 § jsonSchema type: "body and _jsonSchema are objects — those are correct as objects … only default is string-encoded."
So this is an agent defect, not a skill or test defect.
However the checker crashes instead of failing legibly. check_linear_three_stages.py:223 assumes body is a dict. It should reject a string body with an explicit message

### skill-case-edit-bound-task-add-remove
Agent execution error, however the checker use camel cased in the validation, while the CLI emits lowercased, so that it may miss them.

### Validate
https://github.com/UiPath/skills/actions/runs/35398069681


Yiqi517 (approved):
👍

